PR #1549, opened 2016-03-03
stop CommandLineParser from recognizing abbreviations

### LINKED ISSUE #1347: Disable argument abbreviation in joptsimple
Argument abbreviation is causing confusing issues like https://github.com/broadinstitute/gatk/issues/1306, where `-V` means different things in different tools. Let's disable it in `joptsimple` (or, if there's no mechanism to disable it, ask the author to add one).

lbergelson:
previously CommandLineParser would accept non-ambiguous abbreviations for longopts
this caused issues where -v would have different meanings in different tools

I patched joptSimple to add an option to disallow abbreviations, and updated to 5.0-beta-1 which incorporates the necessary changes.
fixes #1347

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/tools/picard/sam/FilterReadsIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/picard/sam/FilterReadsIntegrationTest.java
@@ -112,7 +112,7 @@ public final class FilterReadsIntegrationTest extends CommandLineProgramTest {
         }
 
         if (!writeReadsFile) { // defaults to true
-            args.add("--WRITE_READS_FILE");
+            args.add("--WRITE_READS_FILES");
             args.add("false");
         }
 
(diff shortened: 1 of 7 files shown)


lbergelson:
hmn, looks like tests are failing because some of them were using abbreviations, will fix and then reassign when they're fixed.

lbergelson:
@akiezun fixed the tests

src/test/java/org/broadinstitute/hellbender/tools/picard/sam/FilterReadsIntegrationTest.java
```diff
@@ -112,8 +112,7 @@ public void testReadFilter(
         }
 
         if (!writeReadsFile) { // defaults to true
```
akiezun: Something bogus here - the condition says `if not writeReadsFile then use --WRITE_READS_FILES`

akiezun:
looks fine, but 2 more things before merge:
- fix the bogus things about `writeReadsFile` in `FilterReadsIntegrationTest.java`
- add a test that shows that we do not accept abbreviations, ie that the PR does what it claim "stop CommandLineParser from recognizing abbreviations"

PR #1551, opened 2016-03-04
Upgrade htsjdk to 2.1.1.

cmnbroad:
htsjdk versions older than 2.1.1  would remove NM and MD tags on bam->cram compression, and then automatically regenerate NM tags when reading cram. Starting with 2.1.1, in order to ensure lossless round-tripping, it no longer does either, and restores only the tags present in the compressed file .

As a result, any cramfile read with 2.1.1+ that was generated with older htsjdk versions (or samtools) will fail validation due to missing NM tags. So this PR contains an updated cram file that contains NM tags for the SAMFileValidation tests.


droazen:
:+1:

Issue #1087, opened 2015-11-05
Make CRAM file comparison non-lenient in PrintReadsIntegrationTest


akiezun:
due to a CRAM bug https://github.com/samtools/htsjdk/issues/364 we have to make a few tests lenient.

make them stringent when the cram bug is resolved

cmnbroad:
Once #1084 is merged, AddOrReplaceReadGroupsIntegrationTest has two CRAM tests that are commented out due to this same problem.

akiezun:
#1084 is merged but 2 tests in PrintReadsIntegrationTest are still using lenient comparison because NM tag gets lost (curiously spark version of printreads works fine)

akiezun:
@cmnbroad for you - since you're collecting all things cram

cmnbroad:
The missing NM tags are due to https://github.com/samtools/htsjdk/issues/455. PrintReadsSpark doesn't have this problem now because the tags are lost when going from BAM->CRAM, and the spark tools can't write CRAMs (yet). Once CRAM writing [is enabled](https://github.com/broadinstitute/gatk/tree/cn_spark_write_cram) the same problem happens on Spark.

PR #1561, opened 2016-03-08
Use non-lenient comparison for CRAM.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/1087 now that https://github.com/samtools/htsjdk/pull/487/files is in.


akiezun:
:+1:

Issue #1562, opened 2016-03-08
IntegrationTestSpec doesn't handle CRAM output


cmnbroad:
There are a few problems with IntegrationTestSpec that surface when adding CRAM tests to the Reads/BQSRSparkPipelines:

- generated output filenames contain no sam extension, so outputs always are treated as .bam
- it doesn't have explicit knowledge of the reference file, which is needed to do proper  file comparisons through the SamAssertionUtils assertSamsEqual methods
- there is code that assumes that any expectedFile that doesn't end in ".bam" should use text comparison

We could fix these, but I'm not sure what the incremental value-added of this class is when we can just use TestNG for expected exceptions, etc.; it might make more sense to just eliminate this style of test completely.

cmnbroad:
Closing this based on discussion with @droazen.  We'll keep the code around for easy porting/migration of tests, otherwise we should minimize dependency on it.

PR #1564, opened 2016-03-09
ApplyBQSR CRAM tests.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/1226.

We should resolve whether or not we want to invest in fixing the issues with IntegrationTestSpec (see https://github.com/broadinstitute/gatk/issues/1562) before we do any more of these, since now that we can write CRAMs on SPARK we can also add CRAM tests to ApplyBQSRSpark as well, we just need to decide which way to write the tests.


src/test/java/org/broadinstitute/hellbender/tools/walkers/bqsr/ApplyBQSRIntegrationTest.java
```diff
@@ -15,60 +16,90 @@
...
+
+        //CRAM - input and output crams generated by direct convertsion of the corresponding BAM test files with samtools 1.3
+        tests.add(new Object[]{new ABQSRTest(hiSeqCram, hg18Reference, ".cram", new String[] {"--disableSequenceDictionaryValidation", "true"}, resourceDir + "expected.HiSeq.1mb.1RG.2k_lines.alternate.recalibrated.DIQ.cram")});
+        tests.add(new Object[]{new ABQSRTest(hiSeqCramAligned, hg18Reference, ".cram", new String[] {"-qq", "6", "--disableSequenceDictionaryValidation", "true"}, resourceDir + "expected.HiSeq.1mb.1RG.2k_lines.alternate_allaligned.recalibrated.DIQ.qq6.cram")});
 
         return tests.toArray(new Object[][]{});
     }
 
     @Test(dataProvider = "ApplyBQSRTest")
     public void testPR(ABQSRTest params) throws IOException {
```
droazen: May want to rename this test case

droazen:
:+1: merge when ready

Issue #1560, opened 2016-03-08
spark fails on 'invalid read'


akiezun:
the invalid reads strikes back - i got this when running the ReadsPipelineSpark on qurynamesorted file `hdfs:///user/akiezun/data/CEUTrio.HiSeq.WEx.b37.NA12892.readnamesort.bam`:

```
Job aborted due to stage failure: Task 47 in stage 2.0 failed 4 times, most recent failure: Lost task 47.3 in stage 2.0 (TID 680, dataflow05.broadinstitute.org): java.lang.IllegalArgumentException: 
Invalid interval. Contig:20 start:62720124 end:62720123
	at org.broadinstitute.hellbender.utils.SimpleInterval.validatePositions(SimpleInterval.java:59)
	at org.broadinstitute.hellbender.utils.SimpleInterval.<init>(SimpleInterval.java:34)
	at org.broadinstitute.hellbender.utils.SimpleInterval.<init>(SimpleInterval.java:46)
	at org.broadinstitute.hellbender.engine.spark.BroadcastJoinReadsWithVariants.lambda$join$3d1c3858$1(BroadcastJoinReadsWithVariants.java:27)
	at org.apache.spark.api.java.JavaPairRDD$$anonfun$pairFunToScalaFun$1.apply(JavaPairRDD.scala:1030)
	at org.apache.spark.api.java.JavaPairRDD$$anonfun$pairFunToScalaFun$1.apply(JavaPairRDD.scala:1030)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:328)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:328)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:328)
	at scala.collection.convert.Wrappers$IteratorWrapper.next(Wrappers.scala:30)
	at org.broadinstitute.hellbender.tools.spark.transforms.BaseRecalibratorSparkFn.lambda$apply$26a6df3e$1(BaseRecalibratorSparkFn.java:28)
	at org.apache.spark.api.java.JavaRDDLike$$anonfun$fn$4$1.apply(JavaRDDLike.scala:156)
	at org.apache.spark.api.java.JavaRDDLike$$anonfun$fn$4$1.apply(JavaRDDLike.scala:156)
	at org.apache.spark.rdd.RDD$$anonfun$mapPartitions$1$$anonfun$apply$17.apply(RDD.scala:706)
[...]

PR #1565, opened 2016-03-09
fix 2 bugs in baserecalibrator spark

akiezun:
fixes https://github.com/broadinstitute/gatk/issues/1560 
and another bug that put the whole readgroups into base recalibrator tables (it went undetected because it only manifests itself when bam file header has more readgroups than the actual file has reads from)

for @lbergelson


src/main/java/org/broadinstitute/hellbender/tools/spark/BaseRecalibratorSpark.java
```diff
@@ -93,7 +92,7 @@ protected void runTool( JavaSparkContext ctx ) {
         final RecalibrationReport bqsrReport = BaseRecalibratorSparkFn.apply(rddReadContext, getHeaderForReads(), getReferenceSequenceDictionary(), bqsrArgs);
 
         try ( final PrintStream reportStream = new PrintStream(BucketUtils.createFile(outputTablesPath, getAuthenticatedGCSOptions())) ) {
-            RecalUtils.outputRecalibrationReport(reportStream, bqsrArgs, bqsrReport.getQuantizationInfo(), bqsrReport.getRecalibrationTables(), new StandardCovariateList(bqsrArgs, getHeaderForReads()));
+            RecalUtils.outputRecalibrationReport(reportStream, bqsrArgs, bqsrReport.getQuantizationInfo(), bqsrReport.getRecalibrationTables(), bqsrReport.getCovariates());
```
akiezun: this is the fix for the other bug - the read groups were incorrectly passes in here which resulted in a bogus table

lbergelson: How come we didn't notice it producing bogus tables before?

lbergelson:
:+1: Curious about the "bogus table" though.

PR #1567, opened 2016-03-10
HaplotypeCaller: first runnable version

droazen:
-Ported the remaining pieces of the HaplotypeCaller and assembled them
 into a runnable tool. Fixed many bugs in our ported code in the process
 of doing so.

-Extracted a separate "engine" that does all the work of the HC and is
 separate from the runnable walker.

-Added a new walker class, ReadWindowWalker, as a prospective replacement
 for active region traversal.

-Hooked up the native VECTOR_LOGLESS_CACHING PairHmm to the HaplotypeCaller,
 and activated it by default (this speeds up performance by ~3x in my tests).
 Also added a fallback mode when AVX is not present.


droazen:
For @akiezun to review. Let me know if you want to do an in-person review to expedite things.

Note that I am still working on the HC integration tests, and will add these during the review.

droazen:
I've tested this version in both GVCF and non-GVCF mode, and with and without -L intervals, and it seems to work reasonably. The output, as far as I can tell, is very close to the GATK3 output, but not an exact match. Looking in IGV I suspect that we might have a few boundary artifacts from the new traversal, but I'll have to do more work to confirm this. I've made an effort in the past few days to get the annotations, filtering, etc., closer to GATK3, and this has helped a lot.

Hooking up the native `PairHmm` resulted in a ~3x speedup and hopefully made us more competitive with GATK3 in performance -- I'll post some benchmarking numbers here later.

The `isActive()` machinery was ported, but is not currently used (ie., we consider all sites as active for now). At one point I attempted a standalone tool to do the `isActive()` step, but the overhead of going through the reads twice and writing down a huge sites-level file with the results from the individual loci caused me to abandon that idea. Still, we may want to experiment more with a streamlined version of isActive() in the future if we can't get the performance we want from the current implementation. `ReadWindowWalker` has an entry point to plug in a future `isActive()` check if we choose to add one.

src/main/java/org/broadinstitute/hellbender/engine/ReadWindow.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * The reads are lazily loaded by default (when accessing the reads via {@link #iterator}. Loading all the
+ * reads in the window at once is possible via {@link #loadAllReads}.
+ *
+ * The reads returned will overlap the expanded padded window. It's possible to query whether they are within
+ * the main part of the window via {@link #isContainedWithinWindow} and {@link #startsWithinWindow}.
+ *
+ * The reads in the window can be filtered via {@link #setReadFilter} (no filtering is performed by default).
+ */
+public final class ReadWindow implements Iterable<GATKRead>, Locatable {
```
akiezun: AssemblyRegion is a similar abstraction and we should unify them somehow

src/main/java/org/broadinstitute/hellbender/engine/ReadWindow.java
```diff
@@ -0,0 +1,102 @@
...
+        if ( ! paddedWindow.contains(window) ) {
+            throw new IllegalArgumentException("The padded window must contain the window");
+        }
+
+        this.window = window;
+        this.paddedWindow = paddedWindow;
+        this.readsSource = readsSource;
+    }
+
+    public void setReadFilter( final ReadFilter filter ) {
```
akiezun: document.
Also, is null allowed? If not, make it fail

src/main/java/org/broadinstitute/hellbender/engine/ReadWindow.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public Iterator<GATKRead> iterator() {
+        final Iterator<GATKRead> readsIterator = readsSource.query(paddedWindow);
+
+        return readFilter != null ? new ReadFilteringIterator(readsIterator, readFilter) : readsIterator;
+    }
+
+    public List<GATKRead> loadAllReads() {
```
akiezun: document - what is this method and how it's different from iterator()?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/ConcurrentAFCalculatorProvider.java
```diff
@@ -0,0 +1,38 @@
+package org.broadinstitute.hellbender.tools.walkers.genotyper.afcalc;
+
+import htsjdk.variant.variantcontext.VariantContext;
+
+/**
+ * Produces independent AF calculators per thread.
```
akiezun: threads?

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -43,6 +44,68 @@
...
+    /**
+     * A utility class that computes on the fly average and standard deviation for a stream of numbers.
+     * The number of observations does not have to be known in advance, and can be also very big (so that
+     * it could overflow any naive summation-based scheme or cause loss of precision).
+     * Instead, adding a new number <code>observed</code>
+     * to a sample with <code>add(observed)</code> immediately updates the instance of this object so that
+     * it contains correct mean and standard deviation for all the numbers seen so far. Source: Knuth, vol.2
+     * (see also e.g. http://www.johndcook.com/standard_deviation.html for online reference).
+     */
+    public static class RunningAverage {
```
akiezun: use and Apache class or StatCounter from spark

src/main/java/org/broadinstitute/hellbender/engine/ReadWindow.java
droazen: They are deliberately separate abstractions, since they serve different purposes. `ReadWindow` is supposed to be the unit of eventual parallelism for all `ReadWindowWalkers`, while `AssemblyRegion` is a work unit for the HC specifically. We may eventually settle on a strategy involving creating multiple `AssemblyRegions` per `ReadWindow`, instead of the current 1:1 correspondence.

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
```diff
@@ -35,7 +36,18 @@
         /* PairHMM as implemented for the UnifiedGenotyper. Uses log10 sum functions accurate to only 1E-4 */
         ORIGINAL(() -> new Log10PairHMM(false)),
         /* Optimized version of the PairHMM which caches per-read computations and operations in real space to avoid costly sums of log10'ed likelihoods */
-        LOGLESS_CACHING(() -> new LoglessPairHMM());
+        LOGLESS_CACHING(() -> new LoglessPairHMM()),
+        /* Optimized AVX implementation of LOGLESS_CACHING called through JNI */
+        VECTOR_LOGLESS_CACHING(() -> {
```
lbergelson: As we discussed before, I think we should have a `FASTEST_AVAILABLE` option that tries to use the native code and falls back to the pure java if avx fails. Specifying AVX if it isn't available should fail so that people can detect problems with their systems.

droazen: Done and done

src/main/java/org/broadinstitute/hellbender/engine/ReadWindow.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/engine/ReadWindow.java
droazen: Done

lbergelson:
To review:

- HaplotypeCallerEngine
- GenotypeEngine
- Annotations
- SlidingWindowWalker

src/main/java/org/broadinstitute/hellbender/engine/ReadWindowWalker.java
```diff
@@ -0,0 +1,198 @@
...
+     * @param intervals unmodified intervals for traversal
+     * @param windowSize desired window size; intervals larger than this will be sharded into windows up to this size
+     * @param windowPadding desired window padding; each windowed interval will be padded on both sides by this number of bases
+     * @param readsSource data source for reads
+     * @param dictionary sequence dictionary for reads
+     * @return List of {@link ReadWindow} objects, sharded and padded as necessary
+     */
+    @VisibleForTesting
+    protected static final List<ReadWindow> makeWindows( final List<SimpleInterval> intervals, final int windowSize, final int windowPadding, final ReadsDataSource readsSource, final SAMSequenceDictionary dictionary ) {
+        final List<ReadWindow> windows = new ArrayList<>();
```
akiezun: initialize with capacity

src/main/java/org/broadinstitute/hellbender/engine/ReadWindowWalker.java
```diff
@@ -0,0 +1,198 @@
...
+     * @param dictionary sequence dictionary for reads
+     * @return List of {@link ReadWindow} objects, sharded and padded as necessary
+     */
+    @VisibleForTesting
+    protected static List<ReadWindow> shardInterval( final SimpleInterval interval, final int windowSize, final int windowPadding, final ReadsDataSource readsSource, final SAMSequenceDictionary dictionary ) {
+        if ( ! IntervalUtils.intervalIsOnDictionaryContig(interval, dictionary) ) {
+            throw new IllegalArgumentException("Interval " + interval + " not within the bounds of a contig in the provided dictionary");
+        }
+
+        final List<ReadWindow> shards = new ArrayList<>();
```
akiezun: capacity

src/main/java/org/broadinstitute/hellbender/engine/ReadWindowWalker.java
```diff
@@ -0,0 +1,198 @@
...
+     * Subclasses can override to provide their own filters
+     * Multiple filters can be composed by using {@link org.broadinstitute.hellbender.engine.filters.ReadFilter} composition methods.
+     */
+    public CountingReadFilter makeReadFilter(){
+        return new CountingReadFilter("Wellformed", new WellformedReadFilter(getHeaderForReads()))
+                .and(new CountingReadFilter("Mapped", ReadFilterLibrary.MAPPED));
+    }
+
+    @Override
+    public void traverse() {
```
akiezun: is this method intentionally left not final? If yes, describe what the responsibilities of subclasses are. Otherwise, make final

akiezun: I suggest making traverse final and documenting that it calls `apply` for every window

src/main/java/org/broadinstitute/hellbender/engine/ReadWindowWalker.java
```diff
@@ -0,0 +1,198 @@
...
+
+    @Override
+    public boolean requiresReads() {
+        return true;
+    }
+
+    /**
+     * @return Default value for the {@link #windowSize} parameter, if none is provided on the command line
+     */
+    protected abstract int defaultWindowSize();
```
akiezun: i find these 2 methods (`defaultWindowSize` and `defaultWindowPadding`) misplaced here. If this abstract class does not use them and only the HC subclass does, then they belong in the subclass and not here.

src/main/java/org/broadinstitute/hellbender/engine/ReferenceContext.java
```diff
@@ -90,6 +90,37 @@ public ReferenceContext( final ReferenceDataSource dataSource, final SimpleInter
...
+     * Window boundaries are cropped at contig boundaries, if necessary.
+     *
+     * @param dataSource backing reference data source (may be null if there is no reference)
+     * @param interval our location on the reference (may be null if our location is unknown)
+     * @param window the expanded location on the reference (may be null if our location is unknown). Must be null if interval is null.
+     *               Must contain interval.
+     */
+    public ReferenceContext(final ReferenceDataSource dataSource, final SimpleInterval interval, final SimpleInterval window) {
+        this.dataSource = dataSource;
+        cachedSequence = null;
```
akiezun: change to `this.cachedSequence` so that all lines read the same

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/DepthPerSampleHC.java
```diff
@@ -0,0 +1,95 @@
...
+ * </ul>
+ *
+ * <h3>Related annotations</h3>
+ * <ul>
+ *     <li><b><a href="https://www.broadinstitute.org/gatk/guide/tooldocs/org_broadinstitute_gatk_tools_walkers_annotator_DepthPerAlleleBySample.php">DepthPerAlleleBySample</a></b> calculates depth of coverage for each allele per sample (AD).</li>
+ *     <li><b><a href="https://www.broadinstitute.org/gatk/guide/tooldocs/org_broadinstitute_gatk_tools_walkers_annotator_Coverage.php">Coverage</a></b> gives the filtered depth of coverage for each sample and the unfiltered depth across all samples.</li>
+ * </ul>
+ *
+ */
+public class DepthPerSampleHC extends GenotypeAnnotation implements StandardHCAnnotation {
```
akiezun: make final

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandBiasBySample.java
```diff
@@ -0,0 +1,80 @@
...
+ *     <li>This annotation can only be generated by HaplotypeCaller (it will not work when called from VariantAnnotator).</li>
+ * </ul>
+ *
+ * <h3>Related annotations</h3>
+ * <ul>
+ *     <li><b><a href="https://www.broadinstitute.org/gatk/guide/tooldocs/org_broadinstitute_gatk_tools_walkers_annotator_FisherStrand.php">FisherStrand</a></b> uses Fisher's Exact Test to evaluate strand bias.</li>
+ *     <li><b><a href="https://www.broadinstitute.org/gatk/guide/tooldocs/org_broadinstitute_gatk_tools_walkers_annotator_StrandOddsRatio.php">StrandOddsRatio</a></b> is an updated form of FisherStrand that uses a symmetric odds ratio calculation.</li>
+ * </ul>
+ */
+public class StrandBiasBySample extends GenotypeAnnotation {
```
akiezun: make final

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandBiasTableUtils.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.LogManager;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A class containing convenience methods used in the strand bias annotation calculations
+ */
+public class StrandBiasTableUtils {
```
akiezun: make final and add a private constructor

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandBiasTableUtils.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private final static Logger logger = LogManager.getLogger(StrandBiasTableUtils.class);
+
+    //For now this is only for 2x2 contingency tables
+    protected static final int ARRAY_DIM = 2;
```
akiezun: why are those protected and not private? I find those two consts kind of goofy anyway, given that the getContingencyArray clearly hardwires everything to 2x2

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandBiasTest.java
```diff
@@ -172,6 +184,21 @@ protected static boolean passesMinimumThreshold(final int[] data, final int minC
     }
 
     /**
+     * Helper function to parse the genotype annotation into the SB annotation array
+     * @param arrayList the ArrayList returned from StrandBiasBySample.annotate()
+     * @return the array used by the per-sample Strand Bias annotation
+     */
+    private static int[] encodeSBBS( final List<Integer> arrayList ) {
```
akiezun: why not move this to the StrandBiasUtils class? or move all that code here? Right now it's not clear what belongs where

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/DiploidSNPGenotypeLikelihoods.java
```diff
@@ -0,0 +1,453 @@
...
+ *
+ * for each of the 10 unique diploid genotypes AA, AC, AG, .., TT
+ *
+ * Everything is stored as arrays indexed by DiploidGenotype.ordinal() values in log10 space.
+ *
+ * The priors contain the relative probabilities of each genotype, and must be provided at object creation.
+ * From then on, you can call any of the add() routines to update the likelihoods and posteriors in the above
+ * model.
+ */
+public class DiploidSNPGenotypeLikelihoods implements Cloneable {
```
akiezun: make final

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/DiploidSNPGenotypeLikelihoods.java
```diff
@@ -0,0 +1,453 @@
...
+        log10_1_minus_PCR_error = log10(1.0 - PCR_error_rate);
+        setToZero();
+    }
+
+    /**
+     * Cloning of the object
+     * @return clone
+     * @throws CloneNotSupportedException
+     */
+    protected Object clone() throws CloneNotSupportedException {
```
akiezun: can we make it not Clonable and copy by more civilized means?

src/main/java/org/broadinstitute/hellbender/engine/ReadWindowWalker.java
droazen: Have a closer look at the class -- they are used here for a very important purpose (to give the arguments in question tool-defined default values!)

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/DiploidSNPGenotypeLikelihoods.java
```diff
@@ -0,0 +1,453 @@
...
+                throw new RuntimeException(e);
+            else
+                return false;
+        }
+
+        return true;
+    }
+
+    //
+    // Constant static data
```
akiezun: move all this static stuff to the top of the class

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/DiploidSNPGenotypeLikelihoods.java
```diff
@@ -0,0 +1,453 @@
...
+    //
+    private final static double[] genotypeZeros = new double[DiploidGenotype.values().length];
+    private final static double[] baseZeros = new double[BaseUtils.BASES.length];
+
+    static {
+        for ( DiploidGenotype g : DiploidGenotype.values() ) {
+            genotypeZeros[g.ordinal()] = 0.0;
```
akiezun: does this just set all values to 0.0? if yes, then it's not needed.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/DiploidSNPGenotypeLikelihoods.java
```diff
@@ -0,0 +1,453 @@
...
+        }
+        for ( byte base : BaseUtils.BASES ) {
+            baseZeros[BaseUtils.simpleBaseToBaseIndex(base)] = 0.0;
```
akiezun: same here - what's the purpose of this loop?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/MinimalGenotypingEngine.java
```diff
@@ -0,0 +1,54 @@
...
+import htsjdk.variant.variantcontext.Allele;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.tools.walkers.genotyper.afcalc.AFCalculatorProvider;
+import org.broadinstitute.hellbender.utils.genotyper.SampleList;
+
+
+/**
+ * A stripped-down version of the former UnifiedGenotyper's genotyping strategy implementation,
+ * used only by the HaplotypeCaller for its isActive() determination.
+ */
```
akiezun: make class final and document that it is not supposed to be used for actual genotyping. Maybe it should become a package-private class or even private inside of HaplotypeCallerEngine?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/SNPGenotypeLikelihoodsCalculationModel.java
```diff
@@ -0,0 +1,238 @@
...
+import org.broadinstitute.hellbender.utils.pileup.PileupElement;
+import org.broadinstitute.hellbender.utils.pileup.ReadPileup;
+import org.broadinstitute.hellbender.utils.variant.GATKVCFConstants;
+import org.broadinstitute.hellbender.utils.variant.GATKVariantContextUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class SNPGenotypeLikelihoodsCalculationModel extends GenotypeLikelihoodsCalculationModel {
```
akiezun: i think class is dead. lets delete it until we need it

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/VariantCallContext.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Useful helper class to communicate the results of calculateGenotype to framework
+ */
+public final class VariantCallContext extends VariantContext {
+
+    private static final long serialVersionUID = 1L;
+
+    // Was the site called confidently, either reference or variant?
+    public boolean confidentlyCalled = false;
```
akiezun: make final

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/VariantCallContext.java
```diff
@@ -0,0 +1,50 @@
...
+
+    // Should this site be emitted?
+    public boolean shouldEmit = true;
```
akiezun: make field final or encapsulate

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/FixedAFCalculatorProvider.java
```diff
@@ -97,4 +97,25 @@ public AFCalculator getInstance(final int ploidy, final int maxAltAlleleCount) {
...
+     * Creates a fixed AF calculator provider that is thread safe.
+     *
+     * @param config the caller configuration.
+     *
+     * @throws IllegalArgumentException if any of the input argument is {@code null} or contain invalid configuration
+     *   like zero-samples, zero or negative ploidy or negative-zero maximum number of alleles.
+     *
+     * @return never {@code null}
+     */
+    public static AFCalculatorProvider createThreadSafeProvider( final StandardCallerArgumentCollection config ) {
```
akiezun: why do we need a thread safe anything here?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
```diff
@@ -0,0 +1,244 @@
...
+ *     to the desired levels in the next step of the workflow (GenotypeGVCFs)</li>
+ * </ul>
+ *
+ */
+@CommandLineProgramProperties(
+        summary = "Call germline SNPs and indels via local re-assembly of haplotypes",
+        oneLineSummary = "Call germline SNPs and indels via local re-assembly of haplotypes",
+        programGroup = VariantProgramGroup.class
+)
+public class HaplotypeCaller extends ReadWindowWalker {
```
akiezun: make class final or document and test how it's supposed to be extended

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
```diff
@@ -0,0 +1,244 @@
...
+        hcEngine = new HaplotypeCallerEngine(hcArgs, getHeaderForReads(), referenceArguments.getReferenceFileName());
+
+        // The HC engine will make the right kind (VCF or GVCF) of writer for us
+        vcfWriter = hcEngine.makeVCFWriter(outputVCF, getHeaderForReads().getSequenceDictionary());
+        hcEngine.writeHeader(vcfWriter);
+    }
+
+    @Override
+    public void apply( ReadWindow window, ReferenceContext referenceContext, FeatureContext featureContext ) {
+        for ( final AssemblyRegion assemblyRegion : readWindowToAssemblyRegions(window) ) {
```
akiezun: validate arguments non null

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * Convert a {@link ReadWindow} into a List of {@link AssemblyRegion} for processing by the HaplotypeCaller,
+     * covering the span of the original window
+     *
+     * @param window window to convert into AssemblyRegions
+     * @return a List of {@link AssemblyRegion} for processing by the HaplotypeCaller, covering the span of
+     *         the original window
+     */
+    private List<AssemblyRegion> readWindowToAssemblyRegions( final ReadWindow window ) {
+        List<AssemblyRegion> regions = new ArrayList<>();
```
akiezun: this list is always of size 1 - just return Collections.singletonList instead of making a array list and adding an element etc

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
```diff
@@ -0,0 +1,204 @@
...
+    @Argument(fullName = "captureAssemblyFailureBAM", shortName = "captureAssemblyFailureBAM", doc = "Write a BAM called assemblyFailure.bam capturing all of the reads that were in the active region when the assembler failed for any reason", optional = true)
+    public boolean captureAssemblyFailureBAM = false;
+
+    // Parameters to control read error correction
+    /**
+     * Enabling this argument may cause fundamental problems with the assembly graph itself.
+     */
+    @Hidden
+    @Argument(fullName = "errorCorrectReads", shortName = "errorCorrectReads", doc = "Use an exploratory algorithm to error correct the kmers used during assembly", optional = true)
+    public boolean errorCorrectReads = false;
```
akiezun: seems like this argument should be killed

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
droazen: It's only always size 1 for now -- the 1:1 window to region correspondence is likely a temporary thing.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+ * The core engine for the HaplotypeCaller that does all of the actual work of the tool.
+ *
+ * Usage:
+ * -Pass the HaplotypeCaller args into the constructor, which will initialize the HC engine completely.
+ * -Get the appropriate VCF or GVCF writer (depending on our arguments) from {@link #makeVCFWriter}
+ * -Write the appropriate VCF header via {@link #writeHeader}
+ * -Repeatedly call {@link #callRegion} to call variants in each region, and add them to your writer
+ * -When done, call {@link #shutdown}. Close the writer you got from {@link #makeVCFWriter} yourself.
+ */
+public class HaplotypeCallerEngine {
```
akiezun: make final or document and test for inheritance

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+    private ReferenceConfidenceModel referenceConfidenceModel = null;
+
+    private double log10GlobalReadMismappingRate;
+
+    private AssemblyRegionTrimmer trimmer = new AssemblyRegionTrimmer();
+
+    // the genotyping engine for the isActive() determination
+    private MinimalGenotypingEngine activeRegionEvaluationGenotyperEngine = null;
+
+    // the assembly engine
```
akiezun: not useful comment - redundant with field name

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+    private ReadThreadingAssembler assemblyEngine = null;
+
+    // the likelihoods engine
+    private ReadLikelihoodCalculationEngine likelihoodCalculationEngine = null;
```
akiezun: not useful comment - redundant with field name

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+
+    // the genotyping engine
+    private HaplotypeCallerGenotypingEngine genotypingEngine = null;
```
akiezun: not useful comment - redundant with field name

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+ */
+public class HaplotypeCallerEngine {
+
+    private static final Logger logger = LogManager.getLogger(HaplotypeCallerEngine.class);
+
+    private HaplotypeCallerArgumentCollection hcArgs;
+
+    private SAMFileHeader readsHeader;
+
+    private String reference;
```
akiezun: a comment here would be useful - what kind of 'reference' is this? A full reference string, a path to the file or url, something else?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+ * -Get the appropriate VCF or GVCF writer (depending on our arguments) from {@link #makeVCFWriter}
+ * -Write the appropriate VCF header via {@link #writeHeader}
+ * -Repeatedly call {@link #callRegion} to call variants in each region, and add them to your writer
+ * -When done, call {@link #shutdown}. Close the writer you got from {@link #makeVCFWriter} yourself.
+ */
+public class HaplotypeCallerEngine {
+
+    private static final Logger logger = LogManager.getLogger(HaplotypeCallerEngine.class);
+
+    private HaplotypeCallerArgumentCollection hcArgs;
```
akiezun: hcArgs can  be final

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+
+    private SAMFileHeader readsHeader;
```
akiezun: readsHeader can  be final

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
akiezun: reference can be final

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+import org.broadinstitute.hellbender.utils.read.ReadUtils;
+import org.broadinstitute.hellbender.utils.variant.*;
+import org.broadinstitute.hellbender.utils.variant.writers.GVCFWriter;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.util.*;
+
+/**
+ * The core engine for the HaplotypeCaller that does all of the actual work of the tool.
```
akiezun: Document whether having multiple HCE's is equivalent to using 1 for all `callRegion` calls? Ie how much state is carried over calls to `callRegion`?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+    }
+
+    /**
+     * Create the default read filter for use with the HaplotypeCaller
+     *
+     * @param hcArgs HaplotypeCaller arguments
+     * @param header reads header
+     * @return the default read filter for use with the HaplotypeCaller
+     */
+    public static CountingReadFilter makeStandardHCReadFilter( final HaplotypeCallerArgumentCollection hcArgs, final SAMFileHeader header ) {
```
akiezun: check that arguments not null

akiezun: do you need the whole `HaplotypeCallerArgumentCollection`? Only `MIN_MAPPING_QUALITY_SCORE` is used

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+
+    /**
+     * Create a VCF or GVCF writer as appropriate, given our arguments
+     *
+     * @param outputVCF location to which the vcf should be written
+     * @param readsDictionary sequence dictionary for the reads
+     * @return a VCF or GVCF writer as appropriate, ready to use
+     */
+    public VariantContextWriter makeVCFWriter( final String outputVCF, final SAMSequenceDictionary readsDictionary ) {
+        VariantContextWriter writer = GATKVariantContextUtils.createVCFWriter(new File(outputVCF), readsDictionary, false);
```
akiezun: check that all args are not null

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+        return writer;
+    }
+
+    /**
+     * Writes an appropriate VCF header, given our arguments, to the provided writer
+     *
+     * @param vcfWriter writer to which the header should be written
+     */
+    public void writeHeader( final VariantContextWriter vcfWriter ) {
+        final Set<VCFHeaderLine> headerInfo = new HashSet<>();
```
akiezun: check that all args are not null

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+        headerInfo.add(GATKVCFHeaderLines.getFilterLine(GATKVCFConstants.LOW_QUAL_FILTER_NAME));
+
+        if ( emitReferenceConfidence() ) {
+            headerInfo.addAll(referenceConfidenceModel.getVCFHeaderLines());
+        }
+
+        vcfWriter.writeHeader(new VCFHeader(headerInfo, sampleSet));
+    }
+
+    public ActivityProfileState isActive( final FeatureContext features, final ReferenceContext ref, final AlignmentContext context ) {
```
akiezun: check that all args are not null

akiezun: I wonder if `isActive` (and auxiliary methods) should not go to a different class. That would make it clear that `HaplotypeCallerEngine` is for calling regions and a different thing is for deciding active/non active

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+
+    /**
+     * Generate variant calls for an assembly region
+     *
+     * @param region region to assemble and perform variant calling on
+     * @param features Features overlapping the assembly region
+     * @return List of variants discovered in the region (may be empty)
+     */
+    public List<VariantContext> callRegion( final AssemblyRegion region, final FeatureContext features ) {
+        if ( hcArgs.justDetermineActiveRegions ) {
```
akiezun: check that argument not null

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+                if ( vc.isNotFiltered() ) {
+                    givenAlleles.add(vc); // do something with these VCs during GGA mode
+                }
+            }
+
+            // No alleles found in this region so nothing to do!
+            if ( givenAlleles.isEmpty() ) {
+                return referenceModelForNoVariation(region, true);
+            }
+        } else if( region.size() == 0 ) {
```
akiezun: it seems that this check `if( region.size() == 0 ){ return referenceModelForNoVariation(region, true); }` could me moved up to before the declaration of `givenAlleles`. It would make the code cleaner i think

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+    }
+
+    /**
+     * Generate variant calls for an assembly region
+     *
+     * @param region region to assemble and perform variant calling on
+     * @param features Features overlapping the assembly region
+     * @return List of variants discovered in the region (may be empty)
+     */
+    public List<VariantContext> callRegion( final AssemblyRegion region, final FeatureContext features ) {
```
akiezun: this method is a monster - can you break it up?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+    /**
+     * High-level function that runs the assembler on the given region's reads,
+     * returning a data structure with the resulting information needed
+     * for further HC steps
+     *
+     * @param region the region we should assemble
+     * @param giveAlleles additional alleles we might need to genotype (can be empty)
+     * @return the AssemblyResult describing how to proceed with genotyping
+     */
+    protected AssemblyResultSet assembleReads(final AssemblyRegion region, final List<VariantContext> giveAlleles) {
```
akiezun: why 'protected'?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -0,0 +1,861 @@
...
+    /**
+     * Create an ref model result (ref model or no calls depending on mode) for an active region without any variation
+     * (not is active, or assembled to just ref)
+     *
+     * @param region the region to return a no-variation result
+     * @param needsToBeFinalized should the region be finalized before computing the ref model (should be false if already done)
+     * @return a list of variant contexts (can be empty) to emit for this ref region
+     */
+    private List<VariantContext> referenceModelForNoVariation(final AssemblyRegion region, final boolean needsToBeFinalized) {
+        if ( emitReferenceConfidence() ) {
```
akiezun: flip the if so that the short branch is first. it makes it easier to read

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
droazen: It's easier for the client to pass in the whole arg collection without thinking about which parts need to be extracted (plus this method may use more arguments in the future).

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -0,0 +1,777 @@
...
+    protected static boolean containsVCWithMatchingAlleles(final List<VariantContext> list, final VariantContext vcToTest ) {
+        for( final VariantContext vc : list ) {
+            if( vc.hasSameAllelesAs(vcToTest) ) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    protected static final class Event {
```
akiezun: can this awful class be made private?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HeterogeneousKmerSizeResolution.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Take just the haplotypes from largest kmersize that generated any.
+     */
+    MAX_ONLY,
+
+    /**
+     * Take just the haplotypes from the smallest kmerSize that generated any.
+     */
+    @SuppressWarnings("unused")
```
akiezun: really unused? delete it if yes

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HeterogeneousKmerSizeResolution.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    /**
+     * Indicates whether we should use the minimum kmerSize for the haplotypeGraph or not.
+     *
+     * @return true if we need to use the minimum, false otherwise.
+     */
+    @SuppressWarnings("unused")
+    public boolean useMinimum() {
```
akiezun: delete this - it is really unused. Also unused in gatk3

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/ReadThreadingAssemblerArgumentCollection.java
```diff
@@ -0,0 +1,139 @@
...
+import org.broadinstitute.hellbender.cmdline.Hidden;
+
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Set of arguments related to the {@link org.broadinstitute.hellbender.tools.walkers.haplotypecaller.readthreading.ReadThreadingAssembler}
+ */
+public final class ReadThreadingAssemblerArgumentCollection {
```
akiezun: should this extend `ArgumentCollectionDefinition`?

src/main/java/org/broadinstitute/hellbender/utils/fragments/FragmentUtils.java
```diff
@@ -0,0 +1,85 @@
...
+import htsjdk.samtools.util.QualityUtil;
+import org.apache.commons.lang3.tuple.Pair;
+import org.broadinstitute.hellbender.exceptions.GATKException;
+import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import org.broadinstitute.hellbender.utils.read.ReadUtils;
+
+import java.util.List;
+
+public class FragmentUtils {
```
akiezun: make final and add a private constructor

src/main/java/org/broadinstitute/hellbender/utils/genotyper/ReadLikelihoods.java
```diff
@@ -456,6 +455,27 @@ private BestAllele searchBestAllele(final int sampleIndex, final int readIndex,
         return new BestAllele(sampleIndex,readIndex,bestAlleleIndex,bestLikelihood,secondBestLikelihood);
     }
 
+    public void changeReads(final Map<GATKRead, GATKRead> readRealignments) {
```
akiezun: document what this method does

akiezun:
overall this looks reasonable to me. Adding an integration test is important - at least lock in the variants that are concordant with gatk3 so that we dont lose those while refactoring and fixing etc.

src/main/java/org/broadinstitute/hellbender/engine/ReadWindowWalker.java
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * {@link #windowPadding} controls the number of bases of extra context to fetch on either side of the window. Overlapping reads
+ * are fetched for the padded regions, but they are marked in the {@link ReadWindow} objects as not being within the window proper.
+ *
+ * If multiple sources of reads are specified, they are merged together into a single sorted stream of reads.
+ *
+ * ReadWindowWalker authors must implement the {@link #apply}, {@link #defaultWindowSize}, and {@link #defaultWindowPadding} methods,
+ * and may optionally implement {@link #onTraversalStart} and/or {@link @onTraversalDone}.
+ */
+public abstract class ReadWindowWalker extends GATKTool {
```
akiezun: can you add an example ReadWindowWalker just like we have example walkers of other kinds?

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
```diff
@@ -20,7 +21,7 @@
  */
 public final class VectorLoglessPairHMM extends LoglessPairHMM {
 
-    final static Logger logger = Logger.getLogger(VectorLoglessPairHMM.class);
+    private static final Logger logger = LogManager.getLogger(VectorLoglessPairHMM.class);
     final static Boolean runningOnMac = System.getProperty("os.name", "unknown").toLowerCase().startsWith("mac");
     long threadLocalSetupTimeDiff = 0;
```
akiezun: this is actually thread local?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/DiploidSNPGenotypeLikelihoods.java
```diff
@@ -0,0 +1,453 @@
...
+        return add(obsBase1, qual1, obsBase2, qual2, 1);
+    }
+
+    // -------------------------------------------------------------------------------------
+    //
+    // Dealing with the cache routines
+    //
+    // -------------------------------------------------------------------------------------
+
+    private static ThreadLocal<DiploidSNPGenotypeLikelihoods[][][][][]> CACHE = new ThreadLocal<DiploidSNPGenotypeLikelihoods[][][][][]>() {
```
akiezun: I think all this crazy code should move to a different class. How about `DiploidSNPGenotypeLikelihoodsCache`?

akiezun:
bounce to @droazen

droazen:
Status of branch before today's HaplotypeCaller hack-a-thon:

Our output is currently 97%-99% concordant with GATK3 on our test bam `CEUTrio.HiSeq.WGS.b37.NA12878.20.21.bam`:

    With -ERC GVCF
        GATK3 total:    19910
        GATK4 total:    19916
        Concordant:    19882
        Called in 3 but not 4:       28
        Called in 4 but not 3:       34

    With -ERC NONE:
        GATK3 total:     1014
        GATK4 total:      997
        Concordant:      986
        Called in 3 but not 4:       28
        Called in 4 but not 3:       11

We're about ~15% slower than GATK3 in GVCF mode, and more than twice as slow with -ERC NONE:

    GVCF MODE:
        GATK3: 4.64 min
        GATK4: 5.45 min

    VCF MODE:
        GATK3: 1.73 min
        GATK4: 3.74 min
 
The profile shows plenty of opportunities for optimization, so we should be able to achieve performance parity (in GVCF mode, at least) without too much difficulty.

droazen:
Post hack-a-thon, we're almost ~2x faster than GATK3 in GVCF mode (2.7 min in GATK4 vs. 4.64 min in GATK3) in the `dr_runnable_haplotypecaller_ar` branch, but concordance has dropped from 99% to ~90%. Will have to do some testing to find out why, but the initial results are extremely promising performance-wise, at least. @lbergelson thinks he's found a bug in `SelectVariants` that might be distorting our concordance numbers for GVCFs, as well.

@akiezun and @gspowley also found a bunch of low-hanging-fruit type optimizations that should buy us another 5% or so. It will take some time to merge everything back into this branch and see where we stand overall.

droazen:
This branch is almost in a merge-able state -- comments have been addressed, we now have continuous tests showing 99%+ concordance with GATK 3.4, and we are about twice as fast in GVCF mode. The only remaining holdup is two integration tests that are failing due to slight precision issues/jitter in the QUAL field (possibly due to the native PairHMM). While I work on this final issue, I'll open a separate PR for the pieces of this branch outside of the `haplotypecaller` package itself.

eitanbanks:
Yay!!!!

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -0,0 +1,182 @@
...
+     * Shard an interval into ReadShards using the shardSize and shardPadding arguments
+     *
+     * @param interval interval to shard; must be on the contig according to the provided dictionary
+     * @param shardSize desired shard size; intervals larger than this will be divided into shards of up to this size
+     * @param shardPadding desired shard padding; each shard's interval will be padded on both sides by this number of bases
+     * @param readsSource data source for reads
+     * @param dictionary sequence dictionary for reads
+     * @return List of {@link ReadShard} objects, sharded and padded as necessary
+     */
+    public static List<ReadShard> shardInterval( final SimpleInterval interval, final int shardSize, final int shardPadding, final ReadsDataSource readsSource, final SAMSequenceDictionary dictionary ) {
```
magicDGS: @droazen, could you add another parameter called `shardStep`? It will be useful for #1708, and you can implement another static method that use `shardSize` as `shardSize` for the same behaviour...

Only needs a chek that `shardStep` is not higher than `shardSize` and change line 177 by `start += shardStep`.

droazen:
Rebased this onto the latest master

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/ReferenceConfidenceModel.java
```diff
@@ -0,0 +1,546 @@
...
+     * @param maybeOverlapping a collection of variant contexts that might overlap curPos
+     * @return a VariantContext, or null if none overlaps
+     */
+    @VisibleForTesting
+    VariantContext getOverlappingVariantContext(final Locatable curPos, final Collection<VariantContext> maybeOverlapping) {
+        final SimpleInterval curPosSI = new SimpleInterval(curPos);
+        VariantContext overlaps = null;
+        for ( final VariantContext vc : maybeOverlapping ) {
+            if ( curPosSI.overlaps(vc) ) {
+                if ( overlaps == null || vc.getStart() > overlaps.getStart() ) {
```
sputnikci: [PMD] INFO: These nested if statements could be combined

droazen:
Reviewed in-person with @akiezun today -- we are ready to merge this into gatk-protected as soon as https://github.com/broadinstitute/gatk/pull/1764 is in and we do another snapshot release.

droazen:
Closing in favor of https://github.com/broadinstitute/gatk-protected/pull/452

PR #1573, opened 2016-03-11
ClipReads CRAM tests.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/1228.


src/test/resources/org/broadinstitute/hellbender/tools/expected.clipReadsTestCRAM.QT_10.tmp
```diff
@@ -0,0 +1,11 @@
+--------------------------------------------------------------------------------
```
droazen: Move `ClipReads`-related test files to a `clipreads` subdirectory within `src/test/resources/org/broadinstitute/hellbender/tools/`

src/main/java/org/broadinstitute/hellbender/tools/ClipReads.java
```diff
@@ -460,7 +461,7 @@ private void accumulate(ReadClipperWithData clipper) {
         GATKRead clippedRead = clipper.clipRead(clippingRepresentation);
         if (outputBam != null) {
             outputBam.addRead(clippedRead);
-        } else {
+        } else if (out != null) {
```
droazen: Is there a test that covers the case of `out == null`, and proves that the tool doesn't blow up in that case?

droazen:
Review complete, back to @cmnbroad -- merge after addressing comments

cmnbroad:
The GATK4 version of ClipReads requires an output bam (GATK3 doesn't - it will write the reads to the summary file), but the code wasn't quite handling that change properly, so the code is rationalized for that (fixed an NPE in the case when the STATSOUTPUT arg was null); added tests for that case; moved all the test files into the ClipReads test folder; and removed all the .bai indices which are unused by the tests. Added CRAM tests.

PR #1582, opened 2016-03-16
add JVM arguments to readme

akiezun:


README.md
```diff
@@ -158,6 +158,12 @@ If you are looking for the codebase of the current production version of GATK, p
   on a cluster or the cloud by building a Spark jar with `gradle installSpark` and passing the resulting jar in `build/libs/`
   directly to either `spark-submit` or `gcloud`.
 
+* To pass JVM arguments to GATK (non-spark), use `JAVA_OPTS` like in this example:
```
droazen: And what about with spark tools?

lbergelson: depends how you're running them, JAVA_OPTS will work if you're not using spark-submit, if you're using spark submit you need to use 
```
--conf "spark.executor.extraJavaOptions=-XX:+PrintGCDetails" 
```
and 
```
--driver-java-options "-XX:+PrintGCDetails" 
```

akiezun: given that we recommend using gatk-launch i think spark-submit is not relevant in our readme

akiezun: ok to merge?

lbergelson:
@akiezun This works when you use gatk-launch with sparkRunner SPARK?

lbergelson:
@akiezun I don't think  JAVA_OPTS will work when running with any runner other than DIRECT.  I think it should probably mention that in the doc.  Other than that +1

PR #1555, opened 2016-03-05
utillties needed for CalculateGenotypePosteriors

akiezun:
relatively straightforward port of CGP. I made minimal changes. Test data is sharable and copied from GATK3. Note: this is going to protected for merging before it's usable,

@lbergelson can you review?

PR #1583, opened 2016-03-16
bump to LFS 1.1.1 that fixes a race condition that hurt us

akiezun:
@lbergelson can you look a this?
https://github.com/github/git-lfs/issues/904 is fixed in 1.1.1 so i bumped us to 1.1.1 (i get hit by this bug in https://github.com/broadinstitute/gatk/pull/1555)

### DIFF
--- a/scripts/install_git_lfs.sh
+++ b/scripts/install_git_lfs.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-GIT_LFS_VERSION=v1.1.0
+GIT_LFS_VERSION=v1.1.1
 
 echo "cloning"
 git clone https://github.com/github/git-lfs.git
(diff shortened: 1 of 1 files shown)


lbergelson:
If tests pass :+1:   I was trying to bump to 1.1.2 in https://github.com/broadinstitute/gatk/compare/lb_update_git_lfs_version, but was running into some problems.

akiezun:
nope, this fails. maybe it's because the go version on travis is 1.4 and git lsf 1.1.1 wants 1.6 ? @lbergelson do you know how to bump go version?

lbergelson:
Instead of bumping the go version I think we might do better just installing the prebuilt binary.  I have an alternate branch where I was trying to bump the git-lfs version and encountering the same problems.  I'm trying to just download the latest version and install it now, so we'll see if that works.

akiezun:
ok, closing this then

lbergelson:
I'm pretty sure this is an lfs bug and not something on our end.  https://github.com/github/git-lfs/issues/1090

PR #1584, opened 2016-03-16
Add AVX PairHMM double and single modes and enable FTZ in Java

gspowley:
@droazen 

- Added `AVX_LOGLESS_CACHING_SINGLE` and `AVX_LOGLESS_CACHING_DOUBLE` modes to HaplotypeCaller to enable runtime selection of single-precision or double-precision floating-point AVX PairHMM implementation.
- Added code to enable Flush To Zero (FTZ) in `Main::main()`. 

**Note:** The FTZ setting only applies to the current thread. Enabling FTZ in a `Main` static initializer did not enable FTZ for the HaplotypeCaller execution thread, so we enable FTZ in `Main::main().` [This link](http://www.java-gaming.org/topics/disabling-floating-point-denormals/36491/view.html) has some discussion about using FTZ in Java.


droazen:
@lbergelson Could you take a look at the changes to the build system in this PR?

droazen:
@gspowley Could you close this, and open a new PR against master?

We've also become concerned that if something in the jvm is unsetting FTZ, turning it on at startup won't be sufficient. We added a protective "turn FTZ on" on every call to `jniComputeLikelihoods()`, which at least allows us to produce consistent test results, but we still have no protection outside the PairHMM against FTZ getting disabled.

gspowley:
@droazen Will do. I do want to root cause the jvm unsetting FTZ. I will have some time to focus on debug tomorrow. I'll update #1771 with any progress.

PR #1586, opened 2016-03-17
gradle 2.12

akiezun:
is this how a gradle update is done?


lbergelson:
:+1:  That's all there is to an update if nothing breaks.  You might want to change gradle-2.12-bin -> gradle-2.12-all if you want intellij to not complain about missing source files.  I assume it increases the download time some, but I haven't noticed it being a big difference.

PR #1587, opened 2016-03-17
Use IntelDeflater (libIntelDeflater.so) by default

### LINKED ISSUE #1585: Enable use of IntelDeflater when running GATK4
using IntelDeflater speeds up writing of bams by at least 15% (latest igzip is even faster, though only at compression level 1). We must have a way to use it. 

note: the htsjdk jar does not contain the .so file - it's only available in the zip file

akiezun:
fixes #1585 - getting IntelDeflater from htsjdk zipfile (it's not part of the jar!) and then putting the path to it as a default JVM argument.

@lbergelson can you review? this is a bit of a hack, improvements welcome
Do check that, out of the box (on linux) you get the IntelDeflater. I run this:
```
./gradlew clean installDist
./gatk-launch CountReads -I src/test/resources/large/NA12878.RNAseq.bam
```

and check output for `IntelDeflater`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,12 +27,20 @@ import org.gradle.internal.os.OperatingSystem
 
 mainClassName = "org.broadinstitute.hellbender.Main"
 
+applicationDefaultJvmArgs = ["-Dsamjdk.intel_deflater_so_path=build/libs/libIntelDeflater.so"]
+
 task downloadGsaLibFile(type: Download) {
     src 'http://cran.r-project.org/src/contrib/gsalib_2.1.tar.gz'
     dest "src/main/resources/org/broadinstitute/hellbender/utils/R/gsalib.tar.gz"
     overwrite false
 }
 
+task getIntelDeflater(type:Exec) {
+  workingDir '.'
+  
+  String version="2.1.1"
+  commandLine "/bin/bash", "-c", "wget --no-verbose https://github.com/samtools/htsjdk/archive/" + version + ".zip; unzip -q "+ version +".zip ; mkdir -p build/libs/; cp -v htsjdk-"+ version+"/lib/jni/libIntelDeflater.so build/libs/; rm -f "+ version +".zip; rm -fr htsjdk-"+version+"; "
+}
 
 repositories {
     mavenCentral()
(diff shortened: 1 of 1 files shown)


lbergelson:
Yeesh, inteldeflator isn't packaged in the htsjdk jar?  That's a bug we should fix in htsjdk.

akiezun:
until it's fixed, i want the fastest possible gatk

build.gradle
```diff
@@ -27,12 +27,20 @@ import org.gradle.internal.os.OperatingSystem
...
     src 'http://cran.r-project.org/src/contrib/gsalib_2.1.tar.gz'
     dest "src/main/resources/org/broadinstitute/hellbender/utils/R/gsalib.tar.gz"
     overwrite false
 }
 
+task getIntelDeflater(type:Exec) {
+  workingDir '.'
+  
+  String version="2.1.1"
+  commandLine "/bin/bash", "-c", "wget --no-verbose https://github.com/samtools/htsjdk/archive/" + version + ".zip; unzip -q "+ version +".zip ; mkdir -p build/libs/; cp -v htsjdk-"+ version+"/lib/jni/libIntelDeflater.so build/libs/; rm -f "+ version +".zip; rm -fr htsjdk-"+version+"; "
```
lbergelson: This seems fairly brittle.  Wouldn't it be better to just check the .so into our repo for now? 

I don't know if wget comes standard on osx and this will break any build that doesn't have wget available.  I know the only wget on my system is custom installed from brew.  Maybe newer osx comes with it?

akiezun: i guess we could check the .so in, or make this task ok to fail (is there a way?) because the .so is useless on the mac anyway

akiezun:
@lbergelson what do you prefer?

lbergelson:
@akiezun The so is ~.5mb so it's a little unfortunate to include it in here when it's really part of htsjdk.  I think the best thing to do for right now is to download it, but make the task slightly more robust.  We can use the download task type to download it (like we do with gsalib) and either hardcode a direct url to just the .so file, or have a second task to copy the file to where it needs to be.  (https://github.com/michel-kraemer/gradle-download-task has an example of simultaneous download and unzip)  

I would lean to just hardcoding the link rather than downloading the jar and unzipping it.
i.e. https://github.com/samtools/htsjdk/raw/2.1.1/lib/jni/libIntelDeflater.so.

akiezun:
sounds good. will download the so directly then

akiezun:
@lbergelson looks ok now?

lbergelson:
:+1:  merge when ready

PR #1588, opened 2016-03-17
use async IO by default

akiezun:
Switch to using async IO by default. It's an out-of-the-box win for almost everybody because almost everyone has more than 1 CPU. Difference is write speed is at least 20%.  

Also added a line to the output that indicates what IO we're using (a'la deflater)

On single processors it is slower than syncIO, so you can turn it off by running like this:
```
JAVA_OPTS="-Dsamjdk.use_async_io=false" ./gatk-launch PrintReads .....
```

@lbergelson can you review and test?

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,12 @@ If you are looking for the codebase of the current production version of GATK, p
   on a cluster or the cloud by building a Spark jar with `gradle installSpark` and passing the resulting jar in `build/libs/`
   directly to either `spark-submit` or `gcloud`.
 
+* Note: by default, GATK uses asynchronous IO and will use a separate thread for writting read and variant files. On systems with only 1 CPU this may reduce performance. To turn that option off, run GATK like this:
+
+```
+      JAVA_OPTS="-Dsamjdk.use_async_io=false" ./gatk-launch <rest of command>
+```
+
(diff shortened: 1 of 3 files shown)


lbergelson:
Any way to detect how many cores are available and toggle it off automatically in that case?

lbergelson:
Any way to detect how many cores are available and toggle it off automatically in that case?

lbergelson:
Should we bake this in as a default to gatk-launch as well?

akiezun:
it magically ends up in the gatk scrips that gatk-launch uses

lbergelson:
Yes, but those scripts don't get used when you're running through a sparkjar, so it won't be supplied by spark-submit unless we explicitly add it.

akiezun:
i dont think i want them to be applied to spark. just walkers for now

lbergelson:
since this is effectively a hidden option, could we add something to the readme or wiki about it?

:+1:  other than that

lbergelson:
Using my very sophisticated profiling by running `time` a few times, I see a speedup in wall clock time with this, but also an increase in total cpu time.  As above, it seems reasonable to have it on by default.

akiezun:
added to readme

PR #1589, opened 2016-03-17
ReadFilteringIterator: an iterator that filters reads from another iterator

droazen:
This was extracted from the HaplotypeCaller branch currently under review,
as it's potentially useful for the LocusWalker branch.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/iterators/ReadFilteringIterator.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/iterators/ReadFilteringIterator.java
@@ -1,0 +1,65 @@
+package org.broadinstitute.hellbender.utils.iterators;
+
+import org.broadinstitute.hellbender.engine.filters.ReadFilter;
+import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+/**
+ * An iterator that filters reads from an existing iterator of reads.
+ */
+public class ReadFilteringIterator implements Iterator<GATKRead>, Iterable<GATKRead> {
+
+    private final Iterator<GATKRead> nestedIterator;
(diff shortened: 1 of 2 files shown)


droazen:
@lbergelson can you review?

droazen:
This is necessary in situations where you need to hand a reads iterator to another class, and can't do the reads filtering in-place.

akiezun:
we can use streams instead of iterators?
also, are there no filtering iterators in some apache code or guava?

lbergelson:
The code looks fine to me.  It seems a bit unfortunate to need a custom type here when we could do either the nasty conversion of Iterator -> Stream, or use the guava `Iterators.filter`.

droazen:
Ok, if you guys want to vouch for guava's `Iterators.filter()` we can use that -- I confess that I didn't look thoroughly enough at the existing options before implementing this.

droazen:
Spoke too soon, Louis tried the guava option and it didn't work with our ReadFilters -- unless someone has a working alternative I say we just merge this for now.

lbergelson:
Stupid guava alternate classes with the same names.

lbergelson:
@droazen [Stack overflow](http://stackoverflow.com/questions/24511052/java8-iterator-to-stream) offers:


 ```java
   public static class StreamUtils {

        public static <T> Stream<T> asStream(Iterator<T> sourceIterator) {
            return asStream(sourceIterator, false);
        }

        public static <T> Stream<T> asStream(Iterator<T> sourceIterator, boolean parallel) {
            Iterable<T> iterable = () -> sourceIterator;
            return StreamSupport.stream(iterable.spliterator(), parallel);
        }
    }
```

which is something we should probably include, and allows the reasonable
```java
       StreamUtils.asStream(iter).filter(ReadFilterLibrary.ALLOW_ALL_READS).iterator();

```

droazen:
I'll merge this branch for now, since it's needed elsewhere, and if you want to replace it with something more elegant in the future be my guest :)

PR #1533, opened 2016-02-29
Spark BQSR multiple known sites/getParallelVariants for multple varia…

cmnbroad:
…nt files. Fixes https://github.com/broadinstitute/gatk/issues/1085

PR #1590, opened 2016-03-18
Fix test in BQSRSparkIntegrationTest.

cmnbroad:
One of the non-cloud tests in BQSRSparkIntegrationTest tries to use the API key when its not required; as a result the test fails when the key isn't set even though it should pass. Introduced in the test refactoring that was part of https://github.com/broadinstitute/gatk/pull/1533.


akiezun:
:+1: merge when tests pass

PR #1591, opened 2016-03-18
remove Windows starting script

akiezun:
we don't test on windows and I doubt gatk would work on windows so let's not generate the windows startup to avoid confusing people

@lbergelson can you look at it?


lbergelson:
:+1: Looks good to me.

Issue #1593, opened 2016-03-18
CompareBaseQualitiesSpark: unfriendly stack trace when reference file missing


akiezun:
```
./gatk-launch CompareBaseQualitiesSpark -I HCC1143.bqsr.bam -I2 HCC1143.bqsr.cram -R ../human_g1k_v37.fasta --disableSequenceDictionaryValidation
```

blows up with
```
htsjdk.samtools.SAMException: Error opening file: ../human_g1k_v37.fasta
	at htsjdk.samtools.util.IOUtil.openFileForReading(IOUtil.java:523)
	at htsjdk.samtools.reference.FastaSequenceFile.<init>(FastaSequenceFile.java:59)
	at htsjdk.samtools.reference.ReferenceSequenceFileFactory.getReferenceSequenceFile(ReferenceSequenceFileFactory.java:129)
	at htsjdk.samtools.reference.ReferenceSequenceFileFactory.getReferenceSequenceFile(ReferenceSequenceFileFactory.java:82)
	at htsjdk.samtools.reference.ReferenceSequenceFileFactory.getReferenceSequenceFile(ReferenceSequenceFileFactory.java:70)
	at htsjdk.samtools.reference.ReferenceSequenceFileFactory.getReferenceSequenceFile(ReferenceSequenceFileFactory.java:59)
	at org.broadinstitute.hellbender.engine.datasources.ReferenceFileSource.getReferenceSequenceDictionary(ReferenceFileSource.java:52)
	at org.broadinstitute.hellbender.engine.datasources.ReferenceMultiSource.getReferenceSequenceDictionary(ReferenceMultiSource.java:110)
	at org.broadinstitute.hellbender.engine.spark.GATKSparkTool.initializeReference(GATKSparkTool.java:354)
	at org.broadinstitute.hellbender.engine.spark.GATKSparkTool.initializeToolInputs(GATKSparkTool.java:320)
	at org.broadinstitute.hellbender.engine.spark.GATKSparkTool.runPipeline(GATKSparkTool.java:311)
	at org.broadinstitute.hellbender.engine.spark.SparkCommandLineProgram.doWork(SparkCommandLineProgram.java:38)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:102)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:151)
[...]

Issue #1321, opened 2015-12-09
CompareBaseQualitiesSpark complains about reference for cram files even though i gave it one


akiezun:
`./gatk-launch CompareBaseQualities -I HCC1143.bqsr.bam -I2 HCC1143.bqsr.cram -R ../human_g1k_v37.fasta --disableSequenceDictionaryValidation`

results in 
`A USER ERROR has occurred: A reference file is required when using CRAM files.`

which is bogus because i passed in a reference

droazen:
alpha

akiezun:
I think the problem is that it uses a non-standard way to specify the second read input (using a new field rather than the standard input collection). 

I think this can wait past alpha - not a critical tool and not a standard way of using it. @droazen wdyt?

droazen:
Agreed @akiezun -- moving to beta

droazen:
@cmnbroad if you have time, feel free to have a look

cmnbroad:
I can't reproduce this problem. CompareBaseQualities accepts positional arguments for the two inputs, and rejects -I..-I2 altogether (I don't get the CRAM reference error, it fails on the -I). I've added some integration tests that use 2 CRAMs and a mix of CRAM and BAM, and the reference seems to be accepted fine. Let me know if I'm missing something.

droazen:
@akiezun Can you clarify -- is this bug invalid?

droazen:
@akiezun Perhaps you meant `CompareBaseQualitiesSpark` rather than `CompareBaseQualities`? The spark version takes `-I` and `-I2` args

akiezun:
yes, the tool got renamed in the meantime. The commandline is:
```
/gatk-launch CompareBaseQualitiesSpark -I ./src/test/resources/org/broadinstitute/hellbender/tools/print_reads.bam -I2 ./src/test/resources/org/broadinstitute/hellbender/tools/print_reads.cram -R ./src/test/resources/org/broadinstitute/hellbender/tools/print_reads.fasta --disableSequenceDictionaryValidation
```

the error is `A USER ERROR has occurred: A reference is required for CRAM input`

akiezun:
back to @cmnbroad

PR #1595, opened 2016-03-18
Fix CompareBaseQualitiesSpark reference handling.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/1321. Also added some CRAM tests. I started to fix https://github.com/broadinstitute/gatk/issues/1593 as part of this but there are several code paths that need to change so I'll do a separate PR.


droazen:
:+1: merge when ready

PR #1596, opened 2016-03-18
adding an option for stacktraces on UserExceptions

lbergelson:
setting property GATK_STACKTRACE_ON_USER_EXCEPTION=true will cause UserExceptions to emit stacktraces
the default behavior is to hide the stacktrace on a user exception

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,8 @@ If you are looking for the codebase of the current production version of GATK, p
       Look for the line that says `See the test report at`.
       If TestNG itself crashes there will be no report generated.
 
+* To output stack traces for `UserException` set the environment variable `GATK_STACKTRACE_ON_USER_EXCEPTION=true`
+
 ##General guidelines for GATK4 developers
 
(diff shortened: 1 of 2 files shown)


lbergelson:
@akiezun please review, I didn't add any tests because it's very tricky to test `Main.main` due to the use of System.exit`

akiezun:
:+1: 
I checked it by running like this:
```
GATK_STACKTRACE_ON_USER_EXCEPTION=true  ./gatk-launch CountReads -I ./src/test/resources/org/broadinstitute/hellbender/utils/test/SamAssertionUtilsUnitTest/valid.cram -R ./src/test/resources/problematicFASTA.fasta
```

and another time with `GATK_STACKTRACE_ON_USER_EXCEPTION=false`
(not that there is a stacktrace on the second invocation too because of https://github.com/broadinstitute/gatk/issues/1598

lbergelson:
Erm... I think I introduced #1598.

akiezun:
it's yours then

PR #1600, opened 2016-03-21
updating git-lfs on travis to v1.1.2

### LINKED ISSUE #1404: remove git-lfs workaround when once git lfs pull bug is fixed
We encountered an error with git-lfs in #1378.  This was worked around by simply pulling twice in `scripts/install_git_lfs.sh`, but it shouldn't have to do that.  

Remove this extra pull once a better solution is found in https://github.com/github/git-lfs/issues/904

lbergelson:
his version includes a fix for the issue we were seeing which required multiple pulls
removed the workaround (fixes #1404)

The problem was that we were using an ssh git remote.  git-lfs fails in this case because it can't authenticate.  (unclear to me if this is a bug or not, I don't know if github requires authentication for ssh access to public repos).  In versions <= 1.1.0 git-lfs was falling back to trying over an http connection.  They removed this fallback mechanism in 1.1.1.  See github/git-lfs#1090 for discussion.

### DIFF
--- a/scripts/install_git_lfs.sh
+++ b/scripts/install_git_lfs.sh
@@ -1,38 +1,30 @@
 #!/bin/bash
 
-GIT_LFS_VERSION=v1.1.0
+GIT_LFS_VERSION="1.1.2"
+GIT_LFS_LINK=https://github.com/github/git-lfs/releases/download/v${GIT_LFS_VERSION}/git-lfs-linux-amd64-${GIT_LFS_VERSION}.tar.gz
+GIT_LFS="git-lfs-${GIT_LFS_VERSION}/git-lfs"
+echo "downloading and untarring git-lfs binary" 
+wget -qO- $GIT_LFS_LINK | tar xvz
 
-echo "cloning"
-git clone https://github.com/github/git-lfs.git
-
-echo "cd into git-lfs"
-cd git-lfs
-
-echo "checkout commit e8a5cfb85e27c1284a73a2eb8269bd5f5c4d0955"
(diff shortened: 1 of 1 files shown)


akiezun:
wait, do we need sudo now?

lbergelson:
Ah, I lost that from my commit comment somehow.  Yes.  I talked with travis support recently, and they suggested that we should return to the sudo-enabled builds.  They are on new infrastructure and no longer have the capacity limitation they used to.  They have a ~1 minute startup time, which is slightly longer than the container builds, but they should hopefully make up for it by being somewhat faster nodes.  2 proper cores and 7.5 gigs for ram vs 1.5 virtual cores and ~3.5 gigs.

lbergelson:
It also moves to a more recent ubuntu version so apt-get packages are more up to date when we need them.

lbergelson:
I can split out the travis changes from this branch if you think it's overloading it.  Initially I thought we'd need them because I wanted to update the go version to build from source, but that turned out to not be necessary.  They're orthogonal to the git-lfs changes in the final form.

akiezun:
yes, can you split them? unless using sudo somehow improves things (?)

akiezun:
also, this branch has substantially shorter build times than master - do you know why? is it the sudo or is it the new LFS?

lbergelson:
@akiezun That should be due to the travis changes.  Might shave 20 seconds off with the lfs changes, but nothing significant.

lbergelson:
We'll see if this clean commit takes longer.

lbergelson:
@akiezun Good to merge?

PR #1602, opened 2016-03-21
moving to sudo enabled travis

lbergelson:
travis support recommended that we move to travis with sudo since the capacity problems have been fixed
the travis nodes with sudo are more powerful now and there shouldn't be long startup delays like there used to be
also moving to the newer ubuntu image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,5 @@
-language: java
-addons:
-  apt:
-    sources:
-    - r-packages-precise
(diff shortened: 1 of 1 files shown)


lbergelson:
This seems to speed the tests up by almost 2x...  ~5 an 18 vs ~10 and 30min

akiezun:
:+1:  love fast builds

PR #1616, opened 2016-03-23
squid:S2178 - Short-circuit logic should be used in boolean contexts

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S2178 - Short-circuit logic should be used in boolean contexts

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2178

Please let me know if you have any questions.

M-Ezzat


akiezun:
this is fine - all those used a variable as a second argument to the semantics is equivalent and this change improves readability (less surpirse to the reader)

PR #1619, opened 2016-03-24
build script: all JVM args should be on 1 line to not get lost

akiezun:
all JVM args should be on 1 line to not get lost


lbergelson:
:+1:  squash and merge

PR #1620, opened 2016-03-24
fix inefficient hashcode computation and added tests to 100%

akiezun:
came up in recent profiling - hashCode computation on KMer was very inefficient and wasteful in String creation 

@lbergelson can you review?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/Kmer.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/Kmer.java
@@ -67,7 +67,22 @@ public final class Kmer {
         this.bases = bases;
         this.start = start;
         this.length = length;
-        this.hash = new String(bases, start, length).hashCode();
+        this.hash = hashCode(bases, start, length);
+    }
+
+    /**
+     *  Compute the hashcode for a KMer.
+     *  Equivalent to <code>new String(bases, start, length).hashCode()</code>
+     */
+    private static int hashCode(final byte[] bases, final int start, final int length) {
+        if (length == 0){
+            return 0;
+        }
+        int h = 0;
+        for (int i = start, stop = start + length; i < stop; i++) {
+            h = 31 * h + bases[i];
+        }
+        return h;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/Kmer.java
```diff
@@ -114,7 +129,7 @@ public int length() {
      * differingBases = {'G','G'}
      * @param other                 Other k-mer to test
      * @param maxDistance           Maximum distance to search. If this and other k-mers are beyond this Hamming distance,
-     *                              search is aborted and a null is returned
+     *                              search is aborted and -1 is returned
```
lbergelson: good catch

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/KmerUnitTest.java
```diff
@@ -79,6 +99,30 @@ public void testSubkmer() {
         }
     }
 
+    @Test(expectedExceptions = IllegalArgumentException.class)
+    public void testErrorDifereingPositions() throws Exception {
```
lbergelson: typo testErrorDifferingPositions()

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/Kmer.java
```diff
@@ -67,7 +67,22 @@ public Kmer(final byte[] bases, final int start, final int length) {
         this.bases = bases;
         this.start = start;
         this.length = length;
-        this.hash = new String(bases, start, length).hashCode();
+        this.hash = hashCode(bases, start, length);
+    }
+
+    /**
+     *  Compute the hashcode of this KMer
```
lbergelson: Since it's static it's not technically "computing the hashCode for **this** KMer" .

lbergelson:
Two documentation nitpicks.  Looks good.  It's good that you're fixing this, since i think the entire point of this class was to avoid copying the bases, and it ended up just copying the bases for the hashcode.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/Kmer.java
akiezun: fixed

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/KmerUnitTest.java
akiezun: fixed

PR #1621, opened 2016-03-24
check containsVertex faster and added tests to 100%

akiezun:
This came up during profiling - containsVertex was creating full lists for vertices all the time only to check membership. Using a scan now (no caching but that's possible too if needed later)

Tests for Path at 100% now.

@lbergelson can you review?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/Path.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/Path.java
@@ -122,9 +122,7 @@ public final class Path<T extends BaseVertex, E extends BaseEdge> {
      */
     public boolean containsVertex(final T v) {
         Utils.nonNull(v, "Vertex cannot be null");
-
-        // TODO -- warning this is expensive.  Need to do vertex caching
-        return getVertices().contains(v);
+        return v.equals(getFirstVertex()) || edgesInOrder.stream().map(graph::getEdgeTarget).anyMatch(v::equals);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/Path.java
```diff
@@ -123,8 +123,26 @@ public boolean pathsAreTheSame(final Path<T,E> path) {
     public boolean containsVertex(final T v) {
         Utils.nonNull(v, "Vertex cannot be null");
 
-        // TODO -- warning this is expensive.  Need to do vertex caching
-        return getVertices().contains(v);
+
+        if ( getEdges().isEmpty() ) {
+            return false;
```
lbergelson: Shouldn't this be `return v.equals(lastVertex);` ?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/Path.java
```diff
@@ -123,8 +123,26 @@ public boolean pathsAreTheSame(final Path<T,E> path) {
...
+        } else {
+            boolean first = true;
+            for ( final E e : getEdges() ) {
```
lbergelson: could you change the return type of `getEdges()` to `List`?  It's bothering me that its return type is `Collection`

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/Path.java
```diff
@@ -123,8 +123,26 @@ public boolean pathsAreTheSame(final Path<T,E> path) {
...
+                if ( first ) {
```
lbergelson: The logic might be clearer if this whole "is first" block was lifted out of the loop.

lbergelson: (likewise in the get vertices)

lbergelson:
@akiezun Back to you.  I think there's a minor bug.  Could you add a test that covers that case and show if it's a bug or not?  (Seems like this class should be Iterable, but that's a different issues)

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/Path.java
akiezun: nice catch. fixed and tests added

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/Path.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/Path.java
akiezun: done, replaced all of this with 
```
  return v.equals(getFirstVertex()) || edgesInOrder.stream().map(graph::getEdgeTarget).anyMatch(v::equals);
```

which does not create a list but does create some objects to it may come up on profiling again (in which case we'll rewrite for speed)

akiezun:
back to @lbergelson

lbergelson:
:+1: Much clearer now I think.  Squash and merge.

PR #1622, opened 2016-03-24
findNumberOfRepetitions makes fewer useless arrays

akiezun:
Came in in profiling - findNumberOfRepetitions was creating lots of byte arrays that were discarded right away after comparing of a small range. This PR makes the comparison work on arrays directly.

@lbergelson can you have a look?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
@@ -415,17 +415,21 @@ public final class GATKVariantContextUtils {
      * @param refBasesStartingAtVCWithPad
      * @return
      */
-    public static Pair<List<Integer>,byte[]> getNumTandemRepeatUnits(final VariantContext vc, final byte[] refBasesStartingAtVCWithPad) {
+    public static Pair<List<Integer>, byte[]> getNumTandemRepeatUnits(final VariantContext vc, final byte[] refBasesStartingAtVCWithPad) {
+        Utils.nonNull(vc);
+        Utils.nonNull(refBasesStartingAtVCWithPad);
+
+        if ( ! vc.isIndel() ){ // only indels are tandem repeats
+            return null;
+        }
         final boolean VERBOSE = false;
         final String refBasesStartingAtVCWithoutPad = new String(refBasesStartingAtVCWithPad).substring(1);
-        if ( ! vc.isIndel() ) // only indels are tandem repeats
-            return null;
 
         final Allele refAllele = vc.getReference();
         final byte[] refAlleleBases = Arrays.copyOfRange(refAllele.getBases(), 1, refAllele.length());
 
         byte[] repeatUnit = null;
-        final ArrayList<Integer> lengths = new ArrayList<>();
+        final List<Integer> lengths = new ArrayList<>();
 
         for ( final Allele allele : vc.getAlternateAlleles() ) {
             Pair<int[],byte[]> result = getNumTandemRepeatUnits(refAlleleBases, Arrays.copyOfRange(allele.getBases(), 1, allele.length()), refBasesStartingAtVCWithoutPad.getBytes());
@@ -512,38 +516,53 @@ public final class GATKVariantContextUtils {
     /**
      * Helper routine that finds number of repetitions a string consists of.
      * For example, for string ATAT and repeat unit AT, number of repetitions = 2
-     * @param repeatUnit             Substring
-     * @param testString             String to test
-     * @oaram lookForward            Look for repetitions forward (at beginning of string) or backward (at end of string)
-     * @return                       Number of repetitions (0 if testString is not a concatenation of n repeatUnit's
+     * @param repeatUnit             Non-empty substring represented by byte array
+     * @param testString             String to test (represented by byte array), may be empty
+     * @param leadingRepeats         Look for leading (at the beginning of string) or trailing (at end of string) repetitions
+     * For example:
+     *    GATAT has 2 trailing repeats of A but 0 leading repeats of AT
+     *    ATATG has 0 leading repeats of A but 2 forward repeats of AT
+     *    CCCCCCCC has to forward and 2 trailing repeats of CCC
+     *
+     * @return  Number of repetitions (0 if testString is not a concatenation of n repeatUnit's, including the case of empty testString)
      */
-    public static int findNumberOfRepetitions(byte[] repeatUnit, byte[] testString, boolean lookForward) {
-        int numRepeats = 0;
-        if (lookForward) {
+    public static int findNumberOfRepetitions(byte[] repeatUnit, byte[] testString, boolean leadingRepeats) {
+        Utils.nonNull(repeatUnit, "repeatUnit");
+        Utils.nonNull(testString, "testString");
+        Utils.validateArg(repeatUnit.length != 0, "empty repeatUnit");
+        if (testString.length == 0){
+            return 0;
+        }
+        final int repeatLength = repeatUnit.length;
+        final int testLength = testString.length;
+        final int lengthDifference = testLength - repeatLength;
+
+        if (leadingRepeats) {
+            int numRepeats = 0;
             // look forward on the test string
-            for (int start = 0; start < testString.length; start += repeatUnit.length) {
-                int end = start + repeatUnit.length;
-                byte[] unit = Arrays.copyOfRange(testString,start, end);
-                if(Arrays.equals(unit,repeatUnit))
+            for (int start = 0; start <= lengthDifference; start += repeatLength) {
+                if(Utils.equalRange(testString, start, repeatUnit, 0, repeatLength)) {
                     numRepeats++;
-                else
+                } else {
                     break;
+                }
+            }
+            return numRepeats;
+        } else {
+            // look backward. For example, if repeatUnit = AT and testString = GATAT, number of repeat units is still 2
+            int numRepeats = 0;
+            // look backward on the test string
+            for (int start = lengthDifference; start >= 0; start -= repeatLength) {
+                if (Utils.equalRange(testString, start, repeatUnit, 0, repeatLength)) {
+                    numRepeats++;
+                } else {
+                    break;
+                }
             }
             return numRepeats;
         }
-
-        // look backward. For example, if repeatUnit = AT and testString = GATAT, number of repeat units is still 2
-        // look forward on the test string
-        for (int start = testString.length - repeatUnit.length; start >= 0; start -= repeatUnit.length) {
-            int end = start + repeatUnit.length;
-            byte[] unit = Arrays.copyOfRange(testString,start, end);
-            if(Arrays.equals(unit,repeatUnit))
-                numRepeats++;
-            else
-                break;
-        }
-        return numRepeats;
     }
+
 
     /**
      * Helper function for isTandemRepeat that checks that allele matches somewhere on the reference
--- a/src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
@@ -474,4 +474,36 @@ public final class UtilsUnitTest extends BaseTest {
                 {"romeo+juliette" , "01111010111001" },
         };
     }
+
+    @DataProvider(name = "testEqualRange")
+    public Object[][] testEqualRange(){
+        final Object[][] result = {
+            {"ATGATGATGATG".getBytes(), 0, 3, "ATG".getBytes(), true},
+            {"ATGATGATGATG".getBytes(), 3, 6, "ATG".getBytes(), true},
+            {"ATGATGATGATG".getBytes(), 9, 12, "ATG".getBytes(), true},
+
+            {"ATGATGATGATG".getBytes(), 0, 1, "G".getBytes(), false},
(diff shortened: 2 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -535,17 +535,43 @@ public static int findNumberOfRepetitions(byte[] repeatUnit, byte[] testString,
...
+     *  Checks if arr1 starting at 'from' and ending right before 'to' equals to arr2.
+     *  The 'to' argument may lie outside of arr1 and then the 'elements' beyond arr1 are treated as zero.
+     *  In other words this is equivalent to:
+     <code>
+             byte[] unit = Arrays.copyOfRange(arr1, from, to);    //this allows 'to' to be outside of arr1
+             return Arrays.equals(unit,arr2);
+     </code>
+     *  but avoids making a new array instance.
+     */
+    private static boolean equalToSubarray(final byte[] arr1, final int from, final int to, final byte[] arr2){
```
droazen: This should be in `Utils` rather than embedded within this class, and should (probably) throw if `from` and `to` are out of bounds on `arr1` (unless the calling code actually requires this, which seems unlikely).

akiezun:
how about this - i eliminated the method altogether and using Utils.equalRange

@droazen

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -518,33 +518,37 @@ public static int findRepeatedSubstring(byte[] bases) {
      * @return                       Number of repetitions (0 if testString is not a concatenation of n repeatUnit's
      */
     public static int findNumberOfRepetitions(byte[] repeatUnit, byte[] testString, boolean lookForward) {
```
droazen: Method is deprecated in GATK3 with the message: `@deprecated Move to use TandemRepeatFinder in protected (move to public if needed).`. Should we create a ticket in GATK4 for this task (migrating to `TandemRepeatFinder` here)?

droazen: Also, since this is a public method I think you need `Utils.nonNull()` on the array args.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -518,33 +518,37 @@ public static int findRepeatedSubstring(byte[] bases) {
      * @return                       Number of repetitions (0 if testString is not a concatenation of n repeatUnit's
```
droazen: Add a better explanation for `lookForward` mode vs. backward mode (I was initially confused as to the intended behavior of this method in backward mode), plus an example for the backward case.

droazen: Also, `@oaram` -> `@param`

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -518,33 +518,37 @@ public static int findRepeatedSubstring(byte[] bases) {
...
      */
     public static int findNumberOfRepetitions(byte[] repeatUnit, byte[] testString, boolean lookForward) {
-        int numRepeats = 0;
+        final int repeatLength = repeatUnit.length;
+        final int testLength = testString.length;
+        final int lengthDifference = testLength - repeatLength;
```
droazen: Throw if `lengthDifference` is negative?

droazen: (or, if you prefer, confirm via unit test that the method does something sane in that case, and document that behavior)

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -518,33 +518,37 @@ public static int findRepeatedSubstring(byte[] bases) {
...
+
         if (lookForward) {
+            int numRepeats = 0;
             // look forward on the test string
-            for (int start = 0; start < testString.length; start += repeatUnit.length) {
-                int end = start + repeatUnit.length;
-                byte[] unit = Arrays.copyOfRange(testString,start, end);
-                if(Arrays.equals(unit,repeatUnit))
+            for (int start = 0; start < testLength; start += repeatLength) {
+                if(start <= lengthDifference && Utils.equalRange(testString, start, repeatUnit, 0, repeatLength)) {
```
droazen: You could simplify this a bit if you made `start <= lengthDifference` the loop condition instead of `start < testLength`

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -518,33 +518,37 @@ public static int findRepeatedSubstring(byte[] bases) {
...
                     break;
+                }
+            }
+            return numRepeats;
+        } else {
+            // look backward. For example, if repeatUnit = AT and testString = GATAT, number of repeat units is still 2
+            int numRepeats = 0;
+            // look backward on the test string
+            for (int start = lengthDifference; start >= 0; start -= repeatLength) {
+                if (start <= lengthDifference && Utils.equalRange(testString, start, repeatUnit, 0, repeatLength)) {
```
droazen: The `start <= lengthDifference` test is redundant here, since `start` is initialized to `lengthDifference` and decremented after that.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -518,33 +518,37 @@ public static int findRepeatedSubstring(byte[] bases) {
...
+                if(start <= lengthDifference && Utils.equalRange(testString, start, repeatUnit, 0, repeatLength)) {
                     numRepeats++;
-                else
+                } else {
                     break;
+                }
+            }
+            return numRepeats;
+        } else {
+            // look backward. For example, if repeatUnit = AT and testString = GATAT, number of repeat units is still 2
```
droazen: Move this example to the method-level javadoc

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
```diff
@@ -773,6 +773,22 @@ public void testRepeatDetectorTest(RepeatDetectorTest cfg) {
...
+        Assert.assertEquals(GATKVariantContextUtils.findNumberOfRepetitions("T".getBytes(), "T".getBytes(), true),1);
+        Assert.assertEquals(GATKVariantContextUtils.findNumberOfRepetitions("AT".getBytes(), "ATGATGATCATG".getBytes(), true),1);
+        Assert.assertEquals(GATKVariantContextUtils.findNumberOfRepetitions("CCC".getBytes(), "CCCCCCCC".getBytes(), true),2);
+
+        Assert.assertEquals(GATKVariantContextUtils.findNumberOfRepetitions("AT".getBytes(), "GATAT".getBytes(), false), 2);
+        Assert.assertEquals(GATKVariantContextUtils.findNumberOfRepetitions("ATG".getBytes(), "ATGATGATGATG".getBytes(), false),4);
+        Assert.assertEquals(GATKVariantContextUtils.findNumberOfRepetitions("G".getBytes(), "ATGATGATGATG".getBytes(), false),1);
+        Assert.assertEquals(GATKVariantContextUtils.findNumberOfRepetitions("T".getBytes(), "T".getBytes(), false),1);
+        Assert.assertEquals(GATKVariantContextUtils.findNumberOfRepetitions("AT".getBytes(), "ATGATGATCATG".getBytes(), false),0);
+        Assert.assertEquals(GATKVariantContextUtils.findNumberOfRepetitions("CCC".getBytes(), "CCCCCCCC".getBytes(), false),2);
```
droazen: Add test cases for both the forward and reverse cases with the length of the first byte array > the length of the second, and verify that the method returns 0.

droazen: Also add tests for the case of null arrays.

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -474,4 +474,35 @@ public void testSkimArray(final String original, final String remove) {
...
+            {"ATGATGATGATG".getBytes(), 11, 12, "G".getBytes(), true},
+            {"ATGATGATGATG".getBytes(), 10, 11, "G".getBytes(), false},
+            {"ATGATGATCATG".getBytes(), 10, 12, "AT".getBytes(), false},
+        };
+        return result;
+    }
+
+    @Test(dataProvider = "equalToSubarrayData")
+    public void testEqualToSubarray(byte[] arr1, int from, int to, byte[] arr2, boolean expected) throws Exception {
+        Assert.assertEquals(Utils.equalRange(arr1, from, arr2, 0, to - from), expected);
```
droazen: BUGBUG: You are off by one here -- the number of positions to compare is `to - from + 1`, assuming `from` and `to` form a closed interval.

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -474,4 +474,35 @@ public void testSkimArray(final String original, final String remove) {
...
+
+            {"ATGATGATGATG".getBytes(), 11, 12, "G".getBytes(), true},
+            {"ATGATGATGATG".getBytes(), 10, 11, "G".getBytes(), false},
+            {"ATGATGATCATG".getBytes(), 10, 12, "AT".getBytes(), false},
+        };
+        return result;
+    }
+
+    @Test(dataProvider = "equalToSubarrayData")
+    public void testEqualToSubarray(byte[] arr1, int from, int to, byte[] arr2, boolean expected) throws Exception {
```
droazen: Should be named `testEqualRange`

droazen:
Review complete -- back to @akiezun. Merge after addressing comments.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
akiezun: done https://github.com/broadinstitute/gatk/issues/1635 and fixed

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
akiezun: changed language to leading vs trailing repeats

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
akiezun: already documented `@return  Number of repetitions (0 if testString is not a concatenation of n repeatUnit's)`

I added tests for it

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
akiezun: good point. done

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
akiezun: true. fixed

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
droazen: Make sure you still handle the case where `lengthDifference` == `testLength` correctly, though! (ie., the `repeatLength

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
akiezun: repeatLength == 0 can only be tolerated if testString has length 0. This is still a bogus condition so it'll disallow it (throw) and add tests for it,

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
akiezun: done

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
akiezun: yes, 'to' is exclusive. added a comment

PR #1623, opened 2016-03-24
Smith Waterman performance - small tweaks

akiezun:
came up in profiling - SmithWaterman is a hotspot. This PR lifts some field lookups up from tight loops.

@lbergelson can you look?


lbergelson:
:+1:  I don't see any problems although I'm skeptical about some of the optimization making a difference.

PR #1624, opened 2016-03-24
compute cigar faster when equal strings + added tests

akiezun:
came up in profiling - we were calling full blown SmithWaterman even when strings are equal. Comparison is linear and fast, so let's use as a shortcut

@lbergelson can you review?


src/main/java/org/broadinstitute/hellbender/utils/read/CigarUtils.java
```diff
@@ -268,6 +268,12 @@ public static Cigar calculateCigar(final byte[] refSeq, final byte[] altSeq) {
             return new Cigar(Arrays.asList(new CigarElement(refSeq.length, CigarOperator.D)));
         }
 
+        if (Arrays.equals(refSeq, altSeq)){
```
lbergelson: add a comment explaining that this is a performance optimization

lbergelson:
:+1:  Could you add a note explaining that it's a short circuit for performance reasons?  Merge when ready.

src/main/java/org/broadinstitute/hellbender/utils/read/CigarUtils.java
akiezun: done

PR #1625, opened 2016-03-24
precompute hashcode in MultiDeBruijnVertex and added a bunch of tests

akiezun:
came up on profiling - computing the hashcode over and over is expensive to we precompute it

@lbergelson please review


lbergelson:
:+1:

PR #1630, opened 2016-03-24
tests must use same JVM system properties as gatk-launch

akiezun:
Our launch script gatk-launch now uses a bunch of system properties for things like compression, async reading, deflater. Our testing infrastructure must use the same settings.

for @lbergelson

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,6 +26,8 @@ import de.undercouch.gradle.tasks.download.Download
 import org.gradle.internal.os.OperatingSystem
 
 mainClassName = "org.broadinstitute.hellbender.Main"
+
+//Note: the test suite must use the same defaults. If you change system properties in this list you must also update the one in the test task
 applicationDefaultJvmArgs = ["-Dsamjdk.use_async_io=false", "-Dsamjdk.intel_deflater_so_path=build/libIntelDeflater.so", "-Dsamjdk.compression_level=1"]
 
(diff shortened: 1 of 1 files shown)


lbergelson:
How do you confirm the tests are actually using the intel deflator? Can we add a test on travis that checks that it's using it? 

It would be good if we could specify these in 1 place and have them applied everywhere they need to be.

Looks :+1:  to me though modulo possible test to verify that right deflator is being used.

akiezun:
ha! i put it on the wrong target before. Now I put on the test target and verified the right ones are set by `jinfo -sysprops <PID> | grep -i samjdk` (pid of my test vm)

I also deleted the testAll target because I think it was unused. 
I don't know how to verify that travis is running the right ones but now at least I do know we're running the right ones on the commandline.

akiezun:
back to @lbergelson

lbergelson:
Ooh, good catch, I didn't see that it was in testAll.  Good call to delete that, I didn't even remember we had it...

The pull request builds is failing, I think for because of a bug in asynchronous tribble reading.

https://storage.googleapis.com/hellbender/test/build_reports/5109.2/tests/classes/org.broadinstitute.hellbender.tools.walkers.filters.VariantFiltrationIntegrationTest.html

akiezun:
cool, maybe we discovered a bug this way! i'll have a look

Issue #1638, opened 2016-03-26
Errors in VariantFiltration - possibly due to (or exposed by) async tribble reading 


akiezun:
this happens on the branch for https://github.com/broadinstitute/gatk/pull/1630 (which uses async IO for tests to mimic non-test usage). This bug is either due to or exposed by asynchronous tribble reading.
 
more logs https://travis-ci.org/broadinstitute/gatk/jobs/118507152

test results
https://storage.googleapis.com/hellbender/test/build_reports/5109.2/tests/classes/org.broadinstitute.hellbender.tools.walkers.filters.VariantFiltrationIntegrationTest.html#testClusteredSnps

```
java.lang.RuntimeException: htsjdk.tribble.TribbleException: Exception encountered in worker thread.
	at org.broadinstitute.hellbender.utils.test.IntegrationTestSpec.executeTest(IntegrationTestSpec.java:153)
	at org.broadinstitute.hellbender.utils.test.IntegrationTestSpec.executeTest(IntegrationTestSpec.java:126)
	at org.broadinstitute.hellbender.utils.test.IntegrationTestSpec.executeTest(IntegrationTestSpec.java:108)
	at org.broadinstitute.hellbender.tools.walkers.filters.VariantFiltrationIntegrationTest.testClusteredSnps(VariantFiltrationIntegrationTest.java:36)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:85)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:639)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:821)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1131)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:124)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:108)
	at org.testng.TestRunner.privateRun(TestRunner.java:773)
	at org.testng.TestRunner.run(TestRunner.java:623)
[...]

PR #1649, opened 2016-03-28
turning async off async until tribble is fixed

lbergelson:
turning off async until until we can separate tribble async from bam async

see #1638


droazen:
:+1: merge when ready

PR #1662, opened 2016-03-28
Add GCS-NIO basic unit test

jean-philippe-martin:
This illustrates how to read public and private files.

It also requires a bit of setup as we have to grab a service account credential
and opt into the gcs-nio branch of gcloud-java, for now.

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/utils/nio/NioTests.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/nio/NioTests.java
@@ -1,0 +1,73 @@
+package org.broadinstitute.hellbender.utils.nio;
+
+import com.google.gcloud.AuthCredentials;
+import com.google.gcloud.RetryParams;
+import com.google.gcloud.storage.StorageOptions;
+import com.google.gcloud.storage.contrib.nio.CloudStorageConfiguration;
+import com.google.gcloud.storage.contrib.nio.CloudStorageFileSystem;
+
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.annotations.Test;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Test GCS access via the NIO APIs.
+ */
+public final class NioTests extends BaseTest {
+
+  @Test
+  public void testGcsEnabled() {
+    FileSystem fs = FileSystems.getFileSystem(URI.create("gs://domain-registry-alpha"));
+  }
+
+  @Test
+  public void openPublicFile() throws IOException {
+    InputStream inputStream = Files.newInputStream(Paths.get(URI.create("gs://pgp-harvard-data-public/hu011C57/GS000018120-DID/GS000015172-ASM/manifest.all.sig")));
+    int firstByte = inputStream.read();
+  }
+
+  @Test
+  public void OpenPrivateFile() throws IOException {
+    // this file, potentially unlike the others in the set, is not marked as "Public link".
+    final String privateFile = getGCPTestInputPath() + "org/broadinstitute/hellbender/utils/private_file.txt";
+    final String BUCKET = privateFile.split("/")[2];
+    FileSystem fs = getAuthenticatedGcs(BUCKET);
+    Path path = fs.getPath(privateFile);
+    int firstByte = Files.newInputStream(path).read();
+  }
+
+  private FileSystem getAuthenticatedGcs(String bucket) throws IOException {
+    // 1. Read credentials from disk, as per
+    //    https://github.com/GoogleCloudPlatform/gcloud-java#authentication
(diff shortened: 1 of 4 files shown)


jean-philippe-martin:
Travis failed, which is expected at this point (because of the setup requirements).

lbergelson:
Neat!  Thanks JP.

src/test/java/org/broadinstitute/hellbender/utils/nio/NioTests.java
```diff
@@ -0,0 +1,73 @@
...
+  public void OpenPrivateFile() throws IOException {
+    // this file, potentially unlike the others in the set, is not marked as "Public link".
+    final String privateFile = getGCPTestInputPath() + "org/broadinstitute/hellbender/utils/private_file.txt";
+    final String BUCKET = privateFile.split("/")[2];
+    FileSystem fs = getAuthenticatedGcs(BUCKET);
+    Path path = fs.getPath(privateFile);
+    int firstByte = Files.newInputStream(path).read();
+  }
+
+  private FileSystem getAuthenticatedGcs(String bucket) throws IOException {
```
lbergelson: Is there any way we can set this globally?  I.e. register the newly authenticated `FileSystem` to be the default for `gs:` paths?  It's a bit awkward if we have to manually specify the file system we want to use.

src/test/java/org/broadinstitute/hellbender/utils/nio/NioTests.java
```diff
@@ -0,0 +1,73 @@
...
+public final class NioTests extends BaseTest {
+
+  @Test
+  public void testGcsEnabled() {
+    FileSystem fs = FileSystems.getFileSystem(URI.create("gs://domain-registry-alpha"));
+  }
+
+  @Test
+  public void openPublicFile() throws IOException {
+    InputStream inputStream = Files.newInputStream(Paths.get(URI.create("gs://pgp-harvard-data-public/hu011C57/GS000018120-DID/GS000015172-ASM/manifest.all.sig")));
```
lbergelson: From the doc it looks like it's necessary to always instantiate paths using the `Paths.get(URI)` overload rather than the `Paths.get(String .. String)`, otherwise the default provider is assumed.  (Not so much a question for you JP, just a note to myself and anyone else reading this)

jean-philippe-martin: You're correct. I agree that it's unfortunate.

src/test/java/org/broadinstitute/hellbender/utils/nio/NioTests.java
jean-philippe-martin: I'll have to investigate.

lbergelson: It looks like it might be possible to call:
` CloudStorageFileSystemProvider.setGCloudOptions(authenticatedStorageOptions);`

I'm not sure if FileSystems are created on demand or if they are initialized eagerly, and setting the provider options during execution may be too late.

src/test/java/org/broadinstitute/hellbender/utils/nio/NioTests.java
lbergelson: I wonder if hadoop's Path has the same distinction.  That could explain some of our weird bugs.  I should check that since we use strings and paths and uri's pretty interchangeably for a lot of things.  If the different getters have different meanings that could be problematic in our existing code..

lbergelson:
@jean-philippe-martin I've uploaded a snapshot to our artifactory.  I pushed a new branch with what I believe are all the necessary change here: https://github.com/broadinstitute/gatk/tree/lb_nio_updates

We'll see how it does in travis..

lbergelson:
in case you don't feel like bothering looking at that branch, add     `compile 'com.google.gcloud:gcloud-java:0.1.6-20160404.190825-1'` to your build.gradle dependencies and it should hopefully work.  Still waiting on travis to find out.

lbergelson:
@jean-philippe-martin Tests fail in my branch, but they all look to be legitimate failures, missing environment variables and such.  https://storage.googleapis.com/hellbender/test/build_reports/5237.2/tests/index.html

jean-philippe-martin:
Thanks @lbergelson! You were fast, I didn't even have time to put the files in place. I suspect that's why things failed. They're there, now. The other thing is that `explicitCredentialsAreNotKept` is going to fail if we're logged into gcloud (I suggest we just disable this test by default).

jean-philippe-martin:
Not sure how to get Travis to take up your branch again, so I made a pull request (#1691)
https://github.com/broadinstitute/gatk/pull/1691

even though I don't actually intend to merge this yet (it should stay in a separate branch because it changes the repository)

lbergelson:
@jean-philippe-martin since you have push access to this repo you should be able to force travis to rerun a branch if you're logged in

jean-philippe-martin:
@lbergelson I think you're saying "if you push to my branch, then Travis will build it even if you don't make a pull request. You can see the results from the Travis page". Is that right?

jean-philippe-martin:
Done via the other branch, no need for this one anymore.

Issue #1557, opened 2016-03-07
remove BAQ from BQSR (and validate results)


akiezun:
- [x] remove BAQ from code (provide a code path that skips BAQ)
- [x] run BaseRecalibrator skipping BAQ
- [x] run ApplyBQSR 
- [x] compare base quals to a BAQ-full run
- [x] compare base quals after binning

akiezun:
BAQ skipping branch - https://github.com/broadinstitute/gatk/compare/ak_removeBAQ

akiezun:
ongoing analysis in gsa6:/local/akiezun/gatk4_bqsr_tieout

akiezun:
ongoing analyses in gsa6:/local/akiezun/gatk4_bqsr_deleteBAQ/gatk

akiezun:
update, on small data (our file in git lfs), results with and without baq (respectively, QRead1, QRead2) are small differences (i'm updating the compare tool to do binning), though most bases gain 1 qual.

```
-----------CompareMatrix summary------------
diff	count
-2	4141872
-1	40845652
0	20910238
1	217848

---------CompareMatrix full matrix (non-zero entries) ----------
QRead1	QRead2	diff	count
1	1	0	1397
2	2	0	9335068
3	3	0	5060
4	4	0	18870
5	5	0	18478
7	8	-1	44
8	8	0	85
8	9	-1	132
9	9	0	1664
9	10	-1	875
10	10	0	7777
10	11	-1	3957
11	11	0	13073
11	12	-1	4741
12	12	0	14906
12	13	-1	6591
13	13	0	12765
13	14	-1	4113
14	14	0	12414
14	15	-1	1830
15	15	0	9595
15	16	-1	2225
16	16	0	7853
16	17	-1	5781
17	17	0	4089
17	18	-1	5891
18	17	1	294
18	18	0	6930
18	19	-1	8422
19	19	0	7357
19	20	-1	20324
20	20	0	8385
20	21	-1	24894
21	21	0	13942
21	22	-1	47819
21	23	-2	752
22	21	1	444
22	22	0	62260
22	23	-1	128662
22	24	-2	903
23	22	1	1466
23	23	0	110413
23	24	-1	271693
24	23	1	33597
24	24	0	534204
24	25	-1	1108115
24	26	-2	1352
25	24	1	33405
25	25	0	1291203
25	26	-1	2365657
25	27	-2	87558
26	26	0	856679
26	27	-1	5627145
[...]

PR #1664, opened 2016-03-29
remove BAQ - provide an option to skip BAQ

akiezun:
this PR adds a hidden option to skip BAQ. This is used in the ongoing investigation of BAQ removal in https://github.com/broadinstitute/gatk/issues/1557

@droazen have a look. I could add a test but the recal table would need to be taken on faith.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/recalibration/BaseRecalibrationEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/recalibration/BaseRecalibrationEngine.java
@@ -83,7 +83,11 @@ public final class BaseRecalibrationEngine implements Serializable {
         this.recalArgs = recalArgs;
         this.readsHeader = readsHeader;
 
-        baq = new BAQ(recalArgs.BAQGOP); // setup the BAQ object with the provided gap open penalty
+        if (recalArgs.skipBAQ) {
+            baq = null;
+        } else {
+            baq = new BAQ(recalArgs.BAQGOP); // setup the BAQ object with the provided gap open penalty
+        }
 
         covariates = new StandardCovariateList(recalArgs, readsHeader);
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/utils/recalibration/BaseRecalibrationEngine.java
```diff
@@ -120,11 +124,18 @@ public void processRead( final GATKRead originalRead, final ReferenceDataSource
...
+
+        //Note: this function modifies the isSNP, isInsertion and isDeletion arguments so it can't be skipped, BAQ or no BAQ
+        final int nErrors = calculateIsSNPOrIndel(read, refDS, isSNP, isInsertion, isDeletion);
+        if (recalArgs.skipBAQ){
+            baqArray = flatBAQArray(read);
+        } else {
+            // note for efficiency reasons we don't compute the BAQ array unless we actually have
+            // some error to marginalize over.  For ILMN data ~85% of reads have no error
+            baqArray = nErrors == 0 ? flatBAQArray(read) : calculateBAQArray(read, refDS);
+        }
```
droazen: Simplify the above to the one-liner:

    final byte[] baqArray = (nErrors == 0 || recalArgs.skipBAQ) ? flatBAQArray(read) : calculateBAQArray(read, refDS);

src/main/java/org/broadinstitute/hellbender/utils/recalibration/RecalibrationArgumentCollection.java
```diff
@@ -99,6 +99,9 @@
     @Argument(fullName = "preserve_qscores_less_than", shortName = "preserveQ", doc = "Don't recalibrate bases with quality scores less than this threshold (with -" + StandardArgumentDefinitions.BQSR_TABLE_SHORT_NAME + ")", optional = true)
     public int PRESERVE_QSCORES_LESS_THAN = QualityUtils.MIN_USABLE_Q_SCORE;
 
+    @HiddenOption
```
droazen: Use `@Hidden` rather than `@HiddenOption` (neither actually hides the argument yet, but GATK3 used `@Hidden`, and sticking with that might make the GATKDocs code easier to port)

droazen:
Forgot to mention when I reviewed this 2 days ago: review complete, back to @akiezun  :)

src/main/java/org/broadinstitute/hellbender/utils/recalibration/BaseRecalibrationEngine.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/utils/recalibration/RecalibrationArgumentCollection.java
akiezun: done

PR #1667, opened 2016-03-29
More Spark CRAM writing tests plus misc changes.

cmnbroad:
Most of the CRAM tests were added before we could write CRAM on Spark, so this fills out the CRAM write tests for the remaining Spark tools/pipelines (ApplyBQSRSpark, BQSRPipelineSpark and ReadsPipelineSpark). Also includes one totally opportunistic deletion of an unused zero-length fasta file.


droazen:
@akiezun do you have time to review this one?

akiezun:
sure

akiezun:
hmm. there's going to be a problem with deleting SHUFFLE from spark tools because cram can't use the 2bit reference. Let's chat about what to do about it.

akiezun:
👍  rebase and merge

PR #1668, opened 2016-03-29
removing beta from our gcloud dataproc script

lbergelson:
dataproc is no longer beta, so this is not needed anymore


droazen:
:+1: merge when ready

PR #1678, opened 2016-04-01
fixed comments about low weight pruning

davidbenjamin:
All comments are now consistent with actual behavior in `LowWeightChainPruner`.  No actual code was changed except for which parts of the graph are given dotted lines in the debugging output.  This change,  from `<=` to `<` is also to make everything consistent with `LowWeightChainPruner`.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/BaseGraph.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/BaseGraph.java
@@ -376,7 +376,7 @@ public abstract class BaseGraph<V extends BaseVertex, E extends BaseEdge> extend
         for( final E edge : edgeSet() ) {
             final String edgeString =  String.format("\t%s -> %s ", getEdgeSource(edge).toString(), getEdgeTarget(edge).toString());
             final String edgeLabelString;
-            if (edge.getMultiplicity() > 0 && edge.getMultiplicity() <= pruneFactor){
+            if (edge.getMultiplicity() > 0 && edge.getMultiplicity() < pruneFactor){
                 edgeLabelString = String.format("[style=dotted,color=grey,label=\"%s\"];", edge.getDotLabel());
             } else {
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/BaseGraph.java
```diff
@@ -376,7 +376,7 @@ public final void printGraph(final PrintStream graphWriter, final boolean writeH
         for( final E edge : edgeSet() ) {
             final String edgeString =  String.format("\t%s -> %s ", getEdgeSource(edge).toString(), getEdgeTarget(edge).toString());
             final String edgeLabelString;
-            if (edge.getMultiplicity() > 0 && edge.getMultiplicity() <= pruneFactor){
+            if (edge.getMultiplicity() > 0 && edge.getMultiplicity() < pruneFactor){
```
akiezun: is this PR only about comments?
what about this code change?

davidbenjamin:
@SHuang-Broad as someone familiar with graph-based local assembly, would you mind reviewing this?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/BaseGraph.java
davidbenjamin: This is in `BaseGraph::printGraph`, which is used exclusively for debugging if I understand correctly.

akiezun: just checking. so this code change is as intended?

davidbenjamin: Yes, everything should change to match `LowWeightChainPruner` because it is used in `HaplotypeCaller`'s assembly step.

akiezun:
:+1:

PR #1681, opened 2016-04-01
packaging avx code during spark and shadowJar

### LINKED ISSUE #1576: Native PairHMM so/dylib doesn't get bundled in the shadow jar

lbergelson:
previously avx code was sometimes included if installDist had been run prior to running sparkJar, now sparkJar will always contain the native code
fixes #1576

also changing download of inteldeflator.so to only happen if it doesn't exist already

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -97,70 +97,6 @@ compileTestJava {
   options.compilerArgs = ['-proc:none', '-Xlint:all','-Werror','-Xdiags:verbose']
 }
 
-model {
-  components {
-    VectorLoglessPairHMM(NativeLibrarySpec) {
-      // set location of source files
-      sources.cpp.source {
-        srcDirs "src/main/cpp/VectorLoglessPairHMM"
-        include "*.cc"
-      }
-
-      binaries.withType(SharedLibraryBinarySpec) { binary ->
-        if (System.env.GATK_SKIP_NATIVE_BUILD.toString().toBoolean()) {
-          buildable = false
-          println("INFO: \$GATK_SKIP_NATIVE_BUILD is true, skipping native AVX PairHMM library build.")
-        } else if (!OperatingSystem.current().isLinux() && !OperatingSystem.current().isMacOsX()) {
-          buildable = false
-          println("WARNING: Building the native AVX PairHMM library is not supported on this OS.")
-        } else {
-          cppCompiler.args "-I", "${System.properties['java.home']}/../include"
-          cppCompiler.args "-mavx"
-          cppCompiler.args "-Wall"
-
-          // build with gcov options on Travis
-          if (System.env.CI.toString().toBoolean()) {
-            cppCompiler.args "-g"
-            cppCompiler.args "-O0"
-            cppCompiler.args "--coverage"
-            linker.args "--coverage"
-          } else {
-            cppCompiler.args "-O3"
-          }
-
-          if (OperatingSystem.current().isMacOsX()) {
-            cppCompiler.args "-I", "${System.properties['java.home']}/../include/Darwin"
-          } else {
-            cppCompiler.args "-I", "${System.properties['java.home']}/../include/linux"
-            linker.args "-static-libgcc"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -455,4 +392,70 @@ installDist.dependsOn downloadGsaLibFile, downloadIntelDeflater
 build.dependsOn installDist
 check.dependsOn installDist
 
+model {
```
lbergelson: This had to move lower because the shadow and sparkJar tasks weren't defined yet where it was initially.  It would be better to put the dependOn block in them, but I can't figure out how to reference the tasks defined in the model, I'm sure it's doable, but I can't figure out the magic invocation needed.

build.gradle
```diff
@@ -455,4 +392,70 @@ installDist.dependsOn downloadGsaLibFile, downloadIntelDeflater
...
+                        cppCompiler.args "-I", "${System.properties['java.home']}/../include/Darwin"
+                    } else {
+                        cppCompiler.args "-I", "${System.properties['java.home']}/../include/linux"
+                        linker.args "-static-libgcc"
+                    }
+
+                    task copySharedLib(type: Copy) {
+                        from binary.sharedLibraryFile
+                        into "$compileJava.destinationDir/lib"
+                        dependsOn binary.tasks
```
lbergelson: The only changes here should be that I moved the `copySharedLib.dependsOn binary.task`s into this block, and that we now have `dependsOn` statements for `shadowJar` and `sparkJar` as well as `jar`.  (I couldn't get it to for for `tasks.ofType(Jar)`) if someone knows more I'd be happy to have help.)

build.gradle
```diff
@@ -455,4 +392,71 @@ installDist.dependsOn downloadGsaLibFile, downloadIntelDeflater
...
+                    }
+
+                    jar.dependsOn copySharedLib
+                    shadowJar.dependsOn copySharedLib
+                    sparkJar.dependsOn copySharedLib
+                    test.dependsOn copySharedLib
```
lbergelson: test now depends on it too which makes sense I believe

build.gradle
```diff
@@ -21,9 +21,9 @@ plugins {
 
 }
 
+
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import de.undercouch.gradle.tasks.download.Download
-import org.gradle.internal.os.OperatingSystem
```
droazen: Why not import this? You are still using it below

build.gradle
```diff
@@ -47,6 +47,7 @@ task downloadIntelDeflater(type: Download) {
     //IntelDeflater is not packages with the rest of HTSJDK so we download it by hand
     src 'https://github.com/samtools/htsjdk/raw/2.1.1/lib/jni/libIntelDeflater.so'
     dest 'build/'
+    overwrite false
```
droazen: We don't want to overwrite old copies of the library here?

build.gradle
lbergelson: oh, huh, for some reason intellij fully qualified the names below and dropped the import...  Will fix

droazen:
:+1: merge after addressing comments -- back to @lbergelson

build.gradle
lbergelson: it's a tradeoff between downloading a fresh copy every single time (which extends the installDist time by a reasonable amount) or to only download a fresh copy after a gradle clean (which seems reasonablish to me...)  I don't know if it tracks changes to the src url.  It's possible that it will always redownload if the source url changes which is the ideal situtation.  I'll experiment and see.

build.gradle
lbergelson: done

build.gradle
lbergelson: It doesn't work the way I had hoped.  I've opened an issue with the download plugin authors to see if there is a way to achieve my desired behavior. https://github.com/michel-kraemer/gradle-download-task/issues/42

Reverting.

PR #1695, opened 2016-04-08
asynch IO back on

### LINKED ISSUE #1653: Re-activate async I/O for the samtools package only (not tribble) after next htsjdk release
now that we're on htsjdk 2.2.1 we should switch to asyncIO for bams (not tribble)

- [x] switch to asyncIO for bams (build.gradle)
- [x] switch tests to use asyncIO for bams (build.gradle)
- [x] update readme to say that we're using async IO
- [x] update startup message to clarify which IO is sync/async

measure and report performance impact on (using JdkDeflater and IntelDeflater)
 - [x] PrintReads 
 - [x] BaseRecalibrator

akiezun:
fixes #1653

ok, results are in, with asyncIO GATK4 out of the box beats GATK3 4.53x on PrintReads and 1.48x on BaseRecalibrator!

PrintReads
syncTribble, asyncSamtools, IntelDeflater: 1x baseline (this is the out-of-the box GATK4 config)
syncTribble,  syncSamtools, IntelDeflater:  1.32x baseline
syncTribble, asyncSamtools, JDKDeflater: 1.68x baseline 
syncTribble,  syncSamtools, JDKDeflater:  2.32x baseline

For comparison, GATK3 out of the box is 4.53x baseline on PrintReads

BaseRecalibrator
syncTribble, asyncSamtools, IntelDeflater: 1x baseline (this is the out-of-the box GATK4 config)
syncTribble,  syncSamtools, IntelDeflater:  1x baseline
asyncTribble, asyncSamtools, IntelDeflater:  2.14x baseline
syncTribble, asyncSamtools, JDKDeflater: 1.04x baseline
syncTribble,  syncSamtools, JDKDeflater:  1.03x baseline
asyncTribble, asyncSamtools, JDKDeflater: 2.12x baseline

For comparison, GATK3 out of the box is 1.48x baseline on BaseRecalibrator

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,7 +32,7 @@ import org.gradle.internal.os.OperatingSystem
 mainClassName = "org.broadinstitute.hellbender.Main"
 
 //Note: the test suite must use the same defaults. If you change system properties in this list you must also update the one in the test task
-applicationDefaultJvmArgs = ["-Dsamjdk.use_async_io=false", "-Dsamjdk.intel_deflater_so_path=build/libIntelDeflater.so", "-Dsamjdk.compression_level=1"]
+applicationDefaultJvmArgs = ["-Dsamjdk.use_async_io_samtools=true", "-Dsamjdk.use_async_io_tribble=false", "-Dsamjdk.intel_deflater_so_path=build/libIntelDeflater.so", "-Dsamjdk.compression_level=1"]
 
 //Delete the windows script - we never test on Windows so let's not pretend it works
 startScripts {
@@ -223,7 +223,8 @@ test {
         }
     }
 
-    systemProperty "samjdk.use_async_io", "false"
+    systemProperty "samjdk.use_async_io_samtools", "true"
+    systemProperty "samjdk.use_async_io_tribble", "false"
     systemProperty "samjdk.intel_deflater_so_path", "build/libIntelDeflater.so"
     systemProperty "samjdk.compression_level", "1"
     systemProperty "gatk.spark.debug", System.getProperty("gatk.spark.debug")
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -140,10 +140,13 @@ public abstract class CommandLineProgram {
                         " on " + System.getProperty("os.name") + " " + System.getProperty("os.version") +
                         " " + System.getProperty("os.arch") + "; " + System.getProperty("java.vm.name") +
                         " " + System.getProperty("java.runtime.version") +
-                        "; Version: " + commandLineParser.getVersion() +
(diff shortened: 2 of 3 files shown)


droazen:
Should we create a story to allow Spark tools to take advantage of this speedup at the Hadoop-BAM layer, @akiezun? (if possible)

akiezun:
is it even possible? @tomwhite how would async IO interplay with spark/yarn? 
or are you talking about IntelDeflater + compression level 1, @droazen ?

droazen:
@akiezun @tomwhite Well, I'm not sure it's possible, but since Hadoop-BAM relies on htsjdk we should at least ask the question :)

lbergelson:
Nice speedup!

droazen:
Created https://github.com/broadinstitute/gatk/issues/1696

droazen:
Headline is a bit misleading here, since this only helps `PrintReads` and not `BQSR` :) But it's good that the tribble implementation is no longer causing problems.

build.gradle
```diff
@@ -223,7 +223,8 @@ test {
         }
     }
 
-    systemProperty "samjdk.use_async_io", "false"
+    systemProperty "samjdk.use_async_io_samtools", "true"
+    systemProperty "samjdk.use_async_io_tribble", "false"
     systemProperty "samjdk.intel_deflater_so_path", "build/libIntelDeflater.so"
     systemProperty "samjdk.compression_level", "1"
```
droazen: Would be good if we could somehow enforce consistency between these `systemProperty` settings here and `applicationDefaultJvmArgs` above.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -141,7 +141,8 @@ public Object instanceMainPostParseArgs() {
                         " " + System.getProperty("os.arch") + "; " + System.getProperty("java.vm.name") +
                         " " + System.getProperty("java.runtime.version") +
                         "; Version: " + commandLineParser.getVersion() +
-                        " " + (Defaults.USE_ASYNC_IO ? "asyncIO": "syncIO") +
+                        " " + (Defaults.USE_ASYNC_IO_FOR_SAMTOOLS ? "asyncIOBam": "syncIOBam") +
+                        " " + (Defaults.USE_ASYNC_IO_FOR_TRIBBLE ? "asyncIOTribble": "syncIOTribble") +
```
droazen: This shouldn't be here in the main program name / version String. Recommend instead outputting a `logger.info()` message with the async settings, which is more consistent with the way we communicate other engine-level settings in GATK.

droazen:
Review complete, back to @akiezun. Merge after addressing comments and tests pass.

build.gradle
akiezun: it would be nice indeed. Let's leave if for later.

PR #1706, opened 2016-04-13
updating htsjdk to current snapshop for asynch fix

lbergelson:
also updating build to extract the intelDeflater from the jar instead of downloading it
removed the htsjdkVersion property I had recently added since it was no longer necessary with the update to include the .so in the jar


akiezun:
👍

PR #1528, opened 2016-02-26
Implemented SlidingWindowWalker

magicDGS:
As requested in #1198, I would like to have a walker to iterate over data sources with a set of overlapping windows based on a reference genome (that is, a `SAMSequenceDictionary`).

This first implementation is a very naive one, because it only construct the overlapping windows from the reference (keeping only the ones that overlaps with the provided `intervalsForTraversal`) and then just query in the different sources each of the windows generated.

The next plan is to extends this class and `VariantWalker` to create a `SlidingWindowVariantWalker`, which traverse variants using this sliding-window approach.

Issue #1198, opened 2015-11-29
Request: SlidingWindow walker


magicDGS:
Hello, I would like to ask for the implementation of a SlidingWindowWalker (both for reads and variants), that could be very interesting for other tools. I was thinking that it could be similar to IntervalWalker, but generating the intervals internally using a window size and step size as parameters (provided by the user and without merging overlapping intervals, like suggested on #302). I think that this issue is different from issue #10 and the pull request #890 because it is a more general Walker that could be abstract and extendable by other tools.

lbergelson:
Hi @magicDGS.  This is something we would like to implement, but isn't yet on our timeline.  Do you have use cases for both reads and variants?  We have some ideas for sliding window walkers along reads, but we don't have any examples of when this would be useful for variants.

magicDGS:
Hi @lbergelson, for sliding window along reads some use cases are computing CNVs or mapping statistics over windows; for variants, compute statistics over windows like _Fst_ or heterozygosity is helpful for genomic scans. Probably is not a main goal for the GATK tools, but it will be important for the framework for being used by other software (actually I asked for it because I'm planning to use the framework in a new software).

On the other hand, is it possible to have an idea about the timeline for GATK4?

Thank you very much for everything.

lbergelson:
So the long term timeline is fairly up in the air.  You can check out the alpha and beta milestones for some idea of what's been prioritized.  Alpha milestone is due for completion ~this week.  Beta is much more up in the air and will depend at least in part on feedback from user.  

Incidentally, if you're interested in CNV calling take a look at https://github.com/broadinstitute/gatk-protected/ which has some tools for CNV calling built on this engine.

magicDGS:
Thank you, @lbergelson. Regarding the CNV, I'm not looking for CNV calling _sensu stricto_, I've just give an example. I'm more interested in the SlidingWindow for variants for my software.

I will be interested in collaborate implementing the sliding window walker, but I think that I will need to go too mucho into the details for the framework and it will be more difficult for me.

lbergelson:
Well, we're definitely happy to accept contributions to the code base should you be interested in trying to implement it yourself.  :)  

I estimate that implementing the sliding window walker will be relatively easy to do, but making it perform well may be more difficult since the necessary caching for efficient traversal may not be implemented in the engine yet.  

[...]

PR #1708, opened 2016-04-14
Implemented SlidingWindowWalker

magicDGS:
New implementation of `SlidingWindowWalker` with some ideas from the discussion in #1528.

The thinks that are requested in #1198 still holds, but now it is more general: padding option is added and construction of windows are done by interval.

The code contain a lot of TODO because it relies on changes implemented in #1567, and because it is suppose to be a walker over `ReadWindow` instead of `SimpleInterval`+`ReadsContext` if reads are available.

I think that with these changes it could be general to be extended by `ReadWindowWalker` and by users that needs a different way of "slide" over intervals.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/LocalReadShard.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/LocalReadShard.java
@@ -1,7 +1,6 @@
 package org.broadinstitute.hellbender.engine;
 
 import htsjdk.samtools.SAMSequenceDictionary;
-import htsjdk.samtools.util.Locatable;
 import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.Utils;
@@ -12,10 +11,7 @@ import org.broadinstitute.hellbender.utils.read.GATKRead;
 
 import java.util.Iterator;
 import java.util.List;
-import java.util.Spliterator;
-import java.util.Spliterators;
 import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
 /**
  * A class to represent a shard of reads data, optionally expanded by a configurable amount of padded data.
@@ -101,38 +97,6 @@ public final class LocalReadShard implements Shard<GATKRead> {
     }
 
     /**
-     * @return number of bases of padding to the left of our interval
-     */
-    public int numLeftPaddingBases() {
-        return interval.getStart() - paddedInterval.getStart();
-    }
-
-    /**
-     * @return number of bases of padding to the right of our interval
-     */
-    public int numRightPaddingBases() {
-        return paddedInterval.getEnd() - interval.getEnd();
-    }
-
-    /**
-     * @param loc Locatable to test
-     * @return true if loc is completely contained within this shard's interval, otherwise false
-     */
-    public boolean contains(final Locatable loc) {
-        Utils.nonNull(loc);
-        return interval.contains(loc);
-    }
-
-    /**
-     * @param loc Locatable to test
-     * @return true if loc starts within this shard's interval, otherwise false
-     */
-    public boolean containsStartPosition(final Locatable loc) {
-        Utils.nonNull(loc);
-        return interval.contains(new SimpleInterval(loc.getContig(), loc.getStart(), loc.getStart()));
-    }
-
-    /**
      * @return an iterator over reads in this shard, as filtered using the configured read filter
      *         and downsampled using the configured downsampler; reads are lazily loaded rather than pre-loaded
      *
@@ -162,7 +126,7 @@ public final class LocalReadShard implements Shard<GATKRead> {
      * Note that any read filtering is always performed before any downsampling.
      */
     public List<GATKRead> loadAllReads() {
-        return StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterator(), Spliterator.ORDERED), false).collect(Collectors.toList());
+        return loadAllRecords();
     }
 
     /**
--- a/src/test/java/org/broadinstitute/hellbender/cmdline/argumentcollections/WindowPaddingArgumentCollectionTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/cmdline/argumentcollections/WindowPaddingArgumentCollectionTest.java
@@ -1,0 +1,121 @@
+package org.broadinstitute.hellbender.cmdline.argumentcollections;
+
+import org.apache.commons.lang3.ArrayUtils;
+import org.broadinstitute.hellbender.cmdline.ArgumentCollection;
+import org.broadinstitute.hellbender.cmdline.CommandLineParser;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class WindowPaddingArgumentCollectionTest extends BaseTest{
+
+    private static class FakeToolWPadding {
+        @ArgumentCollection
+        final WindowPaddingArgumentCollection wp;
+
+        FakeToolWPadding(final WindowPaddingArgumentCollection wp) {
+            this.wp = wp;
+        }
+    }
+
+    @DataProvider(name="illegalWPadding")
+    public Object[][] getIllegalPadding() {
+        return new Object[][] {{-10}, {-5}, {-1}};
+    }
+
+    @DataProvider(name="legalWPadding")
+    public Object[][] getLegalPadding() {
+        return new Object[][] {{0}, {1}, {5}};
+    }
+
+    @Test(dataProvider = "illegalWPadding", expectedExceptions = IllegalArgumentException.class)
+    public void testIllegalDefaultWindowSize(final int pad) {
+        final WindowPaddingArgumentCollection ac = new DefaultWindowPaddingArgumentCollection(pad);
+        System.err.println("Window padding: "+ac.getWindowPadding());
+    }
+
+    @Test(dataProvider = "illegalWPadding", expectedExceptions = IllegalArgumentException.class)
+    public void testIllegalRequiredWindowSize(final int pad) {
+        final WindowPaddingArgumentCollection ac = new OptionalWindowPaddingArgumentCollection(pad);
+        System.err.println("Window padding: "+ac.getWindowPadding());
+    }
+
+    @Test(dataProvider = "legalWPadding")
+    public void testLegalDefaultWindowSize(final int pad) {
+        final WindowPaddingArgumentCollection ac = new DefaultWindowPaddingArgumentCollection(pad);
+        Assert.assertEquals(ac.getWindowPadding(), pad);
+    }
+
+    @Test(dataProvider = "legalWPadding")
+    public void testLegalRequiredWindowSize(final int pad) {
+        final WindowPaddingArgumentCollection ac = new OptionalWindowPaddingArgumentCollection(pad);
+        Assert.assertEquals(ac.getWindowPadding(), pad);
+    }
+
+    @DataProvider(name="defaultWPadding")
+    private Object[][] withDefault() {
+        return new Object[][]{
+                {new FakeToolWPadding(new DefaultWindowPaddingArgumentCollection(10)), 10},
+                {new FakeToolWPadding(new DefaultWindowPaddingArgumentCollection(5)),  5}};
+    }
+
+    @DataProvider(name="requiredWPadding")
+    private Object[][] withRequired() {
+        return new Object[][]{
+                {new FakeToolWPadding(new OptionalWindowPaddingArgumentCollection(10)), 10},
+                {new FakeToolWPadding(new OptionalWindowPaddingArgumentCollection(5)),   5}};
+    }
+
+    @DataProvider(name="bothWPadding")
+    private Object[][] getBoth() {
+        return ArrayUtils.addAll(withDefault(), withRequired());
+    }
+
+    @Test(dataProvider = "bothWPadding")
+    public void testNoOptionProvided(final FakeToolWPadding tool, final int defaultValue){
+        String[] args = {};
+        CommandLineParser clp = new CommandLineParser(tool);
+        clp.parseArguments(System.out, args);
+        Assert.assertEquals(tool.wp.getWindowPadding(), defaultValue);
+    }
+
+    @Test(dataProvider = "requiredWPadding")
+    public void provideOptionToRequired(final FakeToolWPadding tool, final int defaultValue) {
+        String[] args = {"--windowPadding", String.valueOf(defaultValue+1)};
+        CommandLineParser clp = new CommandLineParser(tool);
+        clp.parseArguments(System.out, args);
+        Assert.assertEquals(tool.wp.getWindowPadding(), defaultValue+1);
+    }
+
+    @Test(dataProvider = "defaultWPadding", expectedExceptions = UserException.CommandLineException.class)
+    public void provideOptionToDefault(final FakeToolWPadding tool, final int defaultValue) {
+        String[] args = {"--windowPadding", String.valueOf(defaultValue+1)};
+        CommandLineParser clp = new CommandLineParser(tool);
+        clp.parseArguments(System.out, args);
+    }
+
+    @Test(dataProvider = "illegalWPadding", expectedExceptions = UserException.BadArgumentValue.class)
+    public void provideIllegalValue(final int pad) {
+        final FakeToolWPadding tool = new FakeToolWPadding(new OptionalWindowPaddingArgumentCollection(1000));
+        String[] args = {"--windowPadding", String.valueOf(pad)};
+        CommandLineParser clp = new CommandLineParser(tool);
+        clp.parseArguments(System.out, args);
+        int userPadding = tool.wp.getWindowPadding();
+        System.err.println("User padding: "+userPadding);
+    }
+
+    @Test(dataProvider = "legalWPadding")
+    public void provideLegalValue(final int pad) {
+        final FakeToolWPadding tool = new FakeToolWPadding(new OptionalWindowPaddingArgumentCollection(1000));
+        String[] args = {"--windowPadding", String.valueOf(pad)};
+        CommandLineParser clp = new CommandLineParser(tool);
+        clp.parseArguments(System.out, args);
+        int userPadding = tool.wp.getWindowPadding();
+        Assert.assertEquals(userPadding, pad);
+    }
+}
--- a/src/test/java/org/broadinstitute/hellbender/cmdline/argumentcollections/WindowSizeArgumentCollectionTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/cmdline/argumentcollections/WindowSizeArgumentCollectionTest.java
@@ -1,0 +1,121 @@
+package org.broadinstitute.hellbender.cmdline.argumentcollections;
+
+import org.apache.commons.lang3.ArrayUtils;
+import org.broadinstitute.hellbender.cmdline.ArgumentCollection;
+import org.broadinstitute.hellbender.cmdline.CommandLineParser;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class WindowSizeArgumentCollectionTest extends BaseTest{
+
+    private static class FakeToolWSize {
+        @ArgumentCollection
+        final WindowSizeArgumentCollection ws;
+
+        FakeToolWSize(final WindowSizeArgumentCollection ws) {
+            this.ws = ws;
+        }
+    }
+
+    @DataProvider(name="illegalWSize")
+    public Object[][] getIllegalSizes() {
+        return new Object[][] {{-10}, {-1}, {0}};
+    }
+
+    @DataProvider(name="legalWSize")
+    public Object[][] getLegalSizes() {
+        return new Object[][] {{1}, {5}, {10}};
+    }
+
+    @Test(dataProvider = "illegalWSize", expectedExceptions = IllegalArgumentException.class)
+    public void testIllegalDefaultWindowSize(final int ws) {
+        final WindowSizeArgumentCollection ac = new DefaultWindowSizeArgumentCollection(ws);
+        System.err.println("Window size: "+ac.getWindowSize());
+    }
+
+    @Test(dataProvider = "illegalWSize", expectedExceptions = IllegalArgumentException.class)
+    public void testIllegalRequiredWindowSize(final int ws) {
+        final WindowSizeArgumentCollection ac = new OptionalWindowSizeArgumentCollection(ws);
+        System.err.println("Window size: "+ac.getWindowSize());
+    }
+
+    @Test(dataProvider = "legalWSize")
+    public void testLegalDefaultWindowSize(final int ws) {
+        final WindowSizeArgumentCollection ac = new DefaultWindowSizeArgumentCollection(ws);
+        Assert.assertEquals(ac.getWindowSize(), ws);
+    }
+
+    @Test(dataProvider = "legalWSize")
+    public void testLegalRequiredWindowSize(final int ws) {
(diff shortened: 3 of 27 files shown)


magicDGS:
I was thinking after checking your `ReadWindowWalker` implementation and the HaplotypeCaller branch that with this class we can do a general `SlidingWindowWalker`, which could be extended by `ReadWindowWalker` or it could be possible that `HaplotypeCaller` directly implements `SlidingWindowWalker`.

@droazen, could you have a look and tell me what do you think about the idea?

droazen:
@magicDGS The HaplotypeCaller traversal has undergone some changes in the past few weeks to improve performance and bring the output of the tool closer to GATK3. There is now an `AssemblyRegionWalker` that divides the intervals into active and inactive regions, in a greatly simplified version of the GATK3 traversal.

Initially, I did plan on having `AssemblyRegionWalker` extend the former `ReadWindowWalker`, or an adapted version of your `SlidingWindowWalker`, and I did implement it like this at first, but ultimately I collapsed it into a single class for several reasons:

* Inheriting from a more generic traversal type caused usability issues and confusion with respect to the command-line arguments. The `ReadWindow` was the unit of processing for the superclass, but for `AssemblyRegionWalker` it was the unit of I/O and `AssemblyRegion` was the unit of processing, and I couldn't update the docs for `ReadWindowWalker` to clear up the confusion without mentioning `AssemblyRegion`-specific concepts.

* The `ReadShard` / `ReadWindow` was/is **only** there to prove that we can shard the data without introducing calling artifacts, and to provide a unit of parallelism for the upcoming Spark implementation. It's not something we really want to expose to users as a prominent knob, and we may hide it completely in the future once the shard size is tuned for performance.

* Inheriting from a more abstract walker type caused a number of other problems as well: methods that should have been final in the supertype could no longer be made final, with the result that tool implementations could inappropriately override engine initialization/shutdown routines. Also, there were issues with the progress meter, since both the supertype traversal and subtype traversal needed their own progress meter for their different units of processing.

Ultimately it was just too awkward and forced, and the read shard is something that we eventually want to make an internal/encapsulated implementation detail anyway. GATK3 made the mistake, I think, of using long, confusing inheritance chains for its walker types, with the result that you got awkward and forced relationships like `RodWalker` inheriting from `LocusWalker`. It's better, I think, to make each traversal as standalone as possible, especially given the simplicity of writing a new walker type in GATK4.

For all of these reasons we don't want `AssemblyRegionWalker` to inherit from a more abstract traversal type -- it's just going to be its own standalone thing, so that it can evolve freely without affecting anyone else. For `SlidingWindowWalker`, which we still want to merge, I recommend making the traversal do **exactly** what you want for your use case, as clearly and simply as possible, without worrying about serving as a base class for other traversals. Ping me once you're happy with it, and I'll re-review.

magicDGS:
I agree that it is better to keep it simpler. I was proposing this before looking the latest commits in the HC branch.

I will work on this walker without thinking about other cases, but I would like to keep the idea of padding and slide over intervals instead of the genome from the begining. It will be useful for the things that I have in mind.

Should I close this PR until I implement everything, @droazen?

droazen:
@magicDGS You can keep this open -- just push the new version into this same branch and add a comment when you're happy with it and I'll re-review.

magicDGS:
I finished the implementation for the draft `SlidingWindowWalker` (I should implement an example and an integration test, but I would like to wait till some issues are solved).

 made a "TODO" about the way in which the intervals are constructed, because I will need a that `ReadShard` have a way to construct a shard without `ReadSource` (either null or empty source), just in case that the implemented `SlidingWindowWalker` does not require reads. @droazen, could you review and give me some feedback about this, because this class is important for other parts of GATK?

magicDGS:
@droazen, could you review this PR? Before I implement the integration test I would be important that you check if it is possible the change of the `ReadShard` contract (null `ReadsDataSource` are now allowed, and I don't know if this will affect other parts of GATK.

Thank you very much in advance.

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -2,6 +2,7 @@
 
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.util.Locatable;
+import org.apache.avro.test.Simple;
```
droazen: Remove stray import

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
```diff
@@ -0,0 +1,185 @@
...
+import org.broadinstitute.hellbender.engine.filters.WellformedReadFilter;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.IntervalUtils;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * An SlidingWindowWalker is a tool that processes a single window over the genome/intervals at a time, with the ability
```
droazen: `An SlidingWindowWalker` -> `A SlidingWindowWalker`

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
```diff
@@ -0,0 +1,185 @@
...
+ * to query optional overlapping sources of reads, reference data, and/or variants/features.
+ *
+ * The windows are constructed over for each interval with a concrete window and step size, and padding overlapping
```
droazen: `over for each` -> `over each`

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
```diff
@@ -0,0 +1,185 @@
...
+public abstract class SlidingWindowWalker extends GATKTool {
+
+    @Argument(fullName = "disable_all_read_filters", shortName = "f", doc = "Disable all read filters", common = false, optional = true)
+    public boolean disableAllReadFilters = false;
+
+    protected final static String WINDOW_SIZE_ARGUMENT = "windowSize";
+
+    protected final static String WINDOW_STEP_ARGUMENT = "windowStep";
+
+    protected final static String WINDOW_PADDING_ARGUMENT = "windowPadding";
```
droazen: Can you make these into `@Argument`s here in the base class? GATK traversals generally come with the command-line args needed to control them. Tools could provide their own defaults via an overridable method:

       @Argument(fullName="windowSize", shortName="windowSize", doc = "", optional = true)
       protected int windowSize = defaultWindowSize();

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * Get the window size for make the windows
```
droazen: `for make` -> `for making`

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
```diff
@@ -0,0 +1,185 @@
...
+     *
+     * Tools which provides a command-line argument for this parameter should use {@link #WINDOW_SIZE_ARGUMENT}
+     *
+     * @return window-size (must be a positive integer)
+     */
+    protected abstract int getWindowSize();
+
+    /**
+     * Get the window step for make the windows
```
droazen: Explain what window step means, since it's not obvious

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
```diff
@@ -0,0 +1,185 @@
...
+    protected final void onStartup() {
+        super.onStartup();
+        SAMSequenceDictionary dictionary = getBestAvailableSequenceDictionary();
+        // geting window parameters
+        final int windowSize = getWindowSize();
+        final int windowStep = getWindowStep();
+        final int windowPadding = getWindowPadding();
+        // either the
+        if ( windowSize <= 1 ) {
+            throw new UserException.BadArgumentValue("--"+WINDOW_SIZE_ARGUMENT, Integer.toString(windowSize), "window size must be > 0");
```
droazen: Do you mean `must be > 1` here?

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+        if ( windowStep <= 1 ) {
+            throw new UserException.BadArgumentValue("--"+WINDOW_STEP_ARGUMENT, Integer.toString(windowSize), "window step must be > 0");
```
droazen: `must be > 1` ?

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
```diff
@@ -0,0 +1,185 @@
...
+     */
+    @Override
+    protected final void onStartup() {
+        super.onStartup();
+        SAMSequenceDictionary dictionary = getBestAvailableSequenceDictionary();
+        // geting window parameters
+        final int windowSize = getWindowSize();
+        final int windowStep = getWindowStep();
+        final int windowPadding = getWindowPadding();
+        // either the
```
droazen: complete this comment or remove it

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
```diff
@@ -0,0 +1,185 @@
...
+    private List<ReadShard> makeReadShards(final List<SimpleInterval> intervals, final int windowSize, final int windowStep, final int windowPadding, final SAMSequenceDictionary dictionary) {
+        final List<ReadShard> shards = new ArrayList<>();
+        for (final SimpleInterval interval : intervals) {
+            shards.addAll(ReadShard.divideIntervalIntoShards(interval, windowSize, windowStep, windowPadding, reads, dictionary));
+        }
+        return shards;
+    }
+
+    /**
+     * Implementation of window-based traversing. Subclasses can override to provide their own behavior but default
```
droazen: `traversing` -> `traversal`

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -36,12 +37,11 @@
      *
      * @param interval the genomic span covered by this shard
      * @param paddedInterval the span covered by this shard, plus any additional padding on each side (must contain the un-padded interval)
-     * @param readsSource source of reads from which to populate this shard
+     * @param readsSource source of reads from which to populate this shard; it could be null if no data source is available
```
droazen: I think it's ok if you want to allow `ReadShards` to have a null data source, provided you add a `public boolean hasBackingDataSource()` method that returns false when `readsSource == null` to allow the client to query this.

It seems strange, though, to imagine a `SlidingWindowWalker` that does not take reads, and yet takes `ReadShards` in its `apply()` method. I wonder if perhaps we need a `Shard` superclass for `ReadShard` that contains only `interval` and `paddedInterval`? Then your apply method could take a `Shard, ReadsContext, ReferenceContext, FeatureContext`. Or are you happy with the existing implementation?

droazen:
@magicDGS Did a review pass and gave my thoughts on the `ReadShard` issue

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
magicDGS: Yes, I do. Thanks for pointing it out.

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
magicDGS: Done. Also `windowPadding <= 0` is wrong.

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
magicDGS: Removed

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
magicDGS: I'm happy with the implementation because reads are loaded in a lazy way, and that could be important in terms of performance. Nevertheless, my aim was to provide a base class for two classes of `WindowWalker`: one for reads and one for variants, but allowing to do others on demand for reference, other features, etc.

Thus, it is true that is wierd to iterate over `ReadShard` without reads, only needed to keep track of the `interval`/`paddedInterval` in a single classe. Thus, I'm planning to split the implementation in two to classes (`SlidingWindowReadWalker` and `SlidingWindowVariantWalker`), to keep abstractions simple.

By the way, I implemented for the moment the method that you asked for to keep track of the changes and comments.

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
magicDGS: About the arguments, I'm not sure if in this case fit. One of the problems is that some tools implementing a `SlidingWindowWalker` won't like to expose to the final user some parameters. One of the use case that I can imagine is a window walker that does not require any padding; or a walker that requires exactly the same window size and step (as in the case of the `AssemblyRegionWalker`).

Is there a way of hide the `|@Argument` annotations in classes that extends this tool?

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
magicDGS: Done

magicDGS:
Thanks for the help here, @droazen. Now I understand why keeping the walkers separated is a better idea. I will continue working on this soon, separating two `SlidingWindow` for reads and variants, but taking into account your comments.

Do you think that it is better to implement them in two different PRs?

src/main/java/org/broadinstitute/hellbender/engine/SlidingWindowWalker.java
droazen: There's currently no way to hide `@Arguments` directly, but you could effectively do this via `ArgumentCollections`. See, eg., the way `GATKTool` uses `ArgumentCollections` to select between required and optional versions of the same arguments (such as `RequiredReadInputArgumentCollection` vs. `OptionalReadInputArgumentCollection`). You could do the same thing in your `SlidingWindowWalker` to allow tool authors to turn different arguments on and off, by having an "empty" argument collection for each sliding window argument.

This is a bit ugly, but I still think it's better to have the arguments controlling traversal defined in the traversal itself rather than in the tools.

droazen:
@magicDGS It might be best to implement both the reads and variants versions in this PR so that we can make sure the interface makes sense before merging, but it's up to you. Let me know when you're ready for another review.

magicDGS:
This is ready for review again, @droazen. I split the walkers in two: `ReadSliderWalker` and `VariantSliderWalker`, both based on `ReadWalker` and `VariantWalker` to traverse in a sliding-window approach. I also implemented some `ArgumentCollectionDefinition` for the three parameters (window-size, window-step and window-padding).

There are some changes that I would like to discuss with you:

* `ShardSource<T>`: a generic class for lazily load sources of certain type. Now `ReadShard` extends this class, but there is actually no change in the definitions. I wonder if it is worthy to maintain `ReadShard` as a separate class, because it seems that the only usage is in `AssemblyRegionWalker` and it could be easily changed by a `ShardSource<GATKRead>`.
* `FilteringIterator<T>`: another generic class for filter on iteration records. Now `ReadFilteringIterator` extends this class, but again I don't find any usage that requires an specific case instead of a generic.
* There is some repetition in the test code: `ShardSourceUnitTest` and `ReadShardUnitTest` are exactly the same. I didn't want to remove the later, but I think that it is redundant; the same for the `FilteringIteratorUnitTest` adn `ReadFilteringIterator.
* Because of the inclusion of `ArgumentCollectionDefinition`s for window-related parameters, I just want to know if I should include them in `AssemblyRegionWalker` to maintain consistency in the user options.

Back to you @droazen, and thanks again for all the help.

magicDGS:
Can you review when you have time, @droazen? I'll be very thankful if so.

Thanks in advance.

droazen:
Yes, apologies for the delay -- we have several quarterly goals that we're trying to get done within the next 2 weeks. I'll try to make time for this review next week.

magicDGS:
No rush if you are busy. I've just wanted to highlight that this is waiting for revision. Thanks a lot!

magicDGS:
Could it be possible for you to review this if you have time, @droazen?

droazen:
I haven't forgotten -- I'll have time to review this after our alpha-2 release is out.

magicDGS:
Thanks a lot @droazen!!

davidbernick:

magicDGS:
I've just mess up all the history in this PR, sorry... I will try to solve it.

magicDGS:
@davidbernick, I rebased to the current master and solve the conflicts, because the PR was kind of old and new things were implemented (like the downsampler in `ReadShard`.

I think that it is prepared for review, @droazen.

davidbernick:
jenkins test please

magicDGS:
I extracted some classes to a different branch (#2023) to make it easier to review and integrate with the changes for HaplotypeCallerSpark. I will wait for that branch to be accepted to rebase and update this one...

magicDGS:
Hi @droazen, I finished the implementation of the Slider Walkers using the new `Shard` interface from the changes for `HaplotypeCallerSpark` (#2016). I know that it is a lot of code, but I will be very grateful if you can have a look to this PR to accept it.

Thank you in advance!

magicDGS:
Hello @droazen, I rebased the PR to the latest master. Could you have a look to this, please? I will be very thankful if so. Thanks in advance!

magicDGS:
Sorry @droazen, the previous commit had an error in the tests. I'm rebasing/squashing to make a clear PR and when all check pass (except CLOUD), you can review if you have time. Thank you very much.

magicDGS:
Friendly ping here, @droazen! I would appreciate if this could be accepted soon into the framework!

magicDGS:
Hello @droazen. Is there any possibility to review this soon? Thanks in advance!

droazen:
Sorry @magicDGS, we'll get to this (and your other PRs) as soon as we're able -- once we're out of alpha and the GATK3 team joins us here we should be able to respond to external pull requests much more promptly, but until then we'll unfortunately have to live with slow turnaround time on code reviews.

magicDGS:
Thanks for the reply @droazen! I know that it is a lot of work without code that it is not strictly useful within the milestones, and that's why I started this PR time ago (I will need it soon in my work, but I can just use this branch if I rebase from time to time).

I just want to point to this PR again, although I know that it is not trivial code and will take a while to accept it. I'm looking forward for your comments on this. Thanks again!

magicDGS:
Any progress/update with respect to this review, @droazen? Thanks in advance!

src/test/java/org/broadinstitute/hellbender/cmdline/argumentcollections/WindowPaddingArgumentCollectionTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class WindowPaddingArgumentCollectionTest extends BaseTest{
```
magicDGS: Self-reminder: extend `GATKBaseTest` (requires rebase)

src/test/java/org/broadinstitute/hellbender/cmdline/argumentcollections/WindowSizeArgumentCollectionTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class WindowSizeArgumentCollectionTest extends BaseTest{
```
magicDGS: Self-reminder: extend `GATKBaseTest` (requires rebase)

src/test/java/org/broadinstitute/hellbender/cmdline/argumentcollections/WindowStepArgumentCollectionTest.java
```diff
@@ -0,0 +1,122 @@
...
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class WindowStepArgumentCollectionTest extends BaseTest{
```
magicDGS: Self-reminder: extend `GATKBaseTest` (requires rebase)

src/test/java/org/broadinstitute/hellbender/utils/iterators/FilteringIteratorUnitTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.function.Predicate;
+
+public class FilteringIteratorUnitTest extends BaseTest {
```
magicDGS: Self-reminder: extend `GATKBaseTest` (requires rebase)

src/main/java/org/broadinstitute/hellbender/engine/LocalReadShard.java
```diff
@@ -12,10 +11,7 @@
 
 import java.util.Iterator;
 import java.util.List;
-import java.util.Spliterator;
-import java.util.Spliterators;
 import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
 /**
  * A class to represent a shard of reads data, optionally expanded by a configurable amount of padded data.
```
magicDGS: @droazen - looking into the resurrection of this branch, I realized that the `LocalReadShard` disappeared. For the sliding-window walker I required that each window is independent of the others, so I guess that the `MultiIntervalLocalReadShard` is not the option to go.

Is it ok to keep the `LocalReadShard`? Otherwise, I think that starting a new project for not keeping the GATK blow with unnecessary walkers for your methods is the best way to go...

droazen: @magicDGS I think it would be ok to resurrect `LocalReadShard` if you really need it. But you should know that the reason we got rid of it was because we found that there are major performance issues with querying each interval individually. It's much more efficient to query a large number of intervals at once, particularly if the intervals are close together or overlap, to avoid reading and parsing the same parts of the bam file multiple times. Once we moved to `MultiIntervalLocalReadShard`, the performance of both the `HaplotypeCaller` and `Mutect2` increased by about 30-40%.

I'd recommend closing out these old PRs and submitting a new PR for `SlidingWindowWalker` with the walker class itself, an `ExampleSlidingWindowWalker` and `ExampleSlidingWindowWalker`, and whatever utilities you require (but note that the less common code you refactor, the easier it will be for us to merge your PR quickly).

magicDGS:
Closing in favor of a new implementation

PR #1712, opened 2016-04-15
updating gatk-launch to be compatible with gatk-protected

lbergelson:
This makes gatk-launch work with gatk-protected.  It doesn't work without these changes since the folder names it's looking of are different.  Now it looks up the project name from the settings.gradle and uses that to construct the path to the launch script.


droazen:
Gross! (but ok). Merge when ready.

Issue #1203, opened 2015-11-30
delete SHUFFLE from BQSR spark


akiezun:
SHUFFLE is a loser, let's dump it after the broadcast/sharded decision

droazen:
Remove `VariantShard` and `ReferenceShard` as well when we do this, if possible

akiezun:
what is our story for the cloud though? Google references are not 2bit and we require 2bit for BROADCAST?

droazen:
We should currently be able to broadcast .2bit references stored in GCS buckets in the cloud. Whether it makes sense to do this vs. accessing the reference directly via GCS or the Google reference API is a topic for a future comparison.

droazen:
This is blocked until we can make reference broadcast work with cram inputs.

droazen:
Specifically, it is blocked by https://github.com/broadinstitute/gatk/issues/1443

akiezun:
Unmarking as alpha-2. I doubt this will rise enough priority. This is a cleanup task really because the default is BROADCAST already. @cmnbroad @droazen feel free to mark back as alpha-2 if #1443 is closed in time.

droazen:
SHUFFLE is useful to keep around for profiling purposes -- closing.

lbergelson:
Shouldn't we delete it anyway?  I'm not sure it's useful other than to say.  "Doing this badly is slow".

droazen:
It useful to have a baseline during performance benchmarking to answer the question "are we better than a naive shuffle".

PR #1714, opened 2016-04-15
HaplotypeCaller: the engine-level pieces

droazen:
Included in this PR are the engine-level dependencies of the HaplotypeCaller
from the dr_runnable_haplotypecaller branch. This includes all parts
of that branch that lie outside of the haplotypecaller package itself, and
mostly consists of a new traversal (AssemblyRegionWalker), additions to
the genotyper package, and various additions/tweaks to common utilities
and engine-level classes.

The haplotypecaller package itself has been removed in preparation for its
migration to the gatk-protected repository.


droazen:
For @akiezun to review. Frustratingly, github is refusing to display the full diff, which will make a proper review impossible -- I'm going to separate out the deletion of the haplotypecaller package into a separate PR to try to clean this up.

droazen:
I've separated out (and merged) the deletion of the HC package, and now after a rebase the diff here is much more readable (at least, github will actually display it all!). Review away...

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -0,0 +1,182 @@
...
+     * Shard an interval into ReadShards using the shardSize and shardPadding arguments
+     *
+     * @param interval interval to shard; must be on the contig according to the provided dictionary
+     * @param shardSize desired shard size; intervals larger than this will be divided into shards of up to this size
+     * @param shardPadding desired shard padding; each shard's interval will be padded on both sides by this number of bases
+     * @param readsSource data source for reads
+     * @param dictionary sequence dictionary for reads
+     * @return List of {@link ReadShard} objects, sharded and padded as necessary
+     */
+    public static List<ReadShard> shardInterval( final SimpleInterval interval, final int shardSize, final int shardPadding, final ReadsDataSource readsSource, final SAMSequenceDictionary dictionary ) {
```
magicDGS: Same comment as in #1567: could you change the method to include a parameter that is `shardStep`, which should be lower or equal to `shardSize`?

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * The reads are lazily loaded by default (when accessing the reads via {@link #iterator}. Loading all the
+ * reads in the window at once is possible via {@link #loadAllReads}.
+ *
+ * The reads returned will overlap the expanded padded interval. It's possible to query whether they are within
+ * the main part of the shard via {@link #isContainedWithinShard} and {@link #startsWithinShard}.
+ *
+ * The reads in the window can be filtered via {@link #setReadFilter} (no filtering is performed by default).
+ */
+public final class ReadShard implements Iterable<GATKRead>, Locatable {
```
magicDGS: For me it is confusing that `VariantShard`/`ReferenceShard` represent a differen thing than `ReadShard`, having a similar name. Could it be possible to come back to the `ReadWindow` name? It will be important if other sources are lazily load like this, avoiding clashing with the other names.

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: In the case of `AssemblyRegionWalker` it's easier/conceptually cleaner to think in terms of "padding" rather than a "step", because (among other reasons) we need to ensure that the amount of padding around the assembly regions does not exceed the amount of padding around the read shards.

This is one reason I decided to make `AssemblyRegionWalker` standalone rather than inheriting from a more abstract traversal like `ReadWindowWalker` or `SlidingWindowWalker` -- we can label the arguments in a way that is clearest in the context of `AssemblyRegion` traversal, without having to worry about being super-generic to accomodate other, related traversal types, and we also have the option of hiding the sharding as an implementation detail in the future. The sharding is mostly there right now just as a proof-of-concept for the upcoming Spark version of the HaplotypeCaller, anyway (to prove that we can shard without getting calling artifacts). See my latest comment in https://github.com/broadinstitute/gatk/pull/1708 for more on this issue.

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: You raise a good point about `VariantShard`/`ReferenceShard` -- these are only used in an obsolete Spark implementation of BQSR that is slated for deletion, however (https://github.com/broadinstitute/gatk/issues/1203), so we will likely delete those two classes as part of that ticket.

I went with "shard" because "window" to me implies an interval rather than a packet of records, but I'm open to suggestions. Do you still dislike the name `ReadShard` even after learning that the existing `VariantShard`/`ReferenceShard` will likely be deleted?

(Naming is hard! :)  )

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: I'll add: if you need to shard based on a `shardStep` for `SlidingWindowWalker`, just add a separate overload of `shardInterval()` that does what you want. I'd rather have two simple, clean methods than one method that makes everyone happy but is more complex and harder to systematically test.

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
magicDGS: Sorry, I didn't know about the removal. I think that `ReadShard` is a good name, the only problem is that I was confused with the other classes. In addition, if at the end the `ReadShard` is going to be hidden from the API-side, it won't be important.

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
magicDGS: Even if the `SlidingWindowWalker` is not going to be used for the `AssemblyRegionWalker`, I think that the `ReadShard` could be useful in terms of performance for it. In that case, the method will be needed. Padding could be performed in sliding-windows too.

My idea wasn't remove this method, it was mos likely to do the following with it:

```java
public static List<ReadShard> shardInterval( final SimpleInterval interval, final int shardSize, final int shardPadding, final ReadsDataSource readsSource, final SAMSequenceDictionary dictionary ) {
    return shardInterval(interval, shardSize, shardSize, shardPadding, readsSource, dictionary);
}

public static List<ReadShard> shardInterval( final SimpleInterval interval, final int shardSize, final int shardStep, final int shardPadding, final ReadsDataSource readsSource, final SAMSequenceDictionary dictionary ) {
    // unchanged code until line 1777
    start += shardStep;
    // unchanged code untill the end
}
````

This will provide the same behaviour for the rest of the commit, but allows other way of generate the shard intervals. Of course, it could be done in other part of the code. But of course, I won't use `ReadShard` in `SlidingWindowWalker` if it is goin to be hidden, because it will be one of the parameters of the `apply` function.

Just let me know what is the plan for `ReadShard` for implement #1708 in a different way. And many thanks for the help and fruitful discussion about this topic.

Comment (droazen) [inline src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java]: The read sharding may eventually be hidden from the user for `AssemblyRegionWalker` (ie., not exposed through command-line arguments) to avoid possible confusion between shards and assembly regions. Other traversals, like `SlidingWindowWalker`, are of course free to do whatever they want -- if you want to use `ReadShard` for `SlidingWindowWalker` and pass it to your `apply()` method, go right ahead!

I see now what you meant with the `shardInterval()` method -- I'll make the change you need (and add some extra unit tests for the new functionality) when I address the first round of code review comments in this branch, so that you can hopefully use it in `SlidingWindowWalker`.

File: src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java [hunk 3]
```diff
@@ -0,0 +1,609 @@
...
+     * @param readsToAdd a collection of readsToAdd to add to this active region
+     */
+    public void addAll(final Collection<GATKRead> readsToAdd){
+        Utils.nonNull(readsToAdd).forEach(r -> add(r));
+    }
+
+    /**
+     * Get the extension applied to this region
+     *
+     * The extension is >= 0 bp in size, and indicates how much padding this art walker wanted for its regions
```
akiezun: can you add this useful comment to the `extension` field itself too?

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
```diff
@@ -0,0 +1,609 @@
...
+     *
+     * @return the size in bp of the region extension
+     */
+    public int getExtension() { return extension; }
+
+    public SimpleInterval getReadSpanLoc() {
```
akiezun: doc for `getReadSpanLoc`

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
```diff
@@ -0,0 +1,609 @@
...
+     * provided when this region was created.
+     * The returned list is unmodifiable.
+     */
+    public List<ActivityProfileState> getSupportingStates() {
+        return supportingStates;
+    }
+
+    /**
+     * See #getActiveRegionReference but using the span including regions not the extended span
+     */
```
akiezun: add the note about live array in all those methods that return `byte[]`.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
```diff
@@ -0,0 +1,609 @@
...
+    public byte[] getAssemblyRegionReference( final IndexedFastaSequenceFile referenceReader, final int padding ) {
+        return getReference(referenceReader, padding, extendedLoc);
+    }
+
+    /**
+     * Is this region equal to other, excluding any reads in either region in the comparison
+     * @param other the other active region we want to test
+     * @return true if this region is equal, excluding any reads and derived values, to other
+     */
+    public boolean equalsIgnoreReads(final AssemblyRegion other) {
```
akiezun: null check for the argument

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
```diff
@@ -0,0 +1,609 @@
...
+        // TODO: refactor this method so that we don't need to load all reads from the shard into memory at once!
+        final List<GATKRead> windowReads = new ArrayList<>();
+        for ( final GATKRead read : shard ) {
+            windowReads.add(read);
+        }
+
+        final LocusIteratorByState locusIterator = new LocusIteratorByState(windowReads.iterator(), DownsamplingMethod.NONE, false, false, ReadUtils.getSamplesFromHeader(readsHeader), readsHeader);
+        final ActivityProfile activityProfile = new BandPassActivityProfile(null, maxProbPropagationDistance, activeProbThreshold, BandPassActivityProfile.MAX_FILTER_SIZE, BandPassActivityProfile.DEFAULT_SIGMA, readsHeader);
+
+        // First, use our activity profile to determine the bounds of each assembly region:
```
akiezun: extract a new method from this part of the createFromReadShard method (how about `determineRegionBounds`?)

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
```diff
@@ -0,0 +1,609 @@
...
+                final ReferenceBases refBase = new ReferenceBases(new byte[]{referenceContext.getBases()[pileup.getLocation().getStart() - referenceContext.getWindow().getStart()]}, pileupInterval);
+                final ReferenceContext pileupRefContext = new ReferenceContext(new ReferenceMemorySource(refBase, readsHeader.getSequenceDictionary()), pileupInterval);
+
+                final ActivityProfileState profile = evaluator.isActive(pileup, pileupRefContext, features);
+                activityProfile.add(profile);
+            }
+        }
+        assemblyRegions.addAll(activityProfile.popReadyAssemblyRegions(assemblyRegionPadding, minRegionSize, maxRegionSize, true));
+
+        // Then, fill the assembly regions with overlapping reads from the shard:
```
akiezun: Same here - extract to a new method (`fillRegionsWithOverlappingReads`?)

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -0,0 +1,250 @@
...
+     * Initialize data sources for traversal.
+     *
+     * Marked final so that tool authors don't override it. Tool authors should override onTraversalStart() instead.
+     */
+    @Override
+    protected final void onStartup() {
+        super.onStartup();
+
+        if ( readShardSize <= 0 ) {
+            throw new UserException.BadArgumentValue("--readShardSize", Integer.toString(readShardSize), "read shard size must be > 0");
```
akiezun: hard coded constants like `"--readShardSize"` should be extracted to avoid typos (extracting to this class is sufficient given that they are specific to this walker type)

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -0,0 +1,250 @@
...
+     * The default implementation uses the {@link org.broadinstitute.hellbender.engine.filters.WellformedReadFilter} filter with all default options,
+     * as well as the {@link org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary#MAPPED} filter.
+     *
+     * Default implementation of {@link #traverse()} calls this method once before iterating
+     * over the reads and reuses the filter object to avoid object allocation. Nevertheless, keeping state in filter objects is strongly discouraged.
+     *
+     * Subclasses can override to provide their own filters
+     * Multiple filters can be composed by using {@link org.broadinstitute.hellbender.engine.filters.ReadFilter} composition methods.
+     */
+    public CountingReadFilter makeReadFilter(){
```
akiezun: does this need to be a `CountingReadFilter` or is `ReadFilter` enough from the API point of view?

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -0,0 +1,250 @@
...
+
+    /**
+     * @return The boundaries of the read shard we're currently operating within (ignoring any padding).
+     */
+    public SimpleInterval getCurrentReadShardBounds() {
+        return currentReadShard.getInterval();
+    }
+
+    @Override
+    public void traverse() {
```
akiezun: do you want to keep `traverse` non-final? If yes, document what the subclasses should do when they override it. At least say if they are required to call super or not.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * @return The evaluator to be used to determine whether each locus is active or not. Must be implemented by tool authors.
+     */
+    public abstract AssemblyRegionEvaluator assemblyRegionEvaluator();
+
+    /**
+     * Process an individual AssemblyRegion. Must be implemented by tool authors. Each region will be marked as
+     * either "active" or "inactive" using the results of the configured {@link #assemblyRegionEvaluator}.
```
akiezun: what do you mean by `will be marked` here? Will it come pre-marked, or will this method mark it? or something else?

akiezun: Explicitly state that this method will be called on all regions, not only the active ones and it's the subclass's responsibility to deal with that.

src/main/java/org/broadinstitute/hellbender/engine/ProgressMeter.java
```diff
@@ -31,10 +31,10 @@
     public static final double DEFAULT_SECONDS_BETWEEN_UPDATES = 10.0;
 
     /**
-     * We check the current time every time we process this many records
-     * (to cut down on the number of system calls)
+     * We check the current time every time we process this many records,
+     * by default (to cut down on the number of system calls)
      */
-    public static final long RECORDS_BETWEEN_TIME_CHECKS = 1000l;
+    public static final long DEFAULT_RECORDS_BETWEEN_TIME_CHECKS = 1000l;
```
akiezun: can you change to `1000L` ? the lowercase `l` always tricks me

src/main/java/org/broadinstitute/hellbender/engine/ProgressMeter.java
```diff
@@ -69,7 +75,7 @@
     /**
      * Current timestamp in milliseconds as returned by our {@link #timeFunction}.
      *
-     * Updated only every {@link #RECORDS_BETWEEN_TIME_CHECKS} records to cut down
+     * Updated only every {@link #recordsBetweenTimeChecks} records to cut down
      * on system calls.
      */
     private long currentTimeMs = 0l;
```
akiezun: `0l` -> `0L`

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -0,0 +1,182 @@
...
+    public List<GATKRead> loadAllReads() {
+        return StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterator(), Spliterator.ORDERED), false).collect(Collectors.toList());
+    }
+
+    /**
+     * @param loc Locatable to test
+     * @return true if loc is contained within this shard's interval, otherwise false
+     */
+    public boolean isContainedWithinShard( final Locatable loc ) {
+        return interval.contains(loc);
```
akiezun: null check for `loc`

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * @param loc Locatable to test
+     * @return true if loc starts within this shard's interval, otherwise false
+     */
+    public boolean startsWithinShard( final Locatable loc ) {
+        return interval.contains(new SimpleInterval(loc.getContig(), loc.getStart(), loc.getStart()));
```
akiezun: null check for `loc`

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * @return number of bases of padding to the right of our interval
+     */
+    public int numRightPaddingBases() {
+        return paddedInterval.getEnd() - interval.getEnd();
+    }
+
+    /**
+     * Shard an interval into ReadShards using the shardSize and shardPadding arguments
```
akiezun: i'm not a literary man but `Shard an interval into ReadShards using the shardSize and shardPadding arguments` sounds a bit too shard-full for me.

src/test/java/org/broadinstitute/hellbender/engine/ReadShardUnitTest.java
```diff
@@ -0,0 +1,110 @@
...
+        };
+    }
+
+    @Test(dataProvider = "ShardIntervalTestData")
+    public void testShardInterval( final SimpleInterval originalInterval, final int windowSize, final int windowPadding, final ReadsDataSource readsSource, final SAMSequenceDictionary dictionary, List<ReadShard> expectedWindows  ) {
+        final List<ReadShard> windows = ReadShard.shardInterval(originalInterval, windowSize, windowPadding, readsSource, dictionary);
+
+        Assert.assertEquals(windows.size(), expectedWindows.size(), "Wrong number of windows created by ReadWindowWalker.shardInterval()");
+        for ( int i = 0; i < windows.size(); ++i ) {
+            final ReadShard window = windows.get(i);
```
akiezun: can you add a check that each read is actually inside the right boundaries?

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -0,0 +1,182 @@
...
+        this.paddedInterval = paddedInterval;
+        this.readsSource = readsSource;
+    }
+
+    /**
+     * Reads in this shard will be filtered using this filter before being returned
+     *
+     * @param filter filter to use (may be null)
+     */
+    public void setReadFilter( final ReadFilter filter ) {
```
akiezun: why is this a setter rather than a constructor parameter? do you want to adjust filtering after the shard is created?

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
akiezun: this is crying out for a `contains(String, int)` method on `SimpleInterval` or `Locatable`. But it's not a big deal until it shows up on the profiler

akiezun:
Looks fine - @droazen can you indicate which classes are new to this PR (ReadShard etc) and which have been reviewed previously (genotyping classes etc)?

akiezun:
back to @droazen with a question - happy to review more but i need to know where to focus the attention.

droazen:
@akiezun Classes most in need of review/extra scrutiny:

`AssemblyRegion` (especially `createFromReadShard()`)
`AssemblyRegionEvaluator`
`AssemblyRegionWalker`
`ReadShard`
`ReadFilterLibrary` (small change, but needs review)
`DepthPerSampleHC`
`StrandBiasBySample`
`ActivityProfile`
`ActivityProfileState`
`BandPassActivityProfile`
`PairHMM`
`AssemblyRegionUnitTest` (especially `testCreateFromReadShard()`)
`ReadShardUnitTest`
`ActivityProfileUnitTest`
`BandPassActivityProfileUnitTest`

You should do a review pass over the rest as well, but focus on the above.

droazen:
Back to @akiezun

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
```diff
@@ -0,0 +1,609 @@
...
+        final PeekableIterator<GATKRead> reads = new PeekableIterator<>(windowReads.iterator());
+        AssemblyRegion lastRegion = null;
+        for ( final AssemblyRegion region : assemblyRegions ) {
+            if ( lastRegion != null ) {
+                for ( final GATKRead lastRegionRead : lastRegion.getReads() ) {
```
akiezun: this suggests a need for a `AssemblyRegion.getReadsOverlapping(Locatable)` on `AssemblyRegion`

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
```diff
@@ -0,0 +1,609 @@
...
+
+                final ActivityProfileState profile = evaluator.isActive(pileup, pileupRefContext, features);
+                activityProfile.add(profile);
+            }
+        }
+        assemblyRegions.addAll(activityProfile.popReadyAssemblyRegions(assemblyRegionPadding, minRegionSize, maxRegionSize, true));
+
+        // Then, fill the assembly regions with overlapping reads from the shard:
+        final PeekableIterator<GATKRead> reads = new PeekableIterator<>(windowReads.iterator());
+        AssemblyRegion lastRegion = null;
```
akiezun: you mean `previousRegion`

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
```diff
@@ -0,0 +1,609 @@
...
+                }
+            }
+
+            while ( reads.hasNext() ) {
+                final GATKRead read = reads.peek();
+                if ( region.getExtendedSpan().overlaps(read) ) {
+                    region.add(reads.next());
+                }
+                else {
+                    break;
```
akiezun: why is there a break here - is it because there's a sorting assumption? If yes, make this assumption explicit

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionEvaluator.java
```diff
@@ -0,0 +1,21 @@
...
+
+    /**
+     * Given a pileup, returns an ActivityProfileState containing the probability (0.0 to 1.0) that it's an "active" site.
+     *
+     * @param locusPileup reads pileup to examine
+     * @param referenceContext reference base overlapping the pileup locus
+     * @param featureContext features overlapping the pileup locus
+     * @return probability between 0.0 and 1.0 that the site is active
+     */
+    ActivityProfileState isActive( final AlignmentContext locusPileup, final ReferenceContext referenceContext, final FeatureContext featureContext );
```
akiezun: given that we have a class `ReadPileup` the name of `locusPileup` is confusing a bit

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionEvaluator.java
```diff
@@ -0,0 +1,21 @@
...
+@FunctionalInterface
+public interface AssemblyRegionEvaluator {
+
+    /**
+     * Given a pileup, returns an ActivityProfileState containing the probability (0.0 to 1.0) that it's an "active" site.
+     *
+     * @param locusPileup reads pileup to examine
+     * @param referenceContext reference base overlapping the pileup locus
+     * @param featureContext features overlapping the pileup locus
+     * @return probability between 0.0 and 1.0 that the site is active
```
akiezun: this comment suggests that the return type is `double` but it's not

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -0,0 +1,182 @@
...
+     */
+    public List<GATKRead> loadAllReads() {
+        return StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterator(), Spliterator.ORDERED), false).collect(Collectors.toList());
+    }
+
+    /**
+     * @param loc Locatable to test
+     * @return true if loc is contained within this shard's interval, otherwise false
+     */
+    public boolean isContainedWithinShard( final Locatable loc ) {
```
akiezun: rename to `contains` ? Imagine code `shard.isContainedWithinShard(loc)` and compare to `shard.contains(loc)`. The former does not even make sense grammatically.

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
akiezun: I'd rename to something like `createFromInterval` or something. Using `shard` as both a noun and a verb is confusing.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -27,9 +27,9 @@ private ReadFilterLibrary(){ /*no instance*/ }
     public static final ReadFilter NON_ZERO_REFERENCE_LENGTH_ALIGNMENT = read -> read.getCigarElements().stream().anyMatch(c -> c.getOperator().consumesReferenceBases() && c.getLength() > 0);
 
     /**
-     * Reads whose mate maps to the same contig.
+     * Reads that either have a mate that maps to the same contig, or don't have a mapped mate.
      */
-    public static final ReadFilter MATE_ON_SAME_CONTIG = read -> read.isPaired() && ! read.isUnmapped() && ! read.mateIsUnmapped() && read.getContig().equals(read.getMateContig());
+    public static final ReadFilter MATE_ON_SAME_CONTIG = read -> ! read.isPaired() || read.isUnmapped() || read.mateIsUnmapped() || read.getContig().equals(read.getMateContig());
```
akiezun: semantics is now very different yet the name stayed the same. How about renaming to `UNMAPPED_OR_MATE_ON_SAME_CONTIG`?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandBiasBySample.java
```diff
@@ -0,0 +1,80 @@
...
+ *     <li>This annotation can only be generated by HaplotypeCaller (it will not work when called from VariantAnnotator).</li>
+ * </ul>
+ *
+ * <h3>Related annotations</h3>
+ * <ul>
+ *     <li><b><a href="https://www.broadinstitute.org/gatk/guide/tooldocs/org_broadinstitute_gatk_tools_walkers_annotator_FisherStrand.php">FisherStrand</a></b> uses Fisher's Exact Test to evaluate strand bias.</li>
+ *     <li><b><a href="https://www.broadinstitute.org/gatk/guide/tooldocs/org_broadinstitute_gatk_tools_walkers_annotator_StrandOddsRatio.php">StrandOddsRatio</a></b> is an updated form of FisherStrand that uses a symmetric odds ratio calculation.</li>
+ * </ul>
+ */
+public final class StrandBiasBySample extends GenotypeAnnotation {
```
akiezun: in what sense is this `by sample`?

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
```diff
@@ -0,0 +1,493 @@
...
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.Utils;
+
+import java.util.*;
+
+/**
+ * Class holding information about per-base activity scores for
+ * assembly region traversal
+ */
+public class ActivityProfile {
```
akiezun: `final class`

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
```diff
@@ -0,0 +1,493 @@
...
+    protected final List<ActivityProfileState> stateList;
```
akiezun: make these fields private if possible

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
```diff
@@ -0,0 +1,493 @@
...
+    /**
+     * How many profile results are in this profile?
+     * @return the number of profile results
+     */
+    public int size() {
+        return stateList.size();
+    }
+
+    /**
+     * Is this profile empty?
```
akiezun: explain what an empty profile means

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
```diff
@@ -0,0 +1,493 @@
...
+
+    public int getEnd() {
+        return regionStopLoc.getEnd();
+    }
+
+    /**
+     * Get the list of activity profile results in this object
+     * @return a non-null, ordered list of activity profile results
+     */
+    protected List<ActivityProfileState> getStateList() {
```
akiezun: why is this method protected?

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
```diff
@@ -0,0 +1,493 @@
...
+        return stateList;
+    }
+
+    /**
+     * Get the probabilities of the states as a single linear array of doubles
+     * @return a non-null array
+     */
+    protected double[] getProbabilitiesAsArray() {
```
akiezun: same here -make public or private or packega-private if needed for testing

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
```diff
@@ -0,0 +1,493 @@
...
+     * @param justAddedState the state our client provided to use to add to the list
+     * @return a list of derived states that should actually be added to this profile's state list
+     */
+    protected Collection<ActivityProfileState> processState(final ActivityProfileState justAddedState) {
+        if ( justAddedState.getResultState().equals(ActivityProfileState.Type.HIGH_QUALITY_SOFT_CLIPS) ) {
+            // special code to deal with the problem that high quality soft clipped bases aren't added to pileups
+            final List<ActivityProfileState> states = new LinkedList<ActivityProfileState>();
+            // add no more than the max prob propagation distance num HQ clips
+            final int numHQClips = Math.min(justAddedState.getResultValue().intValue(), getMaxProbPropagationDistance());
+            for( int jjj = - numHQClips; jjj <= numHQClips; jjj++ ) {
```
akiezun: `jjj`- > `i`

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
```diff
@@ -0,0 +1,493 @@
...
+     *                        stateList.  Used to close out the active region when we've hit some kind of end (such
+     *                        as the end of the contig)
+     * @return a non-null list of active regions
+     */
+    public List<AssemblyRegion> popReadyAssemblyRegions( final int assemblyRegionExtension, final int minRegionSize, final int maxRegionSize, final boolean forceConversion ) {
+        if ( assemblyRegionExtension < 0 ) throw new IllegalArgumentException("assemblyRegionExtension must be >= 0 but got " + assemblyRegionExtension);
+        if ( minRegionSize < 1 ) throw new IllegalArgumentException("minRegionSize must be >= 1 but got " + minRegionSize);
+        if ( maxRegionSize < 1 ) throw new IllegalArgumentException("maxRegionSize must be >= 1 but got " + maxRegionSize);
+
+        final LinkedList<AssemblyRegion> regions = new LinkedList<>();
```
akiezun: why linked list?

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfileState.java
```diff
@@ -0,0 +1,99 @@
...
+    private final SimpleInterval loc;
+    private double activeProb;
+    private final Type resultState;
+    private final Number resultValue;
+
+    public double isActiveProb() {
+        return activeProb;
+    }
+
+    public void setIsActiveProb( final double activeProb ) {
```
akiezun: validate argument between 0 and 1

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfileState.java
```diff
@@ -0,0 +1,99 @@
...
+        this.activeProb = activeProb;
+    }
+
+    public Type getResultState() {
+        return resultState;
+    }
+
+    public Number getResultValue() {
```
akiezun: document this method

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfileState.java
```diff
@@ -0,0 +1,99 @@
...
+        return resultValue;
+    }
+
+    public enum Type {
```
akiezun: document this enum - what is this for?

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/BandPassActivityProfile.java
```diff
@@ -0,0 +1,160 @@
...
+ * them out of an interval
+ */
+public class BandPassActivityProfile extends ActivityProfile {
+    public static final int MAX_FILTER_SIZE = 50;
+    private final static double MIN_PROB_TO_KEEP_IN_FILTER = 1e-5;
+    public static final double DEFAULT_SIGMA = 17.0;
+
+    private final int filterSize;
+    private final double sigma;
+    private final double[] GaussianKernel;
```
akiezun: lowercase for field names

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/BandPassActivityProfile.java
```diff
@@ -0,0 +1,160 @@
...
+     * Band pass the probabilities in the ActivityProfile, producing a new profile that's band pass filtered
+     * @return a new double[] that's the band-pass filtered version of this profile
+     */
+    @Override
+    protected Collection<ActivityProfileState> processState(final ActivityProfileState justAddedState) {
+        final Collection<ActivityProfileState> states = new LinkedList<>();
+
+        for ( final ActivityProfileState superState : super.processState(justAddedState) ) {
+            if ( superState.isActiveProb() > 0.0 ) {
+                for( int jjj = -filterSize; jjj <= filterSize; jjj++ ) {
```
akiezun: `jjj`-> `i`

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
```diff
@@ -31,11 +32,43 @@
...
         /* Optimized version of the PairHMM which caches per-read computations and operations in real space to avoid costly sums of log10'ed likelihoods */
-        LOGLESS_CACHING(() -> new LoglessPairHMM());
+        LOGLESS_CACHING(() -> {
+            final LoglessPairHMM hmm = new LoglessPairHMM();
+            logger.info("Using the non-hardware-accelerated Java LOGLESS_CACHING PairHMM implementation");
+            return hmm;
+        }),
+        /* Optimized AVX implementation of LOGLESS_CACHING called through JNI. Throws if AVX is not available */
+        AVX_LOGLESS_CACHING(() -> {
+            // Constructor will throw a UserException if AVX is not available
```
akiezun: hmm, can we add a static method `VectorLoglessPairHMM.isAvailable` to avoid exceptions?

src/test/java/org/broadinstitute/hellbender/engine/AssemblyRegionUnitTest.java
```diff
@@ -0,0 +1,485 @@
...
+    @Test(dataProvider = "TrimAssemblyRegionData")
+    public void testTrimAssemblyRegion(final SimpleInterval regionLoc, final int extension, final SimpleInterval desiredSpan, final SimpleInterval expectedAssemblyRegion, final int expectedExtension) {
+        final AssemblyRegion region = new AssemblyRegion(regionLoc, Collections.<ActivityProfileState>emptyList(), true, extension, header);
+        final AssemblyRegion trimmed = region.trim(desiredSpan);
+        Assert.assertEquals(trimmed.getSpan(), expectedAssemblyRegion, "Incorrect region");
+        Assert.assertEquals(trimmed.getExtension(), expectedExtension, "Incorrect region");
+    }
+
+    @Test
+    public void testCreateFromReadShard() {
```
akiezun: looks good

akiezun:
Looks fine. Small comments, back to @droazen

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
droazen: I think this "live array" comment was incorrect -- the `referenceReader.getSubsequenceAt()` call actually allocates a new byte array. Removing this warning.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
droazen: Returning false for null

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
droazen: Method extracted

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
droazen: Method extracted

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
droazen: Declaring constants for all full/short names for all arguments made the argument list hard to read/understand (an effect that would be much worse for the `HaplotypeCaller` arguments themselves). I experimented with a reflection-based solution, but it was a bit silly. Decided ultimately to modify the error messages to describe the problem rather than including the exact argument names.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
droazen: It needs to be a `CountingReadFilter`, because `traverse()` calls `getSummaryLine()` on it.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
droazen: Made `traverse()` final

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
droazen: The regions are pre-marked -- re-wrote the comments here to be clearer and more precise.

droazen: Also re-wrote the class-level docs to improve clarity.

src/main/java/org/broadinstitute/hellbender/engine/ProgressMeter.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/engine/ProgressMeter.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: Done

droazen: (and there will be a `contains()` on `Locatable` soon!)

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: Oh dear -- will adjust the wording here to be less awful.

droazen: Renamed to `divideIntervalIntoShards()`

src/test/java/org/broadinstitute/hellbender/engine/ReadShardUnitTest.java
droazen: Done -- implemented as a separate test. Also improved test coverage for `ReadShard` in general.

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: Yes, the division of the user's intervals into shards is decoupled from the traversal, where things like read filtering get set. Also, one could imagine a hypothetical scenario of wanting to traverse the same shard more than once with different filter settings.

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: Method renamed, requested overload added with additional unit tests.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
droazen: I think this is a rare operation not deserving of a new method + tests-- we only do it here because reads can span multiple regions, so we need to check the reads in the previous region for overlap with the current region.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
droazen: Yes, renamed

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
droazen: There is indeed a sorting assumption, which is now documented.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionEvaluator.java
droazen: It has the advantage of making it clear that the pileup spans only a single locus.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionEvaluator.java
droazen: Fixed

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: Good point -- renamed.

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
droazen: Nope, can't be final -- there is a subclass (`BandPassActivityProfile`)

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
droazen: Must be protected due to subclassing.

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
droazen: inheritance

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
droazen: Needed in subclass, must stay protected

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfile.java
droazen: No good reason -- changed to `ArrayList`

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfileState.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/BandPassActivityProfile.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/BandPassActivityProfile.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
droazen: We *want* an exception if the user requested AVX and it's not available. Only `FASTEST_AVAILABLE` should use a fallback.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandBiasBySample.java
droazen: Take a look at the code -- it creates a map of `sample` -> `alleleLikelihoodMap` before calling into `getContingencyTable()`

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
droazen: This change was to make our `MATE_ON_SAME_CONTIG` filter consistent with GATK3's `BadMateFilter` (albeit with the logic inverted), since the `HaplotypeCaller` needs this filter. Will try to come up with a more accurate name.

droazen: Went with `MATE_ON_SAME_CONTIG_OR_NO_MAPPED_MATE` :(

droazen: Also added exhaustive unit tests

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfileState.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/activityprofile/ActivityProfileState.java
droazen: Done. This code is pretty bad, and will likely be refactored in the future.

droazen:
All comments addressed, back to @akiezun!

akiezun:
LGTM, merge when ready

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
```diff
@@ -0,0 +1,661 @@
...
+
+    private void checkStates(final SimpleInterval activeRegionLoc) {
+        if ( ! this.supportingStates.isEmpty() ) {
+            if ( this.supportingStates.size() != activeRegionLoc.size() ) {
+                throw new IllegalArgumentException("Supporting states wasn't empty but it doesn't have exactly one state per bp in the active region: states " + this.supportingStates.size() + " vs. bp in region = " + activeRegionLoc.size());
+            }
+            SimpleInterval lastStateLoc = null;
+            for ( final ActivityProfileState state : this.supportingStates ) {
+                if ( lastStateLoc != null ) {
+                    if ( state.getLoc().getStart() != lastStateLoc.getStart() + 1 || !state.getLoc().getContig().equals(lastStateLoc.getContig())) {
```
sputnikci: [PMD] INFO: These nested if statements could be combined

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
```diff
@@ -53,4 +123,563 @@ public static void computeAlleleFrequencyPriors(final int N, final double[] prio
...
+        // Add 0.0 removes -0.0 occurrences.
+        final double phredScaledConfidence = (-10.0 * log10Confidence) + 0.0;
+
+        // return a null call if we don't pass the confidence cutoff or the most likely allele frequency is zero
+        if ( !passesEmitThreshold(phredScaledConfidence, outputAlternativeAlleles.siteIsMonomorphic) && !forceSiteEmission()) {
+            // technically, at this point our confidence in a reference call isn't accurately estimated
+            //  because it didn't take into account samples with no data, so let's get a better estimate
+            final double[] AFpriors = getAlleleFrequencyPriors(vc, defaultPloidy, model);
+            final int INDEX_FOR_AC_EQUALS_1 = 1;
+            return limitedContext ? null : estimateReferenceConfidence(vc, stratifiedContexts, AFpriors[INDEX_FOR_AC_EQUALS_1], true, PoFGT0);
```
sputnikci: [PMD] INFO: Ternary operators that can be simplified with || or &&

droazen:
Oh sputnik, is that the best you can do?

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
sputnikci: [PMD] INFO: These nested if statements could be combined

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
sputnikci: [PMD] INFO: Ternary operators that can be simplified with || or &&

PR #1717, opened 2016-04-17
update GATKSparkTool.getRecommendedNumReducers for GCS

lbergelson:
running SparkTools on GCS was failing beceause BucketUtils.dirSize didn't support GCS paths
updating it to support GCS paths and updating GATKSparkTool to pass in the necessary Auth


akiezun:
@lbergelson what;s the command that faild without this fix. I'm running spark on GCS and don't see the problem

akiezun:
assigning to @lbergelson for a clarification

lbergelson:
Incorporated into #1785

PR #1719, opened 2016-04-18
Update build.gradle to make sure that tests are always rerunable.

akiezun:
Before this change you can run tests only once before gradle decides that the task is "UP-TO-DATE". I want to be able to rerun tests multiple times (bonus: guess what I'm testing) and so this caching got in the way. With this change, tasks are always rerunnable.

Test it by running this twice:
```
./gradlew test --tests *VariantFiltrationIntegrationTest*
```

@lbergelson can you review?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,13 +65,13 @@ script:
     echo "Can't run cloud tests without keys so don't run tests";
   else
     bash scripts/install_git_lfs.sh;
-    travis_wait 30 ./gradlew test;
+    travis_wait 30 ./gradlew jacocoTestReport;
   fi
 after_success:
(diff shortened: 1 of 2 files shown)


lbergelson:
This is causing the tests to run twice in travis. There's an `after_success` block that runs ./gradlew jacocoTestReport coveralls`  This is fine in master because it doesn't rerun the tests since they're marked as complete.  With your change it runs them again though.  You should be able to move the jacocoTest report up into the `script` block to prevent this.  Coveralls may have to move too.  Fine to merge once that's fixed.

akiezun:
@lbergelson I think this does it. Can you check?

lbergelson:
👍

Issue #1713, opened 2016-04-15
Implement a test that catches failures to register new annotations in GATKVCFHeaderLines


droazen:

droazen:
If an annotation's header line isn't registered in `GATKVCFHeaderLines`, it can cause a NPE deep in htsjdk when we go to write the header.

akiezun:
do you have a stacktrace?

droazen:
Eg., for `ExcessHet` we needed a line like:

        addInfoLine(new VCFInfoHeaderLine(EXCESS_HET_KEY, 1, VCFHeaderLineType.Float, "Phred-scaled p-value for exact test of excess heterozygosity"));

droazen:
Example stacktrace:

org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCallerIntegrationTest.testGVCFModeIsConcordantWithGATK3Results FAILED
    java.lang.NullPointerException
        at htsjdk.variant.vcf.VCFHeader.removeVCFVersionLines(VCFHeader.java:268)
        at htsjdk.variant.vcf.VCFHeader.<init>(VCFHeader.java:115)
        at htsjdk.variant.vcf.VCFHeader.<init>(VCFHeader.java:139)
        at htsjdk.variant.vcf.VCFHeader.<init>(VCFHeader.java:135)
        at org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCallerEngine.writeHeader(HaplotypeCallerEngine.java:401)
        at org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCaller.onTraversalStart(HaplotypeCaller.java:204)
        at org.broadinstitute.hellbender.engine.GATKTool.doWork(GATKTool.java:515)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:102)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:155)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:174)
        at org.broadinstitute.hellbender.Main.instanceMain(Main.java:67)
        at org.broadinstitute.hellbender.Main.instanceMain(Main.java:74)
        at org.broadinstitute.hellbender.CommandLineProgramTest.runCommandLine(CommandLineProgramTest.java:65)
[...]

PR #1720, opened 2016-04-18
Adding ExcessHet to GATKVCFHeaderLines and future-proofing

akiezun:
fixes https://github.com/broadinstitute/gatk/issues/1713

the tests fail unless the 1 line in `GATKVCFHeaderLines` is added

@droazen please review

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
@@ -130,6 +130,9 @@ public final class VariantAnnotatorEngine {
             }
         }
 
+        if (descriptions.contains(null)){
+            throw new IllegalStateException("getVCFAnnotationDescriptions should not contain null. This error is likely due to an incorrect implementation of getDescriptions() in one or more of the annotation classes");
+        }
         return descriptions;
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
```diff
@@ -130,6 +130,9 @@ private VariantOverlapAnnotator initializeOverlapAnnotator(final FeatureInput<Va
             }
         }
 
+        if (descriptions.contains(null)){
+            throw new IllegalStateException("getVCFAnnotationDescriptions should not contain null");
```
droazen: Make it clear in this error message that the error is due to an incorrect implementation of `getDescriptions()` in one or more of the annotation classes

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/AnnotationUnitTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class AnnotationUnitTest extends BaseTest {
```
droazen: Move these tests to `VariantAnnotatorEngineUnitTest` instead of creating a separate class?

droazen:
:+1: Merge after addressing comments -- back to @akiezun

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
akiezun: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/AnnotationUnitTest.java
akiezun: done

PR #1724, opened 2016-04-18
add fasta bwt index to repo for bwa

akiezun:
@SHuang-Broad this is one for you


SHuang-Broad:
Thank you very much, Adam!

PR #1729, opened 2016-04-19
fix resource leaks in MarkDuplicates

akiezun:
This was debugged while testing https://github.com/samtools/htsjdk/pull/576. Readers allocated in `AbstractMarkDuplicatesCommandLineProgram.openInputs` were never closed. In the asyncIO realm (if we switch to async reading) this is a big problem because one worker thread is then created and abandoned (it keeps living after its master is long gone - unless the master is closed which will inform the worker to finish).

The diffs are much more trivial than it looks in github gui - essentially 2 things were put in try-with-resources (two lines like this) and lot of white-space shifting followed.
`final SamHeaderAndIterator headerAndIterator = openInputs();`

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/AbstractMarkDuplicatesCommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/AbstractMarkDuplicatesCommandLineProgram.java
@@ -4,6 +4,7 @@ import htsjdk.samtools.*;
 import htsjdk.samtools.DuplicateScoringStrategy.ScoringStrategy;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.util.CloseableIterator;
+import htsjdk.samtools.util.CloserUtil;
 import htsjdk.samtools.util.Histogram;
 import org.broadinstitute.hellbender.cmdline.Argument;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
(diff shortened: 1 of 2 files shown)


akiezun:
@lbergelson can you review?

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/AbstractMarkDuplicatesCommandLineProgram.java
```diff
@@ -183,13 +184,20 @@ String getNonCollidingId(final String recordId) {
...
 
-        public SamHeaderAndIterator(final SAMFileHeader header, final CloseableIterator<SAMRecord> iterator) {
+        public SamHeaderAndIterator(final SAMFileHeader header, final CloseableIterator<SAMRecord> iterator, final List<SamReader> readers) {
             this.header = header;
             this.iterator = iterator;
+            this.readers = readers;
+        }
+
+        @Override
+        public void close(){
```
lbergelson: Should close() also close the iterators?

lbergelson:
@akiezun Looks good to me.  It makes `SamHeaderAndIterator` a bit of a weird class, but I don't see a better solution.

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/AbstractMarkDuplicatesCommandLineProgram.java
akiezun: good catch

akiezun:
hmm, i was testing github's new 'squash and merge' and it pushes squashed commits directly to master. I think we should not use it because of this

lbergelson:
@akiezun What's wrong with that?  It just means we don't have a merge commit.

PR #1734, opened 2016-04-19
skip hashtable lookup in Kryo

akiezun:
hashtable lookups are expensive in Kryo and they add up to 15% or more of runtime (top hotspot on Xprof). https://twitter.com/aphyr/status/478638361150636032

This PR turns off reference tracking in Kryo which speeds things up ~7.2mins vs 7.4mins on MarkDuplicatesSpark but it's a bit risky because I think it may result in an infinite loop for cyclic object graphs. We do not have any cyclic object graphs now and so it's fine.

The PR is to open a convo about this.
@tomwhite @droazen @laserson wdyt?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/GATKRegistrator.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/GATKRegistrator.java
@@ -25,7 +25,10 @@ public class GATKRegistrator implements KryoRegistrator {
     @SuppressWarnings({"unchecked", "rawtypes"})
     @Override
     public void registerClasses(Kryo kryo) {
-
+	//NOTE: this speeds up serialization because it avoids a hashtable lookup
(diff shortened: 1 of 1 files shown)


lbergelson:
I think we should set this for specific pipelines rather than globally.  It makes sense to not waste time doing it, but I don't want someone else who writes a spark tool without knowing about this to get bitten by it.

droazen:
I vote no -- the danger (and need to worry about the danger!) is not worth the relatively small gain. `HaplotypeCaller` might very possibly create cyclical object references, so I don't think this is merely hypothetical. I recommend we stick to safe optimizations.

laserson:
+1.  It seems like a very small gain and not worth the future debugging headache.

akiezun:
nevermind. It crashes our tests. Too small benefit anyway for such a risk

PR #1735, opened 2016-04-19
bump to newer htsjdk snapshot and fix spelling of libIntelDeflater.so

lbergelson:
Also added a check that extractIntelDeflator extracts exactly 1 file.  It will kill the build if it's missing.


akiezun:
👍

PR #1737, opened 2016-04-19
Update README.md to add info about dynamic allocation on YARN

akiezun:
added info about dynamic allocation on YARN (I tested it on Google and it works nicely)

@lbergelson please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ If you are looking for the codebase of the current production version of GATK, p
             --conf spark.yarn.executor.memoryOverhead=600
         ```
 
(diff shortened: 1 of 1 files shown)


lbergelson:
neat

README.md
```diff
@@ -111,6 +111,7 @@ If you are looking for the codebase of the current production version of GATK, p
             --conf spark.yarn.executor.memoryOverhead=600
         ```
 
+    * You can also omit the "--num-executors" to enable [dynamic allocation](https://spark.apache.org/docs/latest/job-scheduling.html#dynamic-resource-allocation) if you configure the cluster properly (see the Spark website for instructins).
```
lbergelson: Can you also omit memory and core arguments then?

akiezun: I think not. @tomwhite do you know if you can meaningfully omit  memory and core args with dynamic allocation?

tomwhite: I don't think so. The number of executors is the the thing that's dynamic, not the size of an executor.

Issue #1739, opened 2016-04-20
IntelDeflater misplaced after build and not used on gatk-launch


akiezun:
after a clean build, build/libIntelDeflater.so is a directory (!) and not a file and running gatk does not use the IntelDeflater (plus prints out a (bogus) message) - `Java HotSpot(TM) 64-Bit Server VM warning: You have loaded library /local/akiezun/gatk4_bqsr_deleteIndels_v2/gatk/build/libIntelDeflater.so which might have disabled stack guard. The VM will try to fix the stack guard now.
It's highly recommended that you fix the library with 'execstack -c <libfile>', or link it with '-z noexecstack'`

@lbergelson please fix (ideally prevent from happening in the future or at least enter a ticket to future proof it)

akiezun:
i guess https://github.com/broadinstitute/gatk/issues/1738 is the ticket

Issue #1738, opened 2016-04-20
add test to make sure intel deflator is loadable


lbergelson:
We need a test in travis that the deflator is loadable.  I recently broke it's extraction and nothing failed.  We need to prevent this from happening in the future.

PR #1742, opened 2016-04-20
fix lib intel deflater extraction

lbergelson:
the first commit adds a test, which should fail
the second should fix the test.  It's currently failing though, so I'm not sure what's happening yet.  It works on gsa5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,8 @@ script:
     echo "Can't run cloud tests without keys so don't run tests";
   else
     bash scripts/install_git_lfs.sh;
+    ./gradlew installDist;
+   ./gatk-launch PrintReads -I src/test/resources/NA12878.chr17_69k_70k.dictFix.bam -O /dev/null ;
     travis_wait 30 ./gradlew jacocoTestReport;
   fi
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,14 +59,16 @@ install:
     echo "Can't run cloud tests without keys so don't bother building";
   else
     ./gradlew assemble;
+    ./gradlew installDist;
   fi
 script:
+#basic sanity test to check that gatk-launch actually runs something
+- ./gatk-launch PrintReads -I src/test/resources/NA12878.chr17_69k_70k.dictFix.bam -O output.bam
+#run tests
 - if [[ $TRAVIS_SECURE_ENV_VARS == false && $CLOUD == mandatory ]]; then
     echo "Can't run cloud tests without keys so don't run tests";
   else
     bash scripts/install_git_lfs.sh;
-    ./gradlew installDist;
-   ./gatk-launch PrintReads -I src/test/resources/NA12878.chr17_69k_70k.dictFix.bam -O /dev/null ;
     travis_wait 30 ./gradlew jacocoTestReport;
   fi
 after_success:
--- a/build.gradle
+++ b/build.gradle
@@ -5,8 +5,6 @@ buildscript {
         mavenCentral()
         jcenter() // for shadow plugin
      }
-
-
 }
 
 plugins {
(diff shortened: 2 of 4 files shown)


lbergelson:
@akiezun This should pass.  The issue was that `test` didn't depend on `extractIntelDeflater` and was correctly failing when it wasn't found.

.travis.yml
```diff
@@ -65,6 +65,8 @@ script:
     echo "Can't run cloud tests without keys so don't run tests";
   else
     bash scripts/install_git_lfs.sh;
+    ./gradlew installDist;
```
akiezun: what are these changes for?

build.gradle
```diff
@@ -266,6 +261,19 @@ test {
             logger.lifecycle("Test: " + descriptor + " produced standard out/err: " + event.message )
         }
     }
+
+    testLogging {
```
akiezun: what are these changes for?

akiezun:
back to @lbergelson with 2 questions

src/test/java/org/broadinstitute/hellbender/IntelDeflaterTest.java
```diff
@@ -0,0 +1,24 @@
...
+import htsjdk.samtools.util.zip.DeflaterFactory;
+import org.apache.commons.lang3.SystemUtils;
+import org.testng.Assert;
+import org.testng.SkipException;
+import org.testng.annotations.Test;
+
+/**
+ * Test that it's possible to load libIntelDeflater
+ */
+public class IntelDeflaterTest {
```
droazen: Should extend `BaseTest` and have either a `UnitTest` or `IntegrationTest` suffix.

.travis.yml
lbergelson: This was a change to help debug what was going on by running `PrintReads` and including it's output in the travis log.  I left it because I figured it was a useful thing to have and it adds very little to the test run time.

build.gradle
lbergelson: This forces the test output to always include notification for every skipped or failed test, as well as a stack trace if necessary.  It previously wasn't clear when running single tests or not if they were skipped or if they passed.  Now it's obvious. 

It also adds a line after the whole test suite is run that includes the total number of tests run and the count of succeeded/skipped/failed because that wasn't always clear.

src/test/java/org/broadinstitute/hellbender/IntelDeflaterTest.java
lbergelson: It's not really a unit test, I guess it's sort of an integration test?  What suffix do you prefer?

src/test/java/org/broadinstitute/hellbender/IntelDeflaterTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Test
+    public void testIntelDeflaterIsAvailable(){
+        if ( !SystemUtils.IS_OS_LINUX ) {
+            throw new SkipException("IntelDeflater not available on this platform");
```
lbergelson: @droazen I think this check may fail on POWER systems, but I'm not sure what the test condition should be.

src/test/java/org/broadinstitute/hellbender/IntelDeflaterTest.java
droazen: `IntegrationTest`

src/test/java/org/broadinstitute/hellbender/IntelDeflaterTest.java
droazen: File a ticket and assign to an IBM dev

lbergelson: good idea

lbergelson:
@akiezun Back to you.  I can move the changes to .travis.yml to a different pull request if you prefer.  I think they're useful, but they're essentially unrelated to this.  I think the changes to the test logging should probably stay in this pull request though.

build.gradle
akiezun: cool

.travis.yml
akiezun: ok, fine by me

akiezun:
lgtm, merge when ready

lbergelson:
closed #1738 and closes #1739

Issue #1741, opened 2016-04-20
async writer thread leak in test suite


akiezun:
this guy never dies

```
"SAMFileWriterThread-187" #471 daemon prio=5 os_prio=31 tid=0x00007f818d4b3000 nid=0xd98b waiting on condition [0x0000000126530000]
   java.lang.Thread.State: TIMED_WAITING (parking)
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00000007507b5178> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
        at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.awaitNanos(AbstractQueuedSynchronizer.java:2078)
        at java.util.concurrent.ArrayBlockingQueue.poll(ArrayBlockingQueue.java:418)
        at htsjdk.samtools.util.AbstractAsyncWriter$WriterRunnable.run(AbstractAsyncWriter.java:116)
        at java.lang.Thread.run(Thread.java:745)
```

akiezun:
allocating stack trace
```
   java.lang.Exception
        at htsjdk.samtools.util.AbstractAsyncWriter.<init>(AbstractAsyncWriter.java:42)
        at htsjdk.samtools.AsyncSAMFileWriter.<init>(AsyncSAMFileWriter.java:32)
        at htsjdk.samtools.SAMFileWriterFactory.makeBAMWriter(SAMFileWriterFactory.java:216)
        at htsjdk.samtools.SAMFileWriterFactory.makeBAMWriter(SAMFileWriterFactory.java:188)
        at htsjdk.samtools.SAMFileWriterFactory.makeSAMOrBAMWriter(SAMFileWriterFactory.java:324)
        at htsjdk.samtools.SAMFileWriterFactory.makeWriter(SAMFileWriterFactory.java:343)
        at org.broadinstitute.hellbender.utils.read.ReadUtils.createCommonSAMWriterFromFactory(ReadUtils.java:949)
        at org.broadinstitute.hellbender.utils.read.ReadUtils.createCommonSAMWriter(ReadUtils.java:921)
        at org.broadinstitute.hellbender.engine.GATKTool.createSAMWriter(GATKTool.java:405)
[...]

PR #1743, opened 2016-04-20
close all resources regardless of success

akiezun:
fixes https://github.com/broadinstitute/gatk/issues/1741 by closing the writer regardless of success. Added an `onTraversalDone` template method to `GATKTool` to take care of this. Otherwise we'll always leak resources on exceptions.

@droazen please review

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/ClipReads.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/ClipReads.java
@@ -305,13 +305,20 @@ public final class ClipReads extends ReadWalker {
     public ClippingData onTraversalSuccess(){
         if ( outputStats != null ){
             outputStats.printf(accumulator.toString());
+        }
+        return accumulator;
+    }
+
+    @Override
+    public void closeTool() {
+        if ( outputStats != null ){
             outputStats.close();
         }
         if ( outputBam != null ) {
             outputBam.close();
         }
-        return accumulator;
     }
+
     /**
      * Helper function that adds a seq with name and bases (as bytes) to the list of sequences to be clipped
      *
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
@@ -396,9 +396,10 @@ public final class VariantFiltration extends VariantWalker {
     }
 
     @Override
-    public Object onTraversalSuccess() {
-        writer.close();
-        return null;
+    public void closeTool() {
+        if (writer != null){
+            writer.close();
+        }
     }
 
 }
--- a/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
@@ -512,10 +512,29 @@ public abstract class GATKTool extends CommandLineProgram {
 
     @Override
     protected final Object doWork() {
-        onTraversalStart();
-        progressMeter.start();
-        traverse();
-        progressMeter.stop();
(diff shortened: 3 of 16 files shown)


droazen:
@akiezun Do we really need both `onTraversalDone()` and `onTraversalSuccess()`? This could be a major source of confusion. Is there any tool that calls `onTraversalSuccess()` but not `onTraversalDone()`? Maybe we should just rename `onTraversalSuccess()` back to `onTraversalDone()`, put it in the finally block, and keep things simple?

droazen:
If we do keep both methods, we should make it VERY clear in the docs that `onTraversalDone()` should be for closing resources, and `onTraversalSucess()` should be for producing final outputs, etc.

lbergelson:
Maybe it would be clearer to rename `onTraversalDone()` -> `cleanup()`  It seems useful to offer both options, but it is a bit confusing still.

akiezun:
i'm happy with any of those. I agree that almost no one cares about `onTraversalSuccess` as opposed to `onTraversalDone`. We could indeed just delete the former

droazen:
@akiezun I think there's a case to be made for having the ability to separate closure of resources from generation of final output on success, even if it's not currently needed -- I'd be ok with keeping both methods provided we can settle on the right names to avoid confusion, and provided we update the docs to make it clear when each method should be overridden. @lbergelson's suggestion of `onTraversalSuccess()` and `cleanup()` seems reasonable.

akiezun:
@droazen what do you want to do? I'm fine with whatever works. `cleanup` is fine by me

droazen:
@akiezun I vote that we rename `onTraversalDone()` to `closeTool()`, and keep `onTraversalSuccess()` as-is. Could you make this change, and make sure that in all tools resource closing is happening in `closeTool()`, and final output, etc., is happening in  `onTraversalSuccess()`?

Back to @akiezun

akiezun:
done. I don't see any more resources being closed in `onTraversalSuccess`, all have been moved to `closeTool`. 

back to @droazen

src/main/java/org/broadinstitute/hellbender/tools/ClipReads.java
```diff
@@ -303,15 +303,20 @@ public void apply( GATKRead read, ReferenceContext ref, FeatureContext featureCo
 
     @Override
     public ClippingData onTraversalSuccess(){
+        return accumulator;
+    }
+
+    @Override
+    public void closeTool() {
         if ( outputStats != null ){
             outputStats.printf(accumulator.toString());
```
droazen: Move print statement to `onTraversalSuccess()`, but keep `outputStats.close();` here in `closeTool()`

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
```diff
@@ -396,9 +396,8 @@ private boolean areClusteredSNPs(final FeatureContext featureContext, final Vari
     }
 
     @Override
-    public Object onTraversalSuccess() {
+    public void closeTool() {
         writer.close();
```
droazen: `if ( writer != null )` ?

droazen:
Review complete, back to @akiezun, merge when ready

src/main/java/org/broadinstitute/hellbender/tools/ClipReads.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
akiezun: good catch - now that closeTool is called on exceptional paths, this can be called the field is initialized

PR #1748, opened 2016-04-21
Porting of vector PairHMM to POWER8

t-ogasawara:
* Modified the files so that we can build libVectorLoglessPairHMM.so on Ubuntu/ppc64le platform
* Restored and modified the files for 128-bit vector that are on GATK3
* Added a new file to replace AVX with POWER8 vector instructions
* [Question] Is any unit test included in the repository to test the library?
* Confirmed that the library was built on Ubuntu 15.10/ppc64le
```
./gradlew installAll
:downloadGsaLibFile UP-TO-DATE
:extractIntelDeflater
:compileJava
:processResources
:classes
:compileVectorLoglessPairHMMSharedLibraryVectorLoglessPairHMMCpp
:linkVectorLoglessPairHMMSharedLibrary
:copySharedLib
:jar
:startScripts
:installDist
:sparkJar
:installSpark
:installAll

BUILD SUCCESSFUL
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -422,8 +422,15 @@ model {
                     println("WARNING: Building the native AVX PairHMM library is not supported on this OS.")
                 } else {
                     cppCompiler.args "-I", "${System.properties['java.home']}/../include"
-                    cppCompiler.args "-mavx"
                     cppCompiler.args "-Wall"
+		    if(System.getProperty("os.arch") == "ppc64le") {
+                      cppCompiler.args "-mcpu=power8"
+                      cppCompiler.args "-mtune=power8"
+                      cppCompiler.args "-fopenmp"
+                      linker.args "-lgomp"
+		    } else {
+                      cppCompiler.args "-mavx"
+		    }
 
                     // build with gcov options on Travis
                     if (System.env.CI.toString().toBoolean()) {
@@ -466,8 +473,15 @@ model {
         if (org.gradle.internal.os.OperatingSystem.current().isMacOsX()) {
             clang(Clang)
         } else {
-            gcc(Gcc)
+            gcc(Gcc) {
+	      target("linux_ppc64le")	// tentative fix until gradle supports linux ppc64le
+	    }
         }
+    }
+    platforms {
+        linux_ppc64le { 		// tentative fix until gradle supports linux ppc64le
+	    architecture "ppc64le"
+	}
     }
 }
 
--- a/src/main/cpp/VectorLoglessPairHMM/LoadTimeInitializer.cc
+++ b/src/main/cpp/VectorLoglessPairHMM/LoadTimeInitializer.cc
@@ -20,10 +20,12 @@ LoadTimeInitializer g_load_time_initializer;
 LoadTimeInitializer::LoadTimeInitializer()		//will be called when library is loaded
 {
   ConvertChar::init();
+#if defined(__x86_64__)
   //Very important to get good performance on Intel processors
   //Function: enabling FTZ converts denormals to 0 in hardware
   //Denormals cause microcode to insert uops into the core causing big slowdown
   _MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_ON);
+#endif
 
   //Profiling: times for compute and transfer (either bytes copied or pointers copied)
   m_compute_time = 0;
--- a/src/main/cpp/VectorLoglessPairHMM/README.md
+++ b/src/main/cpp/VectorLoglessPairHMM/README.md
@@ -1,5 +1,4 @@
-This code implements PairHMM using AVX instructions.
+This code implements PairHMM using vector instructions.
 
 Todo:
- * Consider renaming VectorLoglessPairHMM to AVXLoglessPairHMM
- * Add code to report if PairHMM AVX *can run* (AVX is supported and the native library can be loaded)
+ * Add code to report if vector PairHMM *can run* (the vector instruction set is supported and the native library can be loaded)
--- a/src/main/cpp/VectorLoglessPairHMM/avx_function_instantiations.cc
+++ b/src/main/cpp/VectorLoglessPairHMM/avx_function_instantiations.cc
@@ -1,4 +1,6 @@
+#if defined(__x86_64__)
 #define SIMD_ENGINE avx
+#define SIMD_ENGINE_AVX
 
 #include "template.h"
 
@@ -12,3 +14,4 @@
 
 template double compute_full_prob_avxd<double>(testcase* tc, double* nextlog);
 template float compute_full_prob_avxs<float>(testcase* tc, float* nextlog);
+#endif
--- a/src/main/cpp/VectorLoglessPairHMM/common_data_structure.h
+++ b/src/main/cpp/VectorLoglessPairHMM/common_data_structure.h
@@ -2,6 +2,9 @@
 #define COMMON_DATA_STRUCTURE_H
 
 #include "headers.h"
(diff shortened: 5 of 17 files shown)


droazen:
@gspowley Could you or someone else from Intel do a first-pass review of this?

t-ogasawara:
CI didn't fail when I opened the pull request, however, one of the two CI runs failed when I commits the additional changes to two files (one for unit test of VectorPairHMM and one for Java API of VectorPairHMM). I found that PrintReads failed in the log output, but I could not figure out the root cause.
```
$ ./gatk-launch PrintReads -I src/test/resources/NA12878.chr17_69k_70k.dictFix.bam -O output.bam
Missing GATK wrapper script: /home/travis/build/broadinstitute/gatk/build/install/gatk/bin/gatk

To generate the wrapper run:

    /home/travis/build/broadinstitute/gatk/gradlew installDist

The command "./gatk-launch PrintReads -I src/test/resources/NA12878.chr17_69k_70k.dictFix.bam -O output.bam" exited with 3.
```

akiezun:
I'll take this - we have a proposal for the way forward

akiezun:
@t-ogasawara @frank-y-liu @gspowley @paolonarvaez @droazen @lbergelson  please comment on the following proposal.
 
The proposal is that we would spin off native PairHMM as a separate project/repo on github and host AVX code there and have alternative implementations extend that project/repo (by creating repos that depend on the AVX one). 

In other words, now we have 1 repo, broadinstitute/gatk. After the proposed change we'll have 3 repos (all BSD licensed):
1) broadinstitute/gatk
2) broadinstitute/nativePairHMM-AVX
2) broadinstitute/nativePairHMM-PPC

We will duplicate the native code (AVX and PPC will be separate copies of C++ files etc) to simplify the testing burden. The parties interested in working on a specific architecture will contribute code directly to the respective architecture-specific repo and gatk will take occasional updates of those repos.

The gatk repo will depend on the other two. The PPC repo will depend on the AVX repo (and any other native repos will depend on the AVX one)

The avx and ppc repos will have their own build systems and unit tests against the new interface.

The AVX repo will expose something like the following API (to be worked out in detail).

```
//Used to copy references to byteArrays to JNI from reads
public final class JNIReadDataHolderClass {
    public byte[] readBases = null;
    public byte[] readQuals = null;
    public byte[] insertionGOP = null;
    public byte[] deletionGOP = null;
    public byte[] overallGCP = null;
}

//Used to copy references to byteArrays to JNI from haplotypes
public final class JNIHaplotypeDataHolderClass {
    public byte[] haplotypeBases = null;
}

public interface NativePairHMMKernel extends AutoCloseable { 

/**
 * Function to initialize the fields of JNIReadDataHolderClass and JNIHaplotypeDataHolderClass from JVM.
 * C++ code gets FieldIDs for these classes once and re-uses these IDs for the remainder of the program. Field IDs do not
 * change per JVM session
 *
 * @param readDataHolderClass      class type of JNIReadDataHolderClass
 * @param haplotypeDataHolderClass class type of JNIHaplotypeDataHolderClass
 */
void jniInitializeClassFields(Class<JNIReadDataHolderClass> readDataHolderClass, Class<JNIHaplotypeDataHolderClass> haplotypeDataHolderClass);

/**
 * Real compute kernel
 */
void jniComputeLikelihoods(int numReads, int numHaplotypes, JNIReadDataHolderClass[] readDataArray,
                                  JNIHaplotypeDataHolderClass[] haplotypeDataArray, double[] likelihoodArray, int maxNumThreadsToUse);

/**
 * Print final profiling information from native code. 
 */
default void close() { jniClose(); }

void jniClose();
}
```
and a class that implements those as native methods 
```
public class AVXNativePairHMMKernel implements NativePairHMMKernel{

@Override
native void jniInitializeClassFields(Class<JNIReadDataHolderClass> readDataHolderClass, Class<JNIHaplotypeDataHolderClass> haplotypeDataHolderClass);

@Override
native void jniComputeLikelihoods(int numReads, int numHaplotypes, JNIReadDataHolderClass[] readDataArray,
                                  JNIHaplotypeDataHolderClass[] haplotypeDataArray, double[] likelihoodArray, int maxNumThreadsToUse);
@Override
native void jniClose();
}
```
The PPC repo will implement `NativePairHMMKernel` as well and have its own tests

akiezun:
marked as blocked on a decision

t-ogasawara:
@akiezun 
I agree with separation of a native code repo from gatk. But can we have a single source tree for the native code? It will be useful for code maintenance to keep the common files in a single repository since 75% of the code are common between AVX and PPC. The source tree originally had the file set for the 128-bit vector ISA in GATK3, and the PPC porting adds only one file to the file set.

We can split the files into the common and cpu-specific file sets. You propose two repos for the native library. Can you create a single source tree without code duplication (below) by using the two repos?

```
common/    (~1500 lines)
    baseline.cc
    common_data_structure.h
    headers.h
    jni_common.h
    LoadTimeInitializer.{h,cc}
    org_broadinstitute_hellbender_utils_pairhmm_VectorLoglessPairHMM.{h,cc}
    pairhmm-template-kernel.cc
    template.h
    utils.{h,cc}
avx/  (~450 lines, including the file set for the 256-bit vector ISA; enabled for AVX)
    define-double.h
    define-float.h
    function_instantiations.cc (old name: avx_function_instantiations.cc)
    headers_md.h
    shift_template.cc
    template_md.h
    utils_md.{h,cc}
power8/  (~490 lines, including the file set for the 128-bit vector ISA; enabled for POWER8 (optional: SSE))
    define-double.h    (old name: define-sse-double.h)
    define-float.h    (old name: define-sse-float.h)
    function_instantiations.cc (old name: sse_function_instantiations.cc)
    headers_md.h
    power8.h
    shift_template.cc
    template_md.h
    utils_md.{h,cc}
```

droazen:
The problem currently with having everything in one repo is that we have no way of automatically testing on PPC. So an Intel engineer (for example) making a change to the x86 implementation would have to also find a way to manually test on PPC, which adds considerably to the testing/development burden. With separate repos, a change in the AVX repo wouldn't affect the PPC repo until the maintainer of the PPC repo updated it (after doing appropriate testing of their own). There could also still be some level of code sharing between the two implementations, since the PPC repo would depend on the AVX repo.

If someone could provide us with a means of automatically testing on PPC, with ability to see the test status on github, then I think we'd be more open to the idea of a single repo.

t-ogasawara:
@droazen [OSU Open Source Lab](http://osuosl.org/services/powerdev) provides the POWER8 cluster for open-source projects. Is it usable for your testing on PPC? Many open-source projects are using it. 

With the source tree in a single repo that I propose, changes that are specific to AVX will be made only for the files under "avx/" directory, which are not used for building the PPC binary. For example, build.gradle will specify "avx/" when building the binary on x86_64 ("power8/" on ppc64le). If the files under "common/" are changed (e.g., the package name is renamed from hellbender to gatk4), the changes should work on PPC if the tests don't fail on x86_64.

droazen:
@t-ogasawara Thanks for pointing us to the OSU service -- that looks like just what we need! I'll see if I can obtain access for our project. If we can get access and set up automated continuous tests for PPC, then I think we can use a single repo for the two architectures as you propose. Without continuous testing, though, I'm not convinced that `if the files under common are changed, the changes should work on PPC if the tests don't fail on x86_64.` -- at least, we would not be comfortable making changes to common code without tests in place.

akiezun:
closing the PR in favor of the discussion at https://github.com/broadinstitute/gatk/issues/1788

PR #1701, opened 2016-04-12
BWA-MEM call via process builder instead of JNI binding

SHuang-Broad:
Hi Adam, 

This is close-to-minimal implementation.
Would you please review, but don't merge (it's not mature yet)?

Thanks!

Issue #1517, opened 2016-02-24
add a binding to native BWA-MEM


akiezun:
we need a way to bind to a BWA-MEM library to align reads programmatically from within the GATK. 

SVs need it and it would simplify the reads pipeline to have it start with a fasta.

Commandline behavior we need: `bwa mem -K 100000000 -p` (we may not care about multithreading)
* `-K` is undocumented - reading code shows it refers to `fixed_chunk_size` which then sets `actual_chunk_size` @lh3 can you help us understand what it `-K` does?
* `-p` is for interleaved input

akiezun:
@SHuang-Broad @cwhelan this is on your radar already right?

cwhelan:
On our radar, yes. There is some question over whether we'd want to bind straight to BWA or to wrappers from the Snowtools library, but I think that direct BWA bindings would be helpful one way or the other at some point in the future.

akiezun:
can one of you grab it then? up to you how you bind (raw or via snowtools)

lbergelson:
Did dominar write java bindings for BWA for dataflow?  Maybe we could ask someone who was involved in that in case it already exists.

Also, it looks like the highly prolific @lindenb implemented java bindings a long time ago https://github.com/lindenb/jbwa.  I'm not sure how complete they are and the last update was in 2013, but he might be interested in updating them if there is demand.  

It looks like there may be an ancient GSA implementation as well http://gatkforums.broadinstitute.org/gatk/discussion/60/bwa-c-bindings-retired

akiezun:
btw to lower the JNI cost we may want to batch calls to bwa-mem. Need to measure performance and see

lindenb:
jbwa was just a proof-of-concept and a pretext to learn JNI. The algorithm was built on the "simple" API for BWA, but the internal 'mem' algorithm is much more complicated and the last time (2013) I looked at it, it was not easy to use it as a library (e.g. too many things in the `main`). I remember people at http://cloudgene.uibk.ac.at/ used it. 

>  FYI, I try to use jbwa in a MapReduce approach. If you interested, I can
> keep u posted.
> We implemented a framework for the execution of MapReduce jobs
> graphically and therefore new use cases are always nice

akiezun:
I heard that genomeStrip has a binding

lbergelson:
@lindenb Thanks for the info.

cwhelan:
[...]

PR #1750, opened 2016-04-21
BWA on Spark experiment

tomwhite:
This is an experiment to see if it's possible to run BWA-MEM on Spark. (Please don't merge.) The basic idea is that it uses JNI to call BWA-MEM's align function to align a batch of read pairs in one go. I think it should complement the work that @SHuang-Broad has been doing in #1701. It would be great to get your (and @akiezun's) feedback on the direction here.

A few comments
* Building the native libraries is not integrated, and it's not using the Apache 2 licensed code. I think this could use some of the changes in #1701.
* The ref is assumed to be on the local FS for the moment - it should really be loaded from HDFS. Also, the output is a single SAM file on the local FS, not a sharded BAM as for the rest of the GATK Spark tools.
* It is assumed that read pairs are interleaved and reads in a pair are placed in the same split (by setting `hadoopbam.bam.keep-paired-reads-together`). However, that property only works for queryname sorted BAMs, which isn't the case here, so we need to relax that requirement in Hadoop-BAM.
* I haven't tried this on large inputs, so I don't know how well it performs.

To run, I used the following on a cluster

```
./gatk-launch BwaSpark \
  --ref /home/tom/workspace/jbwa/test/ref.fa \
  --input hdfs:///user/$USER/bwa/R.bam \
  --output /tmp/bwa.sam \
  -- \
  --sparkRunner SPARK --sparkMaster yarn-client \
  --driver-memory 3G \
  --num-executors 1 \
  --executor-cores 1 \
  --executor-memory 3G \
  --archives jbwa-native.tar#jbwa-native \
  --conf 'spark.executor.extraLibraryPath=jbwa-native'
```

The interesting bit is the use of Spark's `--archives` flag to copy a tarball of native libraries (which I built manually) to every executor, and unpacks it in the working directory. Then `spark.executor.extraLibraryPath` is set to add that path to the library path of the executor. This means that you don't have to rely on the native libraries being installed on every node in the cluster.


akiezun:
it does not compile. Is somethng missing?

tomwhite:
That's right. It's because I haven't added anything to gradle to build the native libraries. I suspect I could do that by incorporating that part of #1701. Should I go ahead and do that or wait for it to be merged?

SHuang-Broad:
@tomwhite , #1701 is not a JNI binding (I'm working on providing one), but a really naive way of calling bwa from cmdline. Right now if you want to bring in the "correct" branch of BWA, I think messing with @lindenb 's MAKEFILE (bottom lines) and clone only the Apache2 branch would be enough. Per discussion in #1517 , the only difference seem to be `bwa index` performance, which isn't a concern here.

tomwhite:
Thanks @SHuang-Broad, I'll give that a go.

BTW you might also be interested in this script for running BWA on YARN: https://github.com/broadinstitute/gatk/commit/4f8a800e0f8b85b8b44c5c70931be3e48aca6bca.

SHuang-Broad:
Great, thanks @tomwhite !

tomwhite:
@SHuang-Broad @akiezun I've updated the PR to build jbwa, and now the test is passing.

akiezun:
@tomwhite - the JNI libraries are now Apache2 licensed (and using Apache2 BWA) and so we are unblocked in using them (courtesy of the author, @lindenb https://github.com/lindenb/jbwa) . Can you add them to this PR so I can test it easily? Thanks

tomwhite:
@akiezun our comments overlapped. Let me update again to include the latest Apache 2 JNI libraries.

tomwhite:
Updated. Over to you @akiezun.

akiezun:
@tomwhite can this take fastq files as input?

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,210 @@
...
+
+@CommandLineProgramProperties(summary = "Runs BWA",
+        oneLineSummary = "BWA on Spark",
+        programGroup = SparkProgramGroup.class)
+public final class BwaSpark extends GATKSparkTool {
+
+    private static final long serialVersionUID = 1L;
+
+    @Argument(doc = "the reference", shortName = "ref",
+            fullName = "ref", optional = false)
```
akiezun: we have a special argument for the reference

akiezun:
is jbwa distributed on maven central? if yes, can you use that? If not, let's ask the author. @lindenb is jbwa on central?

Also, all jbwa is now Apache2 so we're good linking to it. No problems there.

lindenb:
> is jbwa on central?

@akiezun short answer: no :-)

akiezun:
@tomwhite this looks fine to me as a start for us to experiment with but I wish the install was simpler and so we should either include building of jbwa as part of our build or (preferable) use a pre-built version from maven central. wdyt?

back to @tomwhite

akiezun:
@lindenb - can you put it up? it would be super

lindenb:
@akiezun I've never done this. Do you have any pointer ?

akiezun:
@lbergelson can you help?

lbergelson:
@akiezun @lindenb I can help. It's probably going to be a slightly complicated process though, especially actually building the cross platform jar.  I assume we're targeting OSX and x86-64 to start with, and then hopefully expanding to POWER8 in the future?  

The general idea is to prebuild the c code for whatever platforms we want to support.  Then package that in a structured way into a jar, and write some java code which will detect the platform at runtime and extract the correct executable into a temporary location.  Then we can publish that jar as a maven artifact.

We have an example of how to do the extraction in the `VectorLoglessPairHMM` constructor.  It's not perfect and probably needs a bit of refactoring to make it more general but it's the right idea.  Other libraries that package native code have similar examples.  I.e. Snappy-java https://github.com/xerial/snappy-java

We're going to be performing similar packaging for other native dependencies that we have, so standardizing it is a good idea.

lindenb:
@akiezun @lbergelson FYI: Sorry, I don't have currently the time (and I must say, the need...) to mavenize jbwa on my spare time. Don't wait for me on this please.

akiezun:
@lindenb no worries. @lbergelson we do need a solution of either building jbwa in our build script or (better for everyone but more complex short-term) put it on maven.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
tomwhite: BWA needs to the load the reference as a file, but the current special argument doesn't support that (`ReferenceSource` just allows you to get the bases, not the file). Something to change in the future perhaps.

tomwhite:
> can this take fastq files as input?

@akiezun, no it only takes an unaligned BAM as input. This is because it's hard to bring the reads in a pair together without doing a shuffle, when the reads reside in distinct files. However, it's easy to convert from a pair of fastq file to BAM using Picard. E.g.

```
java -jar picard.jar FastqToSam FASTQ=ERR000589_1.1m.fastq FASTQ2=ERR000589_2.1m.fastq OUTPUT=ERR000589.1m.bam SAMPLE_NAME=ERR000589 SORT_ORDER=queryname
```

tomwhite:
I ran BWA-MEM on 1m read pairs. Here's a comparison for pure BWA-MEM versus the Spark version, using two threads in each case. (The Spark version had a single executor which was allocated two cores via the `--executor-cores` argument.) 

```
$BWA mem -K 100000 -v 3 -t 2 human_g1k_v37.fasta ERR000589_1.1m.fastq ERR000589_2.1m.fastq > ERR242035.1m.sam 2> foo.err
# Real time: 666.345 sec; CPU: 1267.746 sec

./gatk-launch BwaSpark \
  --threads 2 \
  --ref hdfs://bottou01.sjc.cloudera.com/user/$USER/bwa/human_g1k_v37.fasta \
  --input hdfs:///user/$USER/bwa/ERR000589.1m.bam \
  --output hdfs:///user/$USER/bwa/ERR000589-aligned.1m.bam \
  -- \
  --sparkRunner SPARK \
  --sparkMaster yarn-client \
  --driver-memory 3G \
  --num-executors 1 \
  --executor-cores 2 \
  --executor-memory 15G \
  --archives jbwa-native.tar#jbwa-native \
  --conf 'spark.executor.extraLibraryPath=jbwa-native'
BwaSpark - Total time to run tool: 932s
BwaSpark - Time to download reference: 32s 
BwaSpark - Time in bwa: 1575s
```

So BwaSpark was 40% slower overall in terms of elapsed time. Of course, this was only using a single executor, so if we used multiple executors the elapsed time could be made proportionately shorter (and this a valid use case for some). In terms of CPU time in the BWA algorithm itself, it was about 24% slower (1575/1268). The latter could be improved by optimizing the JNI path (e.g. by passing several arrays of primitives or Strings corresponding to the fields in a `ShortRead`, rather than a single array of `ShortRead`s and then translating it).

There are probably other improvements to make too, but I think the summary is that this approach is feasible.

akiezun:
@tomwhite please assign to me when ready for review

tomwhite:
@akiezun thanks. I think this is ready for review now. It would be good to merge something that works, even if there are future performance and usability improvements we can do later.

akiezun:
I agree, @tomwhite, that there's multiple opportunities for JNI speedup in https://github.com/lindenb/jbwa/blob/master/src/main/native/bwajni.c

akiezun:
@tomwhite I did the following to get it to work
1. installed jbwa to `lib` like this (I think that's better than `~`)
```
mkdir lib
cd lib/
mkdir jbwa
git clone https://github.com/lindenb/jbwa   #this uses the apache2 version of bwa so it's all good
cd jbwa/
make
```

and changed build.gradle accordingly

2. add `"-Djava.library.path=lib/jbwa/src/main/native"` to `applicationDefaultJvmArgs` to find the library at runtime (not just in tests)

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,328 @@
...
+        protected Stopwatch initialValue() {
+            return Stopwatch.createUnstarted();
+        }
+    };
+
+    private static Set<Stopwatch> stopwatches = new HashSet<>();
+
+    @Argument(doc = "the reference", shortName = "ref",
+            fullName = "ref", optional = false)
+    private String ref;
```
akiezun: use `StandardArgumentDefinitions.REFERENCE_LONG_NAME` and StandardArgumentDefinitions.REFERENCE_SHORT_NAME` for the reference argument

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,328 @@
...
+
+    @Override
+    public boolean requiresReads() {
+        return true;
+    }
+
+    @Override
+    protected void runTool(final JavaSparkContext ctx) {
+        Stopwatch toolStopwatch = Stopwatch.createStarted();
+        ctx.hadoopConfiguration().setBoolean(BAMInputFormat.KEEP_PAIRED_READS_TOGETHER_PROPERTY, true);
```
akiezun: what reads this property? Can you add a comment?

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,328 @@
...
+        JavaRDD<GATKRead> unalignedReads = getReads();
+        JavaRDD<List<GATKRead>> unalignedPairs = unalignedReads.mapPartitions(iter -> () -> pairwise(iter));
```
akiezun: should this call to `pairwise` just be inlined? It may make it simpler to read I think

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,328 @@
...
+            String name1 = read1.getName();
+            String name2 = read2.getName();
+            byte[] baseQualities1 = SAMUtils.phredToFastq(read1.getBaseQualities()).getBytes();
+            byte[] baseQualities2 = SAMUtils.phredToFastq(read2.getBaseQualities()).getBytes();
+            return new Tuple2<>(
+                    new ShortRead(name1, read1.getBases(), baseQualities1),
+                    new ShortRead(name2, read2.getBases(), baseQualities2));
+        });
+
+        JavaRDD<String> samLines = shortReadPairs.mapPartitions(iter -> () -> {
```
akiezun: can you extract everything up to `samLines` into one method and then the rest into another method? Ie, make it 3 clear steps: read, align, write

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,328 @@
...
+            // provider is loaded using a URL classloader (e.g. in spark-submit).
+            ClassLoader cl = Thread.currentThread().getContextClassLoader();
+            if (cl == null) {
+                throw e;
+            }
+            return FileSystems.newFileSystem(uri, new HashMap<>(), cl).provider().getPath(uri);
+        }
+    }
+
+    static <T> Iterator<T> concat(Iterator<? extends Iterable<T>> iterator) {
```
akiezun: can you add a comment what this method does? Also, make it private or add tests for it

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,328 @@
...
+                    if (subIterator.hasNext()) {
+                        return subIterator.next();
+                    }
+                }
+                return endOfData();
+            }
+        };
+    }
+
+    private Iterator<List<String>> alignChunks(final BwaMem bwaMem, Iterator<Tuple2<ShortRead, ShortRead>> iter, int fixedChunkSize) {
```
akiezun: add 1 sentence what this method does

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,328 @@
...
+                String[] alignments = bwaMem.align(reads1, reads2);
+                stopwatch.stop();
+                return Arrays.asList(alignments);
+            } catch (IOException e) {
+                throw new GATKException(e.toString());
+            }
+        }, numThreads);
+    }
+
+    private <U> Iterator<List<Tuple2<ShortRead, ShortRead>>> chunk(final Iterator<Tuple2<ShortRead, ShortRead>> iterator, int fixedChunkSize) {
```
akiezun: add a short comment for this method - what it does

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,328 @@
...
+            bwaTime += stopwatch.elapsed(TimeUnit.MILLISECONDS);
+        }
+        log.info("Time in bwa: " + (bwaTime / 1000) + "s");
+    }
+
+    // From CreateSequenceDictionary
+
+    public int NUM_SEQUENCES = Integer.MAX_VALUE;
+
+    SAMSequenceDictionary makeSequenceDictionary(final File referenceFile) {
```
akiezun: make private

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,328 @@
...
+        return makeSequenceDictionary(referenceFile.toPath());
+    }
+
+    SAMSequenceDictionary makeSequenceDictionary(final Path referenceFile) {
```
akiezun: private

akiezun: can you delegate to `CreateSequenceDictionary.makeSequenceDictionary` or extract a method for both to use? @droazen is there a way to simply directly load a `.dict` file?

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -0,0 +1,328 @@
...
+                throw new UserException.MalformedFile("Sequence name appears more than once in reference: " + refSeq.getName());
+            }
+            sequenceNames.add(refSeq.getName());
+            ret.add(makeSequenceRecord(refSeq));
+        }
+        return new SAMSequenceDictionary(ret);
+    }
+
+    /**
+     * Create one SAMSequenceRecord from a single fasta sequence
```
akiezun: let's either get rid of this code (by reading a dict directly) or at least document that it's copied from `CreateSequenceDictionary.makeSequenceRecord`

akiezun:
can you add a test for the actual BWA alinment? Ie take some reads and align them and confirm that they get aligned properly (ie same way as if bwa was called on commandline)?

akiezun:
back to @tomwhite - some small refactor, building and testing suggestions.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
akiezun: @droazen just told me how to do it - use `new IndexedFastaSequenceFile(file).getSequenceDictionary()` (it will return null if not there and they you can use `CreateSequenceDictionary.makeSequenceDictionary`. Let's not have that copied code here.

src/test/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkIntegrationTest.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Test
+    public void test() throws Exception {
+        final File expectedSam = new File(getTestDataDir(), "bwa.sam");
+
+        final File ref = getTestFile("ref.fa");
+        final File input = getTestFile("R.bam");
+        final File output = createTempFile("bwa", ".bam");
+        output.delete();
```
sputnikci: \[FindBugs\] INFO: RV: Exceptional return value of java\.io\.File\.delete\(\) ignored in org\.broadinstitute\.hellbender\.tools\.spark\.bwa\.BwaSparkIntegrationTest\.test\(\)

sputnikci: \[FindBugs\] INFO: RV: Exceptional return value of java\.io\.File\.delete\(\) ignored in org\.broadinstitute\.hellbender\.tools\.spark\.bwa\.BwaSparkIntegrationTest\.test\(\)

sputnikci: \[FindBugs\] INFO: RV: Exceptional return value of java\.io\.File\.delete\(\) ignored in org\.broadinstitute\.hellbender\.tools\.spark\.bwa\.BwaSparkIntegrationTest\.test\(\)

sputnikci: \[FindBugs\] INFO: RV: Exceptional return value of java\.io\.File\.delete\(\) ignored in org\.broadinstitute\.hellbender\.tools\.spark\.bwa\.BwaSparkIntegrationTest\.test\(\)

tomwhite:
Thanks for the review @akiezun. I've addressed all your points. I dropped the sequence dictionary code, so now we rely on a .dict file being present (and if it isn't you can generate one using Picard's CreateSequenceDictionary command).

I think the only outstanding issue that is stopping this from being merged is integrating the build of jbwa. Currently you have to manually checkout and build jbwa yourself. The travis build does this automatically, but the local gradle build doesn't. Any thoughts about how to address this?

tomwhite:
The latest command line I used for testing:

```
./gatk-launch BwaSpark \
  --threads 2 \
  --reference hdfs://bottou01.sjc.cloudera.com/user/$USER/bwa/human_g1k_v37.fasta \
  --input hdfs:///user/$USER/bwa/ERR000589.100k.bam \
  --output hdfs:///user/$USER/bwa/ERR000589-aligned.100k.bam \
  --disableSequenceDictionaryValidation true \
  -- \
  --sparkRunner SPARK \
  --sparkMaster yarn-client \
  --driver-memory 3G \
  --num-executors 1 \
  --executor-cores 2 \
  --executor-memory 15G \
  --archives lib/jbwa/jbwa-native.tar#jbwa-native \
  --conf 'spark.executor.extraLibraryPath=jbwa-native'
```

akiezun:
ready for rereview?

tomwhite:
Yes, please take a look @akiezun.

akiezun:
@tomwhite about the building - ideally jbwa would be on maven central and have so and jnilib built in. Until that happens we can only
A) expect people to have it installed (seems bad), or
B) include script to install it, or
C) include pre-built binaries (.so, .jnilib and .jar)

I vote for B so that no steps are magic - (on the mac https://github.com/lindenb/jbwa/pull/5 was required but it's now merged). You can call the install script from gradle like this and make `installDist` depend on `installJBWA`

```
task installJBWA(type: Exec) {
  executable "sh"
  args "-x", "installJBWA.sh"
}
```

my  `installJBWA.sh` is 
```
mkdir -p lib
cd lib/
mkdir -p jbwa
git clone https://github.com/lindenb/jbwa
cd jbwa/
make
#Note: this rename is only needed on the Mac but does not cause problems elsewhere
cp -v src/main/native/libbwajni.so src/main/native/libbwajni.jnilib
``` 
wdyt?

akiezun:
back to @tomwhite to include the install script and we can merge it

droazen:
@akiezun I don't think we should add yet another Frankenstein monster step to our build process like this. I recommend instead that we package the library in Broad artifactory (after obtaining the author's permission), and depend on it from there.

akiezun:
do you know how?

droazen:
I'm sure we can figure it out based on the code we already have in build.gradle to upload gatk snapshots to the artifactory.

This seems much better than some gross shell script and more pain and suffering for people trying to build GATK4.

akiezun:
script is not pretty but there's no pain. It just builds it. And it solves
the problem of mac vs linux. In your scenario, how does that work?

droazen:
We just package both Mac and Linux versions of the library in a jar on the artifactory. Problem solved, no pain for users.

Your solution could potentially cause user pain if the native build fails due to missing dependencies (don't say it won't happen -- you said that about OpenMP and it happened yesterday!). It also doesn't address the packaging issue at all.

akiezun:
so you don't think https://github.com/broadinstitute/gatk/blob/master/scripts/upload_artifact_with_both_lib_and_dylib.sh is nasty and painful? 
(hint: it ssh's to gsa6 and builds and fetches files back! it requires that the user be on Broad network)

droazen:
@akiezun The difference is that `upload_artifact_with_both_lib_and_dylib.sh` is only run by us when we release -- it does not affect builds by users of the GATK, like your approach would.

We need to stop adding garbage to the GATK4 build system before it's too late and ends up like the GATK3 build system. We want GATK4 to be easy and reliable to build from source.

droazen:
Also, we've already agreed as a team that packaging pre-built native libraries for the platforms and OSes that we support into jars and depending on them via maven is the right thing to do for native dependencies of the GATK. We can do this right now for jbwa via the artifactory, with minimal additional effort on our part -- so why are you defending the "hack build.gradle some more and make GATK even harder to reliably build" approach to native code that we've already agreed is bad?

akiezun:
honestly, I really don't care which way we pick as long as it works. Can you make it work?

droazen:
I can help with the "publishing to artifactory" part if you provide me with pre-built Mac and Linux versions of the library and obtain permission from the author.

akiezun:
for the libraries: run my script from a few comments above. @lindenb can we put a pre built versions of this library on https://artifactory.broadinstitute.org/artifactory/webapp/home.html until a maven central artifact is created ?

lindenb:
>  @lindenb can we put a pre built versions of this library on https://artifactory.broadinstitute.org/artifactory/webapp/home.html until a maven central artifact is created ? 

yes, no problem.

droazen:
Alright, I'll try to get it published to the Broad artifactory by end of the week. Once it's there, this branch will need to be modified to depend on the version in artifactory, locate the library on the classpath, and extract and load it.

droazen:
This PR is waiting on https://github.com/broadinstitute/gatk/issues/1838

droazen:
PR resolving https://github.com/broadinstitute/gatk/issues/1838 has been opened:  https://github.com/broadinstitute/gatk/pull/1847. I included a method to extract and load the library from the classpath, as well as a unit test showing how to select between the Linux and Mac versions of the library.

droazen:
@tomwhite #1847 is in, and there's now a `JBWAIntegrationTest` showing how to load and use jbwa in a way that will work on both Linux and Mac. You should be all set -- let me know if there are problems.

PR #1756, opened 2016-04-21
remove SHUFFLE

akiezun:
fixes https://github.com/broadinstitute/gatk/issues/1203

@droazen, please review


droazen:
Why didn't you delete `ReferenceShard` as well? I think it was only used by SHUFFLE.

akiezun:
it's deleted now

akiezun:
it fails because cloud tests used to use shuffle and how they use BROADCAST and that requires a 2bit reference that is not available. What should we do? 
If we're reading the reference from a bucket why do we need BROADCAST?

droazen:
@akiezun Change the cloud tests to broadcast a 2bit reference stored in our test GCS bucket for now? Should work fine.

We need to benchmark the performance of broadcasting a 2bit reference in the cloud vs. reading it directly from a bucket from every worker, but this will require some code changes before we can do the comparison.

droazen:
This conflicts with https://github.com/broadinstitute/gatk/pull/1667, as discussed in person. We should probably close this for now until we have a story for broadcasting the reference when the input is cram rather than bam. However, @akiezun, it looks like you may have made some unrelated changes in this branch that could be split off into a separate PR (eg., "use 2 bit reference from our bucket")?

akiezun:
(funny, I was just typing in the same thing). sure, sounds good.

PR #1760, opened 2016-04-22
make tests run as root

akiezun:
our tests used to fail when run a root (because root can create files that others can't etc). This is a fix for it. Tested on a docker vm running gatk as root.

@lbergelson can you review?

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
@@ -7,6 +7,7 @@ import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.io.IOUtils;
 import org.broadinstitute.hellbender.utils.test.BaseTest;
 import org.testng.Assert;
+import org.testng.SkipException;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -356,18 +356,17 @@ public void testUnsuccessfulValidIndex(final int index, final int length) {
...
+        final File directory = createTempDir("Utils-RRFCD-Dir");
         final File nonExistingFile = createTempFile("Utils-RRFCD-NoFile", ".file");
         nonExistingFile.delete();
         final File nonReadable = createTempFile("Utils-RRFCD-NoReadable", ".file");
-        nonReadable.setReadable(false);
-        return new Object[][]{
-                {directory}, {nonExistingFile}, {nonReadable}
-        };
+        final boolean successSetNotReadable = nonReadable.setReadable(false) && !nonReadable.canRead();
+        if (successSetNotReadable) {
```
lbergelson: Could you add a comment here explaining why this is like this.  I would be confused if I came across this without context.

lbergelson: This would be nicer if it could generate a skip instead of fewer tests.  I don't see a clean way to do it though.  It's a bit weird for the total number of tests to change.  Maybe also add a log message when this happens?

lbergelson:
@akiezun Looks good, needs some explanation for confused future people I think.  We want it to be discoverable when the test suit has slightly different numbers of tests depending on where it runs.

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
akiezun: done

Issue #1711, opened 2016-04-15
GATK_SKIP_NATIVE_BUILD=true doesn't work on POWER systems


lbergelson:
A user building GATK on a POWER system tried to set `GATK_SKIP_NATIVE_BUILD=true`, but it didn't prevent the build from failing.

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'gatk'.
> Exception thrown while executing model rule: NativeComponentModelPlugin.Rules#createBinaries
> Invalid NativePlatform: linux_ppc64

BUILD FAILED
```
the temporary workaround was the following change:
```
$ diff build.gradle.org build.gradle
406c406,408
< VectorLoglessPairHMM(NativeLibrarySpec) {
---
> if(System.properties["os.arch"] != "ppc64")
> {
> VectorLoglessPairHMM(NativeLibrarySpec) {
458a461
> }
```

This is a bug and is likely to cause failures on other systems as well.

lbergelson:
The workaround was merged.  A more idiomatic gradle solution should ultimately be introduced.

PR #1761, opened 2016-04-22
disabled pairHMM binary build on PPC64LE; fixed floating comparison

frank-y-liu:
The fix will enable run gradle build and test on PowerPC while the pairHMM native binary build is being sorted out. It also introduced tolerance when compare two floating point numbers, which caused test failure on PowerPC.


frank-y-liu:
Where to find a GClOUD token in order to pass Travis CI build test? Or can we skip it?

build.gradle
```diff
@@ -406,7 +406,8 @@ check.dependsOn installDist
 
 model {
     components {
-        VectorLoglessPairHMM(NativeLibrarySpec) {
+        if (System.properties["os.arch"] != "ppc64le") {
+	VectorLoglessPairHMM(NativeLibrarySpec) {
```
lbergelson: Could you replace this tab with spaces.  Probably best to set you're IDE to replace tabs with spaces  when working on gatk.

droazen: (with 4 spaces)

lbergelson:
@frank-y-liu Thanks for the pull request!  Looks good to me except for a very minor nitpick about tabs.  In general we always use spaces.  You should be able to set your IDE to autoconvert them.  

I'm happy to merge without the cloud tests. It's a security hazard to let pull request from forks have access to those tokens, so they don't get passed to builds from forks.  We separate the clouds tests explicitly so they can be skipped without breaking the rest of the tests when this happens.  Your code should not effect any of the cloud functionality so I'm not worried if those tests didn't run.

We can give you direct push access as well if that's more convenient for you.  Then the tests will all run.

lbergelson:
Once you've fixed the tabs, could you [squash your commits into a single commit](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) and then I can merge.

lbergelson:
@frank-y-liu There may also be a failing test on power in `IntelDeflaterIntegrationTest`.  I wasn't sure what condition to use to exempt power from that test was.  I can't tell for sure though, if it works for you then I guess it works fine :)

frank-y-liu:
@lbergelson I've made the change, merged the code and am in the process of re-running the test. Will make another PR shortly.  I am fine for not passing cloud tests, if you guys think it's OK.

lbergelson:
@frank-y-liu Thanks, the best thing to do is just update this pull request by force pushing the rebased changes to your same branch.  No need to create a new one.

lbergelson:
Actually, I  can merge it from github's ui.

frank-y-liu:
@lbergelson @droazen While I was running gradle test, it occurred to me that spark tests are hogging a lot of resources. Each Power8 core can support 8 threads. On a big SMP machine, the statement "DEFAULT_SPARK_MASTER=local[*]" means a LOT of spark workers. Since we are running the test on a shared machine, the resource-hungry test is giving other applications some headaches. Is it possible to specify the number of local spark workers via an environmental variable? If the variable is not specified, then we can fall back to default "local[*]".

lbergelson:
@frank-y-liu That's a good idea.  Lets do it as a separate pull request.

lbergelson:
fixes #1711

frank-y-liu:
@lbergelson @droazen Indeed IntelDeflatorTest failed. There are four other failures related to BQSR. Will investigate them later. Thanks!

PR #1762, opened 2016-04-22
Remove the range check in ActivityProfileState.setIsActiveProb()

droazen:
Turns out that the BandPassActivityProfile can generate probabilities
that slightly exceed 1.0 when moving probability mass around. We should
eventually cap at 1.0, but first we must evaluate the effect of such a
cap on the HaplotypeCaller


droazen:
For @akiezun

droazen:
Created ticket https://github.com/broadinstitute/gatk/issues/1763 to eventually fix the underlying issue.

akiezun:
👍

PR #1764, opened 2016-04-22
set FTZ always on jniComputeLikelihoods

akiezun:
this solves a nasty precision issue in the HaplotypeCaller integration tests where tests would pass or fail depending on the order in which they ran!


droazen:
:+1: merge when ready

droazen:
(For third-party spectators: this solves a nasty precision issue in the HaplotypeCaller integration tests where tests would pass or fail depending on the order in which they ran!)

lbergelson:
@akiezun Should we open an issue to understand this better?  I feel like there might be lurking issues.

akiezun:
@gspowley can you help us understand why a reset of the FTZ flag is necessary?

PR #1765, opened 2016-04-23
don't use dataflow SerializableFunction

akiezun:
ongoing effort to reduce dependency on dataflow packages.

@lbergelson please have a look


lbergelson:
👍

PR #1772, opened 2016-04-25
adding gatk-launch to the jar

lbergelson:
packaging gatk-launch in our jar so that it's available to downstream projects
some build.gradle refactoring


lbergelson:
hmn, actually hold off on this, apparently the reason I didn't pull request this last week was because it wasn't ready..

PR #1775, opened 2016-04-26
Ability to override Kryo registrator.  Ugly, but minimal.

tedsharpe:


droazen:
@lbergelson can you review?

lbergelson:
@tedsharpe and I discussed in person.  This approach will cause problems for tests that rely on custom registrators.  Back to you Ted until it grows additional layers of indirection.

tedsharpe:
Back, atcha Louis.  You just set which registrator to use via a static method.  You can change it for any given unit test or suite, or even for a given phase of program execution within a Spark executable.  Crazy!

lbergelson:
@tedsharpe I'm still a bit worried about this.  Does `registerClasses` get called multiple times in the life time of the spark executor?  I had assumed it was invoked once at startup on a global Kryo instance and then that instance was serialized to all of the nodes.   I don't actually know that's the case though.   

Have you tested that this works if you have classes that have different registration requirements?  Could you add a test that shows that it works to reregister during execution?

PR #1776, opened 2016-04-26
user-specified spark core num; disable intelDeflator testing on PPC

frank-y-liu:
1. User can define the number of Spark cores in gradle test by environmental variable GATK_TEST_SPARK_CORES. If the variable is not defined, or the value is bogus, will fall back to default of "local[\*]"
2. Skip intelDeflator test on PPC platforms.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
@@ -1,5 +1,7 @@
 package org.broadinstitute.hellbender.engine.spark;
 
+import org.broadinstitute.hellbender.exceptions.UserException;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import org.apache.spark.SparkConf;
@@ -16,8 +18,9 @@ import java.util.Map;
  */
 public final class SparkContextFactory {
 
-    public static final String DEFAULT_SPARK_MASTER = "local[*]";
+    public static final String DEFAULT_SPARK_MASTER = determineDefaultSparkMaster();
     private static final boolean SPARK_DEBUG_ENABLED = Boolean.getBoolean("gatk.spark.debug");
+    private static final String SPARK_CORES_ENV_VARIABLE = "GATK_TEST_SPARK_CORES";
 
     /**
      * GATK will not run without these properties
@@ -152,5 +155,37 @@ public final class SparkContextFactory {
     private static JavaSparkContext createTestSparkContext(Map<String, String> overridingProperties) {
         final SparkConf sparkConf = setupSparkConf("TestContext", DEFAULT_SPARK_MASTER, DEFAULT_TEST_PROPERTIES, overridingProperties);
         return new JavaSparkContext(sparkConf);
+    }
+
+    /**
+     * Create the default Spark master, determines the number of cores it should use. Applicable to Spark test only.
+     *   Read the specification from the environmental variable GATK_TEST_SPARK_CORES
+     *      If the enviromental variable is not set,  use all available cores as in "local[*]"
+     *      If the value is a positive integer, use the value
+     *      If the value is invalid (strings, empty, etc), throw an UserException
+     *      If the value is a negative interger or zero, throw an UserException
+     */
+    private static String determineDefaultSparkMaster() {
+	final String defaultSparkMasterString = "local[*]";
+	String sparkMasterString;
+
+	String sparkSpecFromEnvironment = System.getenv( SPARK_CORES_ENV_VARIABLE );
+	if ( null == sparkSpecFromEnvironment ) {
+	    sparkMasterString = defaultSparkMasterString;
+	} else {
+	    int numSparkCoresFromEnv = 0;
+	    try {
+		numSparkCoresFromEnv = Integer.parseInt( sparkSpecFromEnvironment );
+	    } catch ( NumberFormatException e ) {
+		throw new UserException("Illegal number of cores specified in " + SPARK_CORES_ENV_VARIABLE + ". Positive integers only");
+	    }
+	    
+	    if ( numSparkCoresFromEnv > 0 ) {
+		sparkMasterString = String.format("local[%d]", numSparkCoresFromEnv );
+	    } else {
+		throw new UserException("Illegal number of cores specified in " + SPARK_CORES_ENV_VARIABLE + ". Number of cores must be positive");
+	    }
+	}
+	return sparkMasterString;
     }
 }
--- a/src/test/java/org/broadinstitute/hellbender/IntelDeflaterIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/IntelDeflaterIntegrationTest.java
@@ -2,6 +2,7 @@ package org.broadinstitute.hellbender;
 
 import htsjdk.samtools.util.zip.DeflaterFactory;
 import org.apache.commons.lang3.SystemUtils;
+import org.broadinstitute.hellbender.utils.NativeUtils;
 import org.broadinstitute.hellbender.utils.test.BaseTest;
 import org.testng.Assert;
 import org.testng.SkipException;
@@ -16,6 +17,10 @@ public class IntelDeflaterIntegrationTest extends BaseTest {
     public void testIntelDeflaterIsAvailable(){
         if ( !SystemUtils.IS_OS_LINUX ) {
             throw new SkipException("IntelDeflater not available on this platform");
+        }
+
+        if ( NativeUtils.isPPCArchitecture() ) {
+            throw new SkipException("IntelDeflater not available for this architecture");
         }
 
         Assert.assertTrue(DeflaterFactory.usingIntelDeflater(), "libIntelDeflater.so was not loaded. " +
--- a/src/main/java/org/broadinstitute/hellbender/utils/NativeUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/NativeUtils.java
@@ -1,0 +1,18 @@
+package org.broadinstitute.hellbender.utils;
+
+import org.apache.commons.lang3.SystemUtils;
+
+/**
+ * Utilities to provide architecture-dependent native functions
+ */
+public final class NativeUtils {
+
+    private NativeUtils(){}
+
+    /**
+     * boolean : returns whether the current platform is PowerPC, LE only
+     */
+    public static boolean isPPCArchitecture() {
+	return SystemUtils.OS_ARCH.equals("ppc64le");
+    }
+}


droazen:
@frank-y-liu Could you please rebase and squash the history in your branch into one commit before we review?

frank-y-liu:
@droazen ready to review?

droazen:
@frank-y-liu Something has gone slightly wrong in this branch git-wise -- it looks like you've duplicated some commits from master, and the merge commits in your history imply that your git workflow needs some tweaking.

In general, you want to always `rebase` rather than `merge` or `pull` (and avoid mixing the two, which can cause problems), since `rebase` produces a much cleaner history. Since you're working in a fork, you should create an "upstream" remote if you haven't already:

`git remote add upstream https://github.com/broadinstitute/gatk.git`

Then when you're working in a branch to which you've made one or more commits, and you want to update your branch with the latest changes from our master, do this:

`git fetch upstream`
`git rebase -i upstream/master`

This will bring up a screen allowing you to "squash" (combine) your work into a single commit that is suitable for merging into our master branch. If you do this with the current version of your branch, and select "squash" for all but the top commit, I believe you'll succeed in repairing your git history. 

Note that after each `rebase`, when you want to push your changes to github you'll need to do a `git push -f` instead of a simple `git push`, since `rebase` changes your commit history.

Try it out and let me know how it goes!

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
```diff
@@ -153,4 +153,26 @@ private static JavaSparkContext createTestSparkContext(Map<String, String> overr
...
+     * Determine the number of cores Spark master should use. Only used in Spark Test
+     *   Read the specification from the environmental variable GATK_TEST_SPARK_CORES
+     *      If the value is a valid positive integer, use it
+     *      If the value is bogus (strings, etc), or the env. var. is not set, use all available cores, as in "local[*]"
+     */
+
+    private static String determineSparkMaster() {
+        int foo = 0;
+        try {
+            foo = Integer.parseInt( System.getenv("GATK_TEST_SPARK_CORES") );
```
droazen: Check the return value of `System.getenv()` for null before passing it to `parseInt()` (it will be null if no environment variable is set).

droazen: Declare `GATK_TEST_SPARK_CORES` as a named constant at the top of the class instead of hardcoding it within this method.

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
```diff
@@ -153,4 +153,26 @@ private static JavaSparkContext createTestSparkContext(Map<String, String> overr
...
+        } catch ( NumberFormatException e ) {}
```
droazen: If `GATK_TEST_SPARK_CORES` is set (ie., `System.getenv()` returned non-null for it) and is non-numeric, throw a `UserException` with a descriptive error message instead of silently swallowing the error.

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
```diff
@@ -153,4 +153,26 @@ private static JavaSparkContext createTestSparkContext(Map<String, String> overr
...
     }
+
+    /**
+     * Determine the number of cores Spark master should use. Only used in Spark Test
+     *   Read the specification from the environmental variable GATK_TEST_SPARK_CORES
+     *      If the value is a valid positive integer, use it
+     *      If the value is bogus (strings, etc), or the env. var. is not set, use all available cores, as in "local[*]"
+     */
+
+    private static String determineSparkMaster() {
```
droazen: Rename to `determineDefaultSparkMaster()`

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
```diff
@@ -153,4 +153,26 @@ private static JavaSparkContext createTestSparkContext(Map<String, String> overr
...
         return new JavaSparkContext(sparkConf);
     }
+
+    /**
+     * Determine the number of cores Spark master should use. Only used in Spark Test
+     *   Read the specification from the environmental variable GATK_TEST_SPARK_CORES
+     *      If the value is a valid positive integer, use it
+     *      If the value is bogus (strings, etc), or the env. var. is not set, use all available cores, as in "local[*]"
+     */
+
```
droazen: Remove stray blank line

droazen:
First-pass review complete -- back to @frank-y-liu

frank-y-liu:
@droazen Let me work on it.

frank-y-liu:
@droazen  revised the code, but got error in rebasing:
`error: could not apply 37f19b8... disabled pairHMM binary build on PPC64LE; fixed floating comparison issue
`
Here is the list I saw when running: `git rebase -i upstream/master`:
```
pick 37f19b8 disabled pairHMM binary build on PPC64LE; fixed floating comparison issue
pick 02a5c45 fixed tab/space
pick 7229770 determine the number of spark cores to use from env. var.; disable IntelDeflator testing
pick 42d6661 don't use dataflow serializablefunction
pick 856243d adding gatk-launch to the jar
pick 71a6feb add IBM to AUTHORS
pick 23d9d6e determine the number of spark cores to use from env. var.
pick 3d0017f disable IntelDeflator testing; fix bug left behind
pick 3c078ad revised determineDefaultSparkMaster() method after review
```

I also don't recognize commits  `42d6`, `8562`.

droazen:
@frank-y-liu In these situations it often helps to do the squash and the rebase onto upstream separately. First, make sure you've cancelled the failed rebase via `git rebase --abort` (`git status` should then show no rebase in progress). Then, rebase onto your current divergence point from master using the following command:

    git rebase -i 37f19b8732ef8690e79d646da5e1db97ed009ac1^

(note the `^` at the end -- it's important!) When the pick/squash screen comes up, change all commits except for the first to "squash", then save. 

With the squash completed, you can then do the rebase onto upstream as a separate step:

    git fetch upstream    
    git rebase -i upstream/master

If you still get conflicts this time, fix them by editing the affected files (`git status` will tell you which files are in conflict), running `git add` on them to mark them as resolved, and then `git rebase --continue`.

droazen:
I'll add: I recommend trying to clean up your branch "the git way" first as described above, since it's good practice, but if that fails you can always go around git! :) Here's a fallback option for you that is guaranteed to succeed, in case the suggestion above doesn't work for you:

1. Copy the final version of your changes to the two files into a text editor of your choice.
2. In your branch, run the command `git reset --hard 814c9b78f1a4d4e90d6e07bd90f3caa6a8ecae5a`. This discards your branch's history and makes it into an exact copy of the GATK master branch.
3. Copy the changes you saved in your text editor back into the two source files.
4. `git commit -a`, then `git push -f`

frank-y-liu:
@droazen Had to do a merge. But did the clean way.

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
```diff
@@ -16,8 +18,9 @@
  */
 public final class SparkContextFactory {
 
-    public static final String DEFAULT_SPARK_MASTER = "local[*]";
+    public static final String DEFAULT_SPARK_MASTER = determineDefaultSparkMaster();
     private static final boolean SPARK_DEBUG_ENABLED = Boolean.getBoolean("gatk.spark.debug");
+    private static final String sparkSpecEnvName = "GATK_TEST_SPARK_CORES";
```
droazen: Name this `SPARK_CORES_ENV_VARIABLE`

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
```diff
@@ -153,4 +156,36 @@ private static JavaSparkContext createTestSparkContext(Map<String, String> overr
         final SparkConf sparkConf = setupSparkConf("TestContext", DEFAULT_SPARK_MASTER, DEFAULT_TEST_PROPERTIES, overridingProperties);
         return new JavaSparkContext(sparkConf);
     }
+
+    /**
+     * Determine the number of cores Spark master should use. Applicable to Spark test only.
```
droazen: Reword method description slightly: "Creates the default Spark master, and determines the number of cores it should use."

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
```diff
@@ -153,4 +156,36 @@ private static JavaSparkContext createTestSparkContext(Map<String, String> overr
...
+
+	String sparkSpecFromEnvironment = System.getenv( sparkSpecEnvName );
+	if ( null == sparkSpecFromEnvironment ) {
+	    sparkMasterString = defaultSparkMasterString;
+	} else {
+	    int numSparkCoresFromEnv = 0;
+	    try {
+		numSparkCoresFromEnv = Integer.parseInt( sparkSpecFromEnvironment );
+	    } catch ( NumberFormatException e ) {
+		throw new UserException("Illegal Spark master specification found in " + sparkSpecEnvName + ". Positive integers only");
```
droazen: "Illegal Spark master specification" -> "Illegal number of cores" (since the environment variables contains only a number, not the full Spark master String)

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
```diff
@@ -153,4 +156,36 @@ private static JavaSparkContext createTestSparkContext(Map<String, String> overr
...
+	    }
+	    
+	    if ( numSparkCoresFromEnv > 0 ) {
+		sparkMasterString = String.format("local[%d]", numSparkCoresFromEnv );
+	    } else {
+		throw new UserException("Illegal Spark master specificiation found in " + sparkSpecEnvName + ". Number of cores must be positive");
```
droazen: "Illegal Spark master specification" -> "Illegal number of cores"

src/test/java/org/broadinstitute/hellbender/IntelDeflaterIntegrationTest.java
```diff
@@ -18,6 +18,10 @@ public void testIntelDeflaterIsAvailable(){
             throw new SkipException("IntelDeflater not available on this platform");
         }
 
+        if ( SystemUtils.OS_ARCH.equals("ppc64le") ) {
```
droazen: Could you extract an `isPPCArchitecture()` method, add it to a new `NativeUtils` class in the `org.broadinstitute.hellbender.utils` package, and call that method from here? I think we might want to do this check in more than one place in the GATK.

droazen:
@frank-y-liu Looks good, only minor comments remaining. We can merge once they're addressed.

After you make the changes, I recommend using `git commit -a --amend` to commit them (this will add them to the existing commit in your branch instead of creating a new commit that you'd then have to squash), and `git push -f` to push them.

droazen:
Back to @frank-y-liu

src/test/java/org/broadinstitute/hellbender/IntelDeflaterIntegrationTest.java
frank-y-liu: @droazen Good idea. Looked under utils/. Should we open a new file named NativeUtils.java, or we should piggy back on Utils.java?

droazen: I think create a new `NativeUtils` class, as I suspect we'll be writing more methods like this in the future.

frank-y-liu: @droazen Sounds good. Should be done soon.

frank-y-liu:
@droazen Done.

droazen:
:+1: looks good! merging

PR #1781, opened 2016-04-27
add cram to testresources

akiezun:
adding cram version of our main test bam

@lbergelson, ok?


lbergelson:
@akiezun 👍 Lets wait for tests though just in case something horrible happens.

cmnbroad:
We might want to reconsider using a .crai file as an index (see https://github.com/samtools/htsjdk/issues/531 and https://github.com/samtools/hts-specs/issues/146). At this point, I would use a .bai (generate the .cram using GATK to get the .bai). It would also be interesting to know if the HC perf is any different using the .bai generated this way.

droazen:
Someone didn't squash and now owes everyone beer :)

akiezun:
hmm, maybe github does - the button said "confirm squash and merge"

droazen:
You're right, github did it for you -- the wonders of the modern age!

PR #1789, opened 2016-04-29
Update README.md - clarified that you can run dataproc from laptop

akiezun:
better sections, clarified that you can run dataproc from laptop (which I did not know until Louis told me)

@lbergelson please review


lbergelson:
looks good, didn't notice any typos..

PR #1792, opened 2016-04-29
save time if assuming 1 sample

akiezun:
`AlignmentContext.splitContextBySampleName` should be a no-op if we assume that there's 1 sample. Saves a few percent of HC. Requires a change to HaplotypeCaller to take advantage of this.

```
diff --git a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
index 97fe54a..c2f58d9 100644
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
@@ -440,7 +440,14 @@ public final class HaplotypeCallerEngine implements AssemblyRegionEvaluator {

         final int ploidy = activeRegionEvaluationGenotyperEngine.getConfiguration().genotypeArgs.samplePloidy;
         final List<Allele> noCall = GATKVariantContextUtils.noCallAlleles(ploidy); // used to noCall all genotypes until the exact model is applied
-        final Map<String, AlignmentContext> splitContexts = context.splitContextBySampleName(readsHeader);
+
+        final Map<String, AlignmentContext> splitContexts;
+        if (sampleSet.size() == 1) {
+            splitContexts = context.splitContextBySampleName(samplesList.getSample(0), readsHeader);
+        } else {
+            splitContexts = context.splitContextBySampleName(readsHeader);
+        }
+
         final GenotypesContext genotypes = GenotypesContext.create(splitContexts.keySet().size());
         final MathUtils.RunningAverage averageHQSoftClips = new MathUtils.RunningAverage();
         final InfiniteRandomMatingPopulationModel genotypingModel = genotypingEngine.getGenotypingModel();
```

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/AlignmentContext.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/AlignmentContext.java
@@ -1,17 +1,17 @@
 package org.broadinstitute.hellbender.engine;
 
 import htsjdk.samtools.SAMFileHeader;
-import htsjdk.samtools.SAMReadGroupRecord;
 import htsjdk.samtools.util.Locatable;
-import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.HasGenomeLocation;
 import org.broadinstitute.hellbender.utils.Utils;
-import org.broadinstitute.hellbender.utils.pileup.PileupElement;
 import org.broadinstitute.hellbender.utils.pileup.ReadPileup;
 import org.broadinstitute.hellbender.utils.read.ReadUtils;
 
-import java.util.*;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
 
 /**
  * Bundles together a pileup and a location.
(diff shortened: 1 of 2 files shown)


akiezun:
@droazen can you examine this?

src/main/java/org/broadinstitute/hellbender/engine/AlignmentContext.java
```diff
@@ -122,10 +122,16 @@ public AlignmentContext stratify(final ReadOrientation type) {
      * Splits the given AlignmentContext into a StratifiedAlignmentContext per sample, but referencd by sample name instead
      * of sample object.
      *
+     * @param assumedSingleSample assumed single sample in the file or null if there is none (eg file contains more than 1 sample).
```
droazen: Wording is a little confusing here -- how about: `If non-null, assume this is the only sample in our pileup and return immediately. If null, get the list of samples from the provided header.`

src/main/java/org/broadinstitute/hellbender/engine/AlignmentContext.java
```diff
@@ -122,10 +122,16 @@ public AlignmentContext stratify(final ReadOrientation type) {
...
+     *                            If this argument is not null then this method returns <code>Collections.singletonMap(assumedSingleSample, this);</code>
      * @return a Map of sample name to StratifiedAlignmentContext
      *
      **/
     public Map<String, AlignmentContext> splitContextBySampleName(final String assumedSingleSample, final SAMFileHeader header) {
+        if (assumedSingleSample != null){
+            return Collections.singletonMap(assumedSingleSample, this);
+        }
```
droazen: Add a unit test for `AlignmentContext` that covers the `assumedSingleSample != null` case.

src/main/java/org/broadinstitute/hellbender/engine/AlignmentContext.java
```diff
@@ -122,10 +122,16 @@ public AlignmentContext stratify(final ReadOrientation type) {
...
      * of sample object.
      *
+     * @param assumedSingleSample assumed single sample in the file or null if there is none (eg file contains more than 1 sample).
+     *                            If this argument is not null then this method returns <code>Collections.singletonMap(assumedSingleSample, this);</code>
      * @return a Map of sample name to StratifiedAlignmentContext
      *
      **/
     public Map<String, AlignmentContext> splitContextBySampleName(final String assumedSingleSample, final SAMFileHeader header) {
+        if (assumedSingleSample != null){
+            return Collections.singletonMap(assumedSingleSample, this);
```
droazen: The previous implementation would return a new `AlignmentContext` for the Map's value in this case -- is any client code relying on the return value being independent from `this`?

droazen:
Back to @akiezun, merge after addressing comments.

src/main/java/org/broadinstitute/hellbender/engine/AlignmentContext.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/engine/AlignmentContext.java
```diff
@@ -140,10 +146,7 @@ public AlignmentContext stratify(final ReadOrientation type) {
             if(sample != null) {
```
akiezun: note that the old code and the new code it still bogus - if `sample == null` then the the line up (`sample.equals`) blows up and so this check is useless. I'm fixing it and adding tests.

src/main/java/org/broadinstitute/hellbender/engine/AlignmentContext.java
akiezun: there were no unit tests for this class at all. I added a bunch now

src/main/java/org/broadinstitute/hellbender/engine/AlignmentContext.java
akiezun: i did a quick search. i don;t think so any of them is making any assumptions one way or the other

PR #1794, opened 2016-04-29
PileupElement : save memory and time by accessing bases and quals directly without copying. 

akiezun:
PileupElement: save memory and time by accessing bases and quals directly without copying. Showed up on profile in HC

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
@@ -14,7 +14,7 @@ import static org.broadinstitute.hellbender.utils.BaseUtils.Base.D;
 /**
  * Represents an individual base in a reads pileup.
  */
-public class PileupElement {
+public final class PileupElement {
 
     private static final EnumSet<CigarOperator> ON_GENOME_OPERATORS =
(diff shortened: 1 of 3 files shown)


akiezun:
@droazen mind looking at this tiny PR?

droazen:
:+1: Looks good, merge at will. Back to @akiezun

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -152,15 +152,15 @@ public int getOffset() {
...
+        return isDeletion() ? DELETION_BASE : read.getBase(offset);
     }
 
     /**
      * Get the base quality score of the base at this aligned position on the genome
      * @return a phred-scaled quality score as a byte
      */
     public byte getQual() {
-        return isDeletion() ? DELETION_QUAL : read.getBaseQualities()[offset];
+        return isDeletion() ? DELETION_QUAL : (byte)read.getBaseQuality(offset);
```
droazen: A question unrelated to this PR: why does `getBaseQuality()` return an int rather than a byte, given that `getBaseQualities()` returns a `byte[]`?

akiezun:
i fixed the int vs byte issue too

PR #1795, opened 2016-05-01
faster PileupElement.isBefore and isAfter queries with no linked list allocation.

akiezun:
those came up clearly on HaplotypeCaller profiles and allocated many megabytes of objects for now reason

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.hellbender.utils.pileup;
 
+import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.CigarElement;
 import htsjdk.samtools.CigarOperator;
 import org.broadinstitute.hellbender.utils.Utils;
@@ -27,7 +28,7 @@ public final class PileupElement {
     public static final byte G_FOLLOWED_BY_INSERTION_BASE = (byte) 90;
 
     private final GATKRead read;         // the read this base belongs to
-    protected final int offset;            // the offset in the bases array for this base
+    private final int offset;            // the offset in the bases array for this base
 
     private final CigarElement currentCigarElement;
     private final int currentCigarOffset;
@@ -128,7 +129,7 @@ public final class PileupElement {
     }
 
     /**
-     * Get the read for this pileup element
+     * Get the read for this pileup element. Returns the live object stored in this pileup element.
      * @return a non-null Read
      */
     public GATKRead getRead() {
@@ -137,7 +138,6 @@ public final class PileupElement {
 
     /**
      * Get the offset of the this element into the read that aligns that read's base to this genomic position.
-     *
      * If the current element is a deletion then offset is the offset of the last base containing offset.
      *
      * @return a valid offset into the read's bases
@@ -208,7 +208,7 @@ public final class PileupElement {
 
     /**
      * Get the mapping quality of the read of this element
-     * @return the mapping quality of the underlying SAM record
+     * @return the mapping quality of the underlying read
      */
     public int getMappingQual() {
         return read.getMappingQuality();
@@ -216,7 +216,7 @@ public final class PileupElement {
 
     @Override
     public String toString() {
-        return String.format("%s @ %d = %c Q%d", getRead().getName(), getOffset(), (char) getBase(), getQual());
+        return String.format("%s @ %d = %c Q%d", read.getName(), offset, (char) getBase(), getQual());
     }
 
     /**
@@ -252,8 +252,8 @@ public final class PileupElement {
      *
      * @return a non-null list of CigarElements
      */
-    public LinkedList<CigarElement> getBetweenPrevPosition() {
-        return atStartOfCurrentCigar() ? getBetween(Direction.PREV) : new LinkedList<>();
+    public List<CigarElement> getBetweenPrevPosition() {
+        return atStartOfCurrentCigar() ? getBetween(Direction.PREV) : Collections.emptyList();
     }
 
     /**
@@ -263,8 +263,8 @@ public final class PileupElement {
      *
      * @return a non-null list of CigarElements
      */
-    public LinkedList<CigarElement> getBetweenNextPosition() {
-        return atEndOfCurrentCigar() ? getBetween(Direction.NEXT) : new LinkedList<>();
+    public List<CigarElement> getBetweenNextPosition() {
+        return atEndOfCurrentCigar() ? getBetween(Direction.NEXT) : Collections.emptyList();
     }
 
     /**
@@ -317,18 +317,31 @@ public final class PileupElement {
         return offsetInCurrentCigar;
     }
 
-    /** for some helper functions */
-    private enum Direction { PREV, NEXT }
+    /**  Direction of search for some helper functions */
+    @VisibleForTesting
+    enum Direction { PREV(-1), NEXT(1);
+        private final int increment;
+
+        public int getIncrement() {
+            return increment;
+        }
+
+        private Direction(final int increment){
+            this.increment = increment;
+        }
+    }
 
     /**
-     * Helper function to get cigar elements between this and either the prev or next genomic position
+     * Helper function to get cigar elements between this and either the prev or next genomic position.
+     * Note that this method stops accumulating elements at the first "on-genome" operator it encounters.
      *
      * @param direction PREVIOUS if we want before, NEXT if we want after
      * @return a non-null list of cigar elements between this and the neighboring position in direction
      */
-    private LinkedList<CigarElement> getBetween(final Direction direction) {
-        final int increment = direction == Direction.NEXT ? 1 : -1;
-        LinkedList<CigarElement> elements = null;
+    private List<CigarElement> getBetween(final Direction direction) {
+        final int increment = direction.getIncrement();
+
+        final List<CigarElement> elements = new ArrayList<>();
         final List<CigarElement> cigarElements = read.getCigarElements();
         final int nCigarElements = cigarElements.size();
         for ( int i = currentCigarOffset + increment; i >= 0 && i < nCigarElements; i += increment) {
@@ -336,25 +349,32 @@ public final class PileupElement {
             if ( ON_GENOME_OPERATORS.contains(elt.getOperator()) ) {
                 break;
             } else {
-                // optimization: don't allocate list if not necessary
-                if ( elements == null ) {
-                    elements = new LinkedList<>();
-                }
-
-                if ( increment > 0 ) {
-                    // to keep the list in the right order, if we are incrementing positively add to the end
-                    elements.add(elt);
-                } else {
-                    // counting down => add to front
-                    elements.addFirst(elt);
-                }
+                elements.add(elt);
             }
         }
-
-        // optimization: elements is null because nothing got added, just return the empty list
-        return elements == null ? new LinkedList<>() : elements;
+        if (increment < 0){
+            Collections.reverse(elements);
+        }
+        return elements;
     }
 
+    /**
+     * Helper function to get cigar operator right next to this position.
+     *
+     * @param direction PREVIOUS if we want before, NEXT if we want after
+     * @return the next Cigar operator in the given direction, or null if there is none
+     */
+    @VisibleForTesting
+    CigarOperator getAdjacentOperator(final Direction direction) {
+        final int increment = direction.getIncrement();
+
+        final List<CigarElement> cigarElements = read.getCigarElements();
+        final int i = currentCigarOffset + increment;
+        if (i < 0 || i >= cigarElements.size()) {
+            return null;
+        }
+        return cigarElements.get(i).getOperator();
+    }
     /**
      * Get the cigar element of the previous genomic aligned position
      *
@@ -367,7 +387,7 @@ public final class PileupElement {
      * @return a CigarElement, or null (indicating that no previous element exists)
      */
     public CigarElement getPreviousOnGenomeCigarElement() {
-        return getNeighboringOnGenomeCigarElement(Direction.PREV);
+        return getNearestOnGenomeCigarElement(Direction.PREV);
     }
 
     /**
@@ -378,16 +398,16 @@ public final class PileupElement {
      * @return a CigarElement, or null (indicating that no next element exists)
      */
     public CigarElement getNextOnGenomeCigarElement() {
-        return getNeighboringOnGenomeCigarElement(Direction.NEXT);
+        return getNearestOnGenomeCigarElement(Direction.NEXT);
     }
 
     /**
      * Helper function to get the cigar element of the next or previous genomic position
      * @param direction the direction to look in
-     * @return a CigarElement, or null if no such element exists
+     * @return nearest on-genome CigarElement or null if no such element exists
      */
-    private CigarElement getNeighboringOnGenomeCigarElement(final Direction direction) {
-        final int increment = direction == Direction.NEXT ? 1 : -1;
+    private CigarElement getNearestOnGenomeCigarElement(final Direction direction) {
+        final int increment = direction.getIncrement();
         final int nCigarElements = read.numCigarElements();
 
         for ( int i = currentCigarOffset + increment; i >= 0 && i < nCigarElements; i += increment) {
@@ -415,35 +435,59 @@ public final class PileupElement {
     /**
      * Does an insertion occur immediately before the current position on the genome?
      *
-     * @return true if yes, false if no
+     * @return true if an insertion occurs immediately before the current position on the genome, false otherwise
      */
-    public boolean isAfterInsertion() { return isAfter(getBetweenPrevPosition(), CigarOperator.I); }
+    public boolean isAfterInsertion() { return isImmediatelyAfter(CigarOperator.I); }
 
     /**
      * Does an insertion occur immediately after the current position on the genome?
      *
-     * @return true if yes, false if no
+     * @return true an insertion occurs immediately after the current position on the genome, false otherwise
      */
-    public boolean isBeforeInsertion() { return isBefore(getBetweenNextPosition(), CigarOperator.I); }
+    public boolean isBeforeInsertion() {
+        return isImmediatelyBefore(CigarOperator.I);
+    }
 
     /**
      * Does a soft-clipping event occur immediately before the current position on the genome?
      *
-     * @return true if yes, false if no
+     * @return true if a soft-clipping event occurs immediately before the current position on the genome, false otherwise
      */
-    public boolean isAfterSoftClip() { return isAfter(getBetweenPrevPosition(), CigarOperator.S); }
+    public boolean isAfterSoftClip() {
+        return isImmediatelyAfter(CigarOperator.S);
+    }
 
     /**
      * Does a soft-clipping event occur immediately after the current position on the genome?
      *
-     * @return true if yes, false if no
+     * @return true if a soft-clipping event occurs immediately after the current position on the genome, false otherwise.
      */
-    public boolean isBeforeSoftClip() { return isBefore(getBetweenNextPosition(), CigarOperator.S); }
+    public boolean isBeforeSoftClip() {
+        return isImmediatelyBefore(CigarOperator.S);
+    }
+
+    /**
+     * @return true if a given operator event occurs immediately before the current position on the genome, false otherwise.
+     */
+    @VisibleForTesting
+    boolean isImmediatelyAfter(final CigarOperator op) {
+        Utils.nonNull(op);
+        return atStartOfCurrentCigar() && getAdjacentOperator(Direction.PREV) == op;
+    }
+
+    /**
+    * @return true if a given operator event occurs immediately after the current position on the genome, false otherwise.
+     */
+    @VisibleForTesting
+    boolean isImmediatelyBefore(final CigarOperator op) {
+        Utils.nonNull(op);
+        return atEndOfCurrentCigar() && getAdjacentOperator(Direction.NEXT) == op;
+    }
 
     /**
      * Does a soft-clipping event occur immediately before or after the current position on the genome?
      *
-     * @return true if yes, false if no
+     * @return true if a soft-clipping event occurs immediately before or after the current position on the genome, false otherwise.
      */
     public boolean isNextToSoftClip() { return isAfterSoftClip() || isBeforeSoftClip(); }
 
@@ -471,25 +515,4 @@ public final class PileupElement {
         return offsetInCurrentCigar == 0;
     }
 
-    /**
-     * Is op the last element in the list of elements?
-     *
-     * @param elements the elements to examine
-     * @param op the op we want the last element's op to equal
-     * @return true if op == last(elements).op
-     */
-    private boolean isAfter(final Deque<CigarElement> elements, final CigarOperator op) {
-        return ! elements.isEmpty() && elements.peekLast().getOperator() == op;
-    }
-
-    /**
-     * Is op the first element in the list of elements?
-     *
-     * @param elements the elements to examine
-     * @param op the op we want the last element's op to equal
-     * @return true if op == first(elements).op
-     */
-    private boolean isBefore(final List<CigarElement> elements, final CigarOperator op) {
-        return ! elements.isEmpty() && elements.get(0).getOperator() == op;
-    }
 }
--- a/src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.hellbender.utils.pileup;
 
+import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.util.Locatable;
 import org.broadinstitute.hellbender.exceptions.UserException;
@@ -61,6 +62,16 @@ public final class ReadPileup implements Iterable<PileupElement>{
      */
     public ReadPileup(final Locatable loc, final List<GATKRead> reads, final List<Integer> offsets) {
         this(loc, readsOffsetsToPileup(reads, offsets));
+    }
+
+    /**
+     * Returns the first element corresponding to the given read or null there is no such element.
+     * @param read or null if elements with no reads are to be retrieved from this pileup.
+     * @return the first element corresponding to the given read or null there is no such element.
+     */
+    @VisibleForTesting
+    PileupElement getElementForRead(final GATKRead read){
+        return getElementStream().filter(el -> Objects.equals(el.getRead(), read)).findAny().orElse(null);
     }
 
     /**
--- a/src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
@@ -13,9 +13,9 @@ import org.testng.Assert;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
-import java.util.Arrays;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.*;
+
+import static htsjdk.samtools.CigarOperator.*;
 
 /**
  * testing of the new (non-legacy) version of LocusIteratorByState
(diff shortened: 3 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -351,11 +346,27 @@ public int getOffsetInCurrentCigar() {
             }
         }
 
-        // optimization: elements is null because nothing got added, just return the empty list
         return elements == null ? new LinkedList<>() : elements;
     }
 
     /**
+     * Helper function to get cigar elements between this and either the prev or next genomic position
```
akiezun: hmm doc needs fixing

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -351,11 +346,27 @@ public int getOffsetInCurrentCigar() {
...
+     *
+     * @param direction PREVIOUS if we want before, NEXT if we want after
+     * @return the neighboring position in given direction or null if there is none.
+     */
+    private final CigarElement getNeighborElement(final Direction direction) {
+        final int increment = direction == Direction.NEXT ? 1 : -1;
```
lbergelson: Increment should be baked into the enum.

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -351,11 +346,27 @@ public int getOffsetInCurrentCigar() {
...
+
+        final List<CigarElement> cigarElements = read.getCigarElements();
+        final int i = currentCigarOffset + increment;
+        if (i >= 0 && i < cigarElements.size()){
+            return cigarElements.get(i);
+        } else {
+            return null;
+        }
+    }
```
droazen: Unit test for this method (make package accessible)

droazen: Check `ON_GENOME_OPERATORS`?

lbergelson:
@akiezun Is there a reason these are `LinkedLists`?  It seems like it should be List instead.  I tried to find usages that depend on the linkedlistness of it, but the only thing I see is that one private method relies on it being a `Deque`, but that could be trivially rewritten.   It seems like you'd avoid even more allocation by switching it to use List and then using the non-allocating `Collections.emptyList()`

`getBetween` uses the linkedness of it to use `addFirst`.  This could be rewritten to either build the list in the reverse order and reverse it, or maybe provide a guava `Lists.reverse` view.

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -415,30 +426,58 @@ private boolean hasOperator(final CigarElement maybeCigarElement, final CigarOpe
     /**
      * Does an insertion occur immediately before the current position on the genome?
      *
-     * @return true if yes, false if no
+     * @return true if an insertion occur immediately before the current position on the genome, false otherwise
```
lbergelson: should be `occurs` in all of these

lbergelson:
There's a very suboptimal operation in `getBetween`  It calls `read.getCigar().getCigarElement(i);` once for every element. `getCigar()` is an expensive operation that makes a copy every time.

droazen:
Address current comments and then I/we will re-review

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
akiezun: done

akiezun:
back to you - i removed the linked list type now. it used to be needed but no more. added some tests for new methods.

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -317,45 +317,69 @@ public int getOffsetInCurrentCigar() {
...
     }
 
     /**
+     * Helper function to get cigar operator right next to this position but only for not-on-genome operators.
+     *
+     * @param direction PREVIOUS if we want before, NEXT if we want after
+     * @return the neighboring position in given direction or null if there is none
+     */
+    @VisibleForTesting
+    CigarOperator getNeighborNotOnGenomeOperator(final Direction direction) {
```
droazen: `getNextNotOnGenomeOperator(Direction)` reads a little less awkwardly

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -317,45 +317,69 @@ public int getOffsetInCurrentCigar() {
...
+            Collections.reverse(elements);
+        }
+        return elements;
     }
 
     /**
+     * Helper function to get cigar operator right next to this position but only for not-on-genome operators.
+     *
+     * @param direction PREVIOUS if we want before, NEXT if we want after
+     * @return the neighboring position in given direction or null if there is none
```
droazen: `the next not-on-genome Cigar operator in the given direction, or null if there is none`

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -387,7 +411,7 @@ public CigarElement getNextOnGenomeCigarElement() {
      * @return a CigarElement, or null if no such element exists
      */
     private CigarElement getNeighboringOnGenomeCigarElement(final Direction direction) {
```
droazen: Make sure the name of this method is consistent with whatever name you end up with for the new method above.

droazen: Wait a sec...this method `getNeighboringOnGenomeCigarElement()` behaves very differently from `getNeighborNotOnGenomeOperator()`, since it iterates until it finds an "on-genome" operator, whereas your new method does not iterate. This is a somewhat confusing state of affairs -- can you clean this up to increase clarity?

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -317,45 +317,69 @@ public int getOffsetInCurrentCigar() {
...
     /**
      * Helper function to get cigar elements between this and either the prev or next genomic position
      *
      * @param direction PREVIOUS if we want before, NEXT if we want after
      * @return a non-null list of cigar elements between this and the neighboring position in direction
      */
-    private LinkedList<CigarElement> getBetween(final Direction direction) {
-        final int increment = direction == Direction.NEXT ? 1 : -1;
-        LinkedList<CigarElement> elements = null;
+    private List<CigarElement> getBetween(final Direction direction) {
```
droazen: Document the behavior of this confusing method -- especially the fact that it stops accumulating elements at the first "on-genome" operator it encounters

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -415,35 +439,59 @@ private boolean hasOperator(final CigarElement maybeCigarElement, final CigarOpe
...
+    }
+
+    /**
+    * @return true if a given operator event occurs immediately after the current position on the genome, false otherwise.
      */
-    public boolean isBeforeSoftClip() { return isBefore(getBetweenNextPosition(), CigarOperator.S); }
+    @VisibleForTesting
+    boolean isImmediatelyBefore(final CigarOperator op) {
+        Utils.nonNull(op);
+        return atEndOfCurrentCigar() && getNeighborNotOnGenomeOperator(Direction.NEXT) == op;
```
droazen: `isAfter()` didn't have this `atEndOfCurrentCigar()` check -- are you sure that your new code is equivalent to the previous version?

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -55,6 +56,16 @@ public ReadPileup(final Locatable loc, final List<GATKRead> reads, final List<In
     }
 
     /**
+     * Returns the element corresponding to the given read of null there is no such element.
```
droazen: of -> or

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -55,6 +56,16 @@ public ReadPileup(final Locatable loc, final List<GATKRead> reads, final List<In
...
+     * @param read or null if elements with no reads are to be retrieved from this pileup.
+     * @return
```
droazen: Complete your javadoc

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -55,6 +56,16 @@ public ReadPileup(final Locatable loc, final List<GATKRead> reads, final List<In
     }
 
     /**
+     * Returns the element corresponding to the given read of null there is no such element.
+     * @param read or null if elements with no reads are to be retrieved from this pileup.
```
droazen: Can there be an element with no read? What if there are multiple such elements? (document)

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
```diff
@@ -155,6 +157,8 @@ public void testPrevAndNextTest(final GATKRead read, final CigarOperator firstOp
         Assert.assertEquals(pe2.getBetweenPrevPosition().size(), ops.size());
         Assert.assertEquals(pe2.getBetweenNextPosition().size(), 0);
         assertEqualsOperators(pe2.getBetweenPrevPosition(), ops);
+        Assert.assertEquals(pe2.getNeighborNotOnGenomeOperator(PileupElement.Direction.PREV), ops.get(ops.size()-1));
+        Assert.assertEquals(pe2.getNeighborNotOnGenomeOperator(PileupElement.Direction.NEXT), null);
```
droazen: These tests for the new `getNeighborNotOnGenomeOperator()` method do not really seem adequate -- can we get a high-quality standalone unit test that properly exercises the method?

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
```diff
@@ -147,6 +147,8 @@ public void testPrevAndNextTest(final GATKRead read, final CigarOperator firstOp
         Assert.assertEquals(pe.getBetweenPrevPosition().size(), 0);
         assertEqualsOperators(pe.getBetweenNextPosition(), ops);
         Assert.assertEquals(pe.getPreviousOnGenomeCigarElement(), null);
+        Assert.assertEquals(pe.getNeighborNotOnGenomeOperator(PileupElement.Direction.PREV), null);
+        Assert.assertEquals(pe.getNeighborNotOnGenomeOperator(PileupElement.Direction.NEXT), ops.get(0));
```
droazen: The neighboring "not on genome" operator is always the first element in ops? What if `ops.get(0)` is an "on genome" operator?

droazen:
Can you please test this change with the `HaplotypeCaller` in protected and make sure nothing changes? In particular, can you run `HaplotypeCallerIntegrationTest` and `HaplotypeCallerEngineUnitTest` and make sure they pass? This PR makes me a little nervous given the centrality of the classes touched, even though the optimization itself is simple enough...

droazen:
Back to @akiezun

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
droazen: Also, where are the tests for `isImmediatelyBefore/After()`? They are `@VisibleForTesting` but not tested!

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -415,35 +439,59 @@ private boolean hasOperator(final CigarElement maybeCigarElement, final CigarOpe
...
      *
-     * @return true if yes, false if no
+     * @return true if a soft-clipping event occurs immediately after the current position on the genome, false otherwise.
+     */
+    public boolean isBeforeSoftClip() {
+        return isImmediatelyBefore(CigarOperator.S);
+    }
+
+    /**
+     * @return true if a given operator event occurs immediately before the current position on the genome, false otherwise.
```
droazen: These `isImmediatelyBefore/After` methods will not work at all if the `op` is an on-genome operator, but this is totally undocumented and unexpected.

akiezun:
hmm, this is turning into a major piece of work. Maybe i'll redo it from scratch and keep minimal.

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
akiezun: renamed the other method to `getAdjacentNotOnGenomeOperator` to indicate that it's the immediate neighbor

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
akiezun: added this text verbatim

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
akiezun: done. it returns the first one it finds or null if it finds none

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
akiezun: it;s a bit painful to test in separation - that's why it was private before

akiezun: added tests for isImmediatelyBefore/After

akiezun: extracted the tests to own method - let me know if this is better for you

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
akiezun: that can't be -  by the way the test is contructed the `ops` list only contains `C` `P` and `S`

akiezun:
FWIW HaplotypeCaller tests pass with these changes

akiezun:
back for rereview @droazen

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
droazen: Document the fact that this method `getNeighboringOnGenomeCigarElement()` iterates until it finds an "on-genome" operator.

droazen: Maybe rename to `getNearestOnGenomeCigarElement()` to make this clearer

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
```diff
@@ -160,6 +162,83 @@ public void testPrevAndNextTest(final GATKRead read, final CigarOperator firstOp
...
+
+        //after the first 'op' from the list
+        state.stepForwardOnGenome();
+        final PileupElement pe2 = state.makePileupElement();
+        Assert.assertEquals(pe2.getAdjacentOperator(PileupElement.Direction.PREV), ops.get(ops.size()-1));
+        Assert.assertEquals(pe2.getAdjacentOperator(PileupElement.Direction.NEXT), null);
+        for (final CigarOperator op : CigarOperator.values()){
+            Assert.assertFalse(pe2.isImmediatelyBefore(op), "immediately before " + op);
+            Assert.assertEquals(pe2.isImmediatelyAfter(op), op ==  ops.get(ops.size()-1), "immediately after " + op);
+        }
```
droazen: This test is really hard to follow and poorly designed -- can you completely replace it with a new test that:

-constructs PileupElements directly using a read that has a cigar with all possible cigar elements (don't use AlignmentStateMachine, please!), and various offsets into that read
-does direct, straightforward assertions on these PileupElements of the form: `pe.isImmediatelyBefore(CigarOperator.S)`, etc.
-includes test cases in which the current offset is not at the beginning/end of the current cigar element

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
```diff
@@ -160,6 +162,83 @@ public void testPrevAndNextTest(final GATKRead read, final CigarOperator firstOp
...
+                    tests.add(new Object[]{read, middleOp});
+                }
+            }
+        }
+
+        return tests.toArray(new Object[][]{});
+    }
+
+    @Test(dataProvider = "PrevAndNextTest_simple")
+    public void testImmediateBeforeAndAfterTest_simple(final GATKRead read, final CigarOperator middleOp) {
```
droazen: Same comment as above -- redesign to construct `PileupElements` directly rather than confusing things by going through `AlignmentStateMachine`

src/test/java/org/broadinstitute/hellbender/utils/pileup/ReadPileupUnitTest.java
```diff
@@ -328,6 +329,13 @@ public void testSimplePileupWithIndelsOffset(){
         final int offset = 4;
         final ReadPileup pu = new ReadPileup(loc, reads, offset);
 
+        final PileupElement firstElem = pu.getElementForRead(read1);
+        Assert.assertNull(firstElem.getAdjacentOperator(PileupElement.Direction.NEXT));
+        Assert.assertNull(firstElem.getAdjacentOperator(PileupElement.Direction.PREV));
+        final PileupElement secondElem = pu.getElementForRead(read2);
+        Assert.assertEquals(secondElem.getAdjacentOperator(PileupElement.Direction.NEXT), CigarOperator.I);
+        Assert.assertNull(secondElem.getAdjacentOperator(PileupElement.Direction.PREV));
```
droazen: Replace with high-quality direct testing on `getAdjacentOperator()` in `PileupElementUnitTest`

droazen:
Back to @akiezun. Code itself looks good now, just need to rename `getNeighboringOnGenomeCigarElement()`. Tests are TERRIBLE though and need to be re-written from scratch.

akiezun:
these things are super hard to create by hand though. I may just abandon this PR - the speedup may not be worth the effort here. I'll see

droazen:
It's not super-hard -- just create a `DataProvider` with a single read containing all the possible cigar elements. Then each test case will be: a `PileupElement` constructed from that read, an offset, expected immediately before element, and expected immediately after element. Super easy!

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
akiezun: renamed getNeighboringOnGenomeCigarElement  -> getNearestOnGenomeCigarElement

akiezun:
added some more direct tests (sorry about the whitespace changes). back to @droazen

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -378,16 +398,16 @@ public CigarElement getPreviousOnGenomeCigarElement() {
...
      */
     public CigarElement getNextOnGenomeCigarElement() {
-        return getNeighboringOnGenomeCigarElement(Direction.NEXT);
+        return getNearestOnGenomeCigarElement(Direction.NEXT);
     }
 
     /**
      * Helper function to get the cigar element of the next or previous genomic position
      * @param direction the direction to look in
      * @return a CigarElement, or null if no such element exists
```
droazen: `a CigarElement` -> `nearest on-genome CigarElement`

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
```diff
@@ -179,13 +258,171 @@ public void testCreatePileupForReadAndOffset1(){
...
+                Assert.assertFalse(pe.isImmediatelyBefore(op), "isImmediatelyBefore " + op);
+            }
+        }
+    }
+
+    @Test
+    public void testIsImmediatelyAfter_afterFirstD() {
+        final GATKRead read = ArtificialReadUtils.createArtificialRead("10M10D10M");
+        final PileupElement pe = PileupElement.createPileupForReadAndOffset(read, 10);
+        Assert.assertFalse(pe.atEndOfCurrentCigar(), "atEndOfCurrentCigar");
```
droazen: Also assert that `pe.atStartOfCurrentCigar()` is true

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
```diff
@@ -179,13 +258,171 @@ public void testCreatePileupForReadAndOffset1(){
...
+            }
+        }
+
+        for (final CigarOperator op : CigarOperator.values()) {
+            Assert.assertFalse(pe.isImmediatelyBefore(op));
+        }
+    }
+
+    @Test
+    public void testIsImmediatelyAfter_insideDs() {
```
droazen: This test case is not "inside" the deletion, is it? Isn't it inside the second M operator instead, since the deletion consumes no bases in the bases array?

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
```diff
@@ -179,13 +258,171 @@ public void testCreatePileupForReadAndOffset1(){
...
+        for (final CigarOperator op : CigarOperator.values()) {
+            Assert.assertFalse(pe.isImmediatelyAfter(op), "isImmediatelyAfter " + op);
+        }
+        for (final CigarOperator op : CigarOperator.values()) {
+            Assert.assertFalse(pe.isImmediatelyBefore(op), "isImmediatelyBefore " + op);
+        }
+    }
+
+    @DataProvider(name = "elementData")
+    private Object[][] elementData() {
```
droazen: Use a more descriptive name for your `DataProvider` -- eg., `adjacentElementTestData` or something similar

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
```diff
@@ -179,13 +258,171 @@ public void testCreatePileupForReadAndOffset1(){
...
+    public void testOperators(GATKRead read, int goodOffset, boolean atStartOfCurrentCigar, boolean atEndOfCurrentCigar, CigarOperator prev, CigarOperator next) throws Exception {
+        final PileupElement pe = PileupElement.createPileupForReadAndOffset(read, goodOffset);
+        Assert.assertEquals(pe.atStartOfCurrentCigar(), atStartOfCurrentCigar, "atStartOfCurrentCigar");
+        Assert.assertEquals(pe.atEndOfCurrentCigar(), atEndOfCurrentCigar, "atEndOfCurrentCigar");
+        Assert.assertEquals(pe.getAdjacentOperator(PileupElement.Direction.PREV), prev, "prev");
+        Assert.assertEquals(pe.getAdjacentOperator(PileupElement.Direction.NEXT), next, "next");
+        for (final CigarOperator op : CigarOperator.values()) {
+            Assert.assertEquals(pe.isImmediatelyAfter(op), atStartOfCurrentCigar && op == prev);
+            Assert.assertEquals(pe.isImmediatelyBefore(op), atEndOfCurrentCigar && op == next);
+        }
```
droazen: No need for a loop here -- replace with two ifs for clarity:

    if ( atStartOfCurrentCigar ) {
        Assert.assertTrue(pe.isImmediatelyAfter(prev);
    }

    if ( atEndOfCurrentCigar ) {
        Assert.assertTrue(pe.isImmediatelyBefore(next);
    }

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
```diff
@@ -179,13 +258,171 @@ public void testCreatePileupForReadAndOffset1(){
...
+                {read, 35, false, false, X, S},
+                {read, 36, false, false, X, S},
+                {read, 37, false, false, X, S},
+                {read, 38, false, false, X, S},
+                {read, 39, false, true, X, S},
+        };
+    }
+
+    @Test(dataProvider = "elementData")
+    public void testOperators(GATKRead read, int goodOffset, boolean atStartOfCurrentCigar, boolean atEndOfCurrentCigar, CigarOperator prev, CigarOperator next) throws Exception {
```
droazen: `testAdjacentElements` might be a better name

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
akiezun: i do want to check that it says `false` for other operators though

droazen:
Back to @akiezun -- minor remaining comments, then squash/merge

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
droazen: Ah yes, good point.

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
```diff
@@ -160,17 +162,94 @@ public void testPrevAndNextTest(final GATKRead read, final CigarOperator firstOp
         Assert.assertEquals(pe2.getNextOnGenomeCigarElement(), null);
     }
 
+    @Test(dataProvider = "PrevAndNextTest")
+    public void testImmediateBeforeAndAfterTest(final GATKRead read, final CigarOperator firstOp, final CigarOperator lastOp, final List<CigarOperator> ops) {
```
droazen: Do you still want to keep these given the new tests? Do they have independent value?

src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
akiezun: done

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
akiezun: done

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
akiezun: correct

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
akiezun: done

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
akiezun: done

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementUnitTest.java
akiezun: i think they are fine to keep, very little cost to them

PR #1796, opened 2016-05-01
faster GATKRead.numCigarElements - avoid object allocation for unmodifable list. Speeds up HC runs.

akiezun:


src/main/java/org/broadinstitute/hellbender/utils/pileup/PileupElement.java
```diff
@@ -51,7 +51,7 @@ public PileupElement(final GATKRead read,
         Utils.nonNull(read, "read is null");
         Utils.nonNull(currentElement, "currentElement is null");
         Utils.validIndex(baseOffset, read.getLength());
-        Utils.validIndex(currentCigarOffset, read.getCigarElements().size());
+        Utils.validIndex(currentCigarOffset, read.numCigarElements());
```
akiezun: Note: this is the expensive call site

lbergelson:
👍

lbergelson:
@akiezun I gave it a thumbs up, but then I just realized it could probably be improved more.

lbergelson:
SamRecord includes a native `getCigarLength` method which says it may be faster in some situations.

lbergelson:
The implementation should probably call that instead of what it does now.

akiezun:
good point - i'll experiment. https://github.com/broadinstitute/gatk/issues/1797

PR #1803, opened 2016-05-03
Miscellaneous test fixes for CompareBaseQualities and CollectIinsertSizeMetrics.

cmnbroad:
The CollectInsertSizeMetrics test cram file used to be zero-length, and was recently regenerated (most likely with Picard since it has BAM contents) but either way its STILL not a CRAM file. Also removed a stray junit dependency that crept in to CompareBaseQualities.


akiezun:
💯 merge when tests pass

akiezun:
i guess they just passed. boom

PR #1804, opened 2016-05-03
Fix Jenkins blowup caused by hacking tests involving external programs

SHuang-Broad:
@droazen , would you please review, before I go out and get enough beer (I hope I won't have to buy another round with this fix)?

Thanks!

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/tools/spark/sv/RunBWAMEMViaCommandLineTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/spark/sv/RunBWAMEMViaCommandLineTest.java
@@ -7,6 +7,7 @@ import org.broadinstitute.hellbender.utils.test.ArgumentsBuilder;
 import org.broadinstitute.hellbender.utils.test.BaseTest;
 import org.testng.Assert;
 import org.testng.SkipException;
+import org.testng.annotations.BeforeMethod;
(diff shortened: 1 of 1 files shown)


src/test/java/org/broadinstitute/hellbender/tools/spark/sv/RunBWAMEMViaCommandLineTest.java
```diff
@@ -28,24 +29,18 @@
...
-            s = getNonNullEnvironmentVariable("PATHTOBWA");
-            if(null==s) {
-                return null;
-            }else{
-                return Paths.get(s);
-            }
-        }else{
-            return Paths.get(s);
-        }
+        return s == null ? (null==(s=System.getenv("PATHTOBWA")) ? null : Paths.get(s)) : Paths.get(s);
```
droazen: This is a little TOO compact! (hard to read). Can you move the `System.getenv("PATHTOBWA")` call back into a separate if statement before the `? :` line?

droazen:
Back to @SHuang-Broad for a quick readability fix.

SHuang-Broad:
As recommended, @droazen . Please take a look again. Thanks!

droazen:
:+1: looks good, merging -- brown drink waived because environment variables are just not fair :)

PR #1813, opened 2016-05-11
Enable multi-threaded AVX PairHMM

gspowley:
Changes to enable multi-threaded native AVX PairHMM using OpenMP. Also includes a performance improvement in the native C++ `Context` class.

`VectorLoglessPairHMM.java` is hardcoded to set the maximum number of PairHMM threads (`maxNumberOfThreads`) to 100. This is the maximum number of threads **allowed** by GATK, not the number of threads **requested**. C code in the native library will query OpenMP for the number of threads available on the platform, and use min(OpenMP threads available, `maxNumberOfThreads`) threads.

**Measured Speedup**
Command
```
./gatk-launch HaplotypeCaller -R src/test/resources/large/human_g1k_v37.20.21.fasta -I src/test/resources/large/CEUTrio.HiSeq.WGS.b37.NA12878.20.21.bam -O out.g.vcf -ERC GVCF
```
1 thread
INFO  PairHMM - Total compute time in PairHMM computeLogLikelihoods() : 36.882098080000006
2 threads
INFO  PairHMM - Total compute time in PairHMM computeLogLikelihoods() : 18.160468659000003
3 threads
INFO  PairHMM - Total compute time in PairHMM computeLogLikelihoods() : 12.541517043
4 threads
INFO  PairHMM - Total compute time in PairHMM computeLogLikelihoods() : 9.727374342000001

**Potential issues**
* The target platform running GATK must have OpenMP installed
* The code has not been tested on Mac

**Todo**
* New Java code to allow the user to specify `maxNumberOfThreads` variable in `VectorLoglessPairHMM.java`.
* Move `maxNumberOfThreads` to the native `initialize` function, once we migrate to the new native library.

### DIFF
--- a/src/main/cpp/VectorLoglessPairHMM/org_broadinstitute_hellbender_utils_pairhmm_VectorLoglessPairHMM.cc
+++ b/src/main/cpp/VectorLoglessPairHMM/org_broadinstitute_hellbender_utils_pairhmm_VectorLoglessPairHMM.cc
@@ -1,3 +1,6 @@
+#ifdef linux
+#include <omp.h>
+#endif
 #include "headers.h"
 #include "jni_common.h"
 #include "org_broadinstitute_hellbender_utils_pairhmm_VectorLoglessPairHMM.h"
@@ -183,7 +186,11 @@ inline JNIEXPORT void JNICALL Java_org_broadinstitute_hellbender_utils_pairhmm_V
 inline void compute_testcases(vector<testcase>& tc_array, unsigned numTestCases, double* likelihoodDoubleArray,
     unsigned maxNumThreadsToUse)
 {
-  #pragma omp parallel for schedule (dynamic,10000) num_threads(maxNumThreadsToUse)
+  // TODO: move number of thread calculation to "initialize" after moving to native library interface
+#ifdef linux
+  int threads = min((int)maxNumThreadsToUse, omp_get_max_threads());
+#endif
+  #pragma omp parallel for schedule(dynamic, 1) num_threads(threads)
   for(unsigned tc_idx=0;tc_idx<numTestCases;++tc_idx)
   {
     float result_avxf = use_double ? 0 : g_compute_full_prob_float(&(tc_array[tc_idx]), 0);
--- a/build.gradle
+++ b/build.gradle
@@ -460,6 +460,8 @@ model {
                         cppCompiler.args "-I", "${System.properties['java.home']}/../include/Darwin"
                     } else {
                         cppCompiler.args "-I", "${System.properties['java.home']}/../include/linux"
+                        cppCompiler.args "-fopenmp"
+                        linker.args "-fopenmp"
                         linker.args "-static-libgcc"
                     }
 
--- a/src/main/cpp/VectorLoglessPairHMM/common_data_structure.h
+++ b/src/main/cpp/VectorLoglessPairHMM/common_data_structure.h
@@ -29,10 +29,10 @@ template<class NUMBER>
 struct ContextBase
 {
   public:
-    NUMBER ph2pr[128];
-    NUMBER INITIAL_CONSTANT;
-    NUMBER LOG10_INITIAL_CONSTANT;
-    NUMBER RESULT_THRESHOLD; 
+    static NUMBER ph2pr[128];
+    static NUMBER INITIAL_CONSTANT;
+    static NUMBER LOG10_INITIAL_CONSTANT;
+    static NUMBER RESULT_THRESHOLD; 
 
     static bool staticMembersInitializedFlag;
     static NUMBER jacobianLogTable[JACOBIAN_LOG_TABLE_SIZE];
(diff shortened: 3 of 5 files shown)


akiezun:
cool stuf! i can take the first look and take it for a spin

gspowley:
great, thanks!

akiezun:
@gspowley can you clarify if openMP is needed to compile this code or is it also needed to run it?

gspowley:
Yes, OpenMP is needed to compile **and** run this code. The `libVectorLoglessPairHMM` shared library depends on an OpenMP shared library.

```
> ldd ./build/classes/main/lib/libVectorLoglessPairHMM.so
	linux-vdso.so.1 =>  (0x00007ffc9b2b3000)
	libstdc++.so.6 => /opt/gcc-4.8.2/lib64/libstdc++.so.6 (0x00007fe67bacf000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fe67b831000)
--> libgomp.so.1 => /opt/gcc-4.8.2/lib64/libgomp.so.1 (0x00007fe67b623000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fe67b405000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fe67b071000)
	/lib64/ld-linux-x86-64.so.2 (0x000000378cc00000)
	libgcc_s.so.1 => /opt/gcc-4.8.2/lib64/libgcc_s.so.1 (0x00007fe67ae5b000)
	librt.so.1 => /lib64/librt.so.1 (0x00007fe67ac52000)
```

I'll look into statically linking OpenMP as an option. Another option would be a clean room source code implementation of the thread pool.

gspowley:
@akiezun BTW, if you want to experiment with different numbers of threads in the current code, you can use the `OMP_THREAD_LIMIT` environment variable to set the maximum number of threads.

For example, this command would run with 2 threads:
```
OMP_THREAD_LIMIT=2 ./gatk-launch HaplotypeCaller ...
```

akiezun:
thanks @gspowley - it works very nicely on linux. We tried and failed on the Mac OSX 10.9 and 10.10. It seems OpenMP is just not available in xcode. We'll have to disable it on the Mac for now.

akiezun:
I like it and want to put it in but we can't unless the code compiles and runs on the Mac. I'd be fine with having Mac use just 1 thread if need be but the code must compile and run. With some hacking @lbergelson was able to compile and run it but it's not something we want users to do. The official XCode on the Mac does not include OpenMP and having our users go through a complex install of an alternative clang is a no-starter for us. Also non-starter is removal of native PairHMM from the Mac. 

So I think we have 2 options:
a) change the code+build to disable OMP on the Mac (ie have Mac be single threaded). The benefit is that we can start running with improved performance right away on linux. That includes production (when we're ready with the HaplotypeCaller)
b) wait until the new intel gatk native library is done and then move the problem of compiling etc to that repo. That would remove the burden of building this from us and shift to the other repo's owners who can then build their library however they want and we'd only pick up compiled versions.

@gspowley @lbergelson @droazen votes?

akiezun:
I think OpenMP is common on Linux so dynamic linking is probably fine (i definitely did not have to do any setup and it just worked). We'll worry about the Mac linking (static/dynamic) once Apple get their act together and updates their XCode to modern clang with openMP.

gspowley:
I can supply code changes to support option (a) above, and then we can work on supporting Mac in the native library.

akiezun:
excellent! thanks @gspowley - looking forward to it.

lbergelson:
I think until we have the intel repository separated out, we should make sure it builds/runs without additional compilers on osx.   I would prefer if we run the same codepath on both osx and linux, because it increases our testing burden and makes the surface for confusing cross platform errors higher.  

If it's important that we include this multithreading before we can separate the repos, then we just need to be sure that we test both osx and linux.  Maybe we should prioritize the lower level pair hmm unit tests so we can run just those on both instead of the complete haplotype caller tests.

akiezun:
thanks @lbergelson BTW can we set up travis on the Mac? If we support the Mac we should test it automatically. https://github.com/broadinstitute/gatk/issues/1816

gspowley:
The code difference between OSX and Linux is pretty small. 

Here's the change in the native code. 
* Linux will define the `threads` variable and use it in the OpenMP pragma.
* OSX will not see the `threads` declaration and the OpenMP pragma will be a comment.

``` C
#ifdef linux
  int threads = min((int)maxNumThreadsToUse, omp_get_max_threads());
#endif
  #pragma omp parallel for schedule(dynamic, 1) num_threads(threads)
```

We also need a minor change in `build.gradle`. I'll submit the changes to this PR and you can decide from there.

lbergelson:
@akiezun We can set up travis on mac.  It will be a bit of a hassle. Do we want to run our whole test suit on mac? Or do we only want to test the pairhmm since ostensibly the rest of it "just works" thanks to the magic of java?

src/main/cpp/VectorLoglessPairHMM/org_broadinstitute_hellbender_utils_pairhmm_VectorLoglessPairHMM.cc
```diff
@@ -1,3 +1,4 @@
+#include <omp.h>
```
akiezun: this include was a problem on the mac too

gspowley:
Sorry I missed that. We're getting a Mac here soon, will catch those in the future.

akiezun:
@gspowley please assign back to me when ready

gspowley:
@akiezun it should be ready now, added `#ifdef linux` around the `#include <omp.h>` last night.

akiezun:
thanks

akiezun:
compiles and runs on the mac now. thanks. 
👍

Issue #1797, opened 2016-05-02
experiment with getCigarLength instead of numCigarElements in GATK read.


akiezun:
experiment with getCigarLength instead of numCigarElements in GATK read. May be faster

came up in #1796

akiezun:
fixed by #1818

PR #1818, opened 2016-05-13
use getCigarLength() rather than getCigar().numCigarElements() 

akiezun:
addresses #1797 
some small speedup and less object allocation when using BAMs

@lbergelson please review

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/read/SAMRecordToGATKReadAdapter.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/read/SAMRecordToGATKReadAdapter.java
@@ -276,7 +276,7 @@ public class SAMRecordToGATKReadAdapter implements GATKRead, Serializable {
      */
     @Override
     public int numCigarElements(){
-        return samRecord.getCigar() == null ? 0 : samRecord.getCigar().numCigarElements();
+        return samRecord.getCigar() == null ? 0 : samRecord.getCigarLength();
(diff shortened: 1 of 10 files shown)


akiezun:
hmm

src/main/java/org/broadinstitute/hellbender/utils/read/SAMRecordToGATKReadAdapter.java
```diff
@@ -276,7 +276,7 @@ public Cigar getCigar() {
      */
     @Override
     public int numCigarElements(){
-        return samRecord.getCigar() == null ? 0 : samRecord.getCigar().numCigarElements();
+        return samRecord.getCigar() == null ? 0 : samRecord.getCigarLength();
```
lbergelson: No need for the ternery here.  samRecord.getCigarLength() already handles the null cigar case.

lbergelson: Actually, maybe that's not true.. let me look again.

lbergelson: Sorry, you were right, it looks like samRecord.getCigarLength can null ptr if there is a null cigar since it just call `getCigar` internally.

lbergelson:
Looks good to me.

Issue #1056, opened 2015-10-27
Rip out the indel recalibration code from BaseRecalibrationEngine


droazen:
After doing this, do another comparison run against GATK3 BQSR to see how much eliminating this code bought us in terms of performance.

akiezun:
To clarify, this means that the `disable_indel_quals` parameter will always be `true`, right @eitanbanks ?

eitanbanks:
That's correct, @akiezun.
However, it's not just stripping out that code.  There are a ton of optimizations that can then be made to the code to simplify it afterwards.  These classes were made very bulky to accommodate the indel calibration, and we should really remove that bulk.  I can help with that.

akiezun:
agreed, just wanted to know the top level starting point.

akiezun:
initial run on 1.2Gb file after ripping out some of the indel code:
CEUTrio.HiSeq.WEx.b37.NA12892.chr10.bam

Mac OS X 10.9.5 x86_64; Java HotSpot(TM) 64-Bit Server VM 1.8.0_25-b17;
4.pre-alpha-71-g50b094b

```
real	4m58.795s
user	5m3.401s
```

this is 2x speedup compared to regular gatk4 and gatk3 (data from see #1033) 

time gatk3 with indels
```
real	9m7.691s
user	9m2.250s
sys	0m22.603s
```

time for GATK3
```
real	11m21.538s
user	17m24.320s
sys	0m8.073s
```

eitanbanks:
@akiezun it's really important to make sure that the results match between the original and newly optimized versions (and if they don't, to make sure we understand why).  Some of the "indel stuff" should probably stay, e.g. the masking of sites using known indels.  Since we aren't running Indel Realigner anymore, there may be some "errors" that we do want to mask because they are alignment artifacts and not sequencing errors.
Pinging @yfarjoun for his thoughts.  We can discuss this at a future methods meeting if you like.

akiezun:
for now this is just the apply step.

vdauwera:
Hey folks, based on this effort, at what point do we start telling users to disable indel quals (or do it for them by default) in the output of BQSR in GATK3? People complain about the file sizes and this would alleviate some of that pain. I know that that's how it's done in production.

akiezun:
btw, i can probably shave off another 10-15% of time off this thing. stay tuned

akiezun:
a little bit shaved off. working on more
real	4m40.226s
user	4m42.279s

[...]

PR #1821, opened 2016-05-14
no indel tables in BQSR tables by default

akiezun:
I want to skip indels by default in BQSR. We know if does not affect results but it's faster and we want speed out of the box.

@lbergelson can you review?


lbergelson:
@akiezun Are there any tests that run now without indelBQSR on?  
When you say doesn't affect results you mean byte identical or analysis shows it's unimportant?  I'm assuming some minor differences but not meaningful.

akiezun:
yes. BothStepsOfBQSRIntegrationTest - the results are qual-for-qual identical  (see https://github.com/broadinstitute/gatk/issues/1056)

lbergelson:
:+

PR #1826, opened 2016-05-17
Changed column names in CalculateTargetBaseCallCoverage

davidbenjamin:
This is to reconcile with formats output by CNV tools in gatk-protected.  CONTIG, START, END, NAME has become contig, start, stop, name.  That's the entire PR.  @lbergelson could you review and make a snapshot once merged?  It should be quick.


lbergelson:
👍

Issue #1643, opened 2016-03-28
Port additional metrics to Spark


droazen:
For this quarter, let's port some additional Picard metrics that are used by production.

droazen:
(This ticket should spawn additional sub-tickets)

droazen:
First task here is to examine the pipeline in https://github.com/broadinstitute/dsde-pipelines/blob/master/genomes_in_the_cloud/single_sample/PairedSingleSampleWf.wdl and determine which metrics used in production still need to be ported to Spark by us.

cmnbroad:
WIP commit [here](https://github.com/broadinstitute/gatk/commit/45bac26ca158546afb3d36fd33195bce9f4f193d) with a proposed approach to factoring Spark metrics collectors so they work both as a standalone tool or on behalf of CollectMultipleMetricsSpark, which reuses a single RDD across multiple metrics tools.

cmnbroad:
@droazen If you get a chance take a look at the branch linked in the comment above (not for code review, just discussion of the approach), and/or we can discuss when I'm in later this week.

droazen:
Adding @jamesemery as a co-assignee along with @cmnbroad. @jamesemery feel free to port one or more metrics to spark once @cmnbroad's framework is fully merged, if you have the time/inclination or need a break from your other spark tickets.

droazen:
This was done

Issue #1237, opened 2015-12-03
enhance ReadsPipelineSpark with metrics


akiezun:
add some metrics to the pipeline to see how well that scales. Metrics should come cheaply because the data RDD is loaded at that point.

droazen:
Task is to hook up metrics to the pipeline and benchmark them after they've been ported. https://github.com/broadinstitute/gatk/issues/1643 captures the actual porting work.

PR #1827, opened 2016-05-17
Spark metrics collector refactoring checkpoint.

cmnbroad:
This is a checkpoint PR for https://github.com/broadinstitute/gatk/issues/1237 and https://github.com/broadinstitute/gatk/issues/1643. This is the first step in refactoring metrics collectors so they can be pipelined in Spark and reuse RDDs, but still share metrics computation code between walker and Spark versions. The next step will be to extend MultilevelCollector to be able to merge its own instances in order to support efficient map and reduce phases for multi level collectors.

Suggested review order:

-MetricsCollectorSpark: interface to be implemented by all Spark collectors
-MetricsArgs:base class for all collector argument sets
-MetricsCollectorToolSpark: base class for all Spark metrics collector tools
-CollectQualityYieldMetrics: Spark version of QualityYieldMetrics using these new interfaces
-CollectInsertSizeMetricsSpark: existing Spark version of InsertSizeMetrics collector ported
to these interfaces
-CollectMultipleMetricsSpark: Spark version of CollectMultipleMetrics; currently only works
on QualityYieldMetrics and InsertSizeMetrics

The rest of the PR is refactoring existing to get QualityYieldMetrics and InsertSizeMetrics to conform to these interfaces (moving CollectInsertSizeMetrics out of the sv package and Program Groups, etc.). Note that the existing InsertSizeMetrics Spark collector doesn’t really share code with the walker
version (and their command line param sets are way out of sync) but this should be fixed separately from these changes as the interfaces evolve.


src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSparkIntegrationTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        validateInsertSizeMetrics(outBase);
+        validateQualityYieldMetrics(outBase);
+    }
+
+    private void validateQualityYieldMetrics(final String outBase) throws IOException {
+        String localOutBase = outBase + ".qualityYieldMetrics" + ".txt";
+        final MetricsFile<InsertSizeMetrics, Comparable<?>> output = new MetricsFile<>();
+
+        try (FileReader reader = new FileReader(localOutBase)) {
```
sputnikci: \[FindBugs\] WARNING: Dm: Found reliance on default encoding in org\.broadinstitute\.hellbender\.tools\.spark\.pipelines\.metrics\.CollectMultipleMetricsSparkIntegrationTest\.validateQualityYieldMetrics\(String\): new java\.io\.FileReader\(String\)

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSparkIntegrationTest.java
```diff
@@ -0,0 +1,131 @@
...
+            output.read(reader);
+            Assert.assertEquals(output.getMetrics().size(), 1);
+        }
+    }
+
+    private void validateInsertSizeMetrics(final String outBase) throws IOException {
+        String localOutBase = outBase + ".insertMetrics" + ".txt";
+        final MetricsFile<InsertSizeMetrics, Comparable<?>> output = new MetricsFile<>();
+
+        try (FileReader reader = new FileReader(localOutBase)) {
```
sputnikci: \[FindBugs\] WARNING: Dm: Found reliance on default encoding in org\.broadinstitute\.hellbender\.tools\.spark\.pipelines\.metrics\.CollectMultipleMetricsSparkIntegrationTest\.validateInsertSizeMetrics\(String\): new java\.io\.FileReader\(String\)

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSparkIntegrationTest.java
sputnikci: \[FindBugs\] WARNING: Dm: Found reliance on default encoding in org\.broadinstitute\.hellbender\.tools\.spark\.pipelines\.metrics\.CollectMultipleMetricsSparkIntegrationTest\.validateQualityYieldMetrics\(String\): new java\.io\.FileReader\(String\)

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSparkIntegrationTest.java
sputnikci: \[FindBugs\] WARNING: Dm: Found reliance on default encoding in org\.broadinstitute\.hellbender\.tools\.spark\.pipelines\.metrics\.CollectMultipleMetricsSparkIntegrationTest\.validateInsertSizeMetrics\(String\): new java\.io\.FileReader\(String\)

droazen:
@akiezun Can you review this one?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -0,0 +1,116 @@
...
+ * the collector in a form that is suitable as a command line argument
+ * collection.
+ * A class that implements this interface, parameterized with the
+ * MetricsArgs-derived class.
+ * </li>
+ *
+ * These two classes can then be used either in the standalone collector tool (see
+ * CollectInsertSizeMetricsSpark as an example), or from CollectMultipleMetricsSpark.
+ *
+ * The general lifecyle of a Spark collector looks like this:
```
akiezun: typo `lifecyle`

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -0,0 +1,116 @@
...
+ *         getHeaderForReads(),
+ *         getReadSourceName(),
+ *         getAuthHolder()
+ *     );
+ *     collector.finishCollection();
+ *
+ */
+public interface MetricsCollectorSpark<T extends MetricsArgs> extends Serializable
+{
+    static final long serialVersionUID = 1l;
```
akiezun: make public, change to `1L`

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Advertise whether or not this collector requires a reference.
```
akiezun: maybe `Advertise` -> `Return`

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -0,0 +1,116 @@
...
+     * @return true if this collector requires that a reference be provided.
+     */
+    default boolean requiresReference() { return false;}
```
akiezun: describe what the default implementation does and what is expected from implementing classes

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -0,0 +1,116 @@
...
+     * at the command line by the user of a standalone metric collector tool.
+     *
+     * This method is primarily for use in a standalone collector tool in order
+     * transfer the command line argument values for that tool. However, it may
+     * never be called if the collector is being used by a pipeline driver such
+     * CollectMultipleMetrics, since the arguments will be provided directly to the
+     * collector's initializeCollector method by the driver.
+     * @return object of type T
+     */
+    default T gatherInputArguments() { return null; };
```
akiezun: describe what the default implementation does and what is expected from implementing classes

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -0,0 +1,116 @@
...
+     * @param defaultHeaders default metrics headers from the containing tool
+     */
+    void initializeCollector(T inputArgs, List<Header> defaultHeaders);
+
+    /**
+     * Return the read filter required for this collector.
+     * @param samHeader the SAMFileHeader for the input BAM
+     * @return the read filter required for this collector
+     */
+    default ReadFilter getCollectorReadFilter(SAMFileHeader samHeader) {
```
akiezun: describe what the default implementation does and what is expected from implementing classes

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -0,0 +1,116 @@
...
+     * @param authHolder authHolder
+     */
+    void collectMetrics(
+            JavaRDD<GATKRead> filteredReads,
+            SAMFileHeader samHeader,
+            String inputBaseName,
+            AuthHolder authHolder
+    );
+
+    default void finishCollection() { }
```
akiezun: add doc. describe what the default implementation does and what is expected from implementing classes

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
akiezun: why not `getReadFilter` ?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -0,0 +1,116 @@
...
+     * is being driven by a standalone tool, these arguments will have been returned
+     * by gatherInputArguments; in the case of a pipeline driver they will be provided
+     * directly by the driver. This method will always be called before either
+     * getReadFilter (so the collector can use the arguments to initialize the
+     * readFilter) or collectMetrics.
+     * @param inputArgs an object that contains the argument values to be used for
+     *                  the collector
+     * @param defaultHeaders default metrics headers from the containing tool
+     */
+    void initializeCollector(T inputArgs, List<Header> defaultHeaders);
```
akiezun: why not `initialize`?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -0,0 +1,116 @@
...
+     * Give the collector a chance to gather and return any input argument values
+     * specific to this collector type, i.e, the argument values specified
+     * at the command line by the user of a standalone metric collector tool.
+     *
+     * This method is primarily for use in a standalone collector tool in order
+     * transfer the command line argument values for that tool. However, it may
+     * never be called if the collector is being used by a pipeline driver such
+     * CollectMultipleMetrics, since the arguments will be provided directly to the
+     * collector's initializeCollector method by the driver.
+     * @return object of type T
```
akiezun: improve doc - what is the returned object?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -0,0 +1,116 @@
...
+ *     collector.initializeCollector(args);
+ *
+ *     ReadFilter filter == collector.getReadFilter();
+ *     collector.collectMetrics(
+ *         getReads().filter(filter),
+ *         getHeaderForReads(),
+ *         getReadSourceName(),
+ *         getAuthHolder()
+ *     );
+ *     collector.finishCollection();
```
akiezun: how does one retrieve the collected metrics from the collector?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.spark.api.java.JavaSparkContext;
+import org.broadinstitute.hellbender.engine.filters.ReadFilter;
+import org.broadinstitute.hellbender.engine.spark.GATKSparkTool;
+import org.broadinstitute.hellbender.exceptions.GATKException;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.metrics.MetricsArgs;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+/**
+ * Base class for standalone Spark metrics collector tools.ß
```
akiezun: bizarre character at the end of line 14

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
```diff
@@ -0,0 +1,94 @@
...
+ *
+ */
+public abstract class MetricsCollectorToolSpark<T extends MetricsArgs>
```
akiezun: I think it's weird that the tool implements `MetricsCollectorSpark`

akiezun: improve the name, maybe revert Spark and Tool, ie `MetricsCollectorSparkTool` to mimic `GATKSparkTool`

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
```diff
@@ -0,0 +1,94 @@
...
+        validateInputState();
+        validateExpectedSortOrder(SAMFileHeader.SortOrder.coordinate, false);
+
+        //  Execute the collector lifecycle
+        T args = gatherInputArguments();
+        if (null == args) {
+            // Null args indicates an incorrectly written collector. Metrics collector arg objects
+            // are always derived from MetricsArgs, so they always have at least an output argument
+            // and should not be null.
+            throw new GATKException.ShouldNeverReachHereException
```
akiezun: this should be an `IllegalStateException`

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
```diff
@@ -0,0 +1,94 @@
...
+        //TODO: temporary references are implemented
+        if (requiresReference()) {
+            throw new UnsupportedOperationException("Requires reference for collector not yet implemented");
+        }
+    }
+
+    /**
+     * Validate the input sort order
+     */
+    protected void validateExpectedSortOrder(
```
akiezun: document what subclasses need to do - call super or not

src/main/java/org/broadinstitute/hellbender/metrics/QualityYieldMetricsArgs.java
```diff
@@ -0,0 +1,18 @@
+package org.broadinstitute.hellbender.metrics;
+
+import org.broadinstitute.hellbender.cmdline.Argument;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+
+/**
+ * MetricsArgs argument collection for QualityYield metrics. All members should be
+ * instantiable as command line arguments.
+ */
+public class QualityYieldMetricsArgs extends MetricsArgs {
```
akiezun: elsewhere in the GATK we call such things 'ArgumentCollection`, not `Args`

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
```diff
@@ -0,0 +1,181 @@
...
+         * type-safe, collector-specific primer that creates and populates an
+         * instance of the collector's input arguments class; creates an instance
+         * of the collector; initializes it with the arguments object; and then
+         * forwards the initialized collector to the executeLifeCycle function
+         */
+        void primeCollector(
+                final String outputBaseName,
+                final Set<MetricAccumulationLevel> metricAccumulationLevel,
+                final List<Header> defaultHeaders,
+                final Consumer<MetricsCollectorSpark<?>> executeCollectorLifecyle
```
akiezun: why not return the collector from the call rather than pass to this consumer instance?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: The result of the metrics collection is an RDD that will usually be reduced to a single object whos' type is specific to the collector (i.e. a QualityYieldMetrics object, or a MultiLevelCollector/derivative) that is serialized out to a file by the collector when the finish method is called. (I think the two implementations in this PR currently do that as part of collectMetrics). We could change this to have the result returned to the caller/driver such as CollectMultipleMetricsSpark, but without any common type ancestor I'm not sure what it could do with it other than pass it back to finish for export. Maybe thats better just because its more explicit. I'm open to suggestions.

akiezun:
I like it. I have small naming-related comments. It's complex stuff and it's always going to be hard to keep spark vs non-spark in synch. @droazen do you want to review this too?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
cmnbroad: Yeah that makes more sense - the continuation is unnecessary.

droazen:
@akiezun I've given feedback on it in person. @lbergelson may want to have a glance over it, though.

akiezun:
assigning to @lbergelson for comments

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
```diff
@@ -0,0 +1,181 @@
...
+public final class CollectMultipleMetricsSpark extends GATKSparkTool {
+
+    private static final long serialVersionUID = 1L;
+
+    @Argument(fullName=StandardArgumentDefinitions.ASSUME_SORTED_LONG_NAME,
+                shortName = StandardArgumentDefinitions.ASSUME_SORTED_SHORT_NAME,
+                doc = "If true (default), then the sort order in the header file will be ignored.")
+    public boolean ASSUME_SORTED = true;
+
+    //TODO: do we need this for Spark ?
```
lbergelson: I would say we do not need this for spark.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
lbergelson: I also think it's a bit strange that this implements MetricsCollectorSpark.  Could it be refactored to just contain a `MetricsCollectorSpark` instead of being one?  Is there a use case where you would want to use the `MetricsCollectorToolSpark` directly as a metrics collector?

lbergelson: There's also the strange intersection of the requiresReference() method from `GATKTool`and `MetricsCollectorSpark` due to this being both.  It seems like they should be compatible though in this case.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
```diff
@@ -0,0 +1,94 @@
...
+                filteredReads,
+                getHeaderForReads(),
+                getReadSourceName(),
+                getAuthHolder()
+        );
+        finishCollection();
+    }
+
+    private void validateInputState() {
+        //TODO: temporary references are implemented
```
lbergelson: temporary *until* references are implemented

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
lbergelson: Is this actually meant to be overriden by subclasses?  I assume it should either be made final or private.  It's duplicated in CollectMultipleMetricsSpark as well, maybe it should be made static and put somewhere public as a utility?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
cmnbroad: The reason for having this tool (which is really a base class that defines a special kind of a tool) declare that it implements this interface is not to allow them to be called from the outside; its that the interface is what defines the requirements of a subclass. If you look at the runTool method below, it needs to be able to call the implementations of those interface methods in order to put the collector through it's lifecycle. I don't think its any different than saying, for example, that a ReadWalker has to implement an "apply" method; nobody outside of the tool calls it; but the base ReadWalker tool depends on subclasses implementing it (and in this case the interface is used in several other places). If I remove it, I have to implement another layer of delegation that I think would add complexity but no real value. Having said that, let me know if you still think its the right thing to do.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
```diff
@@ -0,0 +1,181 @@
...
+                doc = "Base name of output files.")
+    public String outputBaseName;
+
+    @Argument(shortName="LEVEL", doc="The level(s) at which to accumulate metrics.  ")
+    public Set<MetricAccumulationLevel> metricAccumulationLevel = EnumSet.of(MetricAccumulationLevel.ALL_READS);
+
+    @Argument(doc = "List of metrics collectors to apply during the pass through the SAM file.")
+    public List<Collectors> collectors = new ArrayList<>();
+
+    private interface CollectorPrimer {
```
lbergelson: I'm not sure `Primer` is the clearest name.  Possibly `Provider` might be more idiomatic?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
cmnbroad: It is intended to allow subclasses to override the default behavior of requiring coord sorted input. So I think it should remain protected, but the expectedOrder should be removed and the default implementation should just require coord sorted. Or now that I think more about it, it should probably delegate to the collector itself, so it can decide if the input order is sufficient.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
cmnbroad: I wanted something to indicate the main thing these do, which is to populate the collector with its input arguments. Sounds like primer doesn't do that; provider is, as you say, probably more idiomatic

src/main/java/org/broadinstitute/hellbender/metrics/QualityYieldMetrics.java
```diff
@@ -0,0 +1,143 @@
...
+        Q30_BASES    += that.Q30_BASES;
+        PF_Q30_BASES += that.PF_Q30_BASES;
+        Q20_EQUIVALENT_YIELD += that.Q20_EQUIVALENT_YIELD;
+        PF_Q20_EQUIVALENT_YIELD += that.PF_Q20_EQUIVALENT_YIELD;
+
+        return this;
+    }
+
+    //completes the calculations
+    public QualityYieldMetrics finish() {
```
lbergelson: It feels like these merge and finish methods should maybe be extracted to an interface.  Are all metrics classes going to have them?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
```diff
@@ -0,0 +1,181 @@
...
+        void primeCollector(
+                final String outputBaseName,
+                final Set<MetricAccumulationLevel> metricAccumulationLevel,
+                final List<Header> defaultHeaders,
+                final Consumer<MetricsCollectorSpark<?>> executeCollectorLifecyle
+        );
+    }
+
+    // Enum of Collector types supported by this tool.
+    public static enum Collectors implements CollectorPrimer {
```
lbergelson: I'm not a big fan of this hardcoded set of metrics.  

It seems like this is mostly just setting the metrics collectors to use what could be default values.  I.e. `InsertSizeMetricsCollectorSpark` should probably know to use '.insertMetrics' as it's default extension rather than having to define that here.  Maybe `MetricArgs` could provide a function to inject a base filepath?

Could `isArgs.useEnd` just default to SECOND?  

Maybe custom configuration of some collectors is necessary, but it seems like it could be avoided in these cases with some fairly limited extensions of the abstractions.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
```diff
@@ -0,0 +1,181 @@
...
+                    final String outputBaseName,
+                    final Set<MetricAccumulationLevel> metricAccumulationLevel,
+                    final List<Header> defaultHeaders,
+                    final Consumer<MetricsCollectorSpark<?>> executeCollectorLifecyle)
+            {
+                // disambiguate this collector's output files from the other collectors
+                final String localBaseName = outputBaseName + ".qualityYieldMetrics";
+
+                final QualityYieldMetricsArgs isArgs = new QualityYieldMetricsArgs();
+                isArgs.output = new File(localBaseName + ".txt");
```
lbergelson: this shouldn't be called isArgs

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
lbergelson: It just seems a bit strange and confusing that the class implementing this tool calss itself can be used as a MetricsCollector since it essentially just wraps one.  I don't think I would have understood how to implement one without seeing your examples.  

I'd like it if there was a way to have this operate on a provided MetricsCollector but it is a bit awkward to refactor it in a way that forces the subclasses to provide it with one.  

If you think this is the best way then keep it how it is.

src/main/java/org/broadinstitute/hellbender/metrics/QualityYieldMetrics.java
cmnbroad: Right - that will be part of phase II, when we'll have more than reducible collector.

src/main/java/org/broadinstitute/hellbender/metrics/QualityYieldMetricsArgs.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Done, both.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
cmnbroad: Ok, I've removed the "implements MetricsCollectorSpark", at the expense of some duplication of the interface method declarations in the base tool class. We could probably make the whole issue go away by introducing another type to serve as a common ancestor/tag interface, but there is already a fair amount of type complexity so for now we'll go with this simplified version

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorToolSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
cmnbroad: Removed.

droazen:
@cmnbroad merge when ready

PR #1828, opened 2016-05-17
upload snapshots to artifactory on every commit

lbergelson:
by a quirk of travis this includes pull request builds, so if you need a snapshot you can open a pull request and the head of it will be build into a snapshot on artifactory

the version identifier should will appear in the output of the "after_sucess" block on travis
this will only package the linux version of the pairHmm library, to package both versions you must do a manual snapshot

also upgrades gradle to 2.13 to use the new findProperty function

DON'T MERGE UNTIL THE NEW ARTIFACT IS CONFIRMED TO BE IN ARTIFACTORY FOR THIS PR


build.gradle
```diff
@@ -178,24 +177,9 @@ processResources {
...
 
-def String deriveVersion(){
-    def stdout = new ByteArrayOutputStream()
-    try {
-        logger.info("path is $System.env.PATH")
-        exec {
-            commandLine "git", "describe", "--always"
-            standardOutput = stdout;
+final isRelease = Boolean.getBoolean("release")
+version = isRelease ? gitVersion() : gitVersion() + "-SNAPSHOT"
```
droazen: Will this have the "dirty" marker on it even when no tracked files have been changed?

lbergelson:
@droazen Yup... It's just a call to jgit isDirty().  Git defines having untracked stuff as being dirty.  We can remove it entirely if it's bothering you...  Changing it so it only checks if there are unmerged changes to tracked files is slightly tricker though.

droazen:
@lbergelson Ideal behavior is to mark only versions that change tracked files as dirty, but if that's not possible I think we should strip out the "dirty" for now. In my view it's misleading, since it implies that the snapshot contains uncommitted changes, which is not always the case.

lbergelson:
You're objecting to git's own definition of dirty.  I don't know if you have the moral high ground here!

droazen:
I think it's fair to say that most people interpret "dirty" (in a version control context!) to mean "contains uncommitted changes." Distributing jars with "dirty" in their name will cause people to think that if they checked out the commit in the jar name, they would not have all the changes -- which isn't true here. We should use the commonly-accepted definition rather than git's definition.

build.gradle
```diff
@@ -366,66 +350,60 @@ signing {
...
 
-                    scm {
-                        url 'scm:git@github.com:broadinstitute/hellbender.git'
-                        connection 'scm:git@github.com:broadinstitute/hellbender.git'
-                        developerConnection 'scm:git@github.com:broadinstitute/hellbender.git'
-                    }
+                scm {
+                    url 'scm:git@github.com:broadinstitute/hellbender.git'
+                    connection 'scm:git@github.com:broadinstitute/hellbender.git'
+                    developerConnection 'scm:git@github.com:broadinstitute/hellbender.git'
```
droazen: If we're GATK4, and not hellbender, shouldn't these repo URLs be updated?

build.gradle
```diff
@@ -366,66 +350,60 @@ signing {
...
-                }
+            repository(url: "https://oss.sonatype.org/service/local/staging/deploy/maven2/") {
+                authentication(userName: project.findProperty("sonatypeUsername"), password: project.findProperty("sonatypePassword"))
+            }
 
-                snapshotRepository(url: "https://oss.sonatype.org/content/repositories/snapshots") {
-                    authentication(userName: sonatypeUsername, password: sonatypePassword)
-                }
+            snapshotRepository(url: "https://artifactory.broadinstitute.org/artifactory/libs-snapshot-local/") {
+                authentication(userName: System.env.ARTIFACTORY_USERNAME, password: System.env.ARTIFACTORY_PASSWORD)
```
droazen: Gross, but ok for now

build.gradle
```diff
@@ -366,66 +350,60 @@ signing {
...
                     }
                 }
             }
         }
-    } else {
-        doFirst({
-            logger.error( 'Users are not generally supposed to upload archives.  ' +
-                'To upload archives you must specify certain information in your gradle.properties. See the README' +
-                ' for more information.' )
-            throw new Exception("Missing sonatype username or password, please see the README for properties that must be specified for uploading an archive.")}
```
droazen: Why delete this helpful error message?

build.gradle
```diff
@@ -366,66 +350,60 @@ signing {
...
+
 
 
 task installSpark{ dependsOn sparkJar }
 task installAll{  dependsOn installSpark, installDist }
 
 test.dependsOn extractIntelDeflater
 installDist.dependsOn downloadGsaLibFile, extractIntelDeflater
-build.dependsOn installDist
-check.dependsOn installDist
```
droazen: Can you explain why these lines are no longer needed?

droazen:
Back to @lbergelson, merge after addressing comments

build.gradle
lbergelson: oh, heh, good point

build.gradle
lbergelson: I can change them now if you want... I guess there isn't really any reason to keep them this way.  I just didn't want to have to wait for travis again since it's getting late, but If I'm making other changes I can do these too.

droazen: Only if you have time...

build.gradle
lbergelson: The guard was necessary before since checking the properties would crash on every gradle run, even ones that didn't upload.  Now you'll just get a authentication error because of a null username.

build.gradle
lbergelson: They just didn't really do what you'd want.  There's no reason that check should depend on installDist.  We could make build do installAll, but that's what we have installAll for...

build.gradle
lbergelson: actually, I'm going to keep these, because if I change them I'll introduce a typo and I don't have time for 2 builds...

Issue #1819, opened 2016-05-13
Ensure that multi-threaded AVX PairHMM falls back to single-threaded mode when OpenMP is not available or is the wrong version


droazen:
What is the current behavior when OpenMP is not available or is the wrong version on a Linux system? If the answer is "it blows up with a gross error", then let's patch the code so that it gracefully falls back to the single-threaded version, as on a Mac.

droazen:
For @gspowley

gspowley:
Right, there's not a graceful fallback if OpenMP is not supported at this point. I'll start looking into it.

akiezun:
@droazen do you want this for compile time or run time? Both can fail.

droazen:
@akiezun I'm mainly concerned with having a graceful fallback at runtime, though it would also be good if the GATK could be made to compile on Linux when OpenMP is not present / is the wrong version. The latter is a temporary issue, however, pending the move to a separate repository for the native code.

droazen:
The OpenMP stuff is causing problems for us already when the correct versions of the libraries are not present. We are going to back it out in master until this ticket (adding a fallback when OpenMP can't be loaded on Linux) is done.

akiezun:
either fallback to 1 core or a (maybe preferably) statically linked version of OMP libraries so that people can run it on machines that do not have the same exact layout of dynamic OMP libraries as the machine on which the so file was made.

gspowley:
I'd prefer to fix this in the native library. Is that ok? We're focusing on bringing up the native library now.

akiezun:
yes that's fine

droazen:
@gspowley is this fixed now?

gspowley:
It's not fixed yet. The current release of GKL (0.1.1) only has single-threaded AVX PairHMM.

droazen:
Ok, moving to alpha-3 then

gspowley:
The first step of resolving this issue is merged into GKL. The approach is described in https://github.com/Intel-HLS/GKL/pull/25.

**Next steps**
- [ ] Resolve #1946 
- [ ] Resolve https://github.com/Intel-HLS/GKL/issues/26

PR #1829, opened 2016-05-17
Revert "Enable multi-threaded AVX PairHMM"

akiezun:
Reverts broadinstitute/gatk#1813 for now.

We have to revert this because there's no fallback to lack of OMP or incompatible libraries:  https://github.com/broadinstitute/gatk/issues/1819

What happens then is that we build the so file on a one machine and the protected repo uses a different machine and blows up with ```java.lang.UnsatisfiedLinkError: /tmp/libVectorLoglessPairHMM.4842381956752440752.so: /usr/lib/x86_64-linux-gnu/libgomp.so.1: version `GOMP_4.0' not found (required by /tmp/libVectorLoglessPairHMM.4842381956752440752.so)```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -463,8 +463,6 @@ model {
                         cppCompiler.args "-I", "${System.properties['java.home']}/../include/Darwin"
                     } else {
                         cppCompiler.args "-I", "${System.properties['java.home']}/../include/linux"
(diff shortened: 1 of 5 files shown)


droazen:
:+1: merge when tests pass

PR #1831, opened 2016-05-17
remove sputnik

akiezun:
too much noise, too little control

@droazen


droazen:
merge when ready @akiezun

PR #1834, opened 2016-05-18
upgrade spark to 1.6.1 and a few more upgrades

akiezun:
@tomwhite can you review? We have CHD 5.7 running now and 1.6 is available on the cloud so no reason to not upgrade AFAIK

For some reason, the lists returned from `.collect` are no longer mutable so i have to make copies in 1 test


tomwhite:
+1

PR #1825, opened 2016-05-16
Allele-specific annotations and tests

akiezun:
this is the initial port of the Allele Specific annotation for HaplotypeCaller. It mostly focuses on the GVCF mode (ie outputs the 'raw' data). I have a branch in protected https://github.com/broadinstitute/gatk-protected/tree/ak_haplotypecaller_allele_specific_annotations that uses those and I verified that the annotations are correctly output and their values are much closer that before to those from GATK3.5. I did not port any code related to combining the annotations in GenotypeGVCFs or CombinedGVCFs etc. Also, no code for VariantAnnotator or UnifiedGenotyper was ported - gatk4 does not have those tools right now.

@droazen can you review? Sorry, this is a whole bunch of code and it's not the final version yet (in particular, little effort was put into redesigning the framework - that will wait until we have integration tests so we can keep the results stable while improving design and code)

We also need to add tickets to:
- turn dithering off/on in RankSum tests (it's always off for now to simplify testing)
- use AlleleSpecific annotations in the VCF mode
- (later) port code for combining annotations

PR #1840, opened 2016-05-19
new PR for  allele specific annotations

akiezun:
adds originals to https://github.com/broadinstitute/gatk/pull/1825


src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
```diff
@@ -1,329 +1,121 @@
...
+    }
+
+    @Override
+    public Map<String, Object> annotateRawData(final ReferenceContext ref,
                                                final VariantContext vc,
                                                final Map<String, PerReadAlleleLikelihoodMap> perReadAlleleLikelihoodMap ) {
-
-        if ( perReadAlleleLikelihoodMap == null)
-            return new HashMap<>();
+        if ( perReadAlleleLikelihoodMap == null) {
```
droazen: Will this ever be true in GATK4?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
```diff
@@ -1,329 +1,121 @@
...
-        //shortcut for no ref values
-        if (perAltRankSumResults.isEmpty())
-            return annotations;
-        final String annotationString = makeReducedAnnotationString(vc, perAltRankSumResults);
-        annotations.put(getKeyNames().get(0), annotationString);
-        return annotations;
+        return annotationString.toString();
     }
 
+    @SuppressWarnings({"unchecked", "rawtypes"})//FIXME generics here blow up
```
droazen: Gross! Is there no way we can properly parameterize `ReducibleAnnotationData` in the method signature?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeff.java
```diff
@@ -121,47 +55,39 @@ public void initialize ( AnnotatorCompatible walker, GenomeAnalysisEngine toolki
...
     @Override
-    public Map<String, Object> annotate(final RefMetaDataTracker tracker,
-                                        final AnnotatorCompatible walker,
-                                        final ReferenceContext ref,
-                                        final Map<String, AlignmentContext> stratifiedContexts,
+    public Map<String, Object> annotate(final ReferenceContext ref,
                                         final VariantContext vc,
                                         final Map<String, PerReadAlleleLikelihoodMap> perReadAlleleLikelihoodMap ) {
+        Utils.nonNull(vc);
+        final HeterozygosityUtils heterozygosityUtils = new HeterozygosityUtils(vc);
```
droazen: Rename to something that doesn't end in `Utils`, since it looks bizarre to be instantiating a "utils" class like this.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
```diff
@@ -0,0 +1,121 @@
...
+                                         final VariantContext vc,
+                                         final Map<String, PerReadAlleleLikelihoodMap> stratifiedPerReadAlleleLikelihoodMap) {
+        return annotateRawData(ref, vc, stratifiedPerReadAlleleLikelihoodMap);
+    }
+
+    @Override
+    public Map<String, Object> annotateRawData(final ReferenceContext ref,
+                                               final VariantContext vc,
+                                               final Map<String, PerReadAlleleLikelihoodMap> perReadAlleleLikelihoodMap ) {
+        if ( perReadAlleleLikelihoodMap == null) {
```
droazen: Will this ever be true in GATK4? If so, document in the interface javadoc

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
```diff
@@ -0,0 +1,121 @@
...
+            if (i!=0) {
+                annotationString.append(printDelim);
+            }
+            final CompressedDataList<Integer> alleleValues = perAlleleValues.get(vcAlleles.get(i));
+            annotationString.append(alleleValues);
+        }
+        return annotationString.toString();
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})//FIXME generics here blow up
```
droazen: Gross! Is there no way we can properly parameterize `ReducibleAnnotationData` in the method signature?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeff.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    public List<VCFInfoHeaderLine> getDescriptions() { return Collections.singletonList(GATKVCFHeaderLines.getInfoLine(getKeyNames().get(0))); }
+
+    @Override
+    public Map<String, Object> annotate(final ReferenceContext ref,
+                                        final VariantContext vc,
+                                        final Map<String, PerReadAlleleLikelihoodMap> perReadAlleleLikelihoodMap ) {
+        Utils.nonNull(vc);
+        final HeterozygosityUtils heterozygosityUtils = new HeterozygosityUtils(vc);
```
droazen: Rename `HeterozygosityUtils` to something that doesn't end in `Utils`, since it looks bizarre to be instantiating a "utils" class like this.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeff.java
```diff
@@ -0,0 +1,105 @@
...
+ *     <li><b><a href="https://www.broadinstitute.org/gatk/guide/tooldocs/org_broadinstitute_gatk_tools_walkers_annotator_InbreedingCoeff.php">InbreedingCoeff</a></b> outputs a version of this annotation that includes all alternate alleles in a single calculation.</li>
+ *     <li><b><a href="https://www.broadinstitute.org/gatk/gatkdocs/org_broadinstitute_gatk_tools_walkers_annotator_ExcessHet.php">ExcessHet</a></b> estimates excess heterozygosity in a population of samples.</li>
+ * </ul>
+ *
+ */
+//TODO: this can't extend InbreedingCoeff because that one is Standard and it would force this to be output all the time; should fix code duplication nonetheless
+public final class AS_InbreedingCoeff extends InfoFieldAnnotation implements AS_StandardAnnotation {
+
+    public static final int MIN_SAMPLES = 10;
+    private Set<String> founderIds;    //TODO: either use this or enter a bug report
```
droazen: Enter a ticket for this TODO

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_QualByDepth.java
```diff
@@ -0,0 +1,73 @@
...
+    public List<VCFInfoHeaderLine> getDescriptions() {
+        //We only have the finalized key name here because the raw key is internal to GenotypeGVCFs and won't get output in any VCF
+        return Arrays.asList(GATKVCFHeaderLines.getInfoLine(getKeyNames().get(0)));
+    }
+
+    @Override
+    public Map<String, Object> annotate(final ReferenceContext ref,
+                                        final VariantContext vc,
+                                        final Map<String, PerReadAlleleLikelihoodMap> perReadAlleleLikelihoodMap ) {
+        return null;
```
droazen: This annotation does not appear to be in a usable state (`annotate()` and `annotateRawData()` both return null, and `calculateRawData()` is a no-op). Document this fact and explain why (opening a ticket if justified). Also, should there be default implementations of `annotateRawData()`, etc., in the interface that return null?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_QualByDepth.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public Map<String, Object> annotateRawData(final ReferenceContext ref,
+                                               final VariantContext vc,
+                                               final Map<String, PerReadAlleleLikelihoodMap> stratifiedPerReadAlleleLikelihoodMap) {
+        return null;
+    }
+
+    @Override
+    @SuppressWarnings({"unchecked", "rawtypes"})//FIXME
```
droazen: `FIXME` with no explanation -- briefly describe why you were unable to avoid using a raw type for posterity

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSAnnotation.java
```diff
@@ -0,0 +1,192 @@
...
+* 8.5 Amendment and Waiver; Entire Agreement. This Agreement may be amended, supplemented, or otherwise modified only by means of a written instrument signed by all parties. Any waiver of any rights or failure to act in a specific instance shall relate only to such instance and shall not be construed as an agreement to waive any rights or fail to act in any other instance, whether or not similar. This Agreement constitutes the entire agreement among the parties with respect to its subject matter and supersedes prior agreements or understandings between the parties relating to its subject matter.
+* 8.6 Binding Effect; Headings. This Agreement shall be binding upon and inure to the benefit of the parties and their respective permitted successors and assigns. All headings are for convenience only and shall not affect the meaning of any provision of this Agreement.
+* 8.7 Governing Law. This Agreement shall be construed, governed, interpreted and applied in accordance with the internal laws of the Commonwealth of Massachusetts, U.S.A., without regard to conflict of laws principles.
```
droazen: Forgot to delete the license in this one

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSAnnotation.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.*;
+
+/**
+ * Allele-specific implementation of root-mean-squared annotations
+ */
+public abstract class AS_RMSAnnotation extends RMSAnnotation {
+    protected final static Logger logger = Logger.getLogger(AS_RMSAnnotation.class);
+    protected final String splitDelim = "\\|"; //String.split takes a regex, so we need to escape the pipe
+    protected final String printDelim = "|";
+    protected AnnotatorCompatible callingWalker;
```
droazen: I thought the `AnnotatorCompatible` interface was gone?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSAnnotation.java
```diff
@@ -0,0 +1,192 @@
...
+
+
+    @Override
+    public void initialize(final AnnotatorCompatible walker, final GenomeAnalysisEngine toolkit, final Set<VCFHeaderLine> headerLines) {
```
droazen: Oh dear -- I think something has gone wrong in this class (`AS_RMSAnnotation`). It does not appear to have been ported properly.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSMappingQuality.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private final String printFormat = "%.2f";
+
+    private static final Logger logger = Logger.getLogger(AS_RMSMappingQuality.class);
+    public static final String SPLIT_DELIM = "\\|"; //String.split takes a regex, so we need to escape the pipe
+    public static final String PRINT_DELIM = "|";
+
+    public List<VCFInfoHeaderLine> getDescriptions() {
+        //TODO I'm just going to ignore this for now
+//        if (AnnotationUtils.walkerRequiresRawData(callingWalker))
```
droazen: Capture this TODO in a ticket

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSMappingQuality.java
```diff
@@ -0,0 +1,139 @@
...
+        getRMSDataFromPRALM(perReadAlleleLikelihoodMap, myData);
+    }
+
+    @Override
+    public List<String> getKeyNames() { return Arrays.asList(GATKVCFConstants.AS_RMS_MAPPING_QUALITY_KEY); }
+
+    @Override
+    public String getRawKeyName() { return GATKVCFConstants.AS_RAW_RMS_MAPPING_QUALITY_KEY; }
+
+    public void getRMSDataFromPRALM(Map<String, PerReadAlleleLikelihoodMap> perReadAlleleLikelihoodMap, ReducibleAnnotationData<Number> myData) {
```
droazen: Does this have to be public? If so, add javadoc

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSMappingQuality.java
```diff
@@ -0,0 +1,139 @@
...
+                    if (myData.hasAttribute(bestAllele)) {
+                        currSquareSum += (double) myData.getAttribute(bestAllele);
+                    }
+                    myData.putAttribute(bestAllele, currSquareSum + mq * mq);
+                }
+            }
+        }
+    }
+
+    public String makeRawAnnotationString(final List<Allele> vcAlleles, final Map<Allele, Number> perAlleleValues) {
```
droazen: Does this have to be public? If yes, document.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
```diff
@@ -0,0 +1,121 @@
+package org.broadinstitute.hellbender.tools.walkers.annotator;
```
droazen: We should probably move all of the allele-specific annotations into their own subpackage within `annotations`

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
```diff
@@ -0,0 +1,121 @@
...
+                                         final Map<String, PerReadAlleleLikelihoodMap> stratifiedPerReadAlleleLikelihoodMap) {
+        return annotateRawData(ref, vc, stratifiedPerReadAlleleLikelihoodMap);
+    }
+
+    @Override
+    public Map<String, Object> annotateRawData(final ReferenceContext ref,
+                                               final VariantContext vc,
+                                               final Map<String, PerReadAlleleLikelihoodMap> perReadAlleleLikelihoodMap ) {
+        if ( perReadAlleleLikelihoodMap == null) {
+            return null;
```
droazen: Previous version returned an empty `Map` in this case -- is the change intentional / documented in the method contract?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_ReadPosRankSumTest.java
```diff
@@ -0,0 +1,72 @@
...
+ * <li>Uninformative reads are not used in these annotations.</li>
+ * </ul>
+ *
+ * <h3>Related annotations</h3>
+ * <ul>
+ *     <li><b><a href="https://www.broadinstitute.org/gatk/guide/tooldocs/org_broadinstitute_gatk_tools_walkers_annotator_ReadPosRankSumTest.php">ReadPosRankRankSumTest</a></b> outputs a version of this annotation that includes all alternate alleles in a single calculation.</li>
+ * </ul>
+ *
+ */
+//FIXME - this whole class is essentially code duplicated from ReadPosRankSumTest
```
droazen: Create ticket?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
```diff
@@ -0,0 +1,229 @@
...
+
+import java.util.*;
+
+/**
+ * Allele-specific implementation of strand bias annotations
+ */
+public abstract class AS_StrandBiasTest extends StrandBiasTest implements ReducibleAnnotation {
+    private final static Logger logger = Logger.getLogger(AS_StrandBiasTest.class);
+    protected static final String splitDelim = "\\|"; //String.split takes a regex, so we need to escape the pipe
+    protected static final String printDelim = "|";
```
droazen: Same delimiter expressions are repeated in several classes -- should they be declared as a shared constant somewhere?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    @Override
+    public Map<String, Object> annotateRawData(final ReferenceContext ref,
+                                               final VariantContext vc,
+                                               final Map<String, PerReadAlleleLikelihoodMap> perReadAlleleLikelihoodMap ) {
+
+        //for allele-specific annotations we only call from HC and we only use perReadAlleleLikelihoodMap
+        if ( perReadAlleleLikelihoodMap == null) {
+            return new HashMap<>();
```
droazen: You are inconsistent on the return value in this case -- some annotations return an empty `Map`, others return null. Stick to one or the other and document behavior in method contract.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
```diff
@@ -0,0 +1,229 @@
...
+            // calculate the annotation from the stratified per read likelihood map
+            // stratifiedPerReadAllelelikelihoodMap can come from HaplotypeCaller call to VariantAnnotatorEngine
```
droazen: indentation

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
```diff
@@ -0,0 +1,229 @@
...
+        final List<Integer> alleleStrandCounts;
+        if (perAlleleValues.hasAttribute(bestAllele) && perAlleleValues.getAttribute(bestAllele) != null)
+            alleleStrandCounts = perAlleleValues.getAttribute(bestAllele);
+        else {
+            alleleStrandCounts = new ArrayList<>();
+            alleleStrandCounts.add(0,0);
+            alleleStrandCounts.add(1,0);
+        }
+        final boolean isFW = !read.isReverseStrand();
+        if (isFW) {
```
droazen: `isFW` -> `isForwardStrand`

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
lbergelson: If there's a choice between empty map and null I'd definitely vote for empty map.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
```diff
@@ -0,0 +1,229 @@
...
+            if (j < alleleValues.size()-1)
+                annotationString += ",";
+        }
+        return annotationString;
+    }
+
+    @Override
+    @SuppressWarnings({"unchecked", "rawtypes"})//FIXME
+    public void calculateRawData(final VariantContext vc,
+                                 final Map<String, PerReadAlleleLikelihoodMap> pralm,
```
droazen: `pralm` is not the most best parameter name...

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
```diff
@@ -0,0 +1,229 @@
...
+        else {
+            alleleStrandCounts = new ArrayList<>();
+            alleleStrandCounts.add(0,0);
+            alleleStrandCounts.add(1,0);
+        }
+        final boolean isFW = !read.isReverseStrand();
+        if (isFW) {
+            alleleStrandCounts.set(FORWARD, alleleStrandCounts.get(FORWARD) + 1);
+        } else {
+            alleleStrandCounts.set(REVERSE, alleleStrandCounts.get(REVERSE) + 1);
```
droazen: What about the strandless case handled in GATK3?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
```diff
@@ -0,0 +1,229 @@
...
+                readTotal += alleleValues.get(FORWARD);
+                readTotal += alleleValues.get(REVERSE);
+            }
+        }
+        return readTotal > minCount;
+    }
+
+    @Override
+    //This just calls the non-allele-specific code in StrandBiasTest.java
+    protected abstract Map<String, Object> calculateAnnotationFromLikelihoodMap(final Map<String, PerReadAlleleLikelihoodMap> stratifiedPerReadAlleleLikelihoodMap,
```
droazen: Why is this abstract method "overridden" here only to keep it abstract?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
```diff
@@ -0,0 +1,229 @@
...
+                                                                                final VariantContext vc);
+
+    @Override
+    //Allele-specific annotations cannot be called from walkers other than HaplotypeCaller
```
droazen: Where is this constraint enforced?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
```diff
@@ -0,0 +1,229 @@
...
+    protected Map<String, Object> calculateAnnotationFromGTfield(final GenotypesContext genotypes){
+        return new HashMap<>();
+    }
+
+    public static String rawValueAsString(int[][] table) {
+        return table[0][1]+","+table[0][1]+printDelim+table[1][0]+","+table[1][1];
```
droazen: Move to a utility class or make private.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandOddsRatio.java
```diff
@@ -0,0 +1,70 @@
...
+        return Collections.singletonList(GATKVCFConstants.AS_STRAND_ODDS_RATIO_KEY);
+    }
+
+    @Override
+    protected Map<String, Object> calculateAnnotationFromLikelihoodMap(final Map<String, PerReadAlleleLikelihoodMap> stratifiedPerReadAlleleLikelihoodMap,
+                                                                       final VariantContext vc){
+        // either SNP with no alignment context, or indels: per-read likelihood map needed
+        final int[][] table = getContingencyTable(stratifiedPerReadAlleleLikelihoodMap, vc, MIN_COUNT);
+        final double ratio = StrandOddsRatio.calculateSOR(table);
+        return Collections.singletonMap(getKeyNames().get(0), StrandOddsRatio.formattedValue(ratio));
```
droazen: Should there be an interface method for formatting values? Seems like several of the annotations have code like this.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AlleleSpecificAnnotationData.java
```diff
@@ -0,0 +1,50 @@
...
+public final class AlleleSpecificAnnotationData<T> extends ReducibleAnnotationData<T>{
+    private final List<Allele> alleleList;
+    private final Allele refAllele;
+
+    public AlleleSpecificAnnotationData(final List<Allele> inputAlleles, final String inputData) {
+        super(inputData);
+        this.attributeMap = new HashMap<>();
+        inputAlleles.forEach(a -> {attributeMap.put(a, null);});
+        alleleList = inputAlleles;
+        refAllele = alleleList.stream().filter(Allele::isReference).findAny().orElse(null);
```
droazen: This streaming expression is much less readable/straightforward than the previous for loop.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AnnotationUtils.java
```diff
@@ -0,0 +1,23 @@
...
+
+public final class AnnotationUtils {
+    private AnnotationUtils(){}
+
+    /**
+     * Helper function to parse the list into the annotation string
+     * @param valueList the ArrayList returned from StrandBiasBySample.annotate()
+     * @return the array used by the per-sample Strand Bias annotation
+     */
+    protected static String encodeValueList(final List<Double> valueList, final String precisionFormat ) {
```
droazen: This is all you ported of this class? None of the other methods are needed?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/CompressedDataList.java
```diff
@@ -0,0 +1,92 @@
...
+    public void add(final T val){
+        add(val, 1);
+    }
+
+    public void add(final T val, final int count){
+        if(valueCounts.containsKey(val)){
+            valueCounts.put(val, valueCounts.get(val)+count);
+        }
+        else
+            valueCounts.put(val, count);
```
droazen: Missing braces

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/CompressedDataList.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public String toString(){
+        String str = "";
+        Object[] keys = valueCounts.keySet().toArray();
+        Arrays.sort(keys);
+        for (Object i: keys){
+            if(!str.isEmpty())
+                str+=",";
```
droazen: Missing braces

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/CompressedDataList.java
```diff
@@ -0,0 +1,92 @@
...
+            valueCounts.put(val, valueCounts.get(val)+count);
+        }
+        else
+            valueCounts.put(val, count);
+
+    }
+
+    public void add(final CompressedDataList<T> obj){
+        for(Map.Entry<T, Integer> pair : obj.getValueCounts().entrySet()){
+            this.add(pair.getKey(),pair.getValue());
```
droazen: Surprised you didn't streamify this loop

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/HeterozygosityUtils.java
```diff
@@ -0,0 +1,118 @@
...
+     * Get the count of heterozygotes in vc for a specific altAllele (both reference and non-reference hets, e.g. 1/2)
+     * @param altAllele the alternate allele of interest
+     * @return number of hets
+     */
+    protected double getHetCount(final Allele altAllele) {
+        return hetCounts.containsKey(altAllele)? hetCounts.get(altAllele) : 0;
+    }
+
+    protected double getAlleleCount(final Allele allele) {
+        return alleleCounts.containsKey(allele)? alleleCounts.get(allele) : 0;
```
droazen: In these methods, could `hetCounts` or `alleleCounts` be null?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/HeterozygosityUtils.java
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            int idxAB;
+
+            //for each sample
+            for (final Genotype g : genotypes) {
+                if (g.isCalled() && g.hasLikelihoods() && g.getPloidy() == 2)  // only work for diploid samples
+                    sampleCount++;
+                else
+                    continue;
```
droazen: braces

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/HeterozygosityUtils.java
```diff
@@ -0,0 +1,118 @@
...
+     */
+    protected HeterozygosityUtils(final VariantContext vc) {
+        this.vc = vc;
+        doGenotypeCalculations();
+    }
+
+    /**
+     * Get the count of heterozygotes in vc for a specific altAllele (both reference and non-reference hets, e.g. 1/2)
+     */
+    private void doGenotypeCalculations() {
```
droazen: Method overly long, should probably be decomposed into sub-methods

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/HeterozygosityUtils.java
```diff
@@ -0,0 +1,118 @@
...
+import org.broadinstitute.hellbender.utils.MathUtils;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A class containing utility methods used in the calculation of annotations related to cohort heterozygosity, e.g. InbreedingCoefficient and ExcessHet
+ * Stores sample count to make sure we never have to iterate the genotypes more than once.
+ */
+final class HeterozygosityUtils {
```
droazen: Definitely rename

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
```diff
@@ -31,32 +32,64 @@
...
+                                 final ReducibleAnnotationData rawAnnotations){
+        if (pralm.isEmpty()) {
+            return;
+        }
+
+        //put this as a double, like GATK3.5
+        final double squareSum = pralm.values().stream()
+                .flatMap(likelihoodMap -> likelihoodMap.getReads().stream())
+                .map(GATKRead::getMappingQuality)
+                .filter(mq -> mq != QualityUtils.MAPPING_QUALITY_UNAVAILABLE).mapToDouble(mq -> mq * mq).sum();
```
droazen: Shouldn't you be taking the average, like in `MathUtils.rms()`?

droazen: Shouldn't you be taking the average, like in `MathUtils.rms()`?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RankSumTest.java
```diff
@@ -26,7 +26,7 @@ public RankSumTest(final boolean useDithering){
     }
 
     public RankSumTest(){
-        this(true);
+        this(false);
```
droazen: Why this change?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReducibleAnnotation.java
```diff
@@ -0,0 +1,24 @@
...
+
+import htsjdk.variant.variantcontext.VariantContext;
+import org.broadinstitute.hellbender.engine.ReferenceContext;
+import org.broadinstitute.hellbender.utils.genotyper.PerReadAlleleLikelihoodMap;
+
+import java.util.Map;
+
+/**
+ * An interface for annotations that are calculated using raw data across samples, rather than the median (or median of median) of samples values
+ */
```
droazen: Make note of the missing functionality in this interface compared to GATK3, and create a ticket to track this unported code.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -128,6 +128,14 @@ public static double rms(final Collection<Integer> list) {
     }
 
     /**
+     * Computes the sume of squares of the given array or 0.0 if the array is empty.
```
droazen: sume -> sum

droazen: array -> `Collection<Integer>`

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVCFConstants.java
```diff
@@ -10,13 +10,18 @@
 public final class GATKVCFConstants {
```
droazen: Can you check the contents of our `GATKVCFConstants` against the GATK3 version? They should be identical at this point, I would think.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVCFHeaderLines.java
```diff
@@ -98,8 +98,12 @@ private static void addFilterLine(final VCFFilterHeaderLine line) {
         addInfoLine(new VCFInfoHeaderLine(LOW_MQ_KEY, 3, VCFHeaderLineType.Float, "3-tuple: <fraction of reads with MQ=0>,<fraction of reads with MQ<=10>,<total number of reads>"));
```
droazen: Same with our `GATKVCFHeaderLines` -- check against GATK3, make sure we're identical

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeffUnitTest.java
```diff
@@ -0,0 +1,217 @@
...
+        final Map<String, Object> annots = new AS_InbreedingCoeff().annotate(null, test2, null);
+        Assert.assertEquals(annots.keySet(), Collections.singleton(GATKVCFConstants.AS_INBREEDING_COEFFICIENT_KEY), "annots");
+        Assert.assertEquals(annots.values().size(), 1, "size");
+        final String annotString = (String)annots.values().iterator().next();
+        final String[] split = annotString.split(",");
+        Assert.assertEquals(split.length, 2);
+        Assert.assertEquals(Double.parseDouble(split[0]), -0.666666, DELTA_PRECISION, "ic");
+        Assert.assertEquals(Double.parseDouble(split[1]), -0.111129, DELTA_PRECISION, "ic");
+
+        if (false) { //TODO reenable when AS_InbreedingCoeff uses founders
```
droazen: Comment this block out instead of using `if (false)` in the code

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQualityRankSumTestUnitTest.java
```diff
@@ -46,8 +47,17 @@ private GATKRead makeRead(final byte qual) {
         return read;
     }
 
-    @Test
+    @DataProvider(name = "data")
+    private Object[][] data(){
+        return new Object[][]{
+                {},
```
droazen: What the heck is this? :) Why even have a `DataProvider` if your only data is an empty array?

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQualityUnitTest.java
```diff
@@ -13,6 +13,7 @@
 import org.broadinstitute.hellbender.utils.genotyper.PerReadAlleleLikelihoodMap;
 import org.broadinstitute.hellbender.utils.read.ArtificialReadUtils;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
+import org.broadinstitute.hellbender.utils.variant.GATKVCFConstants;
 import org.testng.Assert;
 import org.testng.annotations.Test;
```
droazen: Hmm, how is this unit test class passing? Unless I missed something when reading the code (entirely possible, as looking at the diffs vs. GATK3 is slightly disorienting) we are forgetting to take the average in `RMSMappingQuality` when computing the RMS...

droazen:
First-pass review complete, back to @akiezun. This is a bit messy/spotty, with large blocks of missing GATK3 functionality, some annotations not in a usable state (eg., `AS_QualByDepth`), lots of TODOs in the code, and the branch is not currently compiling. We can probably get it merged in as a work-in-progress, but we should make sure we have tickets to capture the remaining work to be done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
akiezun: yes, it can be null. Documented now

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
akiezun: not that i can see - method parameters are invariant in Java (ie must be exactly the same in all overriding methods) so it's a huge pain here.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeff.java
akiezun: agreed, renamed to `HeterozygosityCalculator`

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeff.java
akiezun: done https://github.com/broadinstitute/gatk/issues/1880

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeffUnitTest.java
```diff
@@ -0,0 +1,300 @@
+/*
```
akiezun: fixed

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_QualByDepth.java
akiezun: this is exactly how it works in gatk3. The actual work happens on combine

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_QualByDepth.java
akiezun: added a comment. generics won't work here because of method overriding

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSAnnotation.java
akiezun: fixed

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSAnnotation.java
akiezun: fixed. this class is dead now

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSMappingQuality.java
akiezun: done https://github.com/broadinstitute/gatk/issues/1881

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSMappingQuality.java
akiezun: nope, made private now

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSMappingQuality.java
akiezun: private now

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RankSumTest.java
akiezun: fixed to return empty map everywhere

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_RMSAnnotation.java
akiezun: AS_RMSAnnotation  class deleted now

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_ReadPosRankSumTest.java
akiezun: https://github.com/broadinstitute/gatk/issues/1882

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
akiezun: probably not, in principle they could be different for every kind of annotation

akiezun: will rename the constants though

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
akiezun: Collections.emptyMap everywhere

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
akiezun: fixed. also removed the dead branch (last else is unreachable, same bug as in gatk3)

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
akiezun: renamed to `stratifiedPerReadAlleleLikelihoodMap`

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
akiezun: strandless was only for reduced reads which are dead in gatk4

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
akiezun: good catch. deleted

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
akiezun: nowhere. we only have 1 thing that makes annotations

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandBiasTest.java
akiezun: it does if we want those classes in a separate package

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_StrandOddsRatio.java
akiezun: meh, probably not worth it for 3 lines

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AnnotationUtils.java
akiezun: not yet. It's a kitchen sink class that should die eventually

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/CompressedDataList.java
akiezun: fixed

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/CompressedDataList.java
akiezun: fixed

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/HeterozygosityUtils.java
akiezun: i think not. i added an arg check

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/HeterozygosityUtils.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/HeterozygosityUtils.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
akiezun: at the 'raw' stage, we just collect the sum of squares. The root part comes in combining.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RankSumTest.java
akiezun: because tests must be reproducible. https://github.com/broadinstitute/gatk/issues/1881 captures the need to make `dither` into a parameter

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReducibleAnnotation.java
akiezun: this is obvious i think - we have not ported the combine part.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReducibleAnnotation.java
droazen: Create a ticket to track it anyway

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVCFConstants.java
akiezun: well, modulo any un-ported stuff like beagle inputs or phase by transmission or whatever else

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVCFHeaderLines.java
akiezun: same here - some beagle stuff changed, rest is the same (some are unported eg FractionInformativeReads)

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeffUnitTest.java
akiezun: why? using if false is far superior IMO - checks the the code compiles and gets refactored with other code.

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQualityRankSumTestUnitTest.java
akiezun: some copy/paste bogosity - removed now

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeffUnitTest.java
droazen: Yes, you're right -- I take it back :)

akiezun:
comments addressed. i rebased too so it compiles now back to @droazen

akiezun:
@droazen let me know if we can merge this

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AlleleSpecificAnnotationData.java
```diff
@@ -0,0 +1,50 @@
+package org.broadinstitute.hellbender.tools.walkers.annotator;
```
droazen: This should be in the allelespecific package.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
```diff
@@ -31,32 +32,64 @@
  * </ul>
  *
  */
-public final class RMSMappingQuality extends InfoFieldAnnotation implements StandardAnnotation {
+public final class RMSMappingQuality extends InfoFieldAnnotation implements StandardAnnotation, ReducibleAnnotation {
```
droazen: Can you explain why `RMSMappingQuality` implements `ReducibleAnnotation` now and has such a large diff? Don't we have `AS_RMSMappingQuality` (which doesn't extend `RMSMappingQuality`)?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/InfoFieldAnnotation.java
```diff
@@ -17,7 +17,7 @@
 
     /**
      * Computes the annotation for the given variant and the likelihoods per read.
-     * Returns a map from annotation keys to values.
+     * Returns a map from annotation keys to values or null if no annotation is to be added.
```
droazen: Is this right? It seems to me you are returning empty maps now

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
```diff
@@ -31,32 +32,64 @@
...
     }
 
     @VisibleForTesting
     static String formatedValue(double rms) {
         return String.format("%.2f", rms);
     }
 
-    public List<String> getKeyNames() { return Collections.singletonList(VCFConstants.RMS_MAPPING_QUALITY_KEY); }
+    @Override
+    public List<String> getKeyNames() { return Collections.singletonList(getRawKeyName()); }
```
droazen: This really doesn't seem right -- the key for this annotation should be `RMS_MAPPING_QUALITY_KEY`, not `RAW_RMS_MAPPING_QUALITY_KEY`, shouldn't it? In GATK3 `getKeyNames()` returns `RMS_MAPPING_QUALITY_KEY` for this annotation...

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReducibleAnnotation.java
```diff
@@ -0,0 +1,27 @@
...
+public interface ReducibleAnnotation extends Annotation {
```
droazen: Add comment that `ReducibleAnnotation` currently lacks any `reduce` functionality, since it hasn't been ported (interface is bizarre/confusing without this comment).

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/MappingQualityRankSumTestUnitTest.java
```diff
@@ -69,18 +73,64 @@ public void testMQ(){
...
+
+        Assert.assertEquals(ann.getDescriptions().size(), 1);
+        Assert.assertEquals(ann.getDescriptions().get(0).getID(), key);
+        Assert.assertEquals(ann.getKeyNames().size(), 1);
+        Assert.assertEquals(ann.getKeyNames().get(0), key);
+    }
+
+    @Test
+    public void testAS_MQRaw(){
+        final AS_RankSumTest ann = new AS_MappingQualityRankSumTest();
```
droazen: I think it would be better to have separate test classes for the AS vs. non-AS annotations (eg., `MappingQualityRankSumTestUnitTest` and `AS_MappingQualityRankSumTestUnitTest` here) to make it easier to see at a glance (by looking at the file listing) which annotations are covered by tests.

droazen:
Back to @akiezun -- a few more TODOs plus some concerns about `RMSMappingQuality` that need answering.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/AlleleSpecificAnnotationData.java
akiezun: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeffUnitTest.java
```diff
@@ -0,0 +1,218 @@
+package org.broadinstitute.hellbender.tools.walkers.annotator;
```
akiezun: i think it would be cleaner to keep all in one package because we may move code around from allele specific to regular etc. but i'm fine with this too

droazen: It's important to have tests be in the same package, because when we make something `@VisibileForTesting` we rely on this.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
akiezun: in gatk3 `RMSMappingQuality implements ReducibleAnnotation` as well
```
public class RMSMappingQuality extends RMSAnnotation implements StandardAnnotation, ActiveRegionBasedAnnotation, ReducibleAnnotation {
````

Comment (akiezun) [inline src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java]: in GATK4 I deleted the `RMSAnnotation` abstract class and the `AS_RMSAnnotation` abstract class because each had exactly one subclass and the generalization feels premature

File: src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReducibleAnnotation.java (hunk 52 above)
Comment (akiezun) [inline src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReducibleAnnotation.java]: done

File: src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeffUnitTest.java (hunk 54 above)
Comment (akiezun) [inline src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/AS_InbreedingCoeffUnitTest.java]: that's why i think it would be better not to have a special package for allele specific stuff

File: src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/MappingQualityRankSumTestUnitTest.java (hunk 53 above)
Comment (akiezun) [inline src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/MappingQualityRankSumTestUnitTest.java]: this way we can test the common functionality much easier - just by using a data provider. otherwise, we'll have a lot of duplicated test logic

File: src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQualityUnitTest.java (hunk 46 above)
Comment (akiezun) [inline src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQualityUnitTest.java]: this passes because we now compute the sum of squares and leave the root computation to the 'reduce' step

File: src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/MappingQualityRankSumTestUnitTest.java [hunk 55]
```diff
@@ -39,8 +41,10 @@ private GATKRead makeRead(final int mq) {
         read.setMappingQuality(mq);
         return read;
     }
-    @Test
+
```
akiezun: oops we lost a `@Test` here - added back

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
akiezun: this behavior matches 3.5 HC when run with `-G StandardHC -G Standard` - this behavior changed from 3.4 and 3.5. HC now prints `RAW_MQ` and not `MQ` (oddly, the header in 3.5 contains `MQ` and `RAW_MQ`)

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
akiezun: gatk3 RMSMappingQuality is the only annotation that is both 'standard' and 'reducible' so that we get both the key and the rawKey. then there's code that asks only reducible annotations for rawKey. If we change this key here, we need to reintroduce that crazy code (otherwise VC blows up because the header is missing `RAW_MQ`)

droazen: Recommend that we be consistent with GATK3 for now, and file a ticket against GATK3 to discuss further and try to figure out why it was designed this way.

akiezun: i found a workaround - return 2 keys here and 2 descriptions in getDescriptions. Works fine

akiezun:
back to @droazen - comments addressed

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQualityUnitTest.java
```diff
@@ -26,14 +27,16 @@ public void testAllNull() throws Exception {
...
     }
 
     @Test
     public void testDescriptions() throws Exception {
         final InfoFieldAnnotation cov = new RMSMappingQuality();
-        Assert.assertEquals(cov.getDescriptions().size(), 1);
+        Assert.assertEquals(cov.getDescriptions().size(), 2);
         Assert.assertEquals(cov.getDescriptions().get(0).getID(), VCFConstants.RMS_MAPPING_QUALITY_KEY);
+        Assert.assertEquals(cov.getDescriptions().get(1).getID(), GATKVCFConstants.RAW_RMS_MAPPING_QUALITY_KEY);
+        Assert.assertEquals(new RMSMappingQuality().getRawKeyName(), GATKVCFConstants.RAW_RMS_MAPPING_QUALITY_KEY);
```
droazen: Add an assert for `getKeyNames()` as well

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/InfoFieldAnnotation.java
droazen: Not sure you've addressed this @akiezun -- is this doc accurate, or should it say empty map instead of null?

droazen:
I still think it's bad that sometimes you have tests for the AS annotations in the same class as its non-AS equivalent, and other times you have the AS tests in a separate class (eg., `AS_QualByDepthUnitTest`). It might be better to have the AS annotations always in a separate class, but extend the non-AS test class to share `DataProviders`, etc. Up to you, but should be consistent at least.

droazen:
Final review complete -- back to @akiezun. Rebase/squash/merge after addressing comments (don't forget the docs in `InfoFieldAnnotation`, which I still think are wrong -- my comment on it is buried in the older comments above).

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQualityUnitTest.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/InfoFieldAnnotation.java
akiezun: fixed now

Issue #1210, opened 2015-12-01
Determine fate of Picard tools


droazen:
Picard is still under (very) active development, while the ported versions of the picard tools in gatk4 are becoming more and more outdated. We need to either make gatk4 the official home of the picard tools and move all current picard development here, or give up on merging gatk and picard. Since having the picard tools and utilities as part of the codebase has been useful in many instances, and since hosting the official versions would also encourage more developers inside and outside the Broad to contribute to gatk4, I am hoping that we can make this marriage work!

vdauwera:
Agreed! Users want this, we want this -- let's please do whatever we need to make it happen!

One barrier to getting this to work, as I understand it, is the need for a build system that allows production/Kathleen to generate new internal releases of the Picard portion independently of the rest of GATK. This is because they often need to update Picard tools but usually don't want to pull in changes to the GATK tools at that time. This was discussed some time ago but I don't know if any work has been done in that direction (apologies if I'm stepping on any toes).

@eitanbanks @yfarjoun @ktibbett @akiezun @geoffjentry

droazen:
@vdauwera Does that really require a solution at the build system level? Couldn't we just have one hellbender jar that is used exclusively for running the Picard tools in production, and a second, separate hellbender jar that is used exclusively for running the GATK tools? Does it really matter if the "Picard" jar actually contains GATK tools, and vice versa, as long as we never run GATK tools from the "Picard" jar, or Picard tools from the "GATK" jar?

vdauwera:
@droazen That actually sounds perfectly fine to me, I think it would satisfy @ktibbet's needs. Forest/trees <-- me not seeing past them.

lbergelson:
It's also totally plausible to be able to build a separate picard jar from
the same project if that's a requirement.

droazen:
@lbergelson Not if there are GATK tools/utilities calling into Picard tools/utilities and vice versa (which there certainly is at this point!).

eitanbanks:
FYI @ktibbett is going to organize a meeting to discuss in person with interested parties

lbergelson:
Should we try to get Tim and Nils involved in the meeting as well? They're
the most prolific committers at the moment.

vdauwera:
Yes the plan is to involve them.

lbergelson:
Good.  We need them on board if we want to avoid forking.

droazen:
[...]

PR #1844, opened 2016-05-19
linked hash set, linked hash map

akiezun:
using linked or sorted data structures to reduce our dependency on the arbitrary changes in iterator orders (happens everytime you move from a jvm to another jvm)

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/ClipReads.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/ClipReads.java
@@ -505,7 +505,7 @@ public final class ClipReads extends ReadWalker {
         public long nRangeClippedBases = 0;
         public long nSeqClippedBases = 0;
 
-        HashMap<String, Long> seqClipCounts = new HashMap<>();
+        SortedMap<String, Long> seqClipCounts = new TreeMap<>();
 
         public ClippingData(List<SeqToClip> clipSeqs) {
             for (SeqToClip clipSeq : clipSeqs) {
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/validation/CompareDuplicatesSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/validation/CompareDuplicatesSpark.java
@@ -22,7 +22,6 @@ import org.broadinstitute.hellbender.utils.read.ReadUtils;
 import org.broadinstitute.hellbender.utils.read.markduplicates.ReadsKey;
 import scala.Tuple2;
 
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -106,7 +105,7 @@ public final class CompareDuplicatesSpark extends GATKSparkTool {
         }
         if (printSummary) {
             MatchType[] values = MatchType.values();
-            HashSet<MatchType> matchTypes = Sets.newHashSet(values);
+            Set<MatchType> matchTypes = Sets.newLinkedHashSet(Sets.newHashSet(values));
             System.out.println("##############################");
             matchTypes.forEach(s -> System.out.println(s + ": " + tagCountMap.getOrDefault(s, 0)));
         }
@@ -138,8 +137,8 @@ public final class CompareDuplicatesSpark extends GATKSparkTool {
         first.sort(new ReadCoordinateComparator(header));
         second.sort(new ReadCoordinateComparator(header));
 
-        Set<GATKRead> firstDupes = Sets.newHashSet();
-        Set<GATKRead> secondDupes = Sets.newHashSet();
+        Set<GATKRead> firstDupes = Sets.newLinkedHashSet();
+        Set<GATKRead> secondDupes = Sets.newLinkedHashSet();
         for (int i = 0; i < size; ++i) {
             GATKRead firstRead = first.get(i);
             GATKRead secondRead = second.get(i);
--- a/src/test/java/org/broadinstitute/hellbender/utils/collections/CountSetUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/collections/CountSetUnitTest.java
@@ -5,10 +5,7 @@ import org.testng.Assert;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Random;
+import java.util.*;
 
 /**
  * Unit tests for {@link CountSet}
@@ -30,7 +27,7 @@ public final class CountSetUnitTest extends BaseTest {
     public void testSingleValueAdd() {
         final int CAPACITY = 10;
         final CountSet subject = new CountSet(CAPACITY);
-        final HashSet<Integer> reasuranceSet = new HashSet<>(CAPACITY);
+        final Set<Integer> reasuranceSet = new LinkedHashSet<>(CAPACITY);
         final int REPEATS = 1000;
         final Random rnd = new Random(13);
         for (int i = 0; i < REPEATS; i++) {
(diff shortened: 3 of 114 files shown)


akiezun:
I'm going to abandon this - Picard tests depend on the order in contrived ways, eg:
```
    private DetailPair getWorstMetrics(final List<DetailPair> metrics) {
        PreAdapterDetailMetrics worstPreAdapterMetrics = null;
        BaitBiasDetailMetrics worstBaitBiasMetrics = null;
        for (final DetailPair m : metrics) {
            if (worstPreAdapterMetrics == null || m.preAdapterMetrics.QSCORE < worstPreAdapterMetrics.QSCORE) worstPreAdapterMetrics = m.preAdapterMetrics;
            if (worstBaitBiasMetrics == null || m.baitBiasMetrics.QSCORE < worstBaitBiasMetrics.QSCORE) worstBaitBiasMetrics = m.baitBiasMetrics;
        }
        return new DetailPair(worstPreAdapterMetrics, worstBaitBiasMetrics);
    }
```

if all `QSCORE` are the same this arbitrarily pick the first one. That backfires when I switch to linked hashmaps and linkedhashsets. Specifically `CollectSequencingArtifactMetricsTest` fails. Given that #1210 (fate of Picard) is unresolved I'm going to not put time into it yet.

droazen:
How about doing this everywhere except in Picard, instead of abandoning it completely? Non-linked hash sets/maps have caused MANY problems in GATK over the years.

ronlevine:
The issue is that only the linked list version of hash sets/maps do has a predictable iteration order. So, any algorithm that is order dependent may get different results between different Java versions. The only downside is the linked list will take up some additional storage.

akiezun:
fine but someone else does it when picard is decided one way or the other

akiezun:
@droazen can you review?

src/main/java/org/broadinstitute/hellbender/tools/ClipReads.java
```diff
@@ -505,7 +505,7 @@ public SeqToClip(String name, byte[] bytez) {
         public long nRangeClippedBases = 0;
         public long nSeqClippedBases = 0;
 
-        HashMap<String, Long> seqClipCounts = new HashMap<>();
+        SortedMap<String, Long> seqClipCounts = new TreeMap<>();
```
droazen: Why did you change `HashMap` to `TreeMap` and not `LinkedHashMap` here?

src/main/java/org/broadinstitute/hellbender/tools/spark/validation/CompareDuplicatesSpark.java
```diff
@@ -106,7 +105,7 @@ protected void runTool(final JavaSparkContext ctx) {
         }
         if (printSummary) {
             MatchType[] values = MatchType.values();
-            HashSet<MatchType> matchTypes = Sets.newHashSet(values);
+            Set<MatchType> matchTypes = Sets.newHashSet(values);
```
droazen: Does this produce a `LinkedHashSet`? If not, should you manually construct one?

src/test/java/org/broadinstitute/hellbender/utils/collections/CountSetUnitTest.java
```diff
@@ -30,7 +27,7 @@ public void testSize(final int capacity) {
     public void testSingleValueAdd() {
         final int CAPACITY = 10;
         final CountSet subject = new CountSet(CAPACITY);
-        final HashSet<Integer> reasuranceSet = new HashSet<>(CAPACITY);
+        final HashSet<Integer> reasuranceSet = new LinkedHashSet<>(CAPACITY);
```
droazen: type on the left side should be `Map`

src/test/java/org/broadinstitute/hellbender/utils/collections/CountSetUnitTest.java
```diff
@@ -118,7 +115,7 @@ public void testMinMax() {
     public void testIncrease() {
         final int CAPACITY = 10;
         final CountSet subject = new CountSet(CAPACITY);
-        final HashSet<Integer> reasuranceSet = new HashSet<>(CAPACITY);
+        final HashSet<Integer> reasuranceSet = new LinkedHashSet<>(CAPACITY);
```
droazen: type on the left side should be `Set`

src/test/java/org/broadinstitute/hellbender/utils/collections/NestedIntegerArrayUnitTest.java
```diff
@@ -57,7 +57,7 @@ public void testPutget1Key() throws Exception {
         arr.put("bozo", 1);
 
         Assert.assertEquals(2, arr.getAllLeaves().size());
-        Assert.assertEquals(Sets.newHashSet("fred", "bozo"), new HashSet<>(arr.getAllValues()));
+        Assert.assertEquals(Sets.newHashSet("fred", "bozo"), new LinkedHashSet<>(arr.getAllValues()));
```
droazen: Can you audit the codebase for uses of `Sets.newHashSet()`, and determine whether they'll cause any issues (eg., by creating non-linked `HashSets`)

src/test/resources/org/broadinstitute/hellbender/tools/ClipReads/expected.clippingReadsTest.withRG.hg19.QT_10_CT_15_X_CCCCC_XF.tmp
```diff
@@ -8,7 +8,7 @@ Percent of clipped bases              24.70
 Number of quality-score clipped bases 126
 Number of range clipped bases         65
 Number of sequence clipped bases      53
-         8 clip sites matching GAGG
         10 clip sites matching CCCCC
+         8 clip sites matching GAGG
```
droazen: Is this the only change in test output across all tools?

droazen:
Review complete, back to @akiezun, merge after addressing comments.

src/main/java/org/broadinstitute/hellbender/tools/ClipReads.java
akiezun: the `seqClipCounts` are better organized when sorted by the String key than just in arbitrary order. So sorted map is better.

src/main/java/org/broadinstitute/hellbender/tools/spark/validation/CompareDuplicatesSpark.java
akiezun: good catch. No, it does not produce a linkedhashset. I changed it in a few places where appropriate

src/test/java/org/broadinstitute/hellbender/utils/collections/CountSetUnitTest.java
akiezun: fixed here and throughout the codebase

src/test/java/org/broadinstitute/hellbender/utils/collections/CountSetUnitTest.java
akiezun: fixed

src/test/java/org/broadinstitute/hellbender/utils/collections/NestedIntegerArrayUnitTest.java
akiezun: yes, all done

src/test/resources/org/broadinstitute/hellbender/tools/ClipReads/expected.clippingReadsTest.withRG.hg19.QT_10_CT_15_X_CCCCC_XF.tmp
akiezun: yes, the only change was to resort those guys (which makes more sense then list them in an arbitrary order)

PR #1846, opened 2016-05-20
MLLib + svd calculations

akiezun:
Pushing MLLib down to public to make it easier for people to use with GATK4. Took SVD along for the ride (to test that it works).


droazen:
@LeeTL1220 can you review this one?

LeeTL1220:
@akiezun As long as you handle the changes to gatk-protected....

akiezun:
yup

PR #1847, opened 2016-05-20
Add dependency on jbwa snapshot, script to release future jbwa snapshots, and method to load the library at runtime

### LINKED ISSUE #1838: Publish jbwa to Broad artifactory
Artifact should contain builds of the library for both Mac and Linux

droazen:
-Published a jbwa snapshot to the Broad artifactory, which we now depend on
 via gradle. This snapshot contains builds of the native jbwa code for both
 Mac and Linux.

-Added utility methods to NativeUtils to load this library at runtime,
 and a test proving that it can be loaded successfully. Also switched
 to the new NativeUtils methods for loading the PairHMM, and confirmed
 that it loads successfully with the HaplotypeCaller in protected

-Included a gradle script to publish a new jbwa snapshot, should it
 become necessary, along with instructions on how to do so.

Resolves #1838

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -148,6 +148,8 @@ dependencies {
         exclude module: 'kryo' // use Spark's version
     }
 
+    compile 'com.github.lindenb:jbwa:bd72b489f9697a857671533c86ad5906967f8672-20160521.042635-1'
+
     //needed for DataflowAssert
     testCompile 'org.hamcrest:hamcrest-all:1.3'
     testCompile 'junit:junit:4.12'
--- a/src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
@@ -1,0 +1,60 @@
+package org.broadinstitute.hellbender;
+
+import com.github.lindenb.jbwa.jni.AlnRgn;
+import com.github.lindenb.jbwa.jni.BwaIndex;
+import com.github.lindenb.jbwa.jni.BwaMem;
+import com.github.lindenb.jbwa.jni.ShortRead;
+import org.broadinstitute.hellbender.utils.NativeUtils;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.SkipException;
+import org.testng.annotations.Test;
+
+import java.io.File;
+
+public class JBWAIntegrationTest extends BaseTest {
+
+    private void skipJBWATestOnUnsupportedPlatforms() {
+        if ( ! NativeUtils.runningOnLinux() && ! NativeUtils.runningOnMac() ) {
+            throw new SkipException("jbwa not available on this platform");
+        }
+        if ( NativeUtils.runningOnPPCArchitecture() ) {
+            throw new SkipException("jbwa not available for this architecture");
+        }
+    }
+
+    @Test
+    public void testJBWAIsLoadable(){
+        skipJBWATestOnUnsupportedPlatforms();
+
+        final String libraryPath = NativeUtils.runningOnLinux() ? "/lib/libbwajni.so" : "/lib/libbwajni_mac.jnilib";
+        Assert.assertTrue(NativeUtils.loadLibraryFromClasspath(libraryPath), "jbwa library was not loaded. " +
+                "This could be due to a configuration error, or your system might not support it.");
+    }
+
+
+    @Test
+    public void testJBWAAlignSingleRead() throws Exception {
+        skipJBWATestOnUnsupportedPlatforms();
+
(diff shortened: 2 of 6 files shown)


droazen:
@akiezun or @tomwhite please review

build.gradle
```diff
@@ -148,6 +148,8 @@ dependencies {
         exclude module: 'kryo' // use Spark's version
     }
 
+    compile 'org.broadinstitute:jbwa:bd72b489f9697a857671533c86ad5906967f8672-20160520.183811-1'
```
droazen: Note that we are temporarily using an "org.broadinstitute" domain for the jbwa snapshots, but we should switch to a domain controlled by the library author when we start publishing jbwa to maven central.

akiezun:
Here's some code you can add to the test to check that jbwa actually works
```
    @Test
    public void testBwaMemAlignSingleRead() throws Exception {
        final String libraryPath = NativeUtils.runningOnLinux() ? "/lib/libbwajni.so" : "/lib/libbwajni_mac.jnilib";
        Assert.assertTrue(NativeUtils.loadLibraryFromClasspath(libraryPath), "jbwa library was not loaded. " +
                "This could be due to a configuration error, or your system might not support it.");

        BwaIndex index= new BwaIndex(new File(b37_reference_20_21));
        final BwaMem bwaMem = new BwaMem(index);
        //real read taken from src/test/resources/large/CEUTrio.HiSeq.WGS.b37.NA12878.20.21.bam
        final String name="20FUKAAXX100202:3:46:9213:168594";
        final byte[] seqs= "GTTTTGTTTACTACAGCTTTGTAGTAAATTTTGAACTCTAAAGTGTTAGTTCTCTAACTTTGTTTGTTTTTCAAGAGTGTTTTGACTCTTCTTACTGCATC".getBytes(); ;
        final byte[] quals= "DGFDGFDHFFFFGFEFHEGFFFGGHEHFHGHHGEGGGGGFGFHGHGHEHGGGFGAEFGDACAHHDHGCGFGGGFGDHGHFFHDDCGGDGEE".getBytes();
        final ShortRead read = new ShortRead(name, seqs, quals);
        final AlnRgn[] align = bwaMem.align(read);
        Assert.assertEquals(align.length, 1);
        Assert.assertEquals(align[0].getChrom(), "20");
        Assert.assertEquals(align[0].getCigar(), "101M");
        Assert.assertEquals(align[0].getMQual(), 60);
        Assert.assertEquals(align[0].getPos(), 9999997-1); #note difference from the bam file (9999997 in bam, 9999996 here)
        Assert.assertEquals(align[0].getNm(), 0);
        bwaMem.dispose();
    }
```

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.testng.Assert;
+import org.testng.SkipException;
+import org.testng.annotations.Test;
+
+public class JBWAIntegrationTest extends BaseTest {
+
+    @Test
+    public void testJBWAIsLoadable(){
+        if ( ! NativeUtils.runningOnLinux() && ! NativeUtils.runningOnMac() ) {
+            throw new SkipException("jbwa not available on this platform");
```
akiezun: what does this have to do with jbwa? this seems to be a generic check not related to jbwa

akiezun:
looks good - please add 1 test to confirm that jbwa is working and clarify what the first 2 ifs in `testJBWAIsLoadable` have to do with jbwa. BWT I tested on the mac and on linux and it works on both.

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        if ( NativeUtils.runningOnPPCArchitecture() ) {
+            throw new SkipException("jbwa not available for this architecture");
```
akiezun: this too - does it have anything to do with jbwa itself? if yes, then how?

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
akiezun: oh, i see. nevermind

droazen:
I've added the additional test you requested, and confirmed that it passes.

The `SkipExceptions` are there to skip JBWA tests on platforms for which we don't have a build of the library -- I've extracted a `skipJBWATestOnUnsupportedPlatforms()` method to make this clearer.

build.gradle
droazen: Changed to `com.github.lindenb`, and uploaded the same snapshot with the updated groupID.

PR #1849, opened 2016-05-20
Move SparkTestUtils to utils/test so that gatk-protected can use it

droazen:
SparkTestUtils contains a useful method that tests whether a class is
serializable/deserializable in Kryo. This method is needed by the
gatk-protected test suite, but is not currently packaged in the gatk-public
jar. Moving to utils/test so that it will be packaged.


droazen:
@akiezun please review

PR #1851, opened 2016-05-24
Update RecalDatum.java

faparpia:
Nudge estimatedQReported to the nearest integer if the departure from this integer is within (0., 1.e-12).  The size of this interval is somewhat arbitrary.


yhcheng:
Hi David,
I think the test failed because the "expected result" is generated from GATK3 that does not have the rounding error tolerance. If GATK3 code is modified with the same change, then the result should be the same.

droazen:
Yes, the test failures are expected.

droazen:
Since the GATK4 version of BQSR is being used in production at the Broad, and this change introduces a difference vs. GATK3 output, we need to check whether Broad production is ok with this change before we can merge.

yhcheng:
Hi @droazen, any progress on checking with Broad production? Thanks!

droazen:
Reminder to self to resolve this before the IBM meeting on 7/21

PR #1853, opened 2016-05-24
Fixes null pointer exception in OverhangFixingManger

meganshand:
which is used by SplitNCigarReads.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/SplitNCigarReads.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/SplitNCigarReads.java
@@ -117,7 +117,9 @@ public final class SplitNCigarReads extends CommandLineProgram {
 
     private SAMFileGATKReadWriter initialize(final SamReader in) {
         final SAMFileHeader outputHeader = ReadUtils.cloneSAMFileHeader(in.getFileHeader());
-        final SAMFileGATKReadWriter outputWriter = new SAMFileGATKReadWriter(ReadUtils.createCommonSAMWriter(OUTPUT, REFERENCE_SEQUENCE, outputHeader, true, false, false));
+        //preSorted is set to false for this writer since ReadCoordinateComparator doesn't match HTSJDK's when an unmapped read has a mapped mate.
+        //Also, we don't want to assume that reads will be written in order by the manager because in deep, deep pileups it won't work
+        final SAMFileGATKReadWriter outputWriter = new SAMFileGATKReadWriter(ReadUtils.createCommonSAMWriter(OUTPUT, REFERENCE_SEQUENCE, outputHeader, false, false, false));
 
         try {
             final IndexedFastaSequenceFile referenceReader = new CachingIndexedFastaSequenceFile(REFERENCE_SEQUENCE);
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
@@ -158,7 +158,7 @@ public class OverhangFixingManager {
 
         // if the new read is on a different contig or we have too many reads, then we need to flush the queue and clear the map
         final boolean tooManyReads = getNReadsInQueue() >= MAX_RECORDS_IN_MEMORY;
-        final boolean encounteredNewContig = getNReadsInQueue() > 0 && !waitingReads.peek().read.getContig().equals(read.getContig());
+        final boolean encounteredNewContig = getNReadsInQueue() > 0 && !waitingReads.peek().read.isUnmapped() && !waitingReads.peek().read.getContig().equals(read.getContig());
 
         if ( tooManyReads || encounteredNewContig ) {
             if ( DEBUG ) {
--- a/src/test/java/org/broadinstitute/hellbender/tools/SplitNCigarReadsIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/SplitNCigarReadsIntegrationTest.java
@@ -53,4 +53,12 @@ public final class SplitNCigarReadsIntegrationTest extends CommandLineProgramTes
                 Arrays.asList(getTestDataDir() +"/" + "expected.splitNCigarReadsSnippet.splitNcigarReads.fixNDN.bam"));
         spec.executeTest("test fix NDN", this);
     }
+
+    @Test //regression test for https://github.com/broadinstitute/gatk/pull/1853
+    public void testSplitsOfUnpairedAndUnmappedReads() throws Exception {
+        IntegrationTestSpec spec = new IntegrationTestSpec(
+                "-R" + b37_reference_20_21 + " -I " + largeFileTestDir + "K-562.duplicateMarked.chr20.bam -O %s",
+                Arrays.asList(largeFileTestDir + "expected.K-562.splitNCigarReads.chr20.bam")); //results created using gatk3.5
+        spec.executeTest("regression test for unmapped and unpaired reads", this);
+    }
 }
--- a/src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManagerUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManagerUnitTest.java
@@ -119,17 +119,34 @@ public final class OverhangFixingManagerUnitTest extends BaseTest {
         Assert.assertEquals(manager.overhangingBasesMismatch(read, readStart, ref, refStart, overhang), expected, new String(read) + " vs. " + new String(ref) + " @" + overhang);
     }
 
-    @Test
-    public void testUnmappedReadsDoNotFail() {
-        // create an unmapped read
+    @DataProvider(name = "unmappedRead")
+    public Object[][] makeUnmappedRead() {
+        final List<Object[]> tests = new ArrayList<>();
         final GATKRead read = ArtificialReadUtils.createRandomRead(100);
         read.setName("foo");
         read.setCigar("*");
         read.setIsUnmapped();
 
-        // try to add it to the manager
+        tests.add(new Object[]{read});
+
+        return tests.toArray(new Object[][]{});
+    }
+
+    @Test(dataProvider = "unmappedRead")
+     public void testUnmappedReadsDoNotFail(final GATKRead read) {
+        // try to add unmapped read to the manager
         final OverhangFixingManager manager = new OverhangFixingManager(getHG19Header(), null, null, null, 100, 1, 30, false);
         manager.addRead(read); // we just want to make sure that the following call does not fail
-        Assert.assertTrue(true);
+    }
+
+
+   /* This tests that if we peek in the waiting reads (which occurs if there are more than one read added) we don't get
+    an NPE from getContig() of an unmapped read.*/
+    @Test(dataProvider = "unmappedRead")
+    public void testTwoUnmappedReadsDoNotFail(final GATKRead read) {
+        // try to add unmapped read to the manager twice
+        final OverhangFixingManager manager = new OverhangFixingManager(getHG19Header(), null, null, null, 100, 1, 30, false);
+        manager.addRead(read); // we need to check when there is an unmapped read waiting to be added (so we need to add two)
+        manager.addRead(read); // we just want to make sure that the following call does not fail
     }
 }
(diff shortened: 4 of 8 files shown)


meganshand:
@droazen Could you please suggest a reviewer?

akiezun:
is this is actual NPE that surfaces when running this tool? if yes, can we have a regression test for it (ie make a test with an actual small input that blows up)

vdauwera:
And it would be lovely to port the fix to GATK3 assuming it's similarly
affected.
On Tue, May 24, 2016 at 10:19 PM Adam Kiezun <notifications@github.com>
wrote:

> is this is actual NPE that surfaces when running this tool? if yes, can we
> have a regression test for it (ie make a test with an actual small input
> that blows up)
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/broadinstitute/gatk/pull/1853#issuecomment-221456201>
>
-- 
Geraldine A. Van der Auwera, PhD
Manager of Outreach and Communications
Data Sciences and Data Engineering
Broad Institute

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -158,7 +158,7 @@ public void addRead(final GATKRead read) {
 
         // if the new read is on a different contig or we have too many reads, then we need to flush the queue and clear the map
         final boolean tooManyReads = getNReadsInQueue() >= MAX_RECORDS_IN_MEMORY;
-        final boolean encounteredNewContig = getNReadsInQueue() > 0 && !waitingReads.peek().read.getContig().equals(read.getContig());
+        final boolean encounteredNewContig = getNReadsInQueue() > 0 && !waitingReads.peek().read.isUnmapped() && !waitingReads.peek().read.getContig().equals(read.getContig());
```
droazen: Can you please add a unit test for `OverhangFixingManager` that fails before this fix and passes after it?

droazen:
Review complete, back to @meganshand

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
meganshand: The unit test that I changed fails before the fix and passes after it. Should I make it a separate test so that the old test is also preserved?

meganshand:
@akiezun Yes this did actually throw a NPE when I ran SplitNCigarReads with a small file. That said, when I tried to run it again with this change I still get an error, but a different one. I'll have to look into it more carefully to see if they are related or not.

@vdauwera I don't think this particular issue is in GATK3, however I'm planning on continuing to poke around in SplitNCigarReads so that will be something I'll keep in mind in the future.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
droazen: Yes, a targeted test for the specific bug in question (with a comment describing it) would be best -- otherwise someone might refactor the existing test one day without realizing that a line was added specifically to detect a particular failure.

meganshand: Ah, got it. Will do.

akiezun:
@meganshand can you make a regression test with this file? At the **very least** include the commandline here.  Otherwise we have a fix for an NPE that no one knows how to trigger.

meganshand:
@droazen and @akiezun I added the separate unit test as well as the (currently broken) regression test. The issue that's preventing the regression test from running is #1865.

src/test/java/org/broadinstitute/hellbender/tools/SplitNCigarReadsIntegrationTest.java
```diff
@@ -53,4 +53,15 @@ public void testSplitsFixNDN() throws Exception {
                 Arrays.asList(getTestDataDir() +"/" + "expected.splitNCigarReadsSnippet.splitNcigarReads.fixNDN.bam"));
         spec.executeTest("test fix NDN", this);
     }
+
+    /**
+     * Currently disabled due to issue: https://github.com/broadinstitute/gatk/issues/1865
+     */
+    @Test(enabled = false) //regression test for https://github.com/broadinstitute/gatk/pull/1853
```
droazen: Have you tested whether setting `presorted = false` in the writer fixes this? This is what GATK3 does, as far as I can tell, to address the sorting issue (this forces the writer to sort the reads for you before writing them out). If that change does fix this test, recommend making it here in this branch rather than introducing a disabled test.

src/test/java/org/broadinstitute/hellbender/tools/SplitNCigarReadsIntegrationTest.java
```diff
@@ -53,4 +53,15 @@ public void testSplitsFixNDN() throws Exception {
...
+    public void testSplitsOfUnpairedAndUnmappedReads() throws Exception {
+        IntegrationTestSpec spec = new IntegrationTestSpec(
+                "-R" + b37_reference_20_21 + " -I " + largeFileTestDir + "K-562.duplicateMarked.chr20.bam -O %s",
```
droazen: How large are the new large files you've introduced?

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManagerUnitTest.java
```diff
@@ -119,16 +119,35 @@ public void testMismatch(final byte[] read, final int readStart, final byte[] re
...
+
+        return tests.toArray(new Object[][]{});
+    }
+
+    @Test(dataProvider = "unmappedRead")
+     public void testUnmappedReadsDoNotFail(final GATKRead read) {
+        // try to add unmapped read to the manager
+        final OverhangFixingManager manager = new OverhangFixingManager(getHG19Header(), null, null, null, 100, 1, 30, false);
+        manager.addRead(read); // we just want to make sure that the following call does not fail
+        Assert.assertTrue(true);
```
droazen: No need for this final assert

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManagerUnitTest.java
```diff
@@ -119,16 +119,35 @@ public void testMismatch(final byte[] read, final int readStart, final byte[] re
...
+
+   /* This tests that if we peek in the waiting reads (which occurs if there are more than one read added) we don't get
+    an NPE from getContig() of an unmapped read.*/
+    @Test(dataProvider = "unmappedRead")
+    public void testTwoUnmappedReadsDoNotFail(final GATKRead read) {
+        // try to add unmapped read to the manager twice
         final OverhangFixingManager manager = new OverhangFixingManager(getHG19Header(), null, null, null, 100, 1, 30, false);
+        manager.addRead(read); // we need to check when there is an unmapped read waiting to be added (so we need to add two)
         manager.addRead(read); // we just want to make sure that the following call does not fail
         Assert.assertTrue(true);
```
droazen: Unnecessary assert

droazen:
Second-pass review complete, back to @meganshand. Main recommendation is to include the `presorted = false` fix in this branch to avoid checking in a disabled test.

src/test/java/org/broadinstitute/hellbender/tools/SplitNCigarReadsIntegrationTest.java
lbergelson: @droazen You can check in the github interface by clicking view on them if you're interested.
@meganshand Shouldn't there be a committed update to src/test/resources/large/.gitattributes to account for the new large files?

meganshand: @droazen They are both about 1 MB (plus a small index file). Should I move them out of the "large" directory?
@lbergelson Yes, I'll add them to .gitattributes.

droazen: @meganshand 1 MB is about the threshold for putting a file into `large` and having it managed by `git lfs`, so I think it's fine to keep them here.

src/test/java/org/broadinstitute/hellbender/tools/SplitNCigarReadsIntegrationTest.java
meganshand: This does fix it. However, is there a reason not to also change ReadCoordinateComparator::compareCoordinates to take the case when an unmapped read's mate is mapped? That would also fix this particular problem.

meganshand:
@droazen I've made the changes. I just have a question about why ReadCoordinateComparator::compareCoordinates doesn't take the case when an unmapped read's mate is mapped into account.

droazen:
@meganshand There is a warning in the docs for `ReadCoordinateComparator` that it should not be used for bam file output that needs to match the ordering of `SAMRecordCoordinateComparator` exactly, since it sorts all unmapped reads after all mapped reads. `ReadCoordinateComparator` is a comparator for `GATKRead`, and that interface does not allow unmapped reads to have a position. Ie., even if an unmapped `SAMRecord` is assigned the position of its mapped mate, calling `getContig()`/`getStart()` on the unmapped read via the `GATKRead` interface will return null/0. This was done mainly for consistency reasons and to simplify client code. Whenever we need bam file order for reads in GATK4, we operate on SAMRecords directly and use either the `SAMRecordCoordinateComparator` from htsjdk or the `HeaderlessSAMRecordCoordinateComparator` (for headerless Spark reads) that produces the same ordering.

I recommend addressing this for this tool via `presorted = false` for now, since the GATK3 version has it set to false as well with the comment: "we don't want to assume that reads will be written in order by the manager because in deep, deep pileups it won't work". This suggests that even if you were to change the comparator used by this tool to behave like `SAMRecordCoordinateComparator`, you'd still have ordering issues in deep coverage areas.

It's worthwhile, though, to open a separate ticket to explore whether `ReadCoordinateComparator` could be changed to exactly match bam file order. Eg., perhaps we could add `getAssignedContig()`, `getAssignedStart()`, etc. methods to `GATKRead` to expose the positions that unmapped reads with mapped mates get assigned for sorting purposes.

meganshand:
@droazen Thank you for the thorough explanation. That makes sense. For now I've changed `presorted = false` and that integration test is now passing (meaning the output now matches GATK3). Back to you!

src/test/resources/large/.gitattributes
```diff
@@ -23,3 +23,5 @@ human_g1k_v37.20.21.fasta.pac filter=lfs diff=lfs merge=lfs -text
 human_g1k_v37.20.21.fasta.ann filter=lfs diff=lfs merge=lfs -text
 human_g1k_v37.20.21.fasta.amb filter=lfs diff=lfs merge=lfs -text
 human_g1k_v37.20.21.fasta.sa filter=lfs diff=lfs merge=lfs -text
+K-562.duplicateMarked.chr20.bam filter=lfs diff=lfs merge=lfs -text
```
droazen: Looks like you're still missing the `bai` here

src/main/java/org/broadinstitute/hellbender/tools/SplitNCigarReads.java
```diff
@@ -117,7 +117,7 @@ protected Object doWork() {
 
     private SAMFileGATKReadWriter initialize(final SamReader in) {
         final SAMFileHeader outputHeader = ReadUtils.cloneSAMFileHeader(in.getFileHeader());
-        final SAMFileGATKReadWriter outputWriter = new SAMFileGATKReadWriter(ReadUtils.createCommonSAMWriter(OUTPUT, REFERENCE_SEQUENCE, outputHeader, true, false, false));
+        final SAMFileGATKReadWriter outputWriter = new SAMFileGATKReadWriter(ReadUtils.createCommonSAMWriter(OUTPUT, REFERENCE_SEQUENCE, outputHeader, false, false, false));
```
droazen: Add a comment explaining why `preSorted` is set to false here, and preserve the warning about deep coverage from GATK3: `"we don't want to assume that reads will be written in order by the manager because in deep, deep pileups it won't work"`

droazen:
@meganshand Two remaining comments to address, then please rebase and squash into a single commit and we'll merge.

meganshand:
@droazen Addressed your comments, squashed and rebased. Ready to merge?

droazen:
:+1: merging

PR #1859, opened 2016-05-26
exclude signature files from spark jar

akiezun:
those have to be excluded or we get "Invalid signature file digest for Manifest main attributes"
I'm not sure how come this only came up now

@droazen can you look?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -200,8 +200,8 @@ tasks.withType(Jar) {
                 'Implementation-Version': version,
                 'Main-Class': project.mainClassName
     }
+    zip64 true
 }
-
 
 test {
     outputs.upToDateWhen { false }  //tests will never be "up to date" so you can always rerun them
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -317,6 +317,11 @@ task sparkJar(type: ShadowJar) {
     description = "Create a combined jar of project and runtime dependencies that excludes provided spark dependencies"
     configurations = [project.configurations.sparkConfiguration]
     classifier = 'spark'
+
+    //Those have to be excluded to prevent "Invalid signature file digest for Manifest main attributes"
+    exclude 'META-INF/*.SF'
+    exclude 'META-INF/*.DSA'
+    exclude 'META-INF/*.RSA'
```
droazen: Can you explain why this is an issue only for the spark jar (and not, say, the shadow jar)?

droazen:
Back to @akiezun to answer a question

build.gradle
akiezun: i have no idea. It's probably because the spark jar is what gets used when running on a cluster and the cluster complains (it did't use to but it does now)

akiezun:
back to @droazen.

akiezun:
nevermind. this may be also fixed by excluding some dependencies. let's close for now, may reopen

PR #1860, opened 2016-05-26
New tool to dump evidence of SVs.

tedsharpe:


droazen:
@tedsharpe Can you nominate a reviewer for this branch?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -0,0 +1,476 @@
...
+import com.esotericsoftware.kryo.DefaultSerializer;
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import org.apache.spark.serializer.KryoRegistrator;
+import org.broadinstitute.hellbender.exceptions.GATKException;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+/**
+ * Various types of read anomalies that provide evidence of genomic breakpoints.
```
cwhelan: I think I'm understanding this more and more as I read through the code, but it would be helpful to expand this comment to give a little more information about the data held by this object.. Ie. each instance of BreakpointEvidence essentially maps to 1 read and/or read pair, etc.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -0,0 +1,476 @@
...
+ */
+public class BreakpointEvidence implements Comparable<BreakpointEvidence> {
+    private final short contigIndex;
+    private final short eventWidth;
+    private final int contigStart;
```
cwhelan: Is contigStart just the start position of the evidence interval? I find the name a little confusing.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -0,0 +1,476 @@
...
+    private final String templateName;
+    private final TemplateEnd templateEnd;
+
+    public enum TemplateEnd {
+        UNPAIRED(""), PAIRED_UNKNOWN("/?"), PAIRED_FIRST("/1"), PAIRED_SECOND("/2"), PAIRED_INTERIOR("/0");
```
cwhelan: What's a PAIRED_INTERIOR read? Is that something we actually see or just defensive code?

Would "ReadPositionInTemplate" maybe be decent alternative name to "TemplateEnd"?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -0,0 +1,476 @@
...
+    protected BreakpointEvidence( final Kryo kryo, final Input input ) {
+        this.contigIndex = input.readShort();
+        this.eventWidth = input.readShort();
+        this.contigStart = input.readInt();
+        this.templateName = kryo.readObject(input, String.class);
+        this.templateEnd = TemplateEnd.values()[input.readByte()];
+    }
+
+    /**
+     * to make a sentinel
```
cwhelan: Would you mind slightly expanding this comment to explain what this is for?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -0,0 +1,476 @@
...
+                return new SameStrandPair(kryo, input);
+            }
+        }
+    }
+
+    @DefaultSerializer(WeirdTemplateSize.Serializer.class)
+    public static final class WeirdTemplateSize extends BreakpointEvidence {
+        private final int templateSize;
+
+        WeirdTemplateSize( final GATKRead read, final ReadMetadata metadata ) {
```
cwhelan: It looks like WeirdTempleSize is not currently hooked up right now (commented out code in `ReadClassifier.checkDiscordantPair()`). Why is that? Also, it seems like if the `WeirdTemplateSize` is _less_ than the expected insert size, we could use a much shorter evidence interval, so you might want to break out those two cases.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -0,0 +1,476 @@
...
+            @Override
+            public MateUnmapped read( final Kryo kryo, final Input input, final Class<MateUnmapped> klass ) {
+                return new MateUnmapped(kryo, input);
+            }
+        }
+    }
+
+    @DefaultSerializer(InterContigPair.Serializer.class)
+    public static final class InterContigPair extends BreakpointEvidence {
+        private final int otherContigIndex;
```
cwhelan: Eventually we'll want to track the actual alignment position of the other read, and possibly some other information, but I guess that can wait for another day.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
tedsharpe: yes.  maybe eventStartPosition?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
tedsharpe: Defensive.  A read that's flagged neither first nor last must be somewhere in the interior of the template.  SAM spec makes it possible.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
tedsharpe: Will do.  But just to clear up the mystery, this constructor is called by the related serializer during deserialization.  I think this is a nice paradigm for getting around the Kryo issues:  every Kryo serializable specifies its own implementation of kryo.Serializer via an annotation (registration will happen automatically), and that implementation of kryo.Serializer is just a few lines of boilerplate that delegates to a serialize( Kryo, Output ) method in the object to serialize and to a constructor <init>( Kryo, Input ) to deserialize.  No magic reflection.  No end runs around access controls.  Just regular old calls to regular old Java methods.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
tedsharpe: Yes.  The information was a dilute soup, and despite efforts to only pick up extremes, the template size evidence was contributing too many intervals all by itself.  We can turn this code back on if we put a weight on BreakpointEvidence and give these guys a lesser weight, or put a rule in place that prohibits calling intervals that are supported only by template size evidence, or both.  Need something a little more sophisticated to pull this in.  (Or maybe it's not needed.)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
tedsharpe: I can put that in.  Easy.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
cwhelan: I think we'll definitely eventually want to turn this on, as it's the primary signal for calling larger deletions (at least when they don't have a split read signal). We can experiment with weighting and/or filtering reads if it's generating too many intervals right now.

tedsharpe: Agreed.  This was simple expediency.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
cwhelan: Ah OK. It's good to be defensive, although I think in other corners of the tool we are assuming paired end Illumina 'innie' reads (ie we gather evidence from 'outies' which only makes sense given that assumption) so I wanted to make sure I understood what was going on.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
cwhelan: Agreed that we don't need to fix it right now for this PR but it'd be good if you could put in a TODO comment at the place where it's disabled and create a github issue (you can label it "SV") saying it should be turned on eventually. And if you happen to have any numbers about the problem or other thoughts you could throw them in the ticket.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+/**
+ * Tool to describe reads that support a hypothesis of a genomic breakpoint.
+ */
+@CommandLineProgramProperties(summary="Find reads that evidence breakpoints.",
+        oneLineSummary="Dump FASTQs for local assembly of putative genomic breakpoints.",
+        programGroup = SparkProgramGroup.class)
+public final class FindBreakpointEvidenceSpark extends GATKSparkTool {
+    private static final long serialVersionUID = 1L;
+    private static final int MIN_MAPQ = 20;
+    private static final int MIN_MATCH_LEN = 45; // minimum length of matched portion of an interesting alignment
```
cwhelan: Just a thought but would it make sense to tie this to the kmer size we're using?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+    //private static final int MAX_FRAGMENT_LEN = 2000;
+    private static final int MAX_COVERAGE = 1000;
+    private static final float ASSEMBLY_TO_MAPPED_SIZE_RATIO = 7.f;
```
cwhelan: Could you document what these constants are? Especially this one and the next one. :)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Seems to me like it probably depends more on the read length.  We could make it 1/3 of the read length.  Just want to reject flaky alignments that don't really localize the read.  It's possible that MAPQ>=20 already does this, but I think I stuck it in there for a reason.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Just an empirical number to size a hashmap to a reasonable capacity for the final assembly relative to the number of reads mapped there in an earlier pass.  I've added documentation.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+                new ReadMetadata.ReadGroupFragmentStatistics(400.f, 75.f);
+        final List<ReadMetadata.ReadGroupFragmentStatistics> stats = new ArrayList<>(nGroups);
+        for ( int idx = 0; idx != nGroups; ++idx ) {
+            stats.add(groupStats);
+        }
+        return new ReadMetadata(header, stats, groupStats, meanBasesPerTemplate);
+    }
+
+    /** mean number of sequenced bases per template (i.e., read length if unpaired, twice the read length if paired) */
+    @VisibleForTesting static int getMeanBasesPerTemplate(final JavaRDD<GATKRead> reads ) {
```
cwhelan: Do you think it's really necessary to run over all the reads to get the mean read length? It seems a bit heavy. Did you measure how expensive it is?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+    protected void runTool( final JavaSparkContext ctx ) {
+        if ( getHeaderForReads().getSortOrder() != SAMFileHeader.SortOrder.coordinate ) {
+            throw new GATKException("The reads must be coordinate sorted.");
+        }
+
+        final Locations locations =
+                new Locations(evidenceDir, intervalFile, qNamesMappedFile, kmerFile, qNamesAssemblyFile);
+        final PipelineOptions pipelineOptions = getAuthenticatedGCSOptions();
+        final SAMFileHeader header = getHeaderForReads();
+        final JavaRDD<GATKRead> reads = getUnfilteredReads();
```
cwhelan: Aren't there some filters we want to apply in all cases? I.e. non-pass-filter reads, maybe others.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+        final HopscotchHashSet<QNameAndInterval> qNamesSet;
+        final int nIntervals;
+        {
```
cwhelan: Why these code blocks? If they are to limit scope would it make more sense to extract the contents into their own methods?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+
+    private static List<byte[]> combineLists( final List<byte[]> list1, final List<byte[]> list2 ) {
+        final List<byte[]> result = new ArrayList<>(list1.size() + list2.size());
+        result.addAll(list1);
+        result.addAll(list2);
+        return result;
+    }
+
+    @VisibleForTesting static byte[][] sortFastqRecs( final List<byte[]> fastqRecs ) {
+        final int nFastqs = SVUtils.iteratorSize(fastqRecs.iterator());
```
cwhelan: why not 'fastqRecs.size()`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+    private static final class QNameFinder implements Function<GATKRead, Iterator<QNameAndInterval>> {
+        private final int maxQnamesPerInterval;
+        private final ReadMetadata metadata;
+        private final List<Interval> intervals;
+        private final int[] qNamesPerInterval;
+        private final Iterator<QNameAndInterval> noName = Collections.emptyIterator();
+        private int intervalsIndex = 0;
+
+        QNameFinder( final int maxQnamesPerInterval, final ReadMetadata metadata, final List<Interval> intervals ) {
+            this.maxQnamesPerInterval = maxQnamesPerInterval;
```
cwhelan: Wouldn't it be possible to precompute an individual maxQNames for each interval? You might save some time collecting reads you don't want.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
cwhelan: There are definitely examples of cases where you can get a high MAPQ on a weak but unique alignment. We encountered this with the bacterial contamination issue. We use OverclippedReadFilter to filter these guys out -- we could apply that here if we want to.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
cwhelan: OK. The same comment would go for some of these other types of evidence -- eg for SplitRead if we could track the SA alternate alignment tags that would be useful for merging breakpoints.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+import org.apache.spark.broadcast.Broadcast;
+import org.broadinstitute.hellbender.cmdline.Argument;
+import org.broadinstitute.hellbender.cmdline.CommandLineProgramProperties;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+import org.broadinstitute.hellbender.cmdline.programgroups.SparkProgramGroup;
+import org.broadinstitute.hellbender.engine.spark.GATKSparkTool;
+import org.broadinstitute.hellbender.exceptions.GATKException;
+import org.broadinstitute.hellbender.tools.spark.utils.HopscotchHashSet;
+import org.broadinstitute.hellbender.tools.spark.utils.MapPartitioner;
+import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
```
cwhelan: I had the same comment for Steve on some of his stuff -- BucketUtils sounds like something specific to GCS. It might be nice to refactor out the stuff we're using (which has nothing to do with GCS) into a different class.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+        for ( int idx = 0; idx != len; ++idx ) {
+            final int result = Integer.compare(arr1[idx] & 0xff, arr2[idx] & 0xff);
+            if ( result != 0 ) return result;
+        }
+        return Integer.compare(arr1.length, arr2.length);
+    }
+
+    /**
+     * Process the input reads, pulling all reads that contain kmers associated with a given breakpoint,
+     * and writing those reads into a separate FASTQ for each breakpoint.
```
cwhelan: Seems like the writing FASTQs happens elsewhere so maybe the comment's out of date.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+    }
+
+    @VisibleForTesting static List<SVKmer> getHighCountKmers( final JavaRDD<GATKRead> reads ) {
+        final int nPartitions = reads.partitions().size();
+        return reads
+                .filter(read ->
+                        !read.isSecondaryAlignment() && !read.isSupplementaryAlignment() &&
+                                !read.isDuplicate() && !read.failsVendorQualityCheck())
+                .mapPartitions(KmerCounter::new, false)
+                .mapToPair(kmerAndCount -> new Tuple2<>(new SVKmer(kmerAndCount), kmerAndCount.getCount()))
```
cwhelan: Would it be more efficient to combine the `mapPartitions` and `mapToPair` into a `mapPartitionsToPair` call?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+            final JavaSparkContext ctx,
+            final HopscotchHashSet<KmerAndInterval> kmerAndIntervalsSet,
+            final JavaRDD<GATKRead> reads ) {
+        final Broadcast<HopscotchHashSet<KmerAndInterval>> broadcastKmerAndIntervalsSet =
+                ctx.broadcast(kmerAndIntervalsSet);
+
+        final List<QNameAndInterval> qNames =
+            reads
+                .filter(read ->
+                        !read.isSecondaryAlignment() && !read.isSupplementaryAlignment() &&
```
cwhelan: This set of filters gets used a few times. It'd be good to extract them into a single filter class and reuse it, I think.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+    /** mean number of sequenced bases per template (i.e., read length if unpaired, twice the read length if paired) */
+    @VisibleForTesting static int getMeanBasesPerTemplate(final JavaRDD<GATKRead> reads ) {
+        return reads
+                .map(ReadCountAndLength::new)
+                .reduce(ReadCountAndLength::new)
+                .getMeanLength();
+    }
+
+    private static void log( final String message ) {
+        System.out.println(dateFormatter.format(System.currentTimeMillis())+message);
```
cwhelan: Can we use a log4j logger for debugging? If not it's probably best to at least turn this off with a flag or something so that we don't accidentally leave a ton of printlns in the code.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+
+        final List<QNameAndInterval> qNames =
+            reads
+                .filter(read ->
+                        !read.isSecondaryAlignment() && !read.isSupplementaryAlignment() &&
+                                !read.isDuplicate() && !read.failsVendorQualityCheck())
+                .mapPartitions(readItr ->
+                        new MapPartitioner<>(readItr,
+                                new QNamesForKmersFinder(broadcastKmerAndIntervalsSet.value())), false)
+                .collect();
```
cwhelan: There are a bunch of RDD operations that end in `collect()`. Since it's working I don't want to re-architect the flow in this PR, but I wonder if it would be possible to speed things up by leaving some of these things as RDDs and then doing things like join operations between them. Let's brainstorm in person.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
droazen: Just chiming in here: yes, please do use log4j with an appropriate logging level (info, debug, etc.) wherever logging output is required, instead of naked println statements.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+        if ( getHeaderForReads().getSortOrder() != SAMFileHeader.SortOrder.coordinate ) {
+            throw new GATKException("The reads must be coordinate sorted.");
+        }
+
+        final Locations locations =
+                new Locations(evidenceDir, intervalFile, qNamesMappedFile, kmerFile, qNamesAssemblyFile);
+        final PipelineOptions pipelineOptions = getAuthenticatedGCSOptions();
+        final SAMFileHeader header = getHeaderForReads();
+        final JavaRDD<GATKRead> reads = getUnfilteredReads();
+        final HopscotchHashSet<QNameAndInterval> qNamesSet;
```
cwhelan: Wouldn't it be more memory-efficient to use a `Map<Interval, HopscotchHashSet<Qname>>`. This denormalization looks a little strange. At least add a comment describing why it is structured this way.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchHashSet.java
```diff
@@ -0,0 +1,542 @@
...
+        this.buckets = (T[])new Object[this.capacity];
+        this.status = new byte[this.capacity];
+    }
+
+    public HopscotchHashSet( final Collection<T> collection ) {
+        this(collection.size());
+        addAll(collection);
+    }
+
+    public HopscotchHashSet( final Collection<T> collection, final float sizeInflator ) {
```
cwhelan: Could you add a comment describing what the use case for sizeInflator is?

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchHashSet.java
```diff
@@ -0,0 +1,542 @@
...
+
+    @Override
+    public Iterator<T> iterator() { return new CompleteIterator(); }
+
+    /**
+     * This special iterator type allows you to traverse individual hash buckets.
+     * This lets you implement a multi-map on this collection by using a <K,V>-ish element type having a hashCode that
+     * depends only on K.
+     */
+    public Iterator<T> bucketIterator( final int hashVal ) { return new BucketIterator(hashVal); }
```
cwhelan: I think it would make things more clear if you made this method (or a sister method) take an argument of type `T` and compute the hash within this object.

cwhelan: To expand on that it feels like since you're essentially hacking this HashSet to act like a HashMap with the use of BucketIterators, it might be nice to actually just provide methods that let you use it like a HashMap explicitly.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
cwhelan: Now that I see how you're using BucketIterators this I understand this a little better, but I think a comment would be still be nice (or, as I suggest below, provide a version of the hash set with API calls that make the map-like behavior explicit).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+        return result;
+    }
+
+    @VisibleForTesting static byte[][] sortFastqRecs( final List<byte[]> fastqRecs ) {
+        final int nFastqs = SVUtils.iteratorSize(fastqRecs.iterator());
+        final byte[][] fastqsArray = new byte[nFastqs][];
+
+        final Iterator<byte[]> itr = fastqRecs.iterator();
+        for ( int idx = 0; idx != nFastqs; ++idx ) {
+            if ( !itr.hasNext() ) throw new GATKException("fastq iterator crapped out at element "+idx);
```
cwhelan: I know we should never hit this but probably better not to put 'crapped' in a message that could bubble out to the user. Actually, could we just get rid of the iterator and do `fastqRecs.get(i)`? Or will this be a LinkedList?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+        try ( final OutputStream writer = new BufferedOutputStream(BucketUtils.createFile(fileName, pipelineOptions)) ) {
+            for ( final byte[] fastqBytes : fastqsArray ) {
+                writer.write(fastqBytes);
+            }
+        } catch ( final IOException ioe ) {
+            throw new GATKException("Can't write "+fileName, ioe);
+        }
+    }
+
+    private static int compareByteArrays( final byte[] arr1, final byte[] arr2 ) {
```
cwhelan: Just document the assumption that the read name is first in the byte array here.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+        if ( itr.hasNext() ) throw new GATKException("unexpected extra element in fastq iterator");
+
+        // since the fastq bytes start with the read name, this will interleave pairs
+        Arrays.sort(fastqsArray, FindBreakpointEvidenceSpark::compareByteArrays);
+
+        return fastqsArray;
+    }
+
+    /** write a FASTQ file for an assembly */
+    private static void writeFastq( final Tuple2<Integer, List<byte[]>> intervalAndFastqs,
```
cwhelan: The next few methods are all about writing the fastqs, I believe, so it might be nice to extract them to their own class to make this one a little less bulky.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+
+        final Broadcast<Set<SVKmer>> broadcastKmerKillSet = ctx.broadcast(kmerKillSet);
+        final Broadcast<HopscotchHashSet<QNameAndInterval>> broadcastQNameAndIntervalsSet =
+                ctx.broadcast(qNamesSet);
+
+        // given a set of template names with interval IDs and a kill set of ubiquitous kmers,
+        // produce a set of interesting kmers for each interval ID
+        final List<KmerAndInterval> kmerIntervals =
+            reads
+            .filter(read -> !read.isDuplicate() && !read.failsVendorQualityCheck() &&
```
cwhelan: Duplicated filtering code again

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+        }
+
+        return kmerIntervals;
+    }
+
+    @VisibleForTesting static List<SVKmer> getHighCountKmers( final JavaRDD<GATKRead> reads ) {
+        final int nPartitions = reads.partitions().size();
+        return reads
+                .filter(read ->
+                        !read.isSecondaryAlignment() && !read.isSupplementaryAlignment() &&
```
cwhelan: duplicate filter

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: We'll use this pass to gather the template size statistics, too, in the not-too-distant future, so it may seem less heavy at that point.  I'm guessing that the JavaRDDLike::sample method won't be faster, but I haven't tested that.  JavaRDDLike::take seems like it could be the right thing, but the documentation seems to imply that something complicated is going on.  I could write a partition mapper that takes that first N elements from each partition.  Does that seem like a good sample?

cwhelan: Yes I think `sample` runs over the whole RDD. I might just switch to `take` and see if that speeds things up. In the vast majority of cases there will only be one read length anyway, aside from hard clipping which we shouldn't have to worry about too much.

Another option would be to take it as an (optional) parameter, or take a path to the picard metrics file that has that info and just parse it out of there.

Just trying to think of ways to cut down on the pipeline runtime.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+        }
+        return new ReadMetadata(header, stats, groupStats, meanBasesPerTemplate);
+    }
+
+    /** mean number of sequenced bases per template (i.e., read length if unpaired, twice the read length if paired) */
+    @VisibleForTesting static int getMeanBasesPerTemplate(final JavaRDD<GATKRead> reads ) {
+        return reads
+                .map(ReadCountAndLength::new)
+                .reduce(ReadCountAndLength::new)
+                .getMeanLength();
```
cwhelan: Also, you could potentially cut a lot of code by using `JavaDoubleRDD.mean`:

    reads.mapToDouble(read -> read.length()).mean()

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+
+    /** helper class for calculating mean template length */
+    @DefaultSerializer(ReadCountAndLength.Serializer.class)
+    private static final class ReadCountAndLength {
+        private final long count;
+        private final long length;
+
+        ReadCountAndLength( final GATKRead read ) {
+            this.count = 1;
+            this.length = (read.isPaired() ? 2 : 1)*read.getLength();
```
cwhelan: Won't both reads in the pair get a ReadCountAndLength fed into them, so aren't you double counting if the reads are paired?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+     */
+    private static final class BreakpointClusterer
+            implements Function<BreakpointEvidence, Iterator<BreakpointEvidence>> {
+        private final int staleEventDistance;
+        private final SortedMap<BreakpointEvidence, Boolean> locMap = new TreeMap<>();
+        private final List<Map.Entry<BreakpointEvidence, Boolean>> reportableEntries = new ArrayList<>(2*MIN_EVIDENCE);
+        private final Iterator<BreakpointEvidence> noEvidence = Collections.emptyIterator();
+        private int currentContig = -1;
+
+        private static final int MIN_EVIDENCE = 15; // minimum evidence count in a cluster
```
cwhelan: This seems like something we might want to play with so it'd be nice to turn this into a command-line parameter.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+            recordSet.add(evidence);
+            return reportableEvidence.iterator();
+        }
+    }
+
+    /**
+     * Minimalistic simple interval.
+     */
+    @DefaultSerializer(Interval.Serializer.class)
+    @VisibleForTesting static final class Interval {
```
cwhelan: What's the reason we can't use `SimpleInterval` here? They hold the same data fields.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+            public Interval read( final Kryo kryo, final Input input, final Class<Interval> klass ) {
+                return new Interval(kryo, input);
+            }
+        }
+    }
+
+    /**
+     * A class to examine a stream of BreakpointEvidence, and group it into Intervals.
+     */
+    private static final class IntervalMapper implements Function<BreakpointEvidence, Iterator<Interval>> {
```
cwhelan: The name IntervalMapper doesn't really capture what this class is doing, for me at least. Can't think of a much better name off the top of my head but even `EvidenceToIntervalMapper` would do more for me.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Sadly, the filtering varies slightly between passes, and must vary for correct behavior.  For example, when producing the FASTQ from a set of template names we want to make sure that we always get both members of each pair exactly once.  So we must filter down to primary lines to avoid dups, but don't want any quality filters or we might exclude a poor quality mate.  On the other hand, when we're looking for reads to support a breakpoint hypothesis secondary and supplementary alignment records are useful, but dups, vendor quality check failures, and unmapped reads are not.  Completely disjoint set of filtering criteria.
Nonetheless, I'll make an effort to organize the filtering a little better and avoid repetition.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
cwhelan: Sounds good to me.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+        }
+
+        @Override
+        public Iterator<KmerAndCount> iterator() { return kmerSet.iterator(); }
+    }
+
+    /**
+     * Eliminates dups, and removes over-represented kmers.
+     */
+    private static final class KmerCleaner {
```
cwhelan: Since this operates on single partitions only, it'd be helpful to note that in the comments and maybe class name. For example, as I read it, MAX_INTERVALS should really be MAX_INTERVALS_WITHIN_PARTITION (which could be a big difference).

cwhelan: Can you create a ticket suggesting that we should test and optimize the parameters you've hard-coded in this class, and maybe investigate alternative ways of filtering kmers and/or reads? I understand why you implemented it but I don't really love the class and find it hard to reason about, if that makes sense.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1160 @@
...
+    }
+
+    /**
+     * Find <intervalId,fastqBytes> pairs for interesting template names.
+     */
+    private static final class ReadsForQNamesFinder {
+        private final HopscotchHashSet<QNameAndInterval> qNamesSet;
+        private final int nIntervals;
+        private final int nReadsPerInterval;
+        private final List<Tuple2<Integer, List<byte[]>>> fastQRecords = new ArrayList<>();
```
cwhelan: Does this have to be a field? It seems to be cleared on each call to `call()` so why not just create a new one as needed in that method? It made me wonder if this was supposed to have mutable state across calls.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/KmerHistogrammer.java
```diff
@@ -0,0 +1,47 @@
...
+import org.broadinstitute.hellbender.engine.spark.GATKSparkTool;
+import org.broadinstitute.hellbender.tools.spark.utils.MapPartitioner;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import scala.Tuple2;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Tool to describe reads that support a hypothesis of a genomic breakpoint.
```
cwhelan: Can you expand on what this tool is and what it's supposed to do? Looks like it's mainly for debugging?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadClassifier.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.function.Function;
+
+// TODO: figure out a way to include funky template size data without swamping ourselves in a sea of evidence
+// TODO: maybe count each such read as fractional evidence?
+
+/**
+ * Figures out what kind of BreakpointEvidence, if any, a read represents.
+ */
+public class ReadClassifier implements Function<GATKRead, Iterator<BreakpointEvidence>> {
+    @VisibleForTesting static final int MIN_SOFT_CLIP_LEN = 30; // minimum length of an interesting soft clip
```
cwhelan: Could you make these into parameters of the command line tool, and add a ticket saying we should test different parameter values and weighting/clustering strategies?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadClassifier.java
```diff
@@ -0,0 +1,124 @@
...
+            evidenceList.add(new BreakpointEvidence.SameStrandPair(read, readMetadata));
+        } else if ( (read.getStart() < read.getMateStart() && read.isReverseStrand()) ||
+                    (read.getStart() > read.getMateStart() && !read.isReverseStrand()) ) {
+            evidenceList.add(new BreakpointEvidence.OutiesPair(read, readMetadata));
+/*      } else {
+            final ReadMetadata.ReadGroupFragmentStatistics stats = readMetadata.getStatistics(read.getReadGroup());
+            final float zIshScore = (Math.abs(read.getFragmentLength()) - stats.getMedianFragmentSize()) /
+                                        stats.getMedianFragmentSizeVariance();
+
+            if ( zIshScore < MIN_ZISH_SCORE || (zIshScore > MAX_ZISH_SCORE && zIshScore < CRAZY_ZISH_SCORE) ) {
```
cwhelan: For our future work to turn this back on, we might want to consider a) ignoring small zish scores, and b) weighting pieces of evidence more highly if they agree with other pieces of evidence w/r/t their mate mapping position for discordant pairs or their SA tag locations for split reads.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
```diff
@@ -0,0 +1,180 @@
...
+        @Override
+        public ReadMetadata read( final Kryo kryo, final Input input, final Class<ReadMetadata> klass ) {
+            return new ReadMetadata(kryo, input);
+        }
+    }
+
+    @DefaultSerializer(ReadGroupFragmentStatistics.Serializer.class)
+    public static final class ReadGroupFragmentStatistics {
+        private final float medianFragmentSize;
+        private final float medianFragmentSizeVariance;
```
cwhelan: I think we'll want to compute in terms of median absolute deviations (MADs) -- can you rename this to medianAbsoluteDeviationFragmentSize?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVConstants.java
```diff
@@ -4,7 +4,7 @@
  * Constants shared among SV tools.
  */
 public final class SVConstants {
-    private SVConstants(){}
+    private SVConstants() {}
 
-    public static final int KMER_SIZE = 63;
+    public static final int KMER_SIZE = 51;
```
cwhelan: Is it worth parameterizing this value?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVUtils.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public static int hashMapCapacity( final int nElements )
+    {
+        return (int)((nElements*4L)/3) + 1;
+    }
+
+    /**
+     * Iterator over 1 or 0 elements.
+     */
+    public static final class SingletonIterator<T> implements Iterator<T> {
```
cwhelan: Is there a need for these, instead of using `Collections.<T> emptyIterator()`, or `org.apache.commons.collections.iterators.SingletonIterator`?

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchHashSet.java
```diff
@@ -0,0 +1,542 @@
...
+
+    @Override
+    public boolean add( final T entry ) {
+        if ( entry == null ) throw new UnsupportedOperationException("This collection cannot contain null.");
+        if ( size == capacity ) resize();
+        try {
+            return insert(entry) == entry;
```
cwhelan: I don't see how `insert(entry) == entry` can be false in the current implementation of `insert`. Is the check necessary? Do we need this version of add that returns a boolean?

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchHashSet.java
```diff
@@ -0,0 +1,542 @@
...
+            status[emptyBucketIndex] = (byte) (nextOffset - toEmptyDistance);
+        }
+        buckets[emptyBucketIndex] = buckets[bucketToMoveIndex];
+        buckets[bucketToMoveIndex] = null;
+        status[bucketToMoveIndex] = 0;
+    }
+
+    @SuppressWarnings("unchecked")
+    private void resize() {
+        if ( buckets == null ) throw new IllegalStateException("OMG I have no buckets.");
```
cwhelan: This looks like it could bubble out to the user so probably best not to OMG.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchHashSet.java
```diff
@@ -0,0 +1,542 @@
...
+
+/**
+ * Hash set implementation that provides low memory overhead with a high load factor by using the hopscotch algorithm.
+ * This is usually a little slower than the JDK's HashSet (neglecting GC time), but it uses much less memory.
+ * It's probably a nice choice for very large collections.
+ * You can make it pretty much equally fast as HashSet by replacing the prime-number table sizing with power-of-2 table
+ * sizing, but then it'll behave just as badly as HashSet given crappy hashCode implementations.
+ */
+@DefaultSerializer(HopscotchHashSet.Serializer.class)
+public final class HopscotchHashSet<T> extends AbstractSet<T> {
```
cwhelan: Could you add some comments on when this will fail? There a few places where it looks like this could throw an IllegalStateException (esp. from the resize method) and I don't really understand whether they are common-ish things the caller should know about or should-never-happen type situations.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/KmerHistogrammer.java
tedsharpe: Oh, sorry.  Yes, it was for debugging.  I'll remove it.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchHashSet.java
tedsharpe: It's false when an equal entry is already present in the set.
We need the add method to complete the implementation of AbstractSet.
We only need the somewhat odd put method because of my abuse of a Set to implement a MultiMap.  (I'll try to refactor that out.)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Oops.  Sorry.  Refactoring has made this very silly and I failed to notice.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Reworked to compute total coverage for each interval.  Eliminates silly read length computation.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Sounds promising, but I can't find it.  Where is the OverclippedReadFilter?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Reworked into separate KmerCounter and KmerReducer objects so that the shuffle can use the more economical KmerAndCount class.  Substantially reduces memory use.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: A MultiMap collection type would certainly be clearer, and I'll try to pull that together.
The key has to be QName, because when I'm gathering the reads for an interval I know the QName and need to map through to which intervals, if any, the read belongs to.  The natural way would be Map<QName,List<IntervalID>>, but the vast majority of QNames are associated with a single interval, and so having a gazillion lists of 1 element takes much more memory than an occasional repeated QName.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
cwhelan: It looks like it wasn't ported over from GATK 3 yet, here it is in that codebase:

https://github.com/broadgsa/gatk/blob/0b73e380436aaa5a41fb3aab97ab651207669f47/public/gatk-engine/src/main/java/org/broadinstitute/gatk/engine/filters/OverclippedReadFilter.java

There's also a version in htsjdk: `htsjdk.samtools.filter.OverclippedReadFilter`, but that's hard to use because it wants a SamRecord rather than a GATKRead. We should probably just port it over into GATK4.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchHashSet.java
tedsharpe: It's for when you're building from a collection, but know that you'll be adding more.  It's fairly silly.  I'll get rid of it.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Eliminated in a refactoring.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Refactoring has made this go away, but I did test writing a functor for mapPartitions that returned after iterating through the first N (N=2) records followed by collect.  That was lightning fast.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: I think I wasn't double counting -- a template name for a pair of reads will generate 2*readLength bases in the assembly -- but it's moot, anyway.  I've refactored to eliminate this step.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: SimpleInterval holds the contig as a String.  Much heavier, and doesn't collate properly.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Yes, we're doing multiple kinds of cleaning when one really good one really ought to suffice.  But with regard to your comment about operating within a single partition:  that is true, but it happens after a shuffle (reduceByKey operation).  So all identical kmers are all together in a single partition (based on their hashCode).  So this operation is, in fact, global in scope.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: No, it's silly.  Will make it local to the method.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVUtils.java
tedsharpe: Replaced with apache commons code (with some misgivings).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Ah.  Well, I'm already being more stringent:  I require 45 bases of M blocks with no requirement on the number of S blocks.  (I'll add in = blocks and X blocks -- that was an oversight.  But including I blocks seems wrong.)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -87,98 +99,186 @@ protected void runTool( final JavaSparkContext ctx ) {
...
-                final int initialSetSize = qNamesSet.size();
-                log("Discovered " + initialSetSize + " mapped template names.");
-
-                // remove intervals that have ridiculously high coverage -- these are mapper artifacts
-                cleanQNames(qNamesSet, intervals, readMetadata.getMeanBasesPerTemplate(), locations, pipelineOptions);
-                log("Killed " + (initialSetSize-qNamesSet.size()) + " intervals that had >" + MAX_COVERAGE + "x coverage.");
+        final JavaRDD<GATKRead> unfilteredReads = getUnfilteredReads();
+        final JavaRDD<GATKRead> allPrimaryLines =
+                unfilteredReads.filter(read -> !read.isSecondaryAlignment() && !read.isSupplementaryAlignment());
+        final int[] nIntervals = new int[1];
```
cwhelan: This is not a huge deal but returning the number of intervals in an array of size 1 is stylistically not a very Java-like thing to do. Could you return a results object or maybe just a `Tuple2<Integer, HopscotchHashSet<QNameAndInterval>>`? Nitpicky, I know, but more in keeping with Java style.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -87,98 +99,186 @@ protected void runTool( final JavaSparkContext ctx ) {
...
-                qNamesSet = getQNames(ctx, broadcastMetadata, intervals, reads);
-                broadcastMetadata.destroy();
-
-                final int initialSetSize = qNamesSet.size();
-                log("Discovered " + initialSetSize + " mapped template names.");
-
-                // remove intervals that have ridiculously high coverage -- these are mapper artifacts
-                cleanQNames(qNamesSet, intervals, readMetadata.getMeanBasesPerTemplate(), locations, pipelineOptions);
-                log("Killed " + (initialSetSize-qNamesSet.size()) + " intervals that had >" + MAX_COVERAGE + "x coverage.");
+        final JavaRDD<GATKRead> unfilteredReads = getUnfilteredReads();
```
cwhelan: Just curious if you tried adding a `cache()` to this RDD? If not let's experiment with that after this PR is merged in.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -0,0 +1,500 @@
...
+        return !read.isPaired() ? TemplateEnd.UNPAIRED :
+               !read.isFirstOfPair() && !read.isSecondOfPair() ? TemplateEnd.PAIRED_UNKNOWN :
+                read.isFirstOfPair() && !read.isSecondOfPair() ? TemplateEnd.PAIRED_FIRST :
+               !read.isFirstOfPair() &&  read.isSecondOfPair() ? TemplateEnd.PAIRED_SECOND :
+                       TemplateEnd.PAIRED_INTERIOR;
+    }
+
+    @DefaultSerializer(SplitRead.Serializer.class)
+    public static final class SplitRead extends BreakpointEvidence {
+        private static final short UNCERTAINTY = 2;
```
cwhelan: Brainstorming a thought here but it seems like we might be able to get rid of this uncertainty if we compared the cigars of the two alignments (the one in the primary read and the one in the supplementary) and computed the microhomology/microinsertion size between them. I'll create a ticket to investigate that.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -0,0 +1,500 @@
...
+            @Override
+            public SplitRead read( final Kryo kryo, final Input input, final Class<SplitRead> klass ) {
+                return new SplitRead(kryo, input);
+            }
+        }
+    }
+
+    @DefaultSerializer(LargeIndel.Serializer.class)
+    public static final class LargeIndel extends BreakpointEvidence {
+        private static final short UNCERTAINTY = 4;
```
cwhelan: Seems like we could get rid of this uncertainty too -- again, for another day.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -0,0 +1,1303 @@
...
+        final List<SAMSequenceRecord> contigs = header.getSequenceDictionary().getSequences();
+        final int nContigs = contigs.size();
+        final int minMapQ = params.minEvidenceMapQ;
+        final int minMatchLen = params.minEvidenceMatchLength;
+        final int minEvidenceCount = params.minEvidenceCount;
+        final JavaRDD<BreakpointEvidence> evidenceRDD =
+                reads
+                    .filter(read ->
+                            read.getMappingQuality() >= minMapQ &&
+                            read.getCigar().getCigarElements()
```
cwhelan: I think you can get better performance by calling `read.getCigarElements()` instead of `read.getCigar().getCigarElements()`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: It would be a good thing to do if we can afford the memory.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
tedsharpe: We could do that, but I'm not convinced it would pay big dividends.  We're about to pull a region of ~300 bases (twice the read length) to assemble, anyway, so nailing the precise location doesn't seem that important.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: It's implemented as
`return samRecord.getCigar() == null ? Collections.emptyList() : samRecord.getCigar().getCigarElements();`
so it won't be any faster today, but it *could* be faster with revised code and I don't see how it could ever be slower, so I'll make that change.

cwhelan:
This looks good to me now. Thanks for the revisions. Please squash/rebase/merge.

PR #1836, opened 2016-05-18
Port Pileup

magicDGS:
Porting Pileup (https://github.com/broadinstitute/gatk/issues/1581) and tests

Issue #589, opened 2015-06-22
add read filtering to command line options


akiezun:
originally from #429 
The request is to be able to specify read filters on the commandline. Each read filter may have additional options. The read filters are to be applied in the order specified on commandline and applied before the tool. 

@vdauwera can clarify additional requirements

droazen:
This is a pretty important engine feature we're missing -- let's attempt for alpha-1.

Issue #121, opened 2015-01-15
Engine should create output writers/streams on behalf of tools that require them


droazen:

akiezun:
to clarify: the requirement is to implement a new annotation for outputs, mark output fields with that annotation, then add code to open the writers to the engine (probably in GATKTool) and close the writers on exit (regardless of fail/success).

lbergelson:
I'm not sure how much of this we want to tackle before dataflowing. We are
probably going to want to provide a standardized output format that will
map nicely to Dataflow. Currently this means a single output pcollection,
although some tools obviously will need multiple output streams.

droazen:
Yes, we shouldn't assume this will necessarily be done any particular way (eg., through annotations) unless we have more working examples of dataflow tools from which to generalize.

akiezun:
we need this independently of dataflow (needs to work one file walkers as well as dataflow)

akiezun:
alpha candidate?

droazen:
yes!

droazen:
This applies to spark too.

droazen:
In https://github.com/broadinstitute/gatk/pull/1146 we've at least centralized SAM/BAM writer creation methods in the superclasses -- this is good enough for alpha. In beta we can add the fancy auto-management of outputs we are dreaming about.

cmnbroad:
There is a [branch/commit](https://github.com/broadinstitute/gatk/commit/2fb990dfeb7f5be20d6d09e184327d3872094f26) here with changes to centralize creation of the output VCFWriter for all (non-Spark) tools in order to centralize index creation, which is an incremental step in this direction.

droazen:
@cmnbroad If you have any ideas about how this should be done, feel free to suggest an approach

lbergelson:
What happened with that branch?

droazen:
It is blocked by another branch currently under review, I believe.

PR #1862, opened 2016-05-27
Port CheckPileup

magicDGS:
Because I'm implementing a program in GATK that comes for a pipeline using `samtools mpileup` and post-processing the file, I need a way to check if the internal `ReadPileup` generated by the tool is indeed the same as the one obtained for samtools.

Thus, I ported the `CheckPileup` tool from GATK3 in this PR.


src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+import java.io.PrintStream;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Compare GATK's internal pileup to a reference Samtools pileup
+ *
+ * <p>At every locus in the input set, compares the pileup data (reference base, aligned base from
+ * each overlapping read, and quality score) generated internally by GATK to a reference pileup data generated
+ * by Samtools. Note that the pileup program has been replaced in Samtools by mpileup, which produces a slightly
```
lbergelson: Does your tool read mpileup format?  Samtools no longer offers the pileup command.  It looks like mpileup produces something very similar to the 6 column format here, but I'm not sure what the differences if any are.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+ * different output format by default.
+ * </p>
+ *
+ * <h3>Format</h3>
```
lbergelson: No need to include a description of the format here.  At most have a link to the pileup format definition.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+public class CheckPileup extends LocusWalker {
+
+    /**
+     * This is the existing pileup against which we'll compare GATK's internal pileup at each genome position in the desired interval.
+     */
+    @Argument(fullName = "pileup", shortName = "pileup", doc="Pileup generated by Samtools")
+    FeatureInput<SAMPileupFeature> pileup;
+
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME, shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "An output file created by the walker. Will overwrite contents if file exists")
+    private PrintStream out;
```
lbergelson: Huh, I didn't realize PrintStream had a string constructor.  This saves a few lines, but it's probably better use a `File` here and manually instantiate the stream in order to be able to throw a clearer `UserException` when an invalid path is given.  Right now it will result in something like 
`A USER ERROR has occurred: Invalid command line: Argument  output has a bad value: /some/invalid/path. Problem constructing PrintStream from the string '/some/invalid/path'.` 

Which isn't that bad, but would be clearer as a `CouldNotCreateOutputFile`  Not a big deal either way.  There's a long neglected plan to make generating and cleaning up output streams cleaner, but it hasn't happened yet.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * By default the program will quit if it encounters an error (such as missing truth data for a given position).
+     * Use this flag to override the default behavior; the program will then simply print an error message and move on
+     * to the next position.
+     */
+    @Argument(fullName="continue_after_error",doc="Continue after encountering an error",optional=true)
+    public boolean continueAfterAnError = false;
```
lbergelson: it seems strange that all the arguments have different levels of visibility.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+
+    @Override
+    public boolean requiresReference() {
+        return true;
+    }
+
+    public long nLoci = 0;
```
lbergelson: these should probably be private

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+        String bQuals = maybeSorted(new String(b.getQuals()), ! orderDependent );
+        if ( ! aQuals.equals(bQuals) ) {
+            return "Quals not equal";
+        }
+
+        return null;
+    }
+
+    /**
+     * Extracts the true pileup data from the given rodSAMPileup.  Note that this implementation
```
lbergelson: Update these references to "rods"

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+     * @return True pileup data.
+     */
+    private SAMPileupFeature getTruePileup( FeatureContext tracker ) {
+        final List<SAMPileupFeature> features = tracker.getValues(pileup);
+        SAMPileupFeature pileupArg = (features.isEmpty()) ? null : features.get(0);
+
+        if( pileupArg == null)
+            return null;
+
+        if( pileupArg.hasPointGenotype() ) {
```
lbergelson: this logic is strange, getPointGenotype just returns  `isPointGenotype() ? this : null` and so does getIndelGenotypes

and it looks like `hasPointGenotype` is defined as `!hasIndelGenotype`

I think this whole block can just be `return pileupArg == null ? null : return pileupArg`  

I don't see how that exception could ever be thrown...  Maybe I'm just missing something dumb though.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+            Arrays.sort(bytes);
+            return new String(bytes);
+        }
+        return x;
+    }
+
+    public String pileupDiff(final ReadPileup a, final SAMPileupFeature b, boolean orderDependent)
+    {
+        if ( a.size() != b.size() ) {
+            return "Sizes not equal";
```
lbergelson: It would be nice to include the sizes so you can see how unequal they are, as well as maybe including the position.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+        }
+        if (IntervalUtils.compareLocatables(a.getLocation(), b, getReferenceDictionary()) != 0 ) {
+            return "Locations not equal";
```
lbergelson: same for the rest of these, probably want more information in the error message so people can see the difference.  You're probably the primary consumer of this though, so do what you think would be useful to you.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
lbergelson: It might be nice to make this output file optional, and have the diff go to stdout if it's not specified.  Fine like this though.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+    {
+        if ( sortMe ) {
+            byte[] bytes = x.getBytes();
+            Arrays.sort(bytes);
+            return new String(bytes);
+        }
+        return x;
+    }
+
+    public String pileupDiff(final ReadPileup a, final SAMPileupFeature b, boolean orderDependent)
```
lbergelson: since the only usage of this method is hardcoded with `orderDependent = true` I would either remove the parameter and simplify the code, or expose it as a command line parameter.

lbergelson:
As part of this, could you port `SAMPileupCodecUnitTest`.  It's shamefully incomplete and only tests `canDecode`, but that's better than nothing.  It would be good to have some unit tests for the codec itself on various pileup lines the different parsing cases.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,226 @@
...
+            return "Locations not equal";
+        }
+
+        String aBases = maybeSorted(new String(a.getBases()), ! orderDependent );
+        String bBases = maybeSorted(b.getBasesAsString(), ! orderDependent );
+        if ( ! aBases.toUpperCase().equals(bBases.toUpperCase()) ) {
+            return "Bases not equal";
+        }
+
+        String aQuals = maybeSorted(new String(a.getBaseQuals()), ! orderDependent );
```
lbergelson: This has the strange property that if these were sorted the quals can be sorted differently than the bases.  It seems like they should be sorted together because each qual goes with a specific base.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,343 @@
...
+ * </p>
+ * <p>The Pileup format was first used by Tony Cox and Zemin Ning at
+ *     the Sanger Institute. It describes the base-pair information at each chromosomal position. This format
+ *     facilitates SNP/indel calling and brief alignment viewing by eye. Note that the pileup program has been replaced
+ *     in Samtools by mpileup, which produces a slightly different output format by default.
+ * </p>
+
+ * <h3>Format</h3>
+ * <p>There are two versions of the original pileup format: the current 6-column format produced by Samtools, and the old
+ * 10/13-column "consensus" format which could be obtained by using the -c argument, now deprecated. </p>
```
lbergelson: do we need to support the deprecated format?  It seems like it would be better not to unless you really need it for something.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,343 @@
...
+/**
+ * Decoder for SAM pileup data.
+ *
+ * <p>
+ *     From the <a href="http://samtools.sourceforge.net/">SAMTools project documentation</a>:
+ * </p>
+ * <p>The Pileup format was first used by Tony Cox and Zemin Ning at
+ *     the Sanger Institute. It describes the base-pair information at each chromosomal position. This format
+ *     facilitates SNP/indel calling and brief alignment viewing by eye. Note that the pileup program has been replaced
+ *     in Samtools by mpileup, which produces a slightly different output format by default.
```
lbergelson: Do we handle single sample mpileups?  We definitely should since it's the standard now.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,343 @@
...
+ *     Escherichia_coli_K12	3995037	*	*\/*	430	0	37	144	*	+A	143	1	0
+ *     Escherichia_coli_K12	3995279	*	*\/*	202	0	36	68	*	+A	67	1	0
+ *     Escherichia_coli_K12	3995281	*	*\/*	239	0	36	67	*	-CG	66	1	0
+ * </pre>
+ * <p>
+ *     See <a href="http://samtools.sourceforge.net/cns0.shtml/">Consensus pileup format (deprecated)</a> for more details.
+ * </p>
+ *
+ * <h3>Caveat</h3>
+ * <p>Handling of indels is questionable at the moment. Proceed with care.</p>
```
lbergelson: this makes me uncomfortable...

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,343 @@
...
+    private static final char fldDelim = '\t';
+    // allocate once and don't ever bother creating them again:
+    private static final String baseA = "A";
+    private static final String baseC = "C";
+    private static final String baseG = "G";
+    private static final String baseT = "T";
+    private static final String emptyStr = ""; // we will use this for "reference" allele in insertions
+
+    // codec file extension
+    protected static final String FILE_EXT = "samp";
```
lbergelson: I think we may also want to accept `.pileup` since that's what the wikipedia page recommends as the most popular file extension for pileups.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,343 @@
...
+                // parsing is pretty straightforward for 6-col format
+                if ( feature.getRef() == '*' ) {   // this indicates an indel -- but it shouldn't occur with vanilla 6-col format
+                    throw new CodecLineParsingException("Found an indel on line = " + line + " but it shouldn't happen in simple pileup format");
+                } else {
+                    parseBasesAndQuals(feature, bases, quals);
+                    feature.setRefBases(tokens[2].toUpperCase());
+                    feature.setEnd(feature.getStart());
+                }
+                break;
+            case consensusSNPTokenCount: // pileup called a SNP or a reference base
```
lbergelson: unless you need them I would just remove the parsing for deprecated pileups and greatly simplify this class

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
```diff
@@ -0,0 +1,289 @@
...
+ *
+ * Allows intake of both simple (6-column) or extended/consensus (10/13-column) pileups. Simple pileup features will
+ * contain only basic information, no observed alleles or variant/genotype inferences, and so shouldn't be used as input
+ * for analysis that requires that information.
+ *
+ * @author mhanna
+ * @version 0.1
+ */
+public class SAMPileupFeature implements Feature {
+    public enum VariantType { NONE, SNP, INSERTION, DELETION, INDEL };
```
lbergelson: If it's ok to drop the deprecated pileup formats than much of this class can also be dropped.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
```diff
@@ -0,0 +1,289 @@
...
+        return !isIndelGenotype();
+    }
+
+    /**
+     * Implements method required by GenotypeList interface. If this object represents an indel genotype, then it
+     * returns itself through this method. If this object is a point genotype, this method returns null.
+     *
+     * @return
+     */
+    public SAMPileupFeature getIndelGenotype() {
```
lbergelson: a lot of these are for some outdated GenotypeList interface that aren't needed anymore and don't really do anything

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
```diff
@@ -0,0 +1,289 @@
...
+        return (isPointGenotype()) ? this : null;
+    }
+
+    /**
+     * Returns true if this object \em is an indel genotype (and thus indel genotype is what it only has).
+     *
+     * @return
+     */
+    public boolean hasIndelGenotype() {
+        return isIndelGenotype();
```
lbergelson: is indel genotype and has indel genotype are redundant

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    /**
+     * Returns true if this object \em is a point genotype (and thus point genotype is what it only has.
+     *
+     * @return
+     */
+    public boolean hasPointGenotype() {
```
lbergelson: same with hasPointGenotype

src/test/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileupIntegrationTest.java
```diff
@@ -0,0 +1,46 @@
...
+public class CheckPileupIntegrationTest extends CommandLineProgramTest {
+    private static final String TEST_DATA_DIRECTORY = publicTestDir + "org/broadinstitute/hellbender/engine/";
+    private static final String TEST_OUTPUT_DIRECTORY = publicTestDir + "org/broadinstitute/hellbender/tools/walkers/qc/pileup/";
+
+    /**
+     * This test runs on a consensus pileup containing 10-column lines for SNPs and 13-column lines for indels
+     */
+    @Test(enabled = false)
+    public void testConsensusPileup() throws IOException {
+        // TODO
```
lbergelson: If you're not including deprecated pileups than you remove this case

src/test/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileupIntegrationTest.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * This test runs on a basic pileup obtained with samtools (version 1.1) and options -B --min-BQ 0
```
lbergelson: would you mind using a more recent samtools to regenerate the test files and use mpileup?

lbergelson:
@magicDGS Review complete for now.  Looks good but I have some nitpicks.  I think they're almost all due to it being ancient gatk3 code that no one has updated in a long time.  I'd recommend dropping the deprecated formats and only supporting mpilup single sample format which should allow for massive simplification of both the Codec and the Feature.  

We need some unit tests for the codec itself since it has a bunch of different potential error cases, and we should have some integration tests for the tool that show it correctly failing on cases with errors.

magicDGS:
Thanks, @lbergelson. I was also thinkinhg that this code is mostly deprecated, but I wanted to ported as is for the first pass review. I just need to support the new mpileup version (unique sample, because if not it is more difficult), because the consensus one is deprecated.

I will update the codec and add some tests for it. In addition, I was thinking to create a list of `PileupElement` inside the feature to make easier to compare the internal pileup, but with "reads" of one base-pair.

On the other hand, I will improve the walker itself. I will tell you when I finished with the changes.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: mpileup and pileup are exactly the same, except the consensus format (not supported). I will remove all code related with "consensus pileup".

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Done both

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: All `public` now.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Simplified, because I'm going to get rid of the consensus format.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Included.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Added TODO, I will decide later if it is worthy to maintain.

lbergelson:
@magicDGS I appreciate the minimal diff initial import, makes it much easier to see what happened.  It's going to be much cleaner once you knock out all of that deprecated code.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Added TODO regarding this issue (I will try to fix using `PileupElement` lists)

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
magicDGS: It's not going to be supported, I deleted the documentation.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
magicDGS: Noted that the codec is only for single-sample mpileup

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
magicDGS: I removed the whole reference to the format, and I'll implement this from scratch.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
magicDGS: I removed "samp" (breaking compatibility with GATK3) and add the recomended ones ("pileup" and "mpileup")

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
magicDGS: Removed.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
magicDGS: Removed.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
magicDGS: Removed.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
magicDGS: Removed both (not needed anymore).

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
magicDGS: Removed both.

src/test/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileupIntegrationTest.java
magicDGS: Removed

src/test/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileupIntegrationTest.java
magicDGS: It should generate exactly the same result, but I will do it asap when I finish the implementation.

vdauwera:
FYI the last person who touched this in GATK3 was probably me -- I had to fix a bug in handing both old & new pileup format a while ago. I remember cringing a lot but my mission was to get it to not crash -- cleaning it up further was way out of scope. 

I say that purely defensively in case anybody does git blame and sees my name pop up ;)

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: When I ported `Pileup` (#1836), I was using also a `PrintStream` (it was thw way GATK3 implement it). Should I open a ticket for change it, @lbergelson? I think your point is right, could clarify user errors...

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Done (also nBases)

magicDGS:
Ready for second pass review, @lbergelson. Now the implementation is much more simple than the previous one, and I added unit tests for the codec.

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
lbergelson: No harm in adding a ticket.  Low priority, but it might help push us into coming up with a better solution where the engine can handle instantiating output streams and cleaning up after them.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
lbergelson: 👍 to that decision

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private long nLoci = 0;
+    private long nBases = 0;
+    private PrintStream out;
+
+    @Override
+    public void onTraversalStart() {
+        try {
+            out = (outFile == null) ? System.out : new PrintStream(outFile);
+        } catch (FileNotFoundException e) {
```
lbergelson: Sorry for this extra boiler plate.  We should improve the output stream handling in the future  part of #121 )

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,166 @@
...
+ * on the format
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class SAMPileupCodec extends AsciiFeatureCodec<SAMPileupFeature> {
+
+    // including also space as delimiter
+    private static final Pattern SPLIT_PATTERN = Pattern.compile("\\t|( +)");
+
+    // although the minimum number of fields is 6 (see specifications), if the coverage is 0 sometimes bases and qualities does not appear
```
lbergelson: thanks for this helpful explanation

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Only files with {@link #SAM_PILEUP_FILE_EXTENSIONS} could be parsed
+     * @param path path the file to test for parsability with this codec
+     * @return {@code true} if the path has the correct file extension {@link #SAM_PILEUP_FILE_EXTENSIONS}, {@code false} otherwise
+     */
+    @Override
+    public boolean canDecode(final String path) {
+        return SAM_PILEUP_FILE_EXTENSIONS.stream().anyMatch(ext -> path.endsWith("."+ext));
```
lbergelson: could you make this case insensitive?

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    public SAMPileupFeature decode(String line) {
+        // Split the line
+        final String[] tokens = SPLIT_PATTERN.split(line.trim(), -1);
+        // check the number of fields
+        if(tokens.length < MINIMUM_FIELDS || tokens.length > MAXIMUM_FIELDS) {
+            throw new CodecLineParsingException("The SAM pileup line didn't have the expected number of columns. " +
```
lbergelson: Could you include the actual text of the line in this comment as well as the number of columns and the expected range?

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,166 @@
...
+                    "Note that this codec is only valid for single-sample pileups.");
+        }
+        // starting parsing
+        final int pos = parseInteger(tokens[1], "position");
```
lbergelson: Would you mind extracting the contig out here too for consitency?

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,166 @@
...
+        if(cov != pileupElements.size()) {
+            throw new CodecLineParsingException("THe SAM pileup line didn't have the same number of elements as the expected coverage: " + cov);
+        }
+        return new SAMPileupFeature(tokens[0], pos, ref, pileupElements);
+
+    }
+
+    @VisibleForTesting
+    protected List<SAMPileupElement> parseBasesAndQuals(final String bases, final String qualities, final byte ref) {
+        // TODO: probably a index out of bounds exception will be thrown
```
lbergelson: Is this still relevant?

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
```diff
@@ -0,0 +1,166 @@
...
+                    break;
+                case '*': // deletion placeholder
+                    pileupElements.add(new SAMPileupElement(BaseUtils.Base.D.base, (byte) SAMUtils.fastqToPhred(qualities.charAt(j++))));
+                    break;
+                case '+':
+                case '-': // start of indel
+                    final String rest = bases.substring(i+1);
+                    final Matcher match = INDEL_REGEXP.matcher(rest);
+                    if(! match.matches()) {
+                        throw new CodecLineParsingException("The SAM pileup line have an indel marker (+/-) without lenght");
```
lbergelson: typo:  "The SAM pileup line *has* an indel marker (+/-) without *length*"

src/test/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileupIntegrationTest.java
```diff
@@ -0,0 +1,38 @@
...
+        final File emptyTemp = createTempFile("empty", "txt");
+        emptyTemp.createNewFile();
+        // pileup was generated with "samtools -f hg19MiniReference -B --min-BQ 0 reads_data_source_test1.bam"
+        IntegrationTestSpec testSpec = new IntegrationTestSpec(
+            " -R " + hg19MiniReference +
+            " -I " + TEST_DATA_DIRECTORY + "reads_data_source_test1.bam" +
+            " -pileup " +  TEST_OUTPUT_DIRECTORY + "reads_data_source_test1.samtools.samp" +
+            " -O %s", Arrays.asList(emptyTemp.toString()));
+
+        testSpec.executeTest("testBasicPileup", this);
```
lbergelson: Could you add an additional test that shows it correctly catching a difference?

src/test/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileupIntegrationTest.java
```diff
@@ -0,0 +1,38 @@
...
+ * Tests only single-sample mpileups (no consensus)
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class CheckPileupIntegrationTest extends CommandLineProgramTest {
+    private static final String TEST_DATA_DIRECTORY = publicTestDir + "org/broadinstitute/hellbender/engine/";
+    private static final String TEST_OUTPUT_DIRECTORY = publicTestDir + "org/broadinstitute/hellbender/tools/walkers/qc/pileup/";
+
+    /**
+     * This test runs on a basic pileup obtained with samtools (version 1.1) and options -B --min-BQ 0
```
lbergelson: is this still accurate?

src/test/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodecUnitTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.broadinstitute.hellbender.utils.BaseUtils.Base.*;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class SAMPileupCodecUnitTest extends BaseTest {
```
lbergelson: Could you add a test that uses `decodeLoc`?

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    public int size() {
+        return pileupElements.size();
+    }
+
+    /**
+     * Returns formatted pileup string for the genomic location as
+     * "location: reference_base observed_base_pile observed_qual_pile"
+     */
+    public String getPileupString() {
```
lbergelson: Could you add a test for this method?  It's super simple but it's important that it doesn't change unexpectedly and people tend to be cavalier about toString sorts of things.

lbergelson:
@magicDGS You really made it much clearner.  Huge improvement over the old code.  Thank you! .  

Back to you with a few typos and request for a few more tests.

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
magicDGS: It is because I forgot to add the try-catch block. If the qualities have less length than the bases because the pileup is malformed, the code won't reach the check of the lengths at the end of the method. I've just added the try-catch

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodec.java
magicDGS: Corrected

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
magicDGS: It is weird that `ReadPileup` and `SAMPileupFeature`, which are similar, have a different behavior regarding this method. I changed now the contract and added a test for it (and solve a bug that I detected in getQualsAsString).

src/test/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileupIntegrationTest.java
magicDGS: No, now it with samtools version 1.3.1. Changed

src/main/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupFeature.java
magicDGS: Actually they are representing the same kind of feature, but `ReadPileup` backed the reads I don't know if it is worthy, because `SAMPileupFeature` will be used only for this tool. From my side, it is just a intermediate step to be sure that the new GATK framework will provide me an easy way of create internally exactly the same pileups as samtools. And it will be more important when #589 provides a way of filter in a similar way as samtools and and #1886 expose the option to do BAQ. Anyway, in my next commit the methods will be normalized (names) and we can discuss about the interface.

src/test/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileupIntegrationTest.java
magicDGS: Done, with BAQ qualities differences.

src/test/java/org/broadinstitute/hellbender/utils/codecs/sampileup/SAMPileupCodecUnitTest.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Created issue

magicDGS:
Thanks for the review, @lbergelson. Back to you with addressed comments and added tests.

I don't know if it is good for the clarity of the framework to create an interface for `Pileup`s (for `ReadPileup` and `SAMPileupFeature`), but let me know if you would like that I include that in this PR.

lbergelson:
@magicDGS It sounds like a common interface isn't necessary. 

Do you think this codec would be useful in htsjdk?  Would other people want to use it?  We could consider moving it there in the future.

lbergelson:
This looks good to me.  👍

lbergelson:
@magicDGS Thanks for the contribution!

magicDGS:
Thanks for reviewing, @lbergelson. I think that the codec could be useful in htsjdk, because I know several people that works with the samtools-pileup format, and it will be nice to be able to use it in the htsjdk API. In addition, it could be useful for include new behaviour and refine the implementation (like retrieving reads from the format).

PR #1863, opened 2016-05-27
moved dependency fixes from protected to here

akiezun:
these are moved over from protected to public. Without them, even count reads does not work after we added mllib.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -136,7 +136,7 @@ dependencies {
 
     compile 'org.jgrapht:jgrapht-core:0.9.1'
     compile 'org.testng:testng:6.9.6' //compile instead of testCompile because it is needed for test infrastructure that needs to be packaged
-    compile 'org.apache.hadoop:hadoop-minicluster:2.7.1' //the version of minicluster should match the version of hadoop
+    compile 'org.apache.hadoop:hadoop-minicluster:2.7.2' //the version of minicluster should match the version of hadoop
 
     compile('org.seqdoop:hadoop-bam:7.5.0') {
         exclude group: 'org.apache.hadoop'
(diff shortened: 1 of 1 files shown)


akiezun:
@droazen can you review? Look at protected build file

build.gradle
```diff
@@ -156,6 +156,19 @@ dependencies {
         exclude module: 'kryo' // use Spark's version
     }
```
droazen: Do you need this as well (from protected)?

    // Comment the next line to disable native code proxies in Spark MLLib
    compile('com.github.fommil.netlib:all:1.1.2')

build.gradle
```diff
@@ -156,6 +156,19 @@ dependencies {
         exclude module: 'kryo' // use Spark's version
```
droazen: protected uses a different version of this dependency (0.26) -- will there be a conflict?

droazen:
@akiezun Can you please test gatk-protected with a local snapshot of this branch, and confirm that protected still builds and tests pass?

droazen:
Back to @akiezun

akiezun:
it won't. Protected is way behind and needs an urgent revup

droazen:
@akiezun In that case, merge when ready and we'll have to deal with the breakage in protected separately. This is going to cause issues for people who need to test out public snapshots of branches with protected, so we should fix soon.

build.gradle
akiezun: i don't know. Let's assume yes

build.gradle
akiezun: protected will get rid of this dependency once they rev up

PR #1867, opened 2016-05-27
Update to htsjdk 2.4.1

droazen:


droazen:
@akiezun please review

akiezun:
Will do. Need to test on spark clusters. Are any changes to gatk-launch required?

droazen:
Hopefully not, but let me know.

droazen:
@akiezun There was indeed a change needed in `gatk-launch` -- I've added this now.

droazen:
This is now an update to htsjdk 2.4.1 rather than 2.4.0. I had to disable the `IntelDeflater` in htsjdk for now until `intel-gkl` is on maven central, since we can't release htsjdk to maven with a dependency that's not in central.

akiezun:
@droazen can we skip this release and wait for the IntelDeflater to come back?

droazen:
@akiezun We can skip this one, as it doesn't contain anything we need right now. I don't think we can skip the next one, however, as it will contain urgent CRAM bug fixes -- and Intel is still weeks away from publishing their GKL to maven central, I believe (@gspowley is this correct? Any chance this could be expedited?)

gspowley:
@droazen Weeks away is a safe assumption. We're going through the process to approve releasing the code as Open Source. I'll see if there is a way to expedite.

akiezun:
@droazen what do you think about passing the deflater to the sam-writer factory and removing GLK as a htsjdk dependency altogether?

droazen:
@akiezun It's a tempting option, provided it's done in a generic way (arg of type `Deflater` and not `IntelDeflater`) and the existing global properties in `Defaults` are removed. It might cause problems for Picard, however, as it would make it difficult to enforce that the Intel deflater be used across all Picard tools (since Picard lacks a centralized engine). Is the IntelDeflater currently on globally in Picard?

Another possibility is that we could create a branch of htsjdk in which the deflater is turned back on (easily done, all the code is still there), depend on snapshots of that branch temporarily in GATK until GKL is in maven central, and update the branch as needed with fixes from master. If it's only a few weeks this might be bearable.

akiezun:
FWIW I poked around a bit with the deflater and the main issue is that if we pass the deflater as a parameter then the default is specified simply by a constructor call and not by a run-time constant (also, the deflater can only be created once the compression level is known which is very late) and so there's no longer a central place where everyone gets their customizable deflaters from. I have a solution that I can submit for review in the next few days (make `DeflaterFactory` a proper factory with an overridable  `makeDeflater` method, make it settable on writerFactory and also and provide a default one that can be set/queried in `BlockCompressedOutputStream.getDeflaterFactory` - that way one can set the default once and for all for everyone and also selectively use different deflaters for different writers is one so desires).

akiezun:
i'm closing this for now

droazen:
@akiezun Or we can just wait a few weeks and depend on a snapshot of htsjdk in the mean time?

PR #1871, opened 2016-06-01
SelectVariants and VariantFiltration not updating AC, AN and AF for --setFilteredGtToNocall

### LINKED ISSUE #1731: SelectVariants and VariantFiltration not updating AC, AN and AF for --setFilteredGtToNocal
Port https://github.com/broadinstitute/gsa-unstable/pull/1347.

ronlevine:
Fixes #1731 

**Summary of changes**
* SelectVariants.java
    * createVCFHeaderLineList() - ensure allele count INFO annotations (AC, AN, AF) are in the output VCF header if --setFilteredGtToNocall  
    * setFilteredGenotypeToNocall() - recompute AC, AN and AF if filtered genotypes are set to no-call, add them to the genotype builder.
* VariantFiltration.java
    * initializeVcfWriter() - ensure allele count INFO annotations (AC, AN, AF) are in the output VCF header if --setFilteredGtToNocall  
    * makeGenotypes() - recompute AC, AN and AF if filtered genotypes are set to no-call, add them to the genotype builder.
* GATKVariantContextUtils.java
    * incrementChromosomeCountsInfo() - count the total and alternate alleles for a genotype
    * updateChromosomeCountsInfo() - update AC, AN and AF with the computed count of total and alternate alleles
* selectVariantsInfoField.vcf* and variantFiltrationInfoField.vcf*
    * User data from https://github.com/broadinstitute/gsa-unstable/issues/1325
* VariantFiltrationIntegrationTest.java
    * Tests that show the fix for https://github.com/broadinstitute/gsa-unstable/issues/1325
* SelectVariantsIntegrationTest.java
    * MD5 changed in testSetFilteredGtoNocall() because AC, AN and AF are always in the VCF if --setFilteredGtToNocall
* GATKVariantContextUtilsUnitTest.java
    * Test new GATKVariantContextUtils methods


ronlevine:
@droazen Please choose a reviewer.

droazen:
@akiezun Would you have time to review this one this week, since you've been looking at Ron's other PRs?

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
```diff
@@ -177,6 +178,9 @@
 
     private final List<Allele> diploidNoCallAlleles = Arrays.asList(Allele.NO_CALL, Allele.NO_CALL);
 
+    // Number of called alleles
+    int calledAlleles = 0;
```
akiezun: make the field private

akiezun: actually, remove this field. It is state that is *not* carried over `apply` calls and thus it is useless. Turn it into a temp and pass directly to methods that need it

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
```diff
@@ -323,9 +335,20 @@ private void filter(final VariantContext vc, final FeatureContext featureContext
...
 
-    private GenotypesContext makeGenotypes(final VariantContext vc) {
+    private GenotypesContext makeGenotypes(final VariantContext vc, final VariantContextBuilder builder ) {
         final GenotypesContext genotypes = GenotypesContext.create(vc.getGenotypes().size());
 
+        //
+        // recompute AC, AN and AF if filtered genotypes are set to no-call
+        //
+        // occurrences of alternate alleles over all genotypes
+        final Map<Allele, Integer> calledAltAlleles = new LinkedHashMap<Allele, Integer>(vc.getNAlleles()-1);
```
akiezun: use diamond type for new, ie `new LinkedHashMap<>(vc.getNAlleles()-1);`

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
```diff
@@ -343,14 +366,19 @@ private GenotypesContext makeGenotypes(final VariantContext vc) {
...
                     genotypes.add(new GenotypeBuilder(g).filters(filters).alleles(diploidNoCallAlleles).make());
                 } else {
                     genotypes.add(new GenotypeBuilder(g).filters(filters).make());
+                    calledAlleles = GATKVariantContextUtils.incrementChromosomeCountsInfo(calledAltAlleles, calledAlleles, g);
                 }
             } else {
                 genotypes.add(g);
             }
         }
+        if ( haveFilteredNoCallAlleles )
```
akiezun: braces

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1675,4 +1675,55 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
...
+    /**
+     * Increment the number of called alternate and reference plus alternate alleles for a genotype
+     *
+     * @param calledAltAlleles number of called alternate alleles for all genotypes
+     * @param calledAlleles    number of called alleles for all genotypes
+     * @param genotype         genotype
+     * @return incremented called alleles
+     * @throws IllegalArgumentException if calledAltAlleles or genotype are null
+     */
+    public static int incrementChromosomeCountsInfo(final Map<Allele, Integer> calledAltAlleles, final int calledAlleles, final Genotype genotype) {
```
akiezun: this api is sub optimal - it does not need the `calledAlleles` argument. Remove the argument and make the method just return the number (clients will then add that number to whatever they need). Rename method to `getChromosomeCounts` (there's no `Info`)

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1675,4 +1675,55 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
...
+        if ( calledAltAlleles == null ) throw new IllegalArgumentException("Called alternate alleles can not be null");
```
akiezun: use `Utils.nonNull` for all null checks

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
akiezun: why is the capacity set to `vc.getNAlleles()-1` if you know you're putting `vc.getAlternateAlleles().size()` keys into the map?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
```diff
@@ -1096,15 +1103,32 @@ private VariantContext setFilteredGenotypeToNocall(final VariantContext vc, fina
         final VariantContextBuilder builder = new VariantContextBuilder(vc);
         final GenotypesContext genotypes = GenotypesContext.create(vc.getGenotypes().size());
 
+        //
+        // recompute AC, AN and AF if filtered genotypes are set to no-call
+        //
+        // occurrences of alternate alleles over all genotypes
+        final Map<Allele, Integer> calledAltAlleles = new LinkedHashMap<Allele, Integer>(vc.getNAlleles()-1);
```
akiezun: same comments as in VariantFiltration

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
```diff
@@ -1096,15 +1103,32 @@ private VariantContext setFilteredGenotypeToNocall(final VariantContext vc, fina
...
             }
             else {
+                // increment the number called alleles and called alternate alleles
+                calledAlleles = GATKVariantContextUtils.incrementChromosomeCountsInfo(calledAltAlleles, calledAlleles, g);
                 genotypes.add(g);
             }
         }
 
+        // if filtered genotypes are set to no-call, output recomputed AC, AN, AF
+        if ( haveFilteredNoCallAlleles )
```
akiezun: braces

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1675,4 +1675,55 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
...
+     * Update the variant context chromosome counts info fields (AC, AN, AF)
+     *
+     * @param calledAltAlleles  number of called alternate alleles for all genotypes
+     * @param calledAlleles     number of called alleles for all genotypes
+     * @param builder           builder for variant context
+     * @throws IllegalArgumentException if calledAltAlleles or builder are null
+     */
+    public static void updateChromosomeCountsInfo(final Map<Allele, Integer> calledAltAlleles, final int calledAlleles,
+                                                  final VariantContextBuilder builder) {
+        if ( calledAltAlleles == null ) throw new IllegalArgumentException("Called alternate alleles can not be null");
```
akiezun: Utils.nonNull

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1675,4 +1675,55 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
...
+        if ( builder == null ) throw new IllegalArgumentException("Variant context builder can not be null");
+
+        builder.attribute(VCFConstants.ALLELE_COUNT_KEY, calledAltAlleles.values().toArray()).
+                attribute(VCFConstants.ALLELE_NUMBER_KEY, calledAlleles);
+        // Add AF is there are called alleles
+        if ( calledAlleles != 0 ) {
+            final Set<Double> alleleFrequency = new LinkedHashSet<Double>(calledAltAlleles.size());
```
akiezun: diamond type

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1675,4 +1675,55 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
...
+            for ( final Integer value : calledAltAlleles.values() ) {
```
akiezun: `int`, not `Integer`

akiezun:
looks good. small edits. back to you @ronlevine. Please rebase before next round.

droazen:
@ronlevine Will you have time to address Adam's comments on this PR?

ronlevine:
Yes. It will have to wait until Monday.

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
ronlevine: Used diamond type, made capacity `vc.getAlternateAlleles().size()`

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
ronlevine: Moved to `makeGenotypes`.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
ronlevine: Used diamond type, made capacity `vc.getAlternateAlleles().size()`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
ronlevine: Added braces.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: Done.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: Used diamond type.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: Would like to keep `Integer` in order to use `value.doubleValue()`.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: Changed method to `int getChromosomeCounts(final Map<Allele, Integer> calledAltAlleles, final Genotype genotype)`. Clients add the return value to `calledAlleles`.

ronlevine:
@droazen Please take a look. Note that I removed the functional  programming in `SelectVariants.setFilteredGenotypeToNocall()`. I could not think of a clean way to implement the added logic.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1349,4 +1349,54 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
...
+     * @throws IllegalArgumentException if calledAltAlleles or builder are null
+     */
+    public static void updateChromosomeCountsInfo(final Map<Allele, Integer> calledAltAlleles, final int calledAlleles,
+                                                  final VariantContextBuilder builder) {
+        Utils.nonNull(calledAltAlleles, "Called alternate alleles can not be null");
+        Utils.nonNull(builder, "Variant context builder can not be null");
+
+        builder.attribute(VCFConstants.ALLELE_COUNT_KEY, calledAltAlleles.values().toArray()).
+                attribute(VCFConstants.ALLELE_NUMBER_KEY, calledAlleles);
+        // Add AF is there are called alleles
```
lbergelson: typo is -> if

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1349,4 +1349,54 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
...
+     *
+     * @param calledAltAlleles  number of called alternate alleles for all genotypes
+     * @param calledAlleles     number of called alleles for all genotypes
+     * @param builder           builder for variant context
+     * @throws IllegalArgumentException if calledAltAlleles or builder are null
+     */
+    public static void updateChromosomeCountsInfo(final Map<Allele, Integer> calledAltAlleles, final int calledAlleles,
+                                                  final VariantContextBuilder builder) {
+        Utils.nonNull(calledAltAlleles, "Called alternate alleles can not be null");
+        Utils.nonNull(builder, "Variant context builder can not be null");
```
lbergelson: you might want a guard against negative calledAlleles here

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
lbergelson: typo: is -> if

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
```diff
@@ -206,6 +207,13 @@ private void initializeVcfWriter() {
         final Set<VCFHeaderLine> hInfo = new LinkedHashSet<>();
         hInfo.addAll(getHeaderForVariants().getMetaDataInInputOrder());
 
+        // need AC, AN and AF since output if set filtered genotypes to no-call
+        if ( setFilteredGenotypesToNocall ) {
```
lbergelson: This code is identical to some code below it, could it be factored into a method in GATKVariantContextUtils?  Or possible added to a new setFilteredGenotypeToNoCallArgumentCollection.

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
```diff
@@ -323,9 +331,23 @@ private void filter(final VariantContext vc, final FeatureContext featureContext
         writer.add(builder.make());
     }
 
-    private GenotypesContext makeGenotypes(final VariantContext vc) {
+    private GenotypesContext makeGenotypes(final VariantContext vc, final VariantContextBuilder builder ) {
```
lbergelson: This would be better as a void method that returns a modified builder.  It's confusing that it modifies the builder and returns genotypes that are used to modify the builder.

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
```diff
@@ -323,9 +331,23 @@ private void filter(final VariantContext vc, final FeatureContext featureContext
...
         final GenotypesContext genotypes = GenotypesContext.create(vc.getGenotypes().size());
 
+        //
+        // recompute AC, AN and AF if filtered genotypes are set to no-call
+        //
+        // occurrences of alternate alleles over all genotypes
```
lbergelson: The new logic here and the logic above are substantially similar.  Is it possible to factor this so that the chromosomeCounts logic is separated from the filtering and applied afterwards.  T
hen you might be able to have a single function that does the update in both VariantFiltration and select variants.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1349,4 +1349,54 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
...
+
+    /**
+     * Get the number of called alleles for a genotype. Also, increment the number of called alternate alleles.
+     *
+     * @param calledAltAlleles number of called alternate alleles for all genotypes
+     * @param genotype         genotype
+     * @return incremented called alleles
+     * @throws IllegalArgumentException if calledAltAlleles or genotype are null
+     */
+    public static int getCalledChromosomeCounts(final Map<Allele, Integer> calledAltAlleles, final Genotype genotype) {
```
lbergelson: It's weird that this method is a non-void method that updates it's input.  Since computing the calledAlleles number is essentially orthogonal to updating could you split the functionality?  It seems like these are packaged together in order to reduce complexity to the caller since they're needed in multiple places, but I think that could be avoided by refactoring so that it's only needed by a single function that can be called in both places.

lbergelson:
@ronlevine It looks to me that there could be some refactoring done here to separate out the AF update logic into a single function call that could be used in multiple places rather than having two sets of fairly complicated and highly similar logic.  Could you see if that's doable?

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: I am not sure if I am following you. What is the method signature you are proposing?

ronlevine: @lbergelson  Ping.

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
ronlevine: I assume you mean the same between `SelectVariants` and `VariantFiltration`. I added the method `void addChromosomeCounts(final Set<VCFHeaderLine> headerLines)` to `GATKVariantContextUtils`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
ronlevine: Done. Changed to `void makeGenotypes(final VariantContext vc, final VariantContextBuilder builder)`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
ronlevine: Made a utility method in `GATKVariantContextUtils` called `setFilteredGenotypeToNocall` that takes `filters` as an input.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: Done

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: Done.

ronlevine:
@lbergelson I refactored the common logic into a utility class. Please take a look.

ronlevine:
@lbergelson Per your suggestion, I forgot to make `setFilteredGenotypeToNocall()` a void function.  This is all set for review.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1349,4 +1351,119 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
 
         return -1;
     }
+
+    /**
+     * Add chromosome Counds (AC, AN and AF) to the VCF header lines
```
lbergelson: typo: counds

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
lbergelson: That was exactly what I meant.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1349,4 +1351,119 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
...
+     *
+     * @param headerLines the VCF header lines
+     */
+    public static void addChromosomeCounts(final Set<VCFHeaderLine> headerLines) {
```
lbergelson: Is `addChromosomeCounts` the right name for this?  I think maybe `addAlleleCountsToHeader` is more accurate if a bit more wordy.

ronlevine: I think so, but your suggestion is more explicit. I'll go with `addAlleleCountsToHeader`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
```diff
@@ -206,6 +207,11 @@ private void initializeVcfWriter() {
         final Set<VCFHeaderLine> hInfo = new LinkedHashSet<>();
         hInfo.addAll(getHeaderForVariants().getMetaDataInInputOrder());
 
+        // need AC, AN and AF since output if set filtered genotypes to no-call
```
lbergelson: `If setFilteredGenotypesToNoCall is true, then allele counts will be recomputed and need to be included in the header`

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1349,4 +1351,119 @@ private static int indexOfSameAllele(final VariantContext vc, final Allele allel
...
+
+    /**
+     * Get the number of called alleles for a genotype. Also, increment the number of called alternate alleles.
+     *
+     * @param calledAltAlleles number of called alternate alleles for all genotypes
+     * @param genotype         genotype
+     * @return incremented called alleles
+     * @throws IllegalArgumentException if {@code calledAltAlleles} or {@code genotype} are {@code null}.
+     */
+    protected static int getCalledChromosomeCounts(final Map<Allele, Integer> calledAltAlleles, final Genotype genotype) {
```
lbergelson: This is a final class so it's weird to have anything be protected.  If this is exposed for testing, could you annotate it with `@VisibleForTesting` and give it default visablity instead of protected.

lbergelson: I think `updateCalledAlleleCounts` may be more clear, and it highlights it that will be modifying the input map.  I still sort of think this should be two different functions.  One that gets the count and one that does the map updating, but if that is to inefficient this is fine.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
lbergelson: Is chromosome counts established nomenclature that I'm not aware of?  It seems like allele counts is more accurate to me since this is all per allele.

ronlevine: Kind of. There are the classes `ChromosomeCounts` and `ChromosomeCountConstants`.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: I'd like to keep the logic as is and not break it into tow functions. The code is more efficient as is.

lbergelson: That's reasonable.  It's effectively private to the class so no one is going to use it wrong accidentally hopefully.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
lbergelson: So there are.  Wait on doing those renames until I talk with a few people here about what the right name is.  I don't want it to be inconsistent, but I think it's confusing how it is at the moment.
1.  Also, I noticed that VariantContextUtils has `calculateChromosomeCounts` which seems to be very similar to the new methods in this pr.  Did we know about those going in to these changes?   Are we duplicating code that already exists?  
2.  I also noticed that ChromosomeCountConstants seems to be totally redundant with ChromosomeCounts.  Could you remove ChromoSomeCountConstants while you're in this code?  It looks like add ChromomeCounts should probably just addAll ChromosomeCounts.getDescriptions as well..

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
```diff
@@ -323,35 +329,27 @@ private void filter(final VariantContext vc, final FeatureContext featureContext
...
+     * Get the genotype filters
+     *
+     * @param vc the variant context
+     * @param g the genotype
+     * @return list of genotype filter names
+     */
+    private List<String> getGenotypeFilters(final VariantContext vc, final Genotype g) {
+        final List<String> filters = new ArrayList<>();
+        if (g.isFiltered()) {
+            filters.add(g.getFilters());
```
lbergelson: `getFilters` returns a single string which is going to cause subtle sorting issues when combined with the jexl filters.  GenotypeBuilder.filters() sorts the list of strings if there are multiple strings, but it will have a single block of existing filters and then separate jexl ones, so it won't end up sorting correctly.

Could you split the filters so that they will end up sorting correctly?  You could run the jexl ones first and then only split if any of those were added for efficiency if you think it's necessary.

lbergelson:
@ronlevine Sorry for the slow response.  I like it much more with the refactoring you did.  I had a few minor comments.  I want to ask @vdauwera  about the names on monday, I find them confusing, but if it's standard nomenclature then we should keep them.  So don't do any renaming until I get back to you about that I guess.

lbergelson:
@ronlevine I consulted with Geraldine.  Keep calling things chromosome counts.  That seems to be the accepted name.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: Well, I'll make it `addChromosomeCountsToHeader`.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: It's effectively private, just made protected for testing. Will remove the access level and use @VisibleForTesting.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
ronlevine: 1. Removed `getCalledChromosomeCounts` and `updateChromosomeCountsInfo` and replaced the `VariantContextUtils .calculateChromosomeCounts`.
2. Removed  `ChromoSomeCountConstants`.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
lbergelson: 👍

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
ronlevine: Paraphrased and embellished your comment.

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
ronlevine: That is subtle. Changed to:
`filters.addAll(Arrays.asList(g.getFilters().split(";")));`

ronlevine:
@lbergelson Incorporated your comments. Back to you.

lbergelson:
@ronlevine Looks good to me.  Feel free to merge after squashing/rebasing if there aren't any issues with the rebase.  Ignore the codacy failure, it's something I was testing but I don't think it's that useful.

Issue #1609, opened 2016-03-22
Automate performance testing (including Spark)


akiezun:
the goal is to have automated performance tests for representative commandlines and datasizes - details of when to run them etc remain to be decided

akiezun:
@lbergelson let's chat with @davidbernick and figure out how to set this up (we need spark ones too)

akiezun:
i will send the right commandlines and accepted times to @davidbernick

akiezun:
Let's start with those 4. You need to fetch and build GATK by clone and `./gradlew clean installAll` and get the data from the location I indicated here. For accepted times, let's first run it a few times and we'll see.

Assigning to @davidbernick 

Non-spark Commandline for BaseRecalibrator (you may want to copy the input files onto a local drive)

```
bamIn=/dsde/working/akiezun/data/CEUTrio.HiSeq.WEx.b37.NA12892.bam
vcfIn=/dsde/working/akiezun/data/dbsnp_138.b37.excluding_sites_after_129.vcf
refIn=/dsde/working/akiezun/data/human_g1k_v37.fasta
./gatk-launch BaseRecalibrator -I $bamIn  -O output.table --useOriginalQualities --knownSites $vcfIn -R $refIn
```

Non-spark Commandline for ApplyBQSR  (you may want to copy the input files onto a local drive)
```
bamIn=/dsde/working/akiezun/data/CEUTrio.HiSeq.WEx.b37.NA12892.bam
refIn=/dsde/working/akiezun/data/human_g1k_v37.fasta
recaFile=/dsde/working/akiezun/data/CEUTrio.HiSeq.WEx.b37.NA12892.noIndelsGATK4.recal
./gatk-launch ApplyBQSR -R $refIn -I $bamIn --useOriginalQualities -O out.bam -bqsr $recalFile --emit_original_quals
```

Spark Commandline for MarkDuplicatesSpark 
```
bamIn=/user/akiezun/CEUTrio.HiSeq.WEx.b37.NA12892.readnamesort.bam 
#Note: bamIn is the same as gs://hellbender/test/resources/benchmark/CEUTrio.HiSeq.WEx.b37.NA12892.readnamesort.bam copied to HDFS
./gatk-launch MarkDuplicatesSpark \
	      --shardedOutput true \
	      -O /user/akiezun/tmp.md.bam \
	      --numReducers 0 \
	      --apiKey <THE_API_KEY> \
	      -I $bamIn \
	      -- \
[...]

PR #1874, opened 2016-06-01
set Spark reducer chunk size to 64MB by default - speeds up our pipelines

akiezun:
@gspowley can you review and/or delegate to Eric and Lucy?

I confirmed the speedup on MarkDuplicatesSpark - from 9:23 down to 8:54 seconds


droazen:
Have you confirmed that this doesn't cause problems for `BaseRecalibratorSpark`? I recall that BQSR was the reason the chunk size was set so low originally.

akiezun:
good point

akiezun:
that reemphasizes that we need https://github.com/broadinstitute/gatk/issues/1609 so bad.

gspowley:
Lucy reviewed and said the change looks correct. She is going to test the change on her cluster to verify it is doing what they expect. I'll post her update later today or tomorrow.

gspowley:
Lucy said the changes look good.

akiezun:
this is on hold - the readspipeline is slower on the cloud with this change (using SSDs). I'll close for now

PR #1878, opened 2016-06-05
Provide option for not capturing stdio of programs launched via ProcessBuilder

SHuang-Broad:
This affects `RunBWAMEMViaCommandLine.java` and `RunSGAViaProcessBuilderOnSpark.java`, where stdio are always captured.
Capturing BWA MEM's stdout is a must because the result is piped to stdout.
For SGA, this helps debugging. But this also slows down the performance.

Hence an option would be useful.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
@@ -62,6 +62,11 @@ public final class RunSGAViaProcessBuilderOnSpark extends GATKSparkTool {
               optional  = true)
     public boolean runCorrectionSteps = false;
 
+    @Argument(doc       = "If true, capture the stdout and stderr of the SGA processes along with the assembly output (valuable for debugging but hurts performance)",
+              shortName = "captureStdIO",
+              fullName  = "captureSGAStdOutAndError",
+              optional  = true)
+    public boolean enableSTDIOCapture = false;
 
     @Override
     public void runTool(final JavaSparkContext ctx){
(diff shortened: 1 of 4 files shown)


droazen:
@cwhelan Can you review?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -62,14 +62,19 @@
               optional  = true)
     public boolean runCorrectionSteps = false;
 
+    @Argument(doc       = "To enable stdout & stderr capture of the SGA processes or not.",
```
cwhelan: I'd change the wording of the doc and arg names a little bit. Maybe something like:

doc = "If true, capture the stdout and stderr of the SGA processes along with the assembly output (valuable for debugging but hurts performance)"
shortName=captureStdIO
fullName=captureSGAStdOutAndError

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSparkUnitTest.java
```diff
@@ -95,12 +96,40 @@ public void writeToLocalTest() throws IOException{
         Assert.assertTrue( FileUtils.contentEqualsIgnoreEOL(copied._2(), originalFASTQFile, StandardCharsets.UTF_8.name()) );
     }
 
+    // Test if disabling stdout stderr works
+    @Test(groups = "sv")
+    public void stdioCaptureTest() throws IOException{
```
cwhelan: Do any of the existing tests actually check that the standard IO is captured when captureStdio is true? I don't see where that is.. If not please add one (you could just add assertions to an existing test, or expand this test to run again with the param set to true).

cwhelan:
Looks pretty straightforward, just a couple of minor comments. Back to you @SHuang-Broad

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSparkUnitTest.java
SHuang-Broad: I did try that once, but the process spawned don't produce stderr (normally, unless something went wrong). So I skipped that half of the test. On second thought, I could deliberately let the program fail (e.g. providing wrong arguments, ask to log), and see if stdout and stderr is captured.
Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
SHuang-Broad: Done.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSparkUnitTest.java
SHuang-Broad: Now done with added test, though admittedly `stdout` is not captured (the process doesn't produce stdout results); but stderr is indeed captured correctly.

cwhelan:
Looks good to me, feel free to rebase and merge.

PR #1900, opened 2016-06-08
Command line read filters for walkers and Spark tools.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/589 (Walkers and Spark tools only, not integrated with Picard tools).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/ArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/ArgumentCollection.java
@@ -13,4 +13,12 @@ import java.lang.annotation.*;
 public @interface ArgumentCollection {
     /** Text that appears for this group of options in text describing usage of the command line program. */
     String doc() default "";
+
+    // The arguments in this collection are dependent on (and only enabled with the presence of)
+    // another argument. This may be used alone or together with dependsOnValue.
+    String dependsOnArgument() default "";
+
+    // The arguments in this collection are dependent on (and only enabled with the presence of)
+    // another argument with this specific value.
+    String dependsOnValue() default "";
 }
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReadFilterArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReadFilterArgumentCollection.java
@@ -1,0 +1,251 @@
+package org.broadinstitute.hellbender.cmdline.argumentcollections;
+
+import htsjdk.samtools.SAMFileHeader;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.broadinstitute.hellbender.cmdline.*;
+import org.broadinstitute.hellbender.engine.filters.*;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.Utils;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.function.BiFunction;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * {@link org.broadinstitute.hellbender.cmdline.ArgumentCollection} for allowing read filters to
+ * be enabled/disabled on the command line.
+ */
+public class ReadFilterArgumentCollection implements ArgumentCollectionDefinition {
+    private static final Logger logger = LogManager.getLogger(ReadFilterArgumentCollection.class);
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Enum of read filters that can be enabled/disabled from the command line
+     */
+    public enum CommandLineReadFilter implements Serializable {
+        /**
+         * No-arg filters: these filters have no arguments and one static instance,
+         * so the supplier function just returns the static instance and ignores
+         * the hdr and args arguments
+         */
+        ALLOW_ALL                   ((hdr, args) -> ReadFilterLibrary.ALLOW_ALL_READS),
+        CIGAR_IS_SUPPORTED          ((hdr, args) -> ReadFilterLibrary.CIGAR_IS_SUPPORTED),
+        GOOD_CIGAR                  ((hdr, args) -> ReadFilterLibrary.GOOD_CIGAR),
+        HAS_READ_GROUP              ((hdr, args) -> ReadFilterLibrary.HAS_READ_GROUP),
+        HAS_MATCHING_BASES_AND_QUALS((hdr, args) -> ReadFilterLibrary.HAS_MATCHING_BASES_AND_QUALS),
+        MAPPED                      ((hdr, args) -> ReadFilterLibrary.MAPPED),
+        MAPPING_QUALITY_AVAILABLE   ((hdr, args) -> ReadFilterLibrary.MAPPING_QUALITY_AVAILABLE),
+        MAPPING_QUALITY_NOT_ZERO    ((hdr, args) -> ReadFilterLibrary.MAPPING_QUALITY_NOT_ZERO),
+        MATE_ON_SAME_CONTIG_OR_NO_MAPPED_MATE((hdr, args) -> ReadFilterLibrary.MATE_ON_SAME_CONTIG_OR_NO_MAPPED_MATE),
+        MATE_DIFFERENT_STRAND       ((hdr, args) -> ReadFilterLibrary.MATE_DIFFERENT_STRAND),
+        NOT_DUPLICATE               ((hdr, args) -> ReadFilterLibrary.NOT_DUPLICATE),
+        PRIMARY_ALIGNMENT           ((hdr, args) -> ReadFilterLibrary.PRIMARY_ALIGNMENT),
+        PASSES_VENDOR_QUALITY_CHECK ((hdr, args) -> ReadFilterLibrary.PASSES_VENDOR_QUALITY_CHECK),
+        READLENGTH_EQUALS_CIGARLENGTH ((hdr, args) -> ReadFilterLibrary.READLENGTH_EQUALS_CIGARLENGTH),
+        SEQ_IS_STORED               ((hdr, args) -> ReadFilterLibrary.SEQ_IS_STORED),
+        VALID_ALIGNMENT_START       ((hdr, args) -> ReadFilterLibrary.VALID_ALIGNMENT_START),
+        VALID_ALIGNMENT_END         ((hdr, args) -> ReadFilterLibrary.VALID_ALIGNMENT_END),
+
+        /**
+         * Read filters with user-defined arguments. The supplier function validates the
+         * user provided values and returns the already-populated filter embedded in the
+         * argument collection.
+         *
+         * For read filters with user-defined arguments that also require a SAMFileHeader,
+         * the supplier function creates a new instance of the read filter using the
+         * given header, populating the other args from the instance embedded in
+         * the read filter argument collection; updates the embedded filter with the
+         * new one and returns it.
+         */
+        ALIGNMENT_AGREES_WITH_HEADER((hdr, args) -> new AlignmentAgreesWithHeaderReadFilter(hdr)),
+        FRAGMENT_LENGTH((hdr, args) -> {
+            validateArgs(CommandLineReadFilter.valueOf("FRAGMENT_LENGTH"), args.fragmentLengthFilter::validate);
+            return args.fragmentLengthFilter;
+        }),
+        LIBRARY((hdr, args) -> {
+            validateArgs(CommandLineReadFilter.valueOf("LIBRARY"), args.libraryFilter::validate);
+            LibraryReadFilter lrf = new LibraryReadFilter(hdr);
+            lrf.libraryToKeep = args.libraryFilter.libraryToKeep;
+            // update the actual argument collection filter just in case...
+            args.libraryFilter = lrf;
+            return lrf;
+        }),
+        MAPPING_QUALITY((hdr, args) -> {
+            validateArgs(CommandLineReadFilter.valueOf("MAPPING_QUALITY"), args.mappingQualityFilter::validate);
+            MappingQualityReadFilter mqf = new MappingQualityReadFilter(args.mappingQualityFilter.minMappingQualityScore);
+            // update the actual argument collection filter just in case...
+            args.mappingQualityFilter = mqf;
+            return mqf;
+        }),
+        PLATFORM((hdr, args) -> {
+            validateArgs(CommandLineReadFilter.valueOf("PLATFORM"), args.platformFilter::validate);
+            PlatformReadFilter pf = new PlatformReadFilter(hdr);
+            pf.PLFilterNames = args.platformFilter.PLFilterNames;
+            // update the actual argument collection filter just in case...
+            args.platformFilter = pf;
+            return pf;
+        }),
+        PLATFORM_UNIT((hdr, args) -> {
+            validateArgs(CommandLineReadFilter.valueOf("PLATFORM_UNIT"), args.platformUnitFilter::validate);
+            PlatformUnitReadFilter pfuf = new PlatformUnitReadFilter(hdr);
+            pfuf.blackListedLanes = args.platformUnitFilter.blackListedLanes;
+            // update the actual argument collection filter just in case...
+            args.platformUnitFilter = pfuf;
+            return pfuf;
+        }),
+        READ_GROUP((hdr, args) -> {
+            validateArgs(CommandLineReadFilter.valueOf("READ_GROUP"), args.readGroupFilter::validate);
+            return args.readGroupFilter;
+        }),
+        READ_GROUP_BLACK_LISTED((hdr, args) -> {
+            validateArgs(CommandLineReadFilter.valueOf("READ_GROUP_BLACK_LISTED"), args.readGroupBlackListFilter::validate);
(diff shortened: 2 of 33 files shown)


src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -65,20 +63,44 @@ public boolean includeNs() {
...
+     * default implementation uses the {@link WellformedReadFilter} and {@link ReadFilterLibrary.MAPPED}filter
+     * with all default options. Subclasses can override to provide alternative filters.
+     * @return List of individual filters to be applied for this tool.
      */
-    public CountingReadFilter makeReadFilter(){
-        return new CountingReadFilter("Wellformed", new WellformedReadFilter(getHeaderForReads()))
-                .and(new CountingReadFilter("Mapped", ReadFilterLibrary.MAPPED));
+    public List<ReadFilterArgumentCollection.CommandLineReadFilter> getDefaultReadFilters() {
+        final List<ReadFilterArgumentCollection.CommandLineReadFilter> defaultFilters = new ArrayList<>();
+        defaultFilters.add(ReadFilterArgumentCollection.CommandLineReadFilter.WELLFORMED);
```
magicDGS: Mapper filter was included in the previous behavior.

cmnbroad: Will fix - good catch thx.

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineParserTest.java
```diff
@@ -693,6 +693,71 @@ public void testSpecialFlags(){
...
+    public void testFailsDependsOnArgumentValue() {
+        TestArgs tArgs = new TestArgs();
+        String args[] = {
+                "--master", "0",
+                "--argAndValueTarget", "foo" // depends on --master 27
+        };
+        final CommandLineParser clp = new CommandLineParser(tArgs);
+        clp.parseArguments(System.err, args);
+    }
+
```
cmnbroad: Need to add a test to verify that the dependsOn argument works with both the short and long name of the argument.

cmnbroad:
Two comments for the reviewer. 1) The ArgumentCollection "dependsOnArgument" attritbutes were added so the command line parser can recognize and complain when the user specifies the argument for a read filter, but not the filter itself, i.e., specifying a maxReadLength value without enabling the corresponding readLength filter. 2) In this PR, the pipeline tools honor the user's command line filters, including disabled filters, for the whole pipeline - are there cases where that isn't the right thing to do ?

magicDGS:
@cmnbroad, I'm thinking about a use case for no filters allowed for the user of the tool. Imagine that you want to pull down some evidence for discordant read pairs (for instance, for SV), and the user provides a read filter which removes this kind of signatures.

Could it be possible to generate an interface for the `ReadFilterArgumentCollection` and implementations for no-exposed/exposed to the final user?

This will be similar to the optional/required arguments for reads, reference and so on...

src/main/java/org/broadinstitute/hellbender/cmdline/ArgumentCollection.java
```diff
@@ -13,4 +13,12 @@
 public @interface ArgumentCollection {
     /** Text that appears for this group of options in text describing usage of the command line program. */
     String doc() default "";
+
+    // The arguments in this collection are dependent on (and only enabled with the presence of)
+    // another argument. This may be used alone or together with dependsOnValue.
+    String dependsOnArgument() default "";
+
+    // The arguments in this collection are dependent on (and only enabled with the presence of)
+    // another argument with this specific value.
```
lbergelson: Can this be specified without dependsOnArgument?

src/main/java/org/broadinstitute/hellbender/cmdline/ArgumentCollection.java
```diff
@@ -13,4 +13,12 @@
 public @interface ArgumentCollection {
     /** Text that appears for this group of options in text describing usage of the command line program. */
     String doc() default "";
+
+    // The arguments in this collection are dependent on (and only enabled with the presence of)
+    // another argument. This may be used alone or together with dependsOnValue.
```
lbergelson: It's a bit strange that these are at the ArgumentCollection level instead of on individual arguments.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReadFilterArgumentCollection.java
```diff
@@ -0,0 +1,251 @@
...
+            return true;
+        }
+    }
+
+    @Argument(fullName = StandardArgumentDefinitions.READ_FILTER_LONG_NAME,
+            shortName = StandardArgumentDefinitions.READ_FILTER_SHORT_NAME,
+            doc = "Filters to apply to reads before analysis",
+            common = true,
+            optional = true)
+    public List<CommandLineReadFilter> enabledCommandLineFilters = new ArrayList<>();
```
lbergelson: How do these interact?  If I enable and disable a filter what happens? Would it be simpler to have 2 arguments, `--readFilter` and `--disableDefaultReadFilters`.  Anything specified in readFilter would be added to the default filters unless `disableDefaultReadFilters` was used.  It would be unambiguous behavior at the expense of having to specify most of the read filters explicitly if you wanted to disable just 1 of them.

src/main/java/org/broadinstitute/hellbender/engine/filters/CommandLineFilter.java
```diff
@@ -0,0 +1,10 @@
+package org.broadinstitute.hellbender.engine.filters;
+
+import java.io.Serializable;
+
+/**
+ * Interface implemented by command-line accessible read filters for argument validation.
+ */
+public interface CommandLineFilter extends Serializable {
```
lbergelson: This name is confusing if you don't know what it means.  Maybe "FilterWithArguments" or something like that?

src/main/java/org/broadinstitute/hellbender/engine/filters/FragmentLengthReadFilter.java
```diff
@@ -20,4 +27,15 @@ public boolean test( final GATKRead read ) {
         //Note fragment length is negative if mate maps to lower position than read so we take absolute value.
         return Math.abs(read.getFragmentLength()) <= maxFragmentLength;
     }
+
+    @Override
+    public String validate() {
+        String message = null;
+        if (maxFragmentLength < 1) {
+            message = "requires a non-zero value for \"" + maxLengthArgName + "\"";
```
lbergelson: Seems like this should be `requires a positive value`

src/main/java/org/broadinstitute/hellbender/engine/filters/PlatformReadFilter.java
```diff
@@ -5,17 +5,24 @@
...
  */
-public final class PlatformReadFilter implements ReadFilter {
+public final class PlatformReadFilter implements ReadFilter, CommandLineFilter, Serializable {
     private static final long serialVersionUID = 1L;
-    @Argument(fullName = "PLFilterName", shortName = "PLFilterName", doc="Keep reads with RG:PL attribute containing this string", optional=true)
-    public Set<String> PLFilterNames;
+
+    private final static String platformArgName = "PLFilterName";
+
+    @Argument(fullName = platformArgName, shortName = "PLFilterName", doc="Keep reads with RG:PL attribute containing this string", optional=true)
```
lbergelson: @cmnbroad It seems like the case in the validate here should be checked automatically if this was optional=false.  Is this due to the bug you were talking about before?  Or is there a problem with required arguments in the read filters?

src/main/java/org/broadinstitute/hellbender/engine/spark/GATKSparkTool.java
```diff
@@ -275,14 +278,46 @@ private boolean hasCramInput() {
...
+     * Multiple filters can be composed by using {@link org.broadinstitute.hellbender.engine.filters.ReadFilter}
+     * composition methods.
      */
     public ReadFilter makeReadFilter() {
-        return new WellformedReadFilter(getHeaderForReads());
+        final ReadFilterArgumentCollection rfc = readArguments.getReadFilterArgumentCollection();
+        return rfc.getMergedCommandLineFilters(getDefaultReadFilters())
+                .stream()
+                .map(f -> f.getFilterInstance(getHeaderForReads(), rfc))
+                .reduce(new WellformedReadFilter(getHeaderForReads()), (f1, f2) -> f1.and(f2));
```
lbergelson: Why is WellFormedReadFilter included specially here?

lbergelson: Code that looks like this shows up in a bunch of places.  Would it be possible to abstract it into ReadFilterArgumentCollection?  It looks like sometimes you need a CountingReadFilter and sometimes you need the read header, but is there any commonality that could be pulled up?

cmnbroad:
@magicDGS I could certainly do that (split out an OptionalReadFilterArgumentCollection), though we'd then need to add a "requiresReadFilters" method to determine which to use. I guess it depends on how common that case would be. An simple alternative would be to just override makeReadFilter and reject any command line filter requests or do any custom filter handling.

src/main/java/org/broadinstitute/hellbender/cmdline/ArgumentCollection.java
cmnbroad: No - "dependsOnValue" only makes sense if "dependsOnArgument" is specified, since it further restricts the dependency. It means "this argument collection depends on the argument(option) specified in "dependsOnArgument", but only when its specified with the value in "dependsOnValue".  So, f you had a "--debug" option that was just a boolean switch, you might use an argument collection with "dependsOnArgument=debug", but, if your "--debug" argument is an enum, you might have an argument collection that "dependsOnArgument=debug" and "dependsOnValue=3".  Perhaps there re better names ? In jopt parlance they would be "dependsOnOption" and "dependsOnArgument".

cmnbroad: Either way we need a test for this case.

src/main/java/org/broadinstitute/hellbender/cmdline/ArgumentCollection.java
cmnbroad: We could add it for Argument as well - I just didn't want to do it for this PR since I think in this case it made more sense to use ArgumentCollection. If I *only* did it for Argument, it would mean that there would be references to the string "--readFilter" or "RF" inside of all the individual read filter classes.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReadFilterArgumentCollection.java
cmnbroad: Somehow I got the idea that we wanted to support disabling of individual default filters (I may have gotten that idea from forum postings since it has been discussed there some). But we could do what you suggest. Currently the code that resolves the defaults with the command line filters has a comment that says that enabled wins if there is a conflict - but I picked that arbitrarily, and if we keep it we should probably log/warn when that happens. I'm open to suggestions.

src/main/java/org/broadinstitute/hellbender/engine/filters/CommandLineFilter.java
cmnbroad: Agreed - it evolved from its original purpose. I think maybe "ValidatingFilter" since it basically adds a validate method, and its not really read filter specific.

src/main/java/org/broadinstitute/hellbender/engine/spark/GATKSparkTool.java
cmnbroad: Right - WellFormed should really be ALLOW_ALL here since its just the default accumulator for the reduction step.

I originally had this as a method in ReadFilterArgumentCollection, but you had to pass in a lambda to wrap the filters to accommodate the CountingReadFilter case, and it seemed a bit complicated for not much reason.  I think I'll put it back though since it would eliminate things like using WellFormed when you shouldn't....

src/main/java/org/broadinstitute/hellbender/cmdline/ArgumentCollection.java
lbergelson: I think the comment should reflect that, and I agree we need a test.

src/main/java/org/broadinstitute/hellbender/cmdline/ArgumentCollection.java
lbergelson: Lets consider adding it to argument later.  I agree for this case it makes more sense to do it this way.

magicDGS:
If you override `makeReadFilter` in a tool with this arguments and the user provides them, either the developer should code a warning message saying that read filters will be ignored or the user will not be aware of what is happening in the processing.

@cmnbroad, I think that it could be a good idea to add a `requiresReadFilters` to `GATKTool`, and like that tools which requires reads does not need to implement all this even if they are not extending `ReadWalker` or `LocusWalker`, for instance. What do you think?

src/main/java/org/broadinstitute/hellbender/engine/filters/PlatformReadFilter.java
cmnbroad: @lbergelson If I make this argument optional=false, then its ALWAYS required, whether you specify the corresponding filter or not (I haven't been able to reproduce the bug I mentioned the other day). Instead, its optional=true, but contained inside of an argument collection that is included with a dependsOnArgument="RF" and dependsOnValue=PLPlatformFilter", which gives the right result (it complains only when you set the argument value but don't specify the filter, or when you specify the filter but don't set the filter argument).

lbergelson: How much more work would it be to have optional be respected, but have the argument be disabled if it dependsOn something that isn't set?

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
```diff
@@ -98,15 +99,42 @@ private SimpleInterval getReadInterval(final GATKRead read) {
...
+     * Default tool implementation of {@link #traverse()} calls this method once before iterating
+     * over the reads and reuses the filter object to avoid object allocation. Nevertheless, keeping state in filter
+     * objects is strongly discouraged.
      *
-     * Subclasses can extend to provide own filters (ie override and call super).
-     * Multiple filters can be composed by using {@link org.broadinstitute.hellbender.engine.filters.ReadFilter} composition methods.
+     * Multiple filters can be composed by using {@link org.broadinstitute.hellbender.engine.filters.ReadFilter}
+     * composition methods.
      */
     public CountingReadFilter makeReadFilter(){
```
lbergelson: Should this be made final? It seems like it would be easy to override this unintentionally and incorrectly and thus break the read filter behavior.

lbergelson: (especially because the meaning of this method is changing and we have lots of code that does override it already..)

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -6,14 +6,14 @@
 
 /**
  * Standard ReadFilters
+ *
+ * TODO: add a note here about how to make these accessible to the command line
```
lbergelson: this needs resolving

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
cmnbroad: If we make it final, then tools will have no way to modify the default behavior, i.e., if a tool absolutely requires that some particular filter never be disabled, or doesn't want the user to modify any of the defaults, etc.

src/main/java/org/broadinstitute/hellbender/engine/filters/PlatformReadFilter.java
lbergelson: It may be possible to leverage jotp's availableIf to do this more easily, see also requiredIf and requiredUnless http://pholser.github.io/jopt-simple/examples.html

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
lbergelson: I thought filters that must never be disabled were the default filters, and overriding them was moving into "I know more than the developers and definitely want to do this" territory?

src/main/java/org/broadinstitute/hellbender/engine/filters/PlatformReadFilter.java
cmnbroad: That looks pretty good - I'll try hooking it up to dependsOn.

droazen:
Discussed in person with @cmnbroad. We decided to go with a different approach that avoids the need for a central registry + connecting code, would be 100% pluggable, and would be easily extensible to variant annotations if we want to add `@Argument` capability there as well:

-replace lamdba read filters with explicit class definitions
-patch the argument system to add the generic ability to discover arguments within plugin classes, and pass on populated instances of those plugin classes to the tool (likely involves only a very small amount of additional code, based on our reading of `CommandLineParser`)

cmnbroad:
Closing in favor of https://github.com/broadinstitute/gatk/pull/1973.

Issue #1427, opened 2016-01-11
Move classes of general utility (new walker types, utilities for working with particular formats, etc.) from protected to public


droazen:
We want the public GATK framework to contain all the generic capabilities for loading and traversing data in various formats -- all new walker types, general-purpose utilities for working with different formats, etc. should ideally live in the framework rather than in protected. Moving these classes to public should also improve code quality by preventing general-purpose code from getting embedded within particular tools, and would encourage the development of clear APIs for using this code.

@akiezun since you love to rip code out, you get this one :) You should consult with @LeeTL1220 and @vruano to identify which classes should be moved. If the classes to be moved have already been adequately reviewed and are covered by tests, they shouldn't require a re-review (though we can do one if @LeeTL1220 requests it).

droazen:
For @akiezun (in consultation with @LeeTL1220)

akiezun:
this is probably best done by @vruano when moving germline CNV code to public.

akiezun:
@droazen @vruano what's the plan there? Is the code stable enough to start moving pieces?
FeatureWalker and TargetWalker seem to be stable, as is the codecs package, SVD+PCA

droazen:
@akiezun Let's do this after the HaplotypeCaller is merged/moved, and https://github.com/broadinstitute/gatk/issues/1676 is done.

akiezun:
i'll grab this

akiezun:
this will continue. moving to alpha-3

akiezun:
closing this - i think PCA and TSV is enough for now

PR #1901, opened 2016-06-09
move PCA and hdf5 from protected

akiezun:
fixed another part of #1427 

notes: 
PCA was moved to the SVD package rather than its own (1 class in a package was a bad smell)
No PoN-related code was moved, only generic HDF5 code


akiezun:
@lbergelson   can you review?

lbergelson:
do we want this in protected?  It means people have to go through arcane installation procedures...

akiezun:
most people get protected. I'd rather push the pain down to public (unless
it doubles the pain and both repos need this).

Oh btw, perhaps we should spin off the HDF5 to a separate project that we
would just inject in gatk
wdyt?

lbergelson:
+1 to spinning off HDF5.  We should package it and make it available as a jar with the native code included.  I'm pretty sure we can do so legally.

lbergelson:
there's an open ticket https://github.com/broadinstitute/gatk-protected/issues/436 to spin off hdf5

akiezun:
i'll close this and spinoff hdf5

PR #1906, opened 2016-06-09
public changes to support GenotypeGVCFs

lbergelson:
These are the changes to public necessary to support my initial port of genotypeGVCFs.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyFailOverAFCalculatorProvider.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyFailOverAFCalculatorProvider.java
@@ -1,0 +1,41 @@
+package org.broadinstitute.hellbender.tools.walkers.genotyper.afcalc;
+
+import org.broadinstitute.hellbender.tools.walkers.genotyper.GenotypeCalculationArgumentCollection;
+import org.broadinstitute.hellbender.utils.Utils;
+
+/**
+ * Provider that defaults to the general ploidy implementation when the preferred one does not handle the required
+ * ploidy.
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+public final class GeneralPloidyFailOverAFCalculatorProvider extends AFCalculatorProvider {
+
+    private final AFCalculator preferred;
+    private final AFCalculatorImplementation preferredImplementation;
+    private final AFCalculator failOver;
+
+    /**
+     * Creates a new AF calculator provider given the genotyping arguments and logger reference.
+     * @param genotypeArgs genotyping parameter collection.
+     * @throws IllegalArgumentException if {@code genotypeArgs} is {@code null}.
+     */
+    public GeneralPloidyFailOverAFCalculatorProvider(final GenotypeCalculationArgumentCollection genotypeArgs) {
+        Utils.nonNull(genotypeArgs);
+        preferredImplementation = AFCalculatorImplementation.bestValue(genotypeArgs.samplePloidy,genotypeArgs.MAX_ALTERNATE_ALLELES, null);
+        preferred = preferredImplementation.newInstance();
+        failOver = AFCalculatorImplementation.EXACT_GENERAL_PLOIDY.newInstance();
+    }
+
+    /**
+     * {@inheritDoc}
+     * @param ploidy {@inheritDoc}
+     * @param maximumAlternativeAlleles {@inheritDoc}
+     * @return {@inheritDoc}
+     */
+    @Override
+    public AFCalculator getInstance(final int ploidy, final int maximumAlternativeAlleles) {
+        return preferredImplementation.usableForParams(ploidy,maximumAlternativeAlleles) ? preferred : failOver;
+    }
+
+}
--- a/src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
@@ -5,9 +5,6 @@ import com.google.cloud.dataflow.sdk.options.PipelineOptionsFactory;
 import com.google.cloud.genomics.dataflow.utils.GCSOptions;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.util.Log;
-import htsjdk.variant.variantcontext.Genotype;
-import htsjdk.variant.variantcontext.VariantContext;
-import htsjdk.variant.vcf.VCFConstants;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.hellbender.engine.AuthHolder;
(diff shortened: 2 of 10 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyFailOverAFCalculatorProvider.java
```diff
@@ -0,0 +1,41 @@
...
+import org.broadinstitute.hellbender.tools.walkers.genotyper.GenotypeCalculationArgumentCollection;
+import org.broadinstitute.hellbender.utils.Utils;
+
+/**
+ * Provider that defaults to the general ploidy implementation when the preferred one does not handle the required
+ * ploidy.
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+public class GeneralPloidyFailOverAFCalculatorProvider extends AFCalculatorProvider {
```
akiezun: final class

src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
```diff
@@ -385,11 +393,29 @@ else if ( Double.isInfinite(expected) ) // NaN == NaN => false unfortunately
...
-        } else
-            Assert.assertEquals(actual, expected, "Attribute " + key);
+            assertEqualsDoubleSmart((Double) notationCorrectedActual, (Double) notationCorrectedExpected, 1e-2, "Attribute " + key);
+        } else {
+            Assert.assertEquals(notationCorrectedActual, notationCorrectedExpected, "Attribute " + key);
+        }
+    }
+
+    /**
+     * This is some  necessary to deal with the fact that variant context attributes are deserialized from vcf as Strings
```
akiezun: `some` -> `sometimes` (?)

akiezun: why not put this in Utils? that way would not have to create BaseTestUnitTest

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyFailOverAFCalculatorProviderUnitTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * Tests {@link GeneralPloidyFailOverAFCalculatorProvider}
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+public class GeneralPloidyFailOverAFCalculatorProviderUnitTest {
```
akiezun: `extends BaseTest` as a way to track all tests

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyFailOverAFCalculatorProviderUnitTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Test(dataProvider="nonThreadSafeConstructorsData")
+    public void testNonThreadSafeConstructors(final int ploidy, final int maxAltAlleles) {
```
akiezun: do we have 'thread safe' constructors?

lbergelson:
@akiezun Do you think we should move the various variant context test routines out of BaseTest?  It seems weird that they live there.

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyFailOverAFCalculatorProviderUnitTest.java
akiezun: can you rename this test and data provider? current names make no sense i think

src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
lbergelson: Should we move these variantcontext methods as well?  It seems like they should all go somewhere else.

akiezun: I think so.

src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
```diff
@@ -403,18 +429,18 @@ public static void assertGenotypesAreEqual(final Genotype actual, final Genotype
...
         Assert.assertEquals(actual.hasPL(), expected.hasPL(), "Genotype hasPL");
-        Assert.assertEquals(actual.hasAD(), expected.hasAD(), "Genotype hasAD");
-        Assert.assertEquals(actual.hasGQ(), expected.hasGQ(), "Genotype hasGQ");
-        Assert.assertEquals(actual.hasDP(), expected.hasDP(), "Genotype hasDP");
+        Assert.assertTrue(Arrays.equals(actual.getPL(), expected.getPL()));
 
         Assert.assertEquals(actual.hasLikelihoods(), expected.hasLikelihoods(), "Genotype haslikelihoods");
         Assert.assertEquals(actual.getLikelihoodsString(), expected.getLikelihoodsString(), "Genotype getlikelihoodsString");
         Assert.assertEquals(actual.getLikelihoods(), expected.getLikelihoods(), "Genotype getLikelihoods");
-        Assert.assertTrue(Arrays.equals(actual.getPL(), expected.getPL()));
```
akiezun: why is the getPL comparison removed?

src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
```diff
@@ -519,5 +547,42 @@ public static void assertEqualVariants(final List<VariantContext> v1, final List
         }
     }
 
+    public static void assertVariantContextsAreEqual( final VariantContext actual, final VariantContext expected, List<String> attributesToIgnore ) {
```
akiezun: i think this one is fine to stay here because it's all asserts

akiezun:
small edits to do - after that, LGTM. back to @lbergelson

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyFailOverAFCalculatorProviderUnitTest.java
lbergelson: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyFailOverAFCalculatorProviderUnitTest.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
lbergelson: @akiezun How do you feel about moving this and the various asserts in base test to a new `VariantContextTestUtils`?  It doesn't feel like this method is generally applicable, and those other methods seem more complex and specialized than should really be in BaseTest

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyFailOverAFCalculatorProvider.java
lbergelson: done

PR #1918, opened 2016-06-15
Moved allele reduction out of AFCalculator and simplified code.

### LINKED ISSUE #1891: Remove extraneous methods from AFCalculator
`AFCalculator` has a couple of methods that don't belong: `reduceScope()`, which is called *before* the AF calculation to reduce its computational burden, and `subsetAlleles()`, which is called *after* the calculation to eliminate alleles that don't exist in called genotypes (in conformance with the dubious VCF spec).

These methods are chronologically distinct from the rest of `AFCalculator` and do not appear to use any private variables.  Thus they could easily be turned into static methods and removed from `AFCalculator`.

Furthermore, `reduceScope()` and `subsetAlleles()` each have two implementations, in `DiploidExactAFCalculator` and `GeneralPloidyExactAFCalculator`.  Since these two cases are complementary, they could easily be merged in a single method with an `if (ploidy == 2) . . .`

Finally, the general ploidy code is more complicated than it needs to be and needs editing.

Beyond general housekeeping, the main motivation here is to untangle the AF/qual code as much as possible *without* changing behavior before introducing the new model into the mix (issue #1697).

davidbenjamin:
Closes #1891.

The bulk of this PR involved extracting `subsetAlleles` and similar methods from `AFCalculator` into a new utility class.  These methods only had two independent implementations, a diploid one with code in `DiploidExactAFCalculator` and a general ploidy one in `GeneralPloidyExactAFCalculator`.

The code has been consolidated and thoroughly rewritten.


davidbenjamin:
@vruano can you review because 1) much of this was your idea and 2) I trust you to spot any dodgy architecture?  @droazen In addition to all gatk public tests passing (of course), all the HaplotyeCaller tests in gatk-protected pass.

davidbenjamin:
I'm pretty sure that the coverage did not decrease to 12%. . .

davidbenjamin:
@vruano This PR needs your ruthlessness.

droazen:
Assigning to @vruano

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.variant.GATKVCFConstants;
+import org.broadinstitute.hellbender.utils.variant.GATKVariantContextUtils;
+
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Created by davidben on 6/8/16.
```
vruano: This comments does not tell much about what the class actually does.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+ */
+public class AlleleSubsettingUtils {
```
vruano: What about make it ```final```.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+    /**
+     * From a given variant context, extract a given subset of alleles, and update genotype context accordingly,
+     * including updating the PL's, and assign genotypes accordingly
+     * @param vc                                variant context with alleles and genotype likelihoods
+     * @param defaultPloidy                     ploidy to use when {@code vc} is missing information for a sample
+     * @param allelesToKeep                     subset of alleles in {@code vc} to keep
+     * @param method                            whether to assign hard genotypes or leave as no-call
+     * @return                                  GenotypesContext with new PLs
+     */
+    public static GenotypesContext subsetAlleles(final VariantContext vc, final int defaultPloidy, final boolean isDiploid,
```
vruano: Instead of having an argument to allow the user to tell you is diploid, could you just check whether it is indeed diploid. If more efficiency you want to allow the user to by pass such a check you could make ```subsetDiploidAlleles``` public; the latter method should fail gracefully (IAE) if encounters something that is not diploid.

vruano: In such case ```subsetGeneralPloidyAlleles``` would be renamed to ```subsetAlleles`` and the ```isDiploid``` argument would be dropped. ```subsetDiploidAlleles``` would be kept as it is.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    /**
+     * Restrict a variant context to its "best" alleles.
+     * @param vc                    variant context with alleles and genotype likelihoods
+     * @param defaultPloidy         ploidy to assume in case that {@code vc} does not contain that information
+     * @param numAltAllelesToKeep
+     * @param isDiploid
+     * @return
+     */
```
vruano: mmm... this is a bit convoluted.. for example what is the actual difference between ```subsetAllele``` and ```reduceNumberOfAlleles```. Yes I know, it is the return type how you specify the alleles to keep .... yet It looks a bit wrong... On the one hand I would expect that these methods would share the name but differ in the signature yet they have different return types which complicate things... I think part of the problem is that Genotypes do not hold a reference to their enclosing VC thus the allele list. 

I need to think this through.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotpyeIterator.java
```diff
@@ -0,0 +1,108 @@
...
+ * of form int[] -> double (to be more precise, IntArrayWrapper -> Double).
+ * For a given ploidy (chromosome count) and number of alleles, we need a form to iterate deterministically
+ * across all possible allele conformations, that is, all possible ways in which N whole numbers will sum to P,
+ * where N is number of alleles and P is number of chromosomes.
+ * For example, with P=2,N=2, ploidy = 2 iterator will produce
+ * alleleCounts = [2 0 ] [1 1] [ 0 2]
+ *
+ *
+ */
+public final class GenotpyeIterator {
```
vruano: Class name typo.

vruano: This functionality is available thru ```GenotypeLikelihoodCalculator```. It does not provide an iterator per se but it caches the different allele-count combinations using ```GenotypeLikelihoodCalculator#genotypeAlleleCountsAt```.

In fact GenotypeAlleleCounts is actually capable of acting as an iterator (check on the protected increase  method). It is not public as to make GenotypeAlleleCounts mostly a constant object. You could change that but I rather you stick to use ```GenotypeLikelihoodCalculator#genotypeAlleleCountsAt```.

vruano: If you prefer the iterator approach you always add it to ```GenotypeLikelihoodCalculator``` and use ```GenotypeLikelihoodCalculator#genotypeAlleleCountsAt``` under the hood.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotpyeIterator.java
```diff
@@ -0,0 +1,108 @@
...
+        hasNext = true;
+    }
+
+    public int[] getAlleleCounts() { return alleleCounts; }
+    public int getLinearIndex() { return linearIndex; }
+    public boolean hasNext() { return hasNext; }
+
+    //TODO: maybe this class is the wrong place.  It kind of makes sense because this is the inverse of getAlleleCounts
+    //TODO: in a sense.
+    public static int getLinearIndex(final int[] alleleCounts, final int numAlleles, final int ploidy) {
```
vruano: This is also available in GenotypeLikelihoodsCalculator: Method ```alleleCountsToIndex```.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GLVectorSizeCache.java
```diff
@@ -0,0 +1,31 @@
+package org.broadinstitute.hellbender.tools.walkers.genotyper;
+
+import java.util.Arrays;
+
+/**
+ * a cache of the PL vector sizes (i.e. the number of possible unphased genotypes) as a function of # of alleles and ploidy
+ *
+ * Created by davidben on 6/9/16.
+ */
+public class GLVectorSizeCache {
```
vruano: It seems that this also would be covered by GenotypeLikelihoodsCalculator

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
vruano: Do we really need to do a distinction between diploid and non-diploid... is this a premature optimization?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+    public static VariantContext reduceNumberOfAlleles(final VariantContext vc, final int defaultPloidy,
```
vruano: Same here.

vruano: I would call this method also ```subsetAlleles```.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+    public static void updateGenotypeCallAfterSubsetting(final List<Allele> originalGT,
+                                                         final GenotypeBuilder gb,
+                                                         final GenotypeAssignmentMethod assignmentMethod,
+                                                         final double[] newLikelihoods,
+                                                         final List<Allele> allelesToUse) {
+        switch ( assignmentMethod ) {
+            case DO_NOT_ASSIGN_GENOTYPES:
+                break;
+            case SET_TO_NO_CALL:
+                gb.alleles(GATKVariantContextUtils.noCallAlleles(2)); // @TODO assumes diploid
```
vruano: In this case perhaps you need to add Diploid to the method name.

vruano: What is the problem with using originalGT.size() to get the ploidy?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+                gb.noGQ();
+                break;
+            case USE_PLS_TO_ASSIGN:
+                if ( newLikelihoods == null || !GATKVariantContextUtils.isInformative(newLikelihoods) ) {
+                    // if there is no mass on the (new) likelihoods, then just no-call the sample
+                    gb.alleles(GATKVariantContextUtils.noCallAlleles(2)); // @TODO assumes diploid
```
vruano: Same.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+                    // find the genotype with maximum likelihoods
+                    final int PLindex = MathUtils.maxElementIndex(newLikelihoods);
+                    GenotypeLikelihoods.GenotypeLikelihoodsAllelePair alleles = GenotypeLikelihoods.getAllelePair(PLindex);
+                    gb.alleles(Arrays.asList(allelesToUse.get(alleles.alleleIndex1), allelesToUse.get(alleles.alleleIndex2)));
+                    gb.log10PError(GenotypeLikelihoods.getGQLog10FromLikelihoods(PLindex, newLikelihoods));
+                }
+                break;
+            case BEST_MATCH_TO_ORIGINAL:
+                final List<Allele> best = new LinkedList<>();
+                final Allele ref = allelesToUse.get(0); // WARNING -- should be checked in input argument
```
vruano: Resolve this warning.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+                for ( final Allele originalAllele : originalGT ) {
+                    best.add(allelesToUse.contains(originalAllele) ? originalAllele : ref);
```
vruano: Maybe we should use Allele.NO_CALL when the allele is not in allelesToUse?

vruano: Or take the best genotype (using the likelihood) that contain a matching count for the alleles present in allelesToUse.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+     * @param defaultPloidy                     ploidy to assume in case that {@code vc} does not contain that information
+     *                                          for a sample.
+     * @param allelesToKeep                      alleles to subset
+     * @param method                            whether to assign hard genotypes or leave as no-call
+     * @return                                  GenotypesContext with new PLs
+     */
+    private static GenotypesContext subsetGeneralPloidyAlleles(final VariantContext vc, final int defaultPloidy,
+                                                               final List<Allele> allelesToKeep, final GenotypeAssignmentMethod method) {
+        Utils.nonNull(vc, "vc is null");
+        Utils.nonNull(allelesToKeep, "allelesToKeep is null");
```
vruano: Check whether input allelesToKeep contains at least the reference.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+     * @param vc                 variant context with genotype likelihoods
+     * @param allelesToKeep       which alleles from the vc are okay to use; *** must be in the same relative order as those in the original VC ***
+     * @param assignGenotypes    assignment strategy for the (subsetted) PLs
+     * @return a new non-null GenotypesContext
+     */
+    private static GenotypesContext subsetDiploidAlleles(final VariantContext vc,
+                                                        final List<Allele> allelesToKeep,
+                                                        final GenotypeAssignmentMethod assignGenotypes) {
+        Utils.nonNull(allelesToKeep, "allelesToKeep is null");
+        Utils.validateArg(allelesToKeep.get(0).isReference(), "First allele must be the reference allele");
```
vruano: What if allelesToKeep is empty?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+
+            final GenotypeBuilder gb = new GenotypeBuilder(g);
+            gb.PL(newLikelihoods);
+
+
+            // if we weren't asked to assign a genotype, then just no-call the sample
+            if (method == GenotypeAssignmentMethod.DO_NOT_ASSIGN_GENOTYPES || method == GenotypeAssignmentMethod.SET_TO_NO_CALL) {
+                gb.alleles(GATKVariantContextUtils.noCallAlleles(ploidy));
+            } else {
+                assignGenotypes(gb, newLikelihoods, allelesToKeep, ploidy);
```
vruano: How is that the genotype assignment method is not relevant to assign the genotypes?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+        if (newAlleles.size() == originalAlleles.size()) {
+            return oldLikelihoods;  //no subsetting
+        } else if (newAlleles.size() == 1) {
+            return new double[] { oldLikelihoods[0]};   //hom ref is the only genotype
+        }
+
+        final int newPLSize = GLVectorSizeCache.getNumLikelihoodElements(newAlleles.size(), ploidy);
+        final double[] newPLs = new double[newPLSize];
+
+        // fill boolean array stating whether each original allele is present in new mapping
```
vruano: I think that You may find useful the AlleleListPermutation class for some of the stuff you are doing here.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+    /**
+     * Create the new GenotypesContext with the subsetted PLs and ADs
+     *
+     * @param originalGs               the original GenotypesContext
+     * @param vc                       the original VariantContext
+     * @param allelesToUse             the actual alleles to use with the new Genotypes
+     * @param assignGenotypes          assignment strategy for the (subsetted) PLs
+     * @return a new non-null GenotypesContext
+     */
+    private static GenotypesContext createDiploidGenotypesWithSubsettedLikelihoods(final GenotypesContext originalGs,
```
vruano: Can we merge diploid and non-diploid code? I assume that there is a non-diploid version somewhere.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,421 @@
...
+     * @return a new non-null GenotypesContext
+     */
+    public static GenotypesContext updatePLsAndAD(final VariantContext selectedVC, final VariantContext originalVC) {
+        final int numNewAlleles = selectedVC.getNAlleles();
+        final int numOriginalAlleles = originalVC.getNAlleles();
+        Utils.validateArg( numNewAlleles <= numOriginalAlleles, "Attempting to fix PLs and AD from what appears to be a *combined* VCF and not a selected one");
+
+        final GenotypesContext oldGs = selectedVC.getGenotypes();
+        return numNewAlleles == numOriginalAlleles ? oldGs :
+                createDiploidGenotypesWithSubsettedLikelihoods(oldGs, originalVC, selectedVC.getAlleles(), GenotypeAssignmentMethod.DO_NOT_ASSIGN_GENOTYPES);
```
vruano: If the code make the diploid assumption the name of the method should say so.

vruano:
Done with the review for now. 

There is a few classes that you can do without since the functionality is implemented in GenotypeLikelihoodsCalculator or GenotypeAlleleCounts.  I think you should consider to use AlleleListPermutation also but that may be less useful, you decide. 

Also I would like you to consolidate diploid and non-diploid code before moving forward and review the code again.

vruano:
Also I need to point out that AlleleSubsettingUtils public method set is quite dizzying... they are a rather uncoheren  collection in terms of the type of parameters they accept and return (VC, GC, GenotypeBuilder, naked arrays....) I wish we would stick to something consistent even if that means changing the using code in places. Sorry for not being more concrete, I hope you get the idea.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotpyeIterator.java
davidbenjamin: @vruano I agree with you that this functionality is redundant with that of `GenotypeAlleleCounts` (as invoked by `GenotypeLikelihoodCalculator`.  This `GenotypeIterator` is simply a port of the former nested `SumIterator` class inside `GeneralPloidyExactAFCalculator`.  I would be quite happy to take the next step and eliminate one of the implementations.

However, it's not clear to me which one to get rid of.  The code in `GenotypeIterator` is more concise and its representation -- `{count of allele 0, count of allele 1, count of allele 2 . . .}` -- is more direct than the representation `{index of first allele to appear in this genotype, count of this allele, index of second allele, . . .}`  Then again, in the case of a large number of alleles this sparse representation would improve performance.  And `GenotypeAlleleCounts` has some additional functionality that may be important.

Since you are much more familiar with this code, I'm going to trust your opinion.  Which one should we keep?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GLVectorSizeCache.java
davidbenjamin: Another consequence of the independent life of the general ploidy code.  I look forward to exercising the delete key.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: I think you're right.  This is a relatively cheap operation (compared to read likelihoods) done once per variant.  Furthermore, it's not actually clear to me if this is an optimization at all i.e. it doesn't seem like the general ploidy code would be wasteful when applied to diploids.  So yes, I think it makes sense to scrap the diploid-specific code.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: I can't think of a problem if you can't.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: It is relevant.  This code preserves existing behavior, but I can try to fix.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: This class brings in all the non-diploid code I could find.  My impression is that the general ploidy code has been in its own corner of the GATK and doesn't necessarily have identical functionality.  I'll try to unify things more.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
vruano: I can't

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: fixed

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotpyeIterator.java
davidbenjamin: Fixed class name typo, although probably will delete the class soon.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: added a check for this error

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GLVectorSizeCache.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotpyeIterator.java
davidbenjamin: Done, via `GenotypeLikelihoodCalculator`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotpyeIterator.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: I'm not sure I understand what you want here. The second option you describe sounds like `case USE_PLS_TO_ASSIGN`.  And the first option sounds risky.  Anyway, is there a use case for `BEST_MATCH_TO_ORIGINAL`?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: fixed to no longer assume diploid

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: fixed

davidbenjamin: And existing behavior is still preserved because the different values of `GenotypeAssignmentMethod` were never actually used in code.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

davidbenjamin:
@vruano In addition to the above comments, I have seriously consolidated code as you requested.  This is ready for a second round of review.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,233 @@
...
+            final GenotypeBuilder gb = new GenotypeBuilder(g);
+            gb.PL(newLikelihoods);
+            GATKVariantContextUtils.makeGenotypeCall(ploidy, gb, method, newLikelihoods, allelesToKeep);
+            newGTs.add(gb.make());
+        }
+        return newGTs;
+    }
+
+
+    //TODO: is this doing the same thing as the method immediately above????  Quite likely, yes!!!
```
vruano: Can you address this TODO?

The body of this method above could simply be:

```
    return subsetAlleles(vc.getGenotypes(), vc.getAlleles(), allelesToUse, assignGenotypes);
```

It seems that one difference is that the method above does not take care of the AD annotation... but that seems rather a bug anyway. Perhaps there is more differences but its seems to me that what we should do here is to do the Union of all those changes across both method.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,233 @@
...
+                newLikelihoods = originalLikelihoods.length == expectedNumLikelihoods ?
+                        MathUtils.normalizeFromLog10(Arrays.stream(subsettedLikelihoodIndices)
+                                .mapToDouble(idx -> originalLikelihoods[idx]).toArray(), false, true) : null;
+            }
+
+            final boolean useNewLikelihoods = newLikelihoods != null && GATKVariantContextUtils.isInformative(newLikelihoods);
+            final GenotypeBuilder gb = useNewLikelihoods ? new GenotypeBuilder(g).PL(newLikelihoods) : new GenotypeBuilder(g).noPL();
+
+            // make new genotype call
+            GATKVariantContextUtils.refreshGenotypeCall(g.getAlleles(), gb, assignGenotypes, newLikelihoods, allelesToUse);
```
vruano: What is the difference between ```refreshGenotypeCall``` and ```makeGenotypeCall``` used in the method below (or above, whatever).

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,233 @@
...
+ * genotypes after subsetting.
+ *
+ * @author David Benjamin &lt;davidben@broadinstitute.org&gt;
+ */
+public final class AlleleSubsettingUtils {
+    private AlleleSubsettingUtils() {}  // prevent instantiation
+    private static final int PL_INDEX_OF_HOM_REF = 0;
+    private static final int MAX_LENGTH_FOR_PL_LOGGING = 100; // if PL vectors longer than this # of elements, don't log them
+
+    private static final GenotypeLikelihoodCalculators GL_CALCS = new GenotypeLikelihoodCalculators();
```
vruano: Mmmm... perhaps I meant for GenptypeLikelihoodCalculators to be non-instantiable? Please fix that and use ```GenotypeLikelihoodCalculators.``` when accessing its static methods like ```getInstance```.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,233 @@
...
+     * @param originalAlleles       List of original alleles
+     * @param newAlleles            New alleles -- must be a subset of {@code originalAlleles}
+     * @return                      old PL indices of new genotypes
+     */
+    private static int[] subsettedGenotypeLikelihoodIndices(final int ploidy, final List<Allele> originalAlleles, final List<Allele> newAlleles) {
+        final int[] result = new int[GenotypeLikelihoods.numLikelihoods(newAlleles.size(), ploidy)];
+        final List<Boolean> isAlleleKept = keptAlleleIndices(originalAlleles, newAlleles);
+        final Permutation<Allele> allelePermutation = new IndexedAlleleList<>(originalAlleles).permutation(new IndexedAlleleList<>(newAlleles));
+
+        final GenotypeLikelihoodCalculator glCalc = GL_CALCS.getInstance(ploidy, originalAlleles.size());
```
vruano: No GL_CALCS but GenotypeLikelihoodCalculators as said above.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,233 @@
...
+        for (int oldPLIndex = 0; oldPLIndex < glCalc.genotypeCount(); oldPLIndex++) {
+            final GenotypeAlleleCounts oldAlleleCounts = glCalc.genotypeAlleleCountsAt(oldPLIndex);
+
+            final boolean onlyNewAlleles = IntStream.range(0, oldAlleleCounts.distinctAlleleCount())
+                    .map(oldAlleleCounts::alleleIndexAt).allMatch(isAlleleKept::get);
+
+            if (onlyNewAlleles) {
+                // make an array in the format described in {@link GenotypeAlleleCounts}:
+                // [(new) index of first allele, count of first allele, (new) index of second allele, count of second allele. . .]
+                final List<Integer> newAlleleCounts = new ArrayList<>(newAlleles.size() * 2);
```
vruano: You can use a ```IntStream``` for this. Notice the existence of ```IntStream.flatMap``` to create a flatten stream of int pairs. E.g.:

```
IntStream.range(0, 10).flatMap(i -> IntStream.of(2 * i, 2 * i + 1)); // equivalent to IntStream.range(0, 21)
```

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,233 @@
...
+    private static int[] subsettedGenotypeLikelihoodIndices(final int ploidy, final List<Allele> originalAlleles, final List<Allele> newAlleles) {
+        final int[] result = new int[GenotypeLikelihoods.numLikelihoods(newAlleles.size(), ploidy)];
+        final List<Boolean> isAlleleKept = keptAlleleIndices(originalAlleles, newAlleles);
+        final Permutation<Allele> allelePermutation = new IndexedAlleleList<>(originalAlleles).permutation(new IndexedAlleleList<>(newAlleles));
+
+        final GenotypeLikelihoodCalculator glCalc = GL_CALCS.getInstance(ploidy, originalAlleles.size());
+        for (int oldPLIndex = 0; oldPLIndex < glCalc.genotypeCount(); oldPLIndex++) {
+            final GenotypeAlleleCounts oldAlleleCounts = glCalc.genotypeAlleleCountsAt(oldPLIndex);
+
+            final boolean onlyNewAlleles = IntStream.range(0, oldAlleleCounts.distinctAlleleCount())
```
vruano: change variable name to ```containsOnlyNewAlleles```.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,233 @@
...
+     * @param newAlleles            New alleles -- must be a subset of {@code originalAlleles}
+     * @return                      old PL indices of new genotypes
+     */
+    private static int[] subsettedGenotypeLikelihoodIndices(final int ploidy, final List<Allele> originalAlleles, final List<Allele> newAlleles) {
+        final int[] result = new int[GenotypeLikelihoods.numLikelihoods(newAlleles.size(), ploidy)];
+        final List<Boolean> isAlleleKept = keptAlleleIndices(originalAlleles, newAlleles);
+        final Permutation<Allele> allelePermutation = new IndexedAlleleList<>(originalAlleles).permutation(new IndexedAlleleList<>(newAlleles));
+
+        final GenotypeLikelihoodCalculator glCalc = GL_CALCS.getInstance(ploidy, originalAlleles.size());
+        for (int oldPLIndex = 0; oldPLIndex < glCalc.genotypeCount(); oldPLIndex++) {
```
vruano: I think that even this ```for``` loop could be implemented using an stream.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,233 @@
...
+            final GenotypeAlleleCounts oldAlleleCounts = glCalc.genotypeAlleleCountsAt(oldPLIndex);
+
+            final boolean onlyNewAlleles = IntStream.range(0, oldAlleleCounts.distinctAlleleCount())
+                    .map(oldAlleleCounts::alleleIndexAt).allMatch(isAlleleKept::get);
```
vruano: I got the impression that some of the operations done here could be moved into the ```Permutation``` heriarchy. For example what about adding a method in ```Permutation``` to check whether an original index is kept?:

```java
public interface Permutation<A> {
  // ...
  boolean isKept(final int fromIndex);
  boolean isKept(final A from);
 
}
```

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -0,0 +1,233 @@
...
+
+    /**
+     * Given an original VariantContext and a list of alleles from that VC to keep,
+     * returns a bitset representing which allele indexes should be kept
+     *
+     * @param originalAlleles      the original alleles
+     * @param allelesToKeep   the list of alleles to keep
+     * @return                a non-null List
+     */
+    private static List<Boolean> keptAlleleIndices(final List<Allele> originalAlleles, final List<Allele> allelesToKeep) {
```
vruano: Perhaps this method can be removed if the adequate functionality is added to ```Permutation```

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
```diff
@@ -603,46 +571,44 @@ protected final boolean passesCallThreshold(final double conf) {
...
+        Utils.nonNull(log10GenotypeLikelihoods, "the input likelihoods cannot be null");
+        Utils.validateArg(log10GenotypeLikelihoods.length == this.configuration.genotypeArgs.samplePloidy + 1, "wrong likelihoods dimensions");
+
+        final double[] log10Priors = log10AlleleFrequencyPriorsSNPs.forTotalPloidy(this.configuration.genotypeArgs.samplePloidy);
+        final double log10ACeq0Likelihood = log10GenotypeLikelihoods[0];
+        final double log10ACeq0Prior = log10Priors[0];
+        final double log10ACeq0Posterior = log10ACeq0Likelihood + log10ACeq0Prior;
+
+        // If the maximum a posteriori AC is 0 then the profile value must be 0.0 as per existing code; it does
+        // not matter whether AC > 0 is at all plausible.
```
vruano: mmm... It seems that you are changing the method (maths) so slightly here... right?
Even if it is for the better I rather you do it this in a separate commit which the appropriate detailed explanaton in the commit message.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/DiploidExactAFCalculator.java
```diff
@@ -1,60 +1,4 @@
...
-import htsjdk.variant.variantcontext.GenotypeLikelihoods;
-import htsjdk.variant.variantcontext.GenotypesContext;
-import htsjdk.variant.variantcontext.VariantContext;
-import org.broadinstitute.hellbender.utils.MathUtils;
-import org.broadinstitute.hellbender.utils.Utils;
-import org.broadinstitute.hellbender.utils.variant.GATKVariantContextUtils;
-
-import java.util.*;
-
 public abstract class DiploidExactAFCalculator extends ExactAFCalculator {
```
vruano: This is now just a empty class... do we need to keep it around?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyExactAFCalculator.java
```diff
@@ -14,17 +14,8 @@
...
     private static final boolean VERBOSE = false;
-    private final GenotypeLikelihoodCalculators calculators = new GenotypeLikelihoodCalculators();
-
-    @Override
-    protected GenotypesContext reduceScopeGenotypes(final VariantContext vc, final int defaultPloidy, final List<Allele> allelesToUse) {
-        Utils.nonNull(vc, "vc is null");
-        Utils.nonNull(allelesToUse, "allelesToUse is null");
-        return subsetAlleles(vc,defaultPloidy,allelesToUse,false);
-    }
+    private static final GenotypeLikelihoodCalculators GL_CALCS = new GenotypeLikelihoodCalculators();
```
vruano: Same as said above.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyExactAFCalculator.java
```diff
@@ -355,396 +310,12 @@ private double computeLofK(final ExactACset set,
...
-     * @param nAlleles                    Number of alleles
-     * @param numChromosomes              Ploidy
-     * @param PLindex                     Index to query
-     * @return                            Allele count conformation, according to iteration order from SumIterator
-     */
-    private int[] getAlleleCountFromPLIndex(final int nAlleles, final int numChromosomes, final int PLindex) {
-        final GenotypeLikelihoodCalculator calculator = calculators.getInstance(numChromosomes, nAlleles);
-        final GenotypeAlleleCounts alleleCounts = calculator.genotypeAlleleCountsAt(PLindex);
-        return alleleCounts.alleleCountsByIndex(nAlleles - 1);
+    //TODO: this probably belongs somewhere else
```
vruano: May be GenotypeLikelihoodCalculator(s) or GenotypeAlleleCounts?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
```diff
@@ -1075,34 +1077,22 @@ private VariantContext subsetRecord(final VariantContext vc, final boolean prese
...
-
-        // if there are fewer alternate alleles now in the selected VC, we need to fix the PL and AD values
-        GenotypesContext newGC = GATKVariantContextUtils.updatePLsAndAD(sub, vc);
-
-        // since the VC has been subset (either by sample or allele), we need to strip out the MLE tags
-        builder.rmAttribute(GATKVCFConstants.MLE_ALLELE_COUNT_KEY);
-        builder.rmAttribute(GATKVCFConstants.MLE_ALLELE_FREQUENCY_KEY);
+        // fix the PL and AD values if sub has fewer alleles than original vc and remove a fraction of the genotypes if needed
+        final GenotypesContext oldGs = sub.getGenotypes();
+        GenotypesContext newGC = sub.getNAlleles() == vc.getNAlleles() ? oldGs :
```
vruano: Make newGC final and the put this initialization in a ```else``` statement following the ```if``` below.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -328,6 +222,78 @@ public static boolean isTransition(final VariantContext context) {
...
+    /**
+     * Add the genotype call (GT) field to GenotypeBuilder using the requested {@link GenotypeAssignmentMethod}
+     *
+     * @param originalGT the original genotype calls, cannot be null
+     * @param gb the builder where we should put our newly called alleles, cannot be null
+     * @param assignmentMethod the method to use to do the assignment, cannot be null
+     * @param genotypeLikelihoods a vector of likelihoods to use if the method requires PLs, should be log10 likelihoods, cannot be null
+     * @param allelesToUse the alleles with respect to which the likelihoods are defined
+     */
+    public static void refreshGenotypeCall(final List<Allele> originalGT,
```
vruano: I don't like the way BEST_MATCH_TO_ORIGINAL is treated differently... we do really need it? Under what circumstance wouldn't we just take the one based on the best PL after subsetting?

vruano:
Much better now. Please address comments and lets see how it looks.
Back to @davidbenjamin

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done; made it a purely static class

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: Whoops, I did that earlier but forgot to push the commit.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: Make genotype call doesn't require an existing call and therefore doesn't work with the `BEST_MATCH` assignment method.  Otherwise they are the same.  I see in your comment below that you think `BEST_MATCH` is dubious and I feel the same way.  If I delete this one, which is never used in `HaplotypeCaller` as far as I can tell, I will be able to merge `refreshGenotypeCall` and `makeGenotypeCall`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: I like this idea a lot.  Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: Nice!  done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
davidbenjamin: This diff makes me do a double-take every time, but I actually didn't touch the method.  The reason it shows up is that it was originall:
```java
if (log10GenotypeLikelihoods == null) {
  throw new IllegalArgumentException("the input likelihoods cannot be null");
} else if (log10GenotypeLikelihoods.length != this.configuration.genotypeArgs.samplePloidy + 1) {
  throw new IllegalArgumentException("wrong likelihoods dimensions");
 } else {
   method goes here. . .
```

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/DiploidExactAFCalculator.java
davidbenjamin: Nope, done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/GeneralPloidyExactAFCalculator.java
davidbenjamin: After all the refactoring since I wrote this `TODO`, it seems like it actually doesn't really belong anywhere else.  It will be deleted when `GeneralPloidyExactAFCalculator` is discarded in the near future.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
davidbenjamin: IMO, under no circumstances.  `BEST_MATCH_TO_ORIGINAL` seems really silly to me and doesn't get used in production code.  I'll delete it, which will let `refreshGenotypeCall` and `makeGenotypeCall` become a single method.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
davidbenjamin: Oh wait, the `if` block requires `newGC` to have been previously calculated.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: done

davidbenjamin:
@vruano I had to roll back making the constructor of `GenotypeLikelihoodCalculators` private and all of its methods static because tests were failing.  Interestingly, individual test classes all passed but running the entire test suite as a whole failed with scary heap errors.  I believe that making this class effectively a singleton object shared by the entire GATK causes its memory requirements to grow out of control somehow.  There has to be a smart way around this, but IMO it's beyond the scope of this ticket.

davidbenjamin:
@droazen FYI I made a snapshot and pointed gatk-protected to it.  All HaplotypeCaller-related unit tests and integration tests pass.

davidbenjamin:
@vruano Back to you.  This refactoring is starting to look really nice thanks to your review.

davidbernick:

src/main/java/org/broadinstitute/hellbender/utils/genotyper/AlleleList.java
```diff
@@ -229,21 +232,26 @@ public A getAllele(final int index) {
...
+    final class ActualPermutation<A extends Allele> implements AlleleListPermutation<A> {
 
         private final AlleleList<A> from;
 
         private final AlleleList<A> to;
 
         private final int[] fromIndex;
 
+        private final boolean[] keptFromIndices;
+
```
vruano: Extra blank lines.

vruano:
Like the changes. You can go ahead: Rebase, Squash and Merge.
Back to @davidbenjamin

src/main/java/org/broadinstitute/hellbender/utils/genotyper/AlleleList.java
davidbenjamin: done

PR #1921, opened 2016-06-17
fix the script to validate-reads-spark-pipeline 

akiezun:
I checked that it runs now with picard 2.0.1 and GATK3.5 (validation fails though)


akiezun:
for @lbergelson - the validation fails now however. See https://github.com/broadinstitute/gatk/issues/1922

lbergelson:
@akiezun Changes look fine to me.

PR #1924, opened 2016-06-19
Porting same fix from GATK3 for issue 1357

SHuang-Broad:
Most changes are trivial, except `computeMaxAcceptableAlleleCount `, which is exactly the same code as in the GATK3 fix.

The part that's missing is changes made to `HaplotypeCallerGenotyingEngine`, which is in GATK-protected, which in turn depends on these classes in GATK. So these changes have to go in first.


SHuang-Broad:
@davidbenjamin would you please review here as well? It should be trivial, as they are almost the same charge as the changes made to the GATK3 code.
Thanks!

davidbenjamin:
@SHuang-Broad Sure, I can do this one too.

PR #1927, opened 2016-06-20
Spark version of BWA+MarkDuplicates

akiezun:
this PR puts together BWA and MarkDuplicates. This is a prelude to BWA+MD+BQSR (but simpler because BQSR requires a 2bit reference and bwa wants a fasta reference).

It extracts the core BWA/Spark code into a BwaSparkEngine and call that from both BwaSpark and the new pipeline. 

It also improves the handling of sorting order for spark writing - adds a way to sort by queryname (relevant for mark duplicates).

@tomwhite can you review?


src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSink.java
```diff
@@ -249,19 +248,36 @@ private static void writeReadsSingle(
...
         return readVoidPairs.map(Tuple2::_1);
     }
 
+    //Returns the comparator to use or null if no sorting is required.
+    private static Comparator<SAMRecord> getSAMRecordComparator(final SAMFileHeader header) {
+        switch (header.getSortOrder()){
+            case coordinate: return new HeaderlessSAMRecordCoordinateComparator(header);
+            case duplicate:  return header.getSortOrder().getComparatorInstance();
+            case queryname:  return header.getSortOrder().getComparatorInstance();
+            case unsorted:   return header.getSortOrder().getComparatorInstance();
```
tomwhite: Make the three cases fall through to a single return statement.

tomwhite:
LGTM

PR #1932, opened 2016-06-22
no coveralls for cloud only tests

akiezun:
trying to fix the coveralls insanity about coverage dropping to 12% all the time. I think it may be because we send results from the cloud tests which are tiny

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,11 +87,7 @@ script:
     travis_wait 30 ./gradlew jacocoTestReport;
   fi
 after_success:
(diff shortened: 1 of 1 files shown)


akiezun:
@lbergelson can you review this? I got very annoyed by the emails with coverage oscillating between 12% and 81%. This is my attempt to fix it. It seems to work but i'll appreciate your review

lbergelson:
@akiezun We'll miss out on that .5% coverage the the cloud tests hit, but should be fine otherwise.  👍  if the pingponging is annoying you.  I have an open ticket with coveralls to address it, but no one has ever responded... so I doubt it will ever be resolved.

PR #1933, opened 2016-06-22
Moved tool to gatk-protected

davidbenjamin:
* moved `CalculateTargetCoverage` to gatk-protected in order to make it aware of `Target` and the associated table reader.  That way it can input the standard target table file used in the rest of gatk-protected.
* Deleted `TargetCollectionUtils` as all its methods are obviated by a parallel PR in gatk-protected.


davidbenjamin:
@LeeTL1220 This and the other PR are needed to get the QScripts with GC working.

LeeTL1220:
@davidbenjamin We need to file an issue to move these back into GATK public.

LeeTL1220:
@davidbenjamin This pull request looks okay, but you will need to coordinate a release with @droazen or @lbergelson , since this is not a typical case.

davidbenjamin:
@LeeTL1220 I made a snapshot and switched build.gradle in gatk-protected accordingly in the companion PR.  I have @lbergelson's permission to move this CLI to proitected, with the understanding that we should later try to move it back to gatk-public along with all the supporting `Target` etc classes.

PR #1939, opened 2016-06-24
AlignContigsAndCallBreakpointsSpark tool

cwhelan:
This tool takes the contigs from an assembly, aligns them to the reference, and calls possible breakpoints from split alignments.

@SHuang-Broad @tedsharpe would both/either of you mind taking a look?


SHuang-Broad:
@tedsharpe , please feel free to review as well. This to me is more about learning than reviewing, though I'll try hard.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+            throw new GATKException("Unable to load reference", e);
+        }
+        bwaMem = new BwaMem(index);
+
+    }
+
+    public List<AssembledBreakpoint> alignContigs (final RunSGAViaProcessBuilderOnSpark.ContigsCollection contigsCollection) {
+        final List<AssembledBreakpoint> assembledBreakpoints = new ArrayList<>();
+
+
```
SHuang-Broad: Extra empty line

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+                                assembledBreakpoint.insertedSequence = "NA";
+                                if (assembledBreakpoint.region1.endInContig < assembledBreakpoint.region2.startInContig - 1) {
+                                    assembledBreakpoint.insertedSequence = getSequence(contigsCollection, contigId, assembledBreakpoint.region1.endInContig, assembledBreakpoint.region2.startInContig);
+                                }
+
+                                assembledBreakpoints.add(assembledBreakpoint);
+                            }
+                        }
+                    }
+                }
```
SHuang-Broad: Indentation seems to be off for this `try` block

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+                        final List<AlignmentRegion> alignmentRegionList = Arrays.stream(alnRgns)
+                                .filter(a -> a.getSecondary() == -1)
+                                .map(AlignmentRegion::new)
+                                .sorted(Comparator.comparing(a -> a.startInContig))
+                                .collect(Collectors.toList());
+
+                        AlignmentRegion current = alignmentRegionList.get(0);
+                        AlignmentRegion previous;
+                        for (int i = 0; i < alignmentRegionList.size(); i++) {
+                            if (i > 0) {
```
SHuang-Broad: is this test for `i` necessary? The loop doesn't seem to be modifying it.

SHuang-Broad: I see now. Should we start from `i=1` then?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+                                if (assembledBreakpoint.region1.endInContig >= assembledBreakpoint.region2.startInContig) {
+                                    assembledBreakpoint.homology = getSequence(contigsCollection, contigId, assembledBreakpoint.region2.startInContig - 1, assembledBreakpoint.region1.endInContig);
+                                }
+
+                                assembledBreakpoint.insertedSequence = "NA";
+                                if (assembledBreakpoint.region1.endInContig < assembledBreakpoint.region2.startInContig - 1) {
+                                    assembledBreakpoint.insertedSequence = getSequence(contigsCollection, contigId, assembledBreakpoint.region1.endInContig, assembledBreakpoint.region2.startInContig);
+                                }
+
+                                assembledBreakpoints.add(assembledBreakpoint);
```
SHuang-Broad: Would it be better if `AssembledBreakpoint` provide a constructor like this:

    AssembledBreakpoint(final String contigId, final AlignmentRegion region1, final AlignmentRegion region2, String insertedSequence, String homology){
            this.contigId = contigId;
            this.region1  = region1;
            this.region2  = region2;
            this.insertedSequence = insertedSequence==null? "NA" : insertedSequence;
            this.homology = homology==null ? "NA" : homology;
        }

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+                    posInContig += cigar.getCigarElement(i).getLength();
+                    i = i + 1;
+                }
+            } else {
+                int i = cigar.getCigarElements().size() - 1;
+                while (cigar.getCigarElement(i).getOperator() == CigarOperator.H || cigar.getCigarElement(i).getOperator() == CigarOperator.S) {
+                    posInContig += cigar.getCigarElement(i).getLength();
+                    i = i - 1;
+                }
+            }
```
SHuang-Broad: What about replacing with 

    int j = forwardStrand ? 0 : cigar.getCigarElements().size() - 1;
    final int offset = forwardStrand ? 1 : -1;
    CigarElement ce = cigar.getCigarElement(j);
    while(ce.getOperator().isClipping()){
        posInContig += ce.getLength();
        j += offset;
        ce = cigar.getCigarElement(j);
    }

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+     *
+     * Inserted sequence contains portions of the contig that are aligned to neither region, and therefore may be inserted in
+     * the sample. For example, a translocation breakpoint with a microinsertion:
+     *
+     * Contig:
+     * ACTGACTGACTGACTGACTGACTGACTGACTGACTGACTGACTG
+     * Alignment regions:
+     * |-----1:100-200-------|
+     *                         |-----2:100-200-----|
+     * Inserted sequence:
```
SHuang-Broad: Should the second alignment region's left vertical bar match the 'C' instead of 'A'?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+
+    /**
+     * Generate a bogus base quality sequence to pass in for the aligned contig (since the jBWA API requires that reads have qualities)
+     * @param length
+     * @return
+     */
+    private byte[] qualSequence(final int length) {
+        final byte[] quals = new byte[length];
+        for (int i = 0; i < quals.length; i++) {
+            quals[i] = 'A';
```
SHuang-Broad: I think this `for` loop could be replaced with `Arrays.fill(quals, (byte)'A');`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+        }
+        return quals;
+    }
+
+    public void close() {
+        bwaMem.dispose();
```
SHuang-Broad: It seems that the constructor constructed the `BwaIndex` object, so I believe `index.close();` is necessary here.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -511,6 +511,22 @@ public String toString(){
             public String toString(){
                 return id;
             }
+
+            @Override
+            public boolean equals(final Object o) {
+                if (this == o) return true;
+                if (o == null || getClass() != o.getClass()) return false;
```
tedsharpe: This is fine with me, but you could consider:
`if ( !(o instanceof ContigID) ) return false;`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -511,6 +511,22 @@ public String toString(){
...
+
+                final ContigID contigID = (ContigID) o;
+
+                return id != null ? id.equals(contigID.id) : contigID.id == null;
```
tedsharpe: No big deal, but there's a JDK method for this if you'd prefer it:  `return Objects.equals(id, contigID.id);`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -511,6 +511,22 @@ public String toString(){
...
+
+            }
+
+            @Override
+            public int hashCode() {
+                return id != null ? id.hashCode() : 0;
```
tedsharpe: Built into JDK as `return Objects.hashCode(id);`.  Up to you.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -524,6 +540,15 @@ public String toString(){
             return StringUtils.join(toListOfStrings(),"\n");
         }
 
+        /**
+         * Pack the entire fasta record on one line so that it's easy for downstream
+         * Spark tools to parse without worrying about partitioning.
+         * @return
```
tedsharpe: return is undocumented.  either add a description for the tag, or remove the `@return`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -524,6 +540,15 @@ public String toString(){
...
+         */
+        public String toOneLineString(){
+            return StringUtils.join(toListOfStrings(),"|");
```
tedsharpe: Should probably document that this can return null.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+        return assembledBreakpoints;
+    }
+
+    /**
+     * Pulls the sequence of the requested contig between the start and end coordinates from the contigsCollection
+     * @param contigsCollection
+     * @param contigId
+     * @param startInContig
+     * @param endInContig
+     * @return
```
tedsharpe: I think our convention is to either document fully, or omit tags altogether.  The `@param` and `@return` tags are just noise, I think.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+            stringBuilder.append(contigId);
+            stringBuilder.append("\t");
+            stringBuilder.append(region1.toString());
+            stringBuilder.append("\t");
+            stringBuilder.append(region2.toString());
+            stringBuilder.append("\t");
+            stringBuilder.append(insertedSequence);
+            stringBuilder.append("\t");
+            stringBuilder.append(homology);
+            return stringBuilder.toString();
```
tedsharpe: IDEA seems to think that this is simpler, and just as efficient if not moreso:
`return contigId + '\t' + region1 + '\t' + region2 + '\t' + insertedSequence + '\t' + homology;`
I don't know whether it's right.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -0,0 +1,121 @@
...
+
+    }
+
+    /**
+     * To ease file format difficulties
+     * @param v
+     * @return
+     */
+    @VisibleForTesting
+    protected static RunSGAViaProcessBuilderOnSpark.ContigsCollection parsePackedFasta(final String v) {
```
SHuang-Broad: the parameter name `v` isn't a very good choice...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ContigAligner implements Closeable {
+
+    final BwaIndex index;
+    final BwaMem bwaMem;
+
+    public ContigAligner(final File referenceFile) {
+        BWANativeLibrary.load();
```
tedsharpe: I think this is probably OK -- the NativeUtils class used by BWANativeLibrary checks to make sure the library isn't already loaded -- but I think it would be clearer to put this in a static initialization block to signal that it's something that we only ever need to do once.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -0,0 +1,121 @@
...
+                    final List<Tuple2<String, ContigAligner.AssembledBreakpoint>> results = new ArrayList<>();
+                    iter.forEachRemaining(cc -> {
+                        final List<ContigAligner.AssembledBreakpoint> breakpointAssemblies = contigAligner.alignContigs(cc._2);
+                        breakpointAssemblies.forEach(b -> results.add(new Tuple2<>(cc._1, b)));
+                    });
+                    return results.iterator();
+                }
+            } catch (IOException e) {
+                throw new GATKException("Cannot run BWA-MEM", e);
+            }
```
SHuang-Broad: Is `ContigAligner.close()` called here? 
If not, I think it should, because C has allocated memory if `ContigAligner contigAligner = new ContigAligner(localRef)` is successful.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -0,0 +1,121 @@
...
+
+        });
+
+        assembledBreakpoints.saveAsTextFile(output);
+
+    }
```
SHuang-Broad: extra blank line

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -0,0 +1,121 @@
...
+                        .coalesce((int) Math.ceil((double) numInputPartitions / (double) NUM_ASSEMBLIES_PER_PARTITION));
+
+        final JavaPairRDD<String, RunSGAViaProcessBuilderOnSpark.ContigsCollection> breakpointIdsToContigsCollection =
+                contigCollectionByBreakpointId.mapValues(AlignContigsAndCallBreakpointsSpark::parsePackedFasta);
+
+        final JavaPairRDD<String, ContigAligner.AssembledBreakpoint> assembledBreakpoints = breakpointIdsToContigsCollection.mapPartitionsToPair(iter -> () -> {
+            try {
+                final String ref = referenceArguments.getReferenceFileName();
+                final File localRef = BucketUtils.isHadoopUrl(ref) ? BwaSparkEngine.localizeReferenceAndBwaIndexFiles(ref) : new File(ref);
+                try (ContigAligner contigAligner = new ContigAligner(localRef)) {
```
SHuang-Broad: `final ContigAligner`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+        try {
+            index = new BwaIndex(referenceFile);
+        } catch (IOException e) {
```
tedsharpe: Local superstition seems to dictate that we noise up the code with `final` everywhere that it can legally go.  If you want to change this to `final IOException e` you should also fix line 82 of AlignContigsAndCallBreakpointsSpark.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Number of assemblies to process per task. We need to make sure that the work is not
+     * over-partitioned since each worker needs to localize the BWA reference index and read it into
+     * memory once per partition.
+     */
+    public static final int NUM_ASSEMBLIES_PER_PARTITION = 200;
+
```
SHuang-Broad: Not a comment against the code or design here, but I was considering edge cases when this could fail. One such scenario is that when SPARK is run in local mode, and there are 1,000 contigs FASTA file to be processed -> 5 partitions. Each one, loading the reference, could blowup the memory usage?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+        bwaMem = new BwaMem(index);
+
+    }
+
+    public List<AssembledBreakpoint> alignContigs (final RunSGAViaProcessBuilderOnSpark.ContigsCollection contigsCollection) {
+        final List<AssembledBreakpoint> assembledBreakpoints = new ArrayList<>();
+
+
+            try {
+                for(final Tuple2<RunSGAViaProcessBuilderOnSpark.ContigsCollection.ContigID, RunSGAViaProcessBuilderOnSpark.ContigsCollection.ContigSequence> contigInfo : contigsCollection.getContents()) {
```
tedsharpe: The type name is so long that it's pretty hard to read.  Maybe consider adding some static imports?  (I don't find it particularly troubling to clutter the local namespace with static imports -- everyone seems to assume an IDE that will show where something is defined these days.)
So if you added this at the top:
`import static org.broadinstitute.hellbender.tools.spark.sv.RunSGAViaProcessBuilderOnSpark.ContigsCollection.*;`
This line could become:
`for(final Tuple2<ContigID, ContigSequence> contigInfo : contigsCollection.getContents()) {`
Don't know whether our standards or your standards prohibit this, but in this case it seems far, far clearer to me.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+                                    assembledBreakpoint.insertedSequence = getSequence(contigsCollection, contigId, assembledBreakpoint.region1.endInContig, assembledBreakpoint.region2.startInContig);
+                                }
+
+                                assembledBreakpoints.add(assembledBreakpoint);
+                            }
+                        }
+                    }
+                }
+        } catch (final IOException e) {
+            throw new GATKException("could not execute BWA");
```
SHuang-Broad: I think `index.close()` and `mem.dispose()` should be called before re-throwing.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+     * @return
+     */
+    private String getSequence(final RunSGAViaProcessBuilderOnSpark.ContigsCollection contigsCollection, final String contigId, final int startInContig, final int endInContig) {
+        for (final Tuple2<RunSGAViaProcessBuilderOnSpark.ContigsCollection.ContigID, RunSGAViaProcessBuilderOnSpark.ContigsCollection.ContigSequence> contig : contigsCollection.getContents()) {
+            if (contig._1.equals(new RunSGAViaProcessBuilderOnSpark.ContigsCollection.ContigID(contigId))) {
+                final String sequence = contig._2.toString();
+                return sequence.substring(startInContig, endInContig);
+            }
+        }
+        throw new GATKException("Could not find contig " + contigId);
```
SHuang-Broad: Again, I think index.close() and mem.dispose() should be called before re-throwing.
I think it would be worth it to have a subclass of Exception here, and have an utility function to catch that exception and close C-allocated resources, then re-throw.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ContigAligner implements Closeable {
+
+    final BwaIndex index;
```
tedsharpe: Appears to be unnecessary to store this as an instance field.  Just keep it local to the constructor?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+            stringBuilder.append(homology);
+            return stringBuilder.toString();
+        }
+    }
+
+    static class AlignmentRegion {
+
+        final Cigar cigar;
+        final boolean forwardStrand;
+        final SimpleInterval interval;
```
tedsharpe: There are conceptually two intervals represented here:  the interval on the assembled contig, and the interval on the reference contig.  Naming could be clearer to keep these straight:  maybe call this `referenceInterval` and change `startInContig` and `endInContig` to `startInAssembledContig` and `endInAssembledContig`?  Or something like that.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+
+
+        public AlignmentRegion(final AlnRgn alnRgn) {
+            this.forwardStrand = alnRgn.getStrand() == '+';
+            this.cigar = TextCigarCodec.decode(alnRgn.getCigar());
+            this.interval = new SimpleInterval(alnRgn.getChrom(), (int) alnRgn.getPos() + 1, (int) (alnRgn.getPos() + 1 + cigar.getReferenceLength()));
+            this.mqual = alnRgn.getMQual();
+            this.contigLength = cigar.getReadLength();
+            this.startInContig = startOfAlignmentInContig();
+            this.endInContig = endOfAlignmentInContig();
```
tedsharpe: My personal feeling is that calling instance methods from the constructor is too error prone.  You're using partial object state to set the rest of the state, and the order becomes crucially important and easy to overlook when maintaining the code.  I'd make the three private methods static, which means you'll have to pass them a forwardStrand flag and a cigar string.  That way they're independent of the state of the object and will always work as intended.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+        final SimpleInterval interval;
+        final int mqual;
+        final int startInContig;
+        final int endInContig;
+        final int contigLength;
+
+
+        public AlignmentRegion(final AlnRgn alnRgn) {
+            this.forwardStrand = alnRgn.getStrand() == '+';
+            this.cigar = TextCigarCodec.decode(alnRgn.getCigar());
```
tedsharpe: I'd consider reversing the cigar right here when it's initialized if we have a reverse strand mapping.  That simplifies getClippedBases, and will probably be more intuitive for users (who won't expect to have to consult the forwardStrand flag for correct cigar interpretation).  I.e., it'll be more BAM-like.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+                    final String contigId = contigInfo._1.toString();
+                    final byte[] sequence = contigInfo._2.toString().getBytes();
+                    final AlnRgn[] alnRgns = bwaAlignSequence(bwaMem, contigId, sequence);
+
+                    // todo: parse cigar for internal indels
+                    if (alnRgns.length > 1) {
+
+                        // filter out secondary alignments, convert to AlignmentRegion objects and sort by alignment start pos
+                        final List<AlignmentRegion> alignmentRegionList = Arrays.stream(alnRgns)
+                                .filter(a -> a.getSecondary() == -1)
```
tedsharpe: I can't find any documentation for AlnRgn, but the provider's sample code in GitHub tests >= 0 to mean "not secondary".  If you found docs somewhere that indicate a 2ndary will always be indicated with the exact value -1, you might want to create a constant and comment it with a pointer to the actual doc.  Otherwise, I think it might be safer to test for `< 0` here.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+                                .map(AlignmentRegion::new)
+                                .sorted(Comparator.comparing(a -> a.startInContig))
+                                .collect(Collectors.toList());
```
tedsharpe: You don't know what kind of list will be produced by `Collectors.toList()` but you're assuming efficient random access (get by index) in the code just below.  If it's a linked list, that code is going to be O(N^2).  If it's an ArrayList, we're creating it with a default (small) size and list construction is going to be O(N^2) as it reallocates to grow the array.  (Neither is important if `alnRgns.length` is small, which it probably will be in most circumstances.)

I realize I'm betraying my C++ background here, but IMO this should really be something like:
`.collect( Collectors.toCollection( () -> new ArrayList<>(alnRgns.length) ) )`
If you filter some elements out of the list, the capacity will be a little generous, but that's OK.  We could even pull the lambda out as a utility function so that we get in the habit of creating lists with an appropriate capacity when we have a reasonable capacity estimate.

And the iteration through the list (four lines below) ought to use an iterator to iterate over the elements so that it will work efficiently with any type of list of any size.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+
+                        AlignmentRegion current = alignmentRegionList.get(0);
```
tedsharpe: You don't know that there is a zero'th element.  It's a pretty safe assumption, because `alnRgns.length > 1` and it would be very weird if all the alignments were secondary so that they all got filtered out.  But I'd at least add a comment something like:
`// Assuming that at least one of the alignments is primary and so there will be at least one element remaining.`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+                        AlignmentRegion previous;
+                        for (int i = 0; i < alignmentRegionList.size(); i++) {
```
tedsharpe: How about:
`Iterator<AlignmentRegion> itr = alignmentRegionList.iterator();
 if ( itr.hasNext() ) {
    AlignmentRegion current = itr.next();
    while ( itr.hasNext() ) {
        final AlignmentRegion previous = current;
        current = itr.next();
        ...
      }
   }`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+
+        private int endOfAlignmentInContig() {
+            return contigLength - getClippedBases(!forwardStrand);
+        }
+
+        private int getClippedBases(final boolean forwardStrand) {
+            int posInContig = 0;
+            if (forwardStrand) {
+                int i = 0;
+                while (cigar.getCigarElement(i).getOperator() == CigarOperator.H || cigar.getCigarElement(i).getOperator() == CigarOperator.S) {
```
tedsharpe: `getOperator().isClipping()` is shorter

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        private int endOfAlignmentInContig() {
+            return contigLength - getClippedBases(!forwardStrand);
+        }
+
+        private int getClippedBases(final boolean forwardStrand) {
+            int posInContig = 0;
+            if (forwardStrand) {
+                int i = 0;
```
tedsharpe: I'd prefer an iterator rather than index arithmetic, but maybe that's just me.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+                                current = alignmentRegionList.get(i);
+                                final AssembledBreakpoint assembledBreakpoint = new AssembledBreakpoint();
+                                assembledBreakpoint.contigId = contigId;
+
+                                assembledBreakpoint.region1 = previous;
+                                assembledBreakpoint.region2 = current;
+
+                                assembledBreakpoint.homology = "NA";
+                                if (assembledBreakpoint.region1.endInContig >= assembledBreakpoint.region2.startInContig) {
+                                    assembledBreakpoint.homology = getSequence(contigsCollection, contigId, assembledBreakpoint.region2.startInContig - 1, assembledBreakpoint.region1.endInContig);
```
tedsharpe: We already have the sequence (`contigInfo._2`).  We can pick off the right stuff with a String substring rather than searching the whole collection for the contig under consideration.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+        final List<AssembledBreakpoint> assembledBreakpoints = new ArrayList<>();
+
+
+            try {
+                for(final Tuple2<RunSGAViaProcessBuilderOnSpark.ContigsCollection.ContigID, RunSGAViaProcessBuilderOnSpark.ContigsCollection.ContigSequence> contigInfo : contigsCollection.getContents()) {
+                    final String contigId = contigInfo._1.toString();
+                    final byte[] sequence = contigInfo._2.toString().getBytes();
+                    final AlnRgn[] alnRgns = bwaAlignSequence(bwaMem, contigId, sequence);
+
+                    // todo: parse cigar for internal indels
```
tedsharpe: There's also the issue about the assembler not necessarily putting together a nice chimeric contig to consider.  (We'd have to explore the graph connectivity info to understand these.)  Is it an issue?  Or do we know that SGA will output chimeras when possible?  File it as an issue to investigate?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,272 @@
...
+        private int getClippedBases(final boolean forwardStrand) {
+            int posInContig = 0;
+            if (forwardStrand) {
+                int i = 0;
+                while (cigar.getCigarElement(i).getOperator() == CigarOperator.H || cigar.getCigarElement(i).getOperator() == CigarOperator.S) {
+                    posInContig += cigar.getCigarElement(i).getLength();
+                    i = i + 1;
+                }
+            } else {
+                int i = cigar.getCigarElements().size() - 1;
```
tedsharpe: You could consider using a ListIterator, rather than index arithmetic.  (It lets you step backward.)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+        return new RunSGAViaProcessBuilderOnSpark.ContigsCollection(fileContents);
+    }
+
+    /**
+     * input format is tab separated BreakpointId, PackedFastaFile
+     * @param assemblyLine
+     * @return
+     */
+    private static Iterable<Tuple2<String, String>> splitAssemblyLine(final String assemblyLine) {
```
tedsharpe: Ditto comment above.  Seems like this belongs in RunSGAProcessBuilderOnSpark.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
tedsharpe: Seems like this functionality -- which requires understanding the file format that RunSGAProcessBuilderOnSpark produced -- ought to be a part of RunSGAProcessBuilderOnSpark.  But maybe that's complicated for some reason.

I guess I'm asking for a method in RunSGAProcessBuilderOnSpark that takes a file name (for an assembly), and returns a `JavaPairRDD<String, ContigCollection>`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
tedsharpe: Do we need to close the BwaIndex?  Or does BwaMem.dispose() do that?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Good catch, done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: close the BwaIndex -- done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: It is closed because it is constructed in a try-with-resources block:

`try (ContigAligner contigAligner = new ContigAligner(localRef)) {`

Which means that close will automatically be called on it when the try block is exited.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Since they're created in the contstructor and closed in the `close()` method, I think it makes sense to let this method throw an exception, and have the client be responsible for closing the resources (since it opened them explicitly). Otherwise if they do the right thing and call close later on, the resources will have already been destroyed. Does that make sense?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
SHuang-Broad: I understand now. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -0,0 +1,121 @@
...
+                        .flatMapToPair(AlignContigsAndCallBreakpointsSpark::splitAssemblyLine)
+                        .coalesce((int) Math.ceil((double) numInputPartitions / (double) NUM_ASSEMBLIES_PER_PARTITION));
+
+        final JavaPairRDD<String, RunSGAViaProcessBuilderOnSpark.ContigsCollection> breakpointIdsToContigsCollection =
+                contigCollectionByBreakpointId.mapValues(AlignContigsAndCallBreakpointsSpark::parsePackedFasta);
+
+        final JavaPairRDD<String, ContigAligner.AssembledBreakpoint> assembledBreakpoints = breakpointIdsToContigsCollection.mapPartitionsToPair(iter -> () -> {
+            try {
+                final String ref = referenceArguments.getReferenceFileName();
+                final File localRef = BucketUtils.isHadoopUrl(ref) ? BwaSparkEngine.localizeReferenceAndBwaIndexFiles(ref) : new File(ref);
```
tedsharpe: Ouch.  Spark doesn't make this easy, but I think there's considerable opportunity for improvement here:  Right now we're localizingReferenceAndBwsIndexFiles once per partition, if I'm reading this correctly.  Once per core would be better, once per executor would be better than that, and once per node would be better yet.  Maybe we can brainstorm?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -0,0 +1,121 @@
...
+    protected void runTool(final JavaSparkContext ctx) {
+
+        final JavaRDD<String> inputAssemblies = ctx.textFile(input).cache();
+
+        final long numInputPartitions = inputAssemblies.count();
+
+        final JavaPairRDD<String, String> contigCollectionByBreakpointId =
+                inputAssemblies
+                        .flatMapToPair(AlignContigsAndCallBreakpointsSpark::splitAssemblyLine)
+                        .coalesce((int) Math.ceil((double) numInputPartitions / (double) NUM_ASSEMBLIES_PER_PARTITION));
```
tedsharpe: Probably doesn't make sense to reduce to fewer than the number of cores, though, right?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
tedsharpe: This means we're making a BwaIndex and a BwaMem once per partition, if I'm understanding correctly.  We can probably do better.  Again, let's brainstorm.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
SHuang-Broad: Quite a nick trick, Ted!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
SHuang-Broad: Would [this](https://github.com/lh3/bwa/blob/master/example.c) help? In particular, line 43.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
SHuang-Broad: Yes, that's correct. Sorry I didn't realize that.

SHuang-Broad: On a separate issue though, if you look at the source code in JBWA, the author does test for "freeing already freed pointers", which is good. So I think it is safe to do so.

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
```diff
@@ -54,4 +57,45 @@ public void testJBWAAlignSingleRead() throws Exception {
         Assert.assertEquals(align[0].getNm(), 0);
         bwaMem.dispose();
```
SHuang-Broad: Again, it seems that `index.close()` is necessary here. Actually, does it make sense to have a static `BwaIndex` as field and have something similar to ["@AfterClass"](http://testng.org/javadocs/org/testng/annotations/AfterClass.html)?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Yes, you're right, much simpler. Done.

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
```diff
@@ -12,6 +12,9 @@
 import org.testng.annotations.Test;
 
 import java.io.File;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
 
 public class JBWAIntegrationTest extends BaseTest {
```
SHuang-Broad: A bit nitpicking here: should the name be `UnitTest` rather than integration test? I though integration test here means something different (at least in GATK3)?

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
```diff
@@ -54,4 +57,45 @@ public void testJBWAAlignSingleRead() throws Exception {
...
+            Assert.assertEquals(alignRgn1.getMQual(), 60);
+            Assert.assertEquals(alignRgn1.getPos(), 999999);
+
+            final AlnRgn alignRgn2 = align[1];
+            Assert.assertEquals(alignRgn2.getChrom(), "20");
+            Assert.assertEquals(alignRgn2.getCigar(), "300S300M");
+            Assert.assertEquals(alignRgn2.getMQual(), 60);
+            Assert.assertEquals(alignRgn2.getPos(), 1000999);
+        } finally {
+            bwaMem.dispose();
```
SHuang-Broad: `index.close()`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: I think it would. Done.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAlignerTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+        final ContigAligner.AlignmentRegion breakpoint2Region2 = breakpoint2.region2;
+        Assert.assertEquals(breakpoint2Region2.interval, new SimpleInterval("21", 27374701, 27375219));
+        Assert.assertTrue(breakpoint2Region2.forwardStrand);
+        Assert.assertEquals(breakpoint2Region2.mqual, 60);
+
+        Assert.assertEquals(breakpoint2.homology, "NA");
+        Assert.assertEquals(breakpoint2.insertedSequence, "NA");
+
+
```
SHuang-Broad: I think `contigAligner.close()` is on call here. Same thing goes for the other test. Or have a `@AfterClass`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Much nicer, thanks. Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: That's nice, thanks. Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
cwhelan: You're right -- I was using an outdated IDEA equals and hashcode template. Fixed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
cwhelan: Updated

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: You're right, sorry. IDEA put them in automatically when I was just trying to add method level descriptor javadoc and I forgot to remove them. Removed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Interesting, I thought StringBuilders were better. But it's simpler to just concatenate strings. Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: Yes it's definitely not ideal and I struggled to find a better way to do this. Let's brainstorm.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -0,0 +1,121 @@
...
+
+        final JavaPairRDD<String, RunSGAViaProcessBuilderOnSpark.ContigsCollection> breakpointIdsToContigsCollection =
+                contigCollectionByBreakpointId.mapValues(AlignContigsAndCallBreakpointsSpark::parsePackedFasta);
+
+        final JavaPairRDD<String, ContigAligner.AssembledBreakpoint> assembledBreakpoints = breakpointIdsToContigsCollection.mapPartitionsToPair(iter -> () -> {
+            try {
+                final String ref = referenceArguments.getReferenceFileName();
+                final File localRef = BucketUtils.isHadoopUrl(ref) ? BwaSparkEngine.localizeReferenceAndBwaIndexFiles(ref) : new File(ref);
+                try (ContigAligner contigAligner = new ContigAligner(localRef)) {
+                    final List<Tuple2<String, ContigAligner.AssembledBreakpoint>> results = new ArrayList<>();
```
tedsharpe: Needs a capacity.  Even a bad guess on how big it'll be is better than no guess.  But take a look at the MapPartition class for a way to make this evaluation lazy, and to avoid the need for a big list entirely.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: changed name to `packedFastaLine` and moved to `RunSGAProcessBuilderOnSpark .ContigsCollection`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Moved line to static initializer

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: Made `contigAligner` final

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: I think we do need to store it as an instance so we can close it later.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Yes, nice suggestion. Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Yep, you're right, that was messy. Refactored to static methods.

tedsharpe:
I'm done reviewing.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
tedsharpe: Yes, that's the place I was looking, too.  If we're taking that as gospel, the reversed test would be `< 0`, not `== -1`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Sounds good, done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Easy enough to make a little method that returns a collector into an ArrayList with a given capacity.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Switched to an iterator approach as you recommended below so this doesn't apply anymore.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Yeah, looks nicer not to have current and previous in scope outside of the loop. Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Already done as per @SHuang-Broad 's suggestion

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Leaving this the way @SHuang-Broad suggested above.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Well that was dumb of me. Thank you for catching that. Done, and removed `getSequence` method.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: Done

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
cwhelan: I didn't write this test but I've added a try/finally here that disposes the mem and closes the index. Some tests don't need it so I didn't move stuff to before / after class methods.

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
cwhelan: I didn't create name this file originally, and I'm not quite sure what the strict definition of integration test is for us. I think I'll leave it for now..

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
cwhelan: done

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAlignerTest.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Yes I guess that was a limitation of my ASCII art. Fixed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: I like the MapPartitioner but I'm not sure how to set it up so that it can close resources. Added a capacity estimate based on NUM_ASSEMBLIES_PER_PARTITION to this results list instead.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: That's certainly possible. I'm still trying to figure out a way to share the BwaIndex between tasks in the same executor, but can't figure out how I would close/free the index if I did that.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
cwhelan: Definitely a possible problem. Filed issue https://github.com/broadinstitute/gatk/issues/1965 to follow up.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: I've created https://github.com/broadinstitute/gatk/issues/1966 to figure out a way to share BwaIndex objects between tasks.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: Created https://github.com/broadinstitute/gatk/issues/1966 to figure out how to share index objects between tasks.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: It seems like we need getExecutorCores() to implement this, right? Added issue https://github.com/broadinstitute/gatk/issues/1967 to fix this when https://github.com/broadinstitute/gatk/pull/1947 is done.

cwhelan: Never mind, I'm now enlightened to `defaultParallelism()`. Done.

cwhelan:
@tedsharpe @SHuang-Broad  I've finished addressing PR comments for now. The one thing I left unchanged was creating one BwaIndex per partition, since I'm not sure how to close the index when all processing is done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -61,61 +58,28 @@ protected void runTool(final JavaSparkContext ctx) {
...
+                contigCollectionByBreakpointId.mapValues(ContigsCollection::fromPackedFasta);
 
-        final JavaPairRDD<String, ContigAligner.AssembledBreakpoint> assembledBreakpoints = breakpointIdsToContigsCollection.mapPartitionsToPair(iter -> () -> {
+        final JavaPairRDD<String, AssembledBreakpoint> assembledBreakpoints = breakpointIdsToContigsCollection.mapPartitionsToPair(iter -> () -> {
             try {
-                final String ref = referenceArguments.getReferenceFileName();
-                final File localRef = BucketUtils.isHadoopUrl(ref) ? BwaSparkEngine.localizeReferenceAndBwaIndexFiles(ref) : new File(ref);
-                try (ContigAligner contigAligner = new ContigAligner(localRef)) {
-                    final List<Tuple2<String, ContigAligner.AssembledBreakpoint>> results = new ArrayList<>();
+                try (ContigAligner contigAligner = new ContigAligner(referenceArguments.getReferenceFileName())) {
```
SHuang-Broad: trivial, but `final`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
```diff
@@ -0,0 +1,86 @@
...
+                inputAssemblies
+                        .flatMapToPair(RunSGAViaProcessBuilderOnSpark::splitAssemblyLine)
+                        .coalesce(numPartitions);
+
+        final JavaPairRDD<String, ContigsCollection> breakpointIdsToContigsCollection =
+                contigCollectionByBreakpointId.mapValues(ContigsCollection::fromPackedFasta);
+
+        final JavaPairRDD<String, AssembledBreakpoint> assembledBreakpoints = breakpointIdsToContigsCollection.mapPartitionsToPair(iter -> () -> {
+            try {
+                try (ContigAligner contigAligner = new ContigAligner(referenceArguments.getReferenceFileName())) {
```
SHuang-Broad: trivial, but `final`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,277 @@
...
+                    final List<AlignmentRegion> alignmentRegionList = Arrays.stream(alnRgns)
+                            .filter(a -> a.getSecondary() < 0)
+                            .map(AlignmentRegion::new)
+                            .sorted(Comparator.comparing(a -> a.startInAssembledContig))
+                            .collect(arrayListCollector(alnRgns.length));
+
+
+                    final Iterator<AlignmentRegion> iterator = alignmentRegionList.iterator();
+                    if ( iterator.hasNext() ) {
+                        AlignmentRegion current = iterator.next();
```
SHuang-Broad: Personally, I find the name "clash" of the two iterators a bit confusing to read.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigAligner.java
```diff
@@ -0,0 +1,277 @@
...
+        }
+    }
+
+    private static class LocalizedReference {
+        static File INSTANCE;
+
+        static {
+            try {
+                INSTANCE = BucketUtils.isHadoopUrl(ContigAligner.referencePath) ? BwaSparkEngine.localizeReferenceAndBwaIndexFiles(ContigAligner.referencePath) : new File(ContigAligner.referencePath);
+            } catch (IOException e) {
```
SHuang-Broad: nitpicking, but `final`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -80,6 +78,21 @@
     private static final DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
     private static final Logger logger = LogManager.getLogger(RunSGAViaProcessBuilderOnSpark.class);
 
+    /**
+     * input format is tab separated BreakpointId, PackedFastaFile
+     * @param assemblyLine
+     * @return
```
SHuang-Broad: It would be nice if `@return` is documented that the returned set can be empty.

SHuang-Broad:
Some really trivial comments. Thanks.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignContigsAndCallBreakpointsSpark.java
cwhelan: done

tedsharpe:
I think we're ready for a rebase, squash, and merge.  @SHuang-Broad do you agree?

SHuang-Broad:
@tedsharpe @cwhelan , yes, agreed.

PR #1941, opened 2016-06-24
remove commandline bwa because we have jni now

akiezun:
this PR removes the commandline BWA (because we have JNI now). 

@lbergelson can you have a look?


akiezun:
@cwhelan @SHuang-Broad should we keep this for now or it is now obsolete?

cwhelan:
We are currently using native BWA only. However we'd like to be able to
modify some things -- for example we'd like to change BWA alignment
parameters for our purposes (
https://github.com/broadinstitute/gatk/issues/1942). If it's too hard to
push changes through jBWA we might have to fall back on the command line
version. What do you think?

SHuang-Broad:
Some updates issued, see discussions in #1942

lbergelson:
@cwhelan @SHuang-Broad So is this +1 or -1 from you guys?  From my perspective this is fine.  I'm not a contributor to jbwa, so we can get stuff merged quickly if we need to and publish versions as necessary.

SHuang-Broad:
Fine from my perspective. Thanks @lbergelson

droazen:
@cwhelan any comments here? If not, send back to @akiezun

SHuang-Broad:
@akiezun , the whole directory `src/test/resources/org/broadinstitute/hellbender/tools/spark/sv/RunBWAMEMViaCommandLine` can be deleted as well.

cwhelan:
@droazen @akiezun Seems like this is fine to delete since it's easy to make changes to jbwa itself.

PR #1947, opened 2016-06-27
Implemented getExecutorCores() and getMemoryPerCore() to command line tools and added sensible local tests

jamesemery:
Currently the behavior is not guaranteed in the situation where the user doesn't specify either spark.executor.cores or spark.executor.memory and is running with autoscaling enabled. Additionally the program will return a different value (the maximum memory available for caching) if the user is running in local mode.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgram.java
@@ -3,11 +3,12 @@ package org.broadinstitute.hellbender.engine.spark;
 import com.google.cloud.dataflow.sdk.options.PipelineOptionsFactory;
 import com.google.cloud.genomics.dataflow.utils.GCSOptions;
 import org.apache.spark.api.java.JavaSparkContext;
+import org.apache.spark.scheduler.SchedulerBackend;
+import org.apache.spark.scheduler.local.LocalBackend;
 import org.broadinstitute.hellbender.cmdline.Argument;
 import org.broadinstitute.hellbender.cmdline.ArgumentCollection;
 import org.broadinstitute.hellbender.cmdline.CommandLineProgram;
 import org.broadinstitute.hellbender.engine.AuthHolder;
-
 import java.io.Serializable;
 
 
@@ -18,6 +19,7 @@ public abstract class SparkCommandLineProgram extends CommandLineProgram impleme
     @Argument(doc = "API Key for google cloud authentication",
             shortName = "apiKey", fullName = "apiKey", optional=true)
     protected String apiKey = null;
+    JavaSparkContext ctx;
 
     @Argument(
             doc = "Name of the program running",
@@ -31,9 +33,15 @@ public abstract class SparkCommandLineProgram extends CommandLineProgram impleme
     public SparkCommandLineArgumentCollection sparkArgs = new SparkCommandLineArgumentCollection();
 
 
+
+    @Override
+    protected void onStartup() {
+        super.onStartup();
+        ctx = SparkContextFactory.getSparkContext(getProgramName(), sparkArgs.getSparkProperties(), sparkArgs.getSparkMaster());
+    }
+
     @Override
     protected Object doWork() {
-        final JavaSparkContext ctx = SparkContextFactory.getSparkContext(getProgramName(), sparkArgs.getSparkProperties(), sparkArgs.getSparkMaster());
         try{
             runPipeline(ctx);
             return null;
@@ -85,4 +93,36 @@ public abstract class SparkCommandLineProgram extends CommandLineProgram impleme
     protected String getProgramName(){
         return programName == null ? getClass().getSimpleName() : programName;
     }
+
+    /**
+     * Returns the number of cores being used by each worker.
+     * In local mode it will return the total number of cores allocated to the process.
+     */
+    protected int getExecutorCores() {
+        int cores;
+        SchedulerBackend schedulerBackend = ctx.sc().schedulerBackend();
+        if (schedulerBackend instanceof LocalBackend) {
+            cores = ((LocalBackend) schedulerBackend).totalCores();
+        } else {
+            cores = ctx.getConf().getInt("spark.executor.cores",1); // Spark defaults to 1 core per executor if no value is set
+        }
+        return cores;
+    }
+
+    /**
+     * Returns the total memory allocated to each core (in MB)
+     *
+     * NOTE: in local mode this returns the ammount of memory available to the primary executor for caching
+     * by default this value will be approximately 0.5 of the available memory for one core. Additionally,
+     * there is no guarantee that this result will be correct if the user never specifies either 'spark.executor.memory'
+     * or 'spark.executor.cores'
+     */
+    public double getMemoryPerCore() {
+        if (ctx.isLocal()) {
+            // In local mode, poll the active executor directly for the memory it has
+            return ((long)ctx.sc().getExecutorMemoryStatus().valuesIterator().next()._1() / ((double)getExecutorCores()))/(1000000.0);
+        }
+        return ctx.sc().executorMemory() / ((double)getExecutorCores());
+    }
+
 }
--- a/src/test/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgramUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgramUnitTest.java
@@ -16,8 +16,10 @@ import java.util.Map;
 
 
 public class SparkCommandLineProgramUnitTest extends BaseTest {
+
     @CommandLineProgramProperties(summary = "test", programGroup = TestSparkProgramGroup.class, oneLineSummary = "line")
     private static class TestSparkCommandLineProgram extends SparkCommandLineProgram {
+        private static final long serialVersionUID = 1L;
         @Override
         protected void runPipeline(JavaSparkContext ctx) {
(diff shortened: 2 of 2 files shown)


droazen:
@tedsharpe please review

droazen:
You have a compile error btw @jamesemery:

```
:compileTestJava/home/travis/build/broadinstitute/gatk/src/test/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgramUnitTest.java:20: warning: [serial] serializable class TestSparkCommandLineProgram has no definition of serialVersionUID
    private static class TestSparkCommandLineProgram extends SparkCommandLineProgram {
                   ^
```

tedsharpe:
Doing another PR review.  Will get to this tomorrow.  Apologies for the delay.

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgram.java
```diff
@@ -85,4 +93,36 @@ protected void afterPipeline(final JavaSparkContext ctx) {
     protected String getProgramName(){
         return programName == null ? getClass().getSimpleName() : programName;
     }
+
+    /**
+     * Returns the number of cores being used by each worker.
+     * In local mode it will return the total number of cores allocated to the process.
+     */
+    protected int getExecutorCores() {
```
tedsharpe: Might as well be public.

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgram.java
```diff
@@ -85,4 +93,36 @@ protected void afterPipeline(final JavaSparkContext ctx) {
...
+        int cores;
+        SchedulerBackend schedulerBackend = ctx.sc().schedulerBackend();
```
tedsharpe: The schedulerBackend method is undocumented as of Spark 1.4.  Maybe not wise to rely on it?

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgram.java
tedsharpe: I think we could replace this method with a single line:
`return ctx.defaultParallelism();`
My testing seems to suggest this gives the right answer locally and on a cluster.  One big gotcha, however: there's a race condition where if you call it too quickly on a brand new JavaSparkContext when on a real cluster, it gives a bogus answer.  I have no idea what to do about that except to document it.

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgram.java
```diff
@@ -85,4 +93,36 @@ protected void afterPipeline(final JavaSparkContext ctx) {
...
+     * by default this value will be approximately 0.5 of the available memory for one core. Additionally,
+     * there is no guarantee that this result will be correct if the user never specifies either 'spark.executor.memory'
+     * or 'spark.executor.cores'
+     */
+    public double getMemoryPerCore() {
+        if (ctx.isLocal()) {
+            // In local mode, poll the active executor directly for the memory it has
+            return ((long)ctx.sc().getExecutorMemoryStatus().valuesIterator().next()._1() / ((double)getExecutorCores()))/(1000000.0);
+        }
+        return ctx.sc().executorMemory() / ((double)getExecutorCores());
```
tedsharpe: The method `executorMemory()` has also become undocumented as of Spark 1.4.  But in any case, this returns total executor memory in MB, without subtracting reserve and user memory.  Whereas `getExecutorMemoryStatus()` returns just the Spark memory (cache and storage).  So the two methods aren't commensurate.
I'd like to see us use `getExecutorMemoryStatus()` both locally and on the cluster.  There are two problems, however:  There's a race condition where if you call it too quickly on a brand new JavaSparkContext when on a real cluster, it gives a bogus answer (it just reports the driver).  The second problem is that it's not clear how to identify which Map entry is for the driver and which entries are for nodes.  But if we can figure out how to do that, we could return a nice report that gave nicely complete information:  driver memory, executor spark memory per core, executor user memory per core, number of executors, number of cores per executor.  A pretty complete description of the execution environment.  This would be really valuable.
Can you figure out how to identify the driver?

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgram.java
```diff
@@ -85,4 +93,36 @@ protected void afterPipeline(final JavaSparkContext ctx) {
...
+     *
+     * NOTE: in local mode this returns the ammount of memory available to the primary executor for caching
+     * by default this value will be approximately 0.5 of the available memory for one core. Additionally,
+     * there is no guarantee that this result will be correct if the user never specifies either 'spark.executor.memory'
+     * or 'spark.executor.cores'
+     */
+    public double getMemoryPerCore() {
+        if (ctx.isLocal()) {
+            // In local mode, poll the active executor directly for the memory it has
+            return ((long)ctx.sc().getExecutorMemoryStatus().valuesIterator().next()._1() / ((double)getExecutorCores()))/(1000000.0);
```
tedsharpe: I think this should be 1024.*1024., not 1,000,000.

src/test/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgramUnitTest.java
```diff
@@ -0,0 +1,69 @@
...
+        @Override
+        protected void runPipeline(JavaSparkContext ctx) {
+        }
+    };
+
+    // Because the exact values here are system dependant in local mode, this tests are just sanity tests
+    @Test (dataProvider = "testMemoryPerCoreData")
+    public void testMemoryPerCore(JavaSparkContext context, double expected) {
+        TestSparkCommandLineProgram prog = new TestSparkCommandLineProgram();
+        prog.ctx = context;
```
tedsharpe: maybe call `onStartup()` rather than poking a value in there.  (that would also let you make ctx private.)

src/test/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgramUnitTest.java
```diff
@@ -0,0 +1,69 @@
...
+        Assert.assertNotEquals(prog.getMemoryPerCore(),0);
+        Assert.assertEquals(prog.getMemoryPerCore()>0,true);
+    }
+
+
+
+    @Test (dataProvider = "testMemoryPerCoreData")
+    public void testCoresPerExecutor(JavaSparkContext context, double expected) {
+        TestSparkCommandLineProgram prog = new TestSparkCommandLineProgram();
+        prog.ctx = context;
```
tedsharpe: `onStartup()`

src/test/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgramUnitTest.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+    };
+
+    // Because the exact values here are system dependant in local mode, this tests are just sanity tests
+    @Test (dataProvider = "testMemoryPerCoreData")
+    public void testMemoryPerCore(JavaSparkContext context, double expected) {
+        TestSparkCommandLineProgram prog = new TestSparkCommandLineProgram();
+        prog.ctx = context;
+        Assert.assertNotEquals(prog.getMemoryPerCore(),0);
+        Assert.assertEquals(prog.getMemoryPerCore()>0,true);
```
tedsharpe: Let's test against the actual expected value.

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgram.java
tedsharpe: But I'd rather have the answer in bytes rather than MB, anyway.

src/test/java/org/broadinstitute/hellbender/engine/spark/SparkCommandLineProgramUnitTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @DataProvider (name = "testMemoryPerCoreData")
+    public Object[][] testMemoryPerCoreData() {
+        return new Object[][] {
+                {makeSparkContext("1", "1G", "3", "4", "12G", true), 3072.0}
+        };
+    }
+
+    private JavaSparkContext makeSparkContext(String driverCores, String driverMemory,
```
tedsharpe: Isn't this whole thing completely irrelevant?

droazen:
@tomwhite Can you give your thoughts on this one? It's turning out to be much trickier than we thought...

Thanks!

tomwhite:
I agree with @tedsharpe that `ctx.defaultParallelism()` is the recommended way to get the number of cores. If there's a race condition, then that is a bug that should be reported (stack trace if possible), so it can be fixed in Spark.

Getting memory is harder. The `executorMemory()` method is probably the best bet (although it is not a public API any more), and for the local case could you use the JVM memory?

droazen:
Perhaps the thing to do here is to move the methods into a utils class like `SparkUtils` with appropriate disclaimers in the method docs, instead of having them in `SparkCommandLineProgram`, which implies that they're 100% rock solid.

lbergelson:
We're going to close this.  It seems like spark doesn't provide a way to get this information in a reasonable way.  The information we can get isn't reliable enough to make decisions about for most purposes.

PR #1953, opened 2016-06-30
Cleanup - squid:S2864 - "entrySet()" should be iterated when both the key and value are needed

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S2864 - "entrySet()" should be iterated when both the key and value are needed

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2864

Please let me know if you have any questions.

M-Ezzat


akiezun:
The title does not correspond to the changes

m-ezzat:
Oh, that's true, I updated it.

akiezun:
i vote 👎 - it's not cleaner I think. It may be faster but you'd need to show me numbers.

PR #1954, opened 2016-06-30
Changes max value output for ExcessHet

meganshand:
Port from [broadinstitue/gsa-unstable#1426](https://github.com/broadinstitute/gsa-unstable/pull/1426)

@akiezun Could you please take a look since there were some changes to ExcessHet between GATK3 and 4?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHet.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHet.java
@@ -26,6 +26,7 @@ import java.util.Map;
  */
 public final class ExcessHet extends InfoFieldAnnotation implements StandardAnnotation {
     private static final double MIN_NEEDED_VALUE = 1.0E-16;
+    public static final double PHRED_SCALED_MIN_P_VALUE = -10.0 * Math.log10(MIN_NEEDED_VALUE);
 
     @Override
     public Map<String, Object> annotate(final ReferenceContext ref,
@@ -58,8 +59,9 @@ public final class ExcessHet extends InfoFieldAnnotation implements StandardAnno
         final double pval = exactTest(hetCount, refCount, homCount);
 
         //If the actual phredPval would be infinity we will probably still filter out just a very large number
-        if (Math.abs(pval - 0.0) < 10e-60) {
-            return Pair.of(Integer.MAX_VALUE, Double.POSITIVE_INFINITY);
+        //Since the method does not guarantee precision for any p-value smaller than 1e-16, we can return the phred scaled version
+        if (pval < 10e-60) {
+            return Pair.of(sampleCount, PHRED_SCALED_MIN_P_VALUE);
         }
         final double phredPval = -10.0 * Math.log10(pval);
 
--- a/src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHetUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHetUnitTest.java
@@ -182,6 +182,9 @@ public final class ExcessHetUnitTest extends BaseTest {
         final double hetsValue = new ExcessHet().calculateEH(allHet, allHet.getGenotypes()).getValue();
 
         Assert.assertTrue(Math.abs(singletonValue) < Math.abs(hetsValue), String.format("singleton=%f allHets=%f", singletonValue, hetsValue));
+
+        //Since all hets is such an extreme case and the sample size is large here, we know that the p-value should be 0
+        Assert.assertEquals(hetsValue, ExcessHet.PHRED_SCALED_MIN_P_VALUE, DELTA_PRECISION, String.format("P-value of 0 should be phred scaled to " + ExcessHet.PHRED_SCALED_MIN_P_VALUE));
     }
 
     @DataProvider(name = "smallSets")


src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHetUnitTest.java
```diff
@@ -182,6 +182,9 @@ public void testAllHetsForLargeCohorts() {
         final double hetsValue = new ExcessHet().calculateEH(allHet, allHet.getGenotypes()).getValue();
 
         Assert.assertTrue(Math.abs(singletonValue) < Math.abs(hetsValue), String.format("singleton=%f allHets=%f", singletonValue, hetsValue));
+
+        //Since all hets is such an extreme case and the sample size is large here, we know that the p-value should be 0
+        Assert.assertTrue(hetsValue == 160.0, String.format("P-value of 0 should be phred scaled to 160.0"));
```
akiezun: and then use that constant here, rather than hardwiring 160

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHet.java
```diff
@@ -58,8 +58,9 @@
         final double pval = exactTest(hetCount, refCount, homCount);
 
         //If the actual phredPval would be infinity we will probably still filter out just a very large number
+        //Since the method does not guarantee precision for any p-value smaller than 1e-16, we can return the phred scaled version
         if (Math.abs(pval - 0.0) < 10e-60) {
-            return Pair.of(Integer.MAX_VALUE, Double.POSITIVE_INFINITY);
+            return Pair.of(sampleCount, -10.0 * Math.log10(MIN_NEEDED_VALUE));
```
akiezun: can you extract `-10.0 * Math.log10(MIN_NEEDED_VALUE)` as a constant? I dont want to keep recomputing it.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHet.java
```diff
@@ -58,8 +58,9 @@
         final double pval = exactTest(hetCount, refCount, homCount);
 
         //If the actual phredPval would be infinity we will probably still filter out just a very large number
+        //Since the method does not guarantee precision for any p-value smaller than 1e-16, we can return the phred scaled version
         if (Math.abs(pval - 0.0) < 10e-60) {
```
akiezun: what's the meaning of `pval - 0.0` ? Is that not just `pval`?

akiezun:
back to @meganshand for some small fixes

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHetUnitTest.java
meganshand: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHet.java
meganshand: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHet.java
meganshand: I think this is the same as (pval < 10e-60) unless the absolute value was to protect against some kind of overflow that I'm not seeing? I changed it to pval < 10e-60.

meganshand:
Changes made, back to @akiezun

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHetUnitTest.java
```diff
@@ -182,6 +182,9 @@ public void testAllHetsForLargeCohorts() {
         final double hetsValue = new ExcessHet().calculateEH(allHet, allHet.getGenotypes()).getValue();
 
         Assert.assertTrue(Math.abs(singletonValue) < Math.abs(hetsValue), String.format("singleton=%f allHets=%f", singletonValue, hetsValue));
+
+        //Since all hets is such an extreme case and the sample size is large here, we know that the p-value should be 0
+        Assert.assertTrue(hetsValue == ExcessHet.getMinPhredScaledValue(), String.format("P-value of 0 should be phred scaled to 160.0"));
```
akiezun: just make the constant public and remove the method. there's no value in the `getMinPhredScaledValue ` method. Also, use the constant in the other place (on the same line!) that hardcodes `160.0`

akiezun:
back to @meganshand to do one more tweak and then rebase/squash/merge.

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHetUnitTest.java
akiezun: one more thing - comparing `doubles` by `==` is always a bad idea. Compare by `assertEquals` and provide a diff threashold

meganshand: fixed

akiezun:
👍

PR #1936, opened 2016-06-23
Enable query-by-interval of VCFs (including blockgzipped VCFs) on Spark.

tomwhite:
Fixes #1651.

Thanks to @droazen for suggesting the use of `-R` as a source for the sequence dictionary.

PR #1956, opened 2016-06-30
add blocked gvcf to attributes

akiezun:
these 2 files were missing from #1936

@lbergelson ok to merge?


lbergelson:
@akiezun Is the failure expected?  It looks like a problem with an lfs file.

lbergelson:
I restarted the travis build.  If it passes then 👍

akiezun:
it passes

Issue #1608, opened 2016-03-22
Beat GATK3 performance (CPU and memory) of GenotypeGVCFs


akiezun:
the goal is to beat GATK3 performance and memory usage of GenotypeGVCFs on realistic datasizes

akiezun:
blocked on #15

akiezun:
first measurement after code in master (file is a wide gvcf with 10k samples and 1k lines : 258M)
GATK3
real	0m31.092s
user	0m46.361s

GATK4
real	0m46.806s
user	0m55.150s

akiezun:
after more aggressive caching of log10 values (a615f8d9c8)
GATK4
real	0m37.982s
user	0m46.861s

droazen:
@lbergelson What input(s) were you using where you saw an order-of-magnitude speedup vs. GATK3?

lbergelson:
I was using a long thin file.

akiezun:
Running with the log10 cache on a non-bp file, 10k samples, 1k lines: 186M

GATK3.5
real	0m26.499s
user	0m42.367s

GATK4
real	0m33.999s
user	0m41.278s

akiezun:
small loop code lifting

GATK4 bp file
real	0m35.082s
user	0m43.495s

GATK4 no-bp file
real	0m31.129s
user	0m40.782s

droazen:
@akiezun Try playing with different values of `VariantWalker.FEATURE_CACHE_LOOKAHEAD` and see if that has an effect.

akiezun:
sharing GenotypeLikelihoodCalculators
GATK4 bp file
real	0m31.917s
user	0m41.398s

GATK4 no-bp file
real	0m26.793s
user	0m36.451s

akiezun:
@droazen i doubt it will help - most work is in the math + vcf decoding (we could move the latter to a different thread eventually i think)

akiezun:
more code lifting out of loops
GATK4 bp file
real	0m32.947s
user	0m41.479s

GATK4 no-bp file
real	0m26.206s
user	0m35.763s

akiezun:
(we're not beating the perf yet, only matching)

akiezun:
@droazen what do you want to do here? We're matching the performance but not beating it at this point. We can either keep this open and continue to investigate or close as 'matching not beating'

droazen:
[...]

PR #1957, opened 2016-06-30
performance optimizations for GenotypeGVCFs

akiezun:
part of work tracked here https://github.com/broadinstitute/gatk/issues/1608

@lbergelson please review. The corresponding protected branch is https://github.com/broadinstitute/gatk-protected/tree/ak_ggvcf_performance

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/ReferenceDiploidExactAFCalculator.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/ReferenceDiploidExactAFCalculator.java
@@ -126,18 +126,20 @@ public final class ReferenceDiploidExactAFCalculator extends DiploidExactAFCalcu
 
     private static void computeLofK(final ExactACset set,
                                     final List<double[]> genotypeLikelihoods,
-                                    final double[] log10AlleleFrequencyPriors, final StateTracker stateTracker) {
+                                    final double[] log10AlleleFrequencyPriors,
+                                    final StateTracker stateTracker) {
 
-        set.getLog10Likelihoods()[0] = 0.0; // the zero case
+        final double[] setLog10Likelihoods = set.getLog10Likelihoods();
+        setLog10Likelihoods[0] = 0.0; // the zero case
         final int totalK = set.getACsum();
 
         // special case for k = 0 over all k
         if ( totalK == 0 ) {
-            for ( int j = 1; j < set.getLog10Likelihoods().length; j++ ) {
-                set.getLog10Likelihoods()[j] = set.getLog10Likelihoods()[j - 1] + genotypeLikelihoods.get(j)[HOM_REF_INDEX];
+            for (int j = 1, n = setLog10Likelihoods.length; j < n; j++ ) {
+                setLog10Likelihoods[j] = setLog10Likelihoods[j - 1] + genotypeLikelihoods.get(j)[HOM_REF_INDEX];
             }
 
-            final double log10Lof0 = set.getLog10Likelihoods()[set.getLog10Likelihoods().length-1];
+            final double log10Lof0 = setLog10Likelihoods[setLog10Likelihoods.length-1];
             stateTracker.setLog10LikelihoodOfAFzero(log10Lof0);
             stateTracker.setLog10PosteriorOfAFzero(log10Lof0 + log10AlleleFrequencyPriors[0]);
             return;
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/ReferenceDiploidExactAFCalculator.java
```diff
@@ -209,12 +211,16 @@ private static void pushData(final ExactACset targetSet,
                                  final List<double[]> genotypeLikelihoods) {
         final int totalK = targetSet.getACsum();
 
-        for ( int j = 1; j < targetSet.getLog10Likelihoods().length; j++ ) {
-            if ( totalK <= 2*j ) { // skip impossible conformations
+        final double[] targetSetLog10Likelihoods = targetSet.getLog10Likelihoods();
+        final double[] dependentSetLog10Likelihoods = dependentSet.getLog10Likelihoods();
+        final int[] counts = targetSet.getACcounts().getCounts();
+
+        for ( int j = (totalK/2)-1, n = targetSetLog10Likelihoods.length; j < n; j++ ) {
```
lbergelson: If `totalK` is 0 (or less) this produces a different result than before.  I'm not sure if that's possible or not.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -308,12 +309,14 @@ public static double log10sumLog10(final double[] log10p, final int start, final
 
 
     /**
-     * A helper class to maintain a cache of log values.
-     * The cache is immutable after creation.
+     * A helper class to maintain a cache of log10 values.
+     * The cache expands when a number is not available.
+     * NOTE: this cache is NOT thread safe and it must not be accessed from multiple threads.
      */
     private static final class Log10Cache {
```
lbergelson: An instance of this cache is you used to back the static method `MathUtils.log10`.  It was previously effectively immutable so it wasn't a problem to have a static cache. Now that this is mutable though I think you should probably guard those instances in ThreadLocals so that it won't be caught in some broken state if multiple threads call `MathUtils.log10`

lbergelson:
@akiezun minor comments

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/ReferenceDiploidExactAFCalculator.java
akiezun: good catch. i think totalK may be 0 sometimes. I'll leave this as `for ( int j = 1, n = targetSetLog10Likelihoods.length; j < n; j++ ) {`

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
akiezun: you're right. done

akiezun: actually, i think the gatk3 code is better than doing the thread local (which requires a map lookup for every call). I'll put the gatk3 version in for the second review

akiezun:
back to @lbergelson for second review. log10 cache is now static (and synchronized at expansion only) as in gatk3 - it looks faster (no map lookup that would be needed in the thread local solution)

lbergelson:
@akiezun My only concern now is that someone takes log10 of a very large number triggering a massive and slow cache expansion.  This caching scheme is good for clustered queries of small values, but terrible for sparse large queries.  Is that a case we need to consider?

akiezun:
The access pattern I saw is that the queries start small and creep up to
tens or hundreds of thousands (Order of the queried numbers). With the
exponential cache expansion, the cache quickly catches up. So it's fine.
We'll revise when needed.

On Wednesday, July 6, 2016, Louis Bergelson <notifications@github.com>
wrote:

> @akiezun <https://github.com/akiezun> My only concern now is that someone
> takes log10 of a very large number triggering a massive and slow cache
> expansion. This caching scheme is good for clustered queries of small
> values, but terrible for sparse large queries. Is that a case we need to
> consider?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/broadinstitute/gatk/pull/1957#issuecomment-230863916>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AB5rL32j0zFsPo1VQ4T6gxnEHvjpYZ02ks5qS_VhgaJpZM4JCXE9>
> .
>


-- 
Sent from Gmail Mobile

lbergelson:
👍

Issue #1512, opened 2016-02-20
incorrect test CollectInsertSizeMetricsTest - not catching an incorrect BAM and CRAM files 


akiezun:
this bam file has no reads: `src/test/resources/org/broadinstitute/hellbender/tools/picard/analysis/CollectInsertSizeMetrics/insert_size_metrics_test.bam`

the cram file also has no reads ``src/test/resources/org/broadinstitute/hellbender/tools/picard/analysis/CollectInsertSizeMetrics/insert_size_metrics_test.cram`

but the corresponding sam file has 52 reads: `src/test/resources/org/broadinstitute/hellbender/tools/picard/analysis/CollectInsertSizeMetrics/insert_size_metrics_test.sam`

However `CollectInsertSizeMetricsTest` does not catch this and incorrectly reports that the test successfully passed. The issue comes from using only a for loop for asserts (the for loop executes 0 times on this file and so it thinks everything is great)

The task here is to:
- fix the test to catch this error, then
- fix the bam file to not have this problem, then
- check and fix all other tests that use the same testing pattern (using the forloop)

I suggest switching to using actual files for such tests (as done in `MeanQualityByCycleIntegrationTest`) to avoid such problems in the future

akiezun:
@cmnbroad can you have a look? Not sure what went wrong with making the bam and cram files here

akiezun:
thanks @SHuang-Broad for pointing out this problem to me

PR #1959, opened 2016-06-30
Spark multilevel metrics collection refactoring.

cmnbroad:
- The main change is the use of a reducible multi level collector; currently there is only one consumer which is InsertizeMetricsCollector.
- There is a long comment and diagram at the top of MetricsCollectorSpark that describes the roles and relationships of the various component classes that make up a collector. Its a good place to start to understand these changes.
- The Spark and walker versions of CollectInsertSizeMetrics now use the same MultiLevelCollectorReducible-derived implementation (InsertSizeCollector), and exactly the same tests and expected results.
- All of the CollectInsertSizeMetricsSpark code and corresponding unit tests from the previous implementation are removed in this PR.
- Fixes https://github.com/broadinstitute/gatk/issues/1512.
- Both CollectMultipleMetrics (Spark and walker) implementations include support and tests for external/custom collectors.


droazen:
@cmnbroad Can you nominate a reviewer for this branch?

cmnbroad:
Nominating @akiezun. @lbergelson knows the metrics code pretty well too.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsArgumentCollection.java
```diff
@@ -16,24 +16,36 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    @Argument(fullName = "histogramPlotFile",
+            shortName="H",
+            doc="File to write insert size Histogram chart to.")
+    public String histogramPlotFile;
 
-    @Argument(doc = "Generate mean, sd and plots by trimming the data down to MEDIAN + maxMADTolerance*MEDIAN_ABSOLUTE_DEVIATION. " +
+    @Argument(doc="Generate mean, sd and plots by trimming the data down to MEDIAN + DEVIATIONS*MEDIAN_ABSOLUTE_DEVIATION. " +
```
akiezun: why this change? the comment is now harder to understand, ie what is `DEVIATIONS` ?

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -1,50 +1,66 @@
...
 
 /**
- * Collects InsertSizeMetrics on the specified accumulationLevels using
+ * Collects InsertSizeMetrics on the specified accumulationLevels
  */
-public final class InsertSizeMetricsCollector
-        extends MultiLevelCollector<InsertSizeMetrics, Integer, InsertSizeMetricsCollectorArgs>
-        implements Serializable {
-
+public class InsertSizeMetricsCollector
```
akiezun: why not final anymore? Is it designed to be subclassed? there are no subclasses now and let's assume the will be none

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -1,50 +1,66 @@
...
-        this.HistogramWidth = HistogramWidth;
-        this.deviations = deviations;
-        setup(accumulationLevels, samRgRecords);
+    InsertSizeMetricsArgumentCollection inputArgs = null;
+
+    public InsertSizeMetricsCollector(
+            final InsertSizeMetricsArgumentCollection inputArgs,
+            final SAMFileHeader samHeader)
+    {
+        super();
```
akiezun: explicit `super()` calls in constructors are not required

akiezun: can any of the arguments be `null`? If not, add checks. If yes, document it.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -1,50 +1,66 @@
...
     }
 
-    // We will pass insertSize and PairOrientation with the DefaultPerRecordCollectorArgs passed to the record collectors
-    // This method is called once Per samRecord
+    /**
+     * Return the read filter for InsertSizeMetrics collector.
+     * @param samHeader SAMFileHeader for the input file
+     * @return ReadFilter to be used to filter records
+     */
+    public ReadFilter getReadFilter(SAMFileHeader samHeader) {
```
akiezun: add null check for the header

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -1,50 +1,66 @@
...
+                    inputArgs.MQPassingThreshold
+            )
+        );
+    }
+
+    // We will pass insertSize and PairOrientation with the DefaultPerRecordCollectorArgs passed to
+    // the record collectors. This method is called once per samRecord
     @Override
     protected InsertSizeMetricsCollectorArgs makeArg(SAMRecord samRecord, ReferenceSequence refSeq) {
         final int insertSize = Math.abs(samRecord.getInferredInsertSize());
```
akiezun: add a comment why the abs call is needed

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -53,135 +69,154 @@ protected InsertSizeMetricsCollectorArgs makeArg(SAMRecord samRecord, ReferenceS
         return new InsertSizeMetricsCollectorArgs(insertSize, orientation);
     }
 
-    /** Make an InsertSizeCollector with the given arguments */
+    // Make an PerUnitInsertSizeMetricsCollector with the given arguments
     @Override
-    protected PerUnitMetricCollector<InsertSizeMetrics, Integer, InsertSizeMetricsCollectorArgs> makeChildCollector(final String sample, final String library, final String readGroup) {
-        return new PerUnitInsertSizeMetricsCollector(sample, library, readGroup);
+    protected PerUnitInsertSizeMetricsCollector makeChildCollector(
+            final String sample,
```
akiezun: check all arguments for null

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -53,135 +69,154 @@ protected InsertSizeMetricsCollectorArgs makeArg(SAMRecord samRecord, ReferenceS
...
-                record.getReadUnmappedFlag() ||
-                record.getMateUnmappedFlag() ||
-                record.getFirstOfPairFlag() ||
-                record.isSecondaryOrSupplementary() ||
-                record.getDuplicateReadFlag() ||
-                record.getInferredInsertSize() == 0) {
-            return;
-        }
-
         super.acceptRecord(record, refSeq);
```
akiezun: if all this method does is call super, then it's useless and we could delete it from here.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -53,135 +69,154 @@ protected InsertSizeMetricsCollectorArgs makeArg(SAMRecord samRecord, ReferenceS
...
-                prefix = this.sample + ".";
+    /**
+     * Combine two InsertSizeMetricsCollector objects and return a single InsertSizeMetricsCollector
+     * object representing the combined results. NOTE: this implementation is destructive in that it
+     * merges the source into the target and returns the merged target as the combined object.
+     * @param target target destination of merged metrics
+     * @param source source of metrics to be merged into target
+     * @return single object representing the combined source and target objects
+     */
+    public InsertSizeMetricsCollector combine(InsertSizeMetricsCollector target, InsertSizeMetricsCollector source) {
```
akiezun: check null for arguments

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -53,135 +69,154 @@ protected InsertSizeMetricsCollectorArgs makeArg(SAMRecord samRecord, ReferenceS
...
+     * Combine two PerUnitInsertSizeMetricsCollector objects and return a single PerUnitInsertSizeMetricsCollector
+     * object representing the combined results.
+     * @param collector1 source PerUnitInsertSizeMetricsCollector
+     * @param collector2 target PerUnitInsertSizeMetricsCollector
+     * @return single PerUnitInsertSizeMetricsCollector object representing the combined source and target objects
+     */
+    public PerUnitInsertSizeMetricsCollector combineUnit(
+            PerUnitInsertSizeMetricsCollector collector1,
+            PerUnitInsertSizeMetricsCollector collector2) {
+        return collector1.combine(collector2);
```
akiezun: null check for args

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -53,135 +69,154 @@ protected InsertSizeMetricsCollectorArgs makeArg(SAMRecord samRecord, ReferenceS
...
+     * @param metricsFile a metricsFile where the collected metrics should be stored
+     * @param inputName the name of the input, for optional inclusion in the metrics file
+     * @param authHolder
+     */
+    public void finish(
+            final MetricsFile<InsertSizeMetrics, Integer> metricsFile,
+            final String inputName,
+            final AuthHolder authHolder)
+    {
+        super.finish();
```
akiezun: this is a strange call. Usually, methods only call super if they want to invoke the method that is overriden by the current method. Here, the call is to a **different** method hierarchy. Can this simply become a virtual call to `finish()` rather than a non-virtual hardwired call to `MultiLevelCollectorReducible.finish()` ?

src/test/resources/org/broadinstitute/hellbender/tools/picard/analysis/CollectInsertSizeMetrics/expectedQualityYieldOnInsertSizeMetrics.txt
```diff
@@ -0,0 +1,10 @@
+## htsjdk.samtools.metrics.StringHeader
+# org.broadinstitute.hellbender.tools.spark.pipelines.metrics.CollectMultipleMetricsSpark  --output /var/folders/3g/0dh8mspj3pz8qpmm4qqqzblr0000gn/T/collectMultiMetricsTest381676565295870422/collectMultiMetrics --collectors CollectInsertSizeMetrics --collectors CollectQualityYieldMetrics --input /Users/cmn/projects/hellbender/src/test/resources/org/broadinstitute/hellbender/tools/picard/analysis/CollectInsertSizeMetrics/insert_size_metrics_test.sam --verbosity ERROR  --assumeSorted true --metricAccumulationLevel ALL_READS --readValidationStringency SILENT --interval_set_rule UNION --interval_padding 0 --bamPartitionSize 0 --disableSequenceDictionaryValidation false --shardedOutput false --numReducers 0 --sparkMaster local[*] --help false --version false --QUIET false
```
akiezun: this file was clearly generated using hellbender test suite. is it the same as a file generated using picard on the same inputs? I think we should (whenever possible) use expected results that were created from picard (or gatk3 when appropriate)

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsArgumentCollection.java
cmnbroad: Also a result of consolidating two similar but not identical implementations. I'll revert this to the previous doc string so it matches the argument name.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: True that. Done.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: Yes. Done.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: I'm not sure why its this way - this code is in the existing walker implementation. The Picard code also does the same thing.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
akiezun: the reason is that the number is negative if the mate maps to lower position than the read. Would be useful to see this as a note here

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -14,41 +14,72 @@
...
  *
  *     CollectorType collector = new CollectorType<CollectorArgType>()
  *     CollectorArgType args = // get metric-specific input arguments
  *
- *     // pass the input arguments back to the collector for initialization
+ *     // pass the input arguments to the collector for initialization
  *     collector.initialize(args);
  *
- *     ReadFilter filter == collector.getReadFilter();
+ *     ReadFilter filter == collector.getReadFilter(samFileHeader);
```
akiezun: is this `samFileHeader` different from `getHeaderForReads()` 2 lines below? I think it's the same

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
```diff
@@ -99,21 +131,16 @@ default ReadFilter getReadFilter(SAMFileHeader samHeader) {
...
-    );
+    void collectMetrics(JavaRDD<GATKRead> filteredReads, SAMFileHeader samHeader);
 
     /**
      * This method is called after collectMetrics has returned and
      * all Spark partitions have completed metrics collection. This
      * gives the collector to serialize the resulting metrics, usually
      * to a metrics file.
+     * @param inputBaseName base name of the input file
+     * @param authHolder
```
akiezun: can `authHolder` be null? If yes, document it.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetrics.java
```diff
@@ -71,4 +69,12 @@
     public int WIDTH_OF_90_PERCENT;
     /** The "width" of the bins, centered around the median, that encompass 100% of all read pairs. */
     public int WIDTH_OF_99_PERCENT;
+
```
akiezun: Why is this not a javadoc?

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -0,0 +1,226 @@
...
+ */
+public final class PerUnitInsertSizeMetricsCollector
+        implements PerUnitMetricCollector<InsertSizeMetrics, Integer, InsertSizeMetricsCollectorArgs>,
+                    Serializable
+{
+    private static final long serialVersionUID = 1L;
+
+    // kryo can't serialize an EnumMap so use a LinkedHashMap (also so we can maintain order
+    // the insertion order of entries since the combining code below counts on it
+    final Map<SamPairUtil.PairOrientation, Histogram<Integer>> Histograms = new LinkedHashMap<>();
```
akiezun: lowercase names for fields

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -0,0 +1,226 @@
...
+    private final double minimumPct;
+
+    // Generate mean, sd and plots by trimming the data down to MEDIAN + DEVIATIONS*MEDIAN_ABSOLUTE_DEVIATION.
+    // This is done because insert size data typically includes enough anomolous values from chimeras and other
+    // artifacts to make the mean and sd grossly misleading regarding the real distribution.
+    private final double deviations;
+
+    //Explicitly sets the Histogram width, overriding automatic truncation of Histogram tail.
+    //Also, when calculating mean and stdev, only bins <= HISTOGRAM_WIDTH will be included.
+    private Integer HistogramWidth;
```
akiezun: lower case. also, why Integer and not int? If null is significant, mention why (or change to `OptionalInt` to make it super explicit

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -0,0 +1,226 @@
...
+
+    public PerUnitInsertSizeMetricsCollector(
+            final String sample,
+            final String library,
+            final String readGroup,
+            final double minimumPct,
+            final double deviations,
+            final Integer HistogramWidth) {
+
+        super();
```
akiezun: remove all those calls to super() in constructors

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -0,0 +1,226 @@
...
+            prefix = this.library + ".";
+        }
+        else if (this.sample != null) {
+            prefix = this.sample + ".";
+        }
+        else {
+            prefix = "All_Reads.";
+        }
+
+        Histograms.put(SamPairUtil.PairOrientation.FR,     new Histogram<Integer>("insert_size", prefix + "fr_count"));
```
akiezun: use diamonds, ie `new Histogram<>`

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    public void finish() { }
+
+    public double getTotalInserts() {
+        return totalInserts;
+    }
+
+    @SuppressWarnings("unchecked")
+    public void addMetricsToFile(final MetricsFile<InsertSizeMetrics, Integer> file) {
```
akiezun: add an `@Override` to make clear that it implements a method from interface (there's no `@Implement` so `@Override` has to do

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -0,0 +1,226 @@
...
+        Histograms.put(SamPairUtil.PairOrientation.FR,     new Histogram<Integer>("insert_size", prefix + "fr_count"));
+        Histograms.put(SamPairUtil.PairOrientation.TANDEM, new Histogram<Integer>("insert_size", prefix + "tandem_count"));
+        Histograms.put(SamPairUtil.PairOrientation.RF,     new Histogram<Integer>("insert_size", prefix + "rf_count"));
+    }
+
+    public void acceptRecord(final InsertSizeMetricsCollectorArgs args) {
+        Histograms.get(args.getPairOrientation()).increment(args.getInsertSize());
+    }
+
+    public void finish() { }
```
akiezun: `@Override`

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -0,0 +1,226 @@
...
+        else {
+            prefix = "All_Reads.";
+        }
+
+        Histograms.put(SamPairUtil.PairOrientation.FR,     new Histogram<Integer>("insert_size", prefix + "fr_count"));
+        Histograms.put(SamPairUtil.PairOrientation.TANDEM, new Histogram<Integer>("insert_size", prefix + "tandem_count"));
+        Histograms.put(SamPairUtil.PairOrientation.RF,     new Histogram<Integer>("insert_size", prefix + "rf_count"));
+    }
+
+    public void acceptRecord(final InsertSizeMetricsCollectorArgs args) {
```
akiezun: `@Override`

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -0,0 +1,226 @@
...
+                    String.format("%s (%s : %s)",
+                            message,
+                            source == null ? "null" : source,
+                            target == null ? "null" : target)
+            );
+        }
+    }
+
+    // Safely determine if two strings are equal, even if one or both are null.
+    private static boolean safeEqualStrings(final String source, final String target) {
```
akiezun: delete this and use `Objects.equals`

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -0,0 +1,226 @@
...
+        implements PerUnitMetricCollector<InsertSizeMetrics, Integer, InsertSizeMetricsCollectorArgs>,
+                    Serializable
+{
+    private static final long serialVersionUID = 1L;
+
+    // kryo can't serialize an EnumMap so use a LinkedHashMap (also so we can maintain order
+    // the insertion order of entries since the combining code below counts on it
+    final Map<SamPairUtil.PairOrientation, Histogram<Integer>> Histograms = new LinkedHashMap<>();
+
+    final String sample;
```
akiezun: why are those fields no private? should be private whenever possible

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -0,0 +1,226 @@
...
+        super();
+
+        this.sample = sample;
+        this.library = library;
+        this.readGroup = readGroup;
+        this.minimumPct = minimumPct;
+        this.HistogramWidth = HistogramWidth;
+        this.deviations = deviations;
+
+        String prefix = null;
```
akiezun: exctract all this to a method to create a prefix

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -53,135 +69,154 @@ protected InsertSizeMetricsCollectorArgs makeArg(SAMRecord samRecord, ReferenceS
...
+    private void writeHistogramPDF(final String inputName) throws RScriptExecutorException {
+        // path to Picard R script for producing histograms in PDF files.
+        final String R_SCRIPT = "insertSizeHistogram.R";
+
+        File histFile = new File(inputArgs.histogramPlotFile);
+        IOUtil.assertFileIsWritable(histFile);
+
+        final RScriptExecutor executor = new RScriptExecutor();
+        executor.addScript(new Resource(R_SCRIPT,
+                org.broadinstitute.hellbender.tools.picard.analysis.CollectInsertSizeMetrics.class));
```
akiezun: why use full name of this class?

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
```diff
@@ -53,135 +69,154 @@ protected InsertSizeMetricsCollectorArgs makeArg(SAMRecord samRecord, ReferenceS
...
+        executor.addArgs(
+                inputArgs.output,           // text-based metrics file
+                histFile.getAbsolutePath(), // PDF graphics file
+                inputName                   // input bam file
+        );
+        if (inputArgs.histogramWidth != null)
```
akiezun: braces here and everywhere

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollectorArgs.java
```diff
@@ -2,13 +2,17 @@
...
 import htsjdk.samtools.SamPairUtil;
 
+import java.io.Serializable;
+
 // Arguments that need to be calculated once per SAMRecord that are then passed to each PerUnitMetricCollector
 // for the given record
-final class InsertSizeMetricsCollectorArgs {
+final class InsertSizeMetricsCollectorArgs implements Serializable {
+
+    public static final long serialVersionUID = 1L;
```
akiezun: `serialVersionUID` should be `private static final long`

src/main/java/org/broadinstitute/hellbender/metrics/MetricsArgumentCollection.java
```diff
@@ -1,19 +1,20 @@
...
 import java.io.File;
+import java.io.Serializable;
 
 /**
  * Base class for defining a set of metrics collector arguments. The members should be
  * instantiable as command line arguments.
  */
-public class MetricsArgumentCollection {
+public class MetricsArgumentCollection implements Serializable {
+    public static final long serialVersionUID = 1L;
```
akiezun: `serialVersionUID` should be `private static final long`

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+ * You can then do these calculations in the makeArg method and they will only be done once per record.
+ *
+ * @param <METRIC_TYPE> The metrics being collected
+ * @param <HISTOGRAM_KEY> If there is are Histograms related to metrics of type <BEAN> then <HKEY> is the key value
+ *                       to these Histograms
+ * @param <ARGTYPE> The type argument passed to individual PerUnitMetricCollector (see SAMRecordMultilevelCollector
+ *                 and PerUnitMetricCollector)
+ * @param <UNIT_COLLECTOR> Class to represent metrics collected at a single level
+ */
+public abstract class MultiLevelCollectorReducible<
```
akiezun: why this name and not `MultiLevelReducibleCollector` ? Classes typically have names that are nouns, adjectives are sometimes used for interfaces like `Serializable`

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+        ARGTYPE,
+        UNIT_COLLECTOR extends PerUnitMetricCollector<METRIC_TYPE, HISTOGRAM_KEY, ARGTYPE>
+        >
+    implements Serializable
+{
+    private static final long serialVersionUID = 1L;
+
+    public static final String UNKNOWN = "unknown";
+
+    // A singleton UNIT_COLLECTOR that will accept all records
```
akiezun: Make this a javadoc - here and elsewhere,  all such comments should be in javadoc style

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+    // TODO: This should be moved into the AllReadDistributor, since it represents the lone
+    // unit collector for that level, but its here since its currently accessed from outside
+    protected UNIT_COLLECTOR allReadCollector;
```
akiezun: does this field need to be exposed like this and non encapsulated?

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+
+    // A list of Distributor that is at most length 4, 1 for each (ALL_READS, SAMPLE, LIBRARY, READ_GROUP)
+    // accumulation levels these will be listed in the order in which their children would be added to a
+    // metric file
+    protected List<Distributor> outputOrderedDistributors;
```
akiezun: exposing field like this is not as good as encapsulating them

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+     * }
+     *
+     * Such a wrapper can be passed directly to JavaRDD.reduce.
+     *
+     * @param source source MultiLevelCollectorReducible to be combined with this object
+     */
+    public void combine(MultiLevelCollectorReducible<METRIC_TYPE, HISTOGRAM_KEY, ARGTYPE, UNIT_COLLECTOR> source)
+    {
+        if ((allReadCollector == null) != (source.allReadCollector == null)) {
+            throw new IllegalArgumentException("MultiLevelCollectors must be the same size to be combined");
```
akiezun: somehow this messase seems weird given that it checks nullness of allReadCollector

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+        protected abstract UNIT_COLLECTOR makeCollector(final SAMReadGroupRecord rg);
+
+        protected abstract UNIT_COLLECTOR makeUnknownCollector();
+
+        public Distributor(final List<SAMReadGroupRecord> rgRecs) {
+            collectors = new LinkedHashMap<>();
+            for(final SAMReadGroupRecord rg : rgRecs) {
+                final String key = getKey(rg);
+                if(!collectors.containsKey(key)) {
+                    collectors.put(key, makeCollector(rg));
```
akiezun: use `putIfAbsent` instead of the `if`

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+        //A Map mapping the key for a specific record (as determined by getKey) to the appropriate collector
+        private final Map<String, UNIT_COLLECTOR> collectors;
+
+        //Given a SAMReadGroupRecord, return the key that identifies the collector for the corresponding SAMRecord
+        protected abstract String getKey(final SAMReadGroupRecord rg);
+
+        //Make a PerUnitMetricCollector for this given Distributor
+        protected abstract UNIT_COLLECTOR makeCollector(final SAMReadGroupRecord rg);
+
+        protected abstract UNIT_COLLECTOR makeUnknownCollector();
```
akiezun: add a comment about what this method does

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+
+        public Distributor(final List<SAMReadGroupRecord> rgRecs) {
+            collectors = new LinkedHashMap<>();
+            for(final SAMReadGroupRecord rg : rgRecs) {
+                final String key = getKey(rg);
```
akiezun: the getKey and makeCollector methods are called during object initialization, which may result in hard to detect bugs because methods typically assume that the object has been created and initialized

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+            }
+        }
+
+        /**
+         * Combine/merge the source distributor into this distributor by merging the collector maps.
+         * @param source source Distributor
+         * @param reMap map entry merge (reMap) function for combining distributor map entries. This
+         *              is usually the combineUnit method of MultiLevelCollectorReducible.
+         */
+        public void combine(Distributor source, BiFunction<UNIT_COLLECTOR, UNIT_COLLECTOR, UNIT_COLLECTOR> reMap) {
```
akiezun: why use `combine` in some places and `merge` in others?

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+            for(final UNIT_COLLECTOR collector : collectors.values()) {
+                collector.addMetricsToFile(file);
+            }
+        }
+    }
+
+    /** A dummy Distributor to handle the ALL_READS accumulation level.  No distribution is required
+     * since there should only ever be one PerUnitMetricCollector for ALL_READS.
+     */
+    private class AllReadsDistributor extends Distributor implements Serializable {
```
akiezun: these non static inner classes make me uncomfortable a bit. Can they be made static ?

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+            super(new ArrayList<>());
+            makeCollector(null);
+        }
+
+        @Override
+        protected String getKey(SAMReadGroupRecord rg) {
+            return null;
+        }
+
+        public void acceptRecord(final ARGTYPE args, final SAMReadGroupRecord rg) {
```
akiezun: `@Override`

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+        }
+
+        @Override
+        protected UNIT_COLLECTOR makeUnknownCollector() {
+            throw new UnsupportedOperationException("Should not happen");
+        }
+
+        @Override
+        public void finish() {
+            allReadCollector.finish();
```
akiezun: is it ok to just ignore the finish method in super class? if yes, add a comment to the super class that it's ok to override and add a comment here that superclass is ignored on purpose

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    /** A dummy Distributor to handle the ALL_READS accumulation level.  No distribution is required
+     * since there should only ever be one PerUnitMetricCollector for ALL_READS.
+     */
+    private class AllReadsDistributor extends Distributor implements Serializable {
+
+        private static final long serialVersionUID = 1L;
+
+        public AllReadsDistributor(final List<SAMReadGroupRecord> rgRecs) {
```
akiezun: the rgRegs are ignored, should you remove them from the constructor's signature?

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+            // NOTE: since we're passing an empty list to the super class constructor,
+            // no collector will actually be added to the collector map; instead this
+            // class assumes a hardcoded allReadsCollector in all of it's other methods
```
akiezun: `it's` -> `its`

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+        if(accumulationLevels.contains(MetricAccumulationLevel.READ_GROUP)) {
+            outputOrderedDistributors.add(new ReadGroupDistributor(samRgRecords));
+        }
+    }
+
+    /**
+     * Construct a argument of ARGTYPE using the given SAMRecord and ReferenceSequence then pass
+     * this value to all collectors that should include this record
+     */
+    public void acceptRecord(final SAMRecord record, final ReferenceSequence refSeq) {
```
akiezun: `@Override`

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
```diff
@@ -0,0 +1,428 @@
...
+     */
+    public void finish() {
+        for(final Distributor distributor : outputOrderedDistributors) {
+            distributor.finish();
+        }
+    }
+
+    /** Get the PerUnitMetricCollector that collects reads for all levels */
+    public UNIT_COLLECTOR getAllReadsCollector() {
+        return allReadCollector;
```
akiezun: should rename the field to `allReadsCollector` or rename this method to `getAllReadCollector` because they are out of synch now

src/main/java/org/broadinstitute/hellbender/metrics/QualityYieldMetrics.java
```diff
@@ -140,4 +140,12 @@ public QualityYieldMetrics finish() {
         finalized = true;
         return this;
     }
+
+    // Return a disambiguating name suffix to be used by the multiple collectors to
```
akiezun: make this into javadoc

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/AlignmentSummaryMetricsCollector.java
```diff
@@ -21,6 +21,8 @@
 import java.util.*;
 
 public final class AlignmentSummaryMetricsCollector extends SAMRecordAndReferenceMultiLevelCollector<AlignmentSummaryMetrics,  Long> {
+    private static final long serialVersionUID = 1L;
```
akiezun: should the class be `Serializable`? right now it has a field `serialVersionUID` but it's not Serializable so the field is useless

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/AlignmentSummaryMetricsCollector.java
```diff
@@ -124,6 +126,8 @@ private boolean isAdapterSequence(final byte[] read) {
     }
 
     private class GroupAlignmentSummaryMetricsPerUnitMetricCollector implements PerUnitMetricCollector<AlignmentSummaryMetrics, Long, SAMRecordAndReference> {
```
akiezun: can this class be static?

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/CollectAlignmentSummaryMetrics.java
```diff
@@ -48,6 +49,10 @@
     
     private static final Logger log = LogManager.getLogger();
 
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "File to write the output to.")
+    public File OUTPUT;
```
akiezun: fields have lowercase names

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/CollectBaseDistributionByCycle.java
```diff
@@ -26,6 +27,10 @@
 public final class CollectBaseDistributionByCycle extends SinglePassSamProgram {
     public static final String R_SCRIPT = "baseDistributionByCycle.R";
 
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "File to write the output to.")
+    public File OUTPUT;
```
akiezun: lowercase

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -40,6 +41,10 @@
     /** The location of the R script to do the plotting. */
     private static final String R_SCRIPT = "gcBias.R";
 
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "File to write the output to.")
+    public File OUTPUT;
```
akiezun: lowercase

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/MeanQualityByCycle.java
```diff
@@ -36,6 +37,10 @@
 public final class MeanQualityByCycle extends SinglePassSamProgram {
     public static final String R_SCRIPT = "meanQualityByCycle.R";
 
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "File to write the output to.")
+    public File OUTPUT;
```
akiezun: lowercase

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/QualityScoreDistribution.java
```diff
@@ -33,6 +34,10 @@
 public final class QualityScoreDistribution extends SinglePassSamProgram {
     public static final String R_SCRIPT = "qualityScoreDistribution.R";
 
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "File to write the output to.")
+    public File OUTPUT;
```
akiezun: lowercase

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -51,6 +52,10 @@
 public final class CollectSequencingArtifactMetrics extends SinglePassSamProgram {
     static final String USAGE = "Produces metrics to quantify single-base sequencing artifacts from a SAM/BAM file";
 
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "File to write the output to.")
+    public File OUTPUT;
```
akiezun: lowercase. 

btw, why was the output field moved from `SinglePassSamProgram` down to subclasses?

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/directed/CollectRnaSeqMetrics.java
```diff
@@ -37,6 +38,10 @@
     private static final String R_SCRIPT = "rnaSeqCoverage.R";
     private static final Logger LOG = LogManager.getLogger();
 
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "File to write the output to.")
+    public File OUTPUT;
```
akiezun: lowercase

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
```diff
@@ -93,68 +94,95 @@
...
+
+    /**
+     * Determine which collectors to run based on commandline args and programmatically
+     * specified custom collectors.
+     */
+    private List<SparkCollectorProvider> getCollectorsToRun() {
+        List<SparkCollectorProvider> collectorsToRun = new ArrayList<>();
+        if (externalCollectors == null) {
+            if (userCollectors.size() == 0) { // run them all
+                collectorsToRun = Arrays.asList(SparkCollectors.values());
```
akiezun: can you just `return Arrays.asList(SparkCollectors.values())`

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSparkTool.java
```diff
@@ -20,7 +20,7 @@
  *
  * - Declare an instance of a collector that implements the MetricsCollectorSpark interface
  * - Declare an instance of the input argument collection (of type T) for the collector
- * - Implement or override the getReadFilter, initialize, collectMetrics and finishCollection
+ * - Implement or override the getReadFilter, initialize, collectMetrics and saveMetrics
```
akiezun: all those should be javadoc links like this `{@link #method}`

src/test/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducibleUnitTest.java
```diff
@@ -0,0 +1,254 @@
...
+ * all of the basic functionality of multi-level distribution based on the MultiLevelCollector
+ * tests, with additional test code to verify the combine functionality used for combing/reducing
+ * multiple collectors into a single collector.
+ */
+public final class MultiLevelCollectorReducibleUnitTest {
+
+    public static File TESTFILE = new File("src/test/resources/org/broadinstitute/hellbender/metrics/test.sam");
+
+    public String noneOrStr(final String str) {
+        final String out;
```
akiezun: `return str == null ? "" : str;`

src/test/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducibleUnitTest.java
```diff
@@ -0,0 +1,254 @@
...
+        public TestArg(final SAMRecord samRecord, final ReferenceSequence refSeq) {
+            this.samRecord = samRecord;
+            this.refSeq    = refSeq;
+        }
+    }
+
+    /** We will just Tally up the number of times records were added to this metric and change FINISHED
+     * to true when FINISHED is called
+     */
+    class TotalNumberMetric extends MultiLevelMetrics {
```
akiezun: `static final class`?

src/test/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducibleUnitTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+        final RecordCountMultiLevelCollector collector1 = new RecordCountMultiLevelCollector(
+                accumulationLevels, in.getFileHeader().getReadGroups());
+        final RecordCountMultiLevelCollector collector2 = new RecordCountMultiLevelCollector(
+                accumulationLevels, in.getFileHeader().getReadGroups());
+
+        //distribute the reads across the two collectors
+        int count = 1;
+        for (final SAMRecord rec : in) {
+            if (count % 2 == 0) {
```
akiezun: how is this supposed to work? `count` is never incremented

akiezun:
this is a whopper. @lbergelson can you have a look at it too?

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
cmnbroad: Because the collectors need be instantiable in multiple contexts, so all of their args need to be in their own argument collection, and available in all contexts. If I leave it in SinglePassSam, it will be a duplicate arg in that context. As the other collectors are moved to this pattern, the individual ones that were moved down will go away and there will only be one again (in the MetricsArgumentCollection base class).

droazen:
Since this new framework is pretty independent of much of the rest of the codebase, I'd recommend we merge once @akiezun's comments are addressed and @lbergelson gives the thumb's up, assuming no one has major issues with the overall design, instead of letting the review drag on too long (especially as @cmnbroad is away next week!)

davidbernick:

akiezun:
@droazen that sounds good to me

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetrics.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: Good point - I moved the R file to a metrics resource folder, and no longer need the qualified name.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/InsertSizeMetricsCollectorArgs.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/MetricsArgumentCollection.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Only to emphasize that its the same as MultiLevelCollector, except its reducible. But someday MultiLevelCollector will be gone so I'll rename it.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Made private.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Made private.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Yeah, that is weird. Reworded and unified with the tests further down.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Done. Separated out initialization into a separate method.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: This one isn't an override in the Distributor hierarchy; its the original in the MultiLevelCollector hierarchy.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Yes, the all reads collector is a special case. However as part of the change to static inner classes, I changed the all reads distributor to use the same collector list the other distributors use, so this method overload is no longer necessary at all.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Comment is removed and no longer necessary as part of the change to static inner classes.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Made them all use allReads.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
cmnbroad: Done, and removed unnecessary warning suppression.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducibleUnitTest.java
cmnbroad: Fixed.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Yes, done, at the expense of making the code look much more complicated because all the type parameters have to be conferred on all of the inner classes, and the need for an explicit reference to the outer class instance to access the overloaded methods. Its the right thing to do though for serialization. The distributor hierarchy could probably be simplified more, but not as part of this pr.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: Yes. In fact since I separated out initialization from construction, they're removed from all of the constructors.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
cmnbroad: This value comes from an optional command line argument that may be specified by the user. Added a comment.

src/main/java/org/broadinstitute/hellbender/metrics/QualityYieldMetrics.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
cmnbroad: I think the only place merge is used is when referring to arguments for the actual merge method on Map. Having said that, some of the comments in other places are not inconsistent. Will unify.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Yes good catch, thats misleading. Fixed.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSpark.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducibleUnitTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/AlignmentSummaryMetricsCollector.java
cmnbroad: No, its not necessary until its ported to the new framework. Removed.

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/AlignmentSummaryMetricsCollector.java
cmnbroad: Not without making a bunch of other changes. All of these (other Picard) metrics classes will all have to be ported to hellbender style and the new metrics framework, but those will be separate PRs.

src/main/java/org/broadinstitute/hellbender/tools/picard/analysis/CollectAlignmentSummaryMetrics.java
cmnbroad: Agreed, but the only change here was to push the definition down from SinglePassSam into the individual classes as an interim step until these classes are all ported to the new framework. If we change it to be lower case, all of the references here and in all of the other classes it got pushed down to would also have to be changed. And this class also has 5 other fields all in upper case, and doesn't conform in other ways. I'd propose we wait and make those changes in the PRs where these tools and classes are made GATK compliant, rather than adding all of those changes to this PR.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectMultipleMetricsSpark.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducibleUnitTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSparkTool.java
cmnbroad: Done.

cmnbroad:
I addressed every comment except two: the suggested "putIfAbsent" change (see the comment there - I don't think we should do it), and I didn't change the case of the OUTPUT field that I moved down from SinglePassSAM to several Picard metrics classes (also see the note) since doing so would require changing all of the references in those classes, and they have many other fields that need similar changes. I'd propose that those changes be done in separate PRs, as those classes are ported to this framework.

cmnbroad:
Added a couple more javadoc updates. Not squashed, otherwise all comments have been addressed (except the two described above).

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -162,65 +151,63 @@ public void addMetricsToFile(final MetricsFile<InsertSizeMetrics, Integer> file)
...
+     * @param sourceCollector InsertSizeMetricsCollector to combine in
      * @return InsertSizeMetricsCollector representing the combination of the source collector with this collector
      */
     public PerUnitInsertSizeMetricsCollector combine(PerUnitInsertSizeMetricsCollector sourceCollector) {
-        String validationMessage = "Internal error combining collectors";
-        validateEquals(this.sample, sourceCollector.sample, validationMessage);
-        validateEquals(this.library, sourceCollector.library, validationMessage);
-        validateEquals(this.readGroup, sourceCollector.readGroup, validationMessage);
+        Utils.nonNull(sourceCollector);
+        Objects.equals(this.sample, sourceCollector.sample);
```
akiezun: hmm, this does not work. Objects.equals just returns false when the args are not equal but this should blow up. Need to wrap in an if statement

src/main/java/org/broadinstitute/hellbender/metrics/QualityYieldMetricsArgumentCollection.java
```diff
@@ -17,6 +17,6 @@
             fullName = StandardArgumentDefinitions.USE_ORIGINAL_QUALITIES_LONG_NAME,
             doc = "If available in the OQ tag, use the original quality scores " +
                     "as inputs instead of the quality scores in the QUAL field.")
-    public boolean useOriginalQualities = false;
+    public boolean useOriginalQualities = true;
```
akiezun: why this change to the default value?

src/main/java/org/broadinstitute/hellbender/metrics/MultiLevelCollectorReducible.java
akiezun: good point. thanks

src/main/java/org/broadinstitute/hellbender/metrics/MetricsArgumentCollection.java
```diff
@@ -1,19 +1,20 @@
...
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 
 import java.io.File;
+import java.io.Serializable;
 
 /**
  * Base class for defining a set of metrics collector arguments. The members should be
  * instantiable as command line arguments.
  */
-public class MetricsArgumentCollection {
```
lbergelson: Should this be an abstract class?  Will anyone want to instantiate one of the directly?

akiezun:
I found 2 issues in the revised code: Object.equals should be wrapped in an if-notEquals-then-blowup and an unexplained change in default. 

Other than those two, 👍 from me

lbergelson:
I didn't do a line by line review since it seemed like Adam did a very thorough job there.  The overall structure looks good to me though.  It seems like metrics collection would be a good candidate for being made a plugin system, as in the ReadFilter branch.  That should definitely be a different pass on this though if necessary.

akiezun:
who's taking over the branch?

droazen:
@akiezun I'll take care of addressing your last few comments on this branch and merging it in.

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
droazen: Fixed

src/main/java/org/broadinstitute/hellbender/metrics/QualityYieldMetricsArgumentCollection.java
droazen: Reverted, we'll see if tests pass

droazen: Tests fail -- taking a look at the failures now

droazen: Turns out that in picard this defaults to true, which is likely why @cmnbroad changed it. Setting it back to true.

droazen:
Tests pass -- merging

PR #1963, opened 2016-07-01
Add support for block gzipped files with a .bgz suffix (as well as

tomwhite:
a .gz suffix), see https://github.com/HadoopGenomics/Hadoop-BAM/pull/106.

Also remove spurious debug.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSource.java
@@ -13,6 +13,7 @@ import org.broadinstitute.hellbender.utils.variant.VariantContextVariantAdapter;
 import org.seqdoop.hadoop_bam.VCFInputFormat;
 import org.seqdoop.hadoop_bam.VariantContextWritable;
 import org.seqdoop.hadoop_bam.util.BGZFCodec;
+import org.seqdoop.hadoop_bam.util.BGZFEnhancedGzipCodec;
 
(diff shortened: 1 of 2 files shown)


tomwhite:
@akiezun can you review this please? Small follow on for #1936

akiezun:
thanks @tomwhite. how do I test it?

tomwhite:
@akiezun if you run `CountVariantsSparkIntegrationTest` without the fix you'll see a warning about using gzip. With the fix (this PR) the warning goes away.

akiezun:
thanks. It works. I'd prefer to test for it more directly. Can you examine the stderr in the tests, like this
(change to CountVariantsSparkIntegrationTest .testCountVariantsWithIntervals)
```
        final ByteArrayOutputStream baosErr = new ByteArrayOutputStream();
        final PrintStream err = System.err;
        try {
            System.setErr(new PrintStream(baosErr));

            this.runCommandLine(args.getArgsArray());

            final String readIn = FileUtils.readFileToString(outputTxt.getAbsoluteFile());
            Assert.assertEquals((int)Integer.valueOf(readIn), expected);
            String errString = baosErr.toString();
            Assert.assertFalse(errString.contains("Warning: using GzipCodec, which is not splittable,"),errString);
        } finally {
            System.setErr(err); //put this back in
        }
```

back to @tomwhite - one small fix and then it's good to go

tomwhite:
I added the test you suggested, @akiezun, and verified that it fails against head.

akiezun:
lgtm

Issue #1903, opened 2016-06-09
Load inteldeflater from intel-gkl, hook up to new htsjdk interface


droazen:

akiezun:
blocked by htsjdk release

akiezun:
@gspowley is already on it - see pull #1935

akiezun:
@gspowley do you have a timeline on this? We need it in alpha-2 (tomorrow).

akiezun:
I'll grab this - I have a solution in branch ak_use_intelDeflaterFactory

PR #1964, opened 2016-07-01
use IntelDeflaterFactory from the GKL

akiezun:
fixes https://github.com/broadinstitute/gatk/issues/1903

also updated the system properties in build.gradle and gatk-launch

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -1,13 +1,14 @@
 package org.broadinstitute.hellbender.cmdline;
 
+import com.intel.gkl.compression.IntelDeflaterFactory;
 import htsjdk.samtools.Defaults;
 import htsjdk.samtools.metrics.Header;
 import htsjdk.samtools.metrics.MetricBase;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.metrics.StringHeader;
+import htsjdk.samtools.util.BlockCompressedOutputStream;
 import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.Log;
-import htsjdk.samtools.util.zip.DeflaterFactory;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.hellbender.utils.LoggingUtils;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -128,6 +129,13 @@ public Object instanceMainPostParseArgs() {
             System.setProperty("java.io.tmpdir", f.getAbsolutePath()); // in loop so that last one takes effect
         }
 
+        //Set defaults (note: setting them here means they are not controllable by the user)
+        BlockCompressedOutputStream.setDefaultDeflaterFactory(new IntelDeflaterFactory());
```
droazen: Add a command-line arg to `CommandLineProgram` that allows the user to opt-out of using the `IntelDeflater`

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -128,6 +129,13 @@ public Object instanceMainPostParseArgs() {
...
+        //Note: IntelDeflaterFactory (incorrectly) only loads the library and sets the usingIntelDeflater if some deflaters are actually created
+        //That means that the reported Deflater will be JdkDeflater even when IntelDeflater will actually be used.
```
droazen: File a bug against htsjdk to fix this

akiezun: i filed a gkl bug report https://github.com/Intel-HLS/GKL/issues/8

akiezun:
i'll grab it

akiezun:
note: This PR does not reinstate the IntelDeflater for Spark. It's now a bit tricky because you need to call `BlockCompressedOutputStream.setDefaultDeflaterFactory(new IntelDeflaterFactory());` on every executor or on every writer factory. Maybe a solution would be to use `IntelDeflaterFactory` directly in Hadoop-BAM. @droazen wdyt?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
akiezun: done

gspowley:
GKL 0.1.1 release in Maven Central resolves https://github.com/Intel-HLS/GKL/issues/8 (not the Spark issue)

akiezun:
thanks @gspowley - using 0.1.1 now and the issue is fixed

Issue #1725, opened 2016-04-18
print tool help by default - save user time


akiezun:
help should be printed by default not only after asking with `--help`

That is, typing this
```
./gatk-launch FilterVcf
```
should be exactly the same as
```
./gatk-launch FilterVcf --help
```

droazen:
Seems like a good idea -- @lbergelson do you agree?

lbergelson:
Sure.  Seems fine.

vdauwera:
I'm a little confused on this one. If the two are the same, what's the point of having `--help`?  What does "help" consist of? Basic usage or additional resources and detailed tool information like list of arguments?

akiezun:
consists of nothing. it's just there for convention

droazen:
Yeah, there are two kinds of people in the world: those who type "--help" first thing when running new software, and those who type nothing expecting help. We should help both kinds of people :)

lbergelson:
What happens if someone wants to write a tool that takes no arguments?

droazen:
We've never had a no-arg tool before, and I'm skeptical that we ever will...

vdauwera:
Ah, fair enough. I also can't imagine a tool with zero inputs.

akiezun:
@lbergelson i can take this one if you want. let me know

akiezun:
Requirements (consistent with picard)
`./gatk-launch CountReads` <- prints error message, usage help, exit code 1
`./gatk-launch CountReads --help` <- no error message, usage help, exit code 1
`./gatk-launch CountReads fred` <- prints error message, usage help, exit code 1
`./gatk-launch CountReads -I src/test/resources/large/expected.NA12878.RNAseq.splitNcigarReads.bam` <- exit code 0

lbergelson:
I think something with --help should exit code 0...

akiezun:
what about no args and `--version`?

akiezun:
just looked as samtools - no args gives 1, --help gives 0, and --version gives 0

lbergelson:
That sounds right to me.  Anything that's explicitly asking for information and then exiting should be 0.  No args is usually an error state.

akiezun:
I agree.

akiezun:
New requirements then:
```
./gatk-launch CountReads <- prints error message, usage help, exit code 1
./gatk-launch CountReads --help <- no error message, usage help, exit code 0
./gatk-launch CountReads fred <- prints error message, usage help, exit code 1
[...]

PR #1970, opened 2016-07-01
display user help if no arg is given

akiezun:
fixes https://github.com/broadinstitute/gatk/issues/1725

@lbergelson please review


akiezun:
needs test update

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -193,7 +194,18 @@ public Object instanceMain(final String[] argv) {
     protected boolean parseArgs(final String[] argv) {
 
         commandLineParser = new CommandLineParser(this);
-        final boolean ret = commandLineParser.parseArguments(System.err, argv);
+        final boolean ret;
+        try{
+            ret = commandLineParser.parseArguments(System.err, argv);
+        } catch (final UserException.CommandLineException e){
+            //The CommandLineException is treated specially - we display help and no blow up
```
lbergelson: I think we should print the help first and then the error message.  Otherwise finding the actually error requires screens worth of scrollback.  I'm not a fan of printing help on every command line error to begin with since I usually just made a typo and don't want a huge dump of information to clobber my screen, but if that's the behavior we've decided on I think should at least make the error the most prominent part.

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -78,14 +84,19 @@ public Object instanceMain(final String[] args) {
...
+     * The entry point to GATK from commandline.
+     * Note: this is the only method that is allowed to call System.exit (because gatk tools may be run from test harness etc)
      */
     public static void main(final String[] args) {
         try {
             final Object result = new Main().instanceMain(args, getPackageList(), "");
             if (result != null) {
               System.out.println("Tool returned:\n" + result);
             }
+        } catch (final UserException.CommandLineException e){
```
lbergelson: I think that stack traces could be helpful for developers for command line exceptions.  I see at least a few places in the code where they're thrown outside of the command line parser, and it would be tricky to track them down without the stack trace.  

It's also a bit weird to have the error be formatted differently (i.e. without the *****s).

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -36,7 +36,13 @@
      * exit value when an unrecoverable {@link UserException} occurs
      */
     private static final int USER_EXCEPTION_EXIT_VALUE = 2;
-    
+
+    /**
+     * exit value when an issue with the commandline is detected, ie {@link UserException.CommandLineException}.
+     * This is the same value as picard uses.
+     */
+    private static final int COMMANDLINE_EXCEPTION_EXIT_VALUE = 1;
```
lbergelson: Do we want this to be the same as for all other exceptions?  I would bump the `ANY_OTHER_EXCEPTION_VALUE` to be 3 if this going to claim 1.

lbergelson:
@akiezun A few comments.  Some of it is just my personal taste, but I think we should at least reverse the ordering of the help and the error message so it's not hidden.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/Main.java
akiezun: good point. done

akiezun:
back to @lbergelson

src/main/java/org/broadinstitute/hellbender/Main.java
akiezun: fixed

lbergelson:
👍

PR #1971, opened 2016-07-01
deleting scripts for building native code

lbergelson:
we don't need these now that intel gkl has replaced our native code build


droazen:
Have we confirmed that the gkl we're pulling down from maven has both so and dylib?

droazen:
Merge once you've confirmed that the new gkl has both so and dylib

lbergelson:
screen shot 2016-07-01 at 3 50 27 pm

PR #1972, opened 2016-07-01
Part 1 of HC memory fixes: add downsampling to AssemblyRegion traversal

### LINKED ISSUE #1642: HaplotypeCaller: hook up downsampling
Let's use a different downsampling strategy than GATK3 for the `HaplotypeCaller` -- perhaps a `ReservoirDownsampler` whose size is proportional to the size of each region.

Need to make sure that whatever strategy we use allows us to deal with high-coverage regions without introducing calling artifacts / missed calls.

### LINKED ISSUE #1911: Make ReadCoordinateComparator match bam file order
The fact that `ReadCoordinateComparator` does not exactly match the ordering of htsjdk's `SAMRecordCoordinateComparator` has been the cause of a few bugs. It sorts all unmapped reads after mapped reads, whereas `SAMRecordCoordinateComparator` sorts unmapped reads that are assigned the positions of their mapped mates with their mapped mates.

The issue is that the `GATKRead` interface does not allow unmapped reads to have a position. Ie., even if an unmapped `SAMRecord` is assigned the position of its mapped mate, calling `getContig()`/`getStart()` on the unmapped read via the `GATKRead` interface will return `null`/`0`. This was done mainly for consistency reasons and to simplify client code.

Perhaps we could add `getAssignedContig()`, `getAssignedStart()`, etc. methods to GATKRead to expose the positions that unmapped reads with mapped mates get assigned for sorting purposes, and use these in `ReadCoordinateComparator`. This should allow us to match `SAMRecordCoordinateComparator` exactly, and then `ReadCoordinateComparator` could be used even when sorting for the purpose of writing a bam.

droazen:
-Hooked up a new PositionalDownsampler to AssemblyRegionWalker controlled
 via an argument --maxReadsPerAlignmentStart (tool sets the default value).

-As a side effect, exposed the "assigned" position of a read in the
 GATKRead interface. This lets us query the nominal positions of unmapped
 reads that have been assigned a position for sorting purposes.

-As a second side effect, modified ReadCoordinateComparator to sort
 unmapped reads with positions in the correct bam/sam file order.

Resolves #1642
Resolves #1911


droazen:
@akiezun please review. I had to split my memory fixes into two PRs, as the other part is not quite ready yet. This one is just downsampling plus related changes.

droazen:
(@lbergelson also feel free to review, if you're around and have the time)

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -199,6 +208,7 @@ public final void traverse() {
             // Since reads in each shard are lazily fetched, we need to pass the filter to the window
             // instead of filtering the reads directly here
             readShard.setReadFilter(countedFilter);
+            readShard.setDownsampler(maxReadsPerAlignmentStart > 0 ? new PositionalDownsampler(maxReadsPerAlignmentStart, getHeaderForReads()) : null);
```
akiezun: would it be cleaner to create a `no-op` downsampler for the `0` case, rather than have a `null` here and always having to deal with that?

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
```diff
@@ -147,9 +160,17 @@ public boolean containsStartPosition( final Locatable loc ) {
      */
     @Override
```
akiezun: enhance the documentation here to state that the filters will be applied before the downsamping

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
droazen: There already is a no-op downsampler (`PassThroughDownsampler`), but I decided not to use it since it adds a bit of overhead (one `List` creation per start position).

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * PositionalDownsampler: Downsample each stack of reads at each alignment start to a size <= a target coverage
+ * using a {@link ReservoirDownsampler}. Stores only O(target coverage) reads in memory at any given time,
+ * provided the client regularly calls {@link #consumeFinalizedItems}.
+ *
+ * Unmapped reads with assigned positions are subject to downsampling in the same way as mapped reads,
+ * but unmapped reads without assigned positions are not subject to downsampling.
+ */
+public class PositionalDownsampler extends ReadsDownsampler {
```
akiezun: make final

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
akiezun: can we have a no-op one that does nothing?

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: The javadoc for `setDownsampler` already states that `Downsampling will be performed after any requested read filtering.` -- do you want this comment copied anywhere else?

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -0,0 +1,146 @@
...
+
+        this.reservoir = new ReservoirDownsampler(targetCoverage);
+        this.finalizedReads = new ArrayList<>();
+        this.header = header;
+        clearItems();
+        resetStats();
+    }
+
+    @Override
+    public void submit( final GATKRead newRead ) {
```
akiezun: check for null input

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -0,0 +1,146 @@
...
+ * Unmapped reads with assigned positions are subject to downsampling in the same way as mapped reads,
+ * but unmapped reads without assigned positions are not subject to downsampling.
+ */
+public class PositionalDownsampler extends ReadsDownsampler {
+
+    private final ReservoirDownsampler reservoir;
+
+    private final SAMFileHeader header;
+
+    private GATKRead lastRead;
```
akiezun: `previousRead` ?

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
droazen: There would still be overhead, since you'd have to wrap the no-op downsampler in a `ReadsDownsamplingIterator`. It seems silly to do this, and is not really "cleaner" or more readable.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+        }
+    }
+
+    private void finalizeReservoir() {
+        finalizedReads.addAll(reservoir.consumeFinalizedItems());
+        reservoir.resetStats();
+    }
+
+    private boolean readHasNoPosition( final GATKRead read ) {
```
akiezun: sounds like a `ReadUtils` kind of method. Move there, document and add tests

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -0,0 +1,146 @@
...
+        // Check actual assigned positions rather than unmapped status, so that unmapped reads with
+        // assigned positions will be processed correctly.
+        return read.getAssignedContig() == null ||
+               read.getAssignedContig().equals(ReadConstants.UNSET_CONTIG) ||
+               read.getAssignedStart() == ReadConstants.UNSET_POSITION;
+    }
+
+    @Override
+    public boolean hasFinalizedItems() {
+        return finalizedReads.size() > 0;
```
akiezun: `! finalizedReads.isEmpty()`

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
akiezun: according to my reading of the doc `read.getAssignedContig()` cannot be null. Can it? (if yes, can you clarify the doc?)

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
```diff
@@ -0,0 +1,91 @@
...
+    private Collection<GATKRead> downsampledReadsCache;
+    private GATKRead nextRead = null;
+    private Iterator<GATKRead> downsampledReadsCacheIterator = null;
+
+    /**
+     * @param iter wrapped iterator from which this iterator will pull reads
+     * @param downsampler downsampler through which the reads will be fed
+     */
+    public ReadsDownsamplingIterator( Iterator<GATKRead> iter, ReadsDownsampler downsampler ) {
+        Utils.nonNull(iter);
```
akiezun: if you have two `nonNull` checks, they need msg strings, otherwise it's hard to debug

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        GATKRead toReturn = nextRead;
+        advanceToNextRead();
+
+        return toReturn;
+    }
+
+    private void advanceToNextRead() {
+        if ( ! readyToReleaseReads() && ! fillDownsampledReadsCache() ) {
```
akiezun: simpler when you flip to 
```
if (readyToReleaseReads() || fillDownsampledReadsCache()) {
  nextRead = downsampledReadsCacheIterator.next();
} else {
  nextRead = null;
}
```

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
```diff
@@ -0,0 +1,91 @@
...
+            nextRead = null;
+        }
+        else {
+            nextRead = downsampledReadsCacheIterator.next();
+        }
+    }
+
+    private boolean readyToReleaseReads() {
```
akiezun: more conventional as `isReadyToReleaseReads`

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
```diff
@@ -0,0 +1,91 @@
...
+
+        if ( ! nestedReadIterator.hasNext() ) {
+            downsampler.signalEndOfInput();
+        }
+
+        // use returned collection directly rather than make a copy, for speed
+        downsampledReadsCache = downsampler.consumeFinalizedItems();
+        downsampledReadsCacheIterator = downsampledReadsCache.iterator();
+
+        return downsampledReadsCacheIterator.hasNext();
```
akiezun: why no null check here? If it's not needed, add a note. Otherwise, add a check

droazen: Guaranteed to be non-null

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
```diff
@@ -0,0 +1,91 @@
...
+        return nextRead != null;
+    }
+
+    @Override
+    public GATKRead next() {
+        if ( nextRead == null ) {
+            throw new NoSuchElementException("next() called when there are no more items");
+        }
+
+        GATKRead toReturn = nextRead;
```
akiezun: final temp

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
```diff
@@ -73,6 +73,21 @@ default boolean isEmpty() {
     void setPosition( final Locatable locatable );
 
     /**
+     * @return The actual contig assigned to the read, regardless of unmapped status. Unlike {@link #getContig}, will
```
akiezun: too many negatives here "Unlike .. will return non-null". It is true to say that this method never returns null?

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
```diff
@@ -73,6 +73,21 @@ default boolean isEmpty() {
...
+     *         return non-null for unmapped reads that are assigned a nominal position (eg., the position of their mates).
+     *         Useful for sorting reads in standard BAM/SAM file order -- for other uses, clients should use {@link #getContig}
+     */
+    String getAssignedContig();
+
+    /**
+     * @return The actual start position assigned to the read, regardless of unmapped status. Unlike {@link #getStart}, will
```
akiezun: it would perhaps be useful for `GATKRead` to explicitly list methods from `Locatable` and explain how the semantics are interpreted, eg the fact that `getStart` for `unmapped reads that are assigned a nominal position` returns 0. Otherwise, it's left sort of unspecified and that's bad.

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
```diff
@@ -73,6 +73,21 @@ default boolean isEmpty() {
...
+     *         return a non-zero position for unmapped reads that are assigned a nominal position (eg., the position of their
+     *         mates). Useful for sorting reads in standard BAM/SAM file order  -- for other uses, clients should use
+     *         {@link #getStart}
+     */
+    int getAssignedStart();
```
akiezun: this method sometimes returns `ReadConstants.UNSET_POSITION` - can you clarify in the doc when that happens?

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -0,0 +1,146 @@
...
+    private void finalizeReservoir() {
+        finalizedReads.addAll(reservoir.consumeFinalizedItems());
+        reservoir.resetStats();
+    }
+
+    private boolean readHasNoPosition( final GATKRead read ) {
+        // Check actual assigned positions rather than unmapped status, so that unmapped reads with
+        // assigned positions will be processed correctly.
+        return read.getAssignedContig() == null ||
+               read.getAssignedContig().equals(ReadConstants.UNSET_CONTIG) ||
```
akiezun: reading the GATKRead code I don't see any mention (in code or doc) of `ReadConstants.UNSET_CONTIG` - the code returns `null`. When is this line ` read.getAssignedContig().equals(ReadConstants.UNSET_CONTIG)` triggered?

droazen: When `SAMRecords` are created, they may be assigned this value by htsjdk. In `GATKRead` we don't expose this value in `getContig()`, but we do now expose it in `getAssignedContig()`. I'll try to clarify in the docs.

src/main/java/org/broadinstitute/hellbender/utils/read/ReadCoordinateComparator.java
```diff
@@ -63,22 +61,22 @@ public int compare( GATKRead first, GATKRead second ) {
...
 
-        final int firstRefIndex = ReadUtils.getReferenceIndex(first, header);
-        final int secondRefIndex = ReadUtils.getReferenceIndex(second, header);
-
-        final int refIndexComparison = firstRefIndex - secondRefIndex;
-        if ( refIndexComparison != 0 ) {
-            return refIndexComparison;
+        final int refIndexDifference = firstRefIndex - secondRefIndex;
+        if ( refIndexDifference != 0 ) {
+            return refIndexDifference;
```
akiezun: what's the value in returning this difference in indices rather than just -1 or 1 (as this comparator does everywhere else)?

src/test/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsamplerUnitTest.java
```diff
@@ -0,0 +1,207 @@
...
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import org.broadinstitute.hellbender.utils.read.ReadCoordinateComparator;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.*;
+
+public class PositionalDownsamplerUnitTest extends BaseTest {
```
akiezun: three methods in the downsampler remain untested - add tests for them:
```
size()
requiresCoordinateSortOrder()
signalNoMoreReadsBefore()
```

src/test/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsamplerUnitTest.java
```diff
@@ -0,0 +1,207 @@
...
+
+    private final SAMFileHeader header = ArtificialReadUtils.createArtificialSamHeader();
+
+    @DataProvider(name = "PositionalDownsamplerTestData")
+    public Object[][] positionalDownsamplerTestData() {
+        return new Object[][] {
+                // reads, downsampling target coverage, expected size and number of postion-based stacks after downsampling, and whether the stream ends with a stack of pure unmapped reads
```
akiezun: `postion` -> `position`

src/test/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsamplerUnitTest.java
```diff
@@ -0,0 +1,207 @@
...
+        }
+        else {
+            Assert.assertNull(downsampler.peekFinalized());
+        }
+
+        Assert.assertFalse(downsampler.hasPendingItems());
+        Assert.assertNull(downsampler.peekPending());
+
+        List<GATKRead> downsampledReads = downsampler.consumeFinalizedItems();
+        Assert.assertFalse(downsampler.hasFinalizedItems() || downsampler.hasPendingItems());
```
akiezun: this line `assertFalse(downsampler.hasFinalizedItems() || downsampler.hasPendingItems());` would be simpler to read as two lines: `assertFalse(downsampler.hasFinalizedItems()) ` and `assertFalse( downsampler.hasPendingItems())`

src/test/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsamplerUnitTest.java
```diff
@@ -0,0 +1,207 @@
...
+        Assert.assertTrue(downsampler.peekFinalized() == null && downsampler.peekPending() == null);
```
akiezun: same here - simpler and easier to debug as two lines `assertNull(downsampler.peekFinalized())` and `assertNull(downsampler.peekPending())`

src/test/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsamplerUnitTest.java
```diff
@@ -0,0 +1,207 @@
...
+            Assert.assertEquals(numReadsActuallyEliminated, numReadsReportedEliminated);
+        }
+
+        downsampler.resetStats();
+        Assert.assertEquals(downsampler.getNumberOfDiscardedItems(), 0);
+
+        // Now test with a PositionalDownsampler wrapped in an iterator, and make sure we get the same results.
+        // It's crucial to reset the random number generator again in order to match the selections made by the
+        // first downsampling pass.
+        Utils.resetRandomGenerator();
```
akiezun: Does this comment suggest that splitting this part (testign the iterator) would be better in a new method?

src/test/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIteratorUnitTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public class ReadsDownsamplingIteratorUnitTest extends BaseTest {
```
akiezun: while you're here, add tests for error conditions: calling `next()` when there are no items and calling `remove`

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
```diff
@@ -0,0 +1,91 @@
...
+public final class ReadsDownsamplingIterator implements Iterator<GATKRead>, Iterable<GATKRead> {
+
+    private final Iterator<GATKRead> nestedReadIterator;
+    private final ReadsDownsampler downsampler;
+    private Collection<GATKRead> downsampledReadsCache;
+    private GATKRead nextRead = null;
+    private Iterator<GATKRead> downsampledReadsCacheIterator = null;
+
+    /**
+     * @param iter wrapped iterator from which this iterator will pull reads
```
akiezun: This constructor needs better documentation - what is the iterator that the clients are supposed to give where? The class doc says nothing about it and this doc here also says nothing

droazen: Not sure what you mean here -- this parameter is simply an iterator of reads that are to be downsampled, as the docs state. What more do you want me to say?

akiezun: A simple description of what this iterator does (in the class doc) would be sufficient here. Something that says that this is an decorator of iterators and takes a iterator and downsamples reads using the given downsampler.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Iterator wrapper around our generic {@link ReadsDownsampler)} interface. Converts the push-style
+ * downsampler interface to a pull model.
+ */
+public final class ReadsDownsamplingIterator implements Iterator<GATKRead>, Iterable<GATKRead> {
+
+    private final Iterator<GATKRead> nestedReadIterator;
+    private final ReadsDownsampler downsampler;
+    private Collection<GATKRead> downsampledReadsCache;
```
akiezun: why is `downsampledReadsCache` an instance field? It's only used within 1 method

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -0,0 +1,146 @@
...
+        // Pass-through reads that have no assigned position, to avoid downsampling all unmapped reads
+        // to the targetCoverage. Unmapped reads that do have an assigned position *will* be subject to
+        // downsampling, however.
+        if ( readHasNoPosition(newRead) ) {
+            finalizedReads.add(newRead);
+        }
+        else {
+            final int reservoirPreviouslyDiscardedItems = reservoir.getNumberOfDiscardedItems();
+            reservoir.submit(newRead);
+            numDiscardedItems += reservoir.getNumberOfDiscardedItems() - reservoirPreviouslyDiscardedItems;
```
akiezun: why do we have this protected field (`numDiscardedItems`) ? It's very unusual in the codebase to have protected fields. Can we encapsulate it for higher clarity and lower reader surprise?

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -0,0 +1,146 @@
...
+    @Override
+    public List<GATKRead> consumeFinalizedItems() {
+        final List<GATKRead> toReturn = finalizedReads;
+        finalizedReads = new ArrayList<>();
+        return toReturn;
+    }
+
+    @Override
+    public boolean hasPendingItems() {
+        return reservoir.hasFinalizedItems();
```
akiezun: can you clarify something here: if a read is finalized in the reservoir, can it become 'pending'? If not, why is this `hasPendingItems` method calling `reservoir.hasFinalizedItems()` ?

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public GATKRead peekFinalized() {
+        return finalizedReads.isEmpty() ? null : finalizedReads.get(0);
+    }
+
+    @Override
+    public GATKRead peekPending() {
+        return reservoir.peekFinalized();
```
akiezun: same question about reads that are finalized in the reservoir - can they become 'pending' in this downsampler?

akiezun:
done with my review - a few small edits and some questions about the downsampler and how it interacts with the reservoir downsampler.

akiezun:
lgtm otherwise 👍

src/main/java/org/broadinstitute/hellbender/engine/ReadShard.java
droazen: Done -- copied the clarification on order of operations to all affected methods.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: I'd rather not, since this gets called once per read, and since the engine is smart enough to not pass around null reads :)

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
akiezun: add a comment then

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: Extracted a public method to `ReadUtils`, added unit tests.

`getAssignedContig()` can return null for certain `GATKRead` implementations (eg., Google Genomics reads). Added a clarification to the docs for that method.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
droazen: Added

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
droazen: Agreed -- done.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
droazen: Eh, think the current name sounds better :)

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
droazen: Improved the wording, and clarified that the return value is dependent on the underlying read implementation.

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
droazen: Agreed, but sounds like a separate ticket to me -- opened https://github.com/broadinstitute/gatk/issues/1976

droazen: For now, added a class-level note about the difference between `getContig()`/`getStart()` and `getAssignedContig()`/`getAssignedStart()`

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/read/ReadCoordinateComparator.java
droazen: Code is a bit simpler this way, and also closer to `SAMRecordCoordinateComparator`, whose `fileOrderCompare()` method we are trying to imitate.

src/test/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsamplerUnitTest.java
droazen: Tests added

src/test/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsamplerUnitTest.java
droazen: Fixed

src/test/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsamplerUnitTest.java
droazen: Agreed -- split this up.

src/test/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsamplerUnitTest.java
droazen: Done

src/test/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsamplerUnitTest.java
droazen: We want to assert that we get the exact same reads from the iterator as from using the downsampler directly, so we can't split it into a separate test. I think making it a separate non-test method would increase confusion.

src/test/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIteratorUnitTest.java
droazen: Tests added

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
droazen: Alright, done

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: The finalized items in the `ReservoirDownsampler` are pending items from the perspective of the enclosing `PositionalDownsampler`, since the `PositionalDownsampler` doesn't know how many more reads at the current start position will be submitted to reservoir before we consume its finalized items.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: Same answer -- the finalized items in the reservoir as pending items from the perspective of the enclosing `PositionalDownsampler`

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
droazen: Paranoia about reference counts / premature garbage collection, mostly :)

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: Since that would change all the other downsamplers, I'd prefer to do it in a separate branch if you feel strongly about it. I think it's fine as-is, though.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: Added a comment at the interface level, adjusted implementations appropriately.

droazen:
@akiezun All comments addressed, back to you for a final look + green light to merge.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
akiezun: just delete the field. it should be a temp

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public List<GATKRead> consumeFinalizedItems() {
+        final List<GATKRead> toReturn = finalizedReads;
+        finalizedReads = new ArrayList<>();
+        return toReturn;
+    }
+
+    @Override
+    public boolean hasPendingItems() {
+        // The finalized items in the ReservoirDownsampler are pending items from the perspective of the
```
akiezun: something i still don't understand fully - the definition of 'pending' is 'read that may be retained after downsampling' and definition of 'finalized' is 'definitely will be retained after downsampling', right? In my reading of the `PositionalDownsampler ` code, all reads that are 'finalized' in the reservoir will survive the positional downsampling too so they are also 'finalized' and not 'pending' (ie we know they will survive).

src/main/java/org/broadinstitute/hellbender/utils/downsampling/Downsampler.java
```diff
@@ -20,6 +20,8 @@
      * immediately whether the item survives the downsampling process, while others will need to see
      * more items before making that determination.
      *
+     * Does not check its argument for null -- it is the client's responsibility not to submit null items.
```
akiezun: removing the check sets a bad precedent i think. Until we show that this check has a measurable performance impact, i think removing checks is not safe. What's stopping us from ditching these checks from everywhere? In profiling I've never seen those checks account for more than a very small fraction of runtime (< 1%)

akiezun:
back to @droazen for final As to my Qs. 👍 when that's done

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
akiezun: whatever. I'd change them - it's bogus as it is now. Separate branch is fine

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReadsDownsamplingIterator.java
droazen: Deleted

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: Alright, fine -- converted to private field with a protected setter

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
droazen: The `ReservoirDownsampler` considers reads that have been selected for inclusion in the reservoir at any given time to be "finalized", since they have survived the *reservoir* downsampling process. The enclosing `PositionalDownsampler` considers these same reads to be "pending", since they have not yet survived the *positional* downsampling process.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/Downsampler.java
droazen: I'll restore these checks.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
akiezun: is it the case though that if a read survived the reservoir downsampling process then it will survive the positional downsampling process?

droazen: No, because it may get eliminated before we move to the next position.

PR #1973, opened 2016-07-04
Command line parser plugins and ReadFilter command line arguments.

cmnbroad:


src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -71,6 +73,14 @@
     public static final String COMMENT = "#";
     public static final String POSITIONAL_ARGUMENTS_NAME = "Positional Argument";
 
+    // Map from (full class) name of each GATKCommandLinePluginDescriptor requested and
+    // found to the actual descriptor instance
+    Map<String, GATKCommandLinePluginDescriptor<?>> pluginDescriptors = new HashMap<>();
```
droazen: Can this be private rather than package-protected (since it has a public accessor)?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -113,7 +123,7 @@ private boolean inArgumentMap(ArgumentDefinition arg){
     private Object positionalArgumentsParent;
 
     // List of all the data members with @Argument annotation
-    private final List<ArgumentDefinition> argumentDefinitions = new ArrayList<>();
+    List<ArgumentDefinition> argumentDefinitions = new ArrayList<>();
```
droazen: Here too -- why does this need to be package visible?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -138,16 +148,26 @@ private String getUsagePreamble() {
         return usagePreamble;
     }
 
-
     public CommandLineParser(final Object callerArguments) {
+        this(callerArguments, null);
+    }
+
+    public CommandLineParser(
```
droazen: Add javadoc for public constructors

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
```diff
@@ -98,15 +105,50 @@ private SimpleInterval getReadInterval(final GATKRead read) {
...
+     *
+     * Multiple filters can be composed by using {@link org.broadinstitute.hellbender.engine.filters.ReadFilter}
+     * composition methods.
      */
     public CountingReadFilter makeReadFilter(){
-          return new CountingReadFilter("Wellformed", new WellformedReadFilter(getHeaderForReads()));
+        // This filter gets returned if all others are disabled.
+        CountingReadFilter defaultFilter = new CountingReadFilter(
+                ReadFilterLibrary.ALLOW_ALL_READS.getClass().getSimpleName(),
+                ReadFilterLibrary.ALLOW_ALL_READS);
```
droazen: Can the `CountingReadFilter` get the class name internally, and simplify this to a 1-arg constructor?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKCommandLinePluginDescriptor.java
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Base class to be used as command line plugin. Subclasses of this class will be
+     * command line accessible.
+     */
+    public abstract Class<?> getPluginClass();
+
+    /**
+     * Package names from which to load the command line plugin classes
+      */
+    public abstract String getPackageName();
```
droazen: Can we make this return a `List<String>`, to allow future support for multiple packages per plugin type?

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReadInputArgumentCollection.java
```diff
@@ -26,6 +26,9 @@
             optional=true)
     public ValidationStringency readValidationStringency = ReadConstants.DEFAULT_READ_VALIDATION_STRINGENCY;
 
+    @Argument(fullName = "disableAllReadFilters", shortName = "f", doc = "Disable all read filters", common = false, optional = true)
+    public boolean disableAllReadFilters = false;
```
droazen: Shouldn't this argument live in the descriptor, along with the other arguments that control selection/exclusion of read filters?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,199 @@
...
+    @Argument(fullName = StandardArgumentDefinitions.READ_FILTER_LONG_NAME,
+            shortName = StandardArgumentDefinitions.READ_FILTER_SHORT_NAME,
+            doc="Read filters to be applied before analysis", optional=true)
+    public final List<String> readFilterNames = new ArrayList<>(); // preserve order
+
+    final String disableReadFilterArgName = "disableReadFilter";
+    @Argument(fullName = disableReadFilterArgName,
+            shortName = "df",
+            doc="Read filters to be disabled before analysis", optional=true)
+    public final Set<String> disableFilters = new HashSet<>();
```
droazen: What if the user disables a filter that's on by default, but supplies argument(s) for the disabled filter?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * A GATKCommandLinePluginDescriptor for the ReadFilter class
+ */
+public class GATKReadFilterPluginDescriptor extends GATKCommandLinePluginDescriptor<ReadFilter> {
+
+    @Argument(fullName = StandardArgumentDefinitions.READ_FILTER_LONG_NAME,
+            shortName = StandardArgumentDefinitions.READ_FILTER_SHORT_NAME,
+            doc="Read filters to be applied before analysis", optional=true)
+    public final List<String> readFilterNames = new ArrayList<>(); // preserve order
```
droazen: What if a tool has a filter on by default that takes arguments?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,199 @@
...
+            throws IllegalAccessException, InstantiationException {
+        final ReadFilter readFilter = (ReadFilter) pluggableClass.newInstance();
+        // use getSimpleName; the classes are all in the same package
+        readFilters.put(pluggableClass.getSimpleName(), readFilter);
+        return readFilter;
+    }
+
+    // Verify that this target plugin class's dependent value has been specified
+    @Override
+    public boolean isDependentArgumentAllowed(Class<?> targetPluginClass) {
```
droazen: I think this needs to become a way of registering plugin classes for which arguments were encountered. Then, after parsing is finished and the plugin system creates a final list of plugins selected (using both the selection/exclusion args and the tool defaults), you can do a second validation pass here to throw if args were specified for a plugin that wasn't selected.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKCommandLinePluginDescriptor.java
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Return the allowable values for the String argument specified by
+     * longArgName. Called by the command line parser to generate a usage
+     * string. If the value is unrecognized, the implementation should
+     * throw IllegalArgumentException.
+     *
+     * @param longArgName
+     * @return Set<String> of allowable values</String>, or null if any value is allowed
+     */
+    public abstract Set<String> getAllowedStringValues(String longArgName);
```
droazen: Needs better name perhaps? Something like `getAllowedValuesForDescriptorArgument()`

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKCommandLinePluginDescriptor.java
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Given a consumer of a list of plugin objects of type T, call the consumer's
+     * accept method with the list actual instances accumulated by the plugin (those
+     * actually specified on the command line). This gives clients of the plugin
+     * objects a chance to retrieve the actual list.
+     *
+     * @param consumer function that accepts a collection of plugin instances
+     */
+    public abstract void getInstances(Consumer<Collection<T>> consumer);
```
droazen: Should this getter just return `Collection<T>`, instead of taking a `Consumer` and returning nothing?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,199 @@
...
+    public Map<String, ReadFilter> readFilters = new HashMap<>();
+
+    /////////////////////////////////////////////////////////
+    // GATKCommandLinePluginDescriptor implementation methods
+
+    @Override
+    public Class<?> getPluginClass() {return org.broadinstitute.hellbender.engine.filters.ReadFilter.class;}
+
+    @Override
+    public String getPackageName() {return "org.broadinstitute.hellbender.engine.filters";};
```
droazen: Declare package/plugin class as constants at the top

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,199 @@
...
+        final ArrayList<ReadFilter> filters = new ArrayList<>(readFilters.size());
+        readFilterNames.forEach(s -> filters.add(getReadFilterForName(s)));
+        consumer.accept(filters);
+    }
+
+    // Find the instance of the given read filter in the instance list.
+    private ReadFilter getReadFilterForName(final String filterName) {
+        ReadFilter rf = null;
+        for (Map.Entry<String, ReadFilter> entry : readFilters.entrySet()) {
+            if (entry.getKey().endsWith(filterName)) {
```
droazen: Can you do a map lookup here?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,199 @@
...
+        readFilterNames.forEach(s -> {
+            ReadFilter trf = readFilters.get(s);
+            if (null == trf) {
+                throw new UserException.CommandLineException("Unrecognized read filter name: " + s);
+            }
+            else {
+                requestedReadFilters.put(s, trf);
+            }
+        });
+        readFilters = requestedReadFilters;
```
droazen: I think all of this validation should be moved to a second pass that takes into account the tool-requested filters, as proposed above.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,199 @@
...
+     * @param samHeader - a SAMFileHeader to initialize read filter instances
+     * @param wrapperFunction - a wrapper function for each read filter; used to wrap read filters in
+     *                        CountingReadFilters
+     * @param mergeAccumulator - identity filter for base accumulator in list reduction (this is the default
+     *                         filter returned if no filters are left after reducing the merged filter list)
+     * @param <T> extends ReadFilter, type returned by the wrapperFunction
+     * @return
+     */
+    public <T extends ReadFilter> T getMergedReadFilters(
+            final List<ReadFilter> defaultFilters,
```
droazen: Rename `defaultFilters` to `toolFilters` to make it clear where they come from

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
```diff
@@ -98,15 +105,50 @@ private SimpleInterval getReadInterval(final GATKRead read) {
...
+
+        GATKReadFilterPluginDescriptor readFilterPlugin =
+                (GATKReadFilterPluginDescriptor)
+                        commandLineParser.getPluginDescriptor(GATKReadFilterPluginDescriptor.class);
+
+        // Unless all filters are disabled, merge the tool's default filters with the users's command line
+        // read filter requests, then initialize the resulting filters and wrap each one in a CountingReadFilter.
+        return readArguments.disableAllReadFilters == true ?
```
droazen: `disableAllReadFilters` should be handled internally in the descriptor validation

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
```diff
@@ -98,15 +105,50 @@ private SimpleInterval getReadInterval(final GATKRead read) {
...
+                defaultFilter :
+                readFilterPlugin.getMergedReadFilters(
+                        getDefaultReadFilters(),
+                        getHeaderForReads(),
+                        f -> new CountingReadFilter(f.getClass().getSimpleName(), f),
+                        (f1 , f2) -> f1.and(f2),
+                        defaultFilter
+                );
```
droazen: Have a simple method that tool authors can call to simplify the task of defining default read filters, that assumes that filters should be wrapped in `CountingReadFilters`, should be `anded` together, etc. Tool authors can always do it manually if they need exotic behavior, but most can just call into the easy method.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -11,41 +12,136 @@
...
+    public static MateOnSameContigOrNoMappedMate MATE_ON_SAME_CONTIG_OR_NO_MAPPED_MATE = new MateOnSameContigOrNoMappedMate();
+    public static MateDifferentStrand MATE_DIFFERENT_STRAND = new MateDifferentStrand();
+    public static NotDuplicate NOT_DUPLICATE = new NotDuplicate();
+    public static NonZeroReferenceLengthAlignment NON_ZERO_REFERENCE_LENGTH_ALIGNMENT = new NonZeroReferenceLengthAlignment();
+    public static PassesVendorQualityCheck PASSES_VENDOR_QUALITY_CHECK = new PassesVendorQualityCheck();
+    public static PrimaryAlignment PRIMARY_ALIGNMENT = new PrimaryAlignment();
+    public static ReadLengthEqualsCigarLength READLENGTH_EQUALS_CIGARLENGTH = new ReadLengthEqualsCigarLength();
+    public static SeqIsStored SEQ_IS_STORED = new SeqIsStored();
+    public static ValidAlignmentStart VALID_ALIGNMENT_START = new ValidAlignmentStart();
+    public static ValidAlignmentEnd VALID_ALIGNMENT_END = new ValidAlignmentEnd();
```
droazen: These should be `static final`

src/main/java/org/broadinstitute/hellbender/engine/filters/FragmentLengthReadFilter.java
```diff
@@ -3,14 +3,19 @@
...
+public final class FragmentLengthReadFilter extends ReadFilter implements Serializable  {
     private static final long serialVersionUID = 1l;
 
-    @Argument(fullName = "maxFragmentLength", shortName = "maxFragment", doc="Keep only read pairs with fragment length at most equal to the given value", optional=true)
-    public int maxFragmentLength = 1000000;
+    @Argument(fullName = "maxFragmentLength",
+            shortName = "maxFragment",
+            doc = "Keep only read pairs with fragment length at most equal to the given value",
+            optional = false)
+    public Integer maxFragmentLength; // no default value, since the command line parser will think its set
```
droazen: Can we allow args in plugins to have default values? This arg, for example, should definitely be optional, as it was originally, as we don't want to require the user to always specify `maxFragmentLength` when using this filter. Should be possible if you add a callback to your descriptor at the point in the arg parsing where it injects values into filter instances, to keep track of which filters had args specified (needed for post-parsing validation when you build the final list of selected filters)

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,199 @@
...
+        final ReadFilter readFilter = (ReadFilter) pluggableClass.newInstance();
+        // use getSimpleName; the classes are all in the same package
+        readFilters.put(pluggableClass.getSimpleName(), readFilter);
+        return readFilter;
+    }
+
+    // Verify that this target plugin class's dependent value has been specified
+    @Override
+    public boolean isDependentArgumentAllowed(Class<?> targetPluginClass) {
+        return readFilterNames.contains(targetPluginClass.getSimpleName());
```
droazen: Filters requested by the tool and filters excluded should be taken into account here (or validation has to be delayed somehow). May need to have the arg parsing system notify via callback whenever it encounters an argument for a particular plugin class, so that you can later determine whether the plugin args the user specified were appropriate given the final selection of plugins.

droazen:
Partial review complete -- will do a second pass after initial revisions.

droazen:
Can you make sure that all read filter args have long-ish/unique short and long names? Single-letter names should be reserved only for important universal GATK args, not plugin args.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/InsertSizeMetricsCollectorSpark.java
```diff
@@ -119,18 +119,35 @@ public InsertSizeMetricsReadFilter(final InsertSizeMetricsArgumentCollection.End
...
+            if(filterSecondaryAlignments)     { tempFilter = tempFilter.and(new ReadFilter() {
+                private static final long serialVersionUID = 1L;
+                @Override public boolean test(final GATKRead read){return !read.isSecondaryAlignment();}});}
+            if(filterSupplementaryAlignments) { tempFilter = tempFilter.and(new ReadFilter() {
+                private static final long serialVersionUID = 1L;
+                @Override public boolean test(final GATKRead read){return !read.isSupplementaryAlignment();}});}
+
+            if(0!=MQThreshold)  { tempFilter = tempFilter.and(new ReadFilter() {
+                private static final long serialVersionUID = 1L;
+                @Override public boolean test(final GATKRead read){return read.getMappingQuality() >= MQThreshold;}});};
```
cmnbroad: Many of these inline filters can be replaced with ReadFilterLibrary instances.

cmnbroad:
Rather than delaying argument validation to a second pass, I changed the sequence so the tool instantiates the descriptor and gives it the tool defaults right from the start, (and then passes the descriptor instance(s) to the arg parser) so it has all the state it needs to validate at arg parsing time. Also, I reverted the removal of the package limitation for plugins, at least temporarily, since searching through all packages looked like it slowed down the integration tests.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: This case will be detected/rejected now during final post-parse descriptor validation.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Done (private).

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKCommandLinePluginDescriptor.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReadInputArgumentCollection.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Rejected at descriptor validation time.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Done, with tests added.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKCommandLinePluginDescriptor.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKCommandLinePluginDescriptor.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Yes; the whole method is eliminated.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Tool defaults are passed in to the descriptor now before parsing, so this method can validate all inconsistencies.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: These are passed in to the constructor now.

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/filters/FragmentLengthReadFilter.java
cmnbroad: Yes, all the individual filters have been updated to have defaults where appropriate.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Done, with tests added for these cases.

cmnbroad:
Let me know if you'd prefer a squash/rebase before review.

droazen:
@lbergelson Would you be able to do the second-pass review here, since I did the first pass? In particular, it would be good if you could scrutinize the changes to the arg parser.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -138,16 +148,38 @@ private String getUsagePreamble() {
...
+     *                          null, no descriptors are loaded.
+     */
+    public CommandLineParser(
+            final Object callerArguments,
+            final List<? extends GATKCommandLinePluginDescriptor<?>> pluginDescriptors) {
         this.callerArguments = callerArguments;
 
-        createArgumentDefinitions(callerArguments);
+        createArgumentDefinitions(callerArguments, null);
+        if (pluginDescriptors != null) {
```
lbergelson: I think we should generally prefer passing an empty list vs a null for something like this.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -138,16 +148,38 @@ private String getUsagePreamble() {
...
+     *                        this command line parser.
+     * @param pluginDescriptors A list of {@link GATKCommandLinePluginDescriptor} objects that
+     *                          should be used by this command line parser to extend the list of
+     *                          command line arguments with dynamically discovered plugins. If
+     *                          null, no descriptors are loaded.
+     */
+    public CommandLineParser(
+            final Object callerArguments,
+            final List<? extends GATKCommandLinePluginDescriptor<?>> pluginDescriptors) {
         this.callerArguments = callerArguments;
```
lbergelson: While you're here could you add null checks for callerArguments?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -138,16 +148,38 @@ private String getUsagePreamble() {
...
+        createArgumentDefinitions(callerArguments, null);
+        if (pluginDescriptors != null) {
+            createCommandLinePluginArgumentDefinitions(pluginDescriptors);
+        }
 
         this.programProperties = this.callerArguments.getClass().getAnnotation(CommandLineProgramProperties.class);
     }
 
-    private List<ArgumentDefinition> createArgumentDefinitions(final Object callerArguments) {
+    private void createArgumentDefinitions(
```
lbergelson: good catch,  don't know how this ended up with a return type...

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -157,20 +189,56 @@ public CommandLineParser(final Object callerArguments) {
...
                 }
             }
         }
-        return null;
     }
 
+    // Find all the instances of plugins specified by the provided plugin descriptors
+    private void createCommandLinePluginArgumentDefinitions(
+            final List<? extends GATKCommandLinePluginDescriptor<?>> requestedPluginDescriptors) {
+        // For each descriptor, create an argument definition for the descriptor itself, then
```
lbergelson: should this read "create argument definitions" instead of being singular?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -195,21 +263,41 @@ public void usage(final PrintStream stream, final boolean printCommon) {
...
-        if (optArgs != null && !optArgs.isEmpty()) {
-            stream.println("\nOptional Arguments:\n");
-            optArgs.stream().forEach(argumentDefinition -> printArgumentUsage(stream, argumentDefinition));
+        // now the contingent/dependent args (those controlled by a plugin descriptor)
+        List<ArgumentDefinition> conditionalArgs = allArgsMap.get(false);
+        if (null != conditionalArgs && !conditionalArgs.isEmpty()) {
+            // group the argdefs by their parent (plugin) class
+            Map<String, List<ArgumentDefinition>> groupedArgsMap = conditionalArgs.stream()
+                    .collect(Collectors.groupingBy(argDef -> argDef.parent.getClass().getSimpleName()));
+            stream.println("\nConditional Arguments:\n");
```
lbergelson: This is nitpicky, but it would be nice if instead of just "Conditional Arguments" these were grouped by plugin descriptor.  I.e. All the read filters are together in a block that says "Read Filter Arguments", and a second plugin would be listed separately.  We could patch that in a later pull request if you want though.

lbergelson: Also, could we sort these alphabetically before printing them?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -349,7 +437,50 @@ private void assertArgumentsAreValid() {
...
+    // specified) and throwing for any that have been set but are not allowed. Note that we don't trim
+    // the list of plugins themselves (just the argument definitions), since the plugin may contain
+    // other arguments that require validation.
+    private void validatePluginArguments() {
+        final List<ArgumentDefinition> actualArgumentDefinitions = new ArrayList<>();
+        for (final ArgumentDefinition argumentDefinition : argumentDefinitions) {
+            if (null == argumentDefinition.controllingDescriptor) {
+                actualArgumentDefinitions.add(argumentDefinition);
+            }
+            else {
```
lbergelson: there's funny line breaking in this block and the one following it, we usually go with` } else {` I think.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -349,7 +437,50 @@ private void assertArgumentsAreValid() {
...
+    // validate any that are plugin class arguments against the controlling descriptor, trimming
+    // the list of argument definitions along the way by removing any that have not been set
+    // (so validation doesn't complain about missing required arguments for plugins that weren't
+    // specified) and throwing for any that have been set but are not allowed. Note that we don't trim
+    // the list of plugins themselves (just the argument definitions), since the plugin may contain
+    // other arguments that require validation.
+    private void validatePluginArguments() {
+        final List<ArgumentDefinition> actualArgumentDefinitions = new ArrayList<>();
+        for (final ArgumentDefinition argumentDefinition : argumentDefinitions) {
+            if (null == argumentDefinition.controllingDescriptor) {
```
lbergelson: Could you extract a method that does this check and name it to make it more clear what this is testing for, maybe `isPluginArgument`.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -493,7 +624,13 @@ private String makeArgumentDescription(final ArgumentDefinition argumentDefiniti
         } else {
             sb.append("Required. ");
         }
-        sb.append(getOptions(getUnderlyingType(argumentDefinition.field)));
+        if (GATKCommandLinePluginDescriptor.class.isAssignableFrom(argumentDefinition.parent.getClass()) &&
```
lbergelson: Please add a comment here explaining this check.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -513,10 +649,22 @@ private String makeArgumentDescription(final ArgumentDefinition argumentDefiniti
         return sb.toString();
     }
 
+    private void usageForPluginDescriptorArgument(final ArgumentDefinition argDef, final StringBuilder sb) {
+        final GATKCommandLinePluginDescriptor<?> descriptor = (GATKCommandLinePluginDescriptor<?>) argDef.parent;
+        // this argument came from a plugin descriptor; delegate to get the list of allowed values
+        final Set<String> allowedValues = descriptor.getAllowedValuesForDescriptorArgument(argDef.getLongName());
+        if (allowedValues == null) {
```
lbergelson: Usually we'd want to prefer empty collections over null values, is there a need to distinguish the two here

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -513,10 +649,22 @@ private String makeArgumentDescription(final ArgumentDefinition argumentDefiniti
...
+            sb.append("Any value allowed");
+        } else {
+            sb.append("Allowed values are: ");
```
lbergelson: Sort these so they're output in alphabetical order (or if you want the plugin descriptor to control the order sort them there.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKCommandLinePluginDescriptor.java
```diff
@@ -0,0 +1,159 @@
...
+ *  validateDependentArgumentAllowed  - once for each plugin argument value that has been
+ *  specified on the command line for a plugin that is controlled by this descriptor
+ *
+ *  validateArguments() - once when argument parsing is complete
+ *  getAllInstances() - whenever the pluggable class consumer wants the resulting plugin instances
+ *
+ *  getAllowedValuesForDescriptorArgument is only called when the command line parser is constructing
+ *  a help/usage message.
+ */
+public abstract class GATKCommandLinePluginDescriptor<T> {
```
lbergelson: We might want to add a `getPluginName()` to support printing the plugin args in an even more user friendly way.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,334 @@
...
+
+    private static final String pluginPackageName = "org.broadinstitute.hellbender.engine.filters";
+    private static final Class<?> pluginBaseClass = org.broadinstitute.hellbender.engine.filters.ReadFilter.class;
+
+    @Argument(fullName = StandardArgumentDefinitions.READ_FILTER_LONG_NAME,
+            shortName = StandardArgumentDefinitions.READ_FILTER_SHORT_NAME,
+            doc="Read filters to be applied before analysis", optional=true)
+    public final List<String> userReadFilterNames = new ArrayList<>(); // preserve order
+
+    final String disableReadFilterArgName = "disableReadFilter";
```
lbergelson: This should probably be static

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,334 @@
...
+    @Argument(fullName = disableReadFilterArgName,
+            shortName = "df",
```
lbergelson: It's strange that this is lower case but read filter is 'RF'

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,334 @@
...
+     * Validate the list of arguments and reduce the list of read filters to those
+     * actually seen on the command line. This is called by the command line parser
+     * after all arguments have been parsed.
+     */
+    @Override
+    public void validateArguments() {
+        // throw if any filter is both enabled *and* disabled by the user
+        final Set<String> enabledAndDisabled = new HashSet<>(userReadFilterNames);
+        enabledAndDisabled.retainAll(disableFilters);
+        enabledAndDisabled.forEach(
```
lbergelson: It's a bit strange to have this as a forEach since it can really only happen once.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,334 @@
...
+        readFilters = requestedReadFilters;
+    }
+
+    /////////////////////////////////////////////////////////
+    // ReadFilter plugin-specific helper methods
+
+    /**
+     * Determine if a particular ReadFilter was disabled on the command line.
+     * @param filterName name of the filter to query
+     * @return true if the name appears the list of disabled filters
```
lbergelson: typo: true if the name appears **in** the list of disabled filters

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -65,20 +64,52 @@ public boolean includeNs() {
...
+     * composition methods.
      */
     public CountingReadFilter makeReadFilter(){
-        return new CountingReadFilter("Wellformed", new WellformedReadFilter(getHeaderForReads()))
-                .and(new CountingReadFilter("Mapped", ReadFilterLibrary.MAPPED));
+        // Unless all filters are disabled, merge the tool's default filters with the users's command line
+        // read filter requests, then initialize the resulting filters and wrap each one in a CountingReadFilter.
+        GATKReadFilterPluginDescriptor readFilterPlugin =
+                (GATKReadFilterPluginDescriptor)
+                        commandLineParser.getPluginDescriptor(GATKReadFilterPluginDescriptor.class);
```
lbergelson: I think it would be nicer if the cast was pushed into `getPluginDescriptor`

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -71,6 +73,14 @@
     public static final String COMMENT = "#";
     public static final String POSITIONAL_ARGUMENTS_NAME = "Positional Argument";
 
+    // Map from (full class) name of each GATKCommandLinePluginDescriptor requested and
+    // found to the actual descriptor instance
+    private Map<String, GATKCommandLinePluginDescriptor<?>> pluginDescriptors = new HashMap<>();
+
+    // Return the plugin instance corresponding to the targetDescriptor class
+    public GATKCommandLinePluginDescriptor<?> getPluginDescriptor(Class<?> targetDescriptor) {
```
lbergelson: I commented below, about this, but I think it would probably be better add a type parameter to this and have it cast the result internally to the target descriptor type.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -11,41 +12,136 @@
...
+    /**
+     * local classes for static read filters
+     */
+    public static class AllowAllReads extends ReadFilter {
+        private static final long serialVersionUID = 1L;
+        @Override public boolean test(final GATKRead read){return true;}}
 
     //Note: do not call getCigar to avoid creation of new Cigar objects
-    public static final ReadFilter READLENGTH_EQUALS_CIGARLENGTH = read -> read.isUnmapped() || read.getLength() == Cigar.getReadLength(read.getCigarElements());
+    public static class CigarIsSupported extends ReadFilter {
```
lbergelson: Should these all have ReadFilter as part of their name?  It's a bit inconsistent on the commandline that some do and some don't.

lbergelson:
What happens if an inactive read filter has the same argument as one of the arguments in a tool, is there an error, or only if there is an active read filter with a name collision? 
 What if two active read filters have the same argument name?  Are there tests for name collision?  I didn't see them but I might have missed it.

lbergelson:
@cmnbroad Back to you.  Some minor cosmetic comments about printing the help output.  I think it looks good though.  This version with the plugin descriptors is so much nicer than the early version.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Good idea. Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: At first I thought not, but I see your point; there can be multiple defs for the descriptor itself. Updated.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Grouped and sorted, both. While I was at it I sorted the required and optional args as well.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Nope. Changed the interface javadoc and the code to used an empty list.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKCommandLinePluginDescriptor.java
cmnbroad: Yeah. Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Good point. Made it upper case.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Agreed. fixed.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
cmnbroad: Yes. Done.

cmnbroad:
Back to @lbergelson - not sure if you wanted to re-review.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
```diff
@@ -1023,6 +1054,13 @@ public boolean isFlag(){
             return field.getType().equals(boolean.class) || field.getType().equals(Boolean.class);
         }
 
+        /**
+         * Determine if this argument definition is controlled by a plugin descripto (and thus subject to
```
lbergelson: missing r in descripto

lbergelson:
@cmnbroad Looks good!  Lets get this merged finally.  Rebase and merge at will.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,337 @@
...
+    private static final Class<?> pluginBaseClass = org.broadinstitute.hellbender.engine.filters.ReadFilter.class;
+
+    @Argument(fullName = StandardArgumentDefinitions.READ_FILTER_LONG_NAME,
+            shortName = StandardArgumentDefinitions.READ_FILTER_SHORT_NAME,
+            doc="Read filters to be applied before analysis", optional=true)
+    public final List<String> userReadFilterNames = new ArrayList<>(); // preserve order
+
+    final static String disableReadFilterArgName = "disableReadFilter";
+    @Argument(fullName = disableReadFilterArgName,
+            shortName = "DF",
```
magicDGS: Sorry for the meddling, but while trying to understand how a plugin will be implemented I found this: the names for the read filter are in `StandardArgumentDefinitions`, the disable read filter in a static field (only the long name, the other is hardcoded in the annotation), and the one for disable all the read filters is inside the annotation. I think that it would be nice to have all this values in `StandardArgumentDefinitions` to be able to have a class to check possible collisions...

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -0,0 +1,337 @@
...
+
+    /**
+     * Return a display name to identify this plugin to the user
+     * @return A short user-friendly name for this plugin.
+     */
+    @Override
+    public String getDisplayName() { return StandardArgumentDefinitions.READ_FILTER_LONG_NAME; }
+
+    /**
+     * @return the class object for the base class of all plugins managed by thid descriptor
```
magicDGS: Again, sorry for being picky: `thid` -> `this`

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Fair enough - its an artifact left over from the first implementation. Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Done

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineParser.java
cmnbroad: Done.

cmnbroad:
Also, good point on the arg name collision. The command line parser handles this (for any args, even across plugins or across anywhere) but I didn't see any tests, so I added tests to both CommandLineParserUnitTest and CommandLineParserPluginUnitTest.

PR #1977, opened 2016-07-05
PileupElementUnitTest no longer skipping a test

lbergelson:
PileupElementUnitTest.testBadOffsetWithinBounds was being skipped because the data provider name was wrong.


akiezun:
👍

PR #1978, opened 2016-07-06
Tweaks to SV breakpoint tools.

tedsharpe:
Filter out low complexity kmers.
Add ability to specify genomic intervals to be added to kmer kill set.
Limit size of fastqs.  Add each read's mapping location to fastq.


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBadGenomicKmersSpark.java
```diff
@@ -122,4 +135,36 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+                        .forEach(kmers::add);
+            }
+            catch ( final IOException ioe ) {
+                throw new GATKException("Can't get reference sequence bases for " + interval, ioe);
+            }
+        }
+        return kmers;
+    }
+
+    private static Collection<SVKmer> uniquefy( final Collection<SVKmer> coll1, final Collection<SVKmer> coll2 ) {
```
cwhelan: I'd usually spell this 'uniquify', not a big deal though.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBadGenomicKmersSpark.java
```diff
@@ -54,32 +57,41 @@ protected void runTool( final JavaSparkContext ctx ) {
         SAMSequenceDictionary dict = null;
         if ( hdr != null ) dict = hdr.getSequenceDictionary();
         final PipelineOptions options = getAuthenticatedGCSOptions();
-        final List<SVKmer> killList = findBadGenomicKmers(ctx, getReference(), options, dict);
+        final ReferenceMultiSource referenceMultiSource = getReference();
+        Collection<SVKmer> killList = findBadGenomicKmers(ctx, kSize, minEntropy, referenceMultiSource, options, dict);
+        if ( hasIntervals() ) {
```
cwhelan: It'd be good to document somewhere what the intervals mean, since we're just using the standard intervals parameter. Someone new to the tool might think that the intervals are the places over which we'll find non-unique kmers or something similar.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -327,11 +339,14 @@ private void addAssemblyQNames(
...
     private static void writeFastq( final Tuple2<Integer, List<byte[]>> intervalAndFastqs,
-                                    final String outputDir ) {
+                                    final String outputDir, final int maxFastqSize ) {
         final List<byte[]> fastqsList = intervalAndFastqs._2;
         SVFastqUtils.sortFastqRecords(fastqsList);
 
-        final String fileName = outputDir + "/assembly" + intervalAndFastqs._1 + ".fastq";
+        String fileName = outputDir + "/assembly" + intervalAndFastqs._1 + ".fastq";
+        final int fastqSize = fastqsList.stream().mapToInt(fastqRec -> fastqRec.length).sum();
+        if ( fastqSize > maxFastqSize ) fileName += ".toobig";
```
cwhelan: If I'm getting this right we're still writing out the big FASTQ file but with a different extension? I think that will still cause a problem for the downstream tool since it tries to load all files in the given directory. Maybe put this in a different directory, or just skip writing it altogether? Sounds like this might be moot with your recent changes anyway.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBadGenomicKmersSpark.java
tedsharpe: Yes.  Your way is correct.  Either a typo or a brain cramp.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBadGenomicKmersSpark.java
tedsharpe: Good point.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: I wanted some output to show that the tool didn't just leave something out by accident.  I'll do something smarter.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizerWithLowComplexityFilter.java
```diff
@@ -0,0 +1,90 @@
...
+
+            if ( ++validBaseCount >= kSize ) {
+                final double entropy = Arrays.stream(baseCounts).mapToDouble(count -> entropies[count]).sum();
+                if ( entropy >= minEntropy ) return tmpKmer;
+            }
+        }
+        return null;
+    }
+
+    private static double[] getEntropies( final int kSize ) {
```
cwhelan: Can you add a comment explaining this? Ie something like "This returns an array where the i'th entry is the contribution to the total entropy of a kmer of having i occurrences of a single base a kmer of the given size"..

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizerWithLowComplexityFilter.java
```diff
@@ -0,0 +1,90 @@
...
+        if ( entropies == null || entropies.length != kSize+1 ) {
+            entropies = new double[kSize+1];
+            final double minusLog2 = -Math.log(2);
+            // note: entropies[0] = 0 by definition, and it's already set to that value.
+            // the same is true for entropies[kSize].
+            for ( int idx = 1; idx != kSize; ++idx ) {
+                final double probability = 1. * idx / kSize;
```
cwhelan: Super nitpicky but I sort of prefer to write "1.0" over "1.". Up to you though.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizerWithLowComplexityFilter.java
```diff
@@ -0,0 +1,90 @@
...
+                new SVKmerizerWithLowComplexityFilter(seq, kSize, minEntropy)).spliterator(), false);
+    }
+
+    public static Stream<SVKmer> stream( final byte[] seq, final int kSize, final double minEntropy ) {
+        return StreamSupport.stream(((Iterable<SVKmer>)() ->
+                new SVKmerizerWithLowComplexityFilter(seq, kSize, minEntropy)).spliterator(), false);
+    }
+
+    protected SVKmer nextKmer( SVKmer tmpKmer, int validBaseCount ) {
+        final double[] entropies = getEntropies(kSize);
```
cwhelan: I find it pretty confusing to have a variable with the same name and type as a static variable on the class. Couldn't we just rename this method to `initializeEntropies` and then access entropies statically? Also this seems like it allows someone to do something bad by constructing two different SVKmerizers with different kSizes -- we might want to throw an exception if that happens, or alternatively allow for different 'entropies' arrays to be looked up statically.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSparkUnitTest.java
```diff
@@ -71,13 +71,8 @@ public void getQNamesTest() {
...
     @Test(groups = "spark")
     public void getKmerIntervalsTest() {
-        final HopscotchSet<SVKmer> killSet = new HopscotchSet<>(2);
-        killSet.add(SVKmerizer.toKmer("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"));
-        killSet.add(SVKmerizer.toKmer("CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC"));
         final List<SVKmer> highCountKmers = FindBreakpointEvidenceSpark.getHighCountKmers(params, reads, locations, null);
-        Assert.assertEquals(highCountKmers.size(), 2);
-        killSet.addAll(highCountKmers);
-        Assert.assertEquals(killSet.size(), 2);
+        Assert.assertEquals(highCountKmers.size(), 0);
```
cwhelan: Are you still testing that `getHighCountKmers` can return some high count kmers somewhere?

cwhelan:
Review done, looks good with just a few small comments.

tedsharpe:
PR comments addressed.

cwhelan:
Thanks for the updates. Looks good to me now.

PR #1979, opened 2016-07-06
Cleaned code in MathUtils and ReadLikelihoods; using Utils methods

davidbenjamin:
This consists almost entirely of using `Utils.nonNull` and `Utils.validateArg` in code ported from GATK 3.  There are less trivial but straightforward simplifications of code in `MathUtils` and `ReadLikelihoods`.  @droazen and @lbergelson is one of you willing to review this mind-numbing PR, or suggest a victim?  It should be quick.


droazen:
@akiezun Do you have time to have a look at this one?

akiezun:
no problem

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
```diff
@@ -146,18 +147,14 @@ public FeatureDataSource( final File featureFile, final FeatureCodec<T, ?> codec
      * @param queryLookaheadBases look ahead this many bases during queries that produce cache misses
      */
     public FeatureDataSource( final File featureFile, final FeatureCodec<T, ?> codec, final String name, final int queryLookaheadBases ) {
-        if ( featureFile == null || codec == null ) {
-            throw new IllegalArgumentException("FeatureDataSource cannot be created from null file/codec");
-        }
-        if ( queryLookaheadBases < 0 ) {
-            throw new IllegalArgumentException("Query lookahead bases must be >= 0");
-        }
+        this.featureFile = Utils.nonNull(featureFile);
```
akiezun: add message here, otherwise it's hard to debug which of the null checks failed

src/main/java/org/broadinstitute/hellbender/engine/ReferenceMemorySource.java
```diff
@@ -95,18 +95,10 @@ public ReferenceSequence queryAndPrefetch( final String contig, final long start
...
-        if (startIndex>=basesBytes.length) {
-            throw new IllegalArgumentException("Asking for start "+start+" on contig "+contig+" but the ReferenceData only has data until "+bases.getInterval().getEnd());
-        }
-        if (startIndex+length>basesBytes.length) {
-            throw new IllegalArgumentException("Asking for stop "+stop+" on contig "+contig+" but the ReferenceData only has data until "+bases.getInterval().getEnd());
-        }
-        if (length<0) {
-            throw new IllegalArgumentException("Asking for stop<start ("+stop+"<"+start+")");
-        }
+        Utils.validateArg(startIndex >= 0, "Asking for start "+start+" on contig "+contig+" but the ReferenceData only has data starting at "+bases.getInterval().getStart());
```
akiezun: replace the first two checks with `Utils.validIndex`

src/main/java/org/broadinstitute/hellbender/tools/exome/Sample.java
```diff
@@ -29,24 +31,12 @@
      *    it contains any {@code null} or some repeated read-group ids.
      */
     public Sample(final String id, final Collection<String> readGroups) {
+        this.id = Utils.nonNull(id);
```
akiezun: add message to distinguish two `nonNull` calls

src/main/java/org/broadinstitute/hellbender/tools/exome/Sample.java
```diff
@@ -29,24 +31,12 @@
...
-                throw new IllegalArgumentException("the input read-groups collection contains null values");
-            } else if (!readGroupSet.add(readGroupId)) {
-                throw new IllegalArgumentException(
-                        String.format("the same read-groups '%s' is found multiple times in the read-groups collection",
-                                readGroupId));
-
-            }
-        }
+        final Set<String> readGroupSet = new LinkedHashSet<>(readGroups);
+        Utils.validateArg(readGroupSet.size() == readGroups.size(), "readGroups contains duplicates");
```
akiezun: technically, the error message used to be more informative by including the name of the read group. Can you add that back?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmer.java
```diff
@@ -43,7 +44,7 @@
      *  Makes an empty SVKmer.  If you call toString on it, it'll look like poly-A.
      */
     public SVKmer( final int kSize ) {
-        if ( kSize < 1 || kSize > 63 ) throw new IllegalArgumentException("K must be between 1 and 63.");
+        Utils.validateArg(kSize >= 1 && kSize <= 64, "K must be between 1 and 63.");
```
akiezun: This is incorrect. It should be `kSize >= 1 && kSize <= 63`, ie value 64 should **not** be allowed

akiezun: that suggests that there's a missing test for value 64. Can you add such a test to confirm that this check blows up on it?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHet.java
```diff
@@ -82,9 +83,7 @@
      */
     @VisibleForTesting
     static double exactTest(final int hetCount, final int refCount, final int homCount) {
-        if (hetCount < 0 || refCount < 0 || homCount < 0) {
-            throw new IllegalArgumentException("Genotype counts cannot be less than 0");
-        }
+        Utils.validateArg(hetCount >= 0 && refCount >= 0 && homCount >= 0, "Genotype counts cannot be less than 0");
```
akiezun: split into 3 checks to make error message more specific

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeLikelihoodCalculator.java
```diff
@@ -507,18 +490,12 @@ public int alleleCountsToIndex(final int ... alleleCountArray) {
...
      */
     private int alleleHeapToIndex() {
-        if (alleleHeap.size() != ploidy) {
-            throw new IllegalArgumentException("the sum of allele counts must be equal to the ploidy of the calculator");
-        }
-        if (alleleHeap.peek() >= alleleCount) {
-            throw new IllegalArgumentException("invalid allele " + alleleHeap.peek() + " more than the maximum " + (alleleCount - 1));
-        }
+        Utils.validateArg(alleleHeap.size() == ploidy, "the sum of allele counts must be equal to the ploidy of the calculator");
+        Utils.validateArg(alleleHeap.peek()< alleleCount, "invalid allele " + alleleHeap.peek() + " more than the maximum " + (alleleCount - 1));
```
akiezun: add space before `<`

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
```diff
@@ -603,46 +598,43 @@ protected final boolean passesCallThreshold(final double conf) {
...
-            final double log10ACgt0Posterior = log10ACgt0Likelihood + log10ACgt0Prior;
-            final double log10PosteriorNormalizationConstant = MathUtils.approximateLog10SumLog10(log10ACeq0Posterior, log10ACgt0Posterior);
-            //TODO End of lousy part.
-
-            final double normalizedLog10ACeq0Posterior = log10ACeq0Posterior - log10PosteriorNormalizationConstant;
-            // This is another condition to return a 0.0 also present in AFCalculator code as well.
-            if (normalizedLog10ACeq0Posterior >= QualityUtils.qualToErrorProbLog10(configuration.genotypeArgs.STANDARD_CONFIDENCE_FOR_EMITTING)) {
-                return 0.0;
+        Utils.nonNull(log10GenotypeLikelihoods, "the input likelihoods cannot be null");
+        Utils.validateArg(log10GenotypeLikelihoods.length == this.configuration.genotypeArgs.samplePloidy + 1, "wrong likelihoods dimensions");
```
akiezun: print the offending argument values

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtils.java
```diff
@@ -35,10 +36,9 @@ public static VariantContext composeGivenAllelesVariantContextFromRod(final Feat
                                                                           final boolean snpsOnly,
                                                                           final Logger logger,
                                                                           final FeatureInput<VariantContext> allelesBinding) {
-        if ( tracker == null ) throw new IllegalArgumentException("the tracker cannot be null");
-        if ( loc == null ) throw new IllegalArgumentException("the location cannot be null");
-        if ( allelesBinding == null ) throw new IllegalArgumentException("the alleles binding cannot be null");
-
+        Utils.nonNull(tracker);
```
akiezun: add message to all null checks (only if there are more then 1 in a method, as here)

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtils.java
```diff
@@ -68,18 +68,17 @@ public static VariantContext composeGivenAllelesVariantContextFromRod(final Feat
      * @return a non-null list of haplotypes
      */
     public static List<Haplotype> composeGivenHaplotypes(final Haplotype refHaplotype, final List<VariantContext> givenHaplotypes, final GenomeLoc activeRegionWindow) {
-        if (refHaplotype == null) throw new IllegalArgumentException("the reference haplotype cannot be null");
-        if (givenHaplotypes == null) throw new IllegalArgumentException("given haplotypes cannot be null");
-        if (activeRegionWindow == null) throw new IllegalArgumentException("active region window cannot be null");
-        if (activeRegionWindow.size() != refHaplotype.length()) throw new IllegalArgumentException("inconsistent reference haplotype and active region window");
+        Utils.nonNull(refHaplotype);
```
akiezun: add message to null checks

src/main/java/org/broadinstitute/hellbender/utils/MRUCachingSAMSequenceDictionary.java
```diff
@@ -31,9 +31,8 @@
      * @param dict a non-null, non-empty sequencing dictionary
      */
     public MRUCachingSAMSequenceDictionary(final SAMSequenceDictionary dict) {
-        if ( dict == null ) throw new IllegalArgumentException("Dictionary cannot be null");
-        if (dict.isEmpty()) throw new IllegalArgumentException("Dictionary cannot have size zero");
-
+        Utils.nonNull( dict == null, "Dictionary cannot be null");
+        Utils.validateArg( dict.size() > 0, "Dictionary cannot have size zero");
```
akiezun: `! dict.isEmpty()` is conventional

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -229,15 +216,8 @@ public static double sumLog10(final double[] log10values) {
     public static int[] vectorDiff(final int[]x, final int[] y) {
```
akiezun: add a space between `[]` and `x`

akiezun: this whole method seems to be the same as `MathArrays.ebeSubtract`. I think we should just delegate

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -280,30 +249,20 @@ public static double log10sumLog10(final double[] log10p, final int start) {
...
-            else {
-                sum += Math.pow(10.0, scaled_val);
-            }
-        }
-        if ( Double.isNaN(sum) || sum == Double.POSITIVE_INFINITY ) {
-            throw new IllegalArgumentException("log10p: Values must be non-infinite and non-NAN");
-        }
-        return maxValue + (sum != 1.0 ? Math.log10(sum) : 0.0);
+        final double sum = 1.0 + IntStream.range(start, finish)
+                .filter(i -> i!= maxElementIndex)
```
akiezun: why were the `Double.NEGATIVE_INFINITY` values skipped before and why are they not skipped now?

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -362,21 +323,10 @@ public static double get(final double difference) {
...
 
         private static void initialize() {
-            if (cache == null) {
-                final int tableSize = (int) (MAX_TOLERANCE / TABLE_STEP) + 1;
-                cache = new double[tableSize];
-                for (int k = 0; k < cache.length; k++) {
-                    cache[k] = Math.log10(1.0 + Math.pow(10.0, -((double) k) * TABLE_STEP));
-                }
-            }
+            final int tableSize = (int) (MAX_TOLERANCE / TABLE_STEP) + 1;
```
akiezun: i think we should put back the null check and make the initialize method synchronized to prevent race conditions in case this cache is accessed via multiple threads
(the null check is needed because another thread may have created the cache after null is checked in `get`)

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -590,8 +525,7 @@ public static double binomialProbability(final int n, final int k, final double
      * binomial Probability(int, int, double) with log applied to result
      */
     public static double log10BinomialProbability(final int n, final int k, final double log10p) {
-        if ( log10p > 1e-18 )
-            throw new IllegalArgumentException("log10p: Log10-probability must be 0 or less");
+        Utils.validateArg(log10p < 1e-18, "log10p: Log10-probability must be 0 or less");
```
akiezun: `1.0e-18` is conventional

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: The original author of this code must have thought that `Double.NEGATIVE_INFINITY` is a symbolic object, but it is perfectly well-behaved as mandated by the Java spec.  Here if it is allowed to pass through we get `Math.pow(10.0, Double.NEGATIVE_INIFINITY - maxValue)` which evaluates to `0`, thus not affecting the sum.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -674,22 +608,14 @@ public static double log10SumLog10(final double[] log10Values, final int start,
 
         // we may decide to just normalize in log space without converting to linear space
         if (keepInLogSpace) {
-            for (int i = 0; i < array.length; i++) {
-                array[i] -= maxValue;
-            }
-            return array;
+            return Arrays.stream(array).map(x -> x - maxValue).toArray();
```
akiezun: i worry a bit about the performance of this stream operations here. Arrays are fast but streams create a lot of garbage and these math operations are heavily hit during some operations

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -748,12 +670,9 @@ public static int maxElementIndex(final double[] array) {
     }
 
     public static int maxElementIndex(final double[] array, final int start, final int endIndex) {
-        if (array == null || array.length == 0)
-            throw new IllegalArgumentException("Array cannot be null!");
-
-        if (start > endIndex) {
-            throw new IllegalArgumentException("Start cannot be after end.");
-        }
+        Utils.nonNull(array, "array may not be null");
```
akiezun: use `Utils.validIndex`

src/main/java/org/broadinstitute/hellbender/utils/genotyper/ReadLikelihoods.java
```diff
@@ -1339,13 +1300,11 @@ private void removeSampleReads(final int sampleIndex, final Collection<GATKRead>
...
         final PerReadAlleleLikelihoodMap result = new PerReadAlleleLikelihoodMap();
-        final int alleleCount = alleles.numberOfAlleles();
         final GATKRead[] sampleReads = readsBySampleIndex[sampleIndex];
-        final int sampleReadCount = sampleReads.length;
-        for (int a = 0; a < alleleCount; a++) {
+        for (int a = 0; a < alleles.numberOfAlleles(); a++) {
             final A allele = alleles.getAllele(a);
             final double[] readLikelihoods = valuesBySampleIndex[sampleIndex][a];
-            for (int r = 0; r < sampleReadCount; r++)
+            for (int r = 0; r < sampleReads.length; r++)
```
akiezun: just making sure you are aware of performance impact of this - accessing `.length` in a tight loop is not free

src/main/java/org/broadinstitute/hellbender/utils/genotyper/ReadLikelihoods.java
```diff
@@ -1339,13 +1300,11 @@ private void removeSampleReads(final int sampleIndex, final Collection<GATKRead>
     public PerReadAlleleLikelihoodMap toPerReadAlleleLikelihoodMap(final int sampleIndex) {
         Utils.validIndex(sampleIndex, samples.numberOfSamples());
         final PerReadAlleleLikelihoodMap result = new PerReadAlleleLikelihoodMap();
-        final int alleleCount = alleles.numberOfAlleles();
         final GATKRead[] sampleReads = readsBySampleIndex[sampleIndex];
-        final int sampleReadCount = sampleReads.length;
-        for (int a = 0; a < alleleCount; a++) {
+        for (int a = 0; a < alleles.numberOfAlleles(); a++) {
```
akiezun: perf comment: this will call `alleles.numberOfAlleles()` on every iteration of the loop. Previously, it was once, before the loop. 

Normally I would not care but these methods may be on the critical path in HC or GenotypeGVCFs. Can you run the HC on your branch to make sure performance is not degraded

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
```diff
@@ -256,14 +257,14 @@ public void computeLog10Likelihoods(final LikelihoodMatrix<Haplotype> logLikelih
...
-        if ( readBases == null ) throw new IllegalArgumentException("readBases cannot be null");
-        if ( readBases.length > maxReadLength ) throw new IllegalArgumentException("readBases is too long, got " + readBases.length + " but max is " + maxReadLength);
-        if ( readQuals.length != readBases.length ) throw new IllegalArgumentException("Read bases and read quals aren't the same size: " + readBases.length + " vs " + readQuals.length);
-        if ( insertionGOP.length != readBases.length ) throw new IllegalArgumentException("Read bases and read insertion quals aren't the same size: " + readBases.length + " vs " + insertionGOP.length);
-        if ( deletionGOP.length != readBases.length ) throw new IllegalArgumentException("Read bases and read deletion quals aren't the same size: " + readBases.length + " vs " + deletionGOP.length);
-        if ( overallGCP.length != readBases.length ) throw new IllegalArgumentException("Read bases and overall GCP aren't the same size: " + readBases.length + " vs " + overallGCP.length);
+        Utils.nonNull(haplotypeBases);
```
akiezun: message for nonNull checks

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessSettings.java
```diff
@@ -85,11 +88,8 @@ public OutputStreamSettings getStderrSettings() {
     }
 
     protected String[] checkCommand(String[] command) {
-        if (command == null)
-            throw new IllegalArgumentException("Command is not allowed to be null");
-        for (String s: command)
-            if (s == null)
-                throw new IllegalArgumentException("Command is not allowed to contain nulls");
+        Utils.nonNull(command);
+        Arrays.stream(command).forEach(s -> Utils.nonNull(s, "Command is not allowed to contain nulls"));
```
akiezun: i'd just convert to list and call containsNoNull

src/main/java/org/broadinstitute/hellbender/utils/variant/writers/HomRefBlock.java
```diff
@@ -164,7 +165,7 @@ public int getMinDP() {
     /** Get the median DP observed within this band
      * If there are an even number of DPs recorded in this band the median is the mean of the two middle values */
     public int getMedianDP() {
-        return (int) Math.round(MathUtils.median(DPs));
+        return (int) Math.round(new Median().evaluate(DPs.stream().mapToDouble(x->x).toArray()));
```
akiezun: rather than doing this dance, you could change the implementation of `MathUtils.median`, right?

note that this may have performance implications and requires a HC run

src/test/java/org/broadinstitute/hellbender/utils/RandomDNAUnitTest.java
```diff
@@ -52,21 +55,15 @@ public void testBases(){
...
     private int[] pairwiseAdd(int[] a, int[] b) {
-        if (a.length != b.length){
-            throw  new IllegalArgumentException();
-        }
-        final int[] results = new int[a.length];
-        for (int i = 0; i < a.length; i++) {
-            results[i] = a[i] + b[i];
-        }
-        return results;
+        Utils.validateArg(a.length == b.length, "lengths must be equal");
```
akiezun: is this `MathArrays.ebeAdd` ?

akiezun:
done with my review. small edits, 1 bug, and a few performance questions. I'd like to see a simple perf run of HC with and without those changes.

akiezun:
back to @davidbenjamin

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -109,9 +110,8 @@ public ReadsDataSource( final File samFile, SamReaderFactory customSamReaderFact
      *                               stringency SILENT is used.
      */
     public ReadsDataSource( final List<File> samFiles, SamReaderFactory customSamReaderFactory ) {
-        if ( samFiles == null || samFiles.isEmpty()) {
-            throw new IllegalArgumentException("ReadsDataSource cannot be created from empty file list");
-        }
+        Utils.nonNull(samFiles);
+        Utils.nonEmpty(samFiles, "ReadsDataSource cannot be created from empty file list");
```
lbergelson: nonEmpty includes a check for nonNull

lbergelson:
So there may be some unfortunate performance implications with some of these changes.  Utils.nonNull(value, message) and it's compatriots will always compute the message even if the error condition is not met.  Using any message which isn't a constant will generate garbage in the form of strings.  In most cases this isn't a problem, but it is not ideal if it's placed in a tight loop.  

We could offset the problem by adding a family of Utils functions that take a lambda String producer instead of a string itself, this would allow the message to be computed only when the error condition is triggered avoiding garbage creation.

akiezun:
let's measure. 2 reruns of HC should tell us all we need to know

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/engine/ReferenceMemorySource.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/exome/Sample.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/exome/Sample.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmer.java
davidbenjamin: Done and done

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ExcessHet.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeLikelihoodCalculator.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/MRUCachingSAMSequenceDictionary.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: Unfortunately, the library method is not defined for `int[]`s, just `double[]`s

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: I got rid of `initialize()` by simply initializing the cache as
```java
private static final double[] cache = IntStream.range(0, (int) (MAX_TOLERANCE / TABLE_STEP) + 1)
                .mapToDouble(k -> Math.log10(1.0 + Math.pow(10.0, -k * TABLE_STEP))).toArray();
```

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/genotyper/ReadLikelihoods.java
davidbenjamin: Reverted; this change isn't important to me.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: I wrote a short method in `MathUtils` to apply a lambda to an array without creating a stream.  We get a compact one-liner with the speed of pure arrays.  I benchmarked this and it gets the same performance as arrays; both are much better than streams.  Even though this new method takes in a `Function<Double, Double>` it appears the JIT compiler is smart enough to realize when to skip the boxing and unboxing.

src/main/java/org/broadinstitute/hellbender/utils/genotyper/ReadLikelihoods.java
davidbenjamin: Reverted.  I was not aware about the cost of accessing `.length`  Why is accessing a public field more expensive than a variable?  Won't the JVM cache it?

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessSettings.java
davidbenjamin: done

src/test/java/org/broadinstitute/hellbender/utils/RandomDNAUnitTest.java
davidbenjamin: Same as above --`ebeAdd` is only defined for `double[]`s.

src/main/java/org/broadinstitute/hellbender/utils/variant/writers/HomRefBlock.java
davidbenjamin: done.  Previous implementation also used a stream, so I don't think an HC run is needed, but I'll be doing one anyway to test `MathUtils::applyToArray`.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: > add a space between [] and x
Done

davidbenjamin:
I implemented @lbergelson's suggestion with a new method: `validateArg(final boolean condition, final Supplier<String> msg)` invoked as eg `Utils.validateArg(n > 0, () -> String.format("You have chosen the worst value ever, name %d,", n))`.

davidbenjamin:
I refactored some of the stream methods to address @akiezun's (well-founded, as running HaplotypeCaller showed) performance concerns.  That, along with @lbergelson's suggestion to lazily evaluate error messages, leaves HaplotypeCaller's performance unaffected.

Tests are passing locally but Travis is giving me a mysterious error:
> compileTestJavaerror: error reading /home/travis/.gradle/caches/modules-2/files-2.1/commons-httpclient/commons-httpclient/3.1/964cd74171f427720480efdec40a7c7f6e58426a/commons-httpclient-3.1.jar; error in opening zip file
error: error reading /home/travis/.gradle/caches/modules-2/files-2.1/commons-httpclient/commons-httpclient/3.1/964cd74171f427720480efdec40a7c7f6e58426a/commons-httpclient-3.1.jar; cannot read zip file

I am fully rebased onto master, including @akiezun's PR from this afternoon.  @droazen and @lbergelson do you have insights?

davidbenjamin:
@akiezun Back to you.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -1127,39 +875,68 @@ public static double normalDistribution(final double mean, final double sd, fina
...
+        double prefactor = log10MultinomialCoefficient(countSum,counts) + log10Gamma(dirichletSum) - log10Gamma(dirichletSum+countSum);
+        return prefactor + new IndexRange(0, counts.length).sum(n -> log10Gamma(counts[n] + params[n]) - log10Gamma(params[n]));
+    }
+
+    /**
+     * The following method implements Arrays.stream(array).map(func).toArray(), which is concise but performs poorly due
+     * to the overhead of creating a stream, especially with small arrays.  Thus we wrap the wordy but fast array code
+     * in the following method which permits concise Java 8 code.
+     *
+     * This method has been benchmarked and performs as well as array-only code.
```
akiezun: add a comment that this method *does not* modify the array but rather creates a new one. There may be applications where modifying the array in place would make sense as well.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -1127,39 +875,68 @@ public static double normalDistribution(final double mean, final double sd, fina
...
-            likelihood -= log10Gamma(dirichletParams[idx]);
+        return result;
+    }
+
+    /**
+     * The following method implements Arrays.stream(array).map(func).toArray(), which is concise but performs poorly due
+     * to the overhead of creating a stream, especially with small arrays.  Thus we wrap the wordy but fast array code
+     * in the following method which permits concise Java 8 code.
+     *
+     * This method has been benchmarked and performs as well as array-only code.
```
akiezun: add a comment that it mutates the array in place

src/main/java/org/broadinstitute/hellbender/utils/IndexRange.java
```diff
@@ -114,6 +115,66 @@ public void forEach(final IntConsumer lambda) {
...
+    }
+
+    /**
+     * Find the elements of this range for which an int -> boolean predicate is true
+     *
+     * @param predicate the int -> boolean predicate
+     * @return
+     */
+    public int[] filter(final IntPredicate predicate) {
+        final int[] result = new int[size()];
```
akiezun: null check for predictate

akiezun: hmm, this seems like a potentially huge array allocation for what may turn out to be a small filtered result. Using an ArrayList (which grows as needed) would be better I think.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -878,7 +890,53 @@ public static double dirichletMultinomial(double[] params, int[] counts) {
...
+     */
+    public static double[] applyToArrayInPlace(final double[] array, final DoubleUnaryOperator func) {
```
akiezun: add a comment that it mutates the array in place

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -878,7 +890,53 @@ public static double dirichletMultinomial(double[] params, int[] counts) {
...
+        for (int m = 0; m < array.length; m++) {
+            array[m] = func.applyAsDouble(array[m]);
+        }
+        return array;
+    }
+
+    public static boolean allMatch(final double[] array, final DoublePredicate pred) {
```
akiezun: add a comment to describe what this method does

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -878,7 +890,53 @@ public static double dirichletMultinomial(double[] params, int[] counts) {
...
+        for (final double x : array) {
```
akiezun: null checks for args

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -878,7 +890,53 @@ public static double dirichletMultinomial(double[] params, int[] counts) {
...
+            if (!pred.test(x)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public static boolean allMatch(final int[] array, final IntPredicate pred) {
```
akiezun: javadoc + null checks

src/main/java/org/broadinstitute/hellbender/utils/IndexRange.java
```diff
@@ -114,6 +115,66 @@ public void forEach(final IntConsumer lambda) {
         }
     }
 
+    /**
+     * Apply an int -> double function to this range, producing a double[]
+     *
+     * @param lambda the int -> double function
+     */
+    public double[] mapToDouble(final IntToDoubleFunction lambda) {
```
akiezun: all new methods in `IndexRange` need unit tests in `IndexRangeUnitTest`

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -878,7 +890,53 @@ public static double dirichletMultinomial(double[] params, int[] counts) {
...
+    }
+
+    /**
+     * The following method implements Arrays.stream(array).map(func).toArray(), which is concise but performs poorly due
+     * to the overhead of creating a stream, especially with small arrays.  Thus we wrap the wordy but fast array code
+     * in the following method which permits concise Java 8 code.
+     *
+     * This method has been benchmarked and performs as well as array-only code.
+     */
+    public static double[] applyToArray(final double[] array, final DoubleUnaryOperator func) {
```
akiezun: applyToArray  needs unit tests

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
akiezun: applyToArrayInPlace needs unit tests

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
akiezun: add unit tests for allMatch

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -589,6 +605,25 @@ public static void containsNoNull(final Collection<?> c, final String message) {
...
     /**
+     * Checks that the collection does not contain a duplicate value (throws an {@link IllegalArgumentException} if it does).
+     * The implementation creates a {@link Set} as an intermediate step or detecting duplicates and returns this Set because
+     * it is sometimes useful to do so.
+     *
+     * @param c collection
+     * @param message A message to emit in case of error, in addition to reporting the first duplicate value found.
+     * @throws IllegalArgumentException if a {@code o == null}
+     */
+    public static <E> Set<E> containsNoDuplicates(final Collection<E> c, final String message) {
```
akiezun: add unit tests for containsNoDuplicates. Also, it's a strange name - does not suggest that it will blow up in my face or return a set, rather the name suggests that it will return a boolean and not blow up.

akiezun:
looking good. a bunch of missing unit tests and comment and small things. Merge when done.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/IndexRange.java
davidbenjamin: done and done

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/IndexRange.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: done and done

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: done and done and done

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
davidbenjamin: done

davidbernick:

PR #1984, opened 2016-07-07
updating to use jbwa from maven central

lbergelson:


droazen:
:+1: merge when ready

PR #1991, opened 2016-07-11
Bump GKL version to 0.1.2

### LINKED ISSUE #1985: Using GATK as a dependency blows up after gkl is included
Using the latest `master` version of GATK4 as dependency in gradle (using [jitpack](https://jitpack.io/)) to build a shadow jar generetes the following error when a custom walker run using `java -jar shadowJar.jar CustomTool`, but also with public tools. Only if `--use_jdk_deflater true` is provided, it works. 

The log is the following (there is no other log):

```
14:57:19.102 INFO  IntelGKLUtils - Trying to load Intel GKL library from:
	jar:file:/Users/daniel/workspaces/gatk4test/build/libs/shadowJar-0.0.1-SNAPSHOT-all.jar!/com/intel/gkl/native/libIntelGKL.dylib
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGILL (0x4) at pc=0x0000000128c014d0, pid=31197, tid=5891
#
# JRE version: Java(TM) SE Runtime Environment (8.0_60-b27) (build 1.8.0_60-b27)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.60-b23 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# C  [libIntelGKL8818190486223479934.dylib+0xe4d0]  _ZN7ContextIfEC2Ev+0x30
#
# Core dump written. Default location: /cores/core or core.31197
#
# An error report file with more information is saved as:
# /Users/daniel/workspaces/gatk4test/hs_err_pid31197.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
Abort trap: 6 (core dumped)
```

To fix it, I tried by excluding `com.intel.gkl` from GATK and add it as a dependency to my program, but it blows up anyway. In addition, I tried a sample program to load the PairHMM fastest implementation by `PairHMM.Implementation.FASTEST_AVAILABLE.makeNewHMM()`, and it also blows up. If I remove completely the dependency in my shadow jar, the command line blows up because the gkl `IntelDeflaterFactory` is not found.

I guess that the error in the library is GKL-related, but in the case of the GATK framework I would like to have a way of using the library without assuming that the final user will have support for the native code or not. Could this be done? I prefer not to remove the faster code by intel because I know that some users will benefit from it.

Just in case it is needed, my system is a Mac OS X (10.11.5) with Darwin Kernel Version 15.5.0 (root:xnu-3248.50.21~8/RELEASE_X86_64 x86_64).

Thank you very much in advance.

gspowley:
Checks for AVX support on the CPU. Fixes #1985

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -172,7 +172,7 @@ dependencies {
     // Dependency change for including MLLib
     compile('com.esotericsoftware:reflectasm:1.10.0:shaded')
 
-    compile('com.intel.gkl:gkl:0.1.1') {
+    compile('com.intel.gkl:gkl:0.1.2') {
         exclude module: 'htsjdk'
     }
     compile 'com.github.lindenb:jbwa:1.0.0'


akiezun:
@magicDGS can you try this?

magicDGS:
@akiezun, I tested the branch in my computer and now I can run the shadow jar. Thanks a lot, @gspowley.

PR #1992, opened 2016-07-11
backport 1 method from protected and renamed 1 file

akiezun:
@lbergelson can you review and make a snapshot with those? It will allow me to cleanup https://github.com/broadinstitute/gatk-protected/pull/610 and delete 2 more classes from there


droazen:
(snapshots of master are made automatically)

akiezun:
do they include all the code that we need? (maybe they do now that we've
moved to using gkl)?

droazen:
Yes, they should, since we're no longer building the native libraries from source!

lbergelson:
So the continuous snapshots broke for a mysterious reason.  https://github.com/broadinstitute/gatk/issues/1993

akiezun:
@samuelklee can you review? this is a change you made in protected and i backported here

samuelklee:
:+1:

PR #1997, opened 2016-07-13
Lower error rate in SGA overlap and better error handling

SHuang-Broad:
@cwhelan would you please review?
Thanks!


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -77,7 +77,7 @@
     // subject to future changes as we see more test cases
     @VisibleForTesting static final int MIN_OVERLAP_IN_FILTER_OVERLAP_ASSEMBLE = 55;
     @VisibleForTesting static final int FILTER_STEP_KMER_FREQUENCY_THREASHOLD = 1;
-    @VisibleForTesting static final double OVERLAP_STEP_ERROR_RATE = 0.05;
+    @VisibleForTesting static final double OVERLAP_STEP_ERROR_RATE = 0.02;
```
cwhelan: While we're changing parameters we might as well turn these into optional command line arguments with these as defaults.

cwhelan:
Now that the rest of the pipeline is in better shape I wanted to try running this to see if it would improve our runtimes and/or results.

It does seem to speed things up on the current data. However, I had one task failure that I don't get with what's in the master branch. Here's the output for that task:

```
cwhelan@svdev-1-m:~$ hdfs dfs -cat hdfs://svdev-1-m:8020/user/cwhelan/outs_sh_sga_para_and_error_handle/NA12878_PCR-_30X/assembly_1/part-00000
13913
[sga preprocess	 successfully executed.
 Module stdout message: 

 Module stderr message: 

, sga index	 successfully executed.
 Module stdout message: 

 Module stderr message: 

, sga filter	 successfully executed.
 Module stdout message: 

 Module stderr message: 

, sga fm-merge	 returned with non-zero exit status, see stderr message for detailed return status.
 Module stdout message: 
is empty
 Module stderr message: 
Program returned exit status 134
/usr/local/bin/sga fm-merge --min-overlap 55 assembly13913.pp.filter.pass.fa


]
```

The command line used was:

```
./gatk-launch RunSGAViaProcessBuilderOnSpark \
	--fullPathToSGA /usr/local/bin/sga \
	--inDir hdfs://svdev-1-m:8020/user/cwhelan/outs/NA12878_PCR-_30X/fastq \
	--outDirPrefix hdfs://svdev-1-m:8020/user/cwhelan/outs_sh_sga_para_and_error_handle/NA12878_PCR-_30X/assembly \
	--subStringToStrip assembly \
	-- \
	--sparkRunner GCS \
	--cluster svdev-1 \
	--driver-memory 30G \
	--executor-memory 8G \
	--conf spark.yarn.executor.memoryOverhead=5000
```

cwhelan:
Never mind, I see now that we're only noticing the failure because of the new error handling.

SHuang-Broad:
@cwhelan , updated with the forced repartition to avoid the partition problem we saw.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -75,20 +75,52 @@
...
+              shortName = "rmTransEdges",
+              fullName  = "transitiveReduction",
+              optional  = true)
+    public static final boolean TURN_ON_TRANSITIVE_REDUCTION = false;
 
 
     // for developer performance debugging use
     private static final DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
     private static final Logger logger = LogManager.getLogger(RunSGAViaProcessBuilderOnSpark.class);
+    private static final int files_per_partition = 5; // performance tuning
```
cwhelan: I think in our usual style we make static final constants all caps: FILES_PER_PARTITION

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -136,17 +168,19 @@ private static void validateAndSaveResults(final JavaPairRDD<Long, SGAAssemblyRe
         final JavaPairRDD<Long, SGAAssemblyResult> cachedResults = results.cache(); // cache because Spark doesn't have an efficient RDD.split(predicate) yet
 
         // save fasta file contents or failure message
-        final JavaPairRDD<Long, SGAAssemblyResult> success = cachedResults.filter(entry -> entry._2().assembledContigs!=null);
-        final JavaPairRDD<Long, SGAAssemblyResult> failure = cachedResults.filter(entry -> entry._2().assembledContigs==null);
+        final JavaPairRDD<Long, SGAAssemblyResult> success = cachedResults.filter(entry -> entry._2().assembledContigs!=null).cache();
+        final JavaPairRDD<Long, SGAAssemblyResult> failure = cachedResults.filter(entry -> entry._2().assembledContigs==null).cache();
```
cwhelan: Why are you caching these two RDDs? You already cached the results RDD.

cwhelan:
Just a couple of minor comments/questions, then this should be fine to merge.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
SHuang-Broad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
SHuang-Broad: The cache call for success is un-reasonable, you are right.
The cache call for failure is to make the `count()` call several lines down and the `saveAsTextFile()` beneath that trigger the filtering only once. Am I misunderstanding something?

cwhelan: Ah, I see. The isEmpty() calls also probably cause the filtering to be triggered, so you can leave the caching. Ted pointed out that we should probably call `unpersist` on cached RDDs after we are done with them -- can you try adding that?

SHuang-Broad: Definitely! Just done.

cwhelan:
Looks good; once you test out the unpersist calls on the cluster go ahead and rebase and merge.

SHuang-Broad:
I've added some non-trivial (code-wise, logical wise very simple) code to single out slow assemblies more obviously. That is, generating another txt file collecting the runtime for those slow ones.
Example here 
/user/shuang/experiments/NA12878_PCR-_30X
/user/shuang/experiments/NA12878_PCR-_30X/assembly_betterLogging_longOnes

cwhelan:
As discussed in person I think it would be a little more helpful to provide an option to either log all of the times or none of them, rather than setting a threshold.

SHuang-Broad:
@cwhelan , just updated with all runtime logged. Same location, with all runtime in one file.

SHuang-Broad:
closing because we probably no longer needs this.

PR #1998, opened 2016-07-13
a bit better asserts to check SVD

akiezun:
added some better svd checking code 

@davidbenjamin this is pretty much straight from your comments in https://github.com/broadinstitute/gatk-protected/pull/609
can you review?

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/utils/svd/SingularValueDecomposerUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/svd/SingularValueDecomposerUnitTest.java
@@ -1,42 +1,33 @@
 package org.broadinstitute.hellbender.utils.svd;
 
-import htsjdk.samtools.util.Log;
 import org.apache.commons.math3.linear.Array2DRowRealMatrix;
 import org.apache.commons.math3.linear.MatrixUtils;
 import org.apache.commons.math3.linear.QRDecomposition;
 import org.apache.commons.math3.linear.RealMatrix;
 import org.broadinstitute.hellbender.engine.spark.SparkContextFactory;
-import org.broadinstitute.hellbender.utils.LoggingUtils;
 import org.broadinstitute.hellbender.utils.test.BaseTest;
 import org.testng.Assert;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
+
+import java.util.stream.IntStream;
 
 
 public final class SingularValueDecomposerUnitTest extends BaseTest {
 
     @DataProvider(name="makeSVDers")
     private Object[][] makeSVDers(){
+        final RealMatrix m1  = new Array2DRowRealMatrix(new double[][]{{ 2.0, 0.0 }, { 0.0, -3.0 }, { 0.0, 0.0 }});
         return new Object[][]{
-                {new ApacheSingularValueDecomposer()},
-                {new OjAlgoSingularValueDecomposer()},
-                {new SparkSingularValueDecomposer(SparkContextFactory.getTestSparkContext())}
+                {new ApacheSingularValueDecomposer(), m1},
+                {new OjAlgoSingularValueDecomposer(), m1},
+                {new SparkSingularValueDecomposer(SparkContextFactory.getTestSparkContext()), m1},
         };
     }
     @Test(dataProvider = "makeSVDers")
-    public void testBasicFunction(final SingularValueDecomposer svder){
-        LoggingUtils.setLoggingLevel(Log.LogLevel.INFO);
-        final Array2DRowRealMatrix m = new Array2DRowRealMatrix(new double[][]{{ 2.0, 0.0 }, { 0.0, -3.0 }, { 0.0, 0.0 }});
-
-        final double[] expectedSingularValues = {3.0, 2.0};
-
+    public void testBasicFunction(final SingularValueDecomposer svder, final RealMatrix m){
         final SVD svd = svder.createSVD(m);
-        Assert.assertEquals(svd.getU().multiply(MatrixUtils.createRealDiagonalMatrix(svd.getSingularValues())).multiply(svd.getV().transpose()), m);
-        Assert.assertTrue(isUnitaryMatrix(svd.getV()), "V is unitary");
-        Assert.assertTrue(isUnitaryMatrix(svd.getU()), "U is unitary");
-        Assert.assertEquals(svd.getV().transpose().multiply(svd.getV()), MatrixUtils.createRealIdentityMatrix(2), "V is unitary");
-        Assert.assertEquals(svd.getU().transpose().multiply(svd.getU()), MatrixUtils.createRealIdentityMatrix(2), "U is unitary");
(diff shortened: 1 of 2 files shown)


src/test/java/org/broadinstitute/hellbender/utils/svd/SingularValueDecomposerUnitTest.java
```diff
@@ -31,11 +31,7 @@ public void testBasicFunction(final SingularValueDecomposer svder){
         final double[] expectedSingularValues = {3.0, 2.0};
 
         final SVD svd = svder.createSVD(m);
-        Assert.assertEquals(svd.getU().multiply(MatrixUtils.createRealDiagonalMatrix(svd.getSingularValues())).multiply(svd.getV().transpose()), m);
-        Assert.assertTrue(isUnitaryMatrix(svd.getV()), "V is unitary");
-        Assert.assertTrue(isUnitaryMatrix(svd.getU()), "U is unitary");
-        Assert.assertEquals(svd.getV().transpose().multiply(svd.getV()), MatrixUtils.createRealIdentityMatrix(2), "V is unitary");
-        Assert.assertEquals(svd.getU().transpose().multiply(svd.getU()), MatrixUtils.createRealIdentityMatrix(2), "U is unitary");
+        assertSVD(svd, m);
```
davidbenjamin: You can get rid of the check on expected singular values.  The checks in `assertSVD` are necessary and sufficient.

src/test/java/org/broadinstitute/hellbender/utils/svd/SingularValueDecomposerUnitTest.java
```diff
@@ -56,5 +52,23 @@ public static boolean isUnitaryMatrix(final RealMatrix m){
...
         return true;
     }
+
+    public void assertSVD(final SVD svd, final RealMatrix m) {
+        final RealMatrix U = svd.getU();
+        final RealMatrix V = svd.getV();
+        final double[] s = svd.getSingularValues();
+
+        final RealMatrix S = MatrixUtils.createRealDiagonalMatrix(svd.getSingularValues());
+        final RealMatrix SPrime = MatrixUtils.inverse(S); //it's diagonal so it's ok to just invert
```
davidbenjamin: `S` is only square if `m` is square; otherwise `S` is has the singular values on its diagonal and is padded with zeroes to become rectangular with the right dimensions.  If `m` has dimension q x r then U is q x q, V is r x r, `S` is q x r, and there are min(q,r) singular values.

This test as currently written will break if `m` is not square (hence the more cumbersome code I sketched in the other ticket).  This case needs to be tested e.g. with a random 2 x 3 matrix.

akiezun: good point. will fix

src/test/java/org/broadinstitute/hellbender/utils/svd/SingularValueDecomposerUnitTest.java
davidbenjamin: Not your code, but while you're at it `isUnitaryMatrix` is needlessly complicated.  Unitary matrices, in particular the U and V of an SVD, are by definition square.  Thus the necessary check is simply that the transpose is the inverse i.e. `m.multiply(m.transpose()) == m.transpose().multiply(m) == IdentityMatrix`.

akiezun: done

src/test/java/org/broadinstitute/hellbender/utils/svd/SingularValueDecomposerUnitTest.java
```diff
@@ -56,5 +52,23 @@ public static boolean isUnitaryMatrix(final RealMatrix m){
...
+
+        Assert.assertEquals(U.multiply(S).multiply(V.transpose()), m);
+        Assert.assertEquals(V.transpose().multiply(V), MatrixUtils.createRealIdentityMatrix(2), "V is unitary");
+        Assert.assertEquals(U.transpose().multiply(U), MatrixUtils.createRealIdentityMatrix(2), "U is unitary");
+
+        Assert.assertTrue(isUnitaryMatrix(U), "matrix U");
+        Assert.assertTrue(isUnitaryMatrix(V), "matrix V");
+        Assert.assertEquals(U.multiply(S).multiply(V.transpose()), m);
+        Assert.assertEquals(V.multiply(SPrime).multiply(U.transpose()), svd.getPinv());
```
davidbenjamin: While correct, the check on the pseudoinverse that I proposed is more complicated than necessary (my apologies).  Instead of verifying the formula for calculating the pseudoinverse from the SVD, we can simply verify the defining properties of the pseudoinverse (Wikipedia: Moore–Penrose_pseudoinverse).  
```java
private void assertPseudoinverse(final RealMatrix m, final RealMatrix p) {
  Assert.assertEquals(m.multiply(p).multiply(m), m);
  Assert.assertEquals(p.multiply(m).multiply(p), p);
  Assert.assertEquals(m.multiply(p).transpose(), m.multiply(p));
  Assert.assertEquals(p.multiply(m).transpose(), p.multiply(m));
}
```
If you like the existing check better it's fine to keep; both are correct.

davidbenjamin:
Back to @akiezun.

src/test/java/org/broadinstitute/hellbender/utils/svd/SingularValueDecomposerUnitTest.java
akiezun: hmm, that may be true in some, but not in the current implementation of SVD in GATK. U may not be square, as per doc for apache http://commons.apache.org/proper/commons-math/apidocs/org/apache/commons/math3/linear/SingularValueDecomposition.html
and mllib https://spark.apache.org/docs/1.2.0/mllib-dimensionality-reduction.html

U is not square in this unit test in this class
M={{2.0,0.0},{0.0,-3.0},{0.0,0.0}}

U = {{0.0,-1.0},{-1.0,0.0},{0.0,0.0}}
S = {{3.0,0.0},{0.0,2.0}}
V = {{0.0,-1.0},{1.0,0.0}}

davidbenjamin: Well, that's interesting.  Not exactly standard on Apache Commons' part, but nothing we can do about it.  Add a comment stating that Apache has an untraditional SVD and a still more untraditional understanding of what a unitary matrix is.

akiezun: done

src/test/java/org/broadinstitute/hellbender/utils/svd/SingularValueDecomposerUnitTest.java
akiezun: i added both for good measure. it's test code, performance does not matter as much

Issue #2000, opened 2016-07-13
New logical name feature in TaggedInputFileArgument breaks hdfs urls that contain port numbers.


tedsharpe:
Trying to use an url of the form hdfs://server:port/path/to/file as an input to a spark program.  This used to work.  Now we're confused by the possibility that we're declaring a logical name.

droazen:
I'll work on a quick fix for this, as it's a fairly serious regression.

akiezun:
maybe we could use a different character than `:` to separate logical names. Maybe `+` (?) Or hardwire a bunch of protocols like `hdfs`, `file`, `gs`, etc?

akiezun:
good idea too

droazen:
I think that won't actually work @lbergelson, since we also have to handle the colon for the port numbers.

droazen:
According to https://tools.ietf.org/html/rfc3986#section-2, there are many disallowed characters in URIs that we could use as a delimiter for logical name, like |. This would be a syntax change vs. GATK3, however.

akiezun:
but the logical name is always first so it can't be confused with anything else

akiezun:
syntax is already different `-I:tumor bam` vs `-I tumor:bam`

droazen:
yes, that's true

lbergelson:
@droazen Maybe I'm misunderstanding, but we'd just split up until a uri specifier?

droazen:
@lbergelson The regex you used in the TileDB branch would ignore the colon for the scheme, but split on the colon for the port number (if there is one), wouldn't it?

lbergelson:
Might be time to abandon regex and write a few lines of logic...

droazen:
I'll have a fix by tomorrow

droazen:
This is more serious than just URIs with port numbers, by the way -- I'm pretty sure all URIs are broken on spark right now.

tedsharpe:
Yes.  That's true.  All Spark URIs.  I didn't provide an accurate description of the problem.

droazen:
We are going to temporarily revert the problematic change until a definitive fix can be developed (https://github.com/broadinstitute/gatk/pull/2007).

droazen:
Closing -- this will be solved by hooking up the new barclay argument tagging feature: https://github.com/broadinstitute/gatk/issues/2473

PR #2007, opened 2016-07-14
Revert "symbolic names for sources of reads"

droazen:
Temporarily backing this change out while we work on a definitive fix for https://github.com/broadinstitute/gatk/issues/2000, to allow Spark tools to work with URI inputs once again. 

I'll open a separate PR to restore this functionality in a way that can co-exist with URI inputs.


davidbernick:

droazen:
@akiezun please review

Issue #2009, opened 2016-07-15
Migrate from the v1beta2 Google Genomics API to the v1 API


droazen:
Request from Google. Migration instructions are here: http://googlegenomics.readthedocs.io/en/latest/migrating_v1beta2_to_v1.html

melissachang:
Hi, we (Google) are still seeing v1beta2 requests from Broad.

genomics.referencesets.search
genomics.references.search
genomics.references.bases.list

https://github.com/broadinstitute/gatk/search?utf8=%E2%9C%93&q=v1beta2

Can this issue be reopened?

jean-philippe-martin:
ReferenceAPISource talks to Genomics API v1beta2, and if nothing else ReferenceAPISourceUnitTest will exercise this code.

https://github.com/broadinstitute/gatk/search?utf8=%E2%9C%93&q=ReferenceAPISource&type=Code

droazen:
@jean-philippe-martin A (hopefully) quick one for you when you have some spare time.

droazen:
Closing -- we've removed support for the legacy Google Genomics API.

PR #2011, opened 2016-07-18
updating google-api-services-genomics off of the beta api

lbergelson:
We got an email complaining that we were the only users in the world still using the beta API.

This updates to the most recent version, they changed the api from `List<String> -> List<Object>`, but the documentation still says it has to be a `String`, so I've just added some explicit casts.  

Fix for #2009

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceAPISource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceAPISource.java
@@ -197,7 +197,7 @@ public class ReferenceAPISource implements ReferenceSource, Serializable {
               if (received.length < interval.size()) {
                   final List<byte[]> blobs = new ArrayList<>();
                   blobs.add(received);
-                  while (result.getNextPageToken() != null) {
+                  while (! result.getNextPageToken().isEmpty()) {
                       listRequest.setPageToken(result.getNextPageToken());
                       result = listRequest.execute();
                       blobs.add(result.getSequence().getBytes());
--- a/src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceAPISource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceAPISource.java
@@ -197,7 +197,7 @@ public class ReferenceAPISource implements ReferenceSource, Serializable {
               if (received.length < interval.size()) {
                   final List<byte[]> blobs = new ArrayList<>();
                   blobs.add(received);
-                  while (! result.getNextPageToken().isEmpty()) {
+                  while (result.getNextPageToken() != null  && !result.getNextPageToken().isEmpty()) {
                       listRequest.setPageToken(result.getNextPageToken());
                       result = listRequest.execute();
                       blobs.add(result.getSequence().getBytes());
--- a/src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
@@ -500,6 +500,7 @@ public final class GoogleGenomicsReadToGATKReadAdapter implements GATKRead, Seri
         return genomicsRead.getInfo() != null && genomicsRead.getInfo().containsKey(attributeName);
     }
 
+    @SuppressWarnings("unchecked")
     private String getRawAttributeValue( final String attributeName, final String targetType ) {
         ReadUtils.assertAttributeNameIsLegal(attributeName);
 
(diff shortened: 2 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceAPISource.java
```diff
@@ -197,7 +197,7 @@ public ReferenceBases getReferenceBases(final PipelineOptions pipelineOptions, f
               if (received.length < interval.size()) {
                   final List<byte[]> blobs = new ArrayList<>();
                   blobs.add(received);
-                  while (result.getNextPageToken() != null) {
+                  while (! result.getNextPageToken().isEmpty()) {
```
akiezun: this https://developers.google.com/resources/api-libraries/documentation/genomics/v1/java/latest/com/google/api/services/genomics/model/ListBasesResponse.html says that getNextPageToken may return null

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
```diff
@@ -517,7 +518,7 @@ private String getRawAttributeValue( final String attributeName, final String ta
             throw new GATKException.ReadAttributeTypeMismatch(attributeName, targetType);
         }
 
-        return rawValue.get(0);
+        return (String)rawValue.get(0);
```
akiezun: add a comment why this cast will succeed (ie link to https://developers.google.com/resources/api-libraries/documentation/genomics/v1/java/latest/com/google/api/services/genomics/model/Read.html#setInfo(java.util.Map). The API is silly (`List<Object>` storing `String`) so we should put the blame for ugly code where it belongs.

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
```diff
@@ -561,7 +562,7 @@ public void setAttribute( final String attributeName, final Integer attributeVal
             return;
         }
 
-        final List<String> encodedValue = Arrays.asList(attributeValue.toString());
+        final List<Object> encodedValue = Arrays.asList(attributeValue.toString());
```
akiezun: this is a singleton list so `Collections.singletonList` is more conventional

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
```diff
@@ -574,7 +575,7 @@ public void setAttribute( final String attributeName, final String attributeValu
             return;
         }
 
-        final List<String> encodedValue = Arrays.asList(attributeValue);
+        final List<Object> encodedValue = Arrays.asList(attributeValue);
```
akiezun: this is a singleton list so Collections.singletonList is more conventional

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
```diff
@@ -587,7 +588,7 @@ public void setAttribute( final String attributeName, final byte[] attributeValu
             return;
         }
 
-        final List<String> encodedValue = Arrays.asList(new String(attributeValue));
+        final List<Object> encodedValue = Arrays.asList(new String(attributeValue));
```
akiezun: this is a singleton list so `Collections.singletonList` is more conventional

akiezun:
looks fine. a few small comments then ok to merge

src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceAPISource.java
lbergelson: oh, you're right...  updated

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
lbergelson: added

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
lbergelson: done

PR #2021, opened 2016-07-19
Changed SplitNCigarReads to produce softclipped supplementary reads

### LINKED ISSUE #1394: Ensure SplitNCigarReads produces valid bam files
Spin out of #1378 

The GATK3 implementation of this tool generates invalid bam files (because the mate pair info gets all screwed up). If possible to fix that within the tool itself that would be fabulous. Otherwise @yfarjoun has a Palantir ticket to look into a post-hoc fix with FixMateInformation. But it would be nicer to generate valid bams in the first place.

jamesemery:
resolves #1394


jamesemery:
This will currently fail all of the SplitNCigarReadsIntegrationTests.java tests because I have yet to validate that the approach is right for this problem and that the output is reasonable before I go about recreating the tests. @meganshand

meganshand:
Thank you @jamesemery! As a first pass at the integration tests we should make sure that the output files actually validate with Picard's ValidateSamFile tool. I can help with that if you have the files. Also we should look at some of the output manually. Can you point me to a location that has the new output files?

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -2,33 +2,24 @@
...
-import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
-import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.exceptions.UserException;
-import org.broadinstitute.hellbender.tools.walkers.rnaseq.OverhangFixingManager;
 import org.broadinstitute.hellbender.transformers.NDNCigarReadTransformer;
 import org.broadinstitute.hellbender.transformers.ReadTransformer;
 import org.broadinstitute.hellbender.utils.GenomeLocParser;
+import org.broadinstitute.hellbender.utils.clipping.ClippingOp;
+import org.broadinstitute.hellbender.utils.clipping.ClippingRepresentation;
```
meganshand: Are these two imports being used?

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,32 +127,48 @@ public void closeTool() {
...
+        if (splitReads.size() < 1) {
             manager.addRead(read);
+            return read;
         }
         //add the last section of the read: from the last N to the the end of the read
         // (it will be done for all the usual cigar string that does not end with N)
-        else if (firstCigarIndex < numCigarElements) {
-            manager.addRead(splitReadBasedOnCigar(read, firstCigarIndex, numCigarElements, null));
+        else if ((firstCigarIndex < numCigarElements) && sectionHasMatch) {
+            splitReads.add(splitReadBasedOnCigar(read, firstCigarIndex, numCigarElements, null));
```
meganshand: It looks like both times you call splitReadBasedOnCigar you pass null in as the manager. Can you remove it from the function itself? That way you can delete lines 205-210 (since it's never touched).

src/main/java/org/broadinstitute/hellbender/utils/clipping/ClippingOp.java
```diff
@@ -39,6 +39,11 @@ public int getLength() {
      * @param originalRead the read to be clipped
      */
     public GATKRead apply(final ClippingRepresentation algorithm, final GATKRead originalRead) {
+        return apply(algorithm, originalRead, true);
+    }
+
+    // Warning, this method is intended for testing purposes
+    GATKRead apply(final ClippingRepresentation algorithm, final GATKRead originalRead, final boolean runAsserts) {
```
meganshand: Since it looks like the only test that uses this is currently turned off, could we wait to make all of these runAssert changes until we want to include that test?

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
```diff
@@ -488,10 +492,122 @@ protected GATKRead hardClipByReferenceCoordinates(final int refStart, final int
...
+
+
+
+    /**
+     * Soft clip the read to the variable region (from refStart to refStop) processing also the clipped bases
+     *
+     * @param read     the read to be clipped
+     * @param refStart the beginning of the variant region (inclusive)
+     * @param refStop  the end of the variant region (inclusive)
+     * @return the read hard clipped to the variant region
```
meganshand: hard -> soft

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
```diff
@@ -488,10 +492,122 @@ protected GATKRead hardClipByReferenceCoordinates(final int refStart, final int
...
+    }
+
+    public static GATKRead softClipBothEndsByReferenceCoordinates(final GATKRead read, final int left, final int right) {
+        return (new ReadClipper(read)).softClipBothEndsByReferenceCoordinates(left, right);
+    }
+
+    /**
+     * Generic functionality to soft clip a read (is analogous to hardClipByReferenceCoordinates())
+     *
+     * Note, it REQUIRES you to give the directionality of your hard clip (i.e. whether you're clipping the
```
meganshand: This one should be soft too, right?

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
```diff
@@ -488,10 +492,122 @@ protected GATKRead hardClipByReferenceCoordinates(final int refStart, final int
...
+     * left of right tail) by specifying either refStart < 0 or refStop < 0.
+     *
+     * @param refStart  first base to clip (inclusive)
+     * @param refStop last base to clip (inclusive)
+     * @return a new read, without the clipped bases
```
meganshand: With soft clipped bases?

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
```diff
@@ -488,10 +492,122 @@ protected GATKRead hardClipByReferenceCoordinates(final int refStart, final int
...
+
+    protected GATKRead softClipByReferenceCoordinates(final int refStart, final int refStop, boolean runAsserts) {
+        if (read.isEmpty()) {
+            return read;
+        }
+
+        final int start;
+        final int stop;
+
+        // Determine the read coordinate to start and stop hard clipping
```
meganshand: hard -> soft

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
```diff
@@ -1120,4 +1120,58 @@ public static boolean validateExpectedSortOrder(
...
+     * properly according to the samspec (example: SA:primaryreadname,position,strand,Cigar,mapQ,NM;...)
+     *
+     * @param primaryRead the primary read in the set of supplemental reads
+     * @param supplementalReads an arbitrarily sorted list of reads to be set as
+     *                          supplemental to eachother and the primary read
+     *
+     * @throws UserException TODO
+     */
+    public static void setReadsAsSupplemental(GATKRead primaryRead, List<GATKRead> supplementalReads) {
+        //TODO deal with the unmapped reads problem
```
meganshand: Can you remind me what the unmapped read problem is?

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReadsUnitTest.java
```diff
@@ -38,6 +42,40 @@ public TestManager( final SAMFileHeader header ) {
...
+        manager = new TestManager(header);
+        GATKRead read3 = ReadClipperTestUtils.makeReadFromCigar("1H2M2D1M2N1M2I1N1M2S1N2S1N2S");
+        SplitNCigarReads.splitNCigarRead(read3, manager);
+        splitReads = manager.getReadsInQueueForTesting();
+        Assert.assertEquals(3, splitReads.size());
+        Assert.assertEquals("1H2M2D1M10S", splitReads.get(0).read.getCigar().toString());
+        Assert.assertEquals("1H3S1M2I6S", splitReads.get(1).read.getCigar().toString());
+        Assert.assertEquals("1H6S1M5S", splitReads.get(2).read.getCigar().toString());
+    }
+
```
meganshand: Nice new tests!

src/test/java/org/broadinstitute/hellbender/utils/clipping/ReadClipperUnitTest.java
```diff
@@ -387,4 +392,60 @@ public void testRevertEntirelySoftclippedReads() {
...
+            for (int i = 0; i < readLength / 2; i++) {
+                GATKRead clippedRead = ReadClipper.softClipBothEndsByReferenceCoordinates(read, alnStart + i, alnEnd - i);
+                Assert.assertTrue(clippedRead.getStart() >= alnStart + i, String.format("Clipped alignment start is less than original read (minus %d): %s -> %s", i, read.getCigar().toString(), clippedRead.getCigar().toString()));
+                Assert.assertTrue(clippedRead.getEnd() <= alnEnd + i, String.format("Clipped alignment end is greater than original read (minus %d): %s -> %s", i, read.getCigar().toString(), clippedRead.getCigar().toString()));
+                assertUnclippedLimits(read, clippedRead);
+            }
+        }
+    }
+
+    // This test depends on issue ##### as it tests the current behavior of the clipping operation
```
meganshand: Can you include the issue number

src/test/java/org/broadinstitute/hellbender/utils/clipping/ReadClipperUnitTest.java
```diff
@@ -387,4 +392,60 @@ public void testRevertEntirelySoftclippedReads() {
...
+    @Test
+    public void testSoftClippingOpEdgeCase() {
```
meganshand: This is useful, thanks for including it.

src/test/java/org/broadinstitute/hellbender/utils/clipping/ReadClipperUnitTest.java
```diff
@@ -387,4 +392,60 @@ public void testRevertEntirelySoftclippedReads() {
...
+        ClippingOp op = new ClippingOp(0, 7);
+        clipper.addOp(op);
+        GATKRead softResult = clipper.clipRead(ClippingRepresentation.SOFTCLIP_BASES);
+        Assert.assertEquals(softResult.getCigar().toString(), "7S1M");
+    }
+
+
+//
+//    Test pending resolution of issue #
+//    @Test
```
meganshand: Instead of commenting this test out, can you switch it to `@Test(enabled = false)`

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsUnitTest.java
```diff
@@ -12,6 +12,7 @@
 import htsjdk.samtools.SAMReadGroupRecord;
 import htsjdk.samtools.reference.IndexedFastaSequenceFile;
 import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.tools.walkers.annotator.allelespecific.AS_StandardAnnotation;
```
meganshand: Don't think you need this.

src/test/java/org/broadinstitute/hellbender/utils/clipping/ReadClipperUnitTest.java
```diff
@@ -387,4 +392,60 @@ public void testRevertEntirelySoftclippedReads() {
...
+    public void testSoftClipBothEndsByReferenceCoordinates() {
+        for (Cigar cigar : cigarList) {
+            GATKRead read = ReadClipperTestUtils.makeReadFromCigar(cigar);
+            int alnStart = read.getStart();
+            int alnEnd = read.getEnd();
+            int readLength = alnStart - alnEnd;
+            for (int i = 0; i < readLength / 2; i++) {
+                GATKRead clippedRead = ReadClipper.softClipBothEndsByReferenceCoordinates(read, alnStart + i, alnEnd - i);
+                Assert.assertTrue(clippedRead.getStart() >= alnStart + i, String.format("Clipped alignment start is less than original read (minus %d): %s -> %s", i, read.getCigar().toString(), clippedRead.getCigar().toString()));
+                Assert.assertTrue(clippedRead.getEnd() <= alnEnd + i, String.format("Clipped alignment end is greater than original read (minus %d): %s -> %s", i, read.getCigar().toString(), clippedRead.getCigar().toString()));
```
meganshand: Are the comments on these two asserts the opposite of what they should be? Clipped alignment start is greater than original and clipped alignment ed is less than the original.

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsUnitTest.java
```diff
@@ -407,6 +408,31 @@ public void testGetAssignedReferenceIndex() {
...
+        Assert.assertEquals(primarySup.getAttributeAsString("SA"), "read2,10001,3,4S,200,;");
+        Assert.assertEquals(primarySup.isSupplementaryAlignment(), false);
+        Assert.assertEquals(secondarySup.getAttributeAsString("SA"), "read1,10000,2,4M,100,;");
+        Assert.assertEquals(secondarySup.isSupplementaryAlignment(), true);
+        GATKRead tertiarySup = ArtificialReadUtils.createArtificialRead(header, "read3", 3, 10003, new byte[] {}, new byte[] {}, "4D");
+        tertiarySup.setMappingQuality(200);
+        sups.add(tertiarySup);
+        ReadUtils.setReadsAsSupplemental(primarySup,sups);
+        Assert.assertEquals(sups.size(), 2);
+        Assert.assertEquals(sups.get(0).getAttributeAsString("SA"), "read1,10000,2,4M,100,;read3,10003,4,4D,200,;");
```
meganshand: These SA tags have a trailing comma before the semicolon.

meganshand:
This looks good, and we should continue to dig into the integration tests. Back to @jamesemery

src/test/java/org/broadinstitute/hellbender/utils/clipping/ReadClipperUnitTest.java
droazen: @meganshand The messages passed to `Assert` methods are only printed on assertion failure, so they always reflect the opposite of the condition you're testing :)

meganshand: Haha, good call :)

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
jamesemery: @meganshand This is true but there was lingering talk that if we wanted to add a setting to the tool that allows for hard-clipped reads then the manager would be necessary to both of those methods. I don't want to eliminate it entirely just yet.

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsUnitTest.java
```diff
@@ -413,24 +413,31 @@ public void testSetReadsAsSupplemental() {
         final SAMFileHeader header = ArtificialReadUtils.createArtificialSamHeader();
         GATKRead primarySup = ArtificialReadUtils.createArtificialRead(header, "read1", 1, 10000, new byte[] {}, new byte[] {}, "4M");
         primarySup.setMappingQuality(100);
+        primarySup.setAttribute("NM",20);
```
meganshand: This is great because we are expecting virtually all of our data to have the NM tag. If you need a bam that has those tags for the integration tests I think the K-562 should have it.

meganshand:
Looking at some output, it looks like some of the tags are no longer valid: namely NH (Number of reported alignments that contains the query in the current record) and NM (Edit distance to the reference, including ambiguous bases but excluding clipping). Would it be possible to not emit them in the first place? Or you could update them to reflect the new split reads, but I'm not sure how difficult that would be. I think once that's fixed we can take a look at the integration test output and update them.

jamesemery:
@meganshand I updated the integration tests. The only difference between versions of the tool that got overwritten by the integration tests seems to relate to overhang fixing being order-dependent. I'm going to open a ticket to farther explore this issue. Otherwise I would say it is ready for final review.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +129,69 @@ public void closeTool() {
 
 
     /**
-     * Goes through the cigar string of the read and create new reads for each consecutive non-N elements (while hard clipping the rest of the read).
-     * For example: for a read with cigar '1H2M2D1M2N1M2I1N1M2S' 3 new reads will be created with cigar strings: 1H2M2D1M, 1M2I and 1M2S
+     * Goes through the cigar string of the read and create new reads for each consecutive non-N elements (while soft clipping the rest of the read) that are supplemental to eachother.
+     * For example: for a read with cigar '1H2M2D1M2N1M2I1N1M2S' 3 new reads will be created with cigar strings: 1H2M2D1M6S, 1H3S1M2I2S and 1H6S1M1S
+     * If the read
```
meganshand: there's a trailing comment here

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +129,69 @@ public void closeTool() {
...
      *
-     * @param read     the read to split
+     * @param read     the read to split (can be null)
+     * @param emitReads   a parameter used to mock behavior for repairing mate cigar string information
```
meganshand: Can you elaborate on emitReads please?

meganshand:
@jamesemery Sorry I got distracted. I'll pick this up again tomorrow.

src/main/java/org/broadinstitute/hellbender/utils/clipping/ClippingOp.java
meganshand: Just double checking, with the new tests do you still need all these runAsserts changes?

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
```diff
@@ -488,10 +492,146 @@ protected GATKRead hardClipByReferenceCoordinates(final int refStart, final int
...
+            if (start < refStart && stop > refStop) {
+                return (new ReadClipper(read)).softClipBothEndsByReferenceCoordinates(refStart - 1, refStop + 1);
+            } else if (start < refStart) {
+                return (new ReadClipper(read)).softClipByReferenceCoordinates(-1, refStart - 1);
+            } else if (stop > refStop) {
+                return (new ReadClipper(read)).softClipByReferenceCoordinates(refStop + 1, -1);
+            }
+            return read;
+        } else {
+            return ReadUtils.emptyRead(read);
```
meganshand: Can this come with a warning message (if not an error) since we shouldn't be clipping away a whole read?

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
```diff
@@ -488,10 +492,146 @@ protected GATKRead hardClipByReferenceCoordinates(final int refStart, final int
...
+     *   Left tail goes from the beginning to the 'left' coordinate (inclusive)
+     *   Right tail goes from the 'right' coordinate (inclusive) until the end of the read
+     *
+     * @param left the coordinate of the last base to be clipped in the left tail (inclusive)
+     * @param right the coordinate of the first base to be clipped in the right tail (inclusive)
+     * @return a new read, without the clipped bases
+     */
+    private GATKRead softClipBothEndsByReferenceCoordinates(final int left, final int right) {
+        if (read.isEmpty() || left == right) {
+            return ReadUtils.emptyRead(read);
```
meganshand: same warning question here

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
```diff
@@ -488,10 +492,146 @@ protected GATKRead hardClipByReferenceCoordinates(final int refStart, final int
...
+        }
+        final GATKRead leftTailRead = softClipByReferenceCoordinates(right, -1);
+
+        // after clipping one tail, it is possible that the consequent hard clipping of adjacent deletions
```
meganshand: soft -> hard? or are you saying that adjacent deletions are hard clipped regardless of whether you're soft clipping or hard clipping?

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
```diff
@@ -488,10 +492,146 @@ protected GATKRead hardClipByReferenceCoordinates(final int refStart, final int
...
+        // make the left cut index no longer part of the read. In that case, clip the read entirely.
+        if (left > leftTailRead.getEnd()) {
+            return ReadUtils.emptyRead(read);
```
meganshand: another place for a possible warning

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReadsIntegrationTest.java
```diff
@@ -16,62 +16,63 @@
  */
 public final class SplitNCigarReadsIntegrationTest extends CommandLineProgramTest {
 
-    @Test
+    @Test (enabled = true)
```
meganshand: You can change these back to just `@Test`

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReadsUnitTest.java
```diff
@@ -60,23 +104,20 @@ public void splitReadAtN() {
                 final TestManager manager = new TestManager(header);
 
                 GATKRead read = ReadClipperTestUtils.makeReadFromCigar(cigar);
-                SplitNCigarReads.splitNCigarRead(read, manager);
+                SplitNCigarReads.splitNCigarRead(read, manager,header,true);
                 List<OverhangFixingManager.SplitRead> splitReads = manager.getReadsInQueueForTesting();
                 final int expectedReads = numOfSplits+1;
                 Assert.assertEquals(splitReads.size(), expectedReads, "wrong number of reads after split read with cigar: " + cigar + " at Ns [expected]: " + expectedReads + " [actual value]: " + splitReads.size());
-                final List<Integer> readLengths = consecutiveNonNElements(read.getCigar().getCigarElements());
+                final int readLengths = consecutiveNonNElements(read.getCigar().getCigarElements());
```
meganshand: Can you change this variable to readLength since it looks like it's not plural anymore.

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsUnitTest.java
meganshand: Can you remove this import?

meganshand:
@jamesemery Looks good, just a couple of small comments. I also asked @lbergelson to take a quick look.

jamesemery:
@lbergelson here is the new version of the tool with all of the integration tests hand-verified verses the old version of the tool. The output now consistently passes validation and there is now an option to turn on (it is off by default) the N-splitting of secondary alignments.

src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
```diff
@@ -0,0 +1,58 @@
...
+
+
+import org.broadinstitute.hellbender.engine.filters.CountingReadFilter;
+import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+import java.util.stream.StreamSupport;
+
+public abstract class TwoPassReadWalker extends ReadWalker {
```
lbergelson: Could you add a class comment here explaining what this class does.

src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
```diff
@@ -0,0 +1,58 @@
...
+                    progressMeter.update(readInterval);
+                });
+    }
+
+    @FunctionalInterface
+    private interface GATKApply{
+        void consume(GATKRead read, ReferenceContext reference, FeatureContext features);
+    }
+
+    abstract protected void firstPassApply(GATKRead read, ReferenceContext bytes, FeatureContext featureContext);
```
lbergelson: Could you add javadoc to these methods that are expected to be overridden

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -12,15 +14,22 @@
 
 import java.io.Serializable;
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
- * The class manages reads and splices and tries to apply overhang clipping when appropriate.
+ * The class manages reads and splices and tries to apply overhang clipping when appropriate. Overhangs correspond to
+ * places where a read is aligned partially over Intronic splits generated by SplitNCigarReads. The class is designed
+ * perform two passes identical passes, where in the first pass every place that the manager will end up changing the
```
lbergelson: some duplicated words here.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -149,43 +173,48 @@ public void addSplicePosition(final String contig, final int start, final int en
...
+     * Add a family of split reads to the manager
      *
-     * @param read  the read to add
+     * @param readGroup the family of reads to add to the manager (assumed be supplementary alignments to eachother)
      */
-    public void addRead(final GATKRead read) {
-        Utils.nonNull(read, "read added to manager is null, which is not allowed");
+    public void addReadGroup(final List<GATKRead> readGroup) {
+        Utils.nonNull(readGroup, "readGroup added to manager is null, which is not allowed");
+        Utils.nonEmpty(readGroup, "readGroup added to manager is empty, which is not allowed");
```
lbergelson: Utils.nonEmpty implies Utils.nonNull

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -68,15 +84,17 @@
      * @param maxMismatchesInOverhangs max number of mismatches permitted in the overhangs before requiring clipping
      * @param maxBasesInOverhangs      max number of bases permitted in the overhangs before deciding not to clip
      * @param doNotFixOverhangs        if true, don't clip overhangs at all
+     * @param processSecondaryReads    if true, allow secondary reads to
```
lbergelson: allow secondary reads to.... ? the suspense is killing me!

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -50,7 +65,8 @@
 
     // the read cache
     private static final int initialCapacity = 5000;
```
lbergelson: initialCapacity should be ALL_CAPS since it's a static constant.

lbergelson: The other values MAX_RECORDS_IN_MEMORY, etc, should probably not be uppercase since they're not really constants in the absolute sense of the word, they're just final fields.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
lbergelson: make it clear what happens to secondary reads if this is false, are they filtered or are they passed through unchanged?

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -12,15 +14,22 @@
...
+ * perform two passes identical passes, where in the first pass every place that the manager will end up changing the
+ * mate strand information is recorded to be repaired with setPredictedMateInformation() when the manager will output
+ * reads to the underlying writer. Running the activateWriting() switches the tool to emit output to the writer.
  * Important note: although for efficiency the manager does try to send reads to the underlying writer in coordinate
  * sorted order, it does NOT guarantee that it will do so in every case!  So unless there's a good reason not to,
  * methods that instantiate this manager should pass in a writer that does not assume the reads are pre-sorted.
  */
 public class OverhangFixingManager {
 
+    private Map<String, Tuple<Integer, String>> mateChangedReads;
```
lbergelson: I think this can be made final

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -149,43 +173,48 @@ public void addSplicePosition(final String contig, final int start, final int en
     }
 
     /**
-     * Add a read to the manager
+     * Add a family of split reads to the manager
      *
-     * @param read  the read to add
+     * @param readGroup the family of reads to add to the manager (assumed be supplementary alignments to eachother)
```
lbergelson: it's not totally clear what a "family" of reads is from this comment

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -149,43 +173,48 @@ public void addSplicePosition(final String contig, final int start, final int en
...
+        waitingReadGroups.add(newReadGroup);
+        waitingReads = waitingReads + newReadGroup.size();
 
-        // add the new read to the queue
-        waitingReads.add(splitRead);
     }
 
     /**
-     * Clean up the list of splices
+     * Clean up the list of splices by removing the lowest half of sequential splices
```
lbergelson: Removing half the splices seems really arbitrary.  Is there a more principled way of doing it? Is there ever a way we can tell which splices are not going to be needed again?

jamesemery: One could remove the splices every time we encounter a new contig. It already will remove the lowest half of the splices which is indexed on each contig so that action is already partially performed. I could add a check that only gets rid of each splice if it is lower than the lowest read in waitingReadGroups, but that adds extra messy dependency to the class for testing and it introduces an edge case if all the splices are not lower than the lowest read.

jamesemery:
@lbergelson I responded to the first wave of comments

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
lbergelson: I assume we have a messy edge case right now were we can potentially eliminated splices that are still in range?

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -1,45 +1,33 @@
...
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.util.stream.StreamSupport;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 
-import static org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions.*;
+import static org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions.OUTPUT_LONG_NAME;
```
lbergelson: In general we don't use static imports in GATK4 because they can cause confusion.   Statically importing StandardArgumentDefinitions is a much debated exception.  Probably better to just not use any static imports though for consistency.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -68,35 +84,41 @@
...
+    public final int getNReadsInQueue() { return waitingReads; }
 
     /**
      * For testing purposes only
      *
      * @return the list of reads currently in the queue
      */
-    public List<SplitRead> getReadsInQueueForTesting() {
-        return new ArrayList<>(waitingReads);
+    public List<List<SplitRead>> getReadsInQueueForTesting() {
```
lbergelson: Can this be made package protected instead of public?

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +143,76 @@ public void closeTool() {
...
+
+        // If it is a secondary alignment, repair its mate-information (assuming its mate was primary) and pass
+        if ( !secondaryAlignments &&  read.isSecondaryAlignment()){
+            manager.addReadGroup(Collections.singletonList(read));
+            return read;
+        }
 
         int firstCigarIndex = 0;
         for ( int i = 0; i < numCigarElements; i++ ) {
             final CigarElement cigarElement = read.getCigar().getCigarElement(i);
```
lbergelson: Instead of using `read.getCigar().getCigarElement(i)` it's better to use `read.getCigarElement(i).  `getCigar` performs a copy of the entire cigar every time which is expensive.  We added the direct getCigarElement recently to avoid that.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +143,76 @@ public void closeTool() {
...
      */
-    public static GATKRead splitNCigarRead(final GATKRead read, OverhangFixingManager manager) {
+    public static GATKRead splitNCigarRead(final GATKRead read, OverhangFixingManager manager, boolean emitReads, SAMFileHeader header, boolean secondaryAlignments) {
         final int numCigarElements = read.numCigarElements();
+        List<GATKRead> splitReads = new ArrayList<>(2);
+        boolean sectionHasMatch = false;
+
+        // Run the tool on dummy mate read to determine what the mate cigar will be upon completion, if manager has a prediction then dont repair
+        if (emitReads && read.hasAttribute("MC")) {
+            read.setAttribute("MC", splitNCigarRead(ArtificialReadUtils.createArtificialRead(header ,TextCigarCodec.decode(read.getAttributeAsString("MC"))), manager, false, header, secondaryAlignments).getCigar().toString());
```
lbergelson: Could you pull out the call to splitNCigarRead  call and and give it's resutl it's own temporary variable for clarity?  This line is just very long and parenthesisful.  

Also, instead of calling `toString` on the cigar it's better to explicitly call `TextCigarCodec.encode()`.  Even though `toString` in this case is just a call through to `encode`, it's bad practice to have code correctness depend on the result of a `toString`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +143,76 @@ public void closeTool() {
...
+     * @param emitReads   a parameter used to mock behavior for repairing mate cigar string information
      */
-    public static GATKRead splitNCigarRead(final GATKRead read, OverhangFixingManager manager) {
+    public static GATKRead splitNCigarRead(final GATKRead read, OverhangFixingManager manager, boolean emitReads, SAMFileHeader header, boolean secondaryAlignments) {
         final int numCigarElements = read.numCigarElements();
+        List<GATKRead> splitReads = new ArrayList<>(2);
+        boolean sectionHasMatch = false;
+
+        // Run the tool on dummy mate read to determine what the mate cigar will be upon completion, if manager has a prediction then dont repair
+        if (emitReads && read.hasAttribute("MC")) {
```
lbergelson: It's a good idea to define a String constant for any Strings you use more than once.  Avoids potential typos.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +143,76 @@ public void closeTool() {
...
      *
-     * @param read     the read to split
+     * @param read     the read to split (can be null)
+     * @param emitReads   a parameter used to mock behavior for repairing mate cigar string information
      */
-    public static GATKRead splitNCigarRead(final GATKRead read, OverhangFixingManager manager) {
+    public static GATKRead splitNCigarRead(final GATKRead read, OverhangFixingManager manager, boolean emitReads, SAMFileHeader header, boolean secondaryAlignments) {
         final int numCigarElements = read.numCigarElements();
+        List<GATKRead> splitReads = new ArrayList<>(2);
+        boolean sectionHasMatch = false;
```
lbergelson: you could move this variable down into the for loop I think

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -202,23 +231,28 @@ private void cleanSplices() {
      * @param read        the read to fix
      * @param splice      the split (bad region to clip out)
      */
-    private void fixSplit(final SplitRead read, final Splice splice) {
+    void fixSplit(final SplitRead read, final Splice splice) {
         // if the read doesn't even overlap the split position then we can just exit
         if ( read.loc == null || !splice.loc.overlapsP(read.loc) ) {
             return;
         }
+        // if the read is secondary, do not clip it by overhang clipping at all
```
lbergelson: This comment seems misleading in the case where `processSecondaryReads == true`

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -251,51 +285,95 @@ protected static boolean isRightOverhang(final GenomeLoc readLoc, final GenomeLo
...
+
     // class to represent the reads with their soft-clip-included GenomeLocs
     public final class SplitRead {
 
+        // Relevant information to determine if the read has been clipped by the manager
+        private final Cigar oldCigar;
+        private final int oldStart;
+
         public GATKRead read;
         public GenomeLoc loc;
```
lbergelson: I think things might be clearer if `loc` was renamed to `unclippedLoc`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -202,23 +231,28 @@ private void cleanSplices() {
      * @param read        the read to fix
      * @param splice      the split (bad region to clip out)
      */
-    private void fixSplit(final SplitRead read, final Splice splice) {
+    void fixSplit(final SplitRead read, final Splice splice) {
```
lbergelson: I'd probably rename `read -> splitRead` and then bind `read.read` and `read.loc` to their own variables in this function for reduced verbosity and ease of understanding.  The logic this function is super confusing as is so anything that makes it easier to read is good.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -251,51 +285,95 @@ protected static boolean isRightOverhang(final GenomeLoc readLoc, final GenomeLo
...
 
+    // Sets the tool to start writing its output to the file writer
+    public void activateWriting() {
+        close();
+        splices.clear();
+        outputToFile = true;
+    }
+
     // class to represent the reads with their soft-clip-included GenomeLocs
     public final class SplitRead {
```
lbergelson: It would be really nice if this class (and the other private classes) could be made static.  It would mean converting `GenomeLoc -> SimpleInterval` though, which I think should be pretty straightforward.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -200,7 +252,26 @@ private static GATKRead splitReadBasedOnCigar(final GATKRead read, final int cig
...
     }
 
+    /**
+     * A method that repairs the NH and NM tags for a group of reads
+     *
+     * @param readFamily         the a list of reads where the first item is the read to be marked as primary
+     * @param header             the file header to associate with the given reads
+     * @return a non-null read representing the section of the original read being split out
+     */
+    public static void repairSuplementaryTags(List<GATKRead> readFamily, SAMFileHeader header) {
```
lbergelson: typo: repairSup**p**lementaryTags

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -200,7 +252,26 @@ private static GATKRead splitReadBasedOnCigar(final GATKRead read, final int cig
...
+        for (GATKRead read : readFamily) {
+            if (readFamily.size()>1) {
```
lbergelson: weirrd spacing here, in general you'd want spaces around `>` like you did below.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -200,7 +252,26 @@ private static GATKRead splitReadBasedOnCigar(final GATKRead read, final int cig
...
+                read.clearAttribute("NH");
+            }
+            read.setAttribute("NM", SequenceUtil.calculateSamNmTagFromCigar(read.convertToSAMRecord(header)));
```
lbergelson: This seems like it will be wrong most of the time.  We usually have `M`s rather than `EQ` and `X`, so calculating NM from only the cigar is not possible.  (That's why we need the NM tag in the first place...)  I think you must have the reference sequence in order to actually calculate NM.

src/main/java/org/broadinstitute/hellbender/utils/clipping/ClippingOp.java
```diff
@@ -40,6 +40,11 @@ public int getLength() {
      * @param originalRead the read to be clipped
      */
     public GATKRead apply(final ClippingRepresentation algorithm, final GATKRead originalRead) {
+        return apply(algorithm, originalRead, true);
+    }
+
+    // Warning, this method is intended for testing purposes
```
lbergelson: Use `@VisibleForTesting` instead of comments.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +143,76 @@ public void closeTool() {
...
+
+            // One of the "Real" operators
+            if (op == CigarOperator.M || op == CigarOperator.EQ || op == CigarOperator.X ||
+                    op == CigarOperator.I || op == CigarOperator.D) {
+                sectionHasMatch = true;
+            }
+
+            if (op == CigarOperator.N) {
+                if (sectionHasMatch) {
+                    if (emitReads == false) {
```
lbergelson: `emitReads == false` -> `!emitReads`

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +143,76 @@ public void closeTool() {
...
+        }
+        manager.setPredictedMateInformation(read);
+
+        // If it is a secondary alignment, repair its mate-information (assuming its mate was primary) and pass
+        if ( !secondaryAlignments &&  read.isSecondaryAlignment()){
+            manager.addReadGroup(Collections.singletonList(read));
+            return read;
+        }
 
         int firstCigarIndex = 0;
```
lbergelson: is `firstCigarElementInSplit` a better name than `firstCigarIndex` ?

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +143,76 @@ public void closeTool() {
...
+            if (op == CigarOperator.M || op == CigarOperator.EQ || op == CigarOperator.X ||
+                    op == CigarOperator.I || op == CigarOperator.D) {
+                sectionHasMatch = true;
+            }
+
+            if (op == CigarOperator.N) {
+                if (sectionHasMatch) {
+                    if (emitReads == false) {
+                        // not passing the manager ensures that no splices get added to the manager for fake reads
+                        splitReads.add(splitReadBasedOnCigar(read, firstCigarIndex, i, null));
```
lbergelson: while you're here, could you change the instances of `read.getCigar().getCigarElement(i)` in `splitReadBaseOnCigar` into `read.getCigarElement()` calls?  It should have a small performance impact.

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
```diff
@@ -488,10 +492,146 @@ protected GATKRead hardClipByReferenceCoordinates(final int refStart, final int
...
         this.addOp(new ClippingOp(start, stop));
         final GATKRead clippedRead = clipRead(ClippingRepresentation.HARDCLIP_BASES);
         this.ops = null;
         return clippedRead;
     }
+
+
+
+    /**
+     * Soft clip the read to the variable region (from refStart to refStop) processing also the clipped bases
```
lbergelson: The new code in this class is mostly copy pasted from the hardClip functions.  Could you add private methods that accept the clipping op to be used then refactor both hard and soft clipping operations to use that shared coded?

lbergelson:
Do you fix MD tags anywhere?  They're definitely invalidated if you split the read, should maybe consider fixing or just removing them if present.

src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
```diff
@@ -0,0 +1,96 @@
+package org.broadinstitute.hellbender.engine;
```
lbergelson: I'm not sure if we want to support this as part of the engine or just make it a package protected class and move it to rnaseq.  If we want to keep it in engine then we should add integration tests for it with an example walker.

src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * A ReadWalker is a tool that processes a single read at a time from one or multiple sources of reads, with
+ * optional contextual information from a reference and/or sets of variants/Features. The TwoPassReadsWalker iterates
+ * through the reads twice, allowing for different processing to be performed on the reads each time.
+ *
+ * If multiple sources of reads are specified, they are merged together into a single sorted stream of reads.
+ *
+ * TwoPassReadWalker authors must implement the firstPassApply() and secondPassApply() methods to process each read,
+ * and may optionally implement onTraversalStart() and/or onTraversalSuccess() and afterFirstPass() to perform
+ * operations between passes. See the SplitNCigarReads walker for an example.
```
lbergelson: you should make this reference a `{@link}`

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -251,51 +285,95 @@ protected static boolean isRightOverhang(final GenomeLoc readLoc, final GenomeLo
...
+     * Writes read groups off the top of waitingReads until the total number of reads in the set is less than the
+     * target queue size. If outputToFile == false, then the it will instead mark the first item to setMateChanged in
+     * mateChagnedReads
+     */
+    private void writeReads(int targetQueueSize) {
         // write out all of the remaining reads
-        while ( ! waitingReads.isEmpty() ) {
-            writer.addRead(waitingReads.poll().read);
+        while ( getNReadsInQueue() > targetQueueSize ) {
+            List<SplitRead> waitingGroup = waitingReadGroups.poll();
```
lbergelson: If we're being really paranoid we might want to make a private class that keeps the count and the read groups and only offers access in a way that they can't get out of synch.  Probably not worth it though since there are very few accesses.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -251,51 +285,95 @@ protected static boolean isRightOverhang(final GenomeLoc readLoc, final GenomeLo
...
+                        .map( r -> r.read )
+                        .collect(Collectors.toList()), header);
+                for (SplitRead splitRead : waitingGroup) {
+                    writer.addRead(splitRead.read);
+                }
+
+                // On the first traversal we want to store reads that would be the mate
+            } else {
+                // Don't mark the readgroup if it is secondary (mate information should always point to the primary alignment)
+                if (!waitingGroup.get(0).read.isSecondaryAlignment() && waitingGroup.get(0).hasBeenOverhangClipped()) {
```
lbergelson: I might pull out a variable with a name in cases like this where you have lots of repetitive call chains like this.  Something like `primaryRead waitingGroup.get(0)`

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -89,13 +77,22 @@
     @Argument(fullName="maxBasesInOverhang", shortName="maxOverhang", doc="max number of bases allowed in the overhang", optional=true)
     protected int MAX_BASES_TO_CLIP = 40;
 
-    @Argument(fullName="doNotFixOverhangs", shortName="doNotFixOverhangs", doc="do not have the walker hard-clip overhanging sections of the reads", optional=true)
+    @Argument(fullName="doNotFixOverhangs", shortName="doNotFixOverhangs", doc="do not have the walker soft-clip overhanging sections of the reads", optional=true)
     protected boolean doNotFixOverhangs = false;
 
+    @Argument(fullName="processSecondaryAlignments", shortName="processSecondaryAlignments", doc="have the walker split secondary alignments (will still repair MC tag without it)", optional=true)
+    protected boolean processSecondaryAlignments = false;
```
lbergelson: Could you fix it so these are default visibility instead of protected.  It's wierd to declare protected variables in a final class.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -104,22 +101,34 @@ public CountingReadFilter makeReadFilter() {
...
+
+    @Override
+    protected void secondPassApply(GATKRead read, ReferenceContext bytes, FeatureContext featureContext) {
+        splitNCigarRead(rnaReadTransform.apply(read),overhangManager, true, header, processSecondaryAlignments);
+    }
+
+    // Activates writing in the manager, which destinguishes each pass
+    @Override
+    protected void afterFirstPass() {
+        super.afterFirstPass();
```
lbergelson: I don't think you need this call, since the super is empty.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +143,76 @@ public void closeTool() {
 
 
     /**
-     * Goes through the cigar string of the read and create new reads for each consecutive non-N elements (while hard clipping the rest of the read).
-     * For example: for a read with cigar '1H2M2D1M2N1M2I1N1M2S' 3 new reads will be created with cigar strings: 1H2M2D1M, 1M2I and 1M2S
+     * Goes through the cigar string of the read and create new reads for each consecutive non-N elements (while soft clipping the rest of the read) that are supplemental to eachother.
```
lbergelson: typo: **each other**

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -134,33 +143,76 @@ public void closeTool() {
...
+     * For example: for a read with cigar '1H2M2D1M2N1M2I1N1M2S' 3 new reads will be created with cigar strings: 1H2M2D1M6S, 1H3S1M2I2S and 1H6S1M1S
```
lbergelson: Is this example correct?  Why is the last one 1S instead of 2S?

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
```diff
@@ -1120,4 +1120,57 @@ public static boolean validateExpectedSortOrder(
...
+     * @param primaryRead the primary read in the set of supplemental reads
+     * @param supplementalReads an arbitrarily sorted list of reads to be set as
+     *                          supplemental to eachother and the primary read
+     *
+     * @throws UserException 
+     */
+    public static void setReadsAsSupplemental(GATKRead primaryRead, List<GATKRead> supplementalReads) {
+        List<String> orderedTags = new LinkedList<>();
+        orderedTags.add(String.format("%s,%d,%s,%s,%s,%s;",
+                ((primaryRead.getName()!=null)?primaryRead.getName():"*"),
```
lbergelson: unnecessary parens and tight spacing.

prefer 
`primaryRead.getName() != null ? primaryRead.getName() : "*"`

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
```diff
@@ -1120,4 +1120,57 @@ public static boolean validateExpectedSortOrder(
...
+     * This tool will set the isSupplemental attribute as well as the 'SA:' tag
+     * properly according to the samspec (example: SA:primaryreadname,position,strand,Cigar,mapQ,NM;...)
+     *
+     * @param primaryRead the primary read in the set of supplemental reads
+     * @param supplementalReads an arbitrarily sorted list of reads to be set as
+     *                          supplemental to eachother and the primary read
+     *
+     * @throws UserException 
+     */
+    public static void setReadsAsSupplemental(GATKRead primaryRead, List<GATKRead> supplementalReads) {
```
lbergelson: This method should probably have some precondition checks, 
i.e. primary nonNull, supplementalReads nonEmpty.

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
```diff
@@ -1120,4 +1120,57 @@ public static boolean validateExpectedSortOrder(
...
+        List<String> orderedTags = new LinkedList<>();
+        orderedTags.add(String.format("%s,%d,%s,%s,%s,%s;",
```
lbergelson: Is it possible to extract a helper method to do this formatting instead of having the same long complex block of formatting code repeated twice?

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReadsUnitTest.java
```diff
@@ -90,49 +140,43 @@ private int numOfNElements(final List<CigarElement> cigarElements){
...
-    private void assertBases(final byte[] actualBase, final byte[] expectedBase, final int startIndex) {
+    private void assertBases(final byte[] actualBase, final byte[] expectedBase) {
         for (int i = 0; i < actualBase.length; i++) {
-            Assert.assertEquals(actualBase[i], expectedBase[startIndex + i], "unmatched bases between: " + Arrays.toString(actualBase) + "\nand:\n" + Arrays.toString(expectedBase) + "\nat position: " + i);
+            Assert.assertEquals(actualBase[i], expectedBase[i], "unmatched bases between: " + Arrays.toString(actualBase) + "\nand:\n" + Arrays.toString(expectedBase) + "\nat position: " + i);
         }
     }
 
-}
+    private class DummyTestWriter implements GATKReadWriter {
```
lbergelson: should be static

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReadsUnitTest.java
```diff
@@ -32,12 +38,58 @@
...
-        public TestManager( final SAMFileHeader header ) {
-            super(header, null, hg19GenomeLocParser, hg19ReferenceReader, 10000, 1, 40, false);
+        public TestManager( final SAMFileHeader header , DummyTestWriter writer) {
+            super(header, writer, hg19GenomeLocParser, hg19ReferenceReader, 10000, 1, 40, false, true);
         }
     }
 
     @Test
+    public void testBogusNSplits() {
+        DummyTestWriter writer = new DummyTestWriter();
```
lbergelson: better to split these into several tests rather than reuse these variables

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReadsUnitTest.java
```diff
@@ -32,12 +38,58 @@
...
+        final SAMFileHeader header = new SAMFileHeader();
```
lbergelson: You could make this header a constant and then add a method to get an active manager.  That Would let you split up the tests without spending a lot of extra lines on boilerplate.

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManagerUnitTest.java
```diff
@@ -146,8 +158,117 @@ an NPE from getContig() of an unmapped read.*/
...
+        Assert.assertEquals(read1b.getMateStart(), 10020);
+        Assert.assertTrue(manager.setPredictedMateInformation(read2));
+        Assert.assertEquals(read2.getMateStart(), 10003);
+
+        // Despite having been split, read 3a should not have been marked as changing
+        Assert.assertFalse(manager.setPredictedMateInformation(read3a));
+        Assert.assertFalse(manager.setPredictedMateInformation(read4));
+    }
+
+    private class OverhangFixingManagerAlwaysSplit10000Reads extends OverhangFixingManager {
```
lbergelson: make all these test classes static if you can

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManagerUnitTest.java
```diff
@@ -146,8 +158,117 @@ an NPE from getContig() of an unmapped read.*/
...
+        read2.setMatePosition("1",10000);
+        read2.setIsSecondOfPair();
+
+        // Second Alignment set as secondary to see that it wont be clipped
+        GATKRead read3a = ArtificialReadUtils.createArtificialRead(hg19Header, "read2", 1, 10000, new byte[]{(byte)'A', (byte)'A', (byte)'A', (byte)'A', (byte)'A', (byte)'A'}, new byte[]{(byte)'A', (byte)'A', (byte)'A', (byte)'A', (byte)'A', (byte)'A'}, "6M");
+        read3a.setMatePosition("1", 10020);
+        read3a.setIsSecondaryAlignment(true);
+        read3a.setIsSecondOfPair();
+        List<GATKRead> readGroup2 = Collections.singletonList(read3a);
+        GATKRead read4 = ArtificialReadUtils.createArtificialRead(hg19Header, "read2", 1, 10020, new byte[]{(byte)'C', (byte)'C', (byte)'A', (byte)'A', (byte)'A', (byte)'A'}, new byte[]{(byte)'C', (byte)'C', (byte)'A', (byte)'A', (byte)'A', (byte)'A'}, "6M");
```
lbergelson: You could save a lot of characters by using "CCAAAA".getBytes()

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -307,10 +385,49 @@ public void setRead(final GATKRead read) {
...
+        // Returns true if either of the required mate information fields have been changed by the clipper
+        public boolean hasBeenOverhangClipped() {
+            return (!oldCigar.equals(read.getCigar())) || (oldStart != read.getStart());
+        }
+
+        // Adds the relevant information for repairing the mate to setMateChanged keyed on a string composed of the start position
+        public void setMateChanged() {
+            if (!read.isUnmapped()) {
+                mateChangedReads.put(read.getName() + (read.isFirstOfPair()? 0: 1) + oldStart,
+                        new Tuple<>(read.getStart(),read.getCigar().toString()));
```
lbergelson: use `TextCigarCodec.encode()` instead of toString.  Better to not rely on the behavior of toString for program correctness.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -307,10 +385,49 @@ public void setRead(final GATKRead read) {
...
+
+        // Returns true if either of the required mate information fields have been changed by the clipper
+        public boolean hasBeenOverhangClipped() {
+            return (!oldCigar.equals(read.getCigar())) || (oldStart != read.getStart());
+        }
+
+        // Adds the relevant information for repairing the mate to setMateChanged keyed on a string composed of the start position
+        public void setMateChanged() {
+            if (!read.isUnmapped()) {
+                mateChangedReads.put(read.getName() + (read.isFirstOfPair()? 0: 1) + oldStart,
```
lbergelson: You should extract a function for getting the key and use it everywhere it is generated.  Otherwise there's a good chance it will break when someone updates one stop but not another.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -307,10 +385,49 @@ public void setRead(final GATKRead read) {
...
         }
+
+        // Returns true if either of the required mate information fields have been changed by the clipper
+        public boolean hasBeenOverhangClipped() {
+            return (!oldCigar.equals(read.getCigar())) || (oldStart != read.getStart());
+        }
+
+        // Adds the relevant information for repairing the mate to setMateChanged keyed on a string composed of the start position
+        public void setMateChanged() {
+            if (!read.isUnmapped()) {
```
lbergelson: should this also check that read is paired?

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -251,51 +285,95 @@ protected static boolean isRightOverhang(final GenomeLoc readLoc, final GenomeLo
...
 
         // we can still mismatch overall if at least half of the bases mismatch
+
         return numMismatchesSeen >= ((spanToTest+1)/2);
     }
 
     /**
      * Close out the manager stream by clearing the read cache
      */
     public void close() {
```
lbergelson: I think close should be renamed to `flush` since it doesn't actually close the stream.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -251,51 +285,95 @@ protected static boolean isRightOverhang(final GenomeLoc readLoc, final GenomeLo
...
+            } else {
+                // Don't mark the readgroup if it is secondary (mate information should always point to the primary alignment)
+                if (!waitingGroup.get(0).read.isSecondaryAlignment() && waitingGroup.get(0).hasBeenOverhangClipped()) {
+                    waitingGroup.get(0).setMateChanged();
+                }
+            }
         }
     }
 
+    // Sets the tool to start writing its output to the file writer
```
lbergelson: This is a public method so this should be javadoc instead of just a comment.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
```diff
@@ -251,51 +285,95 @@ protected static boolean isRightOverhang(final GenomeLoc readLoc, final GenomeLo
...
+    public void activateWriting() {
+        close();
```
lbergelson: Could you log how many changed reads are stored at the time of activation?  

Also, are you allowed to call activateWriting multiple times?  Should that be an error condition or is there an expected usage that you might want to do that?

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
```diff
@@ -1120,4 +1120,57 @@ public static boolean validateExpectedSortOrder(
...
+     * properly according to the samspec (example: SA:primaryreadname,position,strand,Cigar,mapQ,NM;...)
+     *
+     * @param primaryRead the primary read in the set of supplemental reads
+     * @param supplementalReads an arbitrarily sorted list of reads to be set as
+     *                          supplemental to eachother and the primary read
+     *
+     * @throws UserException 
+     */
+    public static void setReadsAsSupplemental(GATKRead primaryRead, List<GATKRead> supplementalReads) {
+        List<String> orderedTags = new LinkedList<>();
```
lbergelson: Don't use a linked list here ( or really anywhere...)  They're so much slower for everything but adding or removing nodes in the middle.  Here you use it for random access.  If you're worried about the resizing cost set the `ArrayList(size)` when you create it since you know that based on the number of reads that were provided.

lbergelson:
@jamesemery Back to you.  Looks good!  A had a bunch of minor comments and refactoring requests to make things a little cleaner and more foolproof for later readers.

src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
jamesemery: The reason it is in the engine is because many of the features that the twoPassWalker needs to access in order to operate are package protected within the engine code.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/OverhangFixingManager.java
jamesemery: Good point! It occurs to me that for the potential future iteration of the tool though that assumption might not necessarily hold if we are overhang clipping supplementary alignments and need to save them for SA tags. One could potentially have a chimeric alignment that is unpaired.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
jamesemery: Right, this is an artifact of having had to reinitialize the reads source and filtering, seems I forgot to change it back.

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
jamesemery: For all of the changes to this class, many of them have been addressed by my SATagBuilder class

jamesemery: This code gets completely rewritten to be nicer and cleaner as well as having useful features for changing these tags.

jamesemery:
@lbergelson Alright, this time i think i have responded to all of the comments, back to you!

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
```diff
@@ -2,6 +2,7 @@
 
 import htsjdk.samtools.CigarElement;
 import htsjdk.samtools.CigarOperator;
+import jdk.management.resource.internal.inst.InitInstrumentation;
```
lbergelson: @jamesemery This got added somehow and it's causing compile errors.

src/main/java/org/broadinstitute/hellbender/utils/clipping/ReadClipper.java
lbergelson: I agree with Megan that this should probably emit a warning when this happens since I don't think it's something anyone would expect to happen.

lbergelson:
@jamesemery Looks good to me.  Could you add those warning messages Megan asked for and fix the compile error and then I think it's good to merge.

PR #2035, opened 2016-07-22
remove direct spark dependency and use ML lib

akiezun:
no reason to depend on spark when we depend on spark ML lib anyway. Simpler to update 1 dependency than two

@lbergelson wdyt?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -145,13 +145,6 @@ dependencies {
     compile('org.apache.hadoop:hadoop-client:2.7.2') // should be a 'provided' dependency
     compile('com.github.jsr203hadoop:jsr203hadoop:1.0.2')
 
-    compile('org.apache.spark:spark-core_2.10:1.6.1') {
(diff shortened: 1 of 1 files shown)


lbergelson:
Seems like a philosophical question.  Do we want to include explicit dependencies on everything we directly depend on, or are we ok depending directly on something else's transitive dependencies.  I think it's probably better to explicitly specify our direct dependencies in most cases.  (I.e. I wouldn't leave out htsjdk just because we depend on hadoop-bam.)  

In this case ml-lib is so closely tied to spark that it's reasonable to argue that it is specifying spark as a direct dependency.  Are there any spark components that are not transitive dependencies of mllib?

I don't feel strongly about this either way.  I think I'd lean towards keeping it, but I'm fine if you prefer to save the lines of code.

akiezun:
agreed about the general case. also agreed that in this case spark and mllib are tightly linked and so i opt for keeping just one and saving ourselves the trouble of keeping the versions in sync.

Issue #2034, opened 2016-07-22
Overlapping read-pairs handling


magicDGS:
While using `CheckPileup` to check the output of samtools mpileup for overlapping reads I found that the internal `ReadPileup` it is different when there are overlapping read-pairs. The latest version of samtools set one of the base pair qualities to 0 and the other to the sum of both qualities, generating a different result.

Thus, when I use a `ReadPileup` that is filtered by quality for get the base counts, I got a different result that it is actually not the correct one, because a single read-pair is counted twice. Could it be possible to implement one method for handling this overlapping read-pairs, although I know that it is a non-trivial issue?

magicDGS:
Closed by #2154

Issue #1879, opened 2016-06-06
Inconsistent constructor arguments ordering in LocusIteratorByState


magicDGS:
As @droazen point out in #1833, constructors for `LocusIteratorByState` have different arguments ordering. It should be fixed eventually.

magicDGS:
@droazen, I will address this in #2041. As you suggested this when I implemented `LocusWalker`, I would like to have some idea about why `DownsamplingMethod` is used as a parameter in the constructor. I think that this is misleading, because independently of the method for downsampling the one that is used by `SamplePartitioner` is  a `ReservoirDownsampler` (if downsampling is performed), so API users could think that they are performing a different downsampling in LIBS that the actual one.

I will keep the constructor in the PR, but I would like some feedback for this either here or in #2041.

magicDGS:
Closed by #2154

PR #2041, opened 2016-07-25
LIBS/LocusWalker refactoring and overlapping read-pairs handling

magicDGS:
As described in #2034, samtools mpileup and the internal pileup generated by `LocusIteratorByState` (LIBS) is not providing the same result because overlapping read-pairs are not taken into account. In this PR I addressed this issue using the same approach as samtools to combine qualities, including new functionality to LIBS and `LocusWalker`:

* Refactoring constructors for LIBS, solving #1879.
* Including an option for ignore overlapping read-pairs in LIBS
* Including command line option in `LocusWalker`to ignore overlapping read-pairs and to downsample with a maximum coverage by sample.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
@@ -7,6 +7,7 @@ import htsjdk.samtools.util.Locatable;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.hellbender.engine.AlignmentContext;
+import org.broadinstitute.hellbender.tools.walkers.qc.Pileup;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.downsampling.DownsamplingMethod;
@@ -77,6 +78,11 @@ public final class LocusIteratorByState implements Iterable<AlignmentContext>, I
      * Should we include reads in the pileup which are aligned with a deletion operator to the reference?
      */
     private final boolean includeReadsWithNsAtLoci;
+
+    /**
+     * Should we ignore read-pair overlaps?
+     */
+    private final boolean ignorePairOverlaps;
 
     /**
      * The next alignment context.  A non-null value means that a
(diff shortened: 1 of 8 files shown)


magicDGS:
@akiezun... because you are working in the performance of LIBS (#2032), could you have a look?

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
```diff
@@ -352,6 +445,37 @@ private void lazyLoadNextAlignmentContext() {
     }
 
     /**
+     * Fix the quality of two elements that comes from an overlaping pair
```
akiezun: Please provide documentation what is being done here. Something more specific than 'fix the quality' - what's the exact algorithm?

akiezun:
I think this one needs to be reviewed by @droazen who is very familiar with LIBS. Unless this is off by default, it may change tool results and thus is non-trivial to review.

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
magicDGS: Sure, I will add it. Basically is the same algorithm as samtools is using, and I can provide a test where `CheckPileup` complains with overlapping reads because of the overlapping read-pairs...

magicDGS:
For clarity for you @droazen (and @akiezun), this will change the result of the tools, but in a way where the output will be the same to the one in `samtools`. And actually, if there are overlapping read-pairs the base counts from the `ReadPileup` may be wrong...

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
magicDGS: Done.

magicDGS:
I was thinking more carefully on this and another option is create methods in `ReadPileup` to fix the overlaps after construction and/or getBaseCounts without overlaps. This won't break the behaviour of LIBS and it is up to the user to change overlaps. But for performance issues, I would like to have a variable in `ReadPileup` for track if the overlaps are corrected/fixed, to avoid recomputation. I can implement this in a different PR or in this one if the basic idea behind this one is not accepted.

magicDGS:
Closing in favor of #2154

PR #2045, opened 2016-07-25
remove unused code from CountSet

akiezun:
CountSet is only used in protected and only the following methods are used:
```
add(int);
isEmpty()
max();
size();
min();
```

so i deleted all the other unused code. 

Whether this whole code could be just deleted and any sorted int set implementation used instead (eg TreeSet<Integer> or any subclass of `IntSortedSet` from fast `it.unimi.dsi.fastutil.ints`) is another question.

@droazen whyt?


droazen:
Closing this one, as it seems bad to have a `CountSet` that doesn't implement `Set`. Instead we are creating a ticket in protected to replace usage of this class with a built-in collection that has an efficient min/max operation.

Issue #365, opened 2015-04-09
Indexing output files on the fly


nh13:
As a bioinformatician, I would like any hellbender tool to output an index along with the primary output file, for any output file format that can be indexed (ex. BAM/VCF/BCF/FASTA), so that I do not need to re-read the output file to create an index (looking at you GenotypeVCFs, tabix, and CombineVCFs).

droazen:
Depends on https://github.com/broadinstitute/hellbender/issues/121

akiezun:
needs a way to prevent data races. Otherwise, multiple concurrent tools write the same index file and bad things happen.

nh13:
How about an indexingoutputstream?

lbergelson:
Are data races actually an issue here?  If you have 2 tools creating the same **output** file, that's a problem on it's own.  I think the data race issue was a problem with tools that auto indexed their **input** files.

nh13:
Basically Picard's ```CREATE_INDEX``` command line option should have an analogous option in hellbender.

akiezun:
@lbergelson yes, i meant 2 concurrent tools making indices for the same **input** file. They will both try to write the same index file, with bad consequences.

droazen:
This ticket only deals with on-the-fly indexing of output files. We already know from our experiences with the old GATK that on-the-fly indexing of input files is a bad idea, so I don't think anyone is advocating for that.

akiezun:
renamed the ticker

droazen:
Let's do this one for alpha-1

cmnbroad:
Walkers:
-ReadWalkers These is already done through the GATKTool createSAMWriter for any BAM outputs that are coordinate sorted.
-VariantWalkers We need something in GATKTool similar to createSAMWriter for creating VCFs that honors the index arg.
-IntervalWalkers Not sure what these will need.
-PicardCommandLineTool These already use a common method for creating BAM writers that honors CREATE_INDEX. We need similar thing for VCFs.
-GATKSparkTools should these default to auto indexing as well ? 

Questions:
Right now we have a --createOutputBAMIndex arg for GATK tools - should this be moved down to ReadWalker ?
PicardCommandLineProgram currently has a --CREATE_INDEX arg, documented as being for BAMs.

Should we unify the name of the indexing argument across all of these, or possibly unify down to two, one for alignment output files and one for variant output files ? Some GATK tools (ie., HaplotypeCaller) create both BAM and VCF outs.

[...]

PR #2046, opened 2016-07-26
Common VCFWriter and honor index/md5 arguments.

cmnbroad:
Common VCFWriter creation, and fix for https://github.com/broadinstitute/gatk/issues/365 (for non-Spark tools).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
@@ -6,18 +6,24 @@ import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.SamReaderFactory;
 import htsjdk.samtools.util.Locatable;
 import htsjdk.tribble.Feature;
+import htsjdk.variant.variantcontext.writer.Options;
+import htsjdk.variant.variantcontext.writer.VariantContextWriter;
 import org.broadinstitute.hellbender.cmdline.Argument;
 import org.broadinstitute.hellbender.cmdline.ArgumentCollection;
 import org.broadinstitute.hellbender.cmdline.CommandLineProgram;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 import org.broadinstitute.hellbender.cmdline.argumentcollections.*;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.SequenceDictionaryUtils;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.io.IOUtils;
 import org.broadinstitute.hellbender.utils.read.ReadUtils;
 import org.broadinstitute.hellbender.utils.read.SAMFileGATKReadWriter;
+import org.broadinstitute.hellbender.utils.variant.GATKVariantContextUtils;
 
 import java.io.File;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
@@ -43,11 +49,30 @@ public abstract class GATKTool extends CommandLineProgram {
     @Argument(fullName = "disableSequenceDictionaryValidation", shortName = "disableSequenceDictionaryValidation", doc = "If specified, do not check the sequence dictionaries from our inputs for compatibility. Use at your own risk!", optional = true)
     private boolean disableSequenceDictionaryValidation = false;
 
-    @Argument(fullName="createOutputBamIndex", shortName="createOutputBamIndex", doc = "If true, create a BAM/CRAM index when writing a coordinate-sorted BAM/CRAM file", optional=true)
+    @Argument(fullName=StandardArgumentDefinitions.CREATE_OUTPUT_BAM_INDEX_LONG_NAME,
+            shortName=StandardArgumentDefinitions.CREATE_OUTPUT_BAM_INDEX_SHORT_NAME,
+            doc = "If true, create a BAM/CRAM index when writing a coordinate-sorted BAM/CRAM file.", optional=true)
     public boolean createOutputBamIndex = true;
 
-    @Argument(fullName="createOutputBamMD5", shortName="createOutputBamMD5", doc = "If true, create a MD5 digest for any BAM/SAM/CRAM file created", optional=true)
+    @Argument(fullName=StandardArgumentDefinitions.CREATE_OUTPUT_BAM_MD5_LONG_NAME,
+            shortName=StandardArgumentDefinitions.CREATE_OUTPUT_BAM_MD5_SHORT_NAME,
+            doc = "If true, create a MD5 digest for any BAM/SAM/CRAM file created", optional=true)
     public boolean createOutputBamMD5 = false;
+
+    @Argument(fullName=StandardArgumentDefinitions.CREATE_OUTPUT_VARIANT_INDEX_LONG_NAME,
+            shortName=StandardArgumentDefinitions.CREATE_OUTPUT_VARIANT_INDEX_SHORT_NAME,
+            doc = "If true, create a VCF index when writing a coordinate-sorted VCF file.", optional=true)
+    public boolean createOutputVariantIndex = true;
+
+    @Argument(fullName=StandardArgumentDefinitions.CREATE_OUTPUT_VARIANT_MD5_LONG_NAME,
+            shortName=StandardArgumentDefinitions.CREATE_OUTPUT_VARIANT_MD5_SHORT_NAME,
+            doc = "If true, create a a MD5 digest any VCF file created.", optional=true)
+    public boolean createOutputVariantMD5 = false;
+
+    @Argument(fullName= StandardArgumentDefinitions.LENIENT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.LENIENT_SHORT_NAME,
+            doc = "Lenient processing of VCF files", common = true, optional = true)
+    protected boolean lenientVCFProcessing = false;
 
     @Argument(fullName="addOutputSAMProgramRecord", shortName="addOutputSAMProgramRecord", doc = "If true, adds a PG tag to created SAM/BAM/CRAM files.", optional=true)
     public boolean addOutputSAMProgramRecord = true;
@@ -418,6 +443,40 @@ public abstract class GATKTool extends CommandLineProgram {
                                 createOutputBamMD5
                         )
         );
+    }
+
+    /**
+     * Creates a VariantContextWriter whose outputFile type is determined by
+     * the outFile's extension, using the best available sequence dictionary for
+     * this tool, and default index, leniency and md5 generation settings.
+     *
+     * @param outFile output File for this writer. May not be null.
+     * @returns VariantContextWriter must be closed by the caller
+     */
+    public VariantContextWriter createVCFWriter(final File outFile) {
+        Utils.nonNull(outFile);
+
+        final SAMSequenceDictionary sequenceDictionary = getBestAvailableSequenceDictionary();
+
+        List<Options> options = new ArrayList<>();
+        if (lenientVCFProcessing) {
+            options.add(Options.ALLOW_MISSING_FIELDS_IN_HEADER);
+        }
+
+        if (createOutputVariantIndex) {
+            if (null == sequenceDictionary) {
+                logger.warn("An variant index will not be created - a sequence dictionary is required to create an output index");
+                // fall through and create without an index
+            } else {
+                options.add(Options.INDEX_ON_THE_FLY);
+            }
+        }
+
+        return GATKVariantContextUtils.createVCFWriter(
+                outFile,
+                sequenceDictionary,
+                createOutputVariantMD5,
+                options.toArray(new Options[options.size()]));
     }
 
     /**
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
@@ -4,9 +4,7 @@ import com.google.common.collect.Sets;
 import htsjdk.tribble.Feature;
 import htsjdk.variant.variantcontext.*;
 import htsjdk.variant.variantcontext.VariantContextUtils.JexlVCMatchExp;
-import htsjdk.variant.variantcontext.writer.Options;
 import htsjdk.variant.variantcontext.writer.VariantContextWriter;
-import htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder;
 import htsjdk.variant.vcf.*;
 import org.broadinstitute.hellbender.cmdline.Argument;
 import org.broadinstitute.hellbender.cmdline.CommandLineProgramProperties;
@@ -16,6 +14,7 @@ import org.broadinstitute.hellbender.engine.*;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.IntervalUtils;
 
+import java.io.File;
 import java.util.*;
 import java.util.stream.Collectors;
 
@@ -199,8 +198,7 @@ public final class VariantFiltration extends VariantWalker {
     }
 
     private void initializeVcfWriter() {
-        //TODO remove hardwiring to output VCFs
-        writer = new VariantContextWriterBuilder().setOutputFile(out).setOutputFileType(VariantContextWriterBuilder.OutputType.VCF).unsetOption(Options.INDEX_ON_THE_FLY).build();
+        writer = createVCFWriter(new File(out));
 
         // setup the header fields
         final Set<VCFHeaderLine> hInfo = new LinkedHashSet<>();
--- a/src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
@@ -2,12 +2,14 @@ package org.broadinstitute.hellbender.utils.variant;
 
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.util.Locatable;
+import htsjdk.tribble.AbstractFeatureReader;
 import htsjdk.tribble.TribbleException;
 import htsjdk.variant.variantcontext.*;
 import htsjdk.variant.variantcontext.writer.Options;
 import htsjdk.variant.variantcontext.writer.VariantContextWriter;
 import htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder;
 import htsjdk.variant.vcf.VCFConstants;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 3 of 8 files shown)


src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -43,12 +48,31 @@
     @Argument(fullName = "disableSequenceDictionaryValidation", shortName = "disableSequenceDictionaryValidation", doc = "If specified, do not check the sequence dictionaries from our inputs for compatibility. Use at your own risk!", optional = true)
     private boolean disableSequenceDictionaryValidation = false;
 
-    @Argument(fullName="createOutputBamIndex", shortName="createOutputBamIndex", doc = "If true, create a BAM/CRAM index when writing a coordinate-sorted BAM/CRAM file", optional=true)
+    @Argument(fullName=StandardArgumentDefinitions.CREATE_OUTPUT_BAM_INDEX_LONG_NAME,
+            shortName=StandardArgumentDefinitions.CREATE_OUTPUT_BAM_INDEX_SHORT_NAME,
+            doc = "If true, create a BAM/CRAM index when writing a coordinate-sorted BAM/CRAM file.", optional=true)
     public boolean createOutputBamIndex = true;
```
lbergelson: in general we avoid boolean flags that default to true.  Could you invert these so they're `dontCreateOutputBamIndex` etc.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -421,6 +445,47 @@ public final SAMFileGATKReadWriter createSAMWriter(final File outputFile, final
...
+
+        final SAMSequenceDictionary sequenceDictionary = getBestAvailableSequenceDictionary();
+        Options options[];
+
+        if (createOutputVariantIndex) {
+            if (null == sequenceDictionary) {
+                logger.warn("An variant index will not be created - a sequence dictionary is required to create an output index");
+                // fall through and create without an index
+                options = lenientVCFProcessing ?
+                        new Options[]{Options.ALLOW_MISSING_FIELDS_IN_HEADER} :
```
lbergelson: This logic is way more complicated than it needs to be.  Can you just use a 'List' instead of the '[]' and simply check each case and append or don't instead of repeating the ternary expressions for case?  Then just convert to array at the end.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -421,6 +445,47 @@ public final SAMFileGATKReadWriter createSAMWriter(final File outputFile, final
     }
 
     /**
+     * Creates a VariantContextWriter who's outputFile type is determined by
+     * the outFile's extension, using the best available sequence dictionary for
+     * this tool, and default index, leniency and md5 generation settings.
+     *
+     * @param outFile output File for this writer. May not be null.
+     * @returns VariantContextWriter must be closed by the caller
```
lbergelson: Why not have GatkTool keep a list of these and close them itself?

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -40,40 +42,69 @@ public static boolean isInformative(final double[] gls) {
...
+
         for (Options opt : options) {
             vcWriterBuilder = vcWriterBuilder.setOption(opt);
         }
 
         return vcWriterBuilder.build();
     }
 
+    // Determine the variant file type from the file extension.
+    // This should really be exposed in htsjdk from AbstractFeatureReader
```
lbergelson: could you open a ticket in htsjdk so we don't forget about it?

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
lbergelson: I guess the existing createSAMWriter() doesn't do it either, but it seems like maybe both of them should keep track of what they've created and try to close them at the end...

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
```diff
@@ -281,4 +321,207 @@ public void testBestSequenceDictionary_fromNothing() throws Exception {
...
+            final GATKTool tool,
+            final boolean createIndex,
+            final boolean createMD5,
+            final boolean lenient) throws IOException
+    {
+        final File tmpDir = createTempDir("createVCFTest");
+        final File outputFile = new File(tmpDir.getAbsolutePath(), "createVCFTest" + outputExtension);
+
+        List<String> args = new ArrayList<>();
+        args.add("--input"); args.add(inputFile.getCanonicalPath());
```
lbergelson: we have some ArgumentsBuilder convenience methods to do this less verbosely if you prefer

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
```diff
@@ -281,4 +321,207 @@ public void testBestSequenceDictionary_fromNothing() throws Exception {
...
+        args.add("--out"); args.add(outputFile.getCanonicalPath());
+        args.add("--createOutputVariantIndex"); args.add(Boolean.toString(createIndex));
+        args.add("--createOutputVariantMD5"); args.add(Boolean.toString(createMD5));
+        if (lenient) {
+            args.add("--lenient");
+        }
+
+        final CommandLineParser clp = new CommandLineParser(tool);
+        clp.parseArguments(System.out, args.toArray(new String[args.size()]));
```
lbergelson: args.getArgsArray()

lbergelson:
@cmnbroad Back you with a minor refactoring request.  It seems to me that GATKTool should try and close it's own writers, but maybe there's a reason not to that I'm missing?

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
droazen: I disagree in this case, actually (though I agree with the general principle of not having boolean args that default to true, in general). If @cmnbroad inverted this arg, it would be inconsistent with `createOutputBamMD5`, etc., below, which would be confusing. Recommend keeping all of these args as `create*` instead of a mix of `dontCreate*` and `create*`

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -421,6 +445,47 @@ public final SAMFileGATKReadWriter createSAMWriter(final File outputFile, final
     }
 
     /**
+     * Creates a VariantContextWriter who's outputFile type is determined by
```
droazen: `who's`

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
droazen: The current contract is that walkers must close their own writers. We can change this (and probably should), but I think it should be done in a separate ticket.

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
```diff
@@ -199,8 +198,7 @@ private String possiblyInvertFilterExpression( final String description ){
     }
 
     private void initializeVcfWriter() {
-        //TODO remove hardwiring to output VCFs
-        writer = new VariantContextWriterBuilder().setOutputFile(out).setOutputFileType(VariantContextWriterBuilder.OutputType.VCF).unsetOption(Options.INDEX_ON_THE_FLY).build();
+        writer = createVCFWriter(new File(out));
```
droazen: Can you have a look at `HaplotypeCallerEngine.makeVCFWriter()` in gatk-protected, and see whether that method can be made to be consistent with your work in this branch, and create a new ticket to patch it? It calls `GATKVariantContextUtils.createVCFWriter()`, but sometimes it constructs a new `GVCFWriter`. You may need a way to create a `GVCFWriter` at the `GATKTool` level, as well as a static entry point not tied to the tool instance (but we can deal with this in a separate ticket)

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -40,40 +42,69 @@ public static boolean isInformative(final double[] gls) {
     }
 
     /**
-     * Creates a VariantContextWriter who's outputFile type is <code>VariantContextWriterBuilder.OutputType.VCF</code>.
+     * Creates a VariantContextWriter who's outputFile type is based on the extension of the output file name.
```
droazen: `who's` -> `whose`

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -40,40 +42,69 @@ public static boolean isInformative(final double[] gls) {
...
+                new VariantContextWriterBuilder().clearOptions().setOutputFile(outFile);
+
+        if (VariantContextWriterBuilder.OutputType.UNSPECIFIED == getVariantFileTypeFromExtension(outFile)) {
+            // the only way the user has to specify an output type is by file extension, and htsjdk
+            // throws if it can't map the file extension to a known vcf type, so fallback to a default
+            // of VCF
+            logger.warn(String.format(
+                    "Can't determine output variant file format from output file extension \"%s\". Defaulting to VCF.",
+                    FilenameUtils.getExtension(outFile.getPath())));
+            vcWriterBuilder = vcWriterBuilder.setOutputFileType(VariantContextWriterBuilder.OutputType.VCF);
```
droazen: Does this not need to be explicitly set in the other case (`VariantContextWriterBuilder.OutputType.UNSPECIFIED != getVariantFileTypeFromExtension(outFile)`)?

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -40,40 +42,69 @@ public static boolean isInformative(final double[] gls) {
...
 
         return vcWriterBuilder.build();
     }
 
+    // Determine the variant file type from the file extension.
+    // This should really be exposed in htsjdk from AbstractFeatureReader
+    private static VariantContextWriterBuilder.OutputType getVariantFileTypeFromExtension(final File outputFile) {
+        final String extension = FilenameUtils.getExtension(outputFile.getPath()).toLowerCase();
+
+        if (extension.equals("vcf")) {
```
droazen: What about the gzipped versions of these?

droazen: Declare these extensions as constants in `VcfUtils`

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -40,40 +42,69 @@ public static boolean isInformative(final double[] gls) {
...
-     * @param outFile output File for this writer
+     * @param outFile output File for this writer. May not be null.
      * @param referenceDictionary required if on the fly indexing is set, otherwise can be null
-     * @param lenientProcessing true if ALLOW_MISSING_FIELDS_IN_HEADER option should be set
-     * @param options variable length list of additional options to be set for this writer
+     * @param createMD5 true if an md5 file should be created
+     * @param options variable length list of additional Options to be set for this writer
      * @returns VariantContextWriter must be closed by the caller
      */
     public static VariantContextWriter createVCFWriter(
```
droazen: This method should be able to create a `GVCFWriter` as well. GATK3 uses the semi-unofficial file extension `.g.vcf` for GVCFs.

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
```diff
@@ -281,4 +321,207 @@ public void testBestSequenceDictionary_fromNothing() throws Exception {
...
+
+        final VariantContextWriter writer = tool.createVCFWriter(outputFile);
+        writer.close();
+
+        final File outFileIndex = new File(outputFile.getAbsolutePath() + indexExtension);
+        final File outFileMD5 = new File(outputFile.getAbsolutePath() + ".md5");
+
+        Assert.assertTrue(outputFile.exists());
+        Assert.assertEquals(outFileIndex.exists(), createIndex);
+        Assert.assertEquals(outFileMD5.exists(), createMD5);
```
droazen: Add error messages for when the asserts fail

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
```diff
@@ -1733,30 +1737,197 @@ public void testTotalPloidy(final int[] ploidies, final int defaultPloidy, final
...
+
+        Options options[] = createIndex ?
+                new Options[] {Options.INDEX_ON_THE_FLY} :
+                new Options[] {};
+        try (final VariantContextWriter writer = GATKVariantContextUtils.createVCFWriter(
+                outputFile,
+                makeSimpleSequenceDictionary(),
+                createMD5,
+                options)) {
+            writeHeaderAndBadVariant(writer, false);
```
droazen: Why are you writing a "bad" variant in this `testCreateVCFWriterWithOptions` test case?

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
```diff
@@ -1733,30 +1737,197 @@ public void testTotalPloidy(final int[] ploidies, final int defaultPloidy, final
...
+            afr = AbstractFeatureReader.getFeatureReader(
+                    resultVCFFile.getAbsolutePath(),
+                    new VCFCodec(),
+                    createIndex);
+        }
+        else {
+            throw new IllegalArgumentException("Unknown file extension in createVCFWriter test validation");
+        }
+
+        Assert.assertNotEquals(null, afr);
```
droazen: `Assert.assertNotNull()`

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
```diff
@@ -1733,30 +1737,197 @@ public void testTotalPloidy(final int[] ploidies, final int defaultPloidy, final
...
+        Assert.assertTrue(outputFile.exists());
+        Assert.assertEquals(outFileIndex.exists(), createIndex);
+        Assert.assertEquals(outFileMD5.exists(), createMD5);
+
+        verifyFileType(outputFile, outputExtension, createIndex);
+    }
+
+    // just make sure we can read the file with the corresponding codec
+    private void verifyFileType(final File resultVCFFile, final String outputExtension, final boolean createIndex) {
+        AbstractFeatureReader<?, ?> afr = null;
```
droazen: Close this reader before the method returns

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
```diff
@@ -1733,30 +1737,197 @@ public void testTotalPloidy(final int[] ploidies, final int defaultPloidy, final
...
+                    createIndex);
+        }
+        else if (outputExtension.equals(".vcf.bgz") ||
+                outputExtension.equals(".vcf.gz") ||
+                outputExtension.equals(".tmp")) {
+            afr = AbstractFeatureReader.getFeatureReader(
+                    resultVCFFile.getAbsolutePath(),
+                    new VCFCodec(),
+                    createIndex);
+        }
```
droazen: Would be simpler to just construct a `FeatureDataSource` and make sure that you don't get an exception, instead of building the reader manually like this.

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
```diff
@@ -1733,30 +1737,197 @@ public void testTotalPloidy(final int[] ploidies, final int defaultPloidy, final
...
+
+        final VariantContextBuilder vcBuilder = new VariantContextBuilder("","chr1", 1, 1, Arrays.asList(Aref));
+        if (writeBadVariant) { // test leniency
+            vcBuilder.attribute("fake", new Object());
+        }
+        final VariantContext vc = vcBuilder.make();
+        writer.add(vc);
     }
+
+    public SAMSequenceDictionary makeSimpleSequenceDictionary() {
```
droazen: `private`

droazen:
Did a review pass on this as well, and added my comments. Back to @cmnbroad

src/main/java/org/broadinstitute/hellbender/tools/walkers/filters/VariantFiltration.java
cmnbroad: Created a ticket https://github.com/broadinstitute/gatk/issues/2102.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
cmnbroad: Yeah. Much better. Done.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
cmnbroad: No, VariantContextWriterBuilder will derive the output type itself based on the file extension, its just that its not very forgiving and throws if it doesn't recognize they extension as a known type (i.e, a name that ends in .tmp). Perhaps I shouldn't be trying to do this at all, and we should just let it throw if such cases. Let me know what you think.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
cmnbroad: I'm reconsidering whether I should be doing this at all, see comment above.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
cmnbroad: I'm not, its just a bad name. The second arg ("false" in this case) determines whether to write the bad variant. I've separated this into two separate methods to make it clearer.

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
cmnbroad: Leaving these as is.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
lbergelson: Is that something we want this method to do?  We'd have to add additional parameters to this method to take the gqPartitions and default ploidy.  I'm not sure any tools that aren't haplotype caller would want to write a gvcf, so I think it would be ok if it just does the wrapping of the output writer manually still?

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
cmnbroad: We have a ticket (https://github.com/broadinstitute/gatk/issues/121) for the work to create and manage these automatically.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
cmnbroad: Added this note to the #2102 ticket

cmnbroad: Yeah - I suggested exactly the same thing to @droazen yesterday. It makes more sense that overloading this one.

cmnbroad:
I've addressed all the code review comments. I'm questioning the value of setting the VCF writer output type by inspecting the output file extension though. I'm wondering if I should just remove this and let htsjdk throw it can't determine the type from the extension.

lbergelson:
@cmnbroad I think the defaulting to vcf if the file extension is ambiguous is probably better behavior for gatk than crashing.   We can read a vcf with no extension, so it seems symmetrical at least.  What's the downside of including it, is it just the fairly minor added complexity?

cmnbroad:
@lbergelson Yes, I was thinking we're just making it easy to do something suboptimal (make file with a random extension). The complexity is pretty minimal - I'll leave it in since it sounds like you think its at least somewhat useful.

lbergelson:
@cmnbroad 👍  Rebase and merge when ready.  I think people expect to be able to write files with any extension they feel like, and the risks are pretty minimal.

PR #2047, opened 2016-07-26
Metrics code serialization fix for sample collectors.

cmnbroad:
This mostly consists of added "implements Serializable" to MultilLevelMetrics and it's subclasses, which wasn't needed for the existing metrics collectors but affected the Example collectors I'm adding. Also contains some miscellaneous reductions of method/field scope that were unnecessarily broad.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
+++ b/src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
@@ -151,11 +151,12 @@ public final class PerUnitInsertSizeMetricsCollector
     }
 
     /**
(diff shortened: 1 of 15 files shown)


droazen:
For @lbergelson to review

src/main/java/org/broadinstitute/hellbender/metrics/PerUnitInsertSizeMetricsCollector.java
```diff
@@ -151,11 +151,12 @@ public void addMetricsToFile(final MetricsFile<InsertSizeMetrics, Integer> file)
     }
 
     /**
-     * Combine this InsertSizeMetricsCollector with sourceCollector and return a combined InsertSizeMetricsCollector.
```
lbergelson: good catch on this doc

lbergelson:
@cmnbroad Looks good.  We can always open up those methods again if we need them to be public in the future.

PR #2053, opened 2016-07-27
fix for race condition in BwaSparkEngine

### LINKED ISSUE #2050: BwaSpark - Race condition in parsing sam records
in `BwaSparkEngine` the method `alignWithBwa` is like this

```
    public JavaRDD<GATKRead> alignWithBWA(final JavaSparkContext ctx, final JavaRDD<GATKRead> unalignedReads, final SAMFileHeader readsHeader) {
        //Note: SparkContext is not serializable so we don't store it in the engine and set this property here. Setting it multiple times is fine.
        // ensure reads in a pair fall in the same partition (input split), so they are processed together
        ctx.hadoopConfiguration().setBoolean(BAMInputFormat.KEEP_PAIRED_READS_TOGETHER_PROPERTY, true);

        final JavaRDD<Tuple2<ShortRead, ShortRead>> shortReadPairs = convertToUnalignedReadPairs(unalignedReads);
        final JavaRDD<String> samLines = align(shortReadPairs);
        final SAMLineParser samLineParser = new SAMLineParser(new DefaultSAMRecordFactory(), ValidationStringency.SILENT, readsHeader, null, null);
        final Broadcast<SAMLineParser> samLineParserBroadcast = ctx.broadcast(samLineParser);
        return samLines.map(r -> new SAMRecordToGATKReadAdapter(samLineParserBroadcast.getValue().parseLine(r)));
    }
```

note that the parser is distributed by broadcast and thus shared by all tasks in an executor. That's a race condition because the parser is mutable (eg the `fields` field in the coded that gets mutated for each decode call). https://github.com/broadinstitute/gatk/issues/2039 may be caused by this bug

akiezun:
fixes #2050 

@tomwhite can you have a look?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
@@ -70,9 +70,18 @@ public final class BwaSparkEngine implements Serializable{
 
         final JavaRDD<Tuple2<ShortRead, ShortRead>> shortReadPairs = convertToUnalignedReadPairs(unalignedReads);
         final JavaRDD<String> samLines = align(shortReadPairs);
-        final SAMLineParser samLineParser = new SAMLineParser(new DefaultSAMRecordFactory(), ValidationStringency.SILENT, readsHeader, null, null);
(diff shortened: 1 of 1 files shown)


akiezun:
or @droazen - can you assign someone?

droazen:
@lbergelson Do you have time to review this one today?

lbergelson:
👍 Looks good to me.

PR #2061, opened 2016-07-28
Use long fragment length as evidence.

tedsharpe:


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadClassifier.java
```diff
@@ -8,19 +8,15 @@
...
  */
 public class ReadClassifier implements Function<GATKRead, Iterator<BreakpointEvidence>> {
     @VisibleForTesting static final int MIN_SOFT_CLIP_LEN = 30; // minimum length of an interesting soft clip
     @VisibleForTesting static final int MIN_INDEL_LEN = 40; // minimum length of an interesting indel
     private static final byte MIN_QUALITY = 15; // minimum acceptable quality in a soft-clip window
-    //private static final float MIN_ZISH_SCORE = -4.f;
-    //private static final float MAX_ZISH_SCORE = 6.f;
-    //private static final float CRAZY_ZISH_SCORE = 100.f;
+    private static final float MAX_ZISH_SCORE = 6.f; // maximum fragment-length "z" score for a normal fragment
+    private static final float MIN_CRAZY_ZISH_SCORE = 100.f; // "z" score that's probably associated with a mapping error
```
cwhelan: Did you need this cap to keep from being overwhelmed still? Does this correspond to some property of our test data or is it just a guess?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
```diff
@@ -4,54 +4,82 @@
...
-        if ( readGroups.size() != statistics.size() ) throw new GATKException("Wrong number of statistics for read groups.");
-        readGroupToFragmentStatistics = new HashMap<>(SVUtils.hashMapCapacity(readGroups.size()));
-        final int nReadGroups = readGroups.size();
-        for ( int readGroupId = 0; readGroupId < nReadGroups; ++readGroupId ) {
-            readGroupToFragmentStatistics.put(readGroups.get(readGroupId).getId(), statistics.get(readGroupId));
+    @VisibleForTesting
+    ReadMetadata( final SAMFileHeader header, final ReadGroupFragmentStatistics stats ) {
+        contigNameToID = buildContigNameToIDMap(header);
+        nPartitions = 1;
+        nReads = 2;
```
cwhelan: Where do these values come from?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
```diff
@@ -120,41 +187,173 @@ public ReadMetadata read( final Kryo kryo, final Input input, final Class<ReadMe
...
+            long bases = 0L;
+            while ( readItr.hasNext() ) {
+                final GATKRead read = readItr.next();
+                reads += 1L;
+                bases += read.getLength();
+                if ( read.isFirstOfPair() && !read.isSecondaryAlignment() && !read.isSupplementaryAlignment() &&
+                        !read.isUnmapped() && !read.mateIsUnmapped() &&
+                        Objects.equals(read.getContig(),read.getMateContig())) {
+                    int tLen = Math.abs(read.getFragmentLength());
+                    if ( tLen > MAX_TRACKED_FRAGMENT_LENGTH ) tLen = MAX_TRACKED_FRAGMENT_LENGTH;
```
cwhelan: Could we relate the MIN_CRAZY_ZISH_SCORE above to the MAX_TRACKED_FRAGMENT_LENGHT here?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
```diff
@@ -120,41 +187,173 @@ public ReadMetadata read( final Kryo kryo, final Input input, final Class<ReadMe
...
+            int medianFragmentSize = 0;
+            while ( medianFragmentSize != counts.length ) {
+                sum += counts[medianFragmentSize];
+                if ( 2*sum >= total ) break;
+                medianFragmentSize += 1;
+            }
             this.medianFragmentSize = medianFragmentSize;
-            this.medianAbsoluteDeviationFragmentSize = medianAbsoluteDeviationFragmentSize;
+
+            sum = counts[medianFragmentSize]/2;
```
cwhelan: Maybe I'm being dense but I'm a little confused by this calculation. Are you assigning half of the reads at the median to the 'over/positive' group and half to the 'under/negative' group?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadClassifier.java
tedsharpe: Yes, the value has to be set pretty high or we get swamped.  I just felt around for a value that would produce a reasonable amount of evidence on NA12878-30X.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
tedsharpe: Just made-up stuff for testing.  I'll add args to the constructor to cover these properties, and then the magic numbers will correspond to test data properties in the testing class and will disappear from here.

tedsharpe: Just made-up stuff for testing.  I'll add args to the constructor to cover these properties, and then the magic numbers will correspond to test data properties in the testing class and will disappear from here.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
tedsharpe: Hmmm.  I don't think so.  That would require knowing the MAD, which is what we're calculating here.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
tedsharpe: That's exactly what I'm doing.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
cwhelan: Sounds good, thanks.

cwhelan:
This looks good to me. Maybe just add a few comments around the places I asked questions and is fine to go in.

PR #2084, opened 2016-08-10
DataSource for reads check for misencoded qualities and -fixMisencodedQuals option (including Spark)

magicDGS:
In the previous framework reads were checked (each 1000 reads) with the `MisencodedBaseQualityReadTransformer` for high qualities and an option was provided for fixing them.

In this PR I implemented the following:

* `MisencodedBaseQualityReadTransformer` with constructor to don't fix qualities and only check for misencoded ones.
* `ReadsDataSource.iterator()` use this transformer for checking by default, with a setter for fix them on request.
* `ReadsSparkSource.getParallelReads()` and `getADAMReads()` use this transformer for checking by default, with a setter for fix them on request.
* `@Argument` in `GATKTool` and `GATKSparkTool` for fix misencoded qualities and set it in the data sources

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
@@ -52,6 +52,10 @@ public abstract class GATKTool extends CommandLineProgram {
     @Argument(fullName="addOutputSAMProgramRecord", shortName="addOutputSAMProgramRecord", doc = "If true, adds a PG tag to created SAM/BAM/CRAM files.", optional=true)
     public boolean addOutputSAMProgramRecord = true;
 
+    @Argument(fullName = "fix_misencoded_quality_scores", shortName="fixMisencodedQuals", doc="Fix mis-encoded base quality scores", optional = true)
+    public boolean fixMissencodedQualityScores = false;
+
+
     /*
      * TODO: Feature arguments for the current tool are currently discovered through reflection via FeatureManager.
      * TODO: Perhaps we should eventually do the same auto-discovery for all input arguments (reads, reference, etc.)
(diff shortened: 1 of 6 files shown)


droazen:
I don't really like the idea of a `--fix_misencoded_quality_scores` arg at the `GATKTool` level hooked into the data source -- I'd prefer to see your read transformer PR (https://github.com/broadinstitute/gatk/pull/2085) get merged, and then make read transformers a plugin that can be turned on/off on the command line, like we can with read filters. I've asked @cmnbroad to add a comment here outlining what would have to be done to make read transformers a plugin.

magicDGS:
Thanks for the suggestion, @droazen! I did this PR before the read filter plugin was included, and actually I was thinking about remove this PR because it is very clunky.

Should I close this and open a discussion about the plugin?

cmnbroad:
It should be fairly easy to create a read transformer plugin; pretty much follow the pattern of read filter plugins: clone and modify GATKReadFilterPluginDescriptor, and add an instance of the new descriptor to the list of plugins passed in to the command line parser in (appropriate) tool base classes. And of course tests!

magicDGS:
Thanks for the feedback, @cmnbroad.  @droazen, should I open a ticket for implement the plugin and close this issue? What's about the checking of the quals?

droazen:
@magicDGS Yes, I think this should be closed in favor of a PR to make `ReadTransformers` into plugins after the merge of https://github.com/broadinstitute/gatk/pull/2085. The plugin framework implemented by @cmnbroad allows for plugin classes to contain command-line arguments, so you could add an argument to the `MisencodedBaseQualityReadTransformer` itself to control whether misencoded quals should be fixed or generate an error.

magicDGS:
Perfect. I will open an issue for this, and close this.

Issue #1886, opened 2016-06-07
Retrieving transformed reads from source


magicDGS:
It could be nice to provide a way to return the reads from source already transformed by one or several `ReadTransformer`. It will be helpful for fixing misencoded qualities or use BAQ quality transformation for API users without iterate over the reads including the transformer. Another interesting feature that could be included once this is implemented is an optional argument in `ReadInputArgumentCollection` for fix qualities or use BAQ qualities.

Use case for this features are, for instance, in the `Pileup` tool, where the output is not comparable to samtools mpileup output without arguments, because by default it use BAQ qualities, and there is no way of setting this yet without hard-coding in each tool.

magicDGS:
This is done in #2085, no?

magicDGS:
Done in https://github.com/broadinstitute/gatk/commit/12c7a2d3ba35370f6cde5c973003aff8e942ae0f for GATKTool. For the spark tools, closing in favor of https://github.com/broadinstitute/gatk/issues/3484

PR #2085, opened 2016-08-11
GATKTool pre/post ReadTransformers methods + ReadWalker/LocusWalker integration 

magicDGS:
In the same way as `makeReadFilter()`, I implemented a function to include read transformers in `ReadWalker` and `LocusWalker` to use already transformed reads in `apply` (as requested in #1886).

This function could be added to the `GATKSparkTool` too, and include it in `getReads()`.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
@@ -9,7 +9,6 @@ import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.filters.WellformedReadFilter;
 import org.broadinstitute.hellbender.transformers.ReadTransformer;
-import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.iterators.IntervalOverlappingIterator;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.locusiterator.LIBSDownsamplingInfo;
@@ -111,32 +110,6 @@ public abstract class LocusWalker extends GATKTool {
     }
 
     /**
-     * Returns the pre-filter read transformer (simple or composite) that will be applied to the reads before calling {@link #apply}.
-     * The default implementation uses the {@link ReadTransformer#identity()}.
-     * Default implementation of {@link #traverse()} calls this method once before iterating over the reads and reuses
-     * the transformer object to avoid object allocation.
-     *
-     * Subclasses can extend to provide own transformers (ie override and call super).
-     * Multiple transformers can be composed by using {@link ReadTransformer} composition methods.
-     */
-    public ReadTransformer makePreReadFilterTransformer() {
-        return ReadTransformer.identity();
-    }
-
-    /**
-     * Returns the post-filter read transformer (simple or composite) that will be applied to the reads before calling {@link #apply}.
-     * The default implementation uses the {@link ReadTransformer#identity()}.
-     * Default implementation of {@link #traverse()} calls this method once before iterating over the reads and reuses
-     * the transformer object to avoid object allocation.
-     *
-     * Subclasses can extend to provide own transformers (ie override and call super).
-     * Multiple transformers can be composed by using {@link ReadTransformer} composition methods.
-     */
-    public ReadTransformer makePostReadFilterTransformer(){
-        return ReadTransformer.identity();
-    }
-
-    /**
      * Marked final so that tool authors don't override it. Tool authors should override onTraversalStart() instead.
      */
     @Override
@@ -163,15 +136,8 @@ public abstract class LocusWalker extends GATKTool {
                                           .map(SAMReadGroupRecord::getSample)
                                           .collect(Collectors.toSet());
         final CountingReadFilter countedFilter = makeReadFilter();
-        // get the transformers
-        final ReadTransformer preTransformer = makePreReadFilterTransformer();
-        final ReadTransformer postTransformer = makePostReadFilterTransformer();
         // get the filter and transformed iterator
-        final Iterator<GATKRead> readIterator = Utils.stream(reads)
-                .map(preTransformer)
-                .filter(countedFilter)
-                .map(postTransformer)
-                .iterator();
+        final Iterator<GATKRead> readIterator = getTransformedReadStream(countedFilter).iterator();
         // get the LIBS
         final LocusIteratorByState libs = new LocusIteratorByState(readIterator, getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
         // prepare the iterator
--- a/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
@@ -8,11 +8,13 @@ import org.broadinstitute.hellbender.engine.filters.CountingReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.filters.WellformedReadFilter;
+import org.broadinstitute.hellbender.transformers.ReadTransformer;
+import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.iterators.IntervalOverlappingIterator;
-import org.broadinstitute.hellbender.utils.iterators.ReadFilteringIterator;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.locusiterator.LIBSDownsamplingInfo;
 import org.broadinstitute.hellbender.utils.locusiterator.LocusIteratorByState;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
 
 import java.util.*;
 import java.util.stream.Collectors;
@@ -22,6 +24,12 @@ import java.util.stream.StreamSupport;
  * A LocusWalker is a tool that processes reads that overlap a single position in a reference at a time from
  * one or multiple sources of reads, with optional contextual information from a reference and/or sets of
  * variants/Features.
+ *
+ * Reads will be:
+ * - Transformed with {@link #makePreReadFilterTransformer()} before filtering.
+ * - Filtered with {@link #makeReadFilter()} before post-transformers.
+ * - Transformed with {@link #makePostReadFilterTransformer()} before processing.
+ * - Include them into the {@link AlignmentContext} before passing to {@link #apply(AlignmentContext, ReferenceContext, FeatureContext)}.
  *
  * LocusWalker authors must implement the apply() method to process each position, and may optionally implement
  * onTraversalStart(), onTraversalSuccess() and/or closeTool().
@@ -77,7 +85,7 @@ public abstract class LocusWalker extends GATKTool {
     /**
      * Returns the default list of CommandLineReadFilters that are used for this tool. The filters returned
      * by this method are subject to selective enabling/disabling by the user via the command line. The
-     * default implementation uses the {@link WellformedReadFilter} and {@link ReadFilterLibrary.MappedReadFilter}filter
+     * default implementation uses the {@link WellformedReadFilter} and {@link ReadFilterLibrary.MappedReadFilter} filter
      * with all default options. Subclasses can override to provide alternative filters.
      *
      * Note: this method is called before command line parsing begins, and thus before a SAMFileHeader is
@@ -103,6 +111,32 @@ public abstract class LocusWalker extends GATKTool {
     }
 
     /**
+     * Returns the pre-filter read transformer (simple or composite) that will be applied to the reads before calling {@link #apply}.
+     * The default implementation uses the {@link ReadTransformer#identity()}.
+     * Default implementation of {@link #traverse()} calls this method once before iterating over the reads and reuses
+     * the transformer object to avoid object allocation.
+     *
+     * Subclasses can extend to provide own transformers (ie override and call super).
+     * Multiple transformers can be composed by using {@link ReadTransformer} composition methods.
+     */
+    public ReadTransformer makePreReadFilterTransformer() {
+        return ReadTransformer.identity();
+    }
+
+    /**
+     * Returns the post-filter read transformer (simple or composite) that will be applied to the reads before calling {@link #apply}.
+     * The default implementation uses the {@link ReadTransformer#identity()}.
+     * Default implementation of {@link #traverse()} calls this method once before iterating over the reads and reuses
+     * the transformer object to avoid object allocation.
+     *
+     * Subclasses can extend to provide own transformers (ie override and call super).
+     * Multiple transformers can be composed by using {@link ReadTransformer} composition methods.
+     */
+    public ReadTransformer makePostReadFilterTransformer(){
+        return ReadTransformer.identity();
+    }
+
+    /**
      * Marked final so that tool authors don't override it. Tool authors should override onTraversalStart() instead.
      */
     @Override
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -121,8 +136,12 @@ public void traverse() {
         CountingReadFilter countedFilter = disableAllReadFilters ?
                 new CountingReadFilter("Allow all", ReadFilterLibrary.ALLOW_ALL_READS ) :
                 makeReadFilter();
+        // get the transformer
+        ReadTransformer transformer = makeReadTransformer();
+        // get the transformed iterator
+        final Iterator<GATKRead> readIterator = StreamSupport.stream(reads.spliterator(), false).map(transformer).iterator();
         // get the LIBS
-        LocusIteratorByState libs = new LocusIteratorByState(new ReadFilteringIterator(reads.iterator(), countedFilter), getDownsamplingMethod(), includeDeletions(), includeNs(), keepUniqueReadListInLibs(), samples, header);
+        LocusIteratorByState libs = new LocusIteratorByState(new ReadFilteringIterator(readIterator, countedFilter), getDownsamplingMethod(), includeDeletions(), includeNs(), keepUniqueReadListInLibs(), samples, header);
```
droazen: Can you document in the class-level docs for `LocusWalker` that read transformers are applied *before* read filters?

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
```diff
@@ -72,9 +73,11 @@ public void traverse() {
         final CountingReadFilter countedFilter = disable_all_read_filters ?
                                                     new CountingReadFilter("Allow all", ReadFilterLibrary.ALLOW_ALL_READS ) :
                                                     makeReadFilter();
+        final ReadTransformer transformer = makeReadTransformer();
 
         StreamSupport.stream(reads.spliterator(), false)
                 .filter(countedFilter)
+                .map(transformer)
```
droazen: Here in `ReadWalker` you are applying the transformers after the filters, which is the opposite of the behavior in `LocusWalker`. The behavior should definitely be consistent across the two walker types. We think that the transformers should be applied before the filters in both cases, so that you have an opportunity to repair bad input that would otherwise be filtered out.

src/main/java/org/broadinstitute/hellbender/tools/walkers/bqsr/ApplyBQSR.java
```diff
@@ -47,18 +47,23 @@
...
     public void onTraversalStart() {
         outputWriter = createSAMWriter(OUTPUT, true);
-        transform = new BQSRReadTransformer(getHeaderForReads(), BQSR_RECAL_FILE, bqsrArgs);
         Utils.warnOnNonIlluminaReadGroups(getHeaderForReads(), logger);
     }
 
     @Override
     public void apply( GATKRead read, ReferenceContext referenceContext, FeatureContext featureContext ) {
-        outputWriter.addRead(transform.apply(read));
+        outputWriter.addRead(read);
```
droazen: I think that `ApplyBQSR` should continue to apply the read transformer manually, rather than via the new `GATKTool` interface, so that it can guarantee that it's applied only to filtered reads (since it could fail on malformed reads). 

There's a bit of an issue here with some read transformers that definitely should be applied before filtering (like `FixMisencodedBaseQualityReads`) to give an opportunity to repair bad input that would otherwise be filtered, and others (like `BQSRReadTransformer`) that should definitely be applied after filtering so that they don't fail on malformed input. We may eventually need a mechanism to specify whether a given transformer should be applied before or after filtering. Thoughts on this?

lbergelson: We could just leave bqsr to apply the transformer explicitly call transform manually during apply.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/SplitNCigarReads.java
```diff
@@ -119,7 +126,7 @@ public void onTraversalStart() {
 
     @Override
     public void apply(GATKRead read, ReferenceContext referenceContext, FeatureContext featureContext) {
-        splitNCigarRead(rnaReadTransform.apply(read),overhangManager);
+        splitNCigarRead(read, overhangManager);
```
droazen: Argh, so it seems like `NDNCigarReadTransformer` is another one that wants to go after read filtering rather than before, since it blows up given an invalid cigar (which would be caught by the default read filters). So out of our three current read filters, two (`NDNCigarReadTransformer` and `BQSRReadTransformer`) want to be applied after filtering, and one (`FixMisencodedBaseQualityReads`) wants to go before filtering. @lbergelson @cmnbroad can you comment on this issue?

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -6,11 +6,13 @@
...
+import org.broadinstitute.hellbender.transformers.ReadTransformer;
 import org.broadinstitute.hellbender.utils.iterators.IntervalOverlappingIterator;
 import org.broadinstitute.hellbender.utils.iterators.ReadFilteringIterator;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.downsampling.DownsamplingMethod;
 import org.broadinstitute.hellbender.utils.locusiterator.LocusIteratorByState;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
 
 import java.util.*;
 import java.util.stream.Collectors;
```
droazen: Can you create a `LocusWalkerUnitTest` and `ReadWalkerUnitTest`, and add basic tests for the new read transformer functionality to those two classes?

droazen:
Review complete, thanks for this contribution! It looks good, but there is an ordering issue of read transformers vs. read filters that we're going to have to discuss and come to a decision on before this is merged.

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
magicDGS: Sure, it was an error. I will do it in the next commit! Thanks!

magicDGS:
Thanks for the review, @droazen. My ideas about the ordering are the following: in #2084, I have the implementation of the fixing/checking of misencoded quals included in the read sources. So the only one that could be applied before filtering is gone if you accept that PR.

Thus, I guess that the transformer could be applied after the filtering. Because transformers could be kind of demanding in computation, it will be interesting to pass directly to them filtered reads. In addition, could be included in the contract that only valid reads for the tool ensure that the read transformers don't blow up.

Whatever the decision is, should I wait for addressing the comments?

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
magicDGS: Documented in the class javadoc.

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
magicDGS: Keep it as now (filter/transformer). I guess that that's why I change it here, because probably the tests were failing.

src/main/java/org/broadinstitute/hellbender/tools/walkers/bqsr/ApplyBQSR.java
magicDGS: Because I kept it with filter/transform, this depends on the final decision.

magicDGS:
I updated the branch with the new changes in the framework, and addressing the comments. Back to you, @droazen.

droazen:
I think there probably needs to be a mechanism in the code to control whether each transformer is applied before or after filtering. Transformers meant to repair bad input need to go before, and those meant to process it in some other way probably need to go after. @lbergelson @cmnbroad @magicDGS suggestions on what form this should take?

magicDGS:
The first idea that comes to my mind is using to methods: `makeReadTransformerBefore` and `makeReadTransformerAfter`, and implementations could control the order of the transformers. Other idea is create two interfaces for transform reads: `ReadRepairTransformer` and `ReadProcessingTransformer`, and make the two methods return the specific interface to do not allow one or the other to be applied before/after.

cmnbroad:
makeBefore/makeAfter seems more flexible in that it makes before/after a property of the way the transformer is applied, rather than of the transformer itself. When we write the plugin descriptor, it can maintain two separate argument lists (for "--preFilterTransformer ..." and "--postFilterTransformer ..."), and then merge them accordingly. It does complicate the tool structure, but its the price for flexibility, and the pattern is not that complex.

I think we should be explicit about what "pre" and "post" are relative to in the method and argument names, i.e., makePreReadFilterTransformer and makePostReadFilterTransformer, or maybe even one makeReadFilterTransformer method that takes a pre/post argument.

droazen:
This problem with this idea is that `ReadTransformers` will be able to be turned on/off by the user. So if a user requests that the `FixMisencodedBaseQualityReads` be turned on, the engine needs to make a decision about whether to apply it before or after the filters.

cmnbroad:
I don't see why; the user would disambiguate it by saying either "--preReadFilterTransformer FixMisencodedBaseQualityReads" "--postReadFilterTransformer FixMisencodedBaseQualityReads".

droazen:
Ah I see -- yes, that could work.

lbergelson:
`--preReadFilterTransformer FixMisencodedBaseQualityReads` <- why we need commandline auto completion

lbergelson:
We could combine the filters and transformers into 1 plugin that can intersperse them.  It would be more complicated but be maximally expressive.

droazen:
I'll take the long version over some cryptic abbreviation like `--PRFT`  :)

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
magicDGS: Done

magicDGS:
I implemented the pre/post methods and added tests for them, @droazen. I guess that this is ready for review.

droazen:
Re-assigning to @cmnbroad to review the pre/post transformer changes.

magicDGS:
Rebased to current master, @cmnbroad.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -26,6 +26,9 @@
  * one or multiple sources of reads, with optional contextual information from a reference and/or sets of
  * variants/Features.
  *
+ * Reads will be filter with {@link #makeReadFilter()} and transformed with {@link #makePostReadFilterTransformer()} before include
```
cmnbroad: "will be filter" -> "will be filtered"

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -137,6 +140,32 @@ protected final LIBSDownsamplingInfo getDownsamplingInfo() {
     }
 
     /**
+     * Returns the pre-filtered read transformer (simple or composite) that will be applied to the reads before calling {@link #apply}.
```
cmnbroad: "Returns the pre-filtered" -> "Returns the pre-filter"

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
```diff
@@ -18,6 +19,9 @@
  *
  * If multiple sources of reads are specified, they are merged together into a single sorted stream of reads.
  *
+ * Reads will be filter with {@link #makeReadFilter()} and transformed with {@link #makePostReadFilterTransformer()} before pass
+ * them to apply().
```
cmnbroad: This comment needs updating - I think it was from before we had pre and post transformers.

src/main/java/org/broadinstitute/hellbender/tools/walkers/bqsr/ApplyBQSR.java
```diff
@@ -47,18 +47,23 @@
     
     private SAMFileGATKReadWriter outputWriter;
 
-    private ReadTransformer transform;
+    /**
+     * Returns the BQSR transformer
```
cmnbroad: BQSR transformer -> BQSR post transformer.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/bqsr/ApplyBQSR.java
magicDGS: Done.

magicDGS:
I updated the javadoc with your comments @cmnbroad. Should I only rebase or also squash my previous commits? Waiting for your answer to continue working on this...

cmnbroad:
Go ahead and squash as well. thx.

magicDGS:
Now it is rebased/squashed. Back to you for the next round, @cmnbroad.

cmnbroad:
@magicDGS Can you move the make*Transformer methods up to GATKTool (at some point we'll have a plugin at that level), and then also integrate these with AssemblyRegionWalker ? We'll want to do the Spark tools as well, but we leave that for a separate PR.

magicDGS:
I moved the transformers up to GATK tool in the last commit (the previous one was fixing the barclay change in the tests).

Integraction with `AssemblyRegionWalker` is more complicated, because it requires to modify the read shard implementation to include transformers there. Maybe we should open an issue regarding that, or I can include a setter for pre/post transformers in the changes for #2372.

Back to you @cmnbroad.

droazen:
@magicDGS Sorry for the delay on this one. We do want to get this merged soon, so that we can take the next step of making these pluggable. @cmnbroad and I had a quick glance at this today, and it looks good now, just a few little changes required. @cmnbroad will add his comments to this PR shortly.

magicDGS:
Thanks for the information @droazen. I will do the changes as soon as @cmnbroad comment it!

src/test/java/org/broadinstitute/hellbender/engine/ReadWalkerUnitTest.java
```diff
@@ -0,0 +1,81 @@
...
+        };
+
+        tool.instanceMain(args);
+
+        Assert.assertEquals(tool.preTransformed, 4);
+        Assert.assertEquals(tool.postTransformed, 1);
+        Assert.assertEquals(tool.totalReads, 5);
+    }
+
+}
```
cmnbroad: I think we need more comprehensive validation in both of these sets of tests (LocusWalker and ReadWalker). We should use a transformer that actually does a transformation, and also validates that the pipeline (pre-transform->filter->post-transform) is executed in the correct order. As it stands, if we were to introduce a regression that changed the execution order of pre, filter, post, I think these would still pass.

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
```diff
@@ -117,6 +123,33 @@ SimpleInterval getReadInterval(final GATKRead read) {
...
+     *
+     * Subclasses can extend to provide own transformers (ie override and call super).
+     * Multiple transformers can be composed by using {@link ReadTransformer} composition methods.
+     */
+    public ReadTransformer makePostReadFilterTransformer(){
+        return ReadTransformer.identity();
+    }
+
+
+    /**
```
magicDGS: Done.

src/test/java/org/broadinstitute/hellbender/engine/ReadWalkerUnitTest.java
magicDGS: I modify slightly the tests to assess that the pre-transform -> filter -> post-transform order is correctly executed, in both `LocusWalker` and `ReadWalker`

magicDGS:
Back to you @cmnbroad, I think that this is almost there!

src/test/java/org/broadinstitute/hellbender/engine/LocusWalkerUnitTest.java
```diff
@@ -0,0 +1,109 @@
...
+    )
+    private static class TestTransformedLocusWalker extends LocusWalker {
+        public int preTransformed = 0;
+        public int postTransformed = 0;
+        public int totalPileup = 0;
+
+        @Override
+        public ReadTransformer makePreReadFilterTransformer() {
+
+            // transforming name of unmapped mates
```
cmnbroad: I think these comments are out-of-sync with the code ?

magicDGS: True. Fixed.

cmnbroad:
Thanks @magicDGS !

Issue #1999, opened 2016-07-13
VariantWalker should prefer the sequence dictionary from the driving variants only if an actual dictionary is present


droazen:
As of https://github.com/broadinstitute/gatk/pull/1424, `VariantWalker` prefers the sequence dictionary from the driving variants in `getBestAvailableSequenceDictionary()` over other sequence dictionaries, as it should. However, if the driving variants input does not have a declared sequence dictionary, but does have an index file, we end up creating an incomplete sequence dictionary from the index with no contig lengths and using it as the "best available" dictionary, even if better dictionaries from the reference or reads are available.

This is discussed in https://github.com/broadinstitute/gatk/pull/1424#discussion_r70674157, and is replicated in the test `ExampleVariantWalkerIntegrationTest.testExampleVariantWalker_UndefinedContigLengthsInDictionary`

PR #2091, opened 2016-08-15
For VariantWalkers, prefer the sequence dictionary from the reference…

cmnbroad:
… over the VCF index if its better. Handle sequence interval validation when no sequence length is available.

Fixes https://github.com/broadinstitute/gatk/issues/1999 and the downstream genomeLoc parser validation fallout.


src/main/java/org/broadinstitute/hellbender/engine/VariantWalker.java
```diff
@@ -57,7 +58,18 @@ void initializeFeatures() {
     public final SAMSequenceDictionary getBestAvailableSequenceDictionary() {
         final SAMSequenceDictionary dictFromDrivingVariants = drivingVariants.getSequenceDictionary();
         if (dictFromDrivingVariants != null){
-            return dictFromDrivingVariants;
+            //If this dictionary looks like it was synthesized from a feature index, see if there is
+            //a better dictionary available from another source (i.e., the reference)
+            final boolean isSyntheticDictionary = dictFromDrivingVariants
+                    .getSequences()
+                    .stream()
+                    .allMatch(seqRec -> seqRec.getSequenceLength() == SAMSequenceRecord.UNKNOWN_SEQUENCE_LENGTH);
```
droazen: Should this be `anyMatch`?

droazen: Maybe this should be extracted to a utility in the same class as the code that builds dictionaries from indices?

droazen:
One comment, then rebase/merge. Back to @cmnbroad

PR #2094, opened 2016-08-16
VariantRecalibrator and ApplyVQSR port part 1 (no integration tests).

cmnbroad:
The first commit has the raw GATK3 files, the second has the ported files. In order to minimize the diffs from GATK3 for the initial port, there are only very minimal style changes. Integration tests will follow in a separate PR.


cmnbroad:
Rebased to master to pick up sequence dictionary fixes. Also, reminder that we need to decide if this can be in the public repo.

droazen:
@cmnbroad I've obtained official confirmation that VQSR can be in the public repo.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
```diff
@@ -492,17 +493,20 @@ public void onTraversalStart() {
...
+        //TODO: this should be refactored/consolidated as part of
+        // https://github.com/broadinstitute/gatk/issues/121 and
+        // https://github.com/broadinstitute/gatk/issues/1116
         Set<VCFHeaderLine> actualLines = null;
         SAMSequenceDictionary sequenceDictionary = null;
         if (hasReference()) {
             File refFile = referenceArguments.getReferenceFile();
             sequenceDictionary= this.getReferenceDictionary();
-            actualLines = withUpdatedContigsAsLines(headerLines, refFile, sequenceDictionary, suppressReferencePath);
+            actualLines = VcfUtils.updateHeaderContigLines(headerLines, refFile, sequenceDictionary, suppressReferencePath);
```
droazen: Add unit tests for the methods you moved to `VcfUtils`

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
```diff
@@ -0,0 +1,628 @@
...
+import org.broadinstitute.hellbender.utils.variant.GATKVariantContextUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Apply a score cutoff to filter variants based on a recalibration table
```
droazen: Document the fact that the previous version could accept multiple vcf inputs, whereas this version accepts only a single input.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
```diff
@@ -0,0 +1,628 @@
...
+    @Argument(fullName="tranches_file", shortName="tranchesFile", doc="The input tranches file describing where to cut the data", optional=true)
+    private File TRANCHES_FILE;
+
+    /////////////////////////////
+    // Outputs
+    /////////////////////////////
+
+    @Argument(fullName= StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName=StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+            doc="The output recal file used by ApplyRecalibration", optional=false)
```
droazen: Description is wrong -- should be `"The output filtered and recalibrated VCF file in which each variant is annotated with its VQSLOD value"`

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
```diff
@@ -0,0 +1,628 @@
...
+    private Double TS_FILTER_LEVEL = null;
+
+    /**
+     *  Filter the input file based on allele-specific recalibration data.  See tool docs for site-level and allele-level filtering details.
+     *  Requires a .recal file produced using an allele-specific run of VariantRecalibrator
+     */
+    @Argument(fullName="useAlleleSpecificAnnotations", shortName="AS", doc="If specified, the tool will attempt to apply a filter to each allele based on the input tranches and allele-specific .recal file.", optional=true)
+    private boolean useASannotations = false;
+
+    @AdvancedOption
```
droazen: We have `@Advanced` in GATK4 -- is there a reason to prefer `@AdvancedOption`. Why do we have both?

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
```diff
@@ -0,0 +1,628 @@
...
+                    }
+                    logger.info(String.format("Read tranche " + t));
+                }
+            }
+            catch(IOException e ) {
+                throw new UserException.CouldNotReadInputFile(TRANCHES_FILE, e);
+            }
+            Collections.reverse(tranches); // this algorithm wants the tranches ordered from best (lowest truth sensitivity) to worst (highest truth sensitivity)
+        }
+
```
droazen: Is `inputNames` not getting initialized? In GATK3 it was initialized with a loop like this:

```
        for( final RodBinding rod : input ) {
            inputNames.add( rod.getName() );
        }
```

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
```diff
@@ -0,0 +1,628 @@
...
+            ignoreInputFilterSet.addAll( IGNORE_INPUT_FILTERS );
+        }
+
+        // setup the header fields
+        VCFHeader inputHeader = (VCFHeader) this.getHeaderForFeatures(getDrivingVariantsFeatureInput());
+        final Set<VCFHeaderLine> inputHeaders = inputHeader.getMetaDataInSortedOrder();
+
+        final Set<VCFHeaderLine> hInfo = new HashSet<>(inputHeaders);
+        VariantRecalibrationUtils.addVQSRStandardHeaderLines(hInfo);
+        if (useASannotations)
```
droazen: Add curly braces here

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
```diff
@@ -0,0 +1,628 @@
...
+    //---------------------------------------------------------------------------------------------------------------
+    //
+    // map
+    //
+    //---------------------------------------------------------------------------------------------------------------
+
+    @Override
+    public void apply(final VariantContext vc, final ReadsContext readsContext, final ReferenceContext ref, final FeatureContext featureContext) {
+
+        final List<VariantContext> VCs =  featureContext.getValues(getDrivingVariantsFeatureInput(), featureContext .getInterval().getStart());
```
droazen: Have you explored the possibility of using the `vc` passed into `apply()` directly, instead of doing a fresh query against the driving feature input? I imagine this query comes with a substantial performance penalty, particularly as caching is disabled for the driving variants feature input. I know that processing one `vc` at at time is not exactly equivalent to processing all vcs that start at this position, but given that the body of `apply()` just loops over the `VCs` returned from this query and processes them individually, it seems like it might be ok to make that change. Can you test it?

droazen: Also, I think with the existing implementation you'd end up processing the same variants multiple times, if there are multiple variants sharing the same start position, which is explicitly allowed in the VCF spec. Can you add a test with multiple variants with the same start? (although the best solution is just to find a way to use the `vc` passed in directly and get rid of the extra query)

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/Tranche.java
```diff
@@ -69,7 +94,46 @@ public String toString() {
...
-    private static double getRequiredDouble(Map<String,String> bindings, String key ) {
+    /**
+     * Returns an appropriately formatted string representing the raw tranches file on disk.
+     *
+     * @param tranches
+     * @return
+     */
+    public static String tranchesString( final List<Tranche> tranches ) {
+        final ByteArrayOutputStream bytes = new ByteArrayOutputStream();
+        final PrintStream stream = new PrintStream(bytes);
```
droazen: Close these streams when done

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantDataManager.java
```diff
@@ -0,0 +1,452 @@
...
+            final VariantDatum datum,
+            final boolean TRUST_ALL_POLYMORPHIC ) {
+        datum.isKnown = false;
+        datum.atTruthSite = false;
+        datum.atTrainingSite = false;
+        datum.atAntiTrainingSite = false;
+        datum.prior = 2.0;
+
+        for( final TrainingSet trainingSet : trainingSets ) {
+            List<VariantContext> vcs = featureContext.getValues(trainingSet.variantSource, targetLoc.getStart());
```
droazen: Is this right? The previous implementation just passed in `targetLoc` in its entirety, instead of just the start position. Should you be calling the `getValues(final FeatureInput<T> featureDescriptor, final SimpleInterval queryInterval)` overload instead?

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantDatum.java
```diff
@@ -11,24 +13,68 @@
...
+                    return -1;
+                } else if ( vd1.loc.getStart() > vd2.loc.getStart() ) {
+                    return 1;
+                }
+                // same start, check end
+                else if ( vd1.loc.getEnd() < vd2.loc.getEnd() ) {
+                    return -1;
+                }  else if ( vd1.loc.getEnd() > vd2.loc.getEnd() ) {
+                    return 1;
+                }
```
droazen: Can you rely on the existing `IntervalUtils.compareLocatables()` here?

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+ * adaptive error model can then be applied to both known and novel variation discovered in the call set of interest to evaluate
+ * the probability that each call is real. The score that gets added to the INFO field of each variant is called the VQSLOD. It is
+ * the log odds of being a true variant versus being false under the trained Gaussian mixture model.
+ * </p>
+ *
+ * <p>
+ * This tool performs the first pass in a two-stage process called VQSR; the second pass is performed by the
+ * <a href='https://www.broadinstitute.org/gatk/guide/tooldocs/org_broadinstitute_gatk_tools_walkers_variantrecalibration_ApplyRecalibration.php'>ApplyRecalibration</a> tool.
+ * In brief, the first pass consists of creating a Gaussian mixture model by looking at the distribution of annotation
+ * In brief, the first pass consists of creating a Gaussian mixture model by looking at the distribution of annotation
```
droazen: This line of documentation got duplicated somehow.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+ * <h3>Output</h3>
+ * <ul>
+ * <li>A recalibration table file that will be used by the ApplyRecalibration tool.</li>
+ * <li>A tranches file which shows various metrics of the recalibration callset for slices of the data.</li>
+ * </ul>
+ *
+ * <h3>Usage example</h3>
+ * <p>Recalibrating SNPs in exome data:</p>
+ * <pre>
+ * ./gatk-launch JAVA_OPTS=-Xmx4g \
```
droazen: A definite TODO is to confirm that our version actually does run in 4g with real inputs...

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+ * for more info on how to download and install R.</li>
+ * </ul>
+ *
+ */
+@CommandLineProgramProperties(
+        summary = "Build a recalibration model to score variant quality for filtering purposes",
+        oneLineSummary = "Build a recalibration model to score variant quality for filtering purposes",
+        programGroup = VariantProgramGroup.class
+)
+public class VariantRecalibrator extends VariantWalker {
```
droazen: We may need to add the ability for our version to accept multiple inputs ("to facilitate scatter-gather parallelism (to enable e.g. running on VCFs generated per-chromosome)"). I ask the relevant people whether this is a requirement or not, and get back to you here.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+     * genes), which varies greatly by the CpG and GC content of the region. See expected Ti/Tv ratios section of the GATK best
+     * practices documentation (http://www.broadinstitute.org/gatk/guide/best-practices) for more information.
+     * Normal values are 2.15 for human whole genome values and 3.2 for human whole exomes. Note
+     * that this parameter is used for display purposes only and isn't used anywhere in the algorithm!
+     */
+    @Argument(fullName="target_titv",
+            shortName="titv",
+            doc="The expected novel Ti/Tv ratio to use when calculating FDR tranches and for display on the optimization curve output figures. (approx 2.15 for whole genome experiments). ONLY USED FOR PLOTTING PURPOSES!",
+            optional=false)
+    private double TARGET_TITV = 2.15;
```
droazen: Whoops, this arg was `required=false` in GATK3, so it should be `optional=true` here.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+            optional=true)
+    private List<String> IGNORE_INPUT_FILTERS = new ArrayList<>();
+
+    @Argument(fullName="ignore_all_filters",
+            shortName="ignoreAllFilters",
+            doc="If specified, the variant recalibrator will ignore all input filters. Useful to rerun the VQSR from a filtered output file.",
+            optional=true)
+    private boolean IGNORE_ALL_FILTERS = false;
+
+    @Argument(fullName="rscriptFile", shortName="rscriptFile", doc="The output rscript file generated by the VQSR to aid in visualization of the input data and learned model", optional=true)
```
droazen: Preserve the old `fullName` (`rscript_file`) unless there's a good reason not to.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+            optional=true)
+    private boolean outputModel = false;
+
+    @Argument(fullName="model_file",
+            shortName = "modelFile",
+            doc="A GATKReport containing the positive and negative model fits",
+            optional=true)
+    private String modelReport = null;
+
+    @HiddenOption
```
droazen: We have both `@Hidden` and `@HiddenOption` -- is either one actually hooked up to the arg system?

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+    private VariantContextWriter recalWriter;
+    private PrintStream tranchesStream;
+    private ArrayList<Double> replicate = new ArrayList<>();
+    private final Set<String> ignoreInputFilterSet = new TreeSet<>();
+    private final VariantRecalibratorEngine engine = new VariantRecalibratorEngine( VRAC );
+    private ExpandingArrayList<VariantDatum> reduceSum = new ExpandingArrayList<>();
+
+    //---------------------------------------------------------------------------------------------------------------
+    //
+    // initialize
```
droazen: `initialize` -> `onTraversalStart`

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+            throw new UserException.CouldNotCreateOutputFile(TRANCHES_FILE, e);
+        }
+
+        for( FeatureInput<VariantContext> variantFeature : resource ) {
+            dataManager.addTrainingSet( new TrainingSet( variantFeature ) );
+        }
+
+        if( !dataManager.checkHasTrainingSet() ) {
+            throw new UserException.CommandLineException(
+                    "No training set found! Please provide sets of known polymorphic loci marked with the training=true ROD binding tag. For example, -resource:hapmap,VCF,known=false,training=true,truth=true,prior=12.0 hapmapFile.vcf" );
```
droazen: Remove mention of "ROD bindings" from these error messages, replacing with GATK4 terminology.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
droazen: I've confirmed that we do in fact need to implement support for multiple variant inputs in VQSR. In practice these will be strictly non-overlapping inputs, but we probably just want to go ahead and implement a general `MultiVariantWalker`, since it will likely be needed by other tools as well. I've opened a new ticket under the alpha-3 VQSR epic for this task (does not need to be addressed in this branch, obviously).

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+            sequenceDictionary = getHeaderForVariants().getSequenceDictionary();
+            if (null != sequenceDictionary) {
+                actualLines = VcfUtils.updateHeaderContigLines(hInfo, null, sequenceDictionary, true);
+            }
+            else {
+                actualLines = hInfo;
+            }
+        }
+        //TODO: integrate lenient argument
+        recalWriter = GATKVariantContextUtils.createVCFWriter(new File(output), getBestAvailableSequenceDictionary(), true);
```
droazen: Will the sequence dictionary selected above match the one returned here by `getBestAvailableSequenceDictionary()`?

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+    //---------------------------------------------------------------------------------------------------------------
+    //
+    // map
+    //
+    //---------------------------------------------------------------------------------------------------------------
+
+    @Override
+    public void apply(VariantContext vc, ReadsContext readsContext, ReferenceContext ref, FeatureContext featureContext) {
+
+        // TODO: this variant came from the drivingvariants input; do we still need to query it again for overlapping ?
```
droazen: My comment to `ApplyVQSR` applies here as well -- we should find a way to use the `vc` passed in to `apply()` directly, instead of doing a separate query against the driving input (which I think will hurt performance, and will incorrectly duplicate variants anyway if any share the same start position, unless I'm mistaken). Processing the vcs at a given start position one at a time instead of all at once should not really make a difference in the results -- if it does, we should understand exactly why. Worst case, we can accumulate variants in `apply()` until a new position is reached -- even that option is preferable to ignoring the variant passed in.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+    /////////////////////////////
+    // Private Member Variables
+    /////////////////////////////
+    private VariantDataManager dataManager;
+    private VariantContextWriter recalWriter;
+    private PrintStream tranchesStream;
+    private ArrayList<Double> replicate = new ArrayList<>();
+    private final Set<String> ignoreInputFilterSet = new TreeSet<>();
+    private final VariantRecalibratorEngine engine = new VariantRecalibratorEngine( VRAC );
+    private ExpandingArrayList<VariantDatum> reduceSum = new ExpandingArrayList<>();
```
droazen: How large does this new `List` potentially get?

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+
+        // TODO: this variant came from the drivingvariants input; do we still need to query it again for overlapping ?
+        reduceSum.addAll( addOverlappingVariants(Collections.singletonList(getDrivingVariantsFeatureInput()), true, featureContext) );
+        if( !aggregate.isEmpty()) {
+            reduceSum.addAll( addOverlappingVariants(aggregate, false, featureContext) );
+        }
+    }
+
+    /**
+     * Using the RefMetaDataTracker find overlapping variants and pull out the necessary information to create the VariantDatum
```
droazen: Update docs -- `RefMetaDataTracker` no longer exists.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+            final VariantContext vc,
+            final Allele refAllele,
+            final Allele altAllele) {
+        final VariantDatum datum = new VariantDatum();
+
+        // Populate the datum with lots of fields from the VariantContext, unfortunately the VC is too big so we just
+        // pull in only the things we absolutely need.
+        datum.referenceAllele = refAllele;
+        datum.alternateAllele = altAllele;
+        dataManager.decodeAnnotations(datum, vc, true); //BUGBUG: when run with HierarchicalMicroScheduler this is
```
droazen: "BUGBUG" comment is obsolete -- the `HierarchicalMicroScheduler` does not exist in GATK4.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+
+        // non-deterministic because order of calls depends on load of machine
+        datum.loc = (isInput ? new SimpleInterval(vc.getContig(), vc.getStart(), vc.getEnd()) : null);
```
droazen: `new SimpleInterval(vc)`

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+
+        datum.originalQual = vc.getPhredScaledQual();
+        datum.isSNP = vc.isSNP() && vc.isBiallelic();
+        datum.isTransition = datum.isSNP && GATKVariantContextUtils.isTransition(vc);
+        datum.isAggregate = !isInput;
+
+        // Loop through the training data sets and if they overlap this locus (and allele, if applicable) then update
+        // the prior and training status appropriately
+        dataManager.parseTrainingSets(featureContext, featureContext.getInterval(), vc, datum, TRUST_ALL_POLYMORPHIC);
```
droazen: `featureContext.getInterval()` is not at all equivalent to the previous `context.getLocation()`, which was a single-base location. Bug?

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,852 @@
...
+                }
+                return true;
+            }
+            catch (Exception e) {
+                //TODO: see https://github.com/broadgsa/gatk-protected/pull/19
+                // If we retain this loop, it needs to be restructured. Currently, any existing state (including data
+                // already written to output streams) is retained for the next attempt. The output streams should
+                // really be re-created and all state should be reset. Also, because we're catching Exception, this
+                // will retry even if we have bad input, such as poorly annotated training data that results in an
+                // IllegalArgumentException.
```
droazen: We should include a test to make sure this retry feature works as intended in GATK4.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibratorArgumentCollection.java
```diff
@@ -2,6 +2,8 @@
 
 
 import org.broadinstitute.hellbender.cmdline.Argument;
+import org.broadinstitute.hellbender.utils.commandline.AdvancedOption;
+import org.broadinstitute.hellbender.utils.commandline.HiddenOption;
 
 /*
  * A bunch of annotations for VQSR.
```
droazen: annotations -> arguments

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibratorArgumentCollection.java
```diff
@@ -21,6 +23,15 @@
...
+     */
+    @Argument(fullName="useAlleleSpecificAnnotations",
+            shortName="AS",
+            doc="If specified, the variant recalibrator will attempt to use the allele-specific versions of the specified annotations.", optional=true)
+    public boolean useASannotations = false;
+
+    /**
      * This parameter determines the maximum number of Gaussians that should be used when building a positive model
      * using the variational Bayes algorithm.
      */
```
droazen: `MAX_GAUSSIANS` should be `@Advanced`

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -588,6 +592,40 @@ public static double log10SumLog10(final double[] log10Values, final int start,
...
+    /**
+     * Calculate f(x) = x^2
+     * @param x the value to square
+     * @return x * x
+     */
+    public static double square(final double x) {
+        return x * x;
+    }
+
+    public static double distanceSquared(final double[] x, final double[] y) {
```
droazen: Unit test for this method?

src/main/java/org/broadinstitute/hellbender/utils/samples/SampleUtils.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Sample utilities class.
+ */
+public class SampleUtils {
+    /**
+     * Given a list of files with sample names it reads all files and creates a list of unique samples from all these files.
+     * @param files list of files with sample names in
+     * @return a collection of unique samples from all files
+     */
+    public static Collection<String> getSamplesFromFiles (Collection<File> files) {
```
droazen: Test for this method?

src/main/java/org/broadinstitute/hellbender/utils/variant/VcfUtils.java
```diff
@@ -40,5 +43,69 @@ private VcfUtils(){}
...
+    /**
+     * Given a set of VCF header lines, update the set with contig
+     * lines from the provided reference dictionary.
+     * @param oldLines
+     * @param referenceFile
+     * @param refDict
+     * @param referenceNameOnly
+     * @return Updated list of VCF header lines.
+     */
+    public static Set<VCFHeaderLine> updateHeaderContigLines(
```
droazen: Tests for these methods

src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantGaussianMixtureModelUnitTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.broadinstitute.hellbender.utils.text.XReadLines;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.util.*;
+
+public final class VariantGaussianMixtureModelUnitTest extends BaseTest {
+    private static int N_VARIANTS = 100;
+    VariantDatum[] variantData1 = new VariantDatum[N_VARIANTS];
```
droazen: `private`

src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibratorModelOutputUnitTest.java
```diff
@@ -0,0 +1,161 @@
...
+        goodGaussianList.add(goodGaussian2);
+
+        List<MultivariateGaussian> badGaussianList = new ArrayList<>();
+        badGaussianList.add(badGaussian1);
+
+        GaussianMixtureModel goodModel = new GaussianMixtureModel(goodGaussianList, shrinkage, dirichlet, priorCounts);
+        GaussianMixtureModel badModel = new GaussianMixtureModel(badGaussianList, shrinkage, dirichlet, priorCounts);
+
+        if (printTables) {
+            System.out.println("Good model mean matrix:");
```
droazen: Tests should use a logger for output.

src/test/java/org/broadinstitute/hellbender/utils/collections/ExpandingArrayListUnitTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.testng.Assert;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+
+import java.util.Arrays;
+
+/**
+ * Basic unit test for RecalData
```
droazen: `RecalData` -> `ExpandingArrayList`

droazen:
Review complete -- back to @cmnbroad

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
cmnbroad: Done here and for VarRecal.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
cmnbroad: They look redundant. I removed all references to the "option"-suffixed ones from this PR and created a ticket to remove the existing usages and remove the containing package. https://github.com/broadinstitute/gatk/issues/2130.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
cmnbroad: Yes, inputNames is currently unreferenced. Removed.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/Tranche.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: Wow. The CLP thinks its provided since it has an initial value. Anyway, fixed.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/samples/SampleUtils.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/variant/VcfUtils.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantDatum.java
cmnbroad: Yes. Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: No. Changed these to Hidden, and created https://github.com/broadinstitute/gatk/issues/2130 to fix the previously existing references.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: Done. I don't see any others beyond these.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: No, good point. I unified this code so all the branches use the same sequence dictionary, which is at least consistent. But I would like to define and provide a unified implementation of this code in another pr so that we're consistent in how we combine the selected sequence dictionary and the (optionally user specified) reference to populate the contig and other metadata lines in the output header (per https://github.com/broadinstitute/gatk/issues/2112). For example, getBestAvailableSequenceDictionary prefers the dict from the variants, but if a reference is provided, *should* we update the output contig lines to reflect the reference like this (GATK3 code) does, even if we're using the dict from the variants ?

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibratorArgumentCollection.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantGaussianMixtureModelUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibratorModelOutputUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/utils/collections/ExpandingArrayListUnitTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibratorArgumentCollection.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: It gets pretty large (it scales with the number of variants used by the model and the number of annotations specified). Its no different than in GATK3 though, since its just replacing the state previously kept in the map accumulator. At the end of traversal they're the same length in both GATK3 and GATK4.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantDataManager.java
cmnbroad: I think this is right, if not a little confusing. In the previous implementation, targetLoc was a GenomeLoc, so I think this matches the behavior when queried against the FeatureContext. Using SimpleInterval here is unnecessary though. Will  clean this up.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: Changed this to accumulate and queue variantContexts with the same start-location, and then use the saved vc/featureContext pairs directly, followed by a single call to query the overlapping aggregates once for the group, which I believe matches the behavior of GATK3.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: This is used to query variants the overlap this location and start at the same position, so I think this is equivalent. Having sad that, I don't see anything in GATK3 that constrains this be a single-base location (getLocation returns a GenomeLoc, which has a start and a stop). It seems like it would be useful to have an explicit (type-level) distinction between a single locus and an interval to make usages like this less ambiguous, if it doesn't already exist.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
cmnbroad: Changed to use the passed-in vc directly.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,873 @@
...
+    // Private Member Variables
+    /////////////////////////////
+    private VariantDataManager dataManager;
+    private VariantContextWriter recalWriter;
+    private PrintStream tranchesStream;
+    private ArrayList<Double> replicate = new ArrayList<>();
+    private final Set<String> ignoreInputFilterSet = new TreeSet<>();
+    private final VariantRecalibratorEngine engine = new VariantRecalibratorEngine( VRAC );
+    private ExpandingArrayList<VariantDatum> reduceSum = new ExpandingArrayList<>();
+    private Queue<ImmutablePair<VariantContext, FeatureContext>> variantsAtLocus = new LinkedList<>();
```
droazen: replace with `ArrayList`

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -0,0 +1,873 @@
...
+        datum.loc = (isInput ? new SimpleInterval(vc) : null);
+
+        datum.originalQual = vc.getPhredScaledQual();
+        datum.isSNP = vc.isSNP() && vc.isBiallelic();
+        datum.isTransition = datum.isSNP && GATKVariantContextUtils.isTransition(vc);
+        datum.isAggregate = !isInput;
+
+        // Loop through the training data sets and if they overlap this locus (and allele, if applicable) then update
+        // the prior and training status appropriately
+        dataManager.parseTrainingSets(featureContext, vc, datum, TRUST_ALL_POLYMORPHIC);
```
droazen: Add a comment that `parseTrainingSets` is only using the start position of the `featureContext`

cmnbroad: Add a comment here that the start loc is retrieved insisde the method.

droazen:
:+1: squash/rebase/merge when ready

PR #2095, opened 2016-08-16
Add gradle distArchive to create a zip containing a standalone GATK distribution + launcher + README

### LINKED ISSUE #1987: Create gradle target that creates distribution-ready GATK archive with launcher script, spark and standard jars, etc.

droazen:
Resolves #1987

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,12 @@ If you are looking for the codebase of the current production version of GATK, p
     * The resulting jar will be in `build/libs` with a name like `gatk-package-VERSION-spark.jar`
     * This jar will not include Spark and Hadoop libraries, in order to allow the versions of Spark and Hadoop installed on your cluster to be used.
 
+* To create a zip archive containing a complete standalone GATK distribution, including our launcher `gatk-launch`, both the local and spark jars, and this README, run
+
+        ./gradlew gatkZipDistribution
+        
+    * The resulting zip file will be in `build` with a name like `gatk-VERSION.zip`
+
 * To remove previous builds, run 
 
         ./gradlew clean
--- a/build.gradle
+++ b/build.gradle
@@ -353,6 +353,27 @@ task sparkJar(type: ShadowJar) {
     classifier = 'spark'
 }
 
+task gatkZipDistribution(type: Zip) {
+    dependsOn shadowJar, sparkJar
+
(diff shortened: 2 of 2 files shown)


droazen:
@lbergelson please review

build.gradle
```diff
@@ -353,6 +353,21 @@ task sparkJar(type: ShadowJar) {
...
+
+    baseName = project.name + "-" + project.version
+    destinationDir file("$buildDir")
+    archiveName baseName + ".zip"
+
+    from(shadowJar.archivePath)
+    from(sparkJar.archivePath)
+    from("settings.gradle")
+    from("gatk-launch")
+    from("README.md")
```
droazen: I can't find a way to get gradle to blow up if one of the files to be included in the archive does not exist -- instead, it happily creates an incomplete archive. @lbergelson do you happen to know of a way to do this?

lbergelson: you could add something like: 
```
    doFirst {
        assert file'gatk-launch").exists()
        assert file("settings.gradle").exists()
        assert file("README.md").exists()
        ...
    }
```

I don't think it's necessary to check that the archives exist since their failure to exist would have been a task failure on the earlier tasks.  It might be worth doing for the other files though.

README.md
```diff
@@ -64,6 +64,12 @@ If you are looking for the codebase of the current production version of GATK, p
     * The resulting jar will be in `build/libs` with a name like `gatk-package-VERSION-spark.jar`
     * This jar will not include Spark and Hadoop libraries, in order to allow the versions of Spark and Hadoop installed on your cluster to be used.
 
+* To create a zip archive containing a complete standalone GATK distribution, including our launcher `gatk-launch`, both the local and spark jars, and this README, run
+
+        ./gradlew distArchive
```
lbergelson: I think this name is not that great.  I'm not sure what's better though, `package`, `createStandalone`, `distComplete`....  not sure.  I don't think it would be terrible to override distZip and make it do what we want instead of something kind of useless..

droazen: How about something really distinctive and unique like `gatkZipDistribution` to set it apart from all the generic-sounding `dist`/`jar` targets?

lbergelson: 👍

PR #2096, opened 2016-08-16
removing gcov references from build.gradle

lbergelson:
We forgot to take this out when we removed native code compilation


droazen:
:+1:

PR #2097, opened 2016-08-16
upgrade to newest shiniest gradle 3.0

lbergelson:
updating wrapper from 2.13 -> 3.0

disable daemon on travis since it's now enabled by default and gradle recommends disabling it on CI servers
remove jacoco version specification since 3.0 specifies a reasonable version by default
update the test result html path on travis since it changed in 3.0


droazen:
:+1:

PR #2098, opened 2016-08-17
New AFCalculator

davidbenjamin:
@droazen I plugged this into the `HaplotypeCaller` integration tests and it looks really good. The concordance tests fail due to a handful of false positives but these are *all* obvious true positive multiallelics that the old qual model behaves stupidly on.

I'm not sure who the right reviewer for this is but I'm guessing it's @ldgauthier and/or @vruano.


davidbenjamin:
A couple unrelated tests failing; hopefully a rebase will clear those up.

vruano:
Will take a look.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeAlleleCounts.java
```diff
@@ -634,4 +648,25 @@ GenotypeAlleleCounts copy() {
...
+    public interface TwoIntConsumer {
+        void accept(final int alleleIndex, final int alleleCount);
+    }
+
+    public void forEachAlleleIndexAndCount(final TwoIntConsumer action) {
+        new IndexRange(0, distinctAlleleCount).forEach(n -> action.accept(sortedAlleleCounts[2*n], sortedAlleleCounts[2*n+1]));
+    }
+
+    @FunctionalInterface
+    public interface TwoIntToDoubleFunction {
```
vruano: Following java.util.function apparent convention I would call this class ```IntToDoubleBiFunction```

vruano: Perhaps this interface should be in a utility package?.... I guess it may be used in some other context.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeAlleleCounts.java
```diff
@@ -634,4 +648,25 @@ GenotypeAlleleCounts copy() {
                         }).collect(Collectors.toList());
     }
 
+
+
+    @FunctionalInterface
+    public interface TwoIntConsumer {
```
vruano: This one would be ```IntBiConsumer```

vruano: Same as below, should it be a top level class in some utilities package?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeCalculationArgumentCollection.java
```diff
@@ -37,6 +38,12 @@ public GenotypeCalculationArgumentCollection( final GenotypeCalculationArgumentC
     }
 
     /**
+     * Use the new allele frequency / QUAL score model
+     */
+    @Argument(fullName = "useNewAFCalculator", shortName = "newQual", doc = "If provided, we will use the new AF model instead of the so-called exact model", optional = true)
+    public boolean USE_NEW_AF_CALCULATOR = false;
```
vruano: With HaplotypeCaller one can choose the implementation using the ```pnrm``` argument which is defined in ```StandardCallerArgumentCollection```. Unfortunately this is not the case for ```GenotypeGVCFs```, this does not make much sense. Could you please add this argument to ```GenotypeGVCFs``` so that the user can use the same argument to select the implementation by name (old or new)? 

Perhaps you want to get ```StandardCallerArgumentCollection``` to delegate the argument definition to ```GenotypeCalculationArgumentCollection``` as is likely that you bump into trouble getting ```GenotypeGVCFs``` to absorve all the arguments in ```StandardCallerArgumentCollection```.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/AFCalculator.java
```diff
@@ -85,7 +85,7 @@ protected abstract AFCalculationResult computeLog10PNonRef(final VariantContext
      * @param maximumAlternativeAlleleCount the maximum alternative allele count it must be able to handle. Has no effect if
      *                                     the current tracker is able to handle that number.
      *
-     * @return never {@code null}
+     * @return never {@code null}, except in the case of {@link NewAFCalculator}, whioch doesn't use a state tracker
```
vruano: What about:

```
* @return {@code null} iff this calculator implementation does not use a state tracker.
```

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * @author David Benjamin &lt;davidben@broadinstitute.org&gt;
+ */
+public final class NewAFCalculator extends AFCalculator {
```
vruano: Why not call it ```VariationalBayesAFCaculator```; Is not going to be new for ever, is it?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
```diff
@@ -0,0 +1,165 @@
...
+    private static final GenotypeLikelihoodCalculators GL_CALCS = new GenotypeLikelihoodCalculators();
```
vruano: ```GL_CALCS``` is used only to access static methods? in that case just make reference to ```GenotypeLikelihoodCalculators``` directly each time. Could I ask you to make ```GenotypeLikelihoodCalculators``` not instantiable (define a private default constructor and make the class final).

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
```diff
@@ -0,0 +1,165 @@
...
+     * Compute the probability of the alleles segregating given the genotype likelihoods of the samples in vc
+     *
+     * @param vc the VariantContext holding the alleles and sample information.  The VariantContext
+     *           must have at least 1 alternative allele
+     * @param refSnpIndelPseudocounts a total hack.  A length-3 vector containing Dirichlet prior pseudocounts to
+     *                                be given to ref, alt SNP, and alt indel alleles.  Hack won't be necessary when we destroy the old AF calculators
+     * @return result (for programming convenience)
+     */
+    @Override
+    public AFCalculationResult getLog10PNonRef(final VariantContext vc, final int defaultPloidy, final int maximumAlternativeAlleles, final double[] refSnpIndelPseudocounts) {
```
vruano: ```refSnpIndelPseudocounts``` is never used, why is it declared?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
```diff
@@ -0,0 +1,165 @@
...
+
+    public NewAFCalculator(final double refPseudocount, final double snpPseudocount, final double indelPseudocount, final int defaultPloidy) {
+        this.refPseudocount = refPseudocount;
+        this.snpPseudocount = snpPseudocount;
+        this.indelPseudocount = indelPseudocount;
+        this.defaultPloidy = defaultPloidy;
+    }
+
+    public AFCalculationResult getLog10PNonRef(final VariantContext vc) {
+        return getLog10PNonRef(vc, defaultPloidy, 0, null);
```
vruano: The default max alt alleles is 0?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
```diff
@@ -0,0 +1,165 @@
...
+            final double[] genotypePosteriors = normalizedGenotypePosteriors(g, glCalc, log10AlleleFrequencies);
+
+            //the total probability
+            log10PNoVariant += Math.log10(genotypePosteriors[HOM_REF_GENOTYPE_INDEX]);
+
+            // per allele non-log space probabilities of zero counts for this sample
+
+
+            // for each allele calculate the total probability of genotypes containing at least one copy of the allele
+            double[] pOfNonZeroAltAlleles = new double[numAlleles];
```
vruano: shouldn't pOfNonZeroAltAlleles be final? actually I think is surprising that you can access it from within the lambda below if it is not final, but it must be compiling.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
```diff
@@ -0,0 +1,165 @@
...
+            new IndexRange(0, glCalc.genotypeCount()).forEach(genotypeIndex ->
+                    glCalc.genotypeAlleleCountsAt(genotypeIndex).forEachAlleleIndexAndCount((alleleIndex, count) ->
+                            pOfNonZeroAltAlleles[alleleIndex] += genotypePosteriors[genotypeIndex]));
+
+            new IndexRange(0, numAlleles).forEach(a -> log10POfZeroCountsByAllele[a] += Math.log10(1 - pOfNonZeroAltAlleles[a]));
```
vruano: I personally disapprove the overuse of streaming. 
Here a traditional for(a = 0; a < length; a++) {} make more sense, not only because is probably more efficient but because is more readable too.

Beside I would use IntStream.range() instead.

src/main/java/org/broadinstitute/hellbender/utils/Dirichlet.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @author David Benjamin &lt;davidben@broadinstitute.org&gt;
+ */
+public class Dirichlet {
+    final double[] alpha;
+
+    public Dirichlet(final double[] alpha) {
+        Utils.nonNull(alpha);
+        Utils.validateArg(alpha.length >= 1, "Dirichlet parameters must have at least one element");
+        this.alpha = Arrays.copyOf(alpha, alpha.length);
```
vruano: ```this.alpha = alpha.clone()```.

vruano: There is no constaints for alpha values (is NaN, +-Inf or negetative values acceptable?)

vruano: Make the single parameter constructor delegate on the two argument constructor.

src/main/java/org/broadinstitute/hellbender/utils/Dirichlet.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public Dirichlet(final Dirichlet prior, final double[] counts) {
+        Utils.nonNull(counts);
+        Utils.validateArg(counts.length == prior.size(), "Counts and prior must have same length.");
+        alpha = IntStream.range(0, prior.size()).mapToDouble(n -> prior.alpha[n] + counts[n]).toArray();
+    }
+
+    public static Dirichlet symmetricDirichlet(final int numStates, final double concentration) {
```
vruano: doc.

vruano: argument verification.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
```diff
@@ -0,0 +1,165 @@
...
+        final double[] dummyFlatPrior = {-1e-10, -1e-10};   //TODO: HACK must be negative for AFCalcResult
+        final double[] log10PosteriorOfNoVariantYesVariant = {log10PNoVariant, Math.log10(1 - Math.pow(10, log10PNoVariant))};
+
+        return new AFCalculationResult(integerAltAlleleCounts, alleles, log10PosteriorOfNoVariantYesVariant, dummyFlatPrior, log10PRefByAllele);
+    }
+
+    private double[] effectiveAlleleCounts(final VariantContext vc, final double[] log10AlleleFrequencies) {
+        final int numAlleles = vc.getNAlleles();
+        Utils.validateArg(numAlleles == log10AlleleFrequencies.length, "number of alleles inconsistent");
+        double[] result = new double[numAlleles];
```
vruano: ```final```

vruano:
Done with the review.
Back to @davidbenjamin.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeAlleleCounts.java
davidbenjamin: done and done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeAlleleCounts.java
davidbenjamin: done and done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeCalculationArgumentCollection.java
davidbenjamin: `GenotypeGVCFs` is in gatk-protected now.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/AFCalculator.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
davidbenjamin: Pretty soon I'm hoping it can be called `AFCalculator` since the others will no longer be with us.   For now, `VariationalBayes` is too much of an implementation detail for my taste.  How about `AlleleFrequencyCalculator`, since the new model infers the population frequencies of the alleles?

vdauwera: I like the idea of a name different enough to make it clear that it's very different from the old implementation, but generic enough to avoid being overly specific. 👍 for `AlleleFrequencyCalculator`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
davidbenjamin: `GenotypeLikelihoodCalculators::getInstance` is not a static method -- it depends on `genotypeTableByPloidy`, which is filled when a `GenotypeLikelihoodCalculators` object is instantiated.  Maybe this should all be refactored but it appears sufficiently non-trivial that I think it belongs in a separate PR.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
davidbenjamin: Yeah, I apologize for the hacks required to make it conform to `AFCalculator`.  Hopefully we'll be able to eliminate `StateTracker` and friends soon and clean this all up.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
davidbenjamin: `maxAltAlleles` is never used (more conforming to the signature).  I added a comment that the `0` is just a dummy value so as not to raise eyebrows.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
davidbenjamin: done, and come to think of it I don't understand why it compiled, either.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
davidbenjamin: Done.  Thank you for being the voice of reason.

src/main/java/org/broadinstitute/hellbender/utils/Dirichlet.java
davidbenjamin: done and done; done; I instead had the two-parameter delegate to the one-parameter constructor, but I think this is in the spirit of your suggestion

src/main/java/org/broadinstitute/hellbender/utils/Dirichlet.java
davidbenjamin: done and done

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/NewAFCalculator.java
davidbenjamin: done

davidbenjamin:
Back to @vruano

src/main/java/org/broadinstitute/hellbender/utils/Dirichlet.java
```diff
@@ -22,19 +23,25 @@
...
+        Utils.validateArg(MathUtils.allMatch(alpha, x -> x >= 0), "Dirichlet parameters may not be negative");
+        Utils.validateArg(MathUtils.allMatch(alpha, Double::isFinite), "Dirichlet parameters must be finite");
+        this.alpha = alpha.clone();
     }
 
     public Dirichlet(final Dirichlet prior, final double[] counts) {
-        Utils.nonNull(counts);
-        Utils.validateArg(counts.length == prior.size(), "Counts and prior must have same length.");
-        alpha = IntStream.range(0, prior.size()).mapToDouble(n -> prior.alpha[n] + counts[n]).toArray();
+        this(MathArrays.ebeAdd(prior.alpha, counts));
```
vruano: a ```null``` prior would results in a NPE.

src/main/java/org/broadinstitute/hellbender/utils/functionalinterfaces/IntBiConsumer.java
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.hellbender.utils.functionalinterfaces;
```
vruano: functional or function is enough.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeCalculationArgumentCollection.java
vruano: Ok, I thought that that was the tool you were updating... so what tool you using to execute the new code? HC? The unified approach to select implemenetation would still apply.

vruano:
Done. Back to @davidbenjamin.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeCalculationArgumentCollection.java
davidbenjamin: Yes, I have been testing in `HaplotypeCaller` i.e. in the GATK4 integration tests.

So right now the toggle for the new qual model is in `GenotypeCalculationArgumentCollection` while the `AFCalculatorImplementation` switch `pnrm` is still in `StandardCallerArgumentCollection`.  This is not unified and is in fact quite clumsy but it seems to me that the alternative is to make the new model fit into the whole apparatus of `AFCalculatorImplementation`, `AFCalculatorProvider` etc.  Since this is all going to be torn down pretty soon I personally prefer the temporary clumsiness over enmeshing the new model in code that will soon be deleted.

If you see things differently let me know and I will put everything under `pnrm`.

src/main/java/org/broadinstitute/hellbender/utils/Dirichlet.java
davidbenjamin: Fixed by deleting the method after making it unused.

src/main/java/org/broadinstitute/hellbender/utils/functionalinterfaces/IntBiConsumer.java
davidbenjamin: done

davidbenjamin:
@vruano back to you -- unless you want me to unify the command line switch as in above comment.

vruano:
If the old implementations are going to go out of the window soon then it may not be worth the effort to unify the command line as eventually there would be no option at all. If this is the case you would need to add the corresponding issue once this PR has been merged. However if these alternative implementation are going to coexist for a while then I rather you unify the cmdline now.

vruano:
Otherwise is good to go and once test pass you can merge.
Back to @davidbenjamin.

davidbenjamin:
>If the old implementations are going to go out of the window soon then it may not be worth the effort to unify the command line as eventually there would be no option at all. If this is the case you would need to add the corresponding issue once this PR has been merged. However if these alternative implementation are going to coexist for a while then I rather you unify the cmdline now.

@ldgauthier Do you have an idea of the timeline and/or required milestones for getting rid of the exact model `AFCalculator`s?

ldgauthier:
@davidbenjamin One factor is that @droazen wanted to make sure nothing changed until the GATK4 HaplotypeCaller got the stamp of approval.  I'm very pleased with the test results so far.  Still interested in the reference sample increase results.  I'd also want to QC a medium-sized callset to make sure the changes make sense, but then I'd be happy.  What about you @vdauwera ?

davidbenjamin:
> Still interested in the reference sample increase results.

I wrote a unit test for that.  QUAL decreases slowly as ref samples increase (e.g. one million ref samples decrease QUAL by about log_10 10^6 = 6), which I believe is the correct behavior.  Overwhelming evidence that a variant is rare *ought* to make any genotyper more skeptical about it.  Note that this will only lose variants with very weak evidence.

As for the command line unification that @vruano suggested, I put in a few hours and it's trickier than expected.  If we use `pnrm` then we have to add an instance of `AFCalculatorImplementation`, which would need a `Supplier<AlleleFrequencyCalculator>`.  But unlike the old exact model `AFCalculator`s the new one knows the heterozygosity priors (needed for Dirichlet prior pseudocounts) and needs to get those in its constructor.  But this would mean that `AFCalculatorImplementation` would have to know about the heterozygosity command line arguments which would would in turn demand that it be refactored as something other than an `Enum`.

As inelegant as the current hack of having a separate argument that overrides `pnrm` is, I think it's actually cleaner than the alternative.

lh3:
> I'm very pleased with the test results so far

What is the test result?

ldgauthier:
@lh3 I gave David some specific examples I had encountered of multiallelic variants where one or more of the alleles gets dropped using the independent alleles model.  The new model preserves the expected alleles.

@davidbenjamin I am perfectly happy with a QUAL decrease of 6 for an additional million reference samples.  That's much less dramatic than the current implementation:
image

In terms of the args, given that it's most complicated than expected, I'm okay with leaving it as is and cleaning it up when we remove the other models.

David, do you have a sense of what the values are for the old model versus new, maybe from the test cases I gave you?  I want to make sure our 30 emission cutoff is still reasonable.

davidbenjamin:
> David, do you have a sense of what the values are for the old model versus new, maybe from the test cases I gave you? I want to make sure our 30 emission cutoff is still reasonable.

@ldgauthier For single-sample biallelics quals are very similar to the old model, so the cutoff of 30 should be fine.  For multi-sample calls the new model is smarter about inferring allele frequency but I expect this effect to be minor.  The definition of qual is the same in both cases, so to the extent that both models are essentially correct (as they both are except for multiallelics) they will agree closely.

lh3:
I haven't understood how multi-allele model exactly works in the old GATK, so can't comment on why it does not perform well. In general, I am supportive of making the new model the default going forward. However:

> when we remove the other models

I would suggest retaining the old model if possible. As I said on the method meeting, the old model takes the full power of population information (by full, I mean under the Wright-Fisher and HWE assumptions, you can't derive a more powerful model in theory). My understanding is that David's current model isn't. This is fine as long as the information from sequence data overwhelms the population information, which is usually true for highCov data. However, when data is thin, the population information will play a more important role. Without thorough evaluations in multiple scenarios, it is not clear when the loss of population information in the new model starts to matter. It would be good to keep the old model as a reference point, at least for biallelic SNPs, until we have more comparison.

ldgauthier:
The independent alleles model has some pretty pernicious error modes, but I understand the value of the "exact" implementation of the old model.  I definitely want to default to the new model, but I understand how other users in the community might still be doing studies with large(ish) numbers of low coverage samples.  It's probably worth keeping the old implementations for them, but usually I let @vdauwera advocate for the users.

Great job @davidbenjamin !  I'm really excited about this!!!  Right now I'm trying to decide which callsets I want to re-run first.  :)

PR #2100, opened 2016-08-17
updating to htsjdk 2.5.0 -> 2.6.0

lbergelson:
updating test files due to change in VCF FT field from htsjdk


lbergelson:
figure we should upgrade to the latest before we release

droazen:
:+1: merge when tests pass

PR #2101, opened 2016-08-17
NIO countreads and supporting code

jean-philippe-martin:


src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -374,6 +374,17 @@ public static String getPathWithoutBucket(String path) {
 
     }
 
+    /**
+     * String -> Path. This *should* not be necessary (use Paths.get(URI.create(...)) instead) , but it currently is
+     * on Spark because using the fat, shaded jar breaks the registration of the GCS FilesystemProvider.
```
lbergelson: Sadness.  Can we fix that somehow?

jean-philippe-martin: I expect we can, yes. If nothing else by manually fixing META-INF/services.
Ideally this would be done automatically by the jar shader.

jean-philippe-martin: Upon investigation it looks like META-INF/services is fine (and the jar
shader's doing the right thing). The problem's elsewhere.

lbergelson: How do the FileSystem providers get registered?  Is there a plugin loader that searches the class path for them or do they register statically on class loading?  Is it possible that there's a class loader issue where spark initializes the FileSystems before it's loaded the classes in the client jar?

jean-philippe-martin: [The FileSystems API page](https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystems.html) describes the mechanism. If our jar file is not in the classpath but is instead loaded dynamically, this could explain the filesystem not being properly installed.

droazen:
Re-assigning to @lbergelson for review, since I'm about to leave on vacation.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
jean-philippe-martin: There doesn't seem to be an easy way around this. It looks related to the way Spark (YARN) starts processes. I could find an [identical question](http://stackoverflow.com/questions/38728479/how-can-i-use-a-jsr-203-file-system-provider-with-apache-spark) on StackOverflow but it has no answer.

lbergelson: @tomwhite Do you have any insight into this?

src/main/java/org/broadinstitute/hellbender/utils/nio/ChannelAsSeekableStream.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Reads the next byte of data from the input stream. The value byte is
+     * returned as an <code>int</code> in the range <code>0</code> to
+     * <code>255</code>. If no byte is available because the end of the stream
+     * has been reached, the value <code>-1</code> is returned. This method
+     * blocks until input data is available, the end of the stream is detected,
+     * or an exception is thrown.
+     * <p>
+     * <p> A subclass must provide an implementation of this method.
```
lbergelson: I think you can cut this line about subclasses out.

src/main/java/org/broadinstitute/hellbender/utils/nio/ChannelAsSeekableStream.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @return the next byte of data, or <code>-1</code> if the end of the
+     * stream is reached.
+     * @throws IOException if an I/O error occurs.
+     */
+    @Override
+    public int read() throws IOException {
+        ByteBuffer buf = ByteBuffer.allocate(1);
```
lbergelson: Is this going to be a performance issue to be allocating a new Buffer with every call to `read()`?  I assume most uses of SeekableStream are going to use the read(byte[]...) but we might want to just preallocate a single buffer that gets reused here?

src/main/java/org/broadinstitute/hellbender/utils/nio/NioBam.java
```diff
@@ -0,0 +1,161 @@
...
+
+    transient byte[] indexCache;
+
+    /** Checks the files exists, then stores them. **/
+    public NioBam(String gcsFilename, String indexGcsFilename) throws IOException {
+        this.bam = gcsFilename;
+        this.index = indexGcsFilename;
+        Path bamPath = BucketUtils.getPathOnGcs(bam);
+        Path bamIndexPath = BucketUtils.getPathOnGcs(index);
+        if (!Files.exists(bamPath)) {
```
lbergelson: What's the latency for Files.exists(bamPath) like?

src/main/java/org/broadinstitute/hellbender/utils/nio/ChannelAsSeekableStream.java
jean-philippe-martin: yup.

src/main/java/org/broadinstitute/hellbender/utils/nio/ChannelAsSeekableStream.java
jean-philippe-martin: I don't expect anyone performance-conscious to repeatedly read a single byte at a time, and avoiding a shared buffer makes the method thread-safe.
But since you're worried I've changed it to allocate at most once. I don't expect code will be calling read() from multiple threads anyways.

src/main/java/org/broadinstitute/hellbender/utils/nio/NioBam.java
jean-philippe-martin: About 200-300ms. The point is that it's a lot less than the latency of starting a Spark job only to have it fail because of a typo in a filename.

This may move somewhere else (especially since I understand there's work on a path class of our own) but I think that the key idea of having a class to hold paths, with the invariant that you can only create one of those if the file exists, is worth keeping. It also means that we do the check only once, since everyone who receives such an object knows the path has been checked already.

lbergelson: Definitely better to find out your file doesn't exist sooner rather than later.  The fabled pull request to add our own path class has and always will be 1 week away.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -358,14 +358,14 @@ public static boolean isFileUrl(String path) {
...
     /**
-     * Given a path of the form "gcs://bucket/folder/folder/file", returns "bucket".
+     * Given a path of the form "gs://bucket/folder/folder/file", returns "bucket".
      */
     public static String getBucket(String path) {
         return path.split("/")[2];
     }
 
     /**
-     * Given a path of the form "gcs://bucket/folder/folder/file", returns "folder/folder/file".
```
lbergelson: good catch

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
```diff
@@ -0,0 +1,141 @@
...
+ */
+public class ReadsIterable implements Iterable<SAMRecord>, Serializable {
+
+    private static final long serialVersionUID = 1L;
+    final String path;
+    final byte[] index;
+    final QueryInterval interval;
+    final boolean removeHeader = true;
+
+    class ReadsIterator implements Iterator<SAMRecord> {
```
lbergelson: We've had mysterious serialization issues with non-static inner classes.  Would you mind making this static?

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleNioCountReads.java
```diff
@@ -0,0 +1,94 @@
...
+    private void countReads(JavaSparkContext ctx) throws IOException {
+
+        try {
+            outputStream = OUTPUT_FILE != null ? new PrintStream(OUTPUT_FILE) : System.out;
+        }
+        catch ( FileNotFoundException e ) {
+            throw new UserException.CouldNotReadInputFile(OUTPUT_FILE, e);
+        }
+
+        NioBam input = new NioBam(path, path + ".bai");
```
lbergelson: We'll need to check for both `path.bam.bai` and `path.bai`.  NioBam should probably have a constructor that takes only the input path and locates the appropriate index.

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
```diff
@@ -0,0 +1,141 @@
...
+        final int BUFSIZE = 200 * 1024 * 1024;
```
lbergelson: Would anyone ever want to tune this?  I.e. should it be exposed somewhere?

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
lbergelson: These iterators should probably be `CloseableIterator` since they own file handles.

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
```diff
@@ -0,0 +1,141 @@
...
+        private SamReader bam;
+        private long count = 0;
+        private SAMRecordIterator query;
+        private SAMRecord nextRecord = null;
+        private boolean done = false;
+
+        public ReadsIterator() throws IOException {
+            List<SAMRecord> ret = new ArrayList<>();
```
lbergelson: ret isn't used anywhere as far as I can tell.

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
```diff
@@ -0,0 +1,141 @@
...
+    private static final long serialVersionUID = 1L;
+    final String path;
+    final byte[] index;
+    final QueryInterval interval;
+    final boolean removeHeader = true;
+
+    class ReadsIterator implements Iterator<SAMRecord> {
+        final int BUFSIZE = 200 * 1024 * 1024;
+        private SamReader bam;
+        private long count = 0;
```
lbergelson: nothing actually accesses count

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
```diff
@@ -0,0 +1,141 @@
...
+        private SAMRecordIterator query;
+        private SAMRecord nextRecord = null;
+        private boolean done = false;
+
+        public ReadsIterator() throws IOException {
+            List<SAMRecord> ret = new ArrayList<>();
+            Path fpath = BucketUtils.getPathOnGcs(path);
+            byte[] data = index;
```
lbergelson: data is a funny name for the index

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
```diff
@@ -0,0 +1,141 @@
...
+            SeekableStream indexInMemory = new ByteArraySeekableStream(data);
+            SeekableByteChannelPrefetcher chan2 = new SeekableByteChannelPrefetcher(Files.newByteChannel(fpath), BUFSIZE);
```
lbergelson: what happened to chan 1?

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
jean-philippe-martin: Changed to `CloseableIterator`.

jean-philippe-martin: Are you *really sure* you want to switch to a static inner class? This class is meant to not be static. The conversion would create some redundant, surprising-looking code.

Those issues you saw were perhaps due to all non-static inner class holding a reference to their parent class. This is certainly surprising if you don't know that's going on, and if the parent class isn't designed to serialize can lead to problems.

Here instead, the design calls for the parent class to be serialized, and it's written with that in mind. I would prefer to leave it as is if at all possible.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleNioCountReads.java
jean-philippe-martin: I was trying to reduce the surprise of having the code access a file you didn't specify. But, ok.

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
jean-philippe-martin: Good catch, thank you!

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
jean-philippe-martin: It was useful for debugging. It's gone now.

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
jean-philippe-martin: renamed to `indexData`.

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
jean-philippe-martin: Renamed to `chan`.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -0,0 +1,390 @@
...
+     * the {@link StandardOpenOption#APPEND APPEND} option, in
+     * which case the position is first advanced to the end. The entity to which
+     * the channel is connected is grown, if necessary, to accommodate the
+     * written bytes, and then the position is updated with the number of bytes
+     * actually written. Otherwise this method behaves exactly as specified by
+     * the {@link WritableByteChannel} interface.
+     *
+     * @param src
+     */
+    @Override
```
lbergelson: This documentation is lies for this class :)

lbergelson: Or at least deceptive, it should say that writing isn't supported.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -0,0 +1,390 @@
...
+     * connected to an entity, typically a file, opened with the {@link
+     * StandardOpenOption#APPEND APPEND} option.
+     *
+     * @param size The new size, a non-negative byte count
+     * @return This channel
+     * @throws NonWritableChannelException If this channel was not opened for writing
+     * @throws ClosedChannelException      If this channel is closed
+     * @throws IllegalArgumentException    If the new size is negative
+     * @throws IOException                 If some other I/O error occurs
+     */
```
lbergelson: This should just say that it's not supported

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -0,0 +1,390 @@
...
+ * SeekableByteChannelPrefetcher wraps an existing SeekableByteChannel to add prefetching.
+ * The prefetching is done on a different thread, so you can write simple code that repeatedly
+ * calls read() to get data, processes it, and then calls read again -- and yet this
+ * simple code overlaps computation and communication for you.
+ * (Of course this is only worthwhile if the underlying SeekableByteChannel doesn't already
+ * implement prefetching).
+ */
+public final class SeekableByteChannelPrefetcher implements SeekableByteChannel {
+
+    final SeekableByteChannel chan;
```
lbergelson: Could you minimize the visibility of the class members.  I don't think we want anyone reaching in and touching our channel if we can avoid it.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -0,0 +1,390 @@
...
+    final int bufSize;
+    final ExecutorService exec;
+    final long size;
+    List<WorkUnit> full = new ArrayList<>();
+    WorkUnit fetching = null;
+    // total number of buffers
+    final int BUF_COUNT = 2;
```
lbergelson: This constant should probably be made static.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
jean-philippe-martin: Fixed :)

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
jean-philippe-martin: Done.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
jean-philippe-martin: Made private.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
jean-philippe-martin: Done.

jean-philippe-martin:
OK, so to be clear - should I press the "merge" button?

droazen:
@jean-philippe-martin @lbergelson is away for several weeks -- if you've addressed all his comments then please go ahead and squash/rebase/merge!

jean-philippe-martin:
squashed + rebased. Will merge once tests are green.

Issue #1280, opened 2015-12-08
ReadSparkSource.getHeader silently manufactures a bogus SAMFileHeader for ADAM files


cmnbroad:
And probably other kinds of files too. The stack below results from handing it a .ADAM file in the MeanQualityByCycleSparkIntegrationTest.test_ADAM test. The ReadSparkSource code is currently delegating to Hadoop-BAM, which is in turn delegating to promiscuous htsjdk code that says anything that doesn't look like known file type [must be a SAM file](https://github.com/samtools/htsjdk/blob/bd92747fd3672635de96473fea6d4b38e8635c8e/src/java/htsjdk/samtools/SAMFileReader.java#L754). It then happily creates a bogus SAMFileHeader from the .ADAM stream.

All 3 layers should probably be more discriminating.

This currently doesn't break any tests. I discovered it when running the HB tests against a local version of htsjdk with a strict setHeader implementation that attempts to resolve all reference names on every setHeader call. That code caused this test to fail because its using the bogus header.

"main@1" prio=5 tid=0x1 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
	  at htsjdk.samtools.SAMTextHeaderCodec.decode(SAMTextHeaderCodec.java:113)
	  at htsjdk.samtools.SAMTextReader.readHeader(SAMTextReader.java:200)
	  at htsjdk.samtools.SAMTextReader.<init>(SAMTextReader.java:63)
	  at htsjdk.samtools.SAMTextReader.<init>(SAMTextReader.java:73)
	  at htsjdk.samtools.SAMFileReader.init(SAMFileReader.java:684)
	  at htsjdk.samtools.SAMFileReader.<init>(SAMFileReader.java:148)
	  at org.seqdoop.hadoop_bam.util.SAMHeaderReader.readSAMHeaderFrom(SAMHeaderReader.java:66)
	  at org.seqdoop.hadoop_bam.util.SAMHeaderReader.readSAMHeaderFrom(SAMHeaderReader.java:47)
	  at org.broadinstitute.hellbender.engine.spark.datasources.ReadsSparkSource.getHeader(ReadsSparkSource.java:195)
	  at org.broadinstitute.hellbender.engine.spark.GATKSparkTool.initializeReads(GATKSparkTool.java:284)
	  at org.broadinstitute.hellbender.engine.spark.GATKSparkTool.initializeToolInputs(GATKSparkTool.java:264)
[...]

PR #2105, opened 2016-08-19
addresses issue #1280 from GATK3 (issue was originally found in GATK3)

pjfan:
addresses issue https://github.com/broadinstitute/gsa-unstable/issues/1280 from GATK3 (issue was originally found in GATK3 but same issue was present in GATK4) now interval padding works for exclude intervals


droazen:
@lbergelson please review

lbergelson:
@pjfan Looks good to me.  Merge when tests pass.

lbergelson:
There were other tests that checked the existing behavior.  Either update those tests with the new behavior, or possibly create a new argument for exclusion padding.  That would be more expressive than the newly introduced behavior that makes it impossible to represent certain intervals at all when you have padding specified.

lbergelson:
@pjfan When you refer to issues in another repo you have to include the entire url to the issue or it will end linking to the incorrect one like it did in your comment.

PR #2106, opened 2016-08-19
updating GenomicsDB from 0.2.1 -> 0.3.0

lbergelson:
this includes the new VCF2TileDB java API
replacing process builder calls to GenomicsDB binaries with usage of VCF2TileDB api in GenomicsDBTestUtils

these tests will likely now fail on power systems since GenomicsDB doesn't bundle a power library, power users should build genomics db and specify it's path on the LD_PATH

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,7 +92,7 @@ compileTestJava {
 }
 
 dependencies {
-    compile 'com.intel:genomicsdb:0.2.1'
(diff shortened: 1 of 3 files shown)


lbergelson:
Look how much cleaner these tests are with the loader api.  It would have saved me so much pain if we'd had it a month ago :/

droazen:
:+1: looks good, merging

PR #2111, opened 2016-08-23
Fix CreateSequenceDictionary (issue #2093)

### LINKED ISSUE #2093: CreateSequenceDictionary generates a compress dict file that can't be used in other GATK tools
If I generate a sequence dictionary with the `CreateSequenceDictionary` tool with GATK4 command line (`./gatk-launch CreateSequenceDictionary -R reference.fasta -O reference.dict`), the generated dictionary is compressed (`cat` points out that is a binary file), and while using it in tools that calls `getReferenceDictionary()` with a provided reference the dictionary is null.

magicDGS:
Fix #2093: `CreateSequenceDictionary` now generates a plain-text dictionary as expected by tools that use it.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
@@ -1,9 +1,6 @@
 package org.broadinstitute.hellbender.tools.picard.sam;
 
-import htsjdk.samtools.SAMFileHeader;
-import htsjdk.samtools.SAMFileWriter;
-import htsjdk.samtools.SAMSequenceDictionary;
-import htsjdk.samtools.SAMSequenceRecord;
+import htsjdk.samtools.*;
 import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.reference.ReferenceSequenceFile;
 import htsjdk.samtools.reference.ReferenceSequenceFileFactory;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
```diff
@@ -83,7 +80,8 @@ protected Object doWork() {
         SAMFileWriter samWriter = null;
         //This writes the header with sequenceDictionary
         try {
-            samWriter = createSAMWriter(OUTPUT, REFERENCE_SEQUENCE, samHeader, false);
+            // do not use createSAMWriter because it does not create a plain text file without a .sam extension
+            samWriter = new SAMFileWriterFactory().makeSAMWriter(samHeader, false, OUTPUT);
```
droazen: Can you add a test proving that the original issue is fixed?

lbergelson: There's an existing test called `CreateSequenceDictionaryIntegrationTest.testBasic()` but it's grossly deficient and doesn't actually assert anything other than that it doesn't crash.  If you could update that to check that it produced the expected sequence dictionary that would be great.

magicDGS: It's what I'm doing! Commit in one minute!

magicDGS:
Back to you with the added test, @droazen and/or @lbergelson! Thanks for having a look to this!

lbergelson:
👍 Thanks for finding this issue and fixing it.

PR #2118, opened 2016-08-25
fix jbwaIntegrationTest so it doesn't skip on ppc

lbergelson:
previously jbwaIntegrationTest was being skipped if running on ppc
this was a hold over from when we didn't yet have a ppc library for it
now it will always run and fail if it's on an unsupported system
it should be compatible with most linux, osx, and power systems now though

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
@@ -4,41 +4,24 @@ import com.github.lindenb.jbwa.jni.AlnRgn;
 import com.github.lindenb.jbwa.jni.BwaIndex;
 import com.github.lindenb.jbwa.jni.BwaMem;
 import com.github.lindenb.jbwa.jni.ShortRead;
-import org.broadinstitute.hellbender.utils.NativeUtils;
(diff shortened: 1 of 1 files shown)


lbergelson:
@cmnbroad Could you review this when you get a chance?  With no David I think you get to be the lucky target of miscellaneous reviews.

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
```diff
@@ -4,41 +4,24 @@
...
-            throw new SkipException("jbwa not available for this architecture");
-        }
-    }
-
     @Test
     public void testJBWAIsLoadable(){
-        skipJBWATestOnUnsupportedPlatforms();
-
         BWANativeLibrary.load();
     }
```
cmnbroad: Do we still need this test ? Its a subset of the others.

cmnbroad:
One minor question/comment. Feel free to merge.

src/test/java/org/broadinstitute/hellbender/JBWAIntegrationTest.java
lbergelson: I think it's useful to have "did the native library load yes/no" in the test output.  If this one fails, then you know where the problem is.  If it passes but others don't, then you have a more complicated situation to look at.

Issue #2033, opened 2016-07-22
Test infrastructure for framework users


magicDGS:
I'm using GATK4 as a framework to implement my own tools, and it will be nice to have a way of perform integration tests using `IntegrationTestSpec`. Nevertheless, it requires the extension of the `CommandLineProgramTest` to run the command, and thus it is extending `BaseTest`.

The issues that this infrastructure generates when trying to use this test classes are the following:

* `BaseTest` loading of `GenomeLocParser` is annotated with `@BeforeClass`, which throws an error because the reference genome (hg19MiniReference) is not present in the repository.
* `CommandLineProgramTest` is using `org.broadinstitute.hellbender.Main` for running the commands, but for custom tools the instanceMain with a different list of packages. Although this could be solved by extending the class by another abstract class.

I propose (and I can implemented if you agree) the following:
* `CommandLineProgramTest` not implementing `BaseTest`.
* `CommandLineProgramTest` as a real abstract class without implementations of `getTestDataDir()` or `runCommandLine()`
* Abstract `GATKCommandLineProgramTest` extending both `CommandLineProgramTest` and `BaseTest`, sited in `org.broadinstitute.hellbender.utils.test` and used in all integrations tests in this repository and the protected repository.

magicDGS:
Related with #2125

PR #2122, opened 2016-08-26
CommandLineProgramTestInterface for API users

magicDGS:
This is a very minimal change of the testing framework to allow users of the framework to use `IntegrationTestSpec` with their own classes.

It solves the problem of a custom  `Main` class to run the command line test in programs using the framework (through overriding default behavior), and the loading of `GenomeLocParser` by the `BaseTest` if the test is simply extending `CommandLineProgramTest`. More details for this issue in #2033.

Now API users could implements and modify default behavior of `CommandLineProgramTestInterface` and use this test classes in `IntegrationTestSpec`.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
@@ -1,0 +1,94 @@
+package org.broadinstitute.hellbender.utils.test;
+
+import htsjdk.samtools.util.Log;
+import org.broadinstitute.hellbender.Main;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+import org.broadinstitute.hellbender.utils.logging.BunnyLog;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Utility interface for CommandLine Program testing.
+ */
+public interface CommandLineProgramTestInterface {
+
+    /**
+     * Returns the name for the tested tool.
+     */
+    public String getTestedToolName();
+
+    /**
+     * For testing support. Given a name of a Main CommandLineProgram and it's arguments, builds the arguments appropriate for calling the
+     * program through Main.
+     *
+     * Default behaviour uses {@link #makeCommandLineArgs(List, String)} with the tool name provided by {@link #getTestedToolName()}.
+     *
+     * @param args List<String> of command line arguments
+     * @return String[] of command line arguments
+     */
+    default String[] makeCommandLineArgs(final List<String> args) {
+        return makeCommandLineArgs(args, getTestedToolName());
+    }
+
+    /**
+     * For testing support. Given a name of a Main CommandLineProgram and it's arguments, builds the arguments appropriate for calling the
+     * program through Main.
+     *
+     * Default behaviour generates a command line in the form "toolname args", with the verbosity parameter returned by {@link #injectDefaultVerbosity(List)}.
+     *
+     * @param args     List<String> of command line arguments
+     * @param toolname name of the tool to test
+     * @return String[] of command line arguments
+     */
+    default String[] makeCommandLineArgs(final List<String> args, final String toolname) {
+        List<String> curatedArgs = injectDefaultVerbosity(args);
+        final String[] commandLineArgs = new String[curatedArgs.size() + 1];
+        commandLineArgs[0] = toolname;
+        int i = 1;
+        for (final String arg : curatedArgs) {
+            commandLineArgs[i++] = arg;
+        }
+        return commandLineArgs;
+    }
+
+    /**
+     * Inject the verbosity parameter into the list.
+     *
+     * Default behaviour look for --verbosity argument; if not found, supply a default value that minimizes the amount of logging output.
+     */
+    default List<String> injectDefaultVerbosity(final List<String> args) {
+
+        // global toggle for BunnyLog output.
+        BunnyLog.setEnabled(false);
+
+        for (String arg : args) {
+            if (arg.equalsIgnoreCase("--" + StandardArgumentDefinitions.VERBOSITY_NAME) || arg.equalsIgnoreCase("-" + StandardArgumentDefinitions.VERBOSITY_NAME)) {
+                return args;
+            }
+        }
+        List<String> argsWithVerbosity = new ArrayList<>(args);
+        argsWithVerbosity.add("--" + StandardArgumentDefinitions.VERBOSITY_NAME);
+        argsWithVerbosity.add(Log.LogLevel.ERROR.name());
+        return argsWithVerbosity;
+    }
+
+    /**
+     * Runs the command line implemented by this test.
+     *
+     * Default behaviour uses {@link Main} with the command line arguments created by {@link #makeCommandLineArgs(List)}.
+     */
+    default Object runCommandLine(final List<String> args) {
+        return new Main().instanceMain(makeCommandLineArgs(args));
+    }
+
+    default Object runCommandLine(final String[] args) {
+        return runCommandLine(Arrays.asList(args));
+    }
+
+    default Object runCommandLine(final ArgumentsBuilder args) {
+        return runCommandLine(args.getArgsList());
+    }
+
+}
--- a/src/main/java/org/broadinstitute/hellbender/CommandLineProgramTest.java
+++ b/src/main/java/org/broadinstitute/hellbender/CommandLineProgramTest.java
@@ -1,7 +1,7 @@
 package org.broadinstitute.hellbender;
 
 import org.broadinstitute.hellbender.utils.test.BaseTest;
-import org.broadinstitute.hellbender.utils.test.CommandLineProgramTestInterface;
+import org.broadinstitute.hellbender.utils.test.CommandLineProgramTester;
 
 import java.io.File;
 import java.util.List;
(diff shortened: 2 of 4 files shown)


lbergelson:
@magicDGS Thanks for this pull request!  This is one of our blindspots since it only affects people who are external to the Broad, thank you for pointing it out and remedying it. 

We've been talking about exposing the package list that `Main` looks for as a properties file in the jar.  That would make it so there's no need to subclass `Main`.  Would that be an alternative solution to this problem?

magicDGS:
@lbergelson, I don't think that this solution will help in this  case, because another error when trying to use `CommandLineProgramTest`is that it extends `BaseTest`, which loads directly a `GenomeLocParser` for a reference that is not present and it blows up in every test.

Regarding the `Main` class, because you point it out here, I would like to have some control over `Main` and how it manages things like errors or logging header. Basically all the things that I'm facing at the moment are, apart of this error using the testing framework, is that the framework have tons of mentions to the GATK itself (error messages pointing to the GATK manual page or bundle tools), and little control over which of them should be expose to the final user. Only as an example, I would like to output a line with the name and version of my software and a short notice about the usage of the GATK framework and which version I'm using (for easier maintenance, and contribution if a bug is found).

lbergelson:
Ah, ok, so you would extend `Main` anyway even if we had a way to control which packages it looks in?  That's good to know.

Can you not use base test because of that GenomeLocParser?  We should fix that.

magicDGS:
I can't use it because of that, and it have lots of variables that I'm not using. I'm doing my own base test class, but I'd love to have a more general base test class in GATK to extend, without that many variables specific for this repository. Thanks a lot for your interest!

Should I do something for this PR? 

So should I

lbergelson:
Just trying to understand what the various issues are.  

I opened a ticket to remove the GenomeLocParser #2125.  GenomeLocParser is essentially a deprecated holdover from gatk3 that we haven't made the effort to completely remove yet.  

Would it be useful if we packaged these small test files in the jar and had them automatically extracted when tests run?  That would provide a basic set of example files that someone developing new tools could play with.  

I think we mostly use BaseTest as a convenient way to find common files and a place to put common test functions like the `getTemp` family.  Those are mostly statics that could be called from anywhere though.  The only other stuff it really does is set up the the test spark context and set test logging verbosity.  Moving the common files out of BaseTest might be a hard sell for people here, but we could extract a superclass that does the important set up and has the various functions without having the project specific files.  We're a bit understaffed at the moment due to vacations, but I'll bring it up once people are back in the office.

src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
```diff
@@ -0,0 +1,94 @@
...
+        argsWithVerbosity.add(Log.LogLevel.ERROR.name());
+        return argsWithVerbosity;
+    }
+
+    /**
+     * Runs the command line implemented by this test.
+     *
+     * Default behaviour uses {@link Main} with the command line arguments created by {@link #makeCommandLineArgs(List)}.
+     */
+    default Object runCommandLine(final List<String> args) {
```
lbergelson: Should this one be abstract so that it forces people to implement it with their version of Main?

src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
```diff
@@ -0,0 +1,94 @@
...
+import org.broadinstitute.hellbender.Main;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+import org.broadinstitute.hellbender.utils.logging.BunnyLog;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Utility interface for CommandLine Program testing.
```
lbergelson: Could you add a comment explaining why someone would implement this?  It's hard to understand the point if you're not thinking about multiple `Main` implementations.

src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+        return commandLineArgs;
+    }
+
+    /**
+     * Inject the verbosity parameter into the list.
+     *
+     * Default behaviour look for --verbosity argument; if not found, supply a default value that minimizes the amount of logging output.
+     */
+    default List<String> injectDefaultVerbosity(final List<String> args) {
```
lbergelson: Are you expecting to override this, or should it just be made into a static method?

src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
```diff
@@ -0,0 +1,94 @@
...
+     * For testing support. Given a name of a Main CommandLineProgram and it's arguments, builds the arguments appropriate for calling the
+     * program through Main.
+     *
+     * Default behaviour generates a command line in the form "toolname args", with the verbosity parameter returned by {@link #injectDefaultVerbosity(List)}.
+     *
+     * @param args     List<String> of command line arguments
+     * @param toolname name of the tool to test
+     * @return String[] of command line arguments
+     */
+    default String[] makeCommandLineArgs(final List<String> args, final String toolname) {
```
lbergelson: Should this be static or is it something you imagine having to override?

src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
```diff
@@ -0,0 +1,94 @@
...
+import org.broadinstitute.hellbender.utils.logging.BunnyLog;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Utility interface for CommandLine Program testing.
+ */
+public interface CommandLineProgramTestInterface {
```
lbergelson: The name isn't great.  It's hard for me to think of a better one though.  `CommandLineProgramTester` is the obvious alternative, but it's not really very good either.  I'll think about it...

lbergelson:
Out of curiousity, couldn't you achieve the same thing already by subclassing CommandLineProgramTest and overriding `makeCommandLineArgs` and `runCommandLine`?

magicDGS:
I did this PR because of the GenomeLocParser; what you suggested was the first thing that I tried. I did this instead of refactoring the BaseTest because I didn't want to cause major changes in the code.

lbergelson:
Oh, right, of course, that makes sense.

src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
magicDGS: I make it abstract just in case someone have a different verbosity level. The same for the rest of the abstract methods, I don't know what other people will need to override in the future.

src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
magicDGS: It can be abstract, but if at the end you change the Main to be used by whatever other program, will be nice to have a default one for the API user that don't care about another main.

src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
magicDGS: Whenever you have an idea for this, just let me know. The alternative that you suggested sounds better than this, so I changed it in the next commit.

src/main/java/org/broadinstitute/hellbender/utils/test/CommandLineProgramTestInterface.java
magicDGS: Sure, added although if you have a different suggestion for the explanation will be great (my English and programming skills aren't the best).

magicDGS:
Back to you @lbergelson, waiting for your suggestion!

cmnbroad:
I've actually been lobbying to get rid of IntegrationTestSpec completely. Although it probably once added value, it has some [problems](https://github.com/broadinstitute/gatk/issues/1562) that limit its usefulness, and the remaining functionality (like expected exception checking) is available directly in the test framework. We have quite a few tests that don't use it at all, and it would be nice to only support one style, so I'd be inclined to discourage rather than encourage it, and focus on resolving the BaseTest issues.

magicDGS:
Thanks for your feedback, @cmnbroad. In my case, I think that `IntegrationTestSpec` is a good way of avoid complicated code to test tool results, but it is true that it have some problems (one that I had was the usage for testing programs where the outputs are determined by a prefix in the command line, but with different suffixes).

I think, from the API user point of view, that a class like `IntegrationTestSpec` to facilitate program output testing (including user exceptions) will be nice for developing purposes. Nevertheless, this is just a convenience that I asked for here, but I can try to solve the issues with the `BaseTest` instead.

By the way, I would love to have this interface in GATK at least for now, because several of my tools rely on the `IntegrationTestSpecs` for development...

droazen:
:+1: on this one from me, although I think we should eventually move away from use of `IntegrationTestSpec`

lbergelson:
👍 to this with the understanding that we may be removing `IntegrationTestSpec` in the future.

magicDGS:
Thanks @droazen and @lbergelson!

PR #2124, opened 2016-08-26
Changed twoPassReadsWalker to reinitialize the reads to ensure equivalent behavior across SAM formats

### LINKED ISSUE #2117: Make the TwoPassReadsWalker accept all types of input files
The twoPassReadsWalker currently only iterates twice through the reads as intended for BAM files. If given a text SAM file as input it will fail to iterate through the file a second time. This has to do with the SAMTextReader not using a seekable stream for the SAM file in htsjdk, which proves to be complicated to fix given the current api implementation there. A workaround for this is to reinitialize ReadsDataSource between passes through the input for twoPassReadsWalker and test that it works for all types of read files.

jamesemery:
resolves #2117

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
@@ -34,11 +34,9 @@ public abstract class TwoPassReadWalker extends ReadWalker {
         traverseReads(countedFilter, this::firstPassApply);
         logger.info("Finished First Pass");
         afterFirstPass();
-        // Need to reinitilize the reads and intervals so they are guarinteed to pass over a file
+        // Need to reinitialize the reads and intervals so they are guaranteed to pass over a file
         initializeReads();
-        if ( hasIntervals() ) {
-            reads.setTraversalBounds(intervalArgumentCollection.getTraversalParameters(getHeaderForReads().getSequenceDictionary()));
-        }
+        setReadTraversalBounds();
         logger.info("Starting SecondPass");
         traverseReads(countedFilter, this::secondPassApply);
 
--- a/src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
@@ -34,6 +34,11 @@ public abstract class TwoPassReadWalker extends ReadWalker {
         traverseReads(countedFilter, this::firstPassApply);
         logger.info("Finished First Pass");
         afterFirstPass();
+        // Need to reinitilize the reads and intervals so they are guarinteed to pass over a file
+        initializeReads();
+        if ( hasIntervals() ) {
+            reads.setTraversalBounds(intervalArgumentCollection.getTraversalParameters(getHeaderForReads().getSequenceDictionary()));
+        }
         logger.info("Starting SecondPass");
         traverseReads(countedFilter, this::secondPassApply);
 
--- a/src/test/java/org/broadinstitute/hellbender/engine/TwoPassReadsWalkerUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/engine/TwoPassReadsWalkerUnitTest.java
@@ -1,0 +1,82 @@
+package org.broadinstitute.hellbender.engine;
+
+
+import org.broadinstitute.hellbender.CommandLineProgramTest;
+import org.broadinstitute.hellbender.cmdline.CommandLineParser;
+import org.broadinstitute.hellbender.cmdline.CommandLineProgramProperties;
+import org.broadinstitute.hellbender.cmdline.TestProgramGroup;
+import org.broadinstitute.hellbender.engine.spark.GATKSparkToolIntegrationTest;
+import org.broadinstitute.hellbender.exceptions.UserException;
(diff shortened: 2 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
```diff
@@ -34,6 +34,11 @@ public void traverse() {
         traverseReads(countedFilter, this::firstPassApply);
         logger.info("Finished First Pass");
         afterFirstPass();
+        // Need to reinitilize the reads and intervals so they are guarinteed to pass over a file
```
lbergelson: typos:  reinitialize and guaranteed

src/main/java/org/broadinstitute/hellbender/engine/TwoPassReadWalker.java
```diff
@@ -34,6 +34,11 @@ public void traverse() {
...
+        initializeReads();
+        if ( hasIntervals() ) {
```
lbergelson: could you extract this as a package private function `setReadTraversalBounds()` in ReadWalker since it's just a copy from their?

src/test/java/org/broadinstitute/hellbender/engine/TwoPassReadsWalkerUnitTest.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Test(dataProvider = "extensions")
+    public void testDifferentFormatEquivalentBehavior(String ext) throws IOException {
+        final TwoPassReadsWalkerUnitTest.dummyTwoPassReadsWalker tool = new TwoPassReadsWalkerUnitTest.dummyTwoPassReadsWalker();
+        final CommandLineParser clp = new CommandLineParser(tool);
+
+        final String[] args = {
+                "-I", getTestDataDir()+ "/count_reads" + ext,
+                "-R", getTestDataDir()+ "/count_reads.fasta"
```
lbergelson: do you need a reference for this?

src/test/java/org/broadinstitute/hellbender/engine/TwoPassReadsWalkerUnitTest.java
```diff
@@ -0,0 +1,82 @@
...
+        };
+
+
+        clp.parseArguments(System.out, args);
+        tool.onStartup();
+        tool.doWork();
+        tool.onShutdown();
+
+        Assert.assertTrue(tool.everythingWorked());
```
lbergelson: instead of asserting tool.everythingWorked(), could you check the individual values instead? .

i.e. `Assert.assertEquals(firstPassCount, 8)` etc... I think that would be a little clearer.

lbergelson: It also solves the problem that in the current situation you might accidentally be reading 0 in both passes and not noticing.

src/test/java/org/broadinstitute/hellbender/engine/TwoPassReadsWalkerUnitTest.java
```diff
@@ -0,0 +1,82 @@
...
+        public boolean everythingWorked() {
+            if (firstPass != secondPass || !betweenTraversals) {
+                return false;
+            }
+            return true;
+        }
+    }
+
+    @Override
+    public String getTestedClassName() {
```
lbergelson: Not actually using this since you're not using runCommandLine.

src/test/java/org/broadinstitute/hellbender/engine/TwoPassReadsWalkerUnitTest.java
```diff
@@ -0,0 +1,82 @@
...
+        tool.onStartup();
+        tool.doWork();
+        tool.onShutdown();
+
+        Assert.assertTrue(tool.everythingWorked());
+    }
+
+    @DataProvider(name = "extensions")
+    public Object[][] makeExtensions() {
+        return new Object[][] {{".bam"}, {".sam"}, {".cram"}};
```
lbergelson: Could you switch this to just specifying the whole file name instead of the extension, it makes it easier to search for where files are used if they whole name is in one piece.

lbergelson:
@jamesemery Looks good.  I had a few very minor comments.  

Could you also add a test that does the same thing but uses a -L interval since you had to do some extra stuff to get that working and it tends to be a bit touchy?
Thank you!

src/test/java/org/broadinstitute/hellbender/engine/TwoPassReadsWalkerUnitTest.java
jamesemery: Yes, apparently cram files can't be read without a reference so it was easier to include it despite it not being relevant to my test.

jamesemery:
@lbergelson Alright, I have responded to all of your comments on the branch and added an explicit interval test (I could not check the same test for .sam files because they are unindexed and don't support interval traversals)

src/test/java/org/broadinstitute/hellbender/engine/TwoPassReadsWalkerUnitTest.java
lbergelson: Ah, yes, of course.  Makes sense.

lbergelson:
👍

PR #2131, opened 2016-08-31
Tabix support for IndexFeatureFile, and move to htsjdk 2.6.1

### LINKED ISSUE #2080: IndexFeatureFile support for tabix index apart of VCF
After my contribution to htsjdk to encapsulate the tabix format in the feature codecs (https://github.com/samtools/htsjdk/pull/669), `IndexFactory.createTabixIndex(final File inputFile, final FeatureCodec<FEATURE_TYPE, SOURCE_TYPE> codec, final SAMSequenceDictionary sequenceDictionary)` could be used for index block-compressed feature files of whatever type which implements the `FeatureCodec.getTabixFormat()`.

I can take this ticket, implementing the `getTabixFormat()` method for the codecs in GATK.

magicDGS:
* Update htsjdk to 2.6.1
* Fix #2080


src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
```diff
@@ -104,13 +105,20 @@ private File determineFileName(final Index index) {
...
-            }
-            else {
-                throw new UserException("This tool only supports indexing of block-compressed files when they are in VCF format");
+            try {
+                // TODO: this could benefit from provided sequence dictionary from reference
+                // TODO: this can be an optional parameter for the tool
+                return IndexFactory
+                        .createIndex(featureFile, codec, IndexFactory.IndexType.TABIX, null);
+            } catch(TribbleException.MalformedFeatureFile e) {
+                throw e;
```
droazen: Wrap this in a `UserException.CouldNotReadInputFile`

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
```diff
@@ -125,6 +133,11 @@ else if ( featureFile.getName().endsWith(GVCF_FILE_EXTENSION) ) {
         }
     }
 
+    /**
+     * @deprecated because it is not used anymore
+     */
+    // TODO: should be removed?
+    @Deprecated
```
droazen: It's private, so go ahead and delete it if it's unused.

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
```diff
@@ -104,13 +105,20 @@ private File determineFileName(final Index index) {
...
-            if ( isVCFCodec(codec) ) {
-                return IndexFactory.createTabixIndex(featureFile, codec, TabixFormat.VCF, null);
-            }
-            else {
-                throw new UserException("This tool only supports indexing of block-compressed files when they are in VCF format");
+            try {
+                // TODO: this could benefit from provided sequence dictionary from reference
+                // TODO: this can be an optional parameter for the tool
+                return IndexFactory
+                        .createIndex(featureFile, codec, IndexFactory.IndexType.TABIX, null);
```
droazen: Add integration tests showing that we can now index block-compressed feature files that are not vcfs (eg., BED files, SAMPileup files).

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
magicDGS: This changed the behaviour and that's why I used the double catch clause. Should I unify the catch?

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
magicDGS: Do you have any suggestion for the TODO? I think that an optional parameter with the sequence dictionary will be interesting for fast indexing...

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
magicDGS: Done, but wrap into `UserException.MalformedFile` instead. Changed also in the tests.

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
magicDGS: Added the tests, but the one for bed files is not enabled.

magicDGS:
I addressed the comments, @droazen. But I found that the way to detect the codec for the files is relying on `canDecode`. But even if the a block-compressed file could be read by a codec through the `AbstractFeatureReader`, the `canDecode` for bed files returns false for block-compressed extension.

This should be change at the htsjdk level if GATK is planing to read block-compressed feature inputs. The same for the codecs implemented in the framework (I did it for the `SAMPileupCodec`), but for instance the table codec could not be used in a compressed way because you can't create a tabix index for it...

src/main/java/org/broadinstitute/hellbender/utils/codecs/ProgressReportingDelegatingCodec.java
```diff
@@ -122,4 +123,12 @@ public boolean canDecode(final String path) {
     public FeatureCodec<A, B> getDelegatee() {
         return delegatee;
     }
+
+    @Override
+    public TabixFormat getTabixFormat() {
+        if (delegatee == null) {
+            throw new IllegalArgumentException("this codec cannot be used without a delegatee.");
```
droazen: Should be `IllegalStateException` (since there are no arguments here).

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
```diff
@@ -265,6 +265,61 @@ public void testBedIndex() {
         Assert.assertEquals(index.getSequenceNames(), Arrays.asList("1", "2", "4"));
     }
 
+    // TODO: this is not enabled because canDecode returns false for bed.gz extension
+    // TODO: should this be changed at the htsjdk level to return true also for a block-compressed extension?
```
droazen: Yes, you're right -- assuming that `AbstractFeatureReader.getFeatureReader()` will work correctly in the block-gzipped case with this codec, `canDecode()` should be returning true. Recommend filing a ticket or PR against htsjdk to fix this.

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
```diff
@@ -265,6 +265,61 @@ public void testBedIndex() {
...
+    @Test(enabled = false)
+    public void testBedGZIndex() {
+        final File ORIG_FILE = getTestFile("test_bed_for_index.bed.gz");
```
droazen: Is this test file gzipped or block-gzipped? (they are not the same thing!)

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
```diff
@@ -265,6 +265,61 @@ public void testBedIndex() {
...
+        final File outName = BaseTest.createTempFile("test_bed_for_index.bed.", TabixUtils.STANDARD_INDEX_EXTENSION);
```
droazen: Can just write `createTempFile()` instead of `BaseTest.createTempFile()`, since it's inherited.

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
```diff
@@ -265,6 +265,61 @@ public void testBedIndex() {
...
+
+        final String[] args = {
+                "--feature_file" ,  ORIG_FILE.getAbsolutePath(),
+                "-O" ,  outName.getAbsolutePath()
+        };
+        final Object res = this.runCommandLine(args);
+        Assert.assertEquals(res, outName.getAbsolutePath());
+
+        final Index index = IndexFactory.loadIndex(res.toString());
+        Assert.assertTrue(index instanceof LinearIndex);
```
droazen: This assert doesn't seem right -- it should be a `TabixIndex`, shouldn't it?

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
```diff
@@ -265,6 +265,61 @@ public void testBedIndex() {
...
+        for (final String chr : Arrays.asList("X", "Y", "MT")){
+            Assert.assertFalse(index.containsChromosome(chr), String.valueOf(chr));
+        }
+
+        Assert.assertEquals(index.getSequenceNames(), Arrays.asList("1", "2", "4"));
+    }
+
+    @Test
+    public void testSAMPileupGZIndex() {
+        final File ORIG_FILE = getTestFile("test_sampileup_for_index.pileup.gz");
```
droazen: Same question here -- is this file block-gzipped, or regular gzipped?

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
```diff
@@ -265,6 +265,61 @@ public void testBedIndex() {
...
+            Assert.assertTrue(index.containsChromosome(String.valueOf(chr)), String.valueOf(chr));
+        }
+        for (int chr: new int[]{5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22}) { //note: unusual loop
+            Assert.assertFalse(index.containsChromosome(String.valueOf(chr)), String.valueOf(chr));
+        }
+        for (final String chr : Arrays.asList("X", "Y", "MT")){
+            Assert.assertFalse(index.containsChromosome(chr), String.valueOf(chr));
+        }
+
+        Assert.assertEquals(index.getSequenceNames(), Arrays.asList("1", "2", "3", "4"));
```
droazen: The test methods here have a lot of repeated code between them -- perhaps you could extract a shared private method that takes the expected index class as an argument?

droazen:
Second-pass review complete -- back to @magicDGS

src/main/java/org/broadinstitute/hellbender/utils/codecs/ProgressReportingDelegatingCodec.java
magicDGS: Sure!

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
magicDGS: I bgzip this and the pileup.

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
magicDGS: I bgzip this and the bed file. If it is not block-compressed, the test will fail (like in `testVCFGZIPIndex()` and `testVCFGZIPIndex_inferredName()`. I added a comment to point it out.

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
magicDGS: Sure, I extracted this and the pileup code.

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
magicDGS: Sorry, copy-paste error. Fixed while extracting the common code from both tests.

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
magicDGS: Done (https://github.com/samtools/htsjdk/pull/704)

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
magicDGS: Done. I change it also in the rest of the tests in this file.

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
magicDGS: The case of the testSAMPileupGZIndex is different, because the expected chromosomes are not the same (3 is included)...

magicDGS:
Adressed the comments -- back to you @droazen. I will squash/rebase when you tell me.

droazen:
:+1: looks good, I've merged this (github's new "squash and merge" feature eliminates the need for PR authors to manually squash at the end of review in most cases)

magicDGS:
Thanks @droazen!

vdauwera:
@droazen I know it's too late now but for the future... I would still recommend requiring the author to do a manual rebase to clean up the commit history, otherwise later it's going to be a pain to do things like assemble coherent and succinct version notes. We end up with many lines in the logs that consist of "addressing comments" and "fixing tests", which are completely uninformative.

vdauwera:
Alright, @droazen just showed me what actually got committed and it's all cleaned up. Yay! So now I approve this feature. As long as people do take advantage of the opportunity to clean up the squashed commit message!

PR #2136, opened 2016-08-31
Created first level class LocalAssemblyContig

SHuang-Broad:
1st step towards bringing in fermi-lite as alternative option for local assembly.
Updated other classes accordingly.
Some teeny clean ups elsewhere.

@cwhelan mind reviewing? Mostly trivial changes, should be very quick.
Thanks!

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignAssembledContigsSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignAssembledContigsSpark.java
@@ -61,8 +61,8 @@ public final class AlignAssembledContigsSpark extends GATKSparkTool {
                 try (final ContigAligner contigAligner = new ContigAligner(referenceFileName)) {
                     final List<AlignmentRegion> results = new ArrayList<>(NUM_ASSEMBLIES_PER_PARTITION * EXPECTED_CONTIGS_PER_ASSEMBLY);
                     iter.forEachRemaining(cc -> {
-                        String breakpointId = cc._1;
-                        final List<AlignmentRegion> contigAlignments = contigAligner.alignContigs(breakpointId, cc._2);
+                        String assemblyid = cc._1;
+                        final List<AlignmentRegion> contigAlignments = contigAligner.alignContigs(assemblyid, cc._2);
                         contigAlignments.forEach(results::add);
                     });
                     return results;
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigsCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigsCollection.java
@@ -12,58 +12,22 @@ import scala.Tuple2;
 
 import java.io.Serializable;
 import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
- * Represents a collection of assembled contigs (not including the variants) produced by "sga assemble".
+ * Represents a collection of assembled contigs (not including the variants) produced by local assembler,
+ * e.g. sga or fermi-lite.
  */
 @VisibleForTesting
 final class ContigsCollection implements Serializable {
     private static final long serialVersionUID = 1L;
     private static final Logger logger = LogManager.getLogger(ContigsCollection.class);
 
-    @VisibleForTesting
-    static final class ContigSequence implements Serializable{
-        private static final long serialVersionUID = 1L;
+    private final List<LocalAssemblyContig> assemblyContents;
 
-        private final String sequence;
-        public ContigSequence(final String sequence){ this.sequence = sequence; }
-
-        @Override
-        public String toString(){
-            return sequence;
-        }
-    }
-
-    @VisibleForTesting
-    static final class ContigID implements Serializable{
-        private static final long serialVersionUID = 1L;
-
-        private final String id;
-        public ContigID(final String idString) { this.id = idString; }
-
-        @Override
-        public String toString(){
-            return id;
-        }
-
-        @Override
-        public boolean equals(final Object o) {
-            if (this == o) return true;
-            if (o == null || getClass() != o.getClass()) return false;
-            final ContigID contigID = (ContigID) o;
-            return Objects.equals(id, contigID.id);
-        }
-
-        @Override
-        public int hashCode() {
-            return Objects.hash(id);
-        }
-    }
-
-    private final List<Tuple2<ContigID, ContigSequence>> contents;
-
-    public List<Tuple2<ContigID, ContigSequence>> getContents(){
-        return contents;
+    public List<LocalAssemblyContig> getContents(){
+        return assemblyContents;
     }
 
     @Override
@@ -76,7 +40,7 @@ final class ContigsCollection implements Serializable {
      * Spark tools to parse without worrying about partitioning. If the ContigCollection
      * is empty this returns null.
      */
-    public String toPackedFasta(){
+    String toPackedFasta(){
         return StringUtils.join(toListOfStrings(),"|");
     }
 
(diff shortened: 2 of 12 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignAssembledContigsSpark.java
```diff
@@ -61,8 +61,8 @@ protected void runTool(final JavaSparkContext ctx) {
                 try (final ContigAligner contigAligner = new ContigAligner(referenceFileName)) {
                     final List<AlignmentRegion> results = new ArrayList<>(NUM_ASSEMBLIES_PER_PARTITION * EXPECTED_CONTIGS_PER_ASSEMBLY);
                     iter.forEachRemaining(cc -> {
-                        String breakpointId = cc._1;
-                        final List<AlignmentRegion> contigAlignments = contigAligner.alignContigs(breakpointId, cc._2);
+                        String assemblyid = cc._1;
```
cwhelan: Nice, thanks for renaming these.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigsCollection.java
```diff
@@ -85,50 +49,52 @@ public String toPackedFasta(){
      * @param packedFastaLine
      */
     @VisibleForTesting
-    protected static ContigsCollection fromPackedFasta(final String packedFastaLine) {
+    protected static ContigsCollection fromPackedFasta(final long assemblyID, final String packedFastaLine) {
```
cwhelan: I'm a little bit confused about `assemblyId`. Why do we need it to create a `ContigsCollection`? And then why store it at the level of the `LocalAssemblyContig`? It doesn't look like it's being accessed there. If we want to track it it seems like it would go better at the ContigsCollection level.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigsCollection.java
```diff
@@ -85,50 +49,52 @@ public String toPackedFasta(){
...
-        }
-        return res;
+        return assemblyContents.stream().flatMap(contig -> Stream.of(contig.contigID, contig.seq)).collect(Collectors.toList());
     }
 
-    public ContigsCollection(final List<String> fileContents){
+    /**
+     * Ctor to be used for SGA output
+     */
+    ContigsCollection(final long assemblyID, final List<String> sgaFASTAFileContents){
```
cwhelan: The SGA output is just a standard FASTA file, right? So can we keep this method name generic?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigsCollection.java
```diff
@@ -85,50 +49,52 @@ public String toPackedFasta(){
...
+            for(int i=0; i<sgaFASTAFileContents.size(); i+=2){
+                assemblyContents.add( new LocalAssemblyContig(assemblyID, sgaFASTAFileContents.get(i), sgaFASTAFileContents.get(i+1)) );
             }
         }
     }
 
     /**
      * Loads an RDD of {@link ContigsCollection} objects keyed by assembly ID from disk. The input file
-     * should be the output of as RunSGAViaProcessBuilderOnSpark.
+     * should be the output of as {@link RunSGAViaProcessBuilderOnSpark}.
```
cwhelan: In the code even before you changed it: there's a word extra in this comment; "the output of as".

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/LocalAssemblyContig.java
```diff
@@ -0,0 +1,95 @@
...
+     * May be {@code null} if assembler doesn't provide this feature
+     */
+    final byte[] quals;
+
+    /**
+     * Read count for the contigs.
+     * TODO: fermi-lite provides support for this, not SGA
+     */
+    final int readCountSupport;
+    static final int READ_COUNT_SUPPORT_DEFAULT = -1;
```
cwhelan: Why not make this a capital-I Integer so that it can be null in the case where we don't have read support?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignAssembledContigsSpark.java
SHuang-Broad: Sorry I didn't catch it while reviewing the previous PR.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigsCollection.java
SHuang-Broad: Actually the individual `LocalAssemblyContig`'s will be accessed in the genotyping step, because not all contigs will be coming from the same assembly.
Does this make things a bit clearer?
In fact I've been thinking of removing `ContigsCollection` all together, but realized that would be too big a PR so just laid back and being lazy...

SHuang-Broad: About removing `ContigsCollection`, another related concern is that when we have fermi-lite assembly as the assembly tool, a similar collection concept might be necessary. But currently the jni binding was simply returning a list of MAG format vertices.
When we request GFA format (a graph) from fermi-lite, the returned struct will be a whole atomic object instead of a container of something.
So it is not a concern there.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigsCollection.java
SHuang-Broad: Yes, admittedly the documentation is bad. I was thinking too much about the different outputs between SGA and fermi-lite.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/LocalAssemblyContig.java
SHuang-Broad: Yes, was simply falling in the trap of primitive-bias and thought about this null-equivalent for it.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/LocalAssemblyContig.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private List<AlignmentRegion> alignmentRecords;
```
cwhelan: You've added a place to put alignmentRecords but we're never actually assigning them or accessing them from there.

In general I'd prefer not to add code that's not being used. It's better to wait until we have an actual need for something to add it to the code base. Some refactorings to clean things up before another set of changes can be useful, but I think the existing code should be modified to take advantage of them when they are done, rather than leaving placeholders. So I'd prefer to either drop these unused fields, or refactor the existing code to use them.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/LocalAssemblyContig.java
```diff
@@ -0,0 +1,95 @@
...
+                        final int readCountSupport, final byte[] quals){
+        this.assemblyID = assemblyID;
+        this.contigID   = contigID;
+        this.seq        = seq;
+        this.alignmentRecords = alignmentRegions;
+        this.quals = (quals == null) ? null : quals;
+        this.readCountSupport = readCountSupport;
+    }
+
+    LocalAssemblyContig(final long assemblyID, final String contigID, final String seq,
```
cwhelan: As mentioned above, this constructor is never called.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -49,12 +47,6 @@
               optional  = false)
     public String pathToAllInterleavedFASTQFiles = null;
 
-    @Argument(doc       = "A substring in the FASTQ file names that needs to be stripped out for retrieving the breakpoint ID",
```
cwhelan: I like that you removed this and made it a constant.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/RunSGAViaProcessBuilderOnSpark.java
```diff
@@ -75,16 +67,45 @@
...
-    @VisibleForTesting static final double ASSEMBLE_STEP_MAX_GAP_DIVERGENCE = 0.0;
-    @VisibleForTesting static final int ASSEMBLE_STEP_MIN_BRANCH_TAIL_LENGTH = 50;
-    @VisibleForTesting static final int CUT_OFF_TERMINAL_BRANCHES_IN_N_ROUNDS = 0;
-    @VisibleForTesting static final boolean TURN_ON_TRANSITIVE_REDUCTION = false;
-
+    //----------------------------------------------------------------------------------------------------------------//
+    // Result-affecting parameters for use in various SGA modules with default values.
+    // Long names are the same as those in SGA documentation, except using camel case in place of dashes.
+    //----------------------------------------------------------------------------------------------------------------//
+    @Argument(doc       = "Minimum overlap parameter value for steps: fm-merge, overlap, and assemble.",
```
cwhelan: I think you also made these changes in the branch sh_sga_para_and_error_handle, no? I'd keep the changes in one branch and wait til it gets merged, otherwise you'll have trouble at rebase time.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ContigsCollection.java
cwhelan: I guess my concern about not adding things until we need them also applies here, then. In this case I guess it might be alright -- at least we are setting the value, rather than leaving it untouched.

SHuang-Broad:
@cwhelan , I'm closing this because most of the changes are incorporated into the genotyping PR and all comments have been addressed there.
Thank you!

Issue #2148, opened 2016-09-06
getDefaultReadFilters should be overridden in SplitNCigarReads not TwoPassReadWalker


lbergelson:
`getDefaultReadFilters` is overridden in `TwoPassReadWalker` but that override should be moved to `SplitNCigarReads` since it's really specific to SplitNCigarReads.

cmnbroad:
Also at the same time the override of makeReadFilter in SplitNCigarReads should be removed (unless it doesn't want the default treatment), since its the base class implementation of makeReadFilter that does the enabling/disabling/merging of user specified read filters.

cmnbroad:
@lbergelson I took this since I was making a PR that had changes here anyway.

Issue #2142, opened 2016-09-02
Replace CollectInsertSizeMetrics custom readFilter arguments with standard read filter args


cmnbroad:
InsertSizeMetricsArgumentCollection has several custom command line arguments that control read filters. Now that there is a standard mechanism for read filter arguments, these custom args should be removed and replaced with tool default read filters that are controlled by standard read filter arguments.

cmnbroad:
Fixed via https://github.com/broadinstitute/gatk/pull/2149.

PR #2149, opened 2016-09-07
Remove and replace custom tool filter args with standard read filters.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/2148, https://github.com/broadinstitute/gatk/issues/2142, and a bug where we were honoring command line order of read filters but not tool default order. Also contains a few TODOs with questions about the suitability of names and/or implementations of some existing read filters.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilter.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilter.java
@@ -14,7 +14,12 @@ import java.util.function.Predicate;
  * based on the implementing class's implementation of test().
  *
  * To be accessible from the command line, subclasses must have a zero-arg constructor and contain
- * an ArgumentCollection.
+ * (optional) ArgumentCollection or Argument annotated fields for filter-specific arguments. Arguments
+ * that are optional=true should have an initial value since the command line parser will not require
+ * the user to provide one; arguments that are optional=false should not have an initial value (and
+ * used boxed types with null default/initial values) since otherwise they will be seen by the command
+ * line parser as having been provided, and the user will not be notified that the value must be provided
+ * on the command line).
  */
 public abstract class ReadFilter implements Predicate<GATKRead>, Serializable {
 
--- a/src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
@@ -20,17 +20,27 @@ public final class ReadFilterLibrary {
         @Override public boolean test(final GATKRead read){return true;}}
 
     //Note: do not call getCigar to avoid creation of new Cigar objects
-    public static class CigarIsSupportedReadFilter extends ReadFilter {
+    public static class CigarContainsNoNOperator extends ReadFilter {
         private static final long serialVersionUID = 1L;
         @Override public boolean test(final GATKRead read){
             return ! CigarUtils.containsNOperator(read.getCigarElements());}}
+
+    public static class FirstOfPairReadFilter extends ReadFilter {
+        private static final long serialVersionUID = 1L;
+        @Override public boolean test (final GATKRead read) {
+            return read.isFirstOfPair();}}
 
     public static class GoodCigarReadFilter extends ReadFilter {
         private static final long serialVersionUID = 1L;
         @Override public boolean test (final GATKRead read) {
             return CigarUtils.isGood(read.getCigar());}}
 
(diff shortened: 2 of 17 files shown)


src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -20,17 +20,28 @@ private ReadFilterLibrary(){ /*no instance*/ }
         @Override public boolean test(final GATKRead read){return true;}}
 
     //Note: do not call getCigar to avoid creation of new Cigar objects
+    //TODO: is "CigarIsSupported" well understood to mean "no Ns", or should this be renamed ?
```
lbergelson: I think it may have originally had extra conditions as well.  Lets rename.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilter.java
```diff
@@ -14,7 +14,11 @@
  * based on the implementing class's implementation of test().
  *
  * To be accessible from the command line, subclasses must have a zero-arg constructor and contain
- * an ArgumentCollection.
+ * an (optional) ArgumentCollection for filter-specific arguments. Arguments that are optional=true
```
lbergelson: It can include raw `@Argument` as well right? Not just `@ArgumentCollection`.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -89,14 +100,36 @@ private ReadFilterLibrary(){ /*no instance*/ }
...
+    public static class ProperlyPairedReadFilter extends ReadFilter {
+        private static final long serialVersionUID = 1L;
+        @Override public boolean test(final GATKRead read) {
+            return read.isProperlyPaired();}}
+
     public static class PrimaryAlignmentReadFilter extends ReadFilter {
         private static final long serialVersionUID = 1L;
-        @Override public boolean test(final GATKRead read){
+        @Override public boolean test(final GATKRead read) {
+            //TODO: should this also be rejecting isSupplementaryAlignment() ?
```
lbergelson: I think this should probably reject supplementary but not duplicates.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -42,23 +53,23 @@ private ReadFilterLibrary(){ /*no instance*/ }
...
     public static class MappingQualityNotZeroReadFilter extends ReadFilter {
         private static final long serialVersionUID = 1L;
-        @Override public boolean test(final GATKRead read){
+        @Override public boolean test(final GATKRead read) {
             return read.getMappingQuality() != 0;}}
 
     /**
      * Reads that either have a mate that maps to the same contig, or don't have a mapped mate.
      */
-    public static class MateOnSameContigOrNoMappedMateReadFilter extends ReadFilter {
```
lbergelson: I think the original name is clearer.  It encapsulates the fact that it accepts reads with either no mate, or reads with an unmapped mate.  It's unclear what the new name does with unpaired reads.

lbergelson:
@cmnbroad Looks good to me modulo a few comments.

lbergelson:
Thanks for a doing that TwoPass update in here.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
cmnbroad: Renaming to CigarContainsNoNOperator.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilter.java
cmnbroad: Yes. Updated.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
cmnbroad: Reverting.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
cmnbroad: I created a ticket to resolve whether we should change this or delete it (since we now have NotSecondary for InsertSizeMetrics) https://github.com/broadinstitute/gatk/issues/2165.

PR #2154, opened 2016-09-08
Pileup-related changes (LIBS/LocusWalker/ReadPileup)

magicDGS:
This PR tries to solve several issues:
* Refactoring constructors for LIBS (#1879)
* Adding maximum depth per sample argument to `LocusWalker` to avoid memory overload
* Fixing `Pileup`/`CheckPileup` tools for command line read filters
* Method for fix overlaps in `ReadPileup` in the same way as samtools (#2034)

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
@@ -2,16 +2,19 @@ package org.broadinstitute.hellbender.engine;
 
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMReadGroupRecord;
+import org.broadinstitute.hellbender.cmdline.Argument;
 import org.broadinstitute.hellbender.cmdline.GATKPlugin.GATKCommandLinePluginDescriptor;
 import org.broadinstitute.hellbender.cmdline.GATKPlugin.GATKReadFilterPluginDescriptor;
 import org.broadinstitute.hellbender.engine.filters.CountingReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.filters.WellformedReadFilter;
+import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.iterators.IntervalOverlappingIterator;
 import org.broadinstitute.hellbender.utils.iterators.ReadFilteringIterator;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.downsampling.DownsamplingMethod;
+import org.broadinstitute.hellbender.utils.locusiterator.LIBSDownsamplingInfo;
 import org.broadinstitute.hellbender.utils.locusiterator.LocusIteratorByState;
 
 import java.util.*;
@@ -29,6 +32,9 @@ import java.util.stream.StreamSupport;
  * @author Daniel Gomez-Sanchez (magicDGS)
  */
 public abstract class LocusWalker extends GATKTool {
+
+    @Argument(fullName = "maxDepthPerSample", shortName = "maxDepthPerSample", doc = "Maximum number of reads to retain per sample per locus. Reads above this threshold will be downsampled. Set to 0 to disable.", optional = true)
+    protected int maxDepthPerSample = defaultMaxDepthPerSample();
 
     /**
      * Should the LIBS keep unique reads? Tools that do should override to return {@code true}.
@@ -61,6 +67,14 @@ public abstract class LocusWalker extends GATKTool {
      */
     public boolean includeNs() {
         return false;
+    }
+
+    /**
+     * Returns default value for the {@link #maxDepthPerSample} parameter, if none is provided on the command line.
+     * Default implementation returns 0 (no downsampling by default).
+     */
+    protected int defaultMaxDepthPerSample() {
+        return 0;
     }
 
     /**
@@ -112,15 +126,14 @@ public abstract class LocusWalker extends GATKTool {
         return defaultFilters;
     }
 
-    /**
-     * Get the information about how to downsample the reads. By default {@link DownsamplingMethod#NONE} is returned.
-     * Subclasses should override it to provide their own downsampling methods.
-     *
-     * @return the downsampling method for the reads
-     */
-    public DownsamplingMethod getDownsamplingMethod() {
-        // TODO: change when downsampling command line options are implemented
-        return DownsamplingMethod.NONE;
+    /** Returns the downsampling info using {@link #maxDepthPerSample} as target coverage. */
+    protected final LIBSDownsamplingInfo getDownsamplingInfo() {
+        if (maxDepthPerSample < 0) {
+            throw new UserException.BadArgumentValue("maxDepthPerSample",
+                    String.valueOf(maxDepthPerSample),
+                    "should be a positive number");
+        }
+        return (maxDepthPerSample == 0) ? LocusIteratorByState.NO_DOWNSAMPLING : new LIBSDownsamplingInfo(true, maxDepthPerSample);
     }
 
     /**
@@ -139,7 +152,7 @@ public abstract class LocusWalker extends GATKTool {
      * Subclasses can override to provide their own behavior but default implementation should be suitable for most uses.
      *
      * The default implementation iterates over all positions in the reference covered by reads for all samples in the read groups, using
-     * the downsampling method provided by {@link #getDownsamplingMethod()}
+     * the downsampling method provided by {@link #getDownsamplingInfo()}
      * and including deletions only if {@link #includeDeletions()} returns {@code true}.
      */
     @Override
@@ -149,9 +162,9 @@ public abstract class LocusWalker extends GATKTool {
         final Set<String> samples = header.getReadGroups().stream()
                                           .map(SAMReadGroupRecord::getSample)
                                           .collect(Collectors.toSet());
-        CountingReadFilter countedFilter = makeReadFilter();
+        final CountingReadFilter countedFilter = makeReadFilter();
         // get the LIBS
-        LocusIteratorByState libs = new LocusIteratorByState(new ReadFilteringIterator(reads.iterator(), countedFilter), getDownsamplingMethod(), includeDeletions(), includeNs(), keepUniqueReadListInLibs(), samples, header);
+        final LocusIteratorByState libs = new LocusIteratorByState(new ReadFilteringIterator(reads.iterator(), countedFilter), getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
         // prepare the iterator
         Spliterator<AlignmentContext> iterator = (hasIntervals()) ? new IntervalOverlappingIterator<>(libs, intervalsForTraversal, header.getSequenceDictionary()).spliterator() : libs.spliterator();
         // iterate over each alignment, and apply the function
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
@@ -5,6 +5,8 @@ import org.broadinstitute.hellbender.cmdline.CommandLineProgramProperties;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 import org.broadinstitute.hellbender.cmdline.programgroups.QCProgramGroup;
 import org.broadinstitute.hellbender.engine.*;
+import org.broadinstitute.hellbender.engine.filters.ReadFilter;
+import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.IntervalUtils;
 import org.broadinstitute.hellbender.utils.codecs.sampileup.SAMPileupFeature;
@@ -64,6 +66,10 @@ public final class CheckPileup extends LocusWalker {
     @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME, shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "Output file (if not provided, defaults to STDOUT).", optional = true)
     public File outFile = null;
 
+    // This is to check for previous versions of samtools without having overlap detection
+    @Argument(fullName = "ignore_overlaps", shortName = "ignore_overlaps", doc = "Disable read-pair overlap detection.", optional = true)
+    public boolean ignoreOverlaps = false;
+
     /**
      * By default the program will quit if it encounters an error (such as missing truth data for a given position).
      * Use this flag to override the default behavior; the program will then simply print an error message and move on
@@ -71,6 +77,16 @@ public final class CheckPileup extends LocusWalker {
      */
     @Argument(fullName="continue_after_error", doc="Continue after encountering an error", optional=true)
     public boolean continueAfterAnError = false;
+
+    // These are the default read filters from samtools
+    @Override
+    public List<ReadFilter> getDefaultReadFilters() {
+        final List<ReadFilter> defaultFilters = super.getDefaultReadFilters();
+        defaultFilters.add(ReadFilterLibrary.NOT_DUPLICATE);
+        defaultFilters.add(ReadFilterLibrary.PASSES_VENDOR_QUALITY_CHECK);
+        defaultFilters.add(ReadFilterLibrary.PRIMARY_ALIGNMENT);
+        return defaultFilters;
+    }
 
     @Override
     public boolean requiresReference() {
@@ -94,6 +110,10 @@ public final class CheckPileup extends LocusWalker {
     public void apply(final AlignmentContext context, final ReferenceContext ref, final FeatureContext featureContext) {
         final ReadPileup pileup = context.getBasePileup();
         final SAMPileupFeature truePileup = getTruePileup(featureContext);
+
+        if (!ignoreOverlaps) {
+            pileup.fixOverlaps();
+        }
 
         if ( truePileup == null ) {
             out.printf("No truth pileup data available at %s%n", pileup.getPileupString((char) ref.getBase()));
--- a/src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
@@ -33,7 +33,7 @@ import java.util.*;
  * That is, the pileup is ordered in a way consistent with the SAM coordinate ordering
  * -- Only aligned reads with at least one on-genomic cigar operator are passed on in the pileups.  That is,
  * unmapped reads or reads that are all insertions (10I) or soft clipped (10S) are not passed on.
- * -- LIBS can perform per-sample downsampling of a variety of kinds.
+ * -- LIBS can perform per-sample downsampling to a coverage target if requested.
  * -- Because of downsampling there's no guarantee that:
  *   -- A read that could be aligned to a position will actually occur in the pileup (downsampled away)
  *   -- A read that appears in a previous pileup that could align to a future position will actually occur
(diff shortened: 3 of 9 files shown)


magicDGS:
@droazen , I guess that this will be for you because you were in the previous PR ( #2041). I think that this is cleaner and does not interfere in the basic behavior for GATK tools; but it will be very useful for my work.

Thanks a lot!

droazen:
Re-assigning to @fleharty

magicDGS:
Hi @fleharty! Just a friendly ping here, because I need this in my work.

fleharty:
@magicDGS reviewing

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -64,6 +66,10 @@
     @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME, shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "Output file (if not provided, defaults to STDOUT).", optional = true)
     public File outFile = null;
 
+    // this is to check for previous versions of samtools without having overlap detection
```
fleharty: Capitalize This

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -72,6 +78,16 @@
     @Argument(fullName="continue_after_error", doc="Continue after encountering an error", optional=true)
     public boolean continueAfterAnError = false;
 
+    // this are the default read filters from samtools
```
fleharty: this -> These

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
```diff
@@ -33,7 +33,7 @@
  * That is, the pileup is ordered in a way consistent with the SAM coordinate ordering
  * -- Only aligned reads with at least one on-genomic cigar operator are passed on in the pileups.  That is,
  * unmapped reads or reads that are all insertions (10I) or soft clipped (10S) are not passed on.
- * -- LIBS can perform per-sample downsampling of a variety of kinds.
+ * -- LIBS can perform per-sample downsampling of a variety of kinds. TODO: it is true?
```
fleharty: Should this still be here?  I think this TODO should be removed.

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
```diff
@@ -95,115 +95,118 @@
...
 
     /**
      * Create a new LocusIteratorByState
      *
-     * @param samIterator the iterator of reads to process into pileups.  Reads must be ordered
-     *                    according to standard coordinate-sorted BAM conventions
-     * @param downsamplingInfo meta-information about how to downsampling the reads
+     * @param samIterator                    the iterator of reads to process into pileups.  Reads must be ordered
+     *                                       according to standard coordinate-sorted BAM conventions
+     * @param downsamplingInfo               meta-information about how to downsampling the reads
```
fleharty: downsampling -> downsample

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
```diff
@@ -95,115 +95,118 @@
...
+     * @param keepUniqueReadListInLIBS       if true, we will keep the unique reads from off the samIterator and make them
```
fleharty: from "off" ?

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -6,6 +6,8 @@
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.BaseUtils;
 import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.fragments.FragmentCollection;
+import org.broadinstitute.hellbender.utils.fragments.FragmentUtils;
```
fleharty: FragmentUtils is not needed, please remove

src/test/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByStateBaseTest.java
```diff
@@ -64,11 +64,9 @@ protected LocusIteratorByState makeLIBS(final List<GATKRead> reads,
         return new LocusIteratorByState(
                 new FakeCloseableIterator<>(reads.iterator()),
                 downsamplingMethod,
-                true,
-                includeNs,
-                keepUniqueReadList,
-                sampleListForSAMWithoutReadGroups(),
-                header);
+                keepUniqueReadList, sampleListForSAMWithoutReadGroups(), header, true,
```
fleharty: Use of new lines is inconsistent, please fix.

src/test/java/org/broadinstitute/hellbender/utils/pileup/ReadPileupUnitTest.java
```diff
@@ -490,4 +491,106 @@ public void testSimplePileupWithOffset(){
...
+        final GATKRead read1 = ArtificialReadUtils.createArtificialRead("5M");
+        final GATKRead read2 = ArtificialReadUtils.createArtificialRead("5M");
+        read1.setBases(new byte[] {'A', 'A', 'A', 'A', 'T'});
+        read1.setBaseQualities(new byte[] {highQual, highQual, highQual, lowQual,
+                SAMUtils.MAX_PHRED_SCORE});
+        read2.setBases(new byte[] {'A', 'T', 'T', 'T', 'T'});
+        read2.setBaseQualities(new byte[] {lowQual, highQual, lowQual, highQual,
+                SAMUtils.MAX_PHRED_SCORE});
+        return new Object[][] {
+                {PileupElement.createPileupForReadAndOffset(read1, 0),
```
fleharty: For each of these tests, could you provide a 1 sentence description of what that test is establishing?

src/test/java/org/broadinstitute/hellbender/utils/pileup/ReadPileupUnitTest.java
```diff
@@ -490,4 +491,106 @@ public void testSimplePileupWithOffset(){
...
+            ReadPileup.fixPairOverlappingQualities(element1, element2);
+            Assert.assertEquals(element1.getQual(), Byte.MAX_VALUE);
+            Assert.assertEquals(element2.getQual(), 0);
+            element1.getRead().setBaseQualities(iArray);
+            logger.debug("Test: fixing ({}) and ({})", element3, element1);
+            ReadPileup.fixPairOverlappingQualities(element3, element1);
+            Assert.assertEquals(element3.getQual(), Byte.MAX_VALUE);
+            Assert.assertEquals(element1.getQual(), 0);
+        }
+        // finally, check what happen if both are the maximum value
```
fleharty: happen -> happens
finally -> Finally

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -237,6 +239,69 @@ public Locatable getLocation() {
...
+                // cap to maximum byte value
+                if (firstQuals[firstElement.getOffset()] < 0) {
+                    firstQuals[firstElement.getOffset()] = Byte.MAX_VALUE;
+                }
+                secondQuals[secondElement.getOffset()] = 0;
+            } else {
+                // if not, we lost confidence in the one with higher quality
+                if (firstElement.getQual() >= secondElement.getQual()) {
+                    firstQuals[firstElement.getOffset()] =
+                            (byte) (0.8 * firstQuals[firstElement.getOffset()]);
```
fleharty: I'm pretty uncomfortable with this 0.8, where does it come from?
I realize that it is in https://github.com/samtools/htslib/blob/master/sam.c
but I also don't see justification there either.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -237,6 +239,69 @@ public Locatable getLocation() {
...
+                    secondQuals[secondElement.getOffset()] = 0;
+                } else {
+                    secondQuals[secondElement.getOffset()] =
+                            (byte) (0.8 * secondQuals[secondElement.getOffset()]);
```
fleharty: Here too.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -237,6 +239,69 @@ public Locatable getLocation() {
...
+     * samtools does {@see tweak_overlap_quality function in
+     * <a href="https://github.com/samtools/htslib/blob/master/sam.c">samtools</a>}.
+     *
+     * Setting the quality of one of the bases to 0 effectively removes the redundant base for
+     * calling. In addition, if the bases overlap we have increased confidence if they agree (or
+     * reduced if they don't). Thus, the algorithm proceeds as following:
+     *
+     * 1. If the bases are the same, the quality of the first element is the sum of both qualities
+     *    and the quality of the second is reduced to 0.
+     * 2. If the bases are different, the base with the highest quality is reduced with a factor of
```
fleharty: Other than samtools, is there a justification for 0.8?

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -237,6 +239,69 @@ public Locatable getLocation() {
...
+            final byte[] firstQuals = firstElement.getRead().getBaseQualities();
+            final byte[] secondQuals = secondElement.getRead().getBaseQualities();
+            if (firstElement.getBase() == secondElement.getBase()) {
+                // if both have the same base, extra confidence in the firts of them
+                firstQuals[firstElement.getOffset()] =
+                        (byte) (firstQuals[firstElement.getOffset()] + secondQuals[secondElement
+                                .getOffset()]);
+                // cap to maximum byte value
+                if (firstQuals[firstElement.getOffset()] < 0) {
+                    firstQuals[firstElement.getOffset()] = Byte.MAX_VALUE;
```
fleharty: is Byte.MAX_VALUE really what we want here?  I don't know about GATK4, but in GATK3 there is a util class that has a Maximum sensible base quality.

src/test/java/org/broadinstitute/hellbender/utils/pileup/ReadPileupUnitTest.java
```diff
@@ -490,4 +491,106 @@ public void testSimplePileupWithOffset(){
...
+    public void testFixOverlappingQualitiesDeletedElements() throws Exception {
+        final GATKRead read = ArtificialReadUtils.createArtificialRead("1M1D1M");
+        read.setBases(new byte[] {'A', 'A'});
+        read.setBaseQualities(new byte[] {60, 60});
+        final GATKRead read2 = ArtificialReadUtils.createArtificialRead("1M1D1M");
+        read.setBases(new byte[] {'A', 'A'});
+        read2.setBaseQualities(new byte[] {40, 40});
+        final GATKRead copy = read.deepCopy();
+        final GATKRead copy2 = read2.deepCopy();
+        final PileupElement deleted = new PileupElement(copy, 1, copy.getCigarElement(1), 1, 0);
```
fleharty: I'm confused.  copy, copy2, read, and read2 appear to be independent of what is happening in deleted and normal.
Could you clarify what's happening here in a comment?

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
magicDGS: I added this TODO here while doing this changes because I found that the per-sample downsampling is using `LevelingDownsampler`, so I don't know if it is true that it performs dowsampling "of a variety of kinds", even if the `DownsamplingMethod` passed to the constructor is different.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
magicDGS: There is no justification, just the samtools behaviour that I'm trying to mimic here.

My work is related in avoid the step of store the mpileup from samtools, that's why I need this as it is. Actually there are other method in `FragmentUtils` to correct for overlapping read pairs, and probably this will be more suitable for some API users. I can add in the javadoc a link to that method to note that there are other strategies to perform this.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
magicDGS: I was thinking to use the maximum quality in the util class (I think that it is 93), but I decided to use `Byte.MAX_VALUE` to allow the API user to distinguish which values were capped. I don't mind if the value is 93 or `Byte.MAX_VALUE` because anyway we can't mimic the cap value from samtools (200) because byte in Java is signed. Should I change it?

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
magicDGS: I also don't like it, but because I want to check how consistent it is with the samtools result, I definitely require that this is in sync with the sam.c code.

src/test/java/org/broadinstitute/hellbender/utils/pileup/ReadPileupUnitTest.java
magicDGS: I will add a comment, but just for clarify it here too: because the method `ReadPileup.fixPairOverlappingQualities()` modify the read inside the `PileupElement`, we need a way to check that anything within the read was changed. Thus, we require a copy to maintain a backup of the original read and check if it was change or not. Is this clear like that, because I guess that is still a bit confusing?

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
magicDGS: Done

src/test/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByStateBaseTest.java
magicDGS: Sorry, it was because of the IDE refactoring. Fixed.

src/test/java/org/broadinstitute/hellbender/utils/pileup/ReadPileupUnitTest.java
magicDGS: Done

src/test/java/org/broadinstitute/hellbender/utils/pileup/ReadPileupUnitTest.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
magicDGS: A suggestion for make this better will be to add a `double` parameter to downgrade the confidence of a mismatch. Like that I can use it with 0.8 to mimic samtools behaviour (and change it if the samtools method is updated), and other API users could use more reasonable values (e.g. PCR error rate correction and/or just set to 0 to do not take it into account).

magicDGS:
Thanks for the review, @fleharty! I addressed the comments (Travis not passing because of the CLOUD tests).

Regarding the downgrade of quality, I think that it should be 0.8 to keep it as in samtools because it is already noted in the javadoc what we are doing. If someone wants to use a different approach should use other method (like `FragmentUtils.adjustQualsOfOverlappingPairedFragments`) and/or their own implementation. Another way to keep it consistent with samtools but give to the client some freedom, a `double` could be added (checking will be required for the range).

Back to you @fleharty. And thanks again for have a look to this!

fleharty:
@magicDGS I'm at a conference right now, I'm going to respond ASAP, hopefully today.

magicDGS:
Thanks @fleharty!

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
magicDGS: I changed it for `QualityUtils.MAX_SAM_QUAL_SCORE`.

src/test/java/org/broadinstitute/hellbender/utils/pileup/ReadPileupUnitTest.java
```diff
@@ -575,17 +580,26 @@ public void testFixPairOverlappingQualitiesCap() {
...
+        // the PileupElement is a reference to the Object passed to the constructor. This is required
+        // because fixing the overlapping qualities change the read stored into PileupElement and we have to check that
+        // it does not change anything from the read.
         final GATKRead copy = read.deepCopy();
         final GATKRead copy2 = read2.deepCopy();
         final PileupElement deleted = new PileupElement(copy, 1, copy.getCigarElement(1), 1, 0);
         final PileupElement normal = new PileupElement(copy2, 0, copy.getCigarElement(0), 0, 0);
         ReadPileup.fixPairOverlappingQualities(deleted, normal);
+        // Copy reads are a reference to the read into the PileupElement, and thus should not be modified.
+        // Thus, we use them to test if it is not modified. This could be check uncommenting the next two lines:
```
magicDGS: Done

magicDGS:
I addressed the latest comments (using `QualityUtils` static value for capping) and added one more test case. Thanks for the review and back to you @fleharty!

magicDGS:
Friendly ping @fleharty! I will really appreaciatte if this can go into before #2185 to avoid the maintenance of the deprecated `PerReadAlleleLikelihoodMap` just for me.

Thanks a lot!

fleharty:
@magicDGS Thanks for the changes, looks great 👍

magicDGS:
Thanks for reviewing, @fleharty. Could this being merge, @droazen?

davidbenjamin:
Restarting the build, which was failing. . .

magicDGS:
The build is failing for CLOUD=mandatory, because it always does for people external to Broad. Is that an issue, @davidbenjamin?

davidbenjamin:
@magicDGS Ah, got it, @droazen's opinion is the only one that matters.

droazen:
Will have a look at this this afternoon

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -30,6 +32,9 @@
  */
 public abstract class LocusWalker extends GATKTool {
 
+    @Argument(fullName = "maxDepthPerSample", shortName = "maxDepthPerSample", doc = "Maximum number of reads to retain per sample. Reads above this threshold will be downsampled. Set to 0 to disable.", optional = true)
```
droazen: "per sample" -> "per sample per locus"

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -112,15 +125,9 @@ public CountingReadFilter makeReadFilter(){
...
-     * Subclasses should override it to provide their own downsampling methods.
-     *
-     * @return the downsampling method for the reads
-     */
-    public DownsamplingMethod getDownsamplingMethod() {
-        // TODO: change when downsampling command line options are implemented
-        return DownsamplingMethod.NONE;
+    /** Returns the downsampling info using {@link #maxDepthPerSample} as target coverage. */
+    protected final LIBSDownsamplingInfo getDownsamplingInfo() {
+        return new LIBSDownsamplingInfo(maxDepthPerSample != 0, maxDepthPerSample);
```
droazen: set the second argument to the `LIBSDownsamplingInfo` to -1 in the case where we're not downsampling, as per the method javadoc

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
```diff
@@ -64,6 +66,10 @@
     @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME, shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "Output file (if not provided, defaults to STDOUT).", optional = true)
     public File outFile = null;
 
+    // This is to check for previous versions of samtools without having overlap detection
+    @Argument(fullName = "ignore_overlaps", shortName = "x", doc = "Disable read-pair overlap detection.", optional = true)
```
droazen: `x` is not a good `shortName` for this argument -- recommend setting `shortname` to `ignore_overlaps` as well.

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
droazen: Just adjust the comment to read "LIBS can perform per-sample downsampling to a coverage target if requested", and remove the TODO.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -237,6 +240,71 @@ public Locatable getLocation() {
...
+    public void fixOverlaps() {
+        final FragmentCollection<PileupElement> fragments = FragmentCollection.create(this);
+        fragments.getOverlappingPairs().stream()
+                .forEach(
+                        elements -> fixPairOverlappingQualities(elements.get(0), elements.get(1))
+                );
+    }
+
+    /**
+     * Fixes the quality of two elements that come from an overlaping pair in the same way as
```
droazen: overlaping -> overlapping

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -237,6 +240,71 @@ public Locatable getLocation() {
...
+                }
+                secondQuals[secondElement.getOffset()] = 0;
+            } else {
+                // if not, we lost confidence in the one with higher quality
+                if (firstElement.getQual() >= secondElement.getQual()) {
+                    firstQuals[firstElement.getOffset()] =
+                            (byte) (0.8 * firstQuals[firstElement.getOffset()]);
+                    secondQuals[secondElement.getOffset()] = 0;
+                } else {
+                    secondQuals[secondElement.getOffset()] =
```
droazen: Declare the magic value `0.8` as a named `static final` constant at the top of the class, with a name that makes it clear the value comes from samtools, and use the constant in this method instead of the literal value.

src/test/java/org/broadinstitute/hellbender/utils/pileup/ReadPileupUnitTest.java
```diff
@@ -490,4 +492,124 @@ public void testSimplePileupWithOffset(){
...
+        // Thus, we use them to test if it is not modified. This could be checked uncommenting the next two lines:
+        // Assert.assertSame(deleted.getRead(), copy);
+        // Assert.assertSame(normal.getRead(), copy2);
+        Assert.assertEquals(copy, read);
+        Assert.assertEquals(copy2, read2);
+        ReadPileup.fixPairOverlappingQualities(normal, deleted);
+        Assert.assertEquals(copy, read);
+        Assert.assertEquals(copy2, read2);
+    }
+
```
droazen: Add at least one unit test that calls `ReadPileup.fixOverlaps()`.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/CheckPileup.java
magicDGS: I used the sames as in samtools mpileup, but it is true that it is not a good short name. I changed it.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
magicDGS: fixed

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
magicDGS: Done, although I guess that the name for the constant (`SAMTOOLS_OVERLAP_LOW_CONFIDENCE`) is not great. Any suggestion?

src/test/java/org/broadinstitute/hellbender/utils/pileup/ReadPileupUnitTest.java
magicDGS: I use exactly the same test `DataProvider` to create a `ReadPileup` with the two `PileupElement` and call `ReadPileup.fixOverlaps()`.  Is that enought?

droazen: Yes, that should be fine

magicDGS:
Rebase/squash and addressed all comments. `CLOUD=mandatory` is not going to pass, but the no-cloud test will. Back to you @droazen. Thanks a lot for reviewing @droazen and @fleharty.

droazen:
:+1: merging

Issue #2139, opened 2016-09-02
Request: Allow API user to include packages for codecs


magicDGS:
`FeatureManager.CODEC_PACKAGES` contains only htsjdk and hellbender codecs, but other users could develop their own codecs, and an static method for add a codec package will be interesing from the API user point of view.

magicDGS:
This requires also a finer control for the codecs, once the configuration-code is implemented, to ignore default packages, and include/exclude single classes.

droazen:
Should be resolved by https://github.com/broadinstitute/gatk/pull/3447

magicDGS:
@droazen - That won't be solved by the current #3447, because there is no way of fine-tune the codecs: I require to being able to add/remove concrete classes, and exclude codecs from a concrete package. An example is a custom codec implementation for some feature, to provide extra-validation for the downstream toolkit. This would be even more useful if HTSJDK is moving to an interface-based library...

droazen:
This ticket is just to make the codec packages configurable, which would be resolved by https://github.com/broadinstitute/gatk/pull/3447. If you need more fine-grained control than this, we could discuss as part of a separate ticket.

droazen:
This is now configurable via Owner -- closing.

PR #2156, opened 2016-09-08
Settable packages for FeatureManager

magicDGS:
Solves #2139

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/FeatureManager.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/FeatureManager.java
@@ -44,9 +44,9 @@ public final class FeatureManager implements AutoCloseable {
     /**
      * We will search these packages at startup to look for FeatureCodecs
      */
-    private static final List<String> CODEC_PACKAGES = Arrays.asList("htsjdk.variant",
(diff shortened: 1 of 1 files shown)


lbergelson:
@magicDGS We like the idea of making this configurable.  We're not sure that this is the best mechanism for doing so though.  We've been thinking about possibly including a standardized mechanism for configuring properties, i.e. some property file in the jar that could then be overridden by downstream projects.  We don't know a great mechanism for doing that though.  

Do you have any thoughts?  Maybe using something like [commons-configuration](https://commons.apache.org/proper/commons-configuration/userguide/user_guide.html)

It seems like we have a number of places that need this sort of configuration, and it would be good if we had a standard way of doing so instead of relying on little hacks for each instance.

droazen:
See also https://github.com/broadinstitute/gatk/issues/2297

magicDGS:
@lbergelson @droazen, what's about a `GATKConf` class containing settable values? Defaults are the current ones and it's up to the API user if a configuration file or hardcoded solution is implemented, or even command line if they  need it.

magicDGS:
Should I close this PR? Although I would like to have something like this soon...

droazen:
Closing in favor of https://github.com/broadinstitute/gatk/pull/2322

PR #2163, opened 2016-09-13
Fixed errors in the GVCF extension for IndexFeatureFile

vdauwera:
GVCF extension was whimsically wrong in a few places.

Will probably require updating some tests.


droazen:
Pushed needed updates to test data here, will merge once tests pass.

lbergelson:
Can we make this test for either `g.vcf` or `gvcf` instead of one or the other?  Both seem reasonable to me.

droazen:
No, there's already been standardization of the extension in practice, so we should follow the de facto standard.

PR #2167, opened 2016-09-14
replacing coveralls with codecov

lbergelson:
wanted to see how codecov.io works.  don't merge this.  It looks like they fixed all the problems we have with coveralls though and actually generate a useful report.  They have a browser extension that lets you view which lines are covered by tests in github.  It seems like it would actually be pretty useful in code reviews.

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,1 +1,0 @@
-service_name: travis-pro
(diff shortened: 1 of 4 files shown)


lbergelson:
@droazen I think we should try out codecov.  I've played with it a little bit and it seems like it would be an improvement over the abysmal coveralls at the very least.  At the most it seems like it might be actually useful during pull requests because it seems like it can produce a useful view of newly covered  and uncovered lines.  It might need some configuration to be maximally useful with minimal spam, and it's hard to know exactly what configuration it should be set with until we have it set up in master.

It reports lower overall code coverage than coveralls did because it includes a notion of "partial coverage"

PR #2168, opened 2016-09-15
Fix a bug responsible for creating duplicate inversion calls

cwhelan:
This PR primarily fixes a bug in the hashCode method of BreakpointAllele that was causing Spark's groupByKey() method to not group identical inversion events together, resulting in duplicate variant calls in the output VCF. 

All changes:

- Fix hashCode of BreakpointAllele to use the ordinal of the InversionType enum, which is consistent across executors
- Added Kryo Serializers to several SV classes
- Made the INSERTED_SEQUENCE_MAPPINGS VCF annotation include the assembly id and contig name for each mapping, which is necessary to read them.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
@@ -1,5 +1,9 @@
 package org.broadinstitute.hellbender.tools.spark.sv;
 
+import com.esotericsoftware.kryo.DefaultSerializer;
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
 import com.github.lindenb.jbwa.jni.AlnRgn;
 import htsjdk.samtools.Cigar;
 import htsjdk.samtools.CigarElement;
@@ -9,9 +13,11 @@ import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.read.CigarUtils;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
 
+import java.io.Serializable;
 import java.util.List;
 import java.util.Objects;
 
+@DefaultSerializer(AlignmentRegion.Serializer.class)
 class AlignmentRegion {
 
     final String contigId;
@@ -67,6 +73,19 @@ class AlignmentRegion {
         } else {
             this.mismatches = 0;
         }
+    }
+
+    public AlignmentRegion(final Kryo kryo, final Input input) {
+        this.contigId = input.readString();
+        this.assemblyId = input.readString();
+        this.forwardStrandCigar = TextCigarCodec.decode(input.readString());
+        this.forwardStrand = input.readBoolean();
+        this.referenceInterval = kryo.readObject(input, SimpleInterval.class);
+        this.mapqual = input.readInt();
+        this.startInAssembledContig = input.readInt();
+        this.endInAssembledContig = input.readInt();
+        this.assembledContigLength = input.readInt();
+        this.mismatches = input.readInt();
     }
 
     public int overlapOnContig(final AlignmentRegion other) {
@@ -167,10 +186,10 @@ class AlignmentRegion {
     }
 
     @Override
-    public boolean equals(Object o) {
+    public boolean equals(final Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
-        AlignmentRegion that = (AlignmentRegion) o;
+        final AlignmentRegion that = (AlignmentRegion) o;
         return forwardStrand == that.forwardStrand &&
                 mapqual == that.mapqual &&
                 startInAssembledContig == that.startInAssembledContig &&
@@ -189,6 +208,33 @@ class AlignmentRegion {
     }
 
     public String toPackedString() {
-        return "" + startInAssembledContig + "-" + endInAssembledContig + ":" + referenceInterval.getContig() + ',' + referenceInterval.getStart() + ',' + (forwardStrand ? '+' : '-') + ',' + TextCigarCodec.encode(forwardStrandCigar) + ',' + mapqual + ',' + mismatches;
+        return assemblyId + "-" + contigId + ":" + startInAssembledContig + "-" + endInAssembledContig + ":" + referenceInterval.getContig() + ',' + referenceInterval.getStart() + ',' + (forwardStrand ? '+' : '-') + ',' + TextCigarCodec.encode(forwardStrandCigar) + ',' + mapqual + ',' + mismatches;
     }
+
+    public static final class Serializer extends com.esotericsoftware.kryo.Serializer<AlignmentRegion> {
+        @Override
+        public void write(final Kryo kryo, final Output output, final AlignmentRegion alignmentRegion ) {
+            alignmentRegion.serialize(kryo, output);
+        }
+
+        @Override
+        public AlignmentRegion read(final Kryo kryo, final Input input, final Class<AlignmentRegion> klass ) {
+            return new AlignmentRegion(kryo, input);
+        }
+    }
+
+    private void serialize(final Kryo kryo, final Output output) {
+        output.writeString(contigId);
+        output.writeString(assemblyId);
+        output.writeString(TextCigarCodec.encode(forwardStrandCigar));
+        output.writeBoolean(forwardStrand);
+        kryo.writeObject(output, referenceInterval);
+        output.writeInt(mapqual);
+        output.writeInt(startInAssembledContig);
+        output.writeInt(endInAssembledContig);
+        output.writeInt(assembledContigLength);
+        output.writeInt(mismatches);
+
+    }
+
 }
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointAllele.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointAllele.java
@@ -1,7 +1,12 @@
 package org.broadinstitute.hellbender.tools.spark.sv;
 
+import com.esotericsoftware.kryo.DefaultSerializer;
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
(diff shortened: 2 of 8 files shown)


cwhelan:
@tedsharpe could you review?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
```diff
@@ -9,10 +13,13 @@
 import org.broadinstitute.hellbender.utils.read.CigarUtils;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
 
+import java.io.Serializable;
 import java.util.List;
 import java.util.Objects;
 
-class AlignmentRegion {
+@DefaultSerializer(AlignmentRegion.Serializer.class)
+class AlignmentRegion implements Serializable {
```
tedsharpe: I'd recommend removing `implements Serializable` and the `serialVersionUID` on the next line.  Since we intend to use Kryo serialization exclusively, mixing in Java serializability just confuses the issue.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
```diff
@@ -69,6 +76,19 @@ public AlignmentRegion(final GATKRead read) {
         }
     }
 
+    public AlignmentRegion(final Kryo kryo, final Input input) {
+        this.contigId = input.readString();
+        this.assemblyId = input.readString();
+        this.forwardStrandCigar = TextCigarCodec.decode(input.readString());
```
tedsharpe: If you haven't already, you should check that `Cigar.toString()` and `TextCigarCodec.decode(String)` are precise inverses in all cases.  Maybe there's a unit test that already does this.  I haven't looked.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointAllele.java
```diff
@@ -1,40 +1,57 @@
...
+        final int start1 = input.readInt();
+        final int end1 = input.readInt();
+        this.leftAlignedLeftBreakpoint = new SimpleInterval(contig1, start1, end1);
+        final String contig2 = input.readString();
+        final int start2 = input.readInt();
+        final int end2 = input.readInt();
+        this.leftAlignedRightBreakpoint = new SimpleInterval(contig2, start2, end2);
+        this.insertedSequence = input.readString();
+        this.homology = input.readString();
+        this.inversionType = InversionType.valueOf(input.readString());
```
tedsharpe: Trivial improvement, but you might as well serialize inversionType as an int.  So this line would become:
`        this.inversionType = InversionType.values()[input.readInt()];`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointAllele.java
```diff
@@ -51,20 +68,56 @@ private InversionType getInversionType(final boolean fiveToThree, final boolean
...
+    private void serialize(final Kryo kryo, final Output output) {
+        output.writeString(leftAlignedLeftBreakpoint.getContig());
+        output.writeInt(leftAlignedLeftBreakpoint.getStart());
+        output.writeInt(leftAlignedLeftBreakpoint.getEnd());
+        output.writeString(leftAlignedRightBreakpoint.getContig());
+        output.writeInt(leftAlignedRightBreakpoint.getStart());
+        output.writeInt(leftAlignedRightBreakpoint.getEnd());
+        output.writeString(insertedSequence);
+        output.writeString(homology);
+        output.writeString(inversionType.toString());
```
tedsharpe: And if you take my suggestion to serialize inversionType as an int, this line would become:
`        output.writeString(inversionType.ordinal());`

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointAlleleTest.java
```diff
@@ -32,4 +32,23 @@ public void testIsInversion() throws Exception {
...
+
+        final AlignmentRegion region3 = new AlignmentRegion("2", "contig-2", TextCigarCodec.decode("100M"), true, new SimpleInterval("1", 10000, 10100), 60, 1, 100, 0);
+        final AlignmentRegion region4 = new AlignmentRegion("2", "contig-2", TextCigarCodec.decode("100M"), false, new SimpleInterval("1", 20100, 20200), 60, 101, 200, 0);
+        final ArrayList<String> insertionMappings2 = new ArrayList<>();
+        insertionMappings2.add("bar");
+        final BreakpointAlignment breakpoint2 = new BreakpointAlignment("contig-2", region3, region4, "", "ACAC", insertionMappings2);
+
+        Assert.assertEquals(breakpoint1.getBreakpointAllele(), breakpoint2.getBreakpointAllele());
+        Assert.assertEquals(breakpoint1.getBreakpointAllele().hashCode(), breakpoint2.getBreakpointAllele().hashCode());
+    }
```
tedsharpe: You could consider adding another test that exercised serialization (though I won't insist on it).  See `BreakpointEvidenceTest.serializationTest()` for inspiration.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointAllele.java
```diff
@@ -51,20 +68,56 @@ private InversionType getInversionType(final boolean fiveToThree, final boolean
...
+            return false;
+        if (homology != null ? !homology.equals(that.homology) : that.homology != null) return false;
+        return inversionType == that.inversionType;
+
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(leftAlignedLeftBreakpoint, leftAlignedRightBreakpoint, insertedSequence, homology, inversionType);
+        int result = leftAlignedLeftBreakpoint != null ? leftAlignedLeftBreakpoint.hashCode() : 0;
```
tedsharpe: You can still use `Objects.hash` if you'd like.  Like this:
`    return Objects.hash(leftAlignedLeftBreakpoint, leftAlignedRightBreakpoint, insertedSequence, homology, inversionType.ordinal());`
or, slightly better:
`    return Objects.hash(leftAlignedLeftBreakpoint, leftAlignedRightBreakpoint, insertedSequence, homology, 2659*inversionType.ordinal());`
Choose some random, moderately sized prime to give your hash better avalanche.

tedsharpe: I have a favor to ask, too:  To avoid a separate PR for a trivial one-line change, could you repair the same non-deterministic behavior in `BreakpointEvidence.hashCode`?
Change line 156 in BreakpointEvidence.java from
`        result = mult * result + templateEnd.hashCode();`
to
`        result = mult * result + templateEnd.ordinal();`

SHuang-Broad:
@cwhelan , while you are at it, I believe the following is unintended:

in `GATKSVVCFHeaderLines.java`

the static initializer for the map has 

```java
vcfHeaderLines.put(INSERTED_SEQUENCE, new VCFInfoHeaderLine(INSERTED_SEQUENCE, VCFHeaderLineCount.UNBOUNDED, VCFHeaderLineType.String, "Inserted sequence at the breakpoint"));
vcfHeaderLines.put(INSERTED_SEQUENCE, new VCFInfoHeaderLine(INSERTED_SEQUENCE_MAPPINGS, VCFHeaderLineCount.UNBOUNDED, VCFHeaderLineType.String, "Alignments of inserted sequence"));
```

The second line should have `INSERTED_SEQUENCE_MAPPINGS` as its key.

SHuang-Broad:
Another issue is that the `INSERTED_SEQUENCE_MAPPINGS` field generated for some of the records are separated by `;`, which is also the separator for different info fields.
One example

```
1      	29720840       	INV_3_TO_5_1_29720840_30878819 	G      	<INV>  	.      	.      	ALIGN_LENGTHS=144;ASSEMBLY_IDS=256;CONTIG_IDS=>contig-3_561_0;END=30878819;HQ_MAPPINGS=1;
INSERTED_SEQUENCE=AAAGCACATAGTAGGTGCACAAAAAAAATTTGTTGAATGAGTGAATGACATTTTATTCACAGATATGCCTTGGTTTCCAGAATGACCCTGTGGTTTGGAGGGTGGCTGCTGTATCTCTATTT;
INSERTED_SEQUENCE_MAPPINGS=160-211:1,30878513,+,159S19M1D33M350S,33,1;219-243:5,108445181,-,218S25M318S,2,0;245-264:1,24507681,-,244S20M297S,0,0;
INV_3_TO_5;MAPPING_QUALITIES=60;MAX_ALIGN_LENGTH=144;SVLEN=1157979;SVTYPE=INV;TOTAL_MAPPINGS=1
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
cwhelan: `Cigar.toString()` calls `TextCigarCodec.encode()` which is pretty straightforward. `TextCigarCodec.decode()` is slightly less straightforward but still looks like it does what we want. I think it's ok to trust that `TextCigarCodec` is doing what it's supposed to. For clarity I changed line 229 to use `TextCigarCodec.encode()` instead of `toString`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointAllele.java
cwhelan: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointAllele.java
cwhelan: Done for the first comment; I like the Objects.hash way better and had just changed it while experimenting on debugging this issue.

cwhelan: Done for the second comment

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointAllele.java
cwhelan: done

cwhelan:
@SHuang-Broad fixed `GATKSVVCFHeaderLines` to use `INSERTED_SEQUENCE_MAPPINGS` as the correct key, thanks for the catch.

cwhelan:
@SHuang-Broad The code in this PR already fixes the issue of using semicolons within annotation fields; I replaced it with a `|` in `CallVariantsFromAlignedContigsSpark:410`

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointAlleleTest.java
cwhelan: Added a very small unit test that round trips a BreakpointAllele

cwhelan:
@SHuang-Broad @tedsharpe Tried to address your comments. Look good now?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
tedsharpe: That will give a reassuring appearance of symmetry.  Thanks.

tedsharpe:
Yup.  Thumbs up from me.

SHuang-Broad:
No more complaints from me either...

PR #2169, opened 2016-09-15
Write splitting-bai files when writing (non-sharded) BAM files from Spark

tomwhite:
Splitting-bai files are used to find split points in BAM files so they can be read in parallel without having to heuristically find suitable split points.

Note that for this change splitting-bai files are only written in non-sharded mode, since sharded mode has issues with the splitting files being read accidentally by the input format. Also, typically each has one split (typically), so there's less need for the splitting index.


lbergelson:
That's excellent!  👍

Issue #2158, opened 2016-09-08
Add tests for AlleleSubsettingUtils


lbergelson:
There is no `AlleleSubsettingUtilsUnitTest`.  Tests for `subsetAlleles` live `GATKVariantContextUtilsUnitTest` and there are no tests for `calculateMostLikelyAlleles`.

fleharty:
@lbergelson Is this issue done, I was looking for things to contribute to, and was going to add these tests, but it looks like they are already there.

droazen:
This one is done! Closing.

PR #2170, opened 2016-09-16
calculateMostLikelyAlleles deals with ties correctly

### LINKED ISSUE #2157: Fix ties in AlleleSubsettingUtils.calculateMostLikelyAlleles
Currently AlleleSubsettingUtils.calculateMostLikelyAlleles can subset to 0 alt alleles in the case that the their are two alleles with equal likelihoods.  This should be fixed so that it chooses as many of the top alleles as it can.

lbergelson:
AlleleSubsettingUtils.calculateMostLikelyAlleles was incorrectly removing all alleles if the best alleles were tied and numAltAllelesToKeep=1
fixes #2157

move tests into AlleleSubsettingUtilsUnitTest from GATKVariantContextUtilsUnitTest as part of #2158
a few typo fixes that I noticed


src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtilsUnitTest.java
```diff
@@ -0,0 +1,219 @@
...
+                new VariantContextBuilder(vcBase).alleles(ACG).genotypes(new GenotypeBuilder(base).alleles(AA).AD(homG3AllelesAD).PL(homG3AllelesPL).
+                        attribute(GATKVCFConstants.STRAND_COUNT_BY_SAMPLE_KEY, homG3AllelesSAC).make()).make(),
+                new VariantContextBuilder(vcBase).alleles(AG).make(),
+                Collections.singletonList(new GenotypeBuilder(base).alleles(AA).PL(new double[]{-20, -40, 0}).AD(new int[]{0, 21}).
+                        attribute(GATKVCFConstants.STRAND_COUNT_BY_SAMPLE_KEY, new int[]{0, 0, 21, 21}).GQ(100).make())});
+
+        return tests.toArray(new Object[][]{});
+    }
+
+
```
lbergelson: @davidbenjamin  everything before here in this file is just moved from GATKVariantContextUtilsUnit test and shouldn't need review.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -171,20 +172,55 @@ else if ( clazz.equals(int[].class) ) {
...
+        final List<Allele> mostLikelyAlleles = new ArrayList<>();
+        for ( int i = 0; i < numAlleles; i++){
+            if( i == 0 || i == nonRefAltAlleleIndex || properAltIndexesToKeep.contains(i) ){
+                mostLikelyAlleles.add(alleles.get(i));
+            }
+        }
+        return mostLikelyAlleles;
+    }
+
+    private static double[] calculateLikelihoodSums(final VariantContext vc, final int defaultPloidy) {
```
lbergelson: there aren't any tests that check the values this is creating, I figure you would probably be the best person for that

lbergelson:
@davidbenjamin Could you review these changes.  I added tests to check that my filtering is doing the right thing, but there aren't any tests that check that the likelihood calculation is correct.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -171,20 +172,56 @@ else if ( clazz.equals(int[].class) ) {
...
+    @VisibleForTesting
+    static List<Allele> filterToMaxNumberOfAltAllelesBasedOnScores(int numAltAllelesToKeep, List<Allele> alleles, double[] likelihoodSums) {
+        final int nonRefAltAlleleIndex = alleles.indexOf(GATKVCFConstants.NON_REF_SYMBOLIC_ALLELE);
+        final int numAlleles = alleles.size();
+        final Set<Integer> properAltIndexesToKeep = IntStream.range(1, numAlleles).filter(n -> n != nonRefAltAlleleIndex).boxed()
+                                                             .sorted(Comparator.comparingDouble((Integer n) -> likelihoodSums[n]).reversed())
+                                                             .limit(numAltAllelesToKeep)
+                                                             .collect(Collectors.toSet());
+
+        final List<Allele> mostLikelyAlleles = new ArrayList<>();
```
davidbenjamin: Personally I prefer
```java
return IntStream.range(0, numAlleles)
    .filter(n -> n == 0 || n == nonRefAltAlleleIndex || properAltIndexesToKeep.contains(n))
```
but I wouldn't even mention it if it didn't match the stream above.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
davidbenjamin: You can file the issue and assign me.  I'm pretty confident that it works for the purposes of your PR getting merged because this is old code that I simply moved into `AlleleSubsettingUtils`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
lbergelson: Done.  Good call.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
lbergelson: Yeah, not your fault that it's untested, but we should have some test for it I think.  I created #2178, but it's not a super high priority.

lbergelson:
@davidbenjamin Thank you for the quick review!  I really appreciate it.

Issue #1233, opened 2015-12-02
UnmarkDuplicates needs a CRAM test


akiezun:

fleharty:
Addresses #1233

fleharty:
closed via #2174

PR #2174, opened 2016-09-16
Adding cram test to UnmarkDuplicatesIntergrationTest

fleharty:
Adds a simple test to test a cram in UnmarkDuplicatesIntergrationTest


fleharty:
Addresses #1233

droazen:
@cmnbroad please review

cmnbroad:
@fleharty Can you replace the .crai with a .bai ? We're trying to discourage proliferation of .crai files due to issues with  the internal .crai->.bai conversion done by htsjdk. Any recent gatk4 will produce a .bai if you run the .cram file through PrintReads, just make sure to submit both the .cram and .crai that are produced, since the output .cram will almost certainly be different than the input one. Otherwise looks good - thanks for adding this test.

fleharty:
@cmnbroad I realized that this test doesn't use the .crai at all, and if I create a .bai it won't use that either.  Perhaps it would be better to remove the index file all-together?

PR #2179, opened 2016-09-19
Added an ExampleAssemblyRegionWalker to provide a full integration test in public for assembly region traversal

### LINKED ISSUE #2172: Create ExampleAssemblyRegionWalker
GATK public doesn't have any example of an `AssemblyRegionWalker` and no integration tests.

This makes it easy to accidentally break the haplotype caller without realizing it.  We should have an ExampleAssemblyRegionWalker and integration tests for it to prevent this.

droazen:
Previously, the only integration tests involving assembly region traversal were in gatk-protected,
which led to breakage.

Resolves #2172


droazen:
@lbergelson please review

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -210,15 +210,19 @@ public SimpleInterval getCurrentReadShardBounds() {
...
      * Note: this method is called before command line parsing begins, and thus before a SAMFileHeader is
      * available through {link #getHeaderForReads}.
      *
      * @return List of default filter instances to be applied for this tool.
      */
     public List<ReadFilter> getDefaultReadFilters() {
-        return Collections.singletonList(new WellformedReadFilter());
+        final List<ReadFilter> defaultFilters = new ArrayList<>(2);
+        defaultFilters.add(new WellformedReadFilter());
+        defaultFilters.add(new ReadFilterLibrary.MappedReadFilter());
```
lbergelson: Why add a mapped read filter here?  Doesn't haplotype caller use the unmapped mates?

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleAssemblyRegionWalker.java
```diff
@@ -0,0 +1,109 @@
...
+        return (locusPileup, referenceContext, featureContext) -> new ActivityProfileState(new SimpleInterval(locusPileup), 1.0);
+    }
+
+    @Override
+    public void onTraversalStart() {
+        try {
+            outputStream = outputFile != null ? new PrintStream(outputFile) : System.out;
+        }
+        catch ( final FileNotFoundException e ) {
+            throw new UserException.CouldNotReadInputFile(outputFile, e);
```
lbergelson: I don't think CountNotReadInput file is the right exception for couldn't create and output stream.  It's a bit strange that `FileNotFoundException` is the exception it throws.

src/test/java/org/broadinstitute/hellbender/tools/examples/ExampleAssemblyRegionWalkerIntegrationTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Test
+    public void testExampleAssemblyRegionWalker() throws Exception {
+        IntegrationTestSpec testSpec = new IntegrationTestSpec(
+                " -R " + b37_reference_20_21 +
+                " -I " + NA12878_20_21_WGS_bam +
+                " -knownVariants " + dbsnp_138_b37_20_21_vcf +
+                " -L 20:10000000-10200000 " +
+                " -O %s",
+                Arrays.asList(TEST_OUTPUT_DIRECTORY + "expected_ExampleAssemblyRegionWalkerIntegrationTest_output.txt")
```
lbergelson: the expected output is about a megabyte.  Maybe cut the region it runs on down?

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
droazen: No, `HaplotypeCaller` filters out all unmapped reads, and the traversal cannot handle them, so this filter should be on by default for AssemblyRegionTraversal.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleAssemblyRegionWalker.java
droazen: Whoops, yes definitely not the right exception type. Will fix.

src/test/java/org/broadinstitute/hellbender/tools/examples/ExampleAssemblyRegionWalkerIntegrationTest.java
droazen: Ok, cut it down to ~250k.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -210,15 +210,19 @@ public SimpleInterval getCurrentReadShardBounds() {
      * Returns the default list of CommandLineReadFilters that are used for this tool. The filters
      * returned by this method are subject to selective enabling/disabling and customization by the
      * user via the command line. The default implementation uses the {@link WellformedReadFilter}
-     * filter with all default options. Subclasses can override to provide alternative filters.
+     * filter with all default options, as well as the {@link ReadFilterLibrary.MappedReadFilter}.
```
droazen: test

PR #2181, opened 2016-09-20
Genomicsdb Spark API - 0.1

kdatta:
Example codes are include to demonstrate the two ways variant contexts can be read to Spark RDD from GenomicsDB. The code takes a loader JSON file, a query JSON file and a hosts file as input.


droazen:
@lbergelson please review

droazen:
@lbergelson What do you want to do with this one? Should we merge this example code?

droazen:
@kdatta Is this example code still current? If not, should we close this? We do eventually want some TileDB example code checked in, but ideally it would take the form of a simple, runnable GATK tool.

kdatta:
Let us close this PR. I am creating a new one which will contain both Java write API and Spark API.

droazen:
Ok, closing!

PR #2182, opened 2016-09-20
MultiVariantWalker implementation

cmnbroad:
MultiVariantWalker implementation, with VariantRecalibrator and ApplyVQSR switched over to the new base class.

GatherVCFs, MergeVCFs, SelectVariants, and possibly SortVCF are all candidates to be changed to MultiVariantWalkers, but are not included here.

Also, this is a branch of the branch that contains the VQSR and ApplyVQSR tests so I could use those tests to verify the MultiVariantWalker. As result there are two commits here, but the tests are already included a separate PR.


src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+     */
+    private MergingIterator<VariantContext> currentIterator;
+
+    /**
+     * Default value for queryLookaheadBases, if none is specified. This is designed to be large enough
+     * so that in typical usage (ie., query intervals with gradually increasing start locations) there will
+     * be a substantial number of cache hits between cache misses, reducing the number of times we need to
+     * repopulate the cache from disk.
+     */
+    public static final int DEFAULT_QUERY_LOOKAHEAD_BASES = 1000;
```
droazen: Can you use `FeatureDataSource.DEFAULT_QUERY_LOOKAHEAD_BASES` directly here?

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+
+    /**
+     * Define the set of incompatibilities that are allowed for SequenceDictionaries in Variant input data sources
+     */
+    private final static Set<SequenceDictionaryUtils.SequenceDictionaryCompatibility> allowedSequenceDictionaryIncompatibilities =
+            new HashSet<>(Arrays.asList(
+                    SequenceDictionaryUtils.SequenceDictionaryCompatibility.COMMON_SUBSET,
+                    SequenceDictionaryUtils.SequenceDictionaryCompatibility.SUPERSET,
+                    SequenceDictionaryUtils.SequenceDictionaryCompatibility.NO_COMMON_CONTIGS));
```
droazen: This can go away

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+
+    /**
+     * Creates a FeatureDataSource backed by the provided File and assigns this data source the specified logical
+     * name. We will look ahead the default number of bases ({@link #DEFAULT_QUERY_LOOKAHEAD_BASES}) during queries
+     * that produce cache misses.
+     *
+     * @param featureFile file containing Features
+     * @param name logical name for this data source (may be null)
+     */
+    public void addFeatureDataSource(final File featureFile, final String name) {
```
droazen: Would requiring that all `FeatureDataSources` be provided at construction time simplify the code at all? If yes, it might be a good change.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+     * Define the set of incompatibilities that are allowed for SequenceDictionaries in Variant input data sources
+     */
+    private final static Set<SequenceDictionaryUtils.SequenceDictionaryCompatibility> allowedSequenceDictionaryIncompatibilities =
+            new HashSet<>(Arrays.asList(
+                    SequenceDictionaryUtils.SequenceDictionaryCompatibility.COMMON_SUBSET,
+                    SequenceDictionaryUtils.SequenceDictionaryCompatibility.SUPERSET,
+                    SequenceDictionaryUtils.SequenceDictionaryCompatibility.NO_COMMON_CONTIGS));
+
+    /**
+     * Creates a FeatureDataSource backed by the provided File and assigns this data source the specified logical
```
droazen: Comments need to be updated with the new class name

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+
+    /**
+     * Merged VCF header used for this (aggregate) source, derived from the individual soruces.
+     */
+    private VCFHeader mergedHeader;
+
+    /**
+     * SAMSequenceDictionary used for this (aggregate) source, derived from the individual sources.
+     */
+    private SAMSequenceDictionary mergedDictionary;
```
droazen: Does the `mergedHeader` contain the `mergedDictionary`? If not, should the dictionary inside the merged header be set to the merged dictionary?

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+     *
+     * @return the sequence dictionary derived from the input sources, or null if no dictionary could be created
+     * from either the header or an index file.
+     */
+    public SAMSequenceDictionary getSequenceDictionary() {
+        validateFeatureSources();
+        if (mergedDictionary == null && null != getHeader()) {
+            validateAllSequenceDictionaries();
+            final VCFHeader header = getHeader();
+            mergedDictionary = header.getSequenceDictionary();
```
droazen: It would be better if at construction time, the merged header were updated with the fully-merged dictionary

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+            if (mergedDictionary == null || mergedDictionary.isEmpty()) {
+                // last resort - try to get a sequence dictionary from one of the feature data sources (which
+                // will create one from the index if necessary)
+                for (FeatureDataSource<VariantContext> fds : featureDataSources) {
```
droazen: The dictionary created from the index will only reflect the variants in that particular `FeatureDataSource`, and may not be usable with the variants from the other data sources, so this fallback may not be a good idea.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+                    }
+                }
+            }
+        }
+        return mergedDictionary;
+    }
+
+    /**
+     * Restricts traversals of this data source via {@link #iterator} to only return Features that overlap the provided
+     //* intervals. Calls to {@link #query(SimpleInterval)} and/or {@link # queryAndPrefetch(SimpleInterval)} are not
```
droazen: Comment formatting slightly off here

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+     * Intervals MUST be non-overlapping and sorted in order of increasing start position, otherwise traversal
+     * results will be incorrect.
+     *
+     * Passing in a null or empty interval List clears the intervals for traversal, making future iterations
+     * over this data source unrestricted by intervals.
+     *
+     * @param intervals Our next full traversal will return only Features overlapping these intervals
+     */
+    public void setIntervalsForTraversal( final List<SimpleInterval> intervals ) {
+        validateFeatureSources();
```
droazen: `validateFeatureSources()` should probably happen at construction time

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+                    .filter(h -> h != null)
+                    .collect(Collectors.toList());
+
+            // Now merge the headers using htsjdk, which is pretty promiscuous, and which only works properly
+            // because of the cross-dictionary validation done in validateAllSequenceDictionaries.
+            mergedHeader = headers.size() > 1 ?
+                    new VCFHeader(VCFUtils.smartMergeHeaders(headers, true)) :
+                    headers.get(0);
+        }
+        return mergedHeader;
```
droazen: As mentioned above, should be created at construction time.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    /**
+     * Permanently close this data source, invalidating any open iteration over it, and making it invalid for future
+     * iterations and queries.
+     */
+    @Override
+    public void close() {
+        validateFeatureSources();
+        featureDataSources.forEach(dataSource -> dataSource.close());
```
droazen: Don't you need to close out any merging iterators that were handed out to clients?

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+    // Cross validate the length and md5 for a pair of sequence records and throw if they're not the same.
+    private void validateSequenceDictionaryRecords(
+            final String sourceDataSourceName,
+            final SAMSequenceDictionary sourceDictionary,
+            final SAMSequenceRecord sourceSequence,
+            final String targetDataSourceName,
+            final SAMSequenceDictionary targetDictionary,
+            final SAMSequenceRecord targetSequence)
+    {
+        if (targetSequence.getSequenceLength() != sourceSequence.getSequenceLength()) {
```
droazen: I think you want to call into the existing `SequenceDictionaryUtils.sequenceRecordsAreEquivalent()` method to do the name and length checks (since it handles some edge cases), and then do the MD5 check on top of it (although checking the MD5s is stricter than we usually are).

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+                    targetSequence.getSequenceName(),
+                    targetSequence.getSequenceLength());
+            throw new UserException.IncompatibleSequenceDictionaries(
+                    msg,
+                    sourceDataSourceName,
+                    sourceDictionary,
+                    targetDataSourceName,
+                    targetDictionary
+            );
+        } else if (!Objects.equals(targetSequence.getMd5(), sourceSequence.getMd5())) {
```
droazen: I think we might want to tolerate the situation where the MD5 is filled in in one record but not the other, and just emit a warning in that case.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+                    targetDataSourceName,
+                    targetDictionary
+            );
+        }
+    }
+
+    /**
+     * Close the iterator currently open over the data sources, if there is one.
+     */
+    private void closeOpenIterationIfNecessary() {
```
droazen: I think your `close()` method should be calling `closeOpenIterationIfNecessary()`

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+        Utils.nonNull(sourceIterator);
+
+        return new CloseableIterator<VariantContext>() {
+            Iterator<VariantContext> delegateIterator = sourceIterator;
+            @Override
+            public void close() { delegateIterator = null; }
+
+            @Override
+            public boolean hasNext() {
+                return delegateIterator.hasNext();
```
droazen: `return delegateIterator != null && delegateIterator.hasNext()`

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+            }
+
+            @Override
+            public VariantContext next() {
```
droazen: throw `NoSuchElementException` if `hasNext()` is false on a call to `next()`

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+                return delegateIterator.next();
+            }
+        };
+    }
+
+    /**
+     * Invalidate the cached VCFHeader and SAMSequenceDictionary whenever the set of backing FeatureDataSources
+     * changed in order to force them to be recreated and cached on demand.
+     */
+    private void invalidateCachedHeaderAndDictionary() {
```
droazen: Can be deleted if initialization of these is tied to construction.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,408 @@
...
+        mergedHeader = null;
+        mergedDictionary = null;
+    }
+
+    /**
+     * Validate the requirement that one or more variant data sources have been added
+     */
+    private void validateFeatureSources() {
```
droazen: This can probably also be deleted if adding data sources is done only at construction.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
```diff
@@ -0,0 +1,103 @@
...
+import org.broadinstitute.hellbender.cmdline.Argument;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Spliterator;
+
+/**
+ * A MultiVariantWalker is a tool that processes one variant at a time from multiple sources of variants, with
+ * optional contextual information from a reference, sets of reads, and/or supplementary sources of Features.
```
droazen: Mention that records from the multiple sources are returned in position order.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * VariantWalker authors must implement the {@link #apply} method to process each variant, and may optionally implement
+ * {@link #onTraversalStart}, {@link #onTraversalSuccess} and/or {@link #closeTool}.
+ */
+public abstract class MultiVariantWalker extends VariantWalkerBase {
+
+    // NOTE: using File rather than FeatureInput<VariantContext> here so that we can keep this driving source
+    //       of variants separate from any other potential sources of Features
+    @Argument(fullName = StandardArgumentDefinitions.VARIANT_LONG_NAME, shortName = StandardArgumentDefinitions.VARIANT_SHORT_NAME, doc = "A VCF file containing variants", common = false, optional = false)
```
droazen: `A VCF file` -> `One or more VCF files`

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    protected final void onStartup() {
+        super.onStartup();
+        if ( hasIntervals() ) {
+            drivingVariants.setIntervalsForTraversal(intervalsForTraversal);
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    protected void initializeDrivingVariants() {
```
droazen: `@Override`

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
```diff
@@ -0,0 +1,103 @@
...
+        drivingVariantFiles.stream().forEach(
+                f -> {
+                    FeatureInput<VariantContext> featureInput = new FeatureInput<>(f, "drivingVariantFile");
+                    if (drivingVariantsFeatureInputs.contains(featureInput)) {
+                        //TODO: should we allow duplicates file names? If so we need to change the FeatureInput hash
```
droazen: I don't think we should allow duplicate file names.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalker.java
```diff
@@ -21,56 +17,21 @@
...
-                return otherDictionary != null ?
-                        otherDictionary :
-                        dictFromDrivingVariants;
-            } else {
-                return dictFromDrivingVariants;
-            }
-        }
-        return super.getBestAvailableSequenceDictionary();
-    }
+    protected Spliterator<VariantContext> getSpliteratorForDrivingVariants() { return drivingVariants.spliterator(); }
```
droazen: `@Override`

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
```diff
@@ -0,0 +1,103 @@
...
+    // NOTE: keeping the driving source of variants separate from other, supplementary FeatureInputs in our FeatureManager
+    // in GATKTool we do add the driving source to the Feature manager but we do need to treat it differently and thus this
+    // field.
+    private MultiVariantDataSource drivingVariants = new MultiVariantDataSource();
+    private List<FeatureInput<VariantContext>> drivingVariantsFeatureInputs = new ArrayList<>(2);
+
+    @Override
+    public boolean requiresFeatures() { return true; }
+
+    protected SAMSequenceDictionary getSequenceDictionaryForDrivingVariants() { return drivingVariants.getSequenceDictionary(); }
```
droazen: `@Override`

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -0,0 +1,127 @@
...
+        return super.getBestAvailableSequenceDictionary();
+    }
+
+    protected abstract void initializeDrivingVariants();
+
+    protected abstract SAMSequenceDictionary getSequenceDictionaryForDrivingVariants();
+
+    protected abstract Spliterator<VariantContext> getSpliteratorForDrivingVariants();
+
+    public abstract VCFHeader getHeaderForVariants();
```
droazen: javadoc for these abstract methods

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -0,0 +1,127 @@
...
+import org.broadinstitute.hellbender.engine.filters.VariantFilter;
+import org.broadinstitute.hellbender.engine.filters.VariantFilterLibrary;
+import org.broadinstitute.hellbender.utils.IndexUtils;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+
+import java.util.Spliterator;
+import java.util.stream.StreamSupport;
+
+/**
+ * Base class for variant walkers, which process one variant at a time from one oe more sources of variants,
```
droazen: `oe` -> `or`

src/main/java/org/broadinstitute/hellbender/utils/SequenceDictionaryUtils.java
```diff
@@ -95,7 +95,12 @@ public static void validateDictionaries( final String name1,
         final boolean requireSuperset = false;
         final boolean checkContigOrdering = false;
 
-        validateDictionaries(name1, dict1, name2, dict2, requireSuperset, checkContigOrdering);
+        final Set<SequenceDictionaryCompatibility> allowedInCompatibilityParams =
```
droazen: These changes to `SequenceDictionaryUtils` can be reverted

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantDataSourceUnitTest.java
```diff
@@ -0,0 +1,145 @@
...
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.Iterator;
+
+public final class MultiVariantDataSourceUnitTest extends BaseTest {
```
droazen: Can you reproduce some of the query test cases from `FeatureDataSourceUnitTest` here?

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantDataSourceUnitTest.java
```diff
@@ -0,0 +1,145 @@
...
+        MultiVariantDataSource multiVariantSource = new MultiVariantDataSource();
+        multiVariantSource.addFeatureDataSource(new File(MULTI_VARIANT_TEST_DIRECTORY, "interleavedVariants_1.vcf"), "interleavedVariants_1.vcf");
+        multiVariantSource.addFeatureDataSource(new File(MULTI_VARIANT_TEST_DIRECTORY, "interleavedVariants_2.vcf"), "interleavedVariants_2.vcf");
+
+        int count = 0;
+        for (final VariantContext vc: multiVariantSource) {
+            count++;
+        };
+        Assert.assertEquals(count, 26);
+    }
```
droazen: Add a test case in which you make assertions about the order in which variants are returned, like in `FeatureDataSourceUnitTest`

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerIntegrationTest.java
```diff
@@ -0,0 +1,157 @@
...
+        args.add(f.getAbsolutePath());});
+        if (interval != null) {
+            args.add("-L");
+            args.add(interval);
+        }
+
+        tool.instanceMain(args.toArray(new String[args.size()]));
+        Assert.assertEquals(tool.count, expectedCount);
+    }
+
```
droazen: Add a test case in which multiple variants from different files are at the same position, and make sure the traversal can handle it.

src/test/java/org/broadinstitute/hellbender/utils/SequenceDictionaryUtilsUnitTest.java
```diff
@@ -214,12 +217,21 @@ public void testSequenceDictionaryValidation( final List<SAMSequenceRecord> firs
                 SequenceDictionaryUtils.getDictionaryAsString(firstDictionary),
                 SequenceDictionaryUtils.getDictionaryAsString(secondDictionary));
         Exception exceptionThrown = null;
+
+        Set<SequenceDictionaryCompatibility> allowedIncompatibilities = Collections.emptySet();
+        if (!requireSuperset) {
+            allowedIncompatibilities = new HashSet<>(Arrays.asList(
+                    SequenceDictionaryUtils.SequenceDictionaryCompatibility.COMMON_SUBSET,
+                    SequenceDictionaryUtils.SequenceDictionaryCompatibility.SUPERSET)
```
droazen: Can be reverted

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Yes. Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: It simplifies things quite a bit. Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Removed.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Yes. This can have the side effect in some cases of emitting a sequence dictionary into the output file that was derived or partially derived from an index, and which contains 0 for sequence length (whereas in the VariantWalker-derived case no dictionary would be emitted since the input header is used). Should we not write such a sequence dictionary out ?

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
cmnbroad: Comment removed.

cmnbroad: Also, this policy is enforced at the MultiVariantWalker level. Should we enforce it at the MultiVariantDataSource level as well ? I don't think there is anything inherent that wouldn't work, so it would be somewhat redundant.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalker.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/utils/SequenceDictionaryUtilsUnitTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/SequenceDictionaryUtils.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerIntegrationTest.java
cmnbroad: Done.

cmnbroad:
@droazen A few questions:

1) The drivingVariants (merged) header is the one used to create the output writer. By pushing the merged sequence dictionary into the merged header, we might write an output file that contains 0 for one or more sequence lengths (if the dictionary came from an index). Should we disallow that? If not then we'll need to be smarter about the merge (I think smartMergeHeader randomly chooses a contig if there is more than one with the same name; we want to force it to choose the non-zero one).

2) If an input file has no embedded dictionary header *and* no index, then we have no dictionary at all, but we need one to create the merging iterator.
- if there is only one input we COULD write special code to not use a merging iterator. Or we could just reject the input. The FeatureDataSourceUnitTest code has such a case.
- If there is more than one input and we have at least one dictionary, but others missing, should we just reject ? This is what I'm currently doing - not sure if there is a sane way to order between these.

cmnbroad:
The updates based on code review are done, but I need feedback on couple of questions above before I can finalize this.

cmnbroad:
Updated per conversation yesterday:
- If a single input is used, no sequence dictionary is required.
- If multiple inputs are used, at least one must have a sequence dictionary.

Other comments:
- We should consider making FeatureDataSource do the same header/dictionary reconciliation that we're doing here
- In theory you should be able to intermix GenomicsDB/file inputs, but there are no tests for that here (separate ticket ?)
- There is currently no protection against writing out a VCF with a header that contains an index-derived dictionary with zero-len sequences
- I added a GATK3 compatibility arg to VariantRecalibrator in this commit since I was using it for perf testing.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -0,0 +1,144 @@
...
+     * have been prepared in {@link #initializeDrivingVariants}
+     * @return VCFHeader to be used for the driving variants.
+     */
+    public abstract VCFHeader getHeaderForVariants();
+
+    /**
+     * Return the primary sequence dictionary to be used for the driving variants for this tool. The value returned
+     * will usually have been prepared in {@link #initializeDrivingVariants}
+     */
+    protected abstract SAMSequenceDictionary getSequenceDictionaryForDrivingVariants();
```
droazen: Shouldn't this always be the same as `getHeaderForVariants().getSequenceDictionary()`?

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -294,7 +298,14 @@
             shortName = "allPoly",
             doc = "Trust that all the input training sets' unfiltered records contain only polymorphic sites to drastically speed up the computation.",
             optional=true)
-    private Boolean TRUST_ALL_POLYMORPHIC = false;
+    private boolean TRUST_ALL_POLYMORPHIC = false;
+
+    // Temporary argument for validation of GATK4 implementation against GATKs results:
+    @Advanced
+    @Argument(fullName="gatk3Compatibility", shortName="gatk3",
+            doc="Set initial random number generator state for compatibility with GATK3 results.",
```
droazen: "compatibility with" -> "an exact match against"

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantDataSourceUnitTest.java
```diff
@@ -0,0 +1,350 @@
...
+
+    @Test
+    public void testGetSequenceDictionaryAlternate() {
+        List<FeatureInput<VariantContext>> featureInputs = new ArrayList<>();
+        featureInputs.add(new FeatureInput<>(baseVariants.getAbsolutePath(), "baseVariants"));
+        featureInputs.add(new FeatureInput<>(baseVariantsAlternateDictionary.getAbsolutePath(), "baseVariantsAlternateDictionary"));
+
+        try (final MultiVariantDataSource multiVariantSource =
+                     new MultiVariantDataSource(featureInputs, FeatureDataSource.DEFAULT_QUERY_LOOKAHEAD_BASES)) {
+            multiVariantSource.getSequenceDictionary();
```
droazen: No assertion here? (and no expected exception)

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerIntegrationTest.java
```diff
@@ -0,0 +1,238 @@
...
+            summary = "TestGATKToolWithFeatures",
+            oneLineSummary = "TestGATKToolWithFeatures",
+            programGroup = TestProgramGroup.class
+    )
+    private static final class TestMultiVariantWalker extends MultiVariantWalker {
+
+        int count = 0;
+        SimpleInterval locus;
+
+        public void apply(
```
droazen: `@Override`

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerIntegrationTest.java
```diff
@@ -0,0 +1,238 @@
...
+    @CommandLineProgramProperties(
+            summary = "TestGATKToolWithFeatures",
+            oneLineSummary = "TestGATKToolWithFeatures",
+            programGroup = TestProgramGroup.class
+    )
+    private static final class TestMultiVariantWalkerIterator extends MultiVariantWalker {
+        String expectedIDOrder[];
+        int count = 0;
+
+        public void apply(
```
droazen: `@Override`

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,372 @@
...
+     */
+    final private List<FeatureDataSource<VariantContext>> featureDataSources = new ArrayList<FeatureDataSource<VariantContext>>();
+
+    /**
+     * Merged VCF header used for this (aggregate) source, derived from the individual soruces.
+     */
+    final private VCFHeader mergedHeader;
+
+    /**
+     * SAMSequenceDictionary used for this (aggregate) source, derived from the individual sources.
```
cmnbroad: Yes, it is, so this be removed. Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,372 @@
...
+        //
+        // 1) Cross validate the sequence dictionaries from each data source (which may be derived from an index
+        //    in the case where its not embedded in the input file) to ensure they're mutually compatible
+        // 2) Create and cache a merged header using versions of the individual headers from each data source that
+        //    have been updated to include the actual dictionary returned from that data source
+        // 3) Retrieve and cache the aggregate sequence dictionary from the merged header
+        //
+        validateAllSequenceDictionaries();
+        mergedHeader = getMergedHeader();
+        mergedDictionary = getMergedSequenceDictionary(mergedHeader);
```
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,372 @@
...
+     * Gets the merged header associated with this data source
+     *
+     * @return header associated with this data source as an Object
+     */
+    public VCFHeader getHeader() {
+        return mergedHeader;
+    }
+
+    /**
+     * Restricts traversals of this data source via {@link #iterator} to only return Features that overlap the provided
```
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,372 @@
...
+     * @return
+     */
+    private Iterator<VariantContext> getMergedIteratorFromDataSources(
+            final Function<FeatureDataSource<VariantContext>, Iterator<VariantContext>> iteratorFromSource) {
+
+        // Tribble documentation states that having multiple iterators open simultaneously over the same FeatureReader
+        // results in undefined behavior
+        closeOpenIterationIfNecessary();
+
+        if (getSequenceDictionary() != null) {
```
cmnbroad: Right. Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,372 @@
...
+     * Permanently close this data source, invalidating any open iteration over it, and making it invalid for future
+     * iterations and queries.
+     */
+    @Override
+    public void close() {
+        closeOpenIterationIfNecessary();
+        featureDataSources.forEach(dataSource -> dataSource.close());
+    }
+
+    private SAMSequenceDictionary getMergedSequenceDictionary(VCFHeader header) {
```
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,372 @@
...
+        // Now merge the headers using htsjdk, which is pretty promiscuous, and which only works properly
+        // because of the cross-dictionary validation done in validateAllSequenceDictionaries.
+        return headers.size() > 1 ?
+                new VCFHeader(VCFUtils.smartMergeHeaders(headers, true)) :
+                headers.get(0);
+    }
+
+    /**
+     * We want the headers that are used to create the merged header to have the same sequence dictionary
+     * that was used returned from the data source and used during validation (which may or may not be the
```
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,372 @@
...
+     * one that was embedded in the input file itself), so get the embedded one from the data source and
+     * update it to include the actual sequence dictionary.
+     */
+    //TODO: should FeatureDataSource do this itself ? Right now it returns a seq dict that may not match
+    // the header it returns.
```
cmnbroad: Comment removed.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
```diff
@@ -0,0 +1,105 @@
...
+    public List<String> drivingVariantFiles = new ArrayList<String>();
+
+    // NOTE: keeping the driving source of variants separate from other, supplementary FeatureInputs in our FeatureManager
+    // in GATKTool we do add the driving source to the Feature manager but we do need to treat it differently and thus this
+    // field.
+    private MultiVariantDataSource drivingVariants;
+    private List<FeatureInput<VariantContext>> drivingVariantsFeatureInputs = new ArrayList<>(2);
+
+    @Override
+    public boolean requiresFeatures() { return true; }
```
cmnbroad: Removed.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
```diff
@@ -0,0 +1,105 @@
...
+                        throw new IllegalArgumentException("Feature inputs must be unique: " + featureInput.toString());
+                    }
+                    drivingVariantsFeatureInputs.add(featureInput);
+
+                    //Add the driving datasource to the feature manager too so that it can be queried. Setting lookahead to 0 to avoid caching.
+                    //Note: we are disabling lookahead here because of windowed queries that need to "look behind" as well.
+                    features.addToFeatureSources(0, featureInput, VariantContext.class);
+                }
+        );
+        drivingVariants = new MultiVariantDataSource(drivingVariantsFeatureInputs, FeatureDataSource.DEFAULT_QUERY_LOOKAHEAD_BASES);
```
cmnbroad: Good catch. Done.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
cmnbroad: Yes, though it took me a minute to convince myself. I'll implement it that way and remove the overrides.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerIntegrationTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerIntegrationTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
cmnbroad: I take it back. It's always the same for a MultiVariantWalker, because we make it so by ensuring the header contains the merged dictionary. But for a VariantWalker with a single input with no embedded dictionary, but with an accompanying index, the header will not have the dictionary, but FeatureDataSource will return the synthesized one . Maybe we should make FeatureDataSource behave like MultiFeatureDataSource (over the TODO above) ? Also, I’m wondering if it better perf-wise to cache the dictionary once rather than regenerating it every time from the header, given that hg38 apparently has thousands of contigs.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantDataSource.java
```diff
@@ -0,0 +1,372 @@
...
+public final class MultiVariantDataSource implements GATKDataSource<VariantContext>, AutoCloseable {
+    private static final Logger logger = LogManager.getLogger(MultiVariantDataSource.class);
+
+    /**
+     * List of FeatureDataSource objects aggregated by this MultiVariantDataSource
+     */
+    final private List<FeatureDataSource<VariantContext>> featureDataSources = new ArrayList<FeatureDataSource<VariantContext>>();
+
+    /**
+     * Merged VCF header used for this (aggregate) source, derived from the individual soruces.
```
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantDataSourceUnitTest.java
cmnbroad: Assertion added.

PR #2187, opened 2016-09-27
SV spark pipeline bash script

SHuang-Broad:
@cwhelan , would you please review? Thanks!

### DIFF
--- a/scripts/sv_spark_pipeline_launch.sh
+++ b/scripts/sv_spark_pipeline_launch.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+
+if [[ -z ${1+x} ]]; then
+    echo "please provide address for output directory you have write access to"
+    exit 1
+else
+    OUTPUT=$1
+    echo "output directory is set by user to '$1'"
+fi
+
+# STAGE 1 args
+ONE_MEGA_BYTE=1000000
+TEN_MEGA_BYTES=10000000
+EXCLUSIONINTERVALS=hdfs://svdev-1-m:8020/reference/GRCh37.kill.intervals
+KMERSTOIGNORE=hdfs://svdev-1-m:8020/reference/Homo_sapiens_assembly38.k51.dups
+FASTQDIR=$OUTPUT/fastq
+EVIDENCE=$OUTPUT/evidence
+INTERVAL=$OUTPUT/intervals
+KMERINTERVALS=$OUTPUT/kmerIntervals
+QNAMEINTERVALSMAPPED=$OUTPUT/qnameIntervalsMapped
+QNAMEINTERVALSFORASSEMBLY=$OUTPUT/qnameIntervalsForAssembly
+MAXFASTQSIZE=$TEN_MEGA_BYTES
+
+# STAGE 2 args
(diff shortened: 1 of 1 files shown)


scripts/sv_spark_pipeline_launch.sh
```diff
@@ -0,0 +1,103 @@
...
+if [[ -z ${1+x} ]]; then
+    echo "please provide address for output directory you have write access to"
+    exit 1
+else
+    OUTPUT=$1
+    echo "output directory is set by user to '$1'"
+fi
+
+# STAGE 1 args
+ONE_MEGA_BYTE=1000000
```
cwhelan: This param doesn't seem to be used

scripts/sv_spark_pipeline_launch.sh
```diff
@@ -0,0 +1,103 @@
...
+TEN_MEGA_BYTES=10000000
```
cwhelan: Since this is only used for max_fastq_size I'd just plug it in there, no need for this variable.

scripts/sv_spark_pipeline_launch.sh
```diff
@@ -0,0 +1,103 @@
...
+EXCLUSIONINTERVALS=hdfs://svdev-1-m:8020/reference/GRCh37.kill.intervals
```
cwhelan: I think it would be good to make the cluster name a command line parameter and replace all occurrences of "svdev-1-m" with it.

cwhelan: Actually replace "svdev-1" with the cluster name param, and then make another variable with the master node name.

scripts/sv_spark_pipeline_launch.sh
```diff
@@ -0,0 +1,103 @@
...
+./gatk-launch CallVariantsFromAlignedContigsSpark \
+    -R $TWOBITREF \
+    --fastaReference $REFERENCE \
+    --inputAssemblies $ASSEMBLY_SUCC \
+    --inputAlignments $ALIGNMENTS \
+    --outputPath $SITEVCFDIR \
+    -- \
+    $SPARKARGS_LOW
+
+./gatk-launch SingleDiploidSampleBiallelicInversionGenotyperSpark \
```
cwhelan: We should hold off on merging this until the genotyper is also merged (so that the script doesn't break for people not in your branch).

SHuang-Broad: Done

scripts/sv_spark_pipeline_launch.sh
SHuang-Broad: Done

scripts/sv_spark_pipeline_launch.sh
SHuang-Broad: Counting repetitive sequences is difficult for me, that's why I put in these names so if we decide to switch the argument, I don't have to count them again...

SHuang-Broad:
@cwhelan , just addressed the comments and some other small changes.

SHuang-Broad:
closing as may be unnecessary

Issue #1988, opened 2016-07-11
Create Spark partitioner that can efficiently load records for overlapping genomic regions


droazen:
We have a somewhat urgent need for a Spark partitioner (or something similar) that can handle loading of records for regions that overlap. For example, we'd like to be able to have a partition with all reads in `1:900-2100`, and a separate partition with reads in `1:1900-3100`, and some reads in the overlapping regions would appear in both partitions. 

Such a partitioner is needed for several current GATK projects, including:

https://github.com/broadinstitute/gatk/issues/1639 Spark version of HaplotypeCaller (the prototype currently uses an inefficient `groupBy` operation to accomplish this)

https://github.com/broadinstitute/gatk/issues/1558 Make pileups in parallel on Spark 

It also might be needed for avoiding boundary artifacts in the SV caller.

droazen:
@tomwhite Would you be able to work on this as your primary GATK ticket for alpha-3?

tomwhite:
@droazen yes.

> the prototype currently uses an inefficient `groupBy` operation

Can you point me in the direction of the code for this please, @lbergelson?

tomwhite:
Here's the work I've done so far to implement an overlaps partitioner: https://github.com/broadinstitute/gatk/compare/tw_overlap_partitioner.

The idea is to coalesce neighbouring partitions so that the task processing them can see (but not process) reads from neighbouring partitions. E.g. the task processing partition _i_ can see reads from partition _i-1_ and partition _i+1_. This is done using Spark 2's _coalesce_ method that takes a PartitionCoalescer: http://spark.apache.org/docs/latest/api/java/org/apache/spark/rdd/RDD.html#coalesce(int,%20boolean,%20scala.Option,%20scala.math.Ordering). In this PR I backported some of that logic so it works with Spark 1.6.

It turns out that this is not faster than the groupByKey, primarily since it has to scan over the input twice: first to find the extents of the reads in each partition, second to process each (coalesced) partition. It should be possible to avoid the first scan by just reading the first read in each partition and using a sequence dictionary to find the end points of contigs. This is a bit more involved, although it should be faster than groupByKey. (I tried caching the input to see if this helped, but it didn't.)

tomwhite:
[...]

Issue #1558, opened 2016-03-07
Make pileups in parallel on Spark


akiezun:
We need a way to create pileups (alignment contexts) in parallel on Spark. Ie the input is a bam file and a vcf file. The engine should provide a way to create an RDD of `AlignmentContext`.

This functionality is important for tools that work on pileups (eg allelic copy number tools).

droazen:
Turns out this may already have been done -- @samuelklee do you have a branch in protected that implements this?

akiezun:
https://github.com/broadinstitute/gatk-protected/tree/sl_het_coverage_spark ?

samuelklee:
Actually, I think this is it: https://github.com/broadinstitute/gatk/compare/sl_het_pulldown_spark

I did take a crack at Spark-ifying the het pulldown in this (ancient) branch, but I can't claim that I knew what I was doing---if I remember correctly, I just tried to adapt some code Adam had in https://github.com/broadinstitute/gatk/tree/ak_coverage_spark.  I don't think I ever implemented proper CIGAR-string parsing either.  Might be better to start from scratch?

droazen:
Reassigning to @jamesemery

droazen:
Blocked by https://github.com/broadinstitute/gatk/issues/1988

droazen:
This ticket should be fairly simple once https://github.com/broadinstitute/gatk/issues/1988 is implemented.

tomwhite:
Fixed by #2190

Issue #2176, opened 2016-09-19
Question: could it be possible to change the logging to sfl4j?


magicDGS:
As explain in the documentation. for logging GATK is using `org.apache.logging.log4j.Logger`.

Nevertheless, because I would like to use GATK4 as a framework, I think that API users could benefit from using [SLF4J](http://www.slf4j.org/) as a plugin system for allow users to decide which system use.

Because it exists a wrapper for log4j, I think that this could be done easily without changing any behaviour or configuration. In addition, I believe that the usage of SLF4J is similar as the one used in log4j.

magicDGS:
As I commented in #2190 (and @tomwhite agrees), this could help on the serialization problems in Spark. In addition, it will be interesting from the point of view that I described in the original comment: more flexibility of GATK as a framework.

What do you think, @droazen?

droazen:
@lbergelson Opinion on this?

lbergelson:
@droazen I don't really have much of an opinion about the loggers.  I think we chose log4j2 pretty arbitrarily and because it seemed popular.  We don't have very complex logging needs so any logger pretty much satisfied our needs.  

If there's a good reason to switch to SLF4J with a log4j backend that seems fine to me.  It makes sense to use the more general solution so that people can use whatever backend they want.  

I do know that we've encountered a lot of issues with having multiple copies of different logging frameworks included as transitive dependencies on spark.  So there may be some hassle switching over, we'd have to run tests on google cloud to make sure we don't start crashing all of a sudden for logger related classpath issues.  

@magicDGS I don't think anyone here cares that much about which logger we're using as long as it doesn't get in our way.  If you want to do the switchover I think we'd be happy to accept a pull request as long as it didn't cause problems with our spark tools.  Are you able to run our spark tools on gcloud dataproc?

magicDGS:
I tried to do it, and I'm afraid that it won't be trivial as I expected: because it is a facade, there is not accesibility to `Logger.setLogLevel()` and this is required to set the verbosity level in the command line.

After explore a bit the code, it seems that [`LoggingUtils`](https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/utils/LoggingUtils.java) is the only place where a concrete implementation should be used.

[...]

PR #2190, opened 2016-09-28
Pileup on Spark

tomwhite:
This addresses #1558 by using an implementation of #1988.

Running on a 6.6GB BAM file:
- PileupSpark with a shuffle took 4.35 minutes on a 6 node cluster
- PileupSpark with no shuffle (using the overlaps partitioner) took 2.96 minutes (30% faster)
- Pileup took 39.2 minutes (on a single node)

@droazen, please take a look and let me know what you think.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -17,6 +17,7 @@ import org.broadinstitute.hellbender.utils.LoggingUtils;
 import org.broadinstitute.hellbender.utils.Utils;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.PrintStream;
 import java.net.InetAddress;
 import java.text.DecimalFormat;
@@ -44,7 +45,13 @@ import java.util.List;
  *
  */
 public abstract class CommandLineProgram {
-    protected final Logger logger = LogManager.getLogger(this.getClass());
+    protected transient Logger logger = LogManager.getLogger(this.getClass());
+
+    private void readObject(java.io.ObjectInputStream in)
+            throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
+        logger = LogManager.getLogger(this.getClass()); // Logger is not serializable (even by Kryo)
+    }
 
     @Argument(common=true, optional=true)
     public List<File> TMP_DIR = new ArrayList<>();
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSpark.java
@@ -1,15 +1,32 @@
 package org.broadinstitute.hellbender.engine.spark;
 
+import com.google.common.base.Function;
+import com.google.common.collect.Iterators;
+import htsjdk.samtools.SAMSequenceDictionary;
 import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.JavaRDD;
+import org.apache.spark.api.java.JavaSparkContext;
+import org.apache.spark.api.java.function.PairFlatMapFunction;
+import org.apache.spark.broadcast.Broadcast;
 import org.broadinstitute.hellbender.engine.ReadContextData;
+import org.broadinstitute.hellbender.engine.Shard;
+import org.broadinstitute.hellbender.engine.ShardBoundary;
 import org.broadinstitute.hellbender.engine.datasources.ReferenceMultiSource;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.IntervalUtils;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.collections.IntervalsSkipList;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
 import org.broadinstitute.hellbender.utils.reference.ReferenceBases;
 import org.broadinstitute.hellbender.utils.variant.GATKVariant;
 import scala.Tuple2;
+
+import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * AddContextDataToRead pairs reference bases and overlapping variants with each GATKRead in the RDD input.
@@ -24,9 +41,24 @@ import scala.Tuple2;
  * {@link org.broadinstitute.hellbender.engine.datasources.ReferenceWindowFunctions} for examples.
  */
 public class AddContextDataToReadSpark {
+    /**
+     * Add context data ({@link ReadContextData}) to reads.
+     * @param ctx the Spark context
+     * @param reads the coordinate-sorted reads
+     * @param referenceSource the reference source
+     * @param variants the coordinate-sorted variants
+     * @param joinStrategy the strategy to use to join context data to reads
+     * @param sequenceDictionary the sequence dictionary for the reads (only used for OVERLAPS_PARTITIONER join strategy, use null otherwise)
+     * @param shardSize the maximum size of each shard, in bases (only used for OVERLAPS_PARTITIONER join strategy, use 0 otherwise)
+     * @param shardPadding amount of extra context around each shard, in bases (only used for OVERLAPS_PARTITIONER join strategy, use 0 otherwise)
+     * @return a RDD of read-context pairs, in coordinate-sorted order
+     */
     public static JavaPairRDD<GATKRead, ReadContextData> add(
-            final JavaRDD<GATKRead> reads, final ReferenceMultiSource referenceDataflowSource,
-            final JavaRDD<GATKVariant> variants, final JoinStrategy joinStrategy) {
+            final JavaSparkContext ctx,
+            final JavaRDD<GATKRead> reads, final ReferenceMultiSource referenceSource,
+            final JavaRDD<GATKVariant> variants, final JoinStrategy joinStrategy,
+            final SAMSequenceDictionary sequenceDictionary,
+            final int shardSize, final int shardPadding) {
         // TODO: this static method should not be filtering the unmapped reads.  To be addressed in another issue.
         JavaRDD<GATKRead> mappedReads = reads.filter(read -> ReadFilterLibrary.MAPPED.test(read));
         JavaPairRDD<GATKRead, Tuple2<Iterable<GATKVariant>, ReferenceBases>> withVariantsWithRef;
@@ -34,15 +66,75 @@ public class AddContextDataToReadSpark {
             // Join Reads and Variants
             JavaPairRDD<GATKRead, Iterable<GATKVariant>> withVariants = BroadcastJoinReadsWithVariants.join(mappedReads, variants);
             // Join Reads with ReferenceBases
-            withVariantsWithRef = BroadcastJoinReadsWithRefBases.addBases(referenceDataflowSource, withVariants);
+            withVariantsWithRef = BroadcastJoinReadsWithRefBases.addBases(referenceSource, withVariants);
         } else if (joinStrategy.equals(JoinStrategy.SHUFFLE)) {
             // Join Reads and Variants
             JavaPairRDD<GATKRead, Iterable<GATKVariant>> withVariants = ShuffleJoinReadsWithVariants.join(mappedReads, variants);
             // Join Reads with ReferenceBases
-            withVariantsWithRef = ShuffleJoinReadsWithRefBases.addBases(referenceDataflowSource, withVariants);
+            withVariantsWithRef = ShuffleJoinReadsWithRefBases.addBases(referenceSource, withVariants);
+        } else if (joinStrategy.equals(JoinStrategy.OVERLAPS_PARTITIONER)) {
+            return addUsingOverlapsPartitioning(ctx, reads, referenceSource, variants, sequenceDictionary, shardSize, shardPadding);
         } else {
             throw new UserException("Unknown JoinStrategy");
         }
         return withVariantsWithRef.mapToPair(in -> new Tuple2<>(in._1(), new ReadContextData(in._2()._2(), in._2()._1())));
+    }
+
+    /**
+     * Add context data ({@link ReadContextData}) to reads, using overlaps partitioning to avoid a shuffle.
+     * @param ctx the Spark context
+     * @param mappedReads the coordinate-sorted reads
+     * @param referenceSource the reference source
+     * @param variants the coordinate-sorted variants
+     * @param sequenceDictionary the sequence dictionary for the reads
+     * @param shardSize the maximum size of each shard, in bases
+     * @param shardPadding amount of extra context around each shard, in bases
+     * @return a RDD of read-context pairs, in coordinate-sorted order
+     */
+    private static JavaPairRDD<GATKRead, ReadContextData> addUsingOverlapsPartitioning(
+            final JavaSparkContext ctx,
+            final JavaRDD<GATKRead> mappedReads, final ReferenceMultiSource referenceSource,
+            final JavaRDD<GATKVariant> variants, final SAMSequenceDictionary sequenceDictionary,
+            final int shardSize, final int shardPadding) {
+        final List<SimpleInterval> intervals = IntervalUtils.getAllIntervalsForReference(sequenceDictionary);
+        // use unpadded shards (padding is only needed for reference bases)
+        final List<ShardBoundary> intervalShards = intervals.stream()
+                .flatMap(interval -> Shard.divideIntervalIntoShards(interval, shardSize, 0, sequenceDictionary).stream())
+                .collect(Collectors.toList());
+
+        final Broadcast<ReferenceMultiSource> bReferenceSource = ctx.broadcast(referenceSource);
+        final IntervalsSkipList<GATKVariant> variantSkipList = new IntervalsSkipList<>(variants.collect());
+        final Broadcast<IntervalsSkipList<GATKVariant>> variantsBroadcast = ctx.broadcast(variantSkipList);
+
+        int maxLocatableSize = Math.min(shardSize, shardPadding);
+        JavaRDD<Shard<GATKRead>> shardedReads = SparkSharder.shard(ctx, mappedReads, GATKRead.class, sequenceDictionary, intervalShards, maxLocatableSize);
+        return shardedReads.flatMapToPair(new PairFlatMapFunction<Shard<GATKRead>, GATKRead, ReadContextData>() {
+            private static final long serialVersionUID = 1L;
+            @Override
+            public Iterable<Tuple2<GATKRead, ReadContextData>> call(Shard<GATKRead> shard) throws Exception {
+                // get reference bases for this shard (padded)
+                SimpleInterval paddedInterval = shard.getInterval().expandWithinContig(shardPadding, sequenceDictionary);
+                ReferenceBases referenceBases = bReferenceSource.getValue().getReferenceBases(null, paddedInterval);
+                final IntervalsSkipList<GATKVariant> intervalsSkipList = variantsBroadcast.getValue();
+                Iterator<Tuple2<GATKRead, ReadContextData>> transform = Iterators.transform(shard.iterator(), new Function<GATKRead, Tuple2<GATKRead, ReadContextData>>() {
+                    @Nullable
+                    @Override
+                    public Tuple2<GATKRead, ReadContextData> apply(@Nullable GATKRead r) {
+                        List<GATKVariant> overlappingVariants;
+                        if (SimpleInterval.isValid(r.getContig(), r.getStart(), r.getEnd())) {
+                            overlappingVariants = intervalsSkipList.getOverlapping(new SimpleInterval(r));
+                        } else {
+                            //Sometimes we have reads that do not form valid intervals (reads that do not consume any ref bases, eg CIGAR 61S90I
+                            //In those cases, we'll just say that nothing overlaps the read
+                            overlappingVariants = Collections.emptyList();
+                        }
+                        return new Tuple2<>(r, new ReadContextData(referenceBases, overlappingVariants));
+                    }
+                });
+                // only include reads that start in the shard
+                return () -> Iterators.filter(transform, r -> r._1().getStart() >= shard.getStart()
+                        && r._1().getStart() <= shard.getEnd());
+            }
+        });
     }
 }
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/LocusWalkerSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/LocusWalkerSpark.java
@@ -1,0 +1,136 @@
+package org.broadinstitute.hellbender.engine.spark;
+
+import com.google.common.base.Function;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterators;
+import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.SAMReadGroupRecord;
+import htsjdk.samtools.SAMSequenceDictionary;
+import org.apache.spark.api.java.JavaRDD;
+import org.apache.spark.api.java.JavaSparkContext;
+import org.apache.spark.api.java.function.FlatMapFunction;
+import org.apache.spark.broadcast.Broadcast;
+import org.broadinstitute.hellbender.cmdline.Argument;
+import org.broadinstitute.hellbender.engine.*;
+import org.broadinstitute.hellbender.engine.datasources.ReferenceMultiSource;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.IntervalUtils;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.downsampling.DownsamplingMethod;
+import org.broadinstitute.hellbender.utils.iterators.IntervalOverlappingIterator;
+import org.broadinstitute.hellbender.utils.locusiterator.LIBSDownsamplingInfo;
+import org.broadinstitute.hellbender.utils.locusiterator.LocusIteratorByState;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import scala.Tuple3;
+
+import javax.annotation.Nullable;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public abstract class LocusWalkerSpark extends GATKSparkTool {
+    private static final long serialVersionUID = 1L;
+
+    @Argument(fullName = "maxDepthPerSample", shortName = "maxDepthPerSample", doc = "Maximum number of reads to retain per sample per locus. Reads above this threshold will be downsampled. Set to 0 to disable.", optional = true)
+    protected int maxDepthPerSample = defaultMaxDepthPerSample();
+
+    /**
+     * Returns default value for the {@link #maxDepthPerSample} parameter, if none is provided on the command line.
+     * Default implementation returns 0 (no downsampling by default).
+     */
+    protected int defaultMaxDepthPerSample() {
+        return 0;
+    }
+
+    @Argument(fullName="readShardSize", shortName="readShardSize", doc = "Maximum size of each read shard, in bases.", optional = true)
+    public int readShardSize = 10000;
+
+    @Argument(fullName="readShardPadding", shortName="readShardPadding", doc = "Each read shard has this many bases of extra context on each side.", optional = true)
+    public int readShardPadding = 1000;
+
+    @Argument(doc = "whether to use the shuffle implementation or overlaps partitioning (the default)", shortName = "shuffle", fullName = "shuffle", optional = true)
+    public boolean shuffle = false;
+
+    private FeatureManager features; // TODO: move up to GATKSparkTool?
+
+    @Override
+    protected void runPipeline(JavaSparkContext sparkContext) {
+        initializeFeatures();
+        super.runPipeline(sparkContext);
+    }
+
+    void initializeFeatures() {
+        features = new FeatureManager(this);
+        if ( features.isEmpty() ) {  // No available sources of Features discovered for this tool
+            features = null;
+        }
+    }
+
+    /** Returns the downsampling info using {@link #maxDepthPerSample} as target coverage. */
+    protected final LIBSDownsamplingInfo getDownsamplingInfo() {
+        if (maxDepthPerSample < 0) {
+            throw new UserException.BadArgumentValue("maxDepthPerSample",
+                    String.valueOf(maxDepthPerSample),
+                    "should be a positive number");
+        }
+        return (maxDepthPerSample == 0) ? LocusIteratorByState.NO_DOWNSAMPLING : new LIBSDownsamplingInfo(true, maxDepthPerSample);
+    }
+
+    /**
+     * Loads alignments and the corresponding reference and features into a {@link JavaRDD} for the intervals specified.
+     *
+     * If no intervals were specified, returns all the alignments.
+     *
+     * @return all alignments from as a {@link JavaRDD}, bounded by intervals if specified.
+     */
+    public JavaRDD<Tuple3<AlignmentContext, ReferenceContext, FeatureContext>> getAlignments(JavaSparkContext ctx) {
+        SAMSequenceDictionary sequenceDictionary = getBestAvailableSequenceDictionary();
+        List<SimpleInterval> intervals = hasIntervals() ? getIntervals() : IntervalUtils.getAllIntervalsForReference(sequenceDictionary);
+        final List<ShardBoundary> intervalShards = intervals.stream()
+                .flatMap(interval -> Shard.divideIntervalIntoShards(interval, readShardSize, readShardPadding, sequenceDictionary).stream())
+                .collect(Collectors.toList());
+        int maxLocatableSize = Math.min(readShardSize, readShardPadding);
+        JavaRDD<Shard<GATKRead>> shardedReads = SparkSharder.shard(ctx, getReads(), GATKRead.class, sequenceDictionary, intervalShards, maxLocatableSize, shuffle);
+        Broadcast<ReferenceMultiSource> bReferenceSource = hasReference() ? ctx.broadcast(getReference()) : null;
+        Broadcast<FeatureManager> bFeatureManager = features == null ? null : ctx.broadcast(features);
+        return shardedReads.flatMap(getAlignmentsFunction(bReferenceSource, bFeatureManager, sequenceDictionary, getHeaderForReads(), getDownsamplingInfo()));
+    }
+
+    /**
+     * Return a function that maps a {@link Shard} of reads into a tuple of alignments and their corresponding reference and features.
+     * @param bReferenceSource the reference source broadcast
+     * @param bFeatureManager the feature manager broadcast
+     * @param sequenceDictionary the sequence dictionary for the reads
+     * @param header the reads header
+     * @param downsamplingInfo the downsampling method for the reads
+     * @return a function that maps a {@link Shard} of reads into a tuple of alignments and their corresponding reference and features.
+     */
+    private static FlatMapFunction<Shard<GATKRead>, Tuple3<AlignmentContext, ReferenceContext, FeatureContext>> getAlignmentsFunction(
+            Broadcast<ReferenceMultiSource> bReferenceSource, Broadcast<FeatureManager> bFeatureManager,
+            SAMSequenceDictionary sequenceDictionary, SAMFileHeader header, LIBSDownsamplingInfo downsamplingInfo) {
+        return (FlatMapFunction<Shard<GATKRead>, Tuple3<AlignmentContext, ReferenceContext, FeatureContext>>) shardedRead -> {
+            SimpleInterval interval = shardedRead.getInterval();
+            SimpleInterval paddedInterval = shardedRead.getPaddedInterval();
+            Iterator<GATKRead> readIterator = shardedRead.iterator();
+            ReferenceDataSource reference = bReferenceSource == null ? null :
+                    new ReferenceMemorySource(bReferenceSource.getValue().getReferenceBases(null, paddedInterval), sequenceDictionary);
+            FeatureManager fm = bFeatureManager == null ? null : bFeatureManager.getValue();
+
+            final Set<String> samples = header.getReadGroups().stream()
+                    .map(SAMReadGroupRecord::getSample)
+                    .collect(Collectors.toSet());
+            LocusIteratorByState libs = new LocusIteratorByState(readIterator, downsamplingInfo, false, samples, header, true, false);
+            IntervalOverlappingIterator<AlignmentContext> alignmentContexts = new IntervalOverlappingIterator<>(libs, ImmutableList.of(interval), sequenceDictionary);
+            Iterator<Tuple3<AlignmentContext, ReferenceContext, FeatureContext>> transform = Iterators.transform(alignmentContexts, new Function<AlignmentContext, Tuple3<AlignmentContext, ReferenceContext, FeatureContext>>() {
+                @Nullable
+                @Override
+                public Tuple3<AlignmentContext, ReferenceContext, FeatureContext> apply(@Nullable AlignmentContext alignmentContext) {
+                    final SimpleInterval alignmentInterval = new SimpleInterval(alignmentContext);
+                    return new Tuple3<>(alignmentContext, new ReferenceContext(reference, alignmentInterval), new FeatureContext(fm, alignmentInterval));
+                }
+            });
+            return () -> transform;
+        };
+    }
+}
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/RangePartitionCoalescer.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/RangePartitionCoalescer.java
@@ -1,0 +1,52 @@
+package org.broadinstitute.hellbender.engine.spark;
+
+import org.apache.spark.Partition;
+import org.apache.spark.rdd.PartitionGroup;
+import org.apache.spark.rdd.RDD;
+import scala.collection.JavaConversions;
+import scala.collection.Seq;
+
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * A {@link PartitionCoalescer} that allows a range of partitions to be coalesced into groups.
+ */
+class RangePartitionCoalescer implements PartitionCoalescer, Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private List<Integer> maxEndPartitionIndexes;
+
+    /**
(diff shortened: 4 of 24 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/PileupSpark.java
```diff
@@ -0,0 +1,152 @@
...
+            // get the samples from the read groups
+            final Set<String> samples = header.getReadGroups().stream()
+                    .map(SAMReadGroupRecord::getSample)
+                    .collect(Collectors.toSet());
+            LocusIteratorByState libs = new LocusIteratorByState(readIterator, DownsamplingMethod.NONE, true, false, false, samples, header);
+            return new IntervalOverlappingIterator<>(libs, ImmutableList.copyOf(intervalIterator), header.getSequenceDictionary());
+        }
+    }
+
+    private static class ShufflePileupFunction implements FlatMapFunction<Tuple2<SimpleInterval, Iterable<GATKRead>>, AlignmentContext> {
```
magicDGS: I think that this could be useful outside the `PileupSpark` tool to create `JavaRDD<AlignmentContext>` easily. I'm working with pileups and this will be very useful for me...

tomwhite: We could expose a method that returns an `JavaRDD<AlignmentContext>`. Also, do you need the reference bases for each context too? That needs a bit more work.

tomwhite:
I've made some improvements to this PR, including:

* Made it easier to use the `joinOverlapping` method by making the function you supply only have to worry about one interval (shard) at a time. This simplifies the callers code, so PileupSpark (for example) is now shorter.
* Added some documentation.

I've also used the same technique to improve `AddContextDataToReadSpark` so that references are filled in on a per shard basis, rather than per read. In tests on a 6.6GB file I managed to get BaseRecalibratorSpark's runtime down from 10.61 minutes to 3.73 minutes, which is over 60% faster.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/PileupSpark.java
magicDGS: Thanks a lot for the answer @tomwhite!

Actually it would be nice to have a way of using an implemented engine for a `LocusWalker` work with Spark. Thus, I think that this requires to have the reference bases for each context. But I don't know if this is out of the scope of this PR or if it is doable here. I would appreciate a lot if it could be done :)

tomwhite:
I've improved PileupSpark so that it more closely mirrors the walker version. In particular, I've created a `LocusWalkerSpark` class that has a (shuffleless) method to obtain a

`JavaRDD<Tuple3<AlignmentContext, ReferenceContext, FeatureContext>>`

to mirror the

`apply(AlignmentContext, ReferenceContext, FeatureContext)`

method on `LocusWalker`.

It should be possible to use a similar approach to make Spark equivalents of other walkers, which will make porting the walkers to Spark easier.

What do you think @magicDGS, @droazen?

magicDGS:
Thanks for this, @tomwhite! Actually this is awesome for me, because I don't know too much about the Spark framework and it makes more easy the implementation of tools compatible with it.

In addition, I think that it will be awesome to have this for other type of walkers. Thanks a lot again!

tomwhite:
BTW this approach works for other walkers as well. I've converted ReadWalker and AssemblyRegionWalker to their Spark versions in this branch (which I'll submit after this goes in): https://github.com/broadinstitute/gatk/compare/tw_spark_walkers.

hammer:
cc @ryan-williams who has done a tremendous amount of work on handling skew when building pileups for https://github.com/hammerlab/guacamole

src/test/java/org/broadinstitute/hellbender/utils/spark/SparkUtilsShardUnitTest.java
```diff
@@ -0,0 +1,355 @@
...
+                new SimpleInterval("1", 2, 4),
+                new SimpleInterval("1", 8, 12),
+                new SimpleInterval("1", 11, 22));
+
+        List<ShardBoundary> shardBoundaries = intervals.stream().map(si -> new ShardBoundary(si, si)).collect(Collectors.toList());
+
+        JavaPairRDD<Locatable, Integer> readsPerInterval =
+                SparkUtils.shard(ctx, reads, TestRead.class, sequenceDictionary, shardBoundaries, false)
+                .flatMapToPair(new CountOverlappingReadsFunction());
+        assertEquals(readsPerInterval.collectAsMap(), ImmutableMap.of(intervals.get(0), 1, intervals.get(1), 7, intervals.get(2), 4));
```
droazen: Declare the Map of expected values separately above, to make it clearer here that you're asserting equality against the same values in both the shuffle and non-shuffle cases. Also applies to the tests below.

src/test/java/org/broadinstitute/hellbender/utils/spark/SparkUtilsShardUnitTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+        assertTrue(it.hasNext());
+        next = it.next();
+        assertEquals(next._1(), intervals.get(2));
+        assertEquals(next._2(), ImmutableList.of(reads.get(7), reads.get(8), reads.get(9), reads.get(10)));
+
+        assertFalse(it.hasNext());
+    }
+
+    static class TestRead implements Locatable {
```
droazen: Make private

src/test/java/org/broadinstitute/hellbender/utils/spark/SparkUtilsShardUnitTest.java
```diff
@@ -0,0 +1,355 @@
...
+        assertEquals(next._1(), intervals.get(1));
+        assertEquals(next._2(), ImmutableList.of(reads.get(2), reads.get(3), reads.get(4), reads.get(5), reads.get(6), reads.get(7), reads.get(8)));
+
+        assertTrue(it.hasNext());
+        next = it.next();
+        assertEquals(next._1(), intervals.get(2));
+        assertEquals(next._2(), ImmutableList.of(reads.get(7), reads.get(8), reads.get(9), reads.get(10)));
+
+        assertFalse(it.hasNext());
+    }
```
droazen: Existing test cases are good, but I think we might also want a slightly larger-scale test that uses more than just a small handful of reads.

src/main/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSpark.java
```diff
@@ -45,4 +63,53 @@
         }
         return withVariantsWithRef.mapToPair(in -> new Tuple2<>(in._1(), new ReadContextData(in._2()._2(), in._2()._1())));
     }
+
+    private static final int SHARD_SIZE = 10000;
+    private static final int SHARD_PADDING = 1000;
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSpark.java
```diff
@@ -45,4 +63,53 @@
...
+
+    public static JavaPairRDD<GATKRead, ReadContextData> add(
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSpark.java
```diff
@@ -45,4 +63,53 @@
...
+            final JavaSparkContext ctx,
+            final JavaRDD<GATKRead> reads, final ReferenceMultiSource referenceSource,
+            final JavaRDD<GATKVariant> variants, final SAMSequenceDictionary sequenceDictionary) {
+        JavaRDD<GATKRead> mappedReads = reads.filter(read -> ReadFilterLibrary.MAPPED.test(read));
+
+        final List<SimpleInterval> intervals = IntervalUtils.getAllIntervalsForReference(sequenceDictionary);
+        // use unpadded shards (padding is only needed for reference bases)
+        final List<ShardBoundary> intervalShards = intervals.stream()
+                .flatMap(interval -> Shard.divideIntervalIntoShards(interval, SHARD_SIZE, SHARD_SIZE, 0, sequenceDictionary).stream())
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSpark.java
```diff
@@ -45,4 +63,53 @@
...
+                        } else {
+                            //Sometimes we have reads that do not form valid intervals (reads that do not consume any ref bases, eg CIGAR 61S90I
+                            //In those cases, we'll just say that nothing overlaps the read
+                            overlappingVariants = Collections.emptyList();
+                        }
+                        return new Tuple2<>(r, new ReadContextData(referenceBases, overlappingVariants));
+                    }
+                });
+                // only include reads that start in the shard
+                return () -> Iterators.filter(transform, r -> r._1().getStart() >= shard.getStart());
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/PileupSpark.java
```diff
@@ -0,0 +1,138 @@
+package org.broadinstitute.hellbender.tools.spark.pipelines;
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/PileupSpark.java
```diff
@@ -0,0 +1,138 @@
...
+import org.broadinstitute.hellbender.utils.pileup.PileupElement;
+import org.broadinstitute.hellbender.utils.pileup.ReadPileup;
+import org.broadinstitute.hellbender.utils.spark.LocusWalkerSpark;
+import scala.Tuple3;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@CommandLineProgramProperties(summary = "Print read alignments in Pileup-style format on Spark",
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/PileupSpark.java
```diff
@@ -0,0 +1,138 @@
...
+        programGroup = SparkProgramGroup.class)
+public final class PileupSpark extends LocusWalkerSpark {
+    private static final long serialVersionUID = 1L;
+
+    private static final String VERBOSE_DELIMITER = "@"; // it's ugly to use "@" but it's literally the only usable character not allowed in read names
+
+    @Override
+    public boolean requiresReads() { return true; }
+
+    @Argument(doc="the output file path", shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME, optional = false)
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/PileupSpark.java
```diff
@@ -0,0 +1,138 @@
...
+    protected String outputFile;
```
tomwhite: Yes. Updated the doc.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/PileupSpark.java
```diff
@@ -0,0 +1,138 @@
...
+    @Argument(fullName = "outputInsertLength", shortName = "outputInsertLength", doc = "Output insert length", optional = true)
+    public boolean outputInsertLength = false;
+
+    @Override
+    public ReadFilter makeReadFilter(){
+        return new WellformedReadFilter(getHeaderForReads())
+                .and(ReadFilterLibrary.MAPPED)
+                .and(ReadFilterLibrary.NOT_DUPLICATE)
+                .and(ReadFilterLibrary.PASSES_VENDOR_QUALITY_CHECK)
+                .and(ReadFilterLibrary.PRIMARY_ALIGNMENT);
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/PileupSpark.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Override
+    protected void runTool(final JavaSparkContext ctx) {
+        getAlignments(ctx).map(pileupFunction(metadata, outputInsertLength, showVerbose)).saveAsTextFile(outputFile);
```
tomwhite: This would indeed be nice, but it means that the function would capture the class instance variables in its closure, which means either that unnecessary data is transferred, or it would fail if there are objects that are not serializable (even if they are not used by the function).

src/main/java/org/broadinstitute/hellbender/utils/spark/LocusWalkerSpark.java
```diff
@@ -0,0 +1,102 @@
+package org.broadinstitute.hellbender.utils.spark;
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/LocusWalkerSpark.java
```diff
@@ -0,0 +1,102 @@
...
+
+import autovalue.shaded.com.google.common.common.base.Function;
+import autovalue.shaded.com.google.common.common.collect.Iterators;
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/LocusWalkerSpark.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public abstract class LocusWalkerSpark extends GATKSparkTool {
+    private static final long serialVersionUID = 1L;
+
+    private static final int SHARD_SIZE = 10000;
+    private static final int SHARD_PADDING = 1000;
```
tomwhite: Done. Something to look at during the performance runs.

src/main/java/org/broadinstitute/hellbender/utils/spark/LocusWalkerSpark.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Argument(doc = "whether to use the shuffle implementation or not", shortName = "shuffle", fullName = "shuffle", optional = true)
+    public boolean shuffle = false;
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/LocusWalkerSpark.java
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * If no intervals were specified, returns all the alignments.
+     *
+     * @return all alignments from as a {@link JavaRDD}, bounded by intervals if specified.
+     */
+    public JavaRDD<Tuple3<AlignmentContext, ReferenceContext, FeatureContext>> getAlignments(JavaSparkContext ctx) {
+        SAMSequenceDictionary sequenceDictionary = getBestAvailableSequenceDictionary();
+        List<SimpleInterval> intervals = hasIntervals() ? getIntervals() : IntervalUtils.getAllIntervalsForReference(sequenceDictionary);
+        final List<ShardBoundary> intervalShards = intervals.stream()
+                .flatMap(interval -> Shard.divideIntervalIntoShards(interval, SHARD_SIZE, SHARD_SIZE, SHARD_PADDING, sequenceDictionary).stream())
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/LocusWalkerSpark.java
```diff
@@ -0,0 +1,102 @@
...
+                .collect(Collectors.toList());
+        JavaRDD<Shard<GATKRead>> shardedReads = SparkUtils.shard(ctx, getReads(), GATKRead.class, sequenceDictionary, intervalShards, shuffle);
+        Broadcast<ReferenceMultiSource> bReferenceSource = hasReference() ? ctx.broadcast(getReference()) : null;
+        Broadcast<FeatureManager> bFeatureManager = features == null ? null : ctx.broadcast(features);
+        return shardedReads.flatMap(getAlignmentsFunction(bReferenceSource, bFeatureManager, sequenceDictionary, getHeaderForReads()));
+    }
+
+    private static FlatMapFunction<Shard<GATKRead>, Tuple3<AlignmentContext, ReferenceContext, FeatureContext>> getAlignmentsFunction(
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/LocusWalkerSpark.java
```diff
@@ -0,0 +1,102 @@
...
+            SimpleInterval paddedInterval = shardedRead.getPaddedInterval();
+            Iterator<GATKRead> readIterator = shardedRead.iterator();
+            ReferenceDataSource reference = bReferenceSource == null ? null :
+                    new ReferenceMemorySource(bReferenceSource.getValue().getReferenceBases(null, paddedInterval), sequenceDictionary);
+            FeatureManager fm = bFeatureManager == null ? null : bFeatureManager.getValue();
+
+            final Set<String> samples = header.getReadGroups().stream()
+                    .map(SAMReadGroupRecord::getSample)
+                    .collect(Collectors.toSet());
+            LocusIteratorByState libs = new LocusIteratorByState(readIterator, DownsamplingMethod.NONE, false, false, false, samples, header);
```
tomwhite: Done. Added a method to allow the downsampling method to be specified.

src/main/java/org/broadinstitute/hellbender/utils/spark/CoalescedRDD.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.spark.TaskContext;
+import org.apache.spark.rdd.PartitionGroup;
+import org.apache.spark.rdd.RDD;
+import scala.collection.Iterator;
+import scala.collection.JavaConversions;
+import scala.collection.Seq;
+import scala.reflect.ClassTag;
+import scala.reflect.ClassTag$;
+
+class CoalescedRDD<T> extends RDD<T> {
```
tomwhite: This a Spark 2 backport. I've added a note.

src/main/java/org/broadinstitute/hellbender/utils/spark/CoalescedRDDPartition.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.spark.rdd.RDD;
+import scala.Option;
+
+import java.io.IOException;
+import java.io.ObjectOutputStream;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+class CoalescedRDDPartition implements Partition {
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/PartitionCoalescer.java
```diff
@@ -0,0 +1,8 @@
+package org.broadinstitute.hellbender.utils.spark;
+
+import org.apache.spark.rdd.PartitionGroup;
+import org.apache.spark.rdd.RDD;
+
+interface PartitionCoalescer {
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/RangePartitionCoalescer.java
```diff
@@ -0,0 +1,44 @@
...
+    private List<Integer> maxEndPartitionIndexes;
+
+    public RangePartitionCoalescer(List<Integer> maxEndPartitionIndexes) {
+        this.maxEndPartitionIndexes = maxEndPartitionIndexes;
+    }
+
+    @Override
+    public PartitionGroup[] coalesce(int maxPartitions, RDD<?> parent) {
+        if (maxPartitions != parent.getNumPartitions()) {
+            throw new IllegalArgumentException("Cannot use " + getClass().getSimpleName() +
```
tomwhite: It's a limitation of the PartitionCoalescer which we are duplicating from Spark. RangePartitionCoalescer is more restrictive, so we have a check to ensure it doesn't change the number of partitions.

src/main/java/org/broadinstitute/hellbender/utils/spark/SparkUtils.java
```diff
@@ -103,4 +123,358 @@ public static boolean pathExists(final JavaSparkContext ctx, final Path targetPa
...
+     * Each shard contains the {@link Locatable} objects that overlap it.
+     * @param ctx the Spark Context
+     * @param locatables the RDD of {@link Locatable}, must be coordinate sorted
+     * @param locatableClass the class of the {@link Locatable} objects in the RDD
+     * @param sequenceDictionary the sequence dictionary to use to find contig lengths
+     * @param intervals the {@link ShardBoundary} objects to create shards for, must be coordinate sorted
+     * @param <L> the {@link Locatable} type
+     * @return an RDD of {@link Shard} of overlapping {@link Locatable} objects
+     */
+    public static <L extends Locatable> JavaRDD<Shard<L>> shard(JavaSparkContext ctx, JavaRDD<L> locatables, Class<L> locatableClass,
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/SparkUtils.java
```diff
@@ -103,4 +123,358 @@ public static boolean pathExists(final JavaSparkContext ctx, final Path targetPa
...
+            private static final long serialVersionUID = 1L;
+            @Override
+            public Shard<L> call(Tuple2<ShardBoundary, Iterable<L>> value) throws Exception {
+                return new ShardBoundaryShard(value._1(), value._2());
+            }
+        });
+    }
+
+    /**
+     * Join an RDD of reads with a set of intervals, and apply a function to process the reads that overlap each interval.
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/SparkUtils.java
```diff
@@ -103,4 +123,358 @@ public static boolean pathExists(final JavaSparkContext ctx, final Path targetPa
...
+     * @param readClass the class of the reads, must be a subclass of {@link Locatable}
+     * @param sequenceDictionary the sequence dictionary to use to find contig lengths
+     * @param intervals the collection of intervals to apply the function to
+     * @param f the function to process intervals and overlapping reads with
+     * @param <R> the read type
+     * @param <I> the interval type
+     * @param <T> the return type of <code>f</code>
+     * @return
+     */
+    private static <R extends Locatable, I extends Locatable, T> JavaRDD<T> joinOverlapping(JavaSparkContext ctx, JavaRDD<R> reads, Class<R> readClass,
```
tomwhite: Done

src/test/java/org/broadinstitute/hellbender/tools/spark/PileupSparkIntegrationTest.java
```diff
@@ -0,0 +1,107 @@
...
+        args.add("--output");
+        args.add(out.getAbsolutePath());
+        args.add("--reference");
+        args.add(b37_reference_20_21);
+        args.add("-L 20:9999900-10000000");
+        if (useShuffle) {
+            args.add("--shuffle");
+        }
+        this.runCommandLine(args.getArgsArray());
+        File expected = new File(TEST_DATA_DIR, "expectedSimplePileup.txt");
```
tomwhite: Yes

src/test/java/org/broadinstitute/hellbender/utils/spark/SparkUtilsShardUnitTest.java
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/engine/FeatureManager.java
```diff
@@ -80,7 +80,7 @@
      * The tool instance containing the FeatureInput argument values that will form the basis of our
      * pool of FeatureDataSources
      */
-    private final CommandLineProgram toolInstance;
+    private transient final CommandLineProgram toolInstance;
```
tomwhite: Possibly. The tests work with this change (this includes tests that load features that aren't in this branch), so it's useful for some cases.

src/main/java/org/broadinstitute/hellbender/utils/spark/SparkUtils.java
```diff
@@ -103,4 +123,358 @@ public static boolean pathExists(final JavaSparkContext ctx, final Path targetPa
...
+     * Create an RDD of {@link Shard} from an RDD of coordinate sorted {@link Locatable}, optionally using a shuffle.
+     * A shuffle is typically only needed for correctness testing, since it usually has a significant performance impact.
+     * @param ctx the Spark Context
+     * @param locatables the RDD of {@link Locatable}, must be coordinate sorted
+     * @param locatableClass the class of the {@link Locatable} objects in the RDD
+     * @param sequenceDictionary the sequence dictionary to use to find contig lengths
+     * @param intervals the {@link ShardBoundary} objects to create shards for, must be coordinate sorted
+     * @param useShuffle whether to use a shuffle or not
+     * @param <L> the {@link Locatable} type
+     * @return an RDD of {@link Shard} of overlapping {@link Locatable} objects
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/SparkUtils.java
```diff
@@ -103,4 +123,358 @@ public static boolean pathExists(final JavaSparkContext ctx, final Path targetPa
...
+     */
+    public static <L extends Locatable> JavaRDD<Shard<L>> shard(JavaSparkContext ctx, JavaRDD<L> locatables, Class<L> locatableClass,
+                                                                SAMSequenceDictionary sequenceDictionary, List<ShardBoundary> intervals, boolean useShuffle) {
+
+        class ShardBoundaryShard implements Shard<L> {
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/SparkUtils.java
```diff
@@ -103,4 +123,358 @@ public static boolean pathExists(final JavaSparkContext ctx, final Path targetPa
...
+                if (currentShard == null) {
+                    return endOfData();
+                }
+                while (peekingReads.hasNext()) {
+                    if (toRightOf(currentShard, peekingReads.peek(), sequenceDictionary)) {
+                        break;
+                    }
+                    R read = peekingReads.next();
+                    if (overlaps(currentShard, read)) {
+                        currentReads.add(read);
```
tomwhite: Yes, it might be. There should be a way to downsample - even skip over partitions that have the same reads in them - but this could be tackled as a follow-on piece of work.

src/main/java/org/broadinstitute/hellbender/utils/spark/SparkUtils.java
```diff
@@ -103,4 +123,358 @@ public static boolean pathExists(final JavaSparkContext ctx, final Path targetPa
         }
     }
 
+    /**
+     * Create an RDD of {@link Shard} from an RDD of coordinate sorted {@link Locatable} <i>without using a shuffle</i>.
+     * Each shard contains the {@link Locatable} objects that overlap it.
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/SparkUtils.java
```diff
@@ -103,4 +123,358 @@ public static boolean pathExists(final JavaSparkContext ctx, final Path targetPa
...
+            public SimpleInterval getPaddedInterval() {
+                return shardBoundary.getPaddedInterval();
+            }
+
+            @Override
+            public Iterator<L> iterator() {
+                return locatables.iterator();
+            }
+        }
+        List<ShardBoundary> paddedIntervals = intervals.stream().map(sb -> new ShardBoundary(sb.getInterval(), sb.getPaddedInterval(), true)).collect(Collectors.toList());
```
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/utils/spark/SparkUtils.java
```diff
@@ -103,4 +123,358 @@ public static boolean pathExists(final JavaSparkContext ctx, final Path targetPa
...
+     * the reads that overlap it. Intervals with no overlapping reads are dropped.
+     */
+    static <R extends Locatable, I extends Locatable> Iterator<Tuple2<I, Iterable<R>>> readsPerShard(Iterator<R> reads, Iterator<I> shards, SAMSequenceDictionary sequenceDictionary) {
+        if (!shards.hasNext()) {
+            return Collections.emptyIterator();
+        }
+        PeekingIterator<R> peekingReads = Iterators.peekingIterator(reads);
+        PeekingIterator<I> peekingShards = Iterators.peekingIterator(shards);
+        Iterator<Tuple2<I, Iterable<R>>> iterator = new AbstractIterator<Tuple2<I, Iterable<R>>>() {
+            // keep track of current and next, since reads can overlap two shards
```
tomwhite: We're assuming that shards are larger than reads (and padding is too). Since shard and padding size is configurable, this shouldn't be a problem. I'll add a fail-fast check for this constraint, as we discussed.

src/test/java/org/broadinstitute/hellbender/utils/spark/SparkUtilsShardUnitTest.java
```diff
@@ -0,0 +1,355 @@
...
+        //               [---------------]
+        //                                   [---------------------]
+        // ---------------------------------------------------------
+        //
+        //     [-----]                                                 <- intervals
+        //                 [---------]
+        //                       [-----------------------]
+        //
+        //                      1                   2
+        //    1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7
```
tomwhite: Done

src/test/java/org/broadinstitute/hellbender/utils/spark/SparkUtilsShardUnitTest.java
tomwhite: The logic is heavily exercised by the integration tests for the various tools that use this feature (there are more to come after this PR), so the small-scale unit tests that cover certain edge cases, plus the integration tests, should be sufficient.

src/main/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSpark.java
```diff
@@ -45,4 +63,53 @@
...
+                    public Tuple2<GATKRead, ReadContextData> apply(@Nullable GATKRead r) {
+                        List<GATKVariant> overlappingVariants;
+                        if (SimpleInterval.isValid(r.getContig(), r.getStart(), r.getEnd())) {
+                            overlappingVariants = intervalsSkipList.getOverlapping(new SimpleInterval(r));
+                        } else {
+                            //Sometimes we have reads that do not form valid intervals (reads that do not consume any ref bases, eg CIGAR 61S90I
+                            //In those cases, we'll just say that nothing overlaps the read
+                            overlappingVariants = Collections.emptyList();
+                        }
+                        return new Tuple2<>(r, new ReadContextData(referenceBases, overlappingVariants));
```
tomwhite: No, it's not used, but the padding will likely give enough context and is a more efficient way of doing this. Address as a follow on?

src/main/java/org/broadinstitute/hellbender/engine/FeatureManager.java
tomwhite: I looked into this more, and it seems that the problem was that Spark was trying to serialize some log4j classes, which are not serializable, and even Kryo can't serialize them. I think there's a fix that makes the logger fields transient and then reinststes them via `readObject` after deserialization. I'll add this to the next PR.

tomwhite:
@droazen thanks for the review. I have now addressed all of your feedback, with the main changes being

* Revert the change to ShardBoundary that uses the padded interval as its interval, and use an anonymous class instead.
* Add a new join strategy (OVERLAPS_PARTITIONER) so that running BQSR can be done using the old way still.
* Add a check for overly long read sizes. If exceeded the job will fail with an exception.
* Be more conservative about the partition end point by using the maximum read length, rather than just the length of the read that happens the start the next partition.
* Rather than making a field in FeatureManager transient, do a better job of reinstating the field that is not serializable (a logger).

src/main/java/org/broadinstitute/hellbender/utils/recalibration/RecalibrationArgumentCollection.java
```diff
@@ -108,6 +108,12 @@
     @Argument(fullName = "computeIndelBQSRTables", shortName = "indelBQSR", doc = "compute indel BQSR tables")
     public boolean computeIndelBQSRTables = false;
 
+    @Argument(fullName="readShardSize", shortName="readShardSize", doc = "Maximum size of each read shard, in bases.", optional = true)
+    public int readShardSize = 10000;
+
+    @Argument(fullName="readShardPadding", shortName="readShardPadding", doc = "Each read shard has this many bases of extra context on each side.", optional = true)
+    public int readShardPadding = 1000;
+
```
droazen: These should not be in `RecalibrationArgumentCollection`, since they are specific to Spark (and to one particular Spark sub-implementation). Can you either move to `BaseRecalibratorSpark`/`BQSRPipelineSpark` or, if you'd prefer to avoid repetition, to your own `OverlapsPartitionerArgumentCollection`/`SparkSharderArgumentCollection` included by BQSR and its various pipelines? 

Also adjust the argument docs to make it clear that they only apply when using the `OVERLAPS_PARTITIONER` join strategy.

src/test/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSparkUnitTest.java
```diff
@@ -34,7 +36,9 @@
     @DataProvider(name = "bases")
     public Object[][] bases() {
         List<Class<?>> classes = Arrays.asList(Read.class, SAMRecord.class);
-        JoinStrategy[] strategies = JoinStrategy.values();
+        EnumSet<JoinStrategy> joinStrategies = EnumSet.allOf(JoinStrategy.class);
+        joinStrategies.remove(JoinStrategy.OVERLAPS_PARTITIONER); // exclude overlaps partitioner, since it needs sequence dictionary not available for mock source
```
droazen: Can you extend the mock so that `ReferenceMultiSource.getReferenceSequenceDictionary()` returns a value, so that we can have test coverage of `JoinStrategy.OVERLAPS_PARTITIONER` in this class? Eg., 

        when(mockSource.getReferenceSequenceDictionary(null)).thenReturn(myDictionary);

src/test/java/org/broadinstitute/hellbender/engine/spark/SparkSharderUnitTest.java
```diff
@@ -0,0 +1,393 @@
...
+        // ---------------------------------------------------------
+        // Reads in partition 3
+        //                                   [-----]
+        //                                           [-----]
+        //                                                   [-----]
+        // ---------------------------------------------------------
+        // Intervals
+        //     [-----]
+        //                 [---------]
+        //                       [-----------------------]
```
droazen: What happened to the partition read extents in this diagram (could have sworn they were here when you presented last week)?

src/main/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSpark.java
```diff
@@ -25,24 +42,87 @@
  */
 public class AddContextDataToReadSpark {
     public static JavaPairRDD<GATKRead, ReadContextData> add(
-            final JavaRDD<GATKRead> reads, final ReferenceMultiSource referenceDataflowSource,
-            final JavaRDD<GATKVariant> variants, final JoinStrategy joinStrategy) {
+            final JavaSparkContext ctx,
+            final JavaRDD<GATKRead> reads, final ReferenceMultiSource referenceSource,
+            final JavaRDD<GATKVariant> variants, final JoinStrategy joinStrategy,
+            final SAMSequenceDictionary sequenceDictionary,
+            final int shardSize, final int shardPadding) {
```
droazen: Document the new `shardSize` and `shardPadding` args, make it clear that they apply only when `JoinStrategy.OVERLAPS_PARTITIONER` is specified, and suggest values to be passed in for these args when other JoinStrategies are being used (0?).

src/main/java/org/broadinstitute/hellbender/engine/spark/LocusWalkerSpark.java
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Maximum size of each read shard, in bases.
+     */
+    private static final int SHARD_SIZE = 10000;
+
+    /**
+     * Each read shard has this many bases of extra context on each side.
+     */
+    private static final int SHARD_PADDING = 1000;
```
droazen: Can you expose these as arguments in `LocusWalkerSpark`, like you did for BQSR?

src/main/java/org/broadinstitute/hellbender/engine/spark/RangePartitionCoalescer.java
```diff
@@ -0,0 +1,52 @@
...
+import scala.collection.Seq;
+
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * A {@link PartitionCoalescer} that allows a range of partitions to be coalesced into groups.
+ */
+class RangePartitionCoalescer implements PartitionCoalescer, Serializable {
```
droazen: Simple unit test for this class? (unless it's a copy of code from Spark2)

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkSharder.java
```diff
@@ -0,0 +1,396 @@
...
+            @Override
+            protected Tuple2<I, Iterable<R>> computeNext() {
+                if (currentShard == null) {
+                    return endOfData();
+                }
+                while (peekingLocatables.hasNext()) {
+                    if (toRightOf(currentShard, peekingLocatables.peek(), sequenceDictionary)) {
+                        break;
+                    }
+                    R read = peekingLocatables.next();
```
droazen: Still some code in here and below that uses the term "read" -- should generalize (eg., `currentLocatable`)

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkSharder.java
```diff
@@ -0,0 +1,396 @@
...
+                    if (read.getContig() != null) {
+                        int size = read.getEnd() - read.getStart() + 1;
+                        if (size > maxLocatableLength) {
+                            throw new UserException(String.format("Max size of locatable exceeded. Max size is %s, but locatable size is %s. Try increasing size and/or padding. Locatable: %s", maxLocatableLength, size, read));
```
droazen: `size and/or padding` -> `shard size and/or padding`

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkSharder.java
```diff
@@ -0,0 +1,396 @@
...
+            final int nextContigIndex;
+            if (i < splitPoints.size() - 1) {
+                next = splitPoints.get(i + 1);
+                nextContigIndex = sequenceDictionary.getSequenceIndex(next.getContig());
+            } else {
+                next = null;
+                nextContigIndex = sequenceDictionary.getSequences().size();
+            }
+            if (intervalContigIndex == nextContigIndex) { // same contig
+                addPartitionReadExtent(extents, i, current.getContig(), current.getStart(), next.getStart() + maxLocatableLength);
```
droazen: So we're now overestimating the extent of each partition by `maxLocatableLength`, and this is set to 1000 by default, correct? Can you briefly explain the performance implications of this? I assume that we'll just potentially pull in a few extra partitions per interval that don't actually contain reads overlapping the interval?

src/main/java/org/broadinstitute/hellbender/utils/recalibration/RecalibrationArgumentCollection.java
tomwhite: Done

src/test/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSparkUnitTest.java
tomwhite: Done

src/test/java/org/broadinstitute/hellbender/engine/spark/SparkSharderUnitTest.java
tomwhite: I was a bit over-zealous in re-writing that test. I've added the extents back to the diagram.

src/main/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSpark.java
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/engine/spark/LocusWalkerSpark.java
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/engine/spark/RangePartitionCoalescer.java
tomwhite: Done

tomwhite:
Thanks for the latest review comments @droazen. I've addressed them all. I'll merge this once the tests pass.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -44,7 +45,13 @@
  *
  */
 public abstract class CommandLineProgram {
-    protected final Logger logger = LogManager.getLogger(this.getClass());
+    protected transient Logger logger = LogManager.getLogger(this.getClass());
+
+    private void readObject(java.io.ObjectInputStream in)
+            throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
+        logger = LogManager.getLogger(this.getClass()); // Logger is not serializable (even by Kryo)
```
tomwhite: Yes, it looks like that would fix it. That change can be handled in #2176 if folks agree about moving to SLF4J (I would support it).

tomwhite: @cmnbroad I'm not sure about the serialization cost (once per task, so maybe not a problem). The issue here is that `FeatureManager` has a `CommandLineProgram` field, and we want to broadcast `FeatureManager`. In the original PR I made the field transient, which seemed to work OK, but changed it after review feedback. We could revisit this if you think it's worth looking at.

PR #2191, opened 2016-09-28
Fix GenomeLocParser error reporting.

cmnbroad:
The current error handling swallows the detailed part of the error string that gets reported to the user, i.e., without this change you see:
A USER ERROR has occurred: Badly formed genome unclippedLoc: Failed to parse Genome Location string: 20:10000000-10010000

With this change:
A USER ERROR has occurred: Badly formed genome unclippedLoc: Parameters to GenomeLocParser are incorrect:The genome loc coordinates 10000000-10010000 exceed the contig size (200000)

I originally removed the entire try/catch block, but there is downstream code that throws IllegalArgumentException, which can benefit from being decorated with the context info.


droazen:
:+1:

PR #2195, opened 2016-10-03
Added CountingVariantFilter

magicDGS:
Following the same implementation as `CountingReadFilter`, this is a way of counting the `VariantContext` that were filtered. This is useful for report the status of `VariantFilter`s.

There are not tests yet because I would like to have the ok on this before spending time on implement them. I think that the community could benefit from using this, and even some tools from the GATK itself.


droazen:
@cmnbroad could you review this one, since you implemented the current read filtering framework?

cmnbroad:
@magicDGS I like your suggestion of factoring out a CountingFilter class that can be reused for both filter types. Ideally, I think we should first get https://github.com/broadinstitute/gatk/pull/2218/files in first (it has some minor changes to CountingReadFilter). Hopefully that will be soon. In the meantime, if you're so inclined, you might try doing the CountingFilter refactoring as part of this PR (which still needs tests), since you'll have two clients for it. That will definitely require some careful refactoring of the existing classes in order to retain the current CountingFilter behavior).

magicDGS:
Thanks for the feedback @cmnbroad! I could wait til #2218 is accepted to continue working and check that nothing is broked...

cmnbroad:
@magicDGS #2218 is merged now.

cmnbroad:
@magicDGS BTW, its probably possible to factor out a generic CountingFilter, but it may not be worth the cost in complexity required. I think the type constraints are (for X in {ReadFilter, VariantFilter}):

- an Xfilter is-a predicate
- a CountingXFilter is-an XFilter
- any filter resulting from AND/OR’ing a CountingXFilter also is-a CountingXFilter
- CountingXFilter reports counts at each granular level

If you have a solution that satisfies these constraints an isn't overly complicated, that would be great! But if not, a separate CountingVariantFilter is also fine.

cmnbroad:
Hi @magicDGS. I went down this same path (composition of CountingFilter) when I originally implemented CountingReadFilter, but I abandoned it for the model we currently have. I think the current model is much simpler in a number of ways. This is an interesting problem, but I would say lets just implement a straight CountingVariantFilter/tests and not try to do the common implementation. 

BTW, when looking at this PR I noticed two bugs in the existing code that we should make sure not to propagate to the Variant one (feel free to fix/test these as part of this PR):

- CountingReadFilter.resetFilterCount only resets the root filter count; it needs an override in BinOp to propagate the reset call to the lhs/rhs operands.
- there is a bug in the getSummary tests/code; you can see the fix [here](https://github.com/broadinstitute/gatk/commit/9ef1458271834aed9b64a5d66f94df33f025eafb)

Thanks!

magicDGS:
@cmnbroad, I opened a PR to fix `CountingReadFilter` and here I push two commits: one removing the complicated implementation and comming back to an independent `CountingVariantFilter`and the second with the tests. Can you review?

magicDGS:
That should be now ready to review, @cmnbroad! I rebased to the latest master and now this should pass the tests.

src/test/java/org/broadinstitute/hellbender/engine/filters/CountingVariantUnitTest.java
```diff
@@ -0,0 +1,308 @@
...
+import htsjdk.variant.variantcontext.VariantContextBuilder;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.*;
+
+public final class CountingVariantUnitTest {
+
+    // Mirrors CountingVariantFilterUnitTest
```
cmnbroad: This file *is* the counting variant unit test - I would just remove this comment since it was relevant in the counting read filter tests, but its not here.)

src/test/java/org/broadinstitute/hellbender/engine/filters/CountingVariantUnitTest.java
```diff
@@ -0,0 +1,308 @@
...
+import htsjdk.variant.variantcontext.Allele;
+import htsjdk.variant.variantcontext.VariantContext;
+import htsjdk.variant.variantcontext.VariantContextBuilder;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.*;
+
+public final class CountingVariantUnitTest {
```
magicDGS: Sure, I think that this was an error caused by some refactoring/rebase... I can't actually remember....

src/test/java/org/broadinstitute/hellbender/engine/filters/CountingVariantUnitTest.java
magicDGS: Copy-paste issue coming from CountingReadFilterUnitTest. Removed!

src/test/java/org/broadinstitute/hellbender/engine/filters/CountingVariantUnitTest.java
```diff
@@ -0,0 +1,308 @@
...
+        filters.add(symbolicFilter);
+
+        // Since we want to ensure that order of the input is honored, we need to test the
+        // structure of the filter rather than the result
+        CountingVariantFilter vf = CountingVariantFilter.fromList(filters);
+
+        Assert.assertTrue(vf.getClass() == CountingVariantFilter.CountingAndVariantFilter.class);
+        CountingVariantFilter.CountingAndVariantFilter andFilter = (CountingVariantFilter.CountingAndVariantFilter) vf;
+
+        // lhs is a Counting and filter; rhs is a counting filter that delegates to GOOD_CIGAR
```
magicDGS: Done.

src/test/java/org/broadinstitute/hellbender/engine/filters/CountingVariantUnitTest.java
```diff
@@ -0,0 +1,308 @@
...
+        Assert.assertTrue(andFilter.lhs.getClass() == CountingVariantFilter.CountingAndVariantFilter.class);
+        Assert.assertTrue(andFilter.rhs.delegateFilter == symbolicFilter);
+        andFilter = (CountingVariantFilter.CountingAndVariantFilter) andFilter.lhs;
+
+        // lhs is a Counting filter that delegates to MAPPING_QUALITY_AVAILABLE; rhs is a
+        // counting filter that delegates to MAPPED
```
magicDGS: Done.

magicDGS:
Addressed comments - back to you @cmnbroad!

I rebased into the latest master, but not squashed to keep track of the changes. I think that it could be squashed at the same time as merging, couldn't it? let me know if you also need me to squash...

Issue #2140, opened 2016-09-02
Request: Packaging of common processing tools for input files


magicDGS:
When working with custom codecs, the input require to be indexed by `IndexFeatureFile`. This require to pack this tool into the software that is implemented, but the package where it is sited is outside any other.

I propose to do not include any tool in `org.broadinstitute.hellbender.tools` to allow the API user to include specific bundled tools in their software, as `IndexFeatureFile` or `CreateSequenceDictionary`. So one package could be just the required for process the input files to make them valid.

magicDGS:
Another idea is to allow to include class names into the instanceMain method.

PR #2204, opened 2016-10-12
Main.instanceMain support for passing CommandLineProgram without a package

magicDGS:
Simple patch to allow passing specific tool classes to `Main.instanceMain` instead of whole packages to solve #2140.

This will allow clients that want their own command line with their tools to include only `IndexFeatureFile` for their own codecs and/or bundle tools like `CreateSequenceDictionary` to pre-process input files.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -16,8 +16,8 @@ import java.util.*;
  *
  * CommandLinePrograms are listed in a single command line interface based on the java package specified to instanceMain.
  *
- * If you want your own single command line program, extend this class and give instanceMain a new list of java packages in which to
- * search for classes that extend CommandLineProgram.
+ * If you want your own single command line program, extend this class and give instanceMain a new list of java packages
+ * in which to search for classes that extend CommandLineProgram and/or a list of single classes.
  *
  */
 public class Main {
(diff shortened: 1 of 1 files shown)


magicDGS:
Because we were discussing about how to manage Main classes for API clients (as part of #2122 to have `CommandLineProgramTester` for custom Main classes), do you think that you can review this PR, @lbergelson?

It is very important for some of my work, because I need to include the `IndexFeatureFile` for custom `Feature` and currently I should add all the hellbender tools to allow them. Thanks in advance!

droazen:
@lbergelson please review

lbergelson:
@magicDGS This seems like a good idea.   I have a few comments.

1.  It seems like there should be an analogous function to `getPackageList()` where you set the list of individual classes, maybe `getNamedToolList()`.

2.  It seems like there's a design flaw in the `Main` class.  I thought that the intent was that people could inherit from `Main` and override `getPackageList()` in order to substitute their own packages.  But for some reason `getPackageList()` is static which is preventing that from happening.  I think we should make it non-static which would allow a subclass of main to just override `getPackageList` and `getNamedToolList()`.  What does your derived main class look like now?

magicDGS:
Actually what you said is correct: it is very painful to extend the `Main` class and use the current implementation. The simplest `Main` class that I'm using it's not extending the GATK one just because of the static methods (see the code [here](https://github.com/magicDGS/thaplv/blob/master/src/main/java/org/magicdgs/thaplv/Main.java)), that's why I would like to include this change.

In few minutes I will commit the changes that you propose, including the method and the change from static. Thanks for the feedback, @lbergelson!

magicDGS:
Back to you @lbergelson, with a new commit with the ideas that you suggested. This will make the managing of my tools much easier. Many thanks!

lbergelson:
@magicDGS 👍 when the tests pass.  I didn't realize how broken this was.  Sorry it was so painful!

magicDGS:
Thanks a lot @lbergelson! The tests passed for CLOUD=false, but not for mandatory.

lbergelson:
Yeah, those cloud tests are messed up for people outside of Broad again.  We need to fix them so they just silently skip.

PR #2215, opened 2016-10-14
Fix bug making HaplotypeCaller fail for non-diploid samples.  Closes #2214.

### LINKED ISSUE #2214: HaplotypeCaller failing or ploidy != 2
This is due to`GATKVAriantContextUtils::makeGenotypeCall` using `GenotypeLikelihoods.getAlleles`, which throws an error because you have to explicitly tell `GenotypeLikelihoods` to initialize ploidies != 2.  We can fix this by explicitly initializing but it seems that `GenotypeLikelihoodCalculators` has the same functionality but more robust.

davidbenjamin:
@takutosato can you review this fix of the bug we were discussing with @droazen?  For comparison, `AlleleSubsettingUtils` uses `GenotypeLikelihoodCalculators` in a similar way.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
@@ -17,7 +17,10 @@ import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.broadinstitute.hellbender.tools.walkers.genotyper.GenotypeAlleleCounts;
 import org.broadinstitute.hellbender.tools.walkers.genotyper.GenotypeAssignmentMethod;
+import org.broadinstitute.hellbender.tools.walkers.genotyper.GenotypeLikelihoodCalculator;
+import org.broadinstitute.hellbender.tools.walkers.genotyper.GenotypeLikelihoodCalculators;
 import org.broadinstitute.hellbender.utils.*;
 
 import java.io.File;
@@ -36,6 +39,8 @@ public final class GATKVariantContextUtils {
     public static final String MERGE_INTERSECTION = "Intersection";
 
     public static final int DEFAULT_PLOIDY = HomoSapiensConstants.DEFAULT_PLOIDY;
+
+    private static final GenotypeLikelihoodCalculators GL_CALCS = new GenotypeLikelihoodCalculators();
 
     public static final double SUM_GL_THRESH_NOCALL = -0.1; // if sum(gl) is bigger than this threshold, we treat GL's as non-informative and will force a no-call.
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -274,9 +279,10 @@ public static void makeGenotypeCall(final int ploidy,
                 gb.alleles(noCallAlleles(ploidy)).noGQ();
             } else {
                 final int maxLikelihoodIndex = MathUtils.maxElementIndex(genotypeLikelihoods);
-                final List<Allele> maxLikelihoodAlleles = GenotypeLikelihoods.getAlleles(maxLikelihoodIndex, ploidy).stream()
-                        .map(allelesToUse::get).collect(Collectors.toList());
-                gb.alleles(maxLikelihoodAlleles);
+                final GenotypeLikelihoodCalculator glCalc = GL_CALCS.getInstance(ploidy, allelesToUse.size());
+                final GenotypeAlleleCounts alleleCounts = glCalc.genotypeAlleleCountsAt(maxLikelihoodIndex);
+
+                gb.alleles(alleleCounts.asAlleleList(allelesToUse));
```
droazen: Can you add a unit test that calls `makeGenotypeCall()` with various ploidies != 2 that fails before the fix and passes after the fix?

davidbenjamin: @droazen Done.  The unit test was already testing ploidy !=2, but was calling the initializer explicitly.  All I had to do was remove that.

takutosato:
Yup I will review

takutosato:
👍 Looks good!

davidbenjamin:
@droazen do you approve?

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
```diff
@@ -1364,8 +1362,6 @@ public void testUpdateGenotypeAfterSubsetting(final int ploidy,
                                                   final List<Allele> allelesToUse,
```
droazen: Might want to rename this test method to reflect the new method name -- eg., `testMakeGenotypeCall`

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -274,9 +279,10 @@ public static void makeGenotypeCall(final int ploidy,
                 gb.alleles(noCallAlleles(ploidy)).noGQ();
             } else {
                 final int maxLikelihoodIndex = MathUtils.maxElementIndex(genotypeLikelihoods);
-                final List<Allele> maxLikelihoodAlleles = GenotypeLikelihoods.getAlleles(maxLikelihoodIndex, ploidy).stream()
-                        .map(allelesToUse::get).collect(Collectors.toList());
-                gb.alleles(maxLikelihoodAlleles);
+                final GenotypeLikelihoodCalculator glCalc = GL_CALCS.getInstance(ploidy, allelesToUse.size());
+                final GenotypeAlleleCounts alleleCounts = glCalc.genotypeAlleleCountsAt(maxLikelihoodIndex);
```
droazen: Have you confirmed that the performance of this method is roughly the same as before?

droazen:
@davidbenjamin A few last comments -- merge after addressing

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
davidbenjamin: Yes, confirmed.  Also, it appears to fix a bug because the htsjdk code had a hard-wired assumption that diploid == biallelic.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -37,6 +40,8 @@
 
     public static final int DEFAULT_PLOIDY = HomoSapiensConstants.DEFAULT_PLOIDY;
 
+    private static final GenotypeLikelihoodCalculators GL_CALCS = new GenotypeLikelihoodCalculators();
```
davidbenjamin: Yes, there is a shared cache but this cache is filled completely in the constructor of `GL_CALCS` and afterwards never mutates.  Thus there is no danger in sharing it between threads.

Issue #2198, opened 2016-10-06
Add an easy way to create composite read filters


droazen:
It would be convenient to have a method that provides an easy way to compose `ReadFilters` from a `List<ReadFilter>`. Eg., it would do something like:

```
        filters.forEach(filter -> filter.setHeader(header));
        ReadFilter combinedFilter = filters.get(0);
        for ( int i = 1; i < filters.size(); ++i ) {
            combinedFilter = combinedFilter.and(filters.get(i));
        }
        return combinedFilter;
```

cmnbroad:
Closed via #2218.

Issue #2175, opened 2016-09-16
Move read filter plugin initialization to GATKTool


lbergelson:
There's lots of duplicate code for initializing read filters in different walker types.  (and at least one bug so far #2173)).  It seems like almost every walker type including feature and variant walkers may want to be able to filter their reads.  We should consolidate the read filter code in GATKTool and add support to FeatureWalker and VariantWalker for filtering their side inputs.

droazen:
Assigning to @cmnbroad to make this happen for this quarter.

droazen:
Should also include an update to gatk-protected to make the tools there use the new API in a more consistent way.

droazen:
As part of this, should hook up to the `ReadsContext` in variant tools using a `ReadFilteringIterator`

cmnbroad:
Closed via #2218.

PR #2218, opened 2016-10-18
Move ReadFilter plugin integration up to GATKTool.

cmnbroad:
- Move command line parser integration of read filters up to GATKTool (https://github.com/broadinstitute/gatk/issues/2175)
- Added fromList method to ReadFilter and CountingReadFilter (https://github.com/broadinstitute/gatk/issues/2198)
- Minor change/rationalization of naming and implementation of BQSR filter methods to match the rest of the framework
- Made a small change to the implementation of the base read filter class to improve clarity/testability
- Opportunistic removal of extraneous imports in unrelated classes


magicDGS:
Hi @cmnbroad, I wonder if as part of this PR the `CountingReadFilter` could be generalized to be used with other kind of filters (like `VariantFilter`, to make easier things like #2195).

Thanks a lot in advance!

droazen:
Can you rebase this one @cmnbroad?

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilter.java
```diff
@@ -25,45 +28,55 @@
...
         private static final long serialVersionUID = 1L;
 
-        //package accesible for testing
-        final protected ReadFilter other;
+        final protected ReadFilter lhs;
+        final protected ReadFilter rhs;
+
+        public ReadFilterBinOp(final ReadFilter lhs, final ReadFilter rhs) {
+            Utils.nonNull(lhs, "AndFilter lhs filter cannot be null");
+            Utils.nonNull(rhs, "AndFilter rhs filter cannot be null");
```
droazen: `AndFilter` -> `BinOpFilter`

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilter.java
```diff
@@ -72,6 +85,28 @@ public boolean test( GATKRead read ) {
...
+     * Return a composite (and) {@code ReadFilter} constructed from a list of {@code ReadFilter}. Each
+     * filter in the list is first initialized with the {@code SAMFileHeader} param. The resulting filter
+     * honors the order of the input list and tests the filter conditions in the same order as the iteration
+     * order of the input list.
+     * @param readFilters If null or empty, the ALLOW_ALL_READS read filter will be returned
+     * @param samHeader {@code SAMFileHeader} used to initialize each filter. May not be null
+     * @return Composite ReadFilter
+     */
+    public static ReadFilter fromList(final List<ReadFilter> readFilters, final SAMFileHeader samHeader) {
+        Utils.nonNull(samHeader != null, "SAMFileHeader must not be null");
```
droazen: Whoops: `Utils.nonNull(samHeader != null` -> `Utils.nonNull(samHeader`

src/main/java/org/broadinstitute/hellbender/engine/filters/CountingReadFilter.java
```diff
@@ -33,6 +38,28 @@ public CountingReadFilter(final ReadFilter readFilter) {
...
+     * Return a composite (and) {@code CountingReadFilter} constructed from a list of
+     * {@link org.broadinstitute.hellbender.engine.filters.ReadFilter}. Each filter in the list is first
+     * initialized with the {@code SAMFileHeader} param. The resulting filter honors the order of the input list
+     * and tests the filter conditions in the same order as the iteration order of the input list.
+     * @param readFilters If null or empty, the ALLOW_ALL_READS read filter will be returned
+     * @param samHeader {@code SAMFileHeader} used to initialize each filter. May not be null
+     * @return Composite CountingReadFilter
+     */
+    public static CountingReadFilter fromList(final List<ReadFilter> readFilters, final SAMFileHeader samHeader) {
+        Utils.validateArg(samHeader != null, "SAMFileHeader must not be null");
```
droazen: Use `Utils.nonNull()` here

src/main/java/org/broadinstitute/hellbender/tools/spark/BaseRecalibratorSpark.java
```diff
@@ -50,8 +50,8 @@
     }
 
     @Override
-    public ReadFilter makeReadFilter() {
-        return BaseRecalibrator.getStandardBQSRReadFilter(getHeaderForReads());
+    public List<ReadFilter> getDefaultReadFilters() {
+        return BaseRecalibrator.getFullBQSRReadFilterList();
```
droazen: `getFullBQSRReadFilterList()` -> `getStandardBQSRReadFilterList()`

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/BQSRPipelineSpark.java
```diff
@@ -95,10 +94,7 @@ protected void runTool(final JavaSparkContext ctx) {
         // is all the filtering that ApplyBQSR wants. BQSR itself wants additional filtering
         // performed, so we do that here.
         //NOTE: this filter doesn't honor enabled/disabled commandline filters
-        final ReadFilter bqsrReadFilter =
-                BaseRecalibrator.makeBQSRSpecificReadFilters()
-                        .stream()
-                        .reduce(ReadFilterLibrary.ALLOW_ALL_READS, (f1, f2) -> f1.and(f2));
+        final ReadFilter bqsrReadFilter = ReadFilter.fromList(BaseRecalibrator.getBasicBQSRReadFilterList(), getHeaderForReads());
```
droazen: `BQSRSpecific` is a more precise descriptor than `Basic`, I think

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -120,6 +125,51 @@
...
+     * and returns a single composite filter resulting from the list by and'ing them together.
+     *
+     * Default tool implementation of {@link #traverse()} calls this method once before iterating
+     * over the reads and reuses the filter object to avoid object allocation. Nevertheless, keeping state in filter
+     * objects is strongly discouraged.
+     *
+     * Multiple filters can be composed by using {@link org.broadinstitute.hellbender.engine.filters.ReadFilter}
+     * composition methods.
+     */
+    public CountingReadFilter makeReadFilter(){
```
droazen: Should this method be final? Is it intended to ever be overridden? It seems like we only ever need to override `getDefaultReadFilters()` (we should make this clearer in the docs in any case).

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectBaseDistributionByCycleSpark.java
```diff
@@ -50,11 +50,6 @@
     @Argument(shortName="F", fullName = "pfReadsOnly", doc="If set to true calculate the base distribution over PF reads only.")
     public boolean pfReadsOnly = false;
 
-    @Override
-    public ReadFilter makeReadFilter() {
-        return ReadFilterLibrary.ALLOW_ALL_READS;
-    }
```
droazen: It seems to me that we still need this override, since the inherited default is `WellformedReadFilter`

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerIntegrationTest.java
```diff
@@ -107,4 +111,33 @@ public void testBestSequenceDictionary_FromVariantIndex() throws Exception {
...
+        final File bamFile = new File(publicTestDir + "org/broadinstitute/hellbender/engine/example_variants_WithReads.bam");
+        final String[] args = {
+                "--variant", vcfFile.getCanonicalPath(),
+                "--input", bamFile.getCanonicalPath(),  // all reads have MAPQ == 0
+                "--readFilter", ReadFilterLibrary.MAPPING_QUALITY_NOT_ZERO.getClass().getSimpleName(),
+                "--hasBackingReadSource", "true",
+                "--hasBackingReads", "false"
+        };
+        tool.instanceMain(args);
+    }
```
droazen: Would be better to have a test that only filters some of the reads, instead of just all or none. Also applies to `ReadsContextUnitTest` above. Eg., make an artificial read filter that only passes reads with certain names.

src/test/java/org/broadinstitute/hellbender/engine/filters/CountingReadFilterUnitTest.java
```diff
@@ -244,5 +247,62 @@ public void testSubFilterCounts(GATKRead[] reads, long totalRejections, long sta
         Assert.assertEquals(badStartAndEnd.getFilteredCount(), startEndRejections);
         Assert.assertEquals(isRayOrEgon.getFilteredCount(), nameRejections);
     }
+
+    @Test
+    public void testFromListNull() {
+        CountingReadFilter rf = CountingReadFilter.fromList(null, ArtificialReadUtils.createArtificialSamHeader(1, 1, 10));
+        Assert.assertTrue(rf.delegateFilter.getClass().getSimpleName().equals(
+                ReadFilterLibrary.AllowAllReadsReadFilter.class.getSimpleName()));
```
droazen: You can assert equality on the class objects themselves using `==`, since they are singletons (here and below).

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
```diff
@@ -348,50 +341,25 @@ public void testPreserveToolDefaultFilterOrder() {
...
+                ReadFilterLibrary.MAPPING_QUALITY_NOT_ZERO.getClass().getSimpleName(),
+                ReadFilterLibrary.PAIRED.getClass().getSimpleName(),
+                ReadFilterLibrary.NONZERO_FRAGMENT_LENGTH_READ_FILTER.getClass().getSimpleName(),
+                ReadFilterLibrary.FIRST_OF_PAIR.getClass().getSimpleName(),
+                ReadFilterLibrary.PROPERLY_PAIRED.getClass().getSimpleName(),
+                ReadFilterLibrary.NOT_DUPLICATE.getClass().getSimpleName(),
+                ReadFilterLibrary.NOT_SECONDARY_ALIGNMENT.getClass().getSimpleName(),
+                ReadFilterLibrary.NOT_SUPPLEMENTARY_ALIGNMENT.getClass().getSimpleName(),
+                ReadFilterLibrary.HAS_MATCHING_BASES_AND_QUALS.getClass().getSimpleName(),
+                ReadFilterLibrary.GOOD_CIGAR.getClass().getSimpleName()
```
droazen: The new expected orders in these tests still reflects the logical order requested by the client, I assume?

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterUnitTest.java
```diff
@@ -110,4 +114,71 @@ public void testDeeperChaining(GATKRead read, boolean expected){
         Assert.assertEquals(readChecksOut.and(r -> false).test(read), false);
         Assert.assertEquals(readChecksOut.or(r -> true).test(read), true);
     }
+
+    @Test
+    public void testFromListNull() {
+        ReadFilter rf = ReadFilter.fromList(null, ArtificialReadUtils.createArtificialSamHeader(1, 1, 10));
+        Assert.assertTrue(rf.getClass().getSimpleName().equals(
+                ReadFilterLibrary.AllowAllReadsReadFilter.class.getSimpleName()));
```
droazen: assert equality on the Class objects themselves

src/test/resources/org/broadinstitute/hellbender/engine/example_variants_WithReads.vcf
```diff
@@ -0,0 +1,8 @@
+##fileformat=VCFv4.1
```
droazen: These new test files should probably be named after the class in whose tests they are used.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -297,33 +293,29 @@ public CountingReadFilter getMergedCountingReadFilter(final SAMFileHeader samHea
...
+            final BiFunction<List<ReadFilter>, SAMFileHeader, T> aggregateFunction) {
 
         Utils.nonNull(samHeader);
-        Utils.nonNull(wrapperFunction);
-        Utils.nonNull(mergeFunction);
+        Utils.nonNull(aggregateFunction);
 
         if (disableAllReadFilters) {
-            return wrapperFunction.apply(ReadFilterLibrary.ALLOW_ALL_READS);
+            return aggregateFunction.apply(null, samHeader);
```
cmnbroad: It does with respect to this PR. That is, the two aggregate functions implementations we have return ALLOW_ALL_READS. However, I'll change the javadoc for the aggregate function argument in this method and codify that it should return a wrapped ALLOW_ALL_READS when passed null.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -120,6 +125,51 @@
...
+        return Collections.singletonList(new GATKReadFilterPluginDescriptor(getDefaultReadFilters()));
+    }
+
+    /**
+     * Returns the default list of CommandLineReadFilters that are used for this tool. The filters returned
+     * by this method are subject to selective enabling/disabling by the user via the command line. The
+     * default implementation returns an empty list. Subclasses can override to provide alternative filters.
+     *
+     * Note: this method is called before command line parsing begins, and thus before a SAMFileHeader is
+     * available through {link #getHeaderForReads}.
```

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -120,6 +125,51 @@
...
+     * @return List of individual filters to be applied for this tool.
+     */
+    public List<ReadFilter> getDefaultReadFilters() {
+        return Collections.singletonList(new WellformedReadFilter());
+    }
+
+    /**
+     * Returns a read filter (simple or composite) that can be applied to reads.
+     * The default implementation combines the default read filters for this tool (returned by
+     * {@link org.broadinstitute.hellbender.engine.GATKTool#getDefaultReadFilters} with any read filter command
```

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -120,6 +125,51 @@
...
+     * line arguments specified by the user; wraps each filter in the resulting list with a CountingReadFilter;
+     * and returns a single composite filter resulting from the list by and'ing them together.
+     *
+     * Default tool implementation of {@link #traverse()} calls this method once before iterating
```
cmnbroad: Changed to a prescriptive comment.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
cmnbroad: Doing so eliminates the chance for a tool (like a multiple metrics collector or a pipeline) to provide custom filter validation/processing post-command line integration unless it overrides traverse. I don't think we have anything that depends on that at the moment though, so I can make it and the GATKSparkTool one final for now.

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
```diff
@@ -11,7 +11,6 @@
 
 import java.util.Collections;
 import java.util.List;
-import java.util.stream.StreamSupport;
```
cmnbroad: No, that should be removed.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -120,6 +125,51 @@
...
+     * over the reads and reuses the filter object to avoid object allocation. Nevertheless, keeping state in filter
+     * objects is strongly discouraged.
+     *
+     * Multiple filters can be composed by using {@link org.broadinstitute.hellbender.engine.filters.ReadFilter}
+     * composition methods.
+     */
+    public CountingReadFilter makeReadFilter(){
+        final GATKReadFilterPluginDescriptor readFilterPlugin =
+                commandLineParser.getPluginDescriptor(GATKReadFilterPluginDescriptor.class);
+        return getHeaderForReads() == null ?
```
cmnbroad: Hm. I don't think we have a clear story on that. It looks like we have some code that assumes that if we have reads, we have a header (and sequence dictionary), and some code that guards. But I don't see anyplace where we enforce that a header is present (and our use of setHeaderStrict implies that we require one). The spec says the header is optional. It's pretty easy to get an NPE with SplitReads by trying a samfile with no header. I propose we settle on a policy, and I'll make a separate ticket to audit/enforce whatever we choose. Let me know what you think.

src/main/java/org/broadinstitute/hellbender/engine/ReadsContext.java
```diff
@@ -40,8 +44,22 @@ public ReadsContext() {
      * @param interval interval over which to query (may be null)
      */
     public ReadsContext( final ReadsDataSource dataSource, final SimpleInterval interval ) {
+        this(dataSource, interval, null);
+    }
+
+    /**
+     * Create a ReadsContext backed by the supplied source of reads. Calls to {@link #iterator}
+     * will return reads overlapping the provided interval. The data source and/or interval
```
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/filters/CountingReadFilter.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/filters/CountingReadFilter.java
```diff
@@ -193,7 +221,7 @@ public String getName() {
     /**
      * Private class for Counting OR filters
      */
-    private final class CountingOrReadFilter extends CountingBinopReadFilter {
+    private static final class CountingOrReadFilter extends CountingBinopReadFilter {
```
cmnbroad: There are no direct tests for the inner ("and"/"or") classes; only indirect ones that verify the results through the public and/or methods. The reason the "and" inner classes are visible for testing is because at some point I added tests that examine the structure of an "and" filter to verify that ordering is preserved during filter execution, because there are usages where read filters are "and-composed" in a specific order for performance reasons. There are no such test for or execution (or any usages of "or" ...AFAIK) since it seemed overkill, though I can add them if you think its worth it.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilter.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilter.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
droazen: If there's a plausible use case for overriding, it's fine to keep it non-final. But I would clarify in the docs that 99%+ of the time tool authors should be overriding `getDefaultReadFilters()`

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
droazen: We should always require a header for reads, and blow up if there's a headerless read input (which is something I've never seen, and am not convinced is actually allowed).

src/main/java/org/broadinstitute/hellbender/engine/filters/CountingReadFilter.java
droazen: Yeah, not necessary to add the overkill tests

src/main/java/org/broadinstitute/hellbender/tools/spark/BaseRecalibratorSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/BQSRPipelineSpark.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectBaseDistributionByCycleSpark.java
cmnbroad: Right. I think it should be in a getDefaultReadFiltersList override, though.

src/test/java/org/broadinstitute/hellbender/engine/ReadsContextUnitTest.java
```diff
@@ -32,4 +31,31 @@ public void testEmptyReadsContext( final ReadsContext readsContext ) {
...
+                        new SimpleInterval("1", 200, 210),
+                        new ReadFilter() {
+                            private static final long serialVersionUID = 1L;
+                            @Override public boolean test(final GATKRead read) {return false;}}), false
+                }
+        };
+    }
+
+    @Test(dataProvider = "ValidReadsContextDataProvider")
+    public void testValidReadsContext(final ReadsContext readsContext, final boolean hasReads ) {
```
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
cmnbroad: Yes. The order verifier used to validate the order in the reverse of the expected order. In this PR, I rewrote it to match reality so I could eliminate this confusion. You now list the expected values in the same order you expect the filter to test them.

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerIntegrationTest.java
cmnbroad: Done.

Issue #1648, opened 2016-03-28
Better GATK-wide support for HDFS and GCS (via Java NIO Path and htsjdk-level patches)


droazen:

droazen:
Depends on https://github.com/broadinstitute/gatk/issues/958

lbergelson:
Also depends on https://github.com/broadinstitute/gatk/issues/1547

droazen:
Replaced by the more specific https://github.com/broadinstitute/gatk/issues/3709

PR #2224, opened 2016-10-20
Convert ReadsDataSource to use Paths, so it can read inputs from the cloud

jean-philippe-martin:
As a result, every walker that takes a BAM as input can now read it from
the cloud directly.

Sample use: ./gatk-launch CountReads -I
"gs://.../CEUTrio.HiSeq.WGS.b37.ch20.4m-12m.NA12878.bam"
(...)
Tool returned:
7275701

Note that this CL doesn't add a prefetcher or multithreading, so the
performance is inferior to that of the NIO proof of concept which was
using both.

cc: @droazen

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
@@ -380,6 +380,7 @@ public final class BucketUtils {
     /**
      * String -> Path. This *should* not be necessary (use Paths.get(URI.create(...)) instead) , but it currently is
      * on Spark because using the fat, shaded jar breaks the registration of the GCS FilesystemProvider.
+     * To transform other types of string URLs into Paths, use IOUtils.getPath instead.
      */
     public static java.nio.file.Path getPathOnGcs(String gcsUrl) {
         final String[] split = gcsUrl.split("/");
@@ -387,7 +388,6 @@ public final class BucketUtils {
         final String pathWithoutBucket = String.join("/", Arrays.copyOfRange(split, 3, split.length));
         return CloudStorageFileSystem.forBucket(BUCKET).getPath(pathWithoutBucket);
     }
-
 
     /**
      * Get an authenticated GCS-backed NIO FileSystem object representing the selected projected and bucket.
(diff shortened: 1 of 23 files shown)


jean-philippe-martin:
Related to #1648

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -388,6 +389,22 @@ public static String getPathWithoutBucket(String path) {
...
     }
 
+    /**
+     * String -> Path. This will accept either
+     * "folder/file", or
+     * "file://folder/file", or
+     * "gs://bucket/folder/file".
+     * It will only accept "hdfs://" if you have an NIO filesystem provider installed for HDFS.
+     */
+    public static java.nio.file.Path getPath(String fileNameOrUrl) {
```
tomwhite: There is a similar method in IOUtils#getPath that should be used instead of adding another one here. It will need to have the GCS special casing added though.

jean-philippe-martin: Sounds good. That one can throw IOException. Are we OK with this pattern?

```java
// some function that doesn't throw IOException
try {
  // code goes here, BucketUtils -> IOUtils
} catch (IOException io) {
  throw new RuntimeException(io);
}
```
?

jean-philippe-martin: `new GATKException(io.getMessage())` it is!

jean-philippe-martin: Actually a bunch of those are in tests. For those I think it's OK to add a new exception type to the test method.

jean-philippe-martin:
Build is failing because of a warning. It's odd that we didn't see this before, it seems unrelated to my change:

```
:compileJava/home/travis/.gradle/caches/modules-2/files-2.1/com.google.cloud/gcloud-java-nio/0.2.8/57e30d28f80ab3a320e560e7b4aaac07baf98c4/
gcloud-java-nio-0.2.8-shaded.jar(com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.class): warning:
Cannot find annotation method 'value()' in type 'AutoService': class file for shaded.cloud-nio.com.google.auto.service.AutoService not found
error: warnings found and -Werror specified
1 error
1 warning
 FAILED
```

droazen:
Hey @jean-philippe-martin, this looks good. I've made some very small changes to avoid the back-and-forth of a review, and rebased the branch onto latest master. 

The main change I made was in `IOUtils.getPath()`. It now traps the `IOException` and throws a `UserException` instead. This has the benefit of not requiring client code to put `throws IOException` (or catch the `IOException`) everywhere, and is more consistent with the rest of the codebase, which typically traps checked exceptions as early as possible and wraps them in unchecked exceptions (either `UserException` or `GATKException`, depending on whether it's likely to be the user's fault or not).

Also made a small change in `NioBam` to make it use a logger instead of `System.out.println()` for debug output.

jean-philippe-martin:
Thank you @droazen !

jean-philippe-martin:
Filed #2331 to add prefetching. As expected this has a big positive impact on performance.

Issue #2225, opened 2016-10-21
Request: parseArgs catch CommandLineException thrown by customCommandLineValidation


magicDGS:
While using a custom tool that validates arguments (e.g., certain range for an `Integer` parameter), a way of handling it is override `CommandLineProgram.customCommandLineValidation()` and return an String with the error found.

Nevertheless, if parsing the arguments throws a `UserException.CommandLineException`, the error is printed with a concrete format (`printDecoratedUserExceptionMessage`) after the usage. This is different from the custom validation, which is printed without any decoration and before the usage. Although this behavior could be desirable, I expect that if my custom validation thrown an `UserException.CommandLineException` it is printed in the same way as other exceptions, and the exception is re-thrown in the same way (for testing purposes, for instance). But the current behaviour just exit without any error printed because the exception is catched in `Main`.

A very minor change is include in the `try` block the custom validation.

PR #2226, opened 2016-10-21
parseArgs catch CommandLineException thrown by customCommandLineValidation

magicDGS:
* Include custom validation in try-catch block to handle custom `UserException.CommandLineException` (solves #2225)
* Unchanged behaviour for the rest of the tools.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -205,9 +205,21 @@ public abstract class CommandLineProgram {
     protected boolean parseArgs(final String[] argv) {
 
         commandLineParser = new CommandLineParser(this, getPluginDescriptors());
-        final boolean ret;
         try{
-            ret = commandLineParser.parseArguments(System.err, argv);
+            final boolean ret = commandLineParser.parseArguments(System.err, argv);
+            commandLine = commandLineParser.getCommandLine();
+            if (!ret) {
+                return false;
+            }
+            final String[] customErrorMessages = customCommandLineValidation();
+            if (customErrorMessages != null) {
+                for (final String msg : customErrorMessages) {
+                    System.err.println(msg);
+                }
+                commandLineParser.usage(System.err, false);
+                return false;
+            }
+            return true;
         } catch (final UserException.CommandLineException e){
             //The CommandLineException is treated specially - we display help and no blow up
             commandLineParser.usage(System.err, true);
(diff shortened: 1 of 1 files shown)


droazen:
@lbergelson please review

lbergelson:
@magicDGS We've inherited customCommandLineValidation() from picard, but it never really caught on as a way to do things in GATK.  The fact that it's output is inconsistent with the rest of the command line parsing is an accident and should be fixed.  

I think the right thing to do here would be to make `customCommandLineValidation()` into a `void` method that either throws a `CommandLineException` or doesn't.  It will take some changes to a few tools but it will make things less confusing in the long run.  Then you can move the call to customCommandLineValidation to just after the call to `parseArgs` like you've done, but we can remove the custom code to print error messages since it will just be handled by the regular exception handling.

magicDGS:
Even if I'll refactore it to be void, the part that I changed is need it: there is no other part which handle the `CommandLineException`  and if the `customCommandLineValidation()` throws the exception no error is printed in the terminal.

I guess that returning a `String[]` in Picard is done to output several errors in the command line to avoid the user to re-run with another bug not reported. Nevertheless, I prefer you approach. I'm changing now the code in this PR to add what you suggested.

Thanks again for make my development smoother, @lbergelson!

lbergelson:
Yeah, being able to print multiple lines of errors at once is nice for the users.  Most tools don't do that though.  The ones that do can wrap the multiple errors into a single exception with a multiline message.  Will changing the method signature and moving the call up into the try block fix the problem your having?

magicDGS:
My problem is fixed with the first commit, because if `customCommandLineValidation()` throws an `UserException.CommandLineException` it is catched and printed (otherwise, it is ignore in `Main`, except for the returning status).

Anyway, I added a commit with the implementation of the void method for all the tools. I guess that it is not a good idea after all, because it could help to print several errors. My first commit deal with the two situations, printing one/several errors if the `String[]` is not null, and if the validation throws an error catching it and handle in the same way as the ones in `CommandLineProgram`. I will rather go for the first and changing the printing of the errors in the array to the same place as the catched exceptions, and decorate it in the same way for not confusing the software user.

What do you thing, @lbergelson?

lbergelson:
@magicDGS I was thinking that we could reproduce essentially the same output using the void method  having any of the methods that could return multiple error message wrap those multiple error messages into a single exception.  

If you'd rather just merge your initial solution we can revisit changing the method signature in a later pr.

magicDGS:
Ok, I removed the commit with the signature change, and I added a very minor one with an added javadoc to the method to show that `UserException.CommandLineException` will be correctly handled now.

Back to you @lbergelson (I guess that the CLOUD=mandatory will fail again). Let me know if you require any other change for accept this.

lbergelson:
@magicDGS 👍 Will merge when tests are done.  Sorry for the back and forth.  I think I might make the separate changes to customCommandLineValidation separately.

magicDGS:
Thanks @lbergelson! Tests passed for CLOUD=false, so back to you for merge.

lbergelson:
👍 thanks

PR #2227, opened 2016-10-21
updating to gradle 3.1

lbergelson:
this will enable use of composite gradle builds which should reduce development pain


droazen:
:+1: merging

PR #2229, opened 2016-10-22
change qss's count type to R

takutosato:
The quality score sum (QSS) format field reports the sum of base qualities for ref and alt allele, and its correct format field count type is 'R' (see 1.2.2 in the [vcf spec](http://samtools.github.io/hts-specs/VCFv4.2.pdf)). We used to use the wrong count type of 1 and that caused a parsing error in the dream challenge evaluation script.


takutosato:
Hi @davidbenjamin, could you review this one-liner?

droazen:
:+1: merging

Issue #2222, opened 2016-10-20
Write a tool to add a sequence dictionary to a VCF


droazen:
The tool should take a VCF plus a source of a sequence dictionary (eg., a fasta reference, a bam, a `.dict` file, or another VCF that already has a dictionary), and add the sequence dictionary to the VCF.

droazen:
For @cmnbroad and alpha-4.

cmnbroad:
Closed via https://github.com/broadinstitute/gatk/pull/2234.

Issue #2223, opened 2016-10-20
Require that VCF inputs to the GATK have a sequence dictionary


droazen:
Depends on #2222, which should be done first.

The error message should point users to the tool that allows them to add a sequence dictionary to their VCF.

droazen:
To clarify this one:

-If VCF(s) are the only tool input(s) (ie., no reference or reads are available), we should require that they all have sequence dictionaries in their headers, and throw if they don't. But if a reference or reads are available, we should use their dictionary and not require the VCF(s) to have one.

-We should also add a `GATKTool`-level `--sequenceDictionary` argument that allows the user to provide a master sequence dictionary in the form of a `.dict` file. When this argument is specified, the sequence dictionary from the `.dict` should be used as the master dictionary everywhere, and we should not require sequence dictionaries in VCF headers.

droazen:
Spun out https://github.com/broadinstitute/gatk/issues/2410 for the `--sequenceDictionary` task.

PR #2232, opened 2016-10-25
Ported Picard tool UpdateVcfSequenceDictionary

magicDGS:
Solves #2222

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/UpdateVcfSequenceDictionary.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/UpdateVcfSequenceDictionary.java
@@ -1,0 +1,110 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


magicDGS:
You are assigned to the issue, @cmnbroad. Can you review, please? Thanks a lot!

cmnbroad:
@magicDGS Thanks for doing this, but I've already [implemented](https://github.com/broadinstitute/gatk/commit/8106d95cf68e24983f1d0c31f041b4cae9f75f51) a native GATK version of this, rather than a port of the Picard tool - I just haven't quite finished the tests.

magicDGS:
But if your implementation is a `VariantWalker` and #2223 is implemented, this will class. By the way, I don't know if there is any plan to include all the tools from Picard, but it seems that the behaviour will be replicated if so, no?

Thanks anyway for the feedback, close the PR in case you think that it is not necessary.

cmnbroad:
Yeah - not sure about the duplication (I didn't even know this Picard tool existed). @droazen whats your preference ?

droazen:
I'd tend to prefer a native GATK tool, but let's see both versions as pull requests and pick the best/most functional one. A key requirement in the ticket is to support many different sources of sequence dictionaries: a fasta reference, a bam, a .dict file, another VCF that already has a dictionary, etc.

droazen:
@cmnbroad Could you review this branch, and give a summary of how this version compares to the one in https://github.com/broadinstitute/gatk/pull/2234? Is one implementation more functional than the other?

magicDGS:
I had a look to the other branch, @droazen, and I think that it is more functional than this one:

* Check if the input already have a sequence dictionary, and only updates if  `--replace` is provided. The version in this PR just overrides the dictionary.
* Check if all the variants agree with the new sequence dictionary, throwing an error if the contig is not present or the variant falls outside the chromosome range. This version does not account at all for that.
* It is a `VariantWalker`, and thus the code is simplest. But the pitfall of this is that if #2223 is implemented, that class will require a dictionary for the input as a `GATKTool`. I'm not sure how that is going to be done, but I guess that it will introduce problems in the class implemented by @cmnbroad.

I think that the other version is more complete and I like it more because it is more concern about putative problems.

cmnbroad:
I agree with everything @magicDGS said. In general, the native GATK tool benefits from all of the automatic framework management and I/O support (like honoring md5, lenient args, etc.). #2234 also has the validation checking mentioned, and has more tests. (It is true that when #2223 goes in, we'll have to override the default sequence dictionary validation behavior, maybe via disableSequenceDictionaryValidation).

So I think we should take the native one, but if we do choose to keep this one for any reason, then I'll still want to do a line-by-line CR before we merge.

lbergelson:
@magicDGS @cmnbroad Going to close this one in favor of #2234.

PR #2234, opened 2016-10-27
UpdateVCFSequenceDictionary tool.

cmnbroad:
If we decide to take both versions of this (Picard port and this one) we might want to rename this one.


src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/UpdateVCFSequenceDictionary.java
```diff
@@ -0,0 +1,150 @@
...
+                  "alignment, dictionary, or reference file. The dictionary must be valid for all variants in the " +
+                  " target file.",
+        oneLineSummary = "Updates the sequence dictionary in a variant file.",
+        programGroup = VariantProgramGroup.class
+)
+public final class UpdateVCFSequenceDictionary extends VariantWalker {
+    static final Logger logger = LogManager.getLogger(UpdateVCFSequenceDictionary.class);
+
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
```
cmnbroad: Remove shortName as its unnecessary.

cmnbroad: Here and below.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/UpdateVCFSequenceDictionary.java
```diff
@@ -0,0 +1,150 @@
...
+        if (samSeqRec == null) {
+            throw new UserException.BadArgumentValue(
+                String.format(
+                    "The input variant file contains a variant (ID: \"%s\") with a reference to a sequence (\"%s\") " +
+                    "that is not present in the provided dictionary",
+                    vc.getID(),
+                    vc.getContig()
+                )
+            );
+        } else if (vc.getEnd() > samSeqRec.getSequenceLength()) {
```
lbergelson: This should probably be a warning instead.

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/UpdateVCFSequenceDictionaryIntegrationTest.java
```diff
@@ -0,0 +1,118 @@
...
+                );
+
+        // Some sequence dictionary sources will contain optional attributes (i.e., if the source is a .dict file,
+        // or if only a reference is presented to the tool using -R, which will in turn cause the framework to retrieve
+        // the dictionary from the .dict file accompanying the reference, the dictionary will likely include md5 and
+        // UR attributes). However, htsjdk doesn't propagate these attributes to the VCF header properly
+        // (https://github.com/samtools/htsjdk/issues/730), and many are stripped out. In order to ensure the
+        // roundtrip comparison succeeds, roundtrip it through a VCF header to match what htsjdk will have written out.
+        VCFHeader sourceVCFHeader = new VCFHeader();
+        sourceVCFHeader.setSequenceDictionary(sourceDictionary); //
```
cmnbroad: Extraneous // at the end

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/UpdateVCFSequenceDictionary.java
magicDGS: Why not checking if it is an structural variant and if not thrown the error?

lbergelson: So on further thought, I don't think it's actually valid at all to have a structural variant off the end.  Maybe leave this as is and if someone shows up with a file that has them we can reconsider.  I was looking at the spec to try to tell for sure, but it seems like end should always be on the reference.  Sorry for the confusion.

PR #2186, opened 2016-09-27
VariantRecalibrator optimizations

cmnbroad:
A handful of simple optimizations for VariantRecalibrator:
- Preallocate arrays when the size is known
- Eliminate unnecessary boxing of doubles
- Lift some loop invariants with unnecessary allocations (this eliminates millions of array allocations on the full SNP test used by GATK3)

The current GATK4 (multi-variant walker) implementation is about 3% faster than GATK3 without these; these bring it to about 6% faster.

PR #2235, opened 2016-10-27
Make new qual model numerically stable

davidbenjamin:
This fixes pathologies that @ldgauthier noticed in a subset of ExAC.  @vruano can you take a look?  Note that the changed test files only change VQSLOD here and there by 0.03 or so and the new values are probably more accurate.


src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/AlleleFrequencyCalculator.java
```diff
@@ -123,37 +129,43 @@ public AFCalculationResult getLog10PNonRef(final VariantContext vc, final int de
...
             }
             final GenotypeLikelihoodCalculator glCalc = GL_CALCS.getInstance(g.getPloidy(), numAlleles);
-            final double[] genotypePosteriors = normalizedGenotypePosteriors(g, glCalc, log10AlleleFrequencies);
+
+            final double[] log10GenotypePosteriors = log10NormalizedGenotypePosteriors(g, glCalc, log10AlleleFrequencies);
 
             new IndexRange(0, glCalc.genotypeCount()).forEach(genotypeIndex ->
                 glCalc.genotypeAlleleCountsAt(genotypeIndex).forEachAlleleIndexAndCount((alleleIndex, count) ->
-                        result[alleleIndex] += genotypePosteriors[genotypeIndex] * count));
+                        log10Result[alleleIndex] = MathUtils.approximateLog10SumLog10(log10Result[alleleIndex], log10GenotypePosteriors[genotypeIndex] + MathUtils.log10(count))));
```
vruano: Why not use MathUtils.log10sumlog10(...) instead of the approximated version?

vruano: I'm using a loaner computer now so I am a bit blind ... perhaps this is not available in GATK4 but otherwise you should use MathUtils.Log10Cache.get(count). If it has been dropped in GATK4, forget about it.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -653,30 +653,20 @@ public static double distanceSquared(final double[] x, final double[] y) {
...
      * @return
      */
     public static double[] normalizeFromLog10(final double[] array, final boolean takeLog10OfOutput, final boolean keepInLogSpace) {
-        // for precision purposes, we need to add (or really subtract, since they're
-        // all negative) the largest value; also, we need to convert to normal-space.
-        double maxValue = arrayMax(array);
-
-        // we may decide to just normalize in log space without converting to linear space
         if (keepInLogSpace) {
+            double maxValue = arrayMax(array);
```
vruano: ```final```

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -653,30 +653,20 @@ public static double distanceSquared(final double[] x, final double[] y) {
...
             return applyToArray(array, x -> x - maxValue);
```
vruano: I though that variables used in lambdas needed to be final, I guess I was mistaken.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -653,30 +653,20 @@ public static double distanceSquared(final double[] x, final double[] y) {
...
-        // default case: go to linear space
-        final double[] normalized = applyToArray(array, x -> Math.pow(10.0, x - maxValue));
-        final double sum = sum(normalized);
-        if (!takeLog10OfOutput) {
-            return applyToArrayInPlace(normalized, x -> x/sum);
         } else {
-            final double log10Sum = Math.log10(sum);
-            return applyToArray(array, x -> x - maxValue - log10Sum);
+            final double log10Sum = log10SumLog10(array);
+            final double[] result = applyToArray(array, x -> x - log10Sum);
```
vruano: It seems that if you merge ```result``` and ```array``` this would be another in-place operation, in that case call of them ```result```.

cmnbroad:
@davidbenjamin I'm wondering if similar changes have been made to GATK3. The reason I ask is that I have a GATK4 branch that contains a temporary command line arg for VQSR that allows it to exactly reproduce the GATK3 results (by conditioning the initial state of the random number generator). We intended to use it to make it easy to compare the results with those produced by GATK3. I'm wondering if this change will cause the results to diverge from GATK3 on the same inputs.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/AlleleFrequencyCalculator.java
davidbenjamin: The approximate version is a lot faster and is still very accurate, but the new qual score is so inexpensive compared to calculating PLs that your suggestion will probably be hardly noticeable in the overall runtime.

davidbenjamin: As for the second comment, `MathUtils.log10(int)` uses the `Log10Cache` in GATK4.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: `array` is the original input, which the calling code might not want to modify.  However, usually the unnormalized array in the calling code *is* simply a temporary variable, so I wrote a version of this method that *does* modify the input in-place.

davidbenjamin: Update: in fact, *every* normalization in the whole code base can be done in-place, so I refactored MathUtils.

davidbenjamin:
@cmnbroad I have not made similar changes to GATK3 but given that this patch is critical to the new qual score model I will want to backport soon.

davidbenjamin:
@vruano in addition to refactoring MAthUtils I also engaged in a few boy scout rule activities.

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -732,23 +727,11 @@ public static double arrayMax(final double[] array) {
...
-     * Checks that the result is a well-formed log10 probability
-     *
      * @param result a supposedly well-formed log10 probability value
-     * @param allowNegativeInfinity should we consider a -Infinity value ok?
      * @return true if result is really well formed
      */
-    public static boolean goodLog10Probability(final double result, final boolean allowNegativeInfinity) {
-        return result <= 0.0 && result != Double.POSITIVE_INFINITY && (allowNegativeInfinity || result != Double.NEGATIVE_INFINITY) && ! Double.isNaN(result);
+    public static boolean goodLog10Probability(final double result) {
+        return result <= 0.0 && result != Double.POSITIVE_INFINITY && !Double.isNaN(result);
```
vruano: if ```result <= 0.0``` then result cannot be +Inf nor a NaN so you can simplify it to just:

``` return result <= 0.0 ```.

Comment (vruano) [inline src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java]: Logical operations involving NaN always return false even ```NaN == NaN```

File: src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/AlleleFrequencyCalculator.java [hunk 6]
```diff
@@ -104,7 +104,7 @@ public AFCalculationResult getLog10PNonRef(final VariantContext vc, final int de
                 final double log10GenotypePosterior = log10GenotypePosteriors[genotype];
                 glCalc.genotypeAlleleCountsAt(genotype).forEachAlleleIndexAndCount((alleleIndex, count) ->
                         log10ProbabilityOfNonZeroAltAlleles[alleleIndex] =
-                                MathUtils.approximateLog10SumLog10(log10ProbabilityOfNonZeroAltAlleles[alleleIndex], log10GenotypePosterior));
+                                MathUtils.log10SumLog10(log10ProbabilityOfNonZeroAltAlleles[alleleIndex], log10GenotypePosterior));
```
davidbenjamin: @vruano I already made the change and the performance is fine, so if you don't mind I'll keep it *un*approximate.  (I would say "exact" but that's an overloaded adjective here).

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/GaussianMixtureModel.java
```diff
@@ -127,56 +127,36 @@ public void expectationStep( final List<VariantDatum> data ) {
...
-        }
+        double sumPK = gaussians.stream().mapToDouble(g -> g.sumProb).sum();
 
-        int gaussianIndex = 0;
-        double[] pGaussianLog10 = new double[gaussians.size()];
-        for( final MultivariateGaussian gaussian : gaussians ) {
-            pGaussianLog10[gaussianIndex++] = Math.log10( gaussian.sumProb / sumPK );
-        }
-        pGaussianLog10 = MathUtils.normalizeFromLog10( pGaussianLog10, true );
+        final double[] pGaussianLog10 = gaussians.stream().mapToDouble(g -> Math.log10(g.sumProb / sumPK)).toArray();
```
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
davidbenjamin: Done and simplified `goodProbability` similarly.

davidbenjamin:
Back to @vruano.

droazen:
@davidbenjamin Whoops, this was not supposed to be merged until after https://github.com/broadinstitute/gatk/pull/2186, and after the `MathUtils` change we talked about. Can you open a second PR with the `MathUtils` change?

vruano:
Sorry I thought I was good to merge.

davidbenjamin:
My fault for not telling @vruano.  I will make the changes to `MathUtils` ASAP.

PR #2237, opened 2016-10-28
PathSeq initial pull

mwalker174:
@cwhelan @tedsharpe please review. There are two tools here:

1) PathSeqKmerSpark - builds kmer library as a hopscotch set from a given reference (done)
2) PathSeqFilterSpark - preprocessing / human read filtering tool (work in progress, currently has read preprocessing steps implemented)

Made modifications to the SV kmer classes: 

1) SVKmerSmall class uses only 1 long (up to 31 bases), largely copied from SVKmer
2) SVKmer and SVKmerSmall are both derived from SVKmerBase
3) SVKmerizer uses generics so it can work with either type, moved some static functions to SVUtils

There are a bunch of new read filters and transformers for the filtering steps, including symmetric DUST. Unit tests included.


src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+    @SuppressWarnings("unchecked")
+    protected void runTool(final JavaSparkContext ctx) {
+
+        header = getHeaderForReads();
+        ctx.hadoopConfiguration().setBoolean(BAMInputFormat.KEEP_PAIRED_READS_TOGETHER_PROPERTY, true);
+
+        final JavaRDD<GATKRead> reads = getReads();
+
+        //Filter secondary/supplementary reads and reads that fail the vendor quality check
+        JavaRDD<GATKRead> primaryReads = reads.filter(read -> !(read.isSecondaryAlignment() || read.failsVendorQualityCheck() || read.isSupplementaryAlignment()));
```
cmnbroad: @mwalker174 If this tool had an override of getDefaultFilter, it could just add the corresponding built-in filters to the set returned there. This would make them subject to being enabled, disabled, and configured via their parameters by the user from the command line, and they would be automatically applied by the engine (you could then remove this line completely - the call to getReads() right above returns the RDD with such filter handling already applied). In theory we could do that for all of the filters below, but it looks like there are some like these that are applied before transformations, then some map/transforms, then some post filtering. We've talked about embracing that as pattern in the engine - this tool may be a good use-case for doing so.

cmnbroad:
@mwalker174 It sounds like this is still a work in progress, so I"m not sure what kind of ffedback you're looking for at this point but I made a couple of suggestions relating to how these tools and the read filters integrate with the framework. FWIW.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizer.java
```diff
@@ -12,23 +12,23 @@
  * Iterator over successive Kmers from a sequence of characters.
  * Silently skips over parts of the sequence that has characters other than A, C, G, or T.
  */
-public class SVKmerizer implements Iterator<SVKmer> {
+public class SVKmerizer<KType extends SVKmerBase> implements Iterator<KType> {
```
cwhelan: Also, do you really need to make this class generic over `KType extends SVKmerBase`? Would it be possible to just make all the return types be `SVKmerBase` instead of `KType`? That way you wouldn't have to do casting in this class. Or would that do something that I'm not seeing like break serialization. Also you wouldn't have to change the subclasses of SVKmerizer.

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+import org.broadinstitute.hellbender.utils.clipping.ClippingOp;
+import org.broadinstitute.hellbender.utils.clipping.ClippingRepresentation;
+import org.broadinstitute.hellbender.utils.clipping.ReadClipper;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+import static org.broadinstitute.hellbender.utils.read.ReadUtils.*;
+import static org.broadinstitute.hellbender.utils.read.ReadUtils.setDeletionBaseQualities;
+
+/**
+ * Created by markw on 10/21/16.
```
cwhelan: We generally remove these auto-generated comments.

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+     * to the end of the read.  This is blatantly stolen from BWA.
+     * <p/>
+     * Walk through the read from the end (in machine cycle order) to the beginning, calculating the
+     * running sum of qTrimmingThreshold - qual.  While we do this, we track the maximum value of this
+     * sum where the delta > 0.  After the loop, clipPoint is either -1 (don't do anything) or the
+     * clipping index in the read (from the end).
+     *
+     */
+    @Override
+    public GATKRead apply(GATKRead read) {
```
cwhelan: Params and local variables in this class can be made final.

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
```diff
@@ -0,0 +1,230 @@
...
+        public float score = 0;
+    }
+
+    @Override
+    public GATKRead apply(final GATKRead read) {
+
+        byte[] q = read.getBases();
+
+        //Find convert to 2-bit and assign random bases to N's
+        //TODO: reject reads with too many N's
```
cwhelan: I think you already have the ambiguous base read filter so you could remove this TODO.

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
```diff
@@ -0,0 +1,230 @@
...
+        Random rng = new Random();
+        rng.setSeed(0);
```
cwhelan: I think you might want to use the Random from `org.broadinstitute.hellbender.utils.Utils`, which centralizes managing  seeds and resetting the generator.

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
```diff
@@ -0,0 +1,230 @@
...
+                } else {
+                    logger.warn("Invalid base " + (char) q[i] + " in read " + read.getName() + ", assigning to random base.");
+                    q[i] = (byte)rng.nextInt(4);
+                }
+            }
+        }
+
+        //Start main algorithm
+        List<Tuple<Integer,Integer>> res = new ArrayList<>();
+        LinkedList<PStruct> P = new LinkedList<>();
```
cwhelan: I think it would add a lot to the readability/maintainability of this algorithm if you could assign some more descriptive names to variables and structures. (I realize you probably based this on someone else's implementation or description). P, PStruct, rv, rw, cv, cw, and L are all pretty impenetrable names. Based on your understanding of the algorithm, do you think you could do some renaming here? Failing that, a link to somewhere that explains the algorithm and would let people decipher it would definitely help.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerBase.java
```diff
@@ -0,0 +1,66 @@
...
+    // This is most quickly and easily done with a lookup table.
+    protected static final long[] BYTEWISE_REVERSE_COMPLEMENT;
+    static {
+        BYTEWISE_REVERSE_COMPLEMENT = new long[256];
+        for ( int idx = 0; idx != 256; ++idx ) {
+            BYTEWISE_REVERSE_COMPLEMENT[idx] = reverseComplementByteValueAsLong(idx);
+        }
+    }
+
+    protected abstract void serialize(final Kryo kryo, final Output output );
```
tedsharpe: Ought to remove this.  It's not really a feature of a kmer, but an implementation detail to be handled in each concrete class that needs it.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
```diff
@@ -0,0 +1,190 @@
...
+    public SVKmerSmall( final int kSize ) {
+        Utils.validateArg(kSize >= 1 && kSize < 32, "K must be between 1 and 31.");
+        valLow = 0;
+    }
+
+    public SVKmerSmall( final SVKmerSmall that ) { this.valLow = that.valLow; }
+
+    private SVKmerSmall( final long valLow ) { this.valLow = valLow; }
+
+    protected SVKmerSmall( final Kryo kryo, final Input input ) {
```
tedsharpe: should be private

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
```diff
@@ -0,0 +1,190 @@
...
+        valLow = input.readLong();
+    }
+
+    protected void serialize( final Kryo kryo, final Output output ) {
```
tedsharpe: should be private

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
```diff
@@ -0,0 +1,190 @@
...
+ */
+@DefaultSerializer(SVKmerSmall.Serializer.class)
+public class SVKmerSmall extends SVKmerBase implements Comparable<SVKmerSmall> {
+    // these are treated as K-bit unsigned integers
+    private final long valLow; // Kmer bits
+
+    // Lookup table for reverse-complementing each possible byte value.
+    // Each pair of bits represents a base, so you have to reverse bits pairwise and then invert all bits.
+    // This is most quickly and easily done with a lookup table.
+    private static final long[] BYTEWISE_REVERSE_COMPLEMENT;
```
tedsharpe: left in by accident -- already in base class

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
```diff
@@ -0,0 +1,190 @@
...
+        // we'll produce the string in reverse order and reverse it at the end
+        long val = valLow;
+        for ( int nChars = kSize/2; nChars > 0; --nChars ) {
+            // grab the two least significant bits to index into the BASE_CHARS array
+            sb.append(BaseUtils.BASE_CHARS[(int)val & 3]);
+            // roll the whole mess down two bits
+            val >>= 2;
+        }
+        // this for loop will have one more iteration than the previous one if kSize is odd
+        for ( int nChars = (kSize+1)/2; nChars > 0; --nChars ) {
```
tedsharpe: this loop was necessary when there were two longs, but it's superfluous in this class.  correction above does it all in the first for loop.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Not an override.  An SVKmerSmall doesn't know what K is, so it has to be supplied.
+     */
+    public final String toString( final int kSize ) {
+        final StringBuilder sb = new StringBuilder(kSize);
+
+        // we'll produce the string in reverse order and reverse it at the end
+        long val = valLow;
+        for ( int nChars = kSize/2; nChars > 0; --nChars ) {
```
tedsharpe: you can do the whole job in this loop -- just fix the boundary
`for ( int idx = 0; idx != kSize; ++idx )`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
```diff
@@ -0,0 +1,190 @@
...
+            // move 'em down
+            val >>= 2;
+        }
+        // we built the string in least-significant to most-significant bit order.  reverse it now.
+        return sb.reverse().toString();
+    }
+
+
+    // Reverse-complement a long by taking the reverse-complement of each of its bytes in reverse order.
+    private static long reverseComplement( long val ) {
```
tedsharpe: this could go into the base class

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmer.java
```diff
@@ -208,11 +169,6 @@ public final String toString( final int kSize ) {
         return sb.reverse().toString();
     }
 
-    private static long reverseComplementByteValueAsLong( final int bIn ) {
-        // this turns the 8 bits [b1 b2 b3 b4 b5 b6 b7 b8] into [~b7 ~b8 ~b5 ~b6 ~b3 ~b4 ~b1 ~b2]
-        return ~(((bIn & 3) << 6) | (((bIn >> 2) & 3) << 4) | (((bIn >> 4) & 3) << 2) | ((bIn >> 6) & 3)) & 0xffL;
-    }
-
     // Reverse-complement a long by taking the reverse-complement of each of its bytes in reverse order.
     private static long reverseComplement( long val ) {
```
tedsharpe: move to base class

src/main/java/org/broadinstitute/hellbender/engine/filters/MarkedDuplicateReadFilter.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Set;
+
+public final class MarkedDuplicateReadFilter extends ReadFilter implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    //Filters out optical duplicate reads (marked with the OD tag)
+    @Override
+    public boolean test( final GATKRead read ) {
+        if (read.hasAttribute(MarkDuplicatesSparkUtils.OPTICAL_DUPLICATE_TOTAL_ATTRIBUTE_NAME)
```
tedsharpe: is it possible for a read to have an OD=1 tag without its flag being set to indicate that it's a duplicate?  in other words, since you're testing isDuplicate() can you eliminate this test?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizer.java
tedsharpe: I think my muddled thinking sent Mark in the wrong direction.  I agree that this class doesn't need to be generic -- it can just deal with SVKmerBase.  Serialization should work fine -- it finds the correct serializer based on the actual class.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+                    Iterator<GATKRead> iter = p.iterator();
+                    while (iter.hasNext()) {
+                        size++;
+                        iter.next();
+                    }
+                    boolean notPair = (size==1);
+                    ArrayList<GATKRead> newPair = new ArrayList<>(size);
+                    iter = p.iterator();
+                    while (iter.hasNext()) {
+                        newPair.add(CleanRead(iter.next(),notPair));
```
tedsharpe: Yikes.  Not good.  The lambda will capture `this` -- the pointer to the PathSeqFilterSpark tool itself.  See suggestion for fix below.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+        writeReads(ctx, output, readsCleaned);
+    }
+
+    private boolean BaseQualityFilterFn(GATKRead read) {
+        return false;
+    }
+
+    private GATKRead CleanRead(GATKRead read, boolean notPaired) {
+
+        GATKRead new_read = new SAMRecordToGATKReadAdapter(new SAMRecord(header));
```
tedsharpe: I'm pretty sure you can pass `null` for a header as long as you avoid methods that try to turn reference names into reference indices or vice versa (and since you're producing unmapped reads, you should be good to go).  If I'm right about that, then this method can be made static and you won't capture the `this` pointer in the lambda that calls CleanRead.  This header/headerless, reference name/index morass is one of my least favorite parts of GATK.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+        HopscotchSet<SVKmerSmall> kmerLibSet = (HopscotchSet<SVKmerSmall>)kryo.readClassAndObject(input);
+        Broadcast<HopscotchSet<SVKmerSmall>> kmerLibBroadcast = ctx.broadcast(kmerLibSet);
+        ContainsKmerReadFilter readKmerFilter = new ContainsKmerReadFilter(kmerLibBroadcast,KMER_SIZE);
+        final JavaRDD<GATKRead> readsKmerFiltered = readsLengthFiltered.filter(read -> !readKmerFilter.test(read));
+
+        //TODO: bwa filtering against user-specified reference
+
+        //"Clean" reads by resetting all flags and setting pairedness flags according to the remaining read set
+        final JavaRDD<GATKRead> readsCleaned = readsKmerFiltered
+                .groupBy(read -> read.getName()) //group pairs using read name into set of Iterable<GATKRead>'s
```
tedsharpe: This could be really, really expensive for an input with a lot of bacterial reads.  Shuffling a large number of GATKReads is crazy expensive.  (But maybe you're confident that the filtered set will be small?)
I don't have code, or even a detailed design to try to circumvent this, but a sketch of an idea might go something like this:
Persist the readsKmerFiltered RDD.  Map it to just the read names, and shuffle them. Within each partition save just the unpaired names (or just the paired names -- whichever set you expect to be smaller).  Broadcast the set of unpaired (or paired) names.  Now pass over the same RDD you persisted earlier, looking up each name in the set to tell you how to "clean" it.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+                .values()   //Get Iterables
+                .map(p -> {     //Determine if Iterable represents 1 or 2 reads and set flags accordingly
+                    int size = 0;
```
tedsharpe: This lambda is a little messy, IMO.  How about something like:
```
                    final Iterator<GATKRead> readItr = p.iterator();
                    final GATKRead firstRead = readItr.next();
                    if ( !readItr.hasNext() ) return Collections.singletonList(CleanRead(firstRead,true));
                    final List<GATKRead> readList = new ArrayList<>(2);
                    readList.add(CleanRead(firstRead,false));
                    do {
                        readList.add(CleanRead(readItr.next(),false));
                    }
                    while(readItr.hasNext());
                    return readList;
```

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
tedsharpe: Hate to contradict, but I think it would be a mistake to use the Utils RNG:  It will lead to thread contention and stochastic behavior.  I think it's better as it stands.

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
```diff
@@ -0,0 +1,230 @@
...
+
+        byte[] q = read.getBases();
+
+        //Find convert to 2-bit and assign random bases to N's
+        //TODO: reject reads with too many N's
+        Random rng = new Random();
+        rng.setSeed(0);
+        for (int i = 0; i < q.length; i++) {
+            if (q[i] == 'N') {
+                q[i] = (byte)rng.nextInt(4);
```
tedsharpe: Perhaps consider lazily instantiating the RNG here.  (Most reads don't have N's, so you can save some clicks by not instantiating an RNG unless you need one.)  I.e., initialize rng to null up above, and then:
`if ( rng == null ) { rng = new Random(); rng.setSeed(0); }`

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
```diff
@@ -0,0 +1,230 @@
...
+    public GATKRead apply(final GATKRead read) {
+
+        byte[] q = read.getBases();
+
+        //Find convert to 2-bit and assign random bases to N's
+        //TODO: reject reads with too many N's
+        Random rng = new Random();
+        rng.setSeed(0);
+        for (int i = 0; i < q.length; i++) {
+            if (q[i] == 'N') {
```
tedsharpe: Maybe handle lower case 'n', too?

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
```diff
@@ -0,0 +1,230 @@
...
+                    q[i] = 0x0;
+                } else if (q[i] == 'T' || q[i] == 't') {
+                    q[i] = 0x1;
+                } else if (q[i] == 'C' || q[i] == 'c') {
+                    q[i] = 0x2;
+                } else if (q[i] == 'G' || q[i] == 'g') {
+                    q[i] = 0x3;
+                } else {
+                    logger.warn("Invalid base " + (char) q[i] + " in read " + read.getName() + ", assigning to random base.");
+                    q[i] = (byte)rng.nextInt(4);
```
tedsharpe: if you do the lazy instantiation thing, you'll need to do it here, too.

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
```diff
@@ -0,0 +1,230 @@
...
+                }
+            }
+        }
+
+        //Start main algorithm
+        List<Tuple<Integer,Integer>> res = new ArrayList<>();
```
tedsharpe: Using a generic class here (i.e., Tuple) will cause a lot of integer boxing and unboxing.  Consider replacing it with a custom private class that uses ints rather than Integers.  E.g.:
```private static final class Interval { public int a; public int b; }```

File: src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java [hunk 25]
```diff
@@ -0,0 +1,230 @@
...
+
+    @Argument(fullName="Window size", shortName="dustW", optional=true)
+    public int W = 64;
+
+    @Argument(fullName="Score threshold", shortName="dustT", optional=true)
+    public float T = 20;
+
+    private static final int NUM_TRIPLETS = 64;
+
+    private class PStruct {
```
tedsharpe: should be static and final

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
tedsharpe: Yeah.  It's really pretty impenetrable.  Are the comments in this code the only comments in the pseudo-code from the paper?  (If so, shame on the DUST author.)
I'm sure it's going to cost you some time to clean this up enough to make it comprehensible, and I apologize, but we'd be remiss as reviewers to let this go.

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
```diff
@@ -0,0 +1,230 @@
...
+                while (p1 != null && p1.start < wstart) {
+                    P.removeLast();
+                    p1 = P.isEmpty() ? null : P.getLast();
+                }
+            }
+            //End: SAVE_MASKED_REGIONS
+
+            wstart++;
+        }
+
```
tedsharpe: Could you extract all the code up to here as a separate method that returns `res`?  It would be useful in other contexts to be able to find out what the DUST-masked intervals are without having to transform the read.

src/main/java/org/broadinstitute/hellbender/engine/filters/AmbiguousBaseReadFilter.java
```diff
@@ -0,0 +1,35 @@
...
+import org.broadinstitute.hellbender.cmdline.Argument;
+import org.broadinstitute.hellbender.tools.spark.transforms.markduplicates.MarkDuplicatesSparkUtils;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import org.broadinstitute.hellbender.utils.read.ReadUtils;
+
+import java.io.Serializable;
+import java.util.HashSet;
+import java.util.Set;
+
+public final class AmbiguousBaseReadFilter extends ReadFilter implements Serializable {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/engine/filters/AmbiguousBaseReadFilter.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public AmbiguousBaseReadFilter() { }
+
+    public AmbiguousBaseReadFilter( final float n_frac ) { this.N_FRAC = n_frac; }
+
+    //Filters out reads with more than a threshold number of N's
+    @Override
+    public boolean test( final GATKRead read ) {
+        int num_N = 0;
+        for (int i = 0; i < read.getLength(); i++) {
```
mwalker174: Done

mwalker174: Done, using the first suggestion.

src/main/java/org/broadinstitute/hellbender/engine/filters/AmbiguousBaseReadFilter.java
```diff
@@ -0,0 +1,35 @@
...
+            if (read.getBase(i) == 'N') {
```
mwalker174: Done - seems more robust to me.

src/main/java/org/broadinstitute/hellbender/engine/filters/AmbiguousBaseReadFilter.java
```diff
@@ -0,0 +1,35 @@
...
+                num_N++;
+            }
+        }
+        return (num_N < Math.ceil(read.getLength()*N_FRAC));
```
mwalker174: Done. Also changing the behavior slightly this version (<= instead of <)

src/main/java/org/broadinstitute/hellbender/engine/filters/BaseQualityReadFilter.java
```diff
@@ -0,0 +1,36 @@
+package org.broadinstitute.hellbender.engine.filters;
+
+import org.broadinstitute.hellbender.cmdline.Argument;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import java.io.Serializable;
+
+public final class BaseQualityReadFilter extends ReadFilter implements Serializable {
```
mwalker174: I have removed this filter in favor of using BaseQualityReadTransformer to mask low-quality bases with 'N' and then applying AmbiguousBaseReadFilter to filter reads with too many 'N's - thus taking into account N's due to both low base quality as well as low-complexity subsequences.

src/main/java/org/broadinstitute/hellbender/engine/filters/BaseQualityReadFilter.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    @Argument(fullName="baseQualThresh", shortName="baseQualThresh", optional=true)
```
mwalker174: Class removed (see response to Ted's comment above)

src/main/java/org/broadinstitute/hellbender/engine/filters/BaseQualityReadFilter.java
```diff
@@ -0,0 +1,36 @@
...
+    public int BASE_QUALITY_THRESHOLD = 15;
+
+    @Argument(fullName="baseQualFrac", shortName="baseQualFrac", optional=true)
+    public float BASE_QUALITY_MAX_FRAC = 0.05f;
```
mwalker174: Noted. Class removed (see above).

src/main/java/org/broadinstitute/hellbender/engine/filters/BaseQualityReadFilter.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public BaseQualityReadFilter(final int quality_threshold, final float quality_frac) {
+        this.BASE_QUALITY_THRESHOLD = quality_threshold;
+        this.BASE_QUALITY_MAX_FRAC = quality_frac;
+    }
+
+    //Filters out reads with at least BASE_QUALITY_MAX_COUNT fraction of base qualities that are below BASE_QUALITY_THRESHOLD
+    @Override
+    public boolean test( final GATKRead read ) {
+        short numBelowThreshold = 0;
```
mwalker174: Noted. Class removed (see above).

src/main/java/org/broadinstitute/hellbender/engine/filters/BaseQualityReadFilter.java
```diff
@@ -0,0 +1,36 @@
...
+        if (read.getBaseQualityCount() > 0) {
+            for (int q : read.getBaseQualities()) {
```
mwalker174: Noted. Class removed (see above).

mwalker174: Noted. Class removed (see above).

src/main/java/org/broadinstitute/hellbender/engine/filters/BaseQualityReadFilter.java
```diff
@@ -0,0 +1,36 @@
...
+                if (q < BASE_QUALITY_THRESHOLD)
+                    numBelowThreshold++;
+            }
+        }
+        return numBelowThreshold < Math.ceil(BASE_QUALITY_MAX_FRAC*read.getLength());
```
mwalker174: Yes I think so. I must have been thinking of casting the RHS to an int rather than the other way around.

src/main/java/org/broadinstitute/hellbender/engine/filters/ContainsKmerReadFilter.java
```diff
@@ -0,0 +1,38 @@
...
+import org.broadinstitute.hellbender.tools.spark.utils.HopscotchSet;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import org.junit.Before;
+
+import java.io.Serializable;
+
+/**
+ * Keep reads that contain at least one kmer from a broadcasted HopscotchSet of SVKmerSmall's
+ */
+public class ContainsKmerReadFilter extends ReadFilter implements Serializable {
```
mwalker174: Moved class to org.broadinstitute.hellbender.tools.spark.pathseq since this filter is somewhat specialized.

src/main/java/org/broadinstitute/hellbender/engine/filters/ContainsKmerReadFilter.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private static final long serialVersionUID = 1L;
+    private Broadcast<HopscotchSet<SVKmerSmall>> kmerLibBroadcast;
```
mwalker174: Done, following Ted's suggestion below.

src/main/java/org/broadinstitute/hellbender/engine/filters/MarkedDuplicateReadFilter.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.io.Serializable;
+import java.util.HashSet;
+import java.util.Set;
+
+public final class MarkedDuplicateReadFilter extends ReadFilter implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    //Filters out optical duplicate reads (marked with the OD tag)
```
mwalker174: The intended behavior is to remove optical duplicates. Firstly, PCR duplicates are detected using the position of the read alignment and we are not assuming the input BAM is aligned. Second, it is possible to have identical independent reads, particularly with RNAseq data, which is a possible use case.

I renamed the class to MarkedOpticalDuplicateReadFilter and corrected the behavior by looking for the OD tag and not looking at the flags.

src/main/java/org/broadinstitute/hellbender/engine/filters/MarkedDuplicateReadFilter.java
```diff
@@ -0,0 +1,31 @@
...
+    @Override
+    public boolean test( final GATKRead read ) {
+        if (read.hasAttribute(MarkDuplicatesSparkUtils.OPTICAL_DUPLICATE_TOTAL_ATTRIBUTE_NAME)
+                && read.getAttributeAsInteger(MarkDuplicatesSparkUtils.OPTICAL_DUPLICATE_TOTAL_ATTRIBUTE_NAME) != 0) {
+            return false;
+        }
+        else if (read.isDuplicate()) {
```
mwalker174: This line has been removed.

src/main/java/org/broadinstitute/hellbender/engine/filters/MarkedDuplicateReadFilter.java
```diff
@@ -0,0 +1,31 @@
...
+            return false;
+        }
+        else {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+import org.seqdoop.hadoop_bam.BAMInputFormat;
+import scala.Tuple2;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@CommandLineProgramProperties(summary = "Read preprocessing and host filtering on reads from a BAM file",
```
mwalker174: Done. Added a high-level overview and a list of the filtering steps, as well as input and output specifications.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+        oneLineSummary = "PathSeqFilter on Spark",
+        programGroup = SparkProgramGroup.class)
+public final class PathSeqFilterSpark extends GATKSparkTool {
+
+    private static final long serialVersionUID = 1L;
+    private SAMFileHeader header;
+
+    private static final String maxLengthArgName = "pathseq_maxReadLength";
```
mwalker174: I suppose not - I thought some of my arguments might clash with other tools' but I found some that seem different enough without using the prefix.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+    @Argument(fullName = maxLengthArgName,
```
mwalker174: This was just because I was using the existing ReadLengthFilter that requires an upper bound. I've done away with the argument and instead passed Integer.INT_MAX to the filter.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
mwalker174: Overriding getDefaultFilter could be very elegant. I could use that if transformations could be worked in. But it may get confusing if, say, you wanted to use a filter twice. For example if the tool looked like ReadLengthFilter -> ClipRead -> ReadLengthFilter, you would need a way to generate distinct arguments for each ReadLengthFilter. Maybe you've already thought of a way to do this.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+
+        final JavaRDD<GATKRead> reads = getReads();
+
+        //Filter secondary/supplementary reads and reads that fail the vendor quality check
+        JavaRDD<GATKRead> primaryReads = reads.filter(read -> !(read.isSecondaryAlignment() || read.failsVendorQualityCheck() || read.isSupplementaryAlignment()));
+
+        //Filter duplicate reads
+        final OpticalDuplicateFinder finder =
+                new OpticalDuplicateFinder(opticalDuplicatesArgumentCollection.READ_NAME_REGEX, opticalDuplicatesArgumentCollection.OPTICAL_DUPLICATE_PIXEL_DISTANCE, null);
+        JavaRDD<GATKRead> markedReads = MarkDuplicatesSpark.mark(primaryReads, getHeaderForReads(),MarkDuplicatesScoringStrategy.SUM_OF_BASE_QUALITIES, finder, getRecommendedNumReducers());
```
mwalker174: I will ask Mark about this. It looks like this functionality is implemented in the Walker version but not the Spark version of MarkDuplicates.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+
+        ReadFilter markedFilter = new MarkedDuplicateReadFilter();
+        final JavaRDD<GATKRead> markedFilteredReads = markedReads.filter(read -> markedFilter.test(read));
```
mwalker174: Neat! Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+
+        //Apply DUST masking
+        ReadTransformer dustTransformer = new DUSTReadTransformer();
```
mwalker174: Line deleted since I'm using the ReadSparkifier.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+        final JavaRDD<GATKRead> readsDUSTMasked = markedFilteredReads.map(read -> dustTransformer.apply(read));
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+        final JavaRDD<GATKRead> readsAmbigFiltered = readsBQFiltered.filter(read -> ambigFilter.test(read));
+
+        //Filter reads with less than MIN_READ_LENGTH bases
+        ReadFilter readLengthFilter = new ReadLengthReadFilter(MIN_READ_LENGTH,MAX_READ_LENGTH);
+        final JavaRDD<GATKRead> readsLengthFiltered = readsAmbigFiltered.filter(read -> readLengthFilter.test(read));
+
+        //Load Kmer hopscotch set and filter reads containing > 0 matching kmer
+        final PipelineOptions options = getAuthenticatedGCSOptions();
+        Input input = new Input(BucketUtils.openFile(KMER_LIB_PATH, options));
+        Kryo kryo=new Kryo();
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+        HopscotchSet<SVKmerSmall> kmerLibSet = (HopscotchSet<SVKmerSmall>)kryo.readClassAndObject(input);
+        Broadcast<HopscotchSet<SVKmerSmall>> kmerLibBroadcast = ctx.broadcast(kmerLibSet);
+        ContainsKmerReadFilter readKmerFilter = new ContainsKmerReadFilter(kmerLibBroadcast,KMER_SIZE);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
mwalker174: Done. Take a look at the retainPairs() function.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
mwalker174: I realized I don't need to do any read cleaning - the downstream BWA step takes care of that by converting to a Simple Read and back. For now, I am only going to retain read pairs, since that is how they are expected in BWASparkEngine. I have a similar lambda in retainPairs() however.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+        //TODO: bwa filtering against user-specified reference
+
+        //"Clean" reads by resetting all flags and setting pairedness flags according to the remaining read set
+        final JavaRDD<GATKRead> readsCleaned = readsKmerFiltered
+                .groupBy(read -> read.getName()) //group pairs using read name into set of Iterable<GATKRead>'s
+                .values()   //Get Iterables
+                .map(p -> {     //Determine if Iterable represents 1 or 2 reads and set flags accordingly
+                    int size = 0;
+                    Iterator<GATKRead> iter = p.iterator();
+                    while (iter.hasNext()) {
```
mwalker174: This statement was removed, but a similar one now lives in retainPairs() (see response to Ted's comment).

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+                        size++;
+                        iter.next();
+                    }
+                    boolean notPair = (size==1);
```
mwalker174: Line removed but noted.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+                    ArrayList<GATKRead> newPair = new ArrayList<>(size);
```
mwalker174: Line removed but noted.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
mwalker174: Again this lambda was removed but the new one I wrote is in a static function now.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+                    iter = p.iterator();
+                    while (iter.hasNext()) {
+                        newPair.add(CleanRead(iter.next(),notPair));
+                    }
+                    return newPair;
+                })
+                .flatMap(p -> p);   //Flatten back from Iterables to GAKTReads
```
mwalker174: Line removed but noted.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,192 @@
...
+
+        writeReads(ctx, output, readsCleaned);
+    }
+
+    private boolean BaseQualityFilterFn(GATKRead read) {
```
mwalker174: Function removed but noted.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
mwalker174: The intention is to remove all the attributes (except read group). I realized that the BWA alignment downstream will do all this so CleanRead no longer exists.

mwalker174: Code deleted but noted.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqKmerSpark.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /** Get the list of distinct kmers in the reference, and write them to a file. */
+    @Override
+    protected void runTool( final JavaSparkContext ctx ) {
+        final SAMFileHeader hdr = getHeaderForReads();
+        SAMSequenceDictionary dict = null;
+        if ( hdr != null ) dict = hdr.getSequenceDictionary();
+        final PipelineOptions options = getAuthenticatedGCSOptions();
+        final ReferenceMultiSource referenceMultiSource = getReference();
+        List<SVKmerSmall> kmerList = findKmers(ctx, kSize, referenceMultiSource, options, dict);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqKmerSpark.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Turn a text file of overlapping records from a reference sequence into an RDD, and do a classic map/reduce:
+     * Kmerize, mapping map to the collection of all canonicalized kmers, removing duplicates, and returning result to
+     * the driver.
+     */
+    @VisibleForTesting static List<SVKmerSmall> processRefRDD( final int kSize,
+                                                          final JavaRDD<byte[]> refRDD ) {
+        return refRDD.flatMap(seq -> SVKmerizer.stream(seq, kSize, new SVKmerSmall(kSize))
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqKmerSpark.java
```diff
@@ -0,0 +1,109 @@
...
+                .map(kmer -> kmer.canonical(kSize))
+                .collect(Collectors.toList())).distinct().collect();
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqKmerSpark.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private static Collection<SVKmer> uniquify(final Collection<SVKmer> coll1, final Collection<SVKmer> coll2 ) {
```
mwalker174: Nope - removed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerBase.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
mwalker174: Yes - removed this copy.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Returns a SVKmerSmall that is a canonical representation of this one.
+     * An odd-K SVKmerSmall is in canonical form if its middle base is A or C.
+     * The reverse-complement of a non-canonical SVKmerSmall is a canonical SVKmerSmall, and vice versa.  (Think about it.)
+     * Canonical form is not defined for even-K Kmers (too expensive to compute routinely).
+     */
+    public SVKmerSmall canonical( final int kSize ) {
+        Utils.validateArg( (kSize & 1) != 0, "K must be odd to canonicalize.");
+        // for odd-size kmers, the high bit of the middle base is in least significant position in valHigh.
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerSmall.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizer.java
mwalker174: Done. The superclass is now SVKmer, extended by SVKmerShort and SVKmerLong. It required a lot more changes in some of the existing SV tools but I think results in overall cleaner code.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizer.java
```diff
@@ -12,23 +12,23 @@
...
-    public SVKmerizer( final CharSequence seq, final int kSize ) {
+    public SVKmerizer( final CharSequence seq, final int kSize, KType kmer ) {
         this.seq = seq;
         this.kSize = kSize;
-        this.nextKmer = nextKmer(new SVKmer(kSize), 0);
+        this.nextKmer = nextKmer(kmer, 0);
     }
 
-    protected SVKmerizer( final int kSize, final CharSequence seq ) {
+    protected SVKmerizer( final int kSize, final CharSequence seq, KType kmer ) {
```
mwalker174: This was a result of using generics - Java requires there to be an argument with the generic type. No longer relevant since we're not using generics.

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
+package org.broadinstitute.hellbender.transformers;
+
+import htsjdk.samtools.Cigar;
+import htsjdk.samtools.SAMUtils;
+import htsjdk.samtools.util.StringUtil;
+import org.broadinstitute.hellbender.cmdline.Argument;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+import org.broadinstitute.hellbender.tools.ClipReads;
+import org.broadinstitute.hellbender.utils.clipping.ClippingOp;
+import org.broadinstitute.hellbender.utils.clipping.ClippingRepresentation;
+import org.broadinstitute.hellbender.utils.clipping.ReadClipper;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
```
mwalker174: Done.

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+
+import static org.broadinstitute.hellbender.utils.read.ReadUtils.*;
+import static org.broadinstitute.hellbender.utils.read.ReadUtils.setDeletionBaseQualities;
+
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+    private int qTrimmingThreshold = 15;
+
+    /**
+     * Clip bases from the read from
+     * <p/>
+     * argmax_x{ \sum{i = x + 1}^l (qTrimmingThreshold - qual)
+     * <p/>
+     * to the end of the read.  This is blatantly stolen from BWA.
+     * <p/>
+     * Walk through the read from the end (in machine cycle order) to the beginning, calculating the
```
mwalker174: Modified comment to specify how both ends are trimmed.

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+     * running sum of qTrimmingThreshold - qual.  While we do this, we track the maximum value of this
+     * sum where the delta > 0.  After the loop, clipPoint is either -1 (don't do anything) or the
+     * clipping index in the read (from the end).
+     *
+     */
+    @Override
+    public GATKRead apply(GATKRead read) {
+        GATKRead readClippedRightEnd = clipReadRightEnd(read);
+        return clipReadLeftEnd(readClippedRightEnd);
```
mwalker174: The strand is unknown because the input BAM could be unaligned.

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+        byte[] quals = read.getBaseQualities();
+
+        int clipSum = 0, lastMax = -1, clipPoint = -1; // -1 means no clip
+        for (int i = readLength - 1; i >= 0; i--) {
+            clipSum += (qTrimmingThreshold - quals[i]);
+            if (clipSum >= 0 && (clipSum >= lastMax)) {
+                lastMax = clipSum;
+                clipPoint = i;
+            }
+        }
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+        byte[] quals = read.getBaseQualities();
+
+        int clipSum = 0, lastMax = -1, clipPoint = -1; // -1 means no clip
+        for (int i = 0; i < readLength; i++) {
+            clipSum += (qTrimmingThreshold - quals[i]);
+            if (clipSum >= 0 && (clipSum >= lastMax)) {
+                lastMax = clipSum;
+                clipPoint = i;
+            }
+        }
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+
+        if (clipPoint != -1) {
+            int newLength = clipPoint;
+            final byte[] newBases = new byte[newLength];
+            final byte[] newQuals = new byte[newLength];
+            final int copyStart = 0;
+            System.arraycopy(read.getBases(), copyStart, newBases, 0, newLength);
+            System.arraycopy(read.getBaseQualities(), copyStart, newQuals, 0, newLength);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+        }
+
+        if (clipPoint != -1) {
+            int newLength = readLength - 1 - clipPoint;
+            final byte[] newBases = new byte[newLength];
+            final byte[] newQuals = new byte[newLength];
+            final int copyStart = clipPoint + 1;
+            System.arraycopy(read.getBases(), copyStart, newBases, 0, newLength);
+            System.arraycopy(read.getBaseQualities(), copyStart, newQuals, 0, newLength);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
mwalker174: Keeping it the way it is. It's not perfect because the sequence of replacement bases will be the same for every read, but I don't see any better simple solution. It shouldn't be an issue for PathSeq assuming the input reads don't have N's.

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
mwalker174: Implemented a nested IntTuple class.

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
mwalker174: I copied the pseudocode (almost) verbatim and so I wanted to match the variable names. I think it is more readable, though, with some new names. 

The algorithm itself is a little complicated to explain in comments because there are some clever optimizations that require some math. I've added more comments to describe the variables and what each subroutine does, but anyone who wants to modify the code will need to familiarize him/herself with the manuscript first. I added the reference at the top of the function.

Also, the save_masked_regions subroutine appears twice in the algorithm, so I created a function for it. One issue though - two of the arguments are the lists perfect_intervals and final_intervals. Both are modified in the function, resulting in the calling function's instances of these variables to also be changed. This is the desired behavior in this case, but it's my impression this is not good practice. Any thoughts on this?

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
mwalker174: Added function:
`    public static Collection<Tuple2<Integer,Integer>> sDUST(final GATKRead read, final int window, final float tscore)`

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
tedsharpe: The reverse strand flag will never be set in an unaligned BAM, so the read will be right clipped, which is correct because the read will be presented in machine cycle order.  (Reads only get reverse complemented by the aligner when it discovers that they map to the non-canonical strand of the reference.)
Actually, you may need to consider this fact when you're re-writing the reads as unmapped reads:  you might want to RC any of the output reads that have been RC'd.

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
tedsharpe: It's not ideal, but partly it's just the way Java works.  Lists passed as arguments could be modifiable or unmodifiable, and the subroutine might modify them or not, and there's nothing that signals which Lists are unmodifiable nor any way to mark which passed objects will be left unmodified.  I'd just document it and not worry about it.

tedsharpe:
Yes.  Makes perfect sense.

src/main/java/org/broadinstitute/hellbender/transformers/DUSTReadTransformer.java
mwalker174: Okay added a comment.

mwalker174:
@cwhelan @tedsharpe Done with revising - please check. 

I added a few lines to PathSeqFilterSpark to do the BWA alignment with the BWASparkEngine class and filter out the mapped reads.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -0,0 +1,268 @@
...
+                                                itr.next();
+                                                return itr.hasNext();})
+                                            .flatMap(pairedRead -> pairedRead);
+        return pairedReads;
+    }
+
+    /**
+     * This version of retainPairs() groups read name Strings instead of the GATKReads themselves for faster shuffling.
+     */
+    private static JavaRDD<GATKRead> retainPairs_ShuffleByName(final JavaSparkContext ctx, JavaRDD<GATKRead> reads) {
```
cwhelan: This version of the method doesn't seem to be used -- do you want to keep it?

mwalker174: Yes I'd like to keep it for future performance testing on a distributed architecture.

cwhelan:
Looks good to me now, feel free to rebase and merge.

PR #2185, opened 2016-09-23
Eliminate the deprecated PerReadAlleleLikelihoodMap.  Closes #1467.

davidbenjamin:
@ldgauthier this cleans up something you have previously grumbled about in the annotation code.  Do you feel like reviewing?

PR #2239, opened 2016-10-30
Fix an oversight in recent refactoring of annotation code

davidbenjamin:
Two annotations, allele depth and total depth, consider whether reads are informative relative to the alleles in the output `VariantContext`, which is in general a subset of the alleles contains in the `ReadLikelihoods`.  In PR #2185 I overlooked this and forgot to subset the likelihoods' alleles to those of the vc, which was the previous behavior (see the diff from that PR for the two annotations in this PR).  This PR duplicates the old behavior.  This fixes failures in the HaplotypeCaller integration tests introduced by the previous PR.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/DepthPerAlleleBySample.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/DepthPerAlleleBySample.java
@@ -12,6 +12,7 @@ import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.genotyper.ReadLikelihoods;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * Depth of coverage of each allele per sample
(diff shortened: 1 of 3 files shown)


davidbenjamin:
@takutosato This is needed to rev gatk public, which in turn is needed for the Mutect2 refactoring I'm working on.  Can you review?

takutosato:
Yessir

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/DepthPerAlleleBySample.java
```diff
@@ -58,7 +59,9 @@ public void annotate(final ReferenceContext ref,
         for ( final Allele allele : vc.getAlleles() ) {
             alleleCounts.put(allele, 0);
         }
-        likelihoods.bestAlleles(g.getSampleName()).stream()
+        final Map<Allele, List<Allele>> alleleSubset = alleles.stream().collect(Collectors.toMap(a -> a, a -> Arrays.asList(a)));
```
takutosato: You could use a method reference here: `Collectors.toMap(a -> a, Arrays::asList)`. I'm not saying it's better, just wanted to mention that that's an option too.

takutosato: This is not the code change in this PR, but In #2185 you kept the error message "not a strict subset of _per read allele map alleles_" in DepthPerAlleleBySample and others. It's not a big deal obviously but wanted to bring it up in case it wasn't intentional.

davidbenjamin: done and done

davidbenjamin:
Back to @takutosato

takutosato:
👍  good to go!

droazen:
@davidbenjamin Can you craft a simple unit test for these annotations to make sure this stays fixed?

PR #2241, opened 2016-11-01
Restore old numerics of VQSR temporarily

davidbenjamin:
@drozen @cmnbroad this is for you.


droazen:
@cmnbroad please review, and confirm that with this branch we can still match GATK3 output

cmnbroad:
These changes look good to me in terms of reverting to GATK3-matching results.

On a separate note though, the optimizations on my branch are largely aimed at reducing unnecessary allocations and boxing, since this (GMM expectation step) is the main hotspot in VQSR, and we iterate through it millions of times. That may conflict with the use of streams as introduced in these PRs (I don't think they introduce any boxing since they use DoubleStreams, but may still result in the same number of array allocation(s) that were in the original code). When I merge in my changes I'll do some more profiling. The stream code certainly looks nicer so it would be nice if we can keep it.

droazen:
I definitely say kill the pretty streaming code if it's harming performance in these critical sections -- no shame in a straightforward loop!

davidbenjamin:
@cmnbroad `MathUtils` has methods like `applyToArray(double[], <lambda>)` that are concise like streams without the overhead.  `IndexRange` has something similar.  Between these and maybe introducing eg `applyToList(List<E>, EToDoubleFunction)` I'd have to imagine we can have both speed and elegance.

PR #2242, opened 2016-11-01
Close readers in ReferenceFileSource

droazen:
This class was leaking file handles badly on Spark


droazen:
@LeeTL1220 or @lbergelson please review

lbergelson:
We should add a unit test for getAllReferenceBases()

Issue #2177, opened 2016-09-19
Request: improvements in CreateSequenceDictionary


magicDGS:
`CreateSequenceDictionary` have several things that could be improved:
- The ouptut is necessary, but a default value could be created within the tool. Because the framework reads the sequence dictionary from the same path and a default extension (**.dict**), it would be nice to allow the creation without specifying the output file and create it according to the convention.
- An option to create the **.fai** index could be included. The `samtools faidx` indexing is well [documented](http://www.htslib.org/doc/faidx.html) and it looks simple to implement. I don't know if this should be implemented here or in htsjdk.

I can contribute with both if the GATK developers think that this will improve the tool. I think that the first point is good because is fairly simple and consistent with the framework. The second could be an improvement to be able to create the two required indexes using the same bundled tool.

droazen:
Closing as obsolete

magicDGS:
It is not really obsolete, but I will port the issue to Picard/HTSJDK. Create all indexes needed for GATK with a single tool will be nice, including the `.fai` index and the imageIndexFile for bwa (with the same approach as in #2452, or even better if https://github.com/broadinstitute/gatk-bwamem-jni/issues/4 is implemented).

magicDGS:
Issue opened in HTSJD: https://github.com/samtools/htsjdk/issues/830

droazen:
We've removed all issues related to Picard tools from our issue tracker -- this is what is meant by "obsolete". Of course, you're free to port any of these issues to the Picard/htsjdk issue tracker.

magicDGS:
Thanks for the feedback, @droazen!

PR #2243, opened 2016-11-01
CreateSequenceDictionary default output file name

magicDGS:
This solves partially #2177, although generating the _.fai_ is not implemented yet.

* Use `IOUtil.DICT_FILE_EXTENSION` as dictionary extension in `CachingIndexedFastaSequenceFile.checkAndCreate`
* No required output argument for `CreateSequenceDictionary`, using as default the name that is used internally in the GATK framework (e.g., __basic.dict__ for __basic.fa__).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.hellbender.tools.picard.sam;
 
+import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.*;
 import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.reference.ReferenceSequenceFile;
@@ -67,14 +68,22 @@ public final class CreateSequenceDictionary extends PicardCommandLineProgram {
             URI = "file:" + REFERENCE_SEQUENCE.getAbsolutePath();
         }
         if (OUTPUT == null) {
-            // determine the name for the dict file in the same way as CachingIndexedFastaSequenceFile.checkAndCreate
-            final String name = REFERENCE_SEQUENCE.getName();
-            final String fastaExt = ReferenceSequenceFileFactory.FASTA_EXTENSIONS.stream()
-                    .filter(name::endsWith).findFirst().orElseGet(() -> "");
-            OUTPUT = new File(REFERENCE_SEQUENCE.getParentFile(),
-                    REFERENCE_SEQUENCE.getName().replace(fastaExt, IOUtil.DICT_FILE_EXTENSION));
+            if (OUTPUT == null) {
+                // TODO: use the htsjdk method implemented in https://github.com/samtools/htsjdk/pull/774
+                OUTPUT = getDefaultDictionaryForReferenceSequence(REFERENCE_SEQUENCE);
+                logger.info("Output dictionary will be written in ", OUTPUT);
+            }
         }
         return null;
+    }
+
+    // TODO: this method will be in htsjdk (https://github.com/samtools/htsjdk/pull/774)
+    private static File getDefaultDictionaryForReferenceSequence(final File fastaFile) {
+        final String name = fastaFile.getName();
+        final String extension = ReferenceSequenceFileFactory.FASTA_EXTENSIONS.stream().filter(name::endsWith).findFirst()
+                .orElseGet(() -> {throw new IllegalArgumentException("File is not a supported reference file type: " + fastaFile.getAbsolutePath());});
+        final int extensionIndex = name.length() - extension.length();
+        return new File(fastaFile.getParentFile(), name.substring(0, extensionIndex) + IOUtil.DICT_FILE_EXTENSION);
     }
 
     /**
--- a/src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
@@ -4,6 +4,7 @@ import htsjdk.samtools.*;
 import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.reference.ReferenceSequenceFile;
 import htsjdk.samtools.reference.ReferenceSequenceFileFactory;
+import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.StringUtil;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
@@ -36,7 +37,7 @@ public final class CreateSequenceDictionary extends PicardCommandLineProgram {
     @Argument(
 	    shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
             fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
-            doc = "Output a dict file containing only the sequence dictionary")
+            doc = "Output a dict file containing only the sequence dictionary. By default it will use the base name of the input reference with the .dict extension", optional = true)
     public File OUTPUT;
 
     @Argument(doc = "Put into AS field of sequence dictionary entry if supplied", optional = true)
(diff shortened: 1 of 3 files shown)


src/test/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionaryIntegrationTest.java
```diff
@@ -35,6 +35,22 @@ public void testBasic() throws Exception {
         IntegrationTestSpec.assertEqualTextFiles(outputDict, new File(BASIC_FASTA.getAbsolutePath() + ".dict"));
     }
 
+    @Test
+    public void testNoOutput() throws Exception {
+        final File expectedDict = new File(TEST_DATA_DIR, "basic.dict");
```
droazen: Should use `createTempFile()` here

src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
```diff
@@ -63,6 +66,11 @@
         if (URI == null) {
             URI = "file:" + REFERENCE_SEQUENCE.getAbsolutePath();
         }
+        if (OUTPUT == null) {
+            // determine the name for the dict file in the same way as CachingIndexedFastaSequenceFile.checkAndCreate
+            final String fastaExt = REFERENCE_SEQUENCE.getAbsolutePath().endsWith("fa") ? "\\.fa$" : "\\.fasta$";
```
cmnbroad: @magic Can you use FilenameUtils.getBaseName from apache commons-io here instead?

src/test/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionaryIntegrationTest.java
```diff
@@ -35,6 +35,22 @@ public void testBasic() throws Exception {
         IntegrationTestSpec.assertEqualTextFiles(outputDict, new File(BASIC_FASTA.getAbsolutePath() + ".dict"));
     }
 
+    @Test
+    public void testNoOutput() throws Exception {
```
cmnbroad: Can you change the name of this method to something that reflects the fact that its using default output ? "testNoOutput" sounds like it does something completly different..

src/test/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionaryIntegrationTest.java
magicDGS: This is not an output file to create, but the expected output file for the tool. It is exists under `TEST_DATA_DIR`.

src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
magicDGS: I used the same code as in `CachingIndexedFastaSequenceFile.checkAndCreate()` to maintain the behaviour of that class. In addition, if I will port this to Picard it does not have (I guess) teh commons-io dependency.

Should I change it anyway?

src/test/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionaryIntegrationTest.java
magicDGS: Changed to "testDefaultOutputFile".

magicDGS:
Addressed some of the comments. Waiting for the answer about the usage of apache commons-io. Back to you @cmnbroad.

src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
magicDGS: While I was doing the same in Picard, I realized that I could support all the `FASTA_EXTENSIONS` instead of only "fa" and "fasta". The last commit contains this solution.

magicDGS:
@cmnbroad, now this is prepared. I'm not using commons-io but the all `FASTA_EXTENSIONS` for detecting what to modify in the name.

src/test/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionaryIntegrationTest.java
droazen: Ah, I see -- ideally we shouldn't be creating temporary files within `src/test/resources` during tests. Might be preferable to copy the reference to a temporary location as returned by `createTempFile()`, and run the test from there.

magicDGS: Done

magicDGS:
Because the Picard implementation is accepted already, last changes include the current implementation.

magicDGS:
Can you have a look to this, @cmnbroad? This is a very simple PR and it is already reviewed in Picard...

src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
```diff
@@ -63,9 +67,25 @@
         if (URI == null) {
             URI = "file:" + REFERENCE_SEQUENCE.getAbsolutePath();
         }
+        if (OUTPUT == null) {
+            if (OUTPUT == null) {
```
lbergelson: this seems redundant

src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
```diff
@@ -63,9 +67,25 @@
...
+                // TODO: use the htsjdk method implemented in https://github.com/samtools/htsjdk/pull/774
+                OUTPUT = getDefaultDictionaryForReferenceSequence(REFERENCE_SEQUENCE);
```
lbergelson: Can you update to using the htsjdk method now?

src/main/java/org/broadinstitute/hellbender/utils/fasta/CachingIndexedFastaSequenceFile.java
```diff
@@ -159,7 +160,7 @@ public static CachingIndexedFastaSequenceFile checkAndCreate(final File fastaFil
 
         // determine the name for the dict file
         final String fastaExt = fastaFile.getAbsolutePath().endsWith("fa") ? "\\.fa$" : "\\.fasta$";
-        final File dictFile = new File(fastaFile.getAbsolutePath().replaceAll(fastaExt, ".dict"));
+        final File dictFile = new File(fastaFile.getAbsolutePath().replaceAll(fastaExt, IOUtil.DICT_FILE_EXTENSION));
```
lbergelson: Is replaceAll really what you want?  Don't you want to just replace the last one?

lbergelson: oh, never mind, it only matches the end.

lbergelson: Can you use IOUtils.replaceExtension instead?

lbergelson:
@magicDGS very simple comments 👍

src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
magicDGS: Requires an htsjdk release.

src/main/java/org/broadinstitute/hellbender/tools/picard/sam/CreateSequenceDictionary.java
magicDGS: I don't know how this ends here. Fixed.

src/main/java/org/broadinstitute/hellbender/utils/fasta/CachingIndexedFastaSequenceFile.java
magicDGS: I just kept the previous behaviour, but using the `IOUtil.DICT_FILE_EXTENSION` for keeping in sync with htsjdk.

Nevertheless, I added a TODO to use the htsjdk method to find the default dictionary name for the reference. Than may be simpler, to keep also in sync with `CreateSequenceDictionary`.

magicDGS:
Back to you @lbergelson. I think that we can wait for the htsjdk release for refactoring `CachingIndexedFastaSequenceFile`.

lbergelson:
👍

PR #2246, opened 2016-11-02
Main improvements for custom toolkits

magicDGS:
* No change in the actual behaviour of GATK
* Several protected methods for customize some parameters of the toolkit:
    - `String getCommandLineName()`: returns the name from the command line toolkit
    - `void handleResult(Object)`: handle the output of the tool
* `main(String[] args)` code moved to a non-static final method (`mainEntry(String[])`) called inside the static one. This allow that the changes from override the customization methods could be apply with custom instances.

With this changes a custom main could be simplify a lot. As an example:

```java
public class Main extends org.broadinstitute.hellbender.Main {

    @Override
    protected String getCommandLineName() { return "MyCustomName"; }

    @Override
    protected List<String> getPackageList() {
        return Arrays.asList("org.custom.tools");
    }

    @Override
    protected List<Class<? extends CommandLineProgram>> getClassList() {
        return Arrays.asList(CreateSequenceDictionary.class, IndexFeatureFile.class);
    }

    public static void main(final String[] args) {
        new Main().mainEntry(args);
    }

}
```

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -16,9 +16,15 @@ import java.util.*;
  *
  * CommandLinePrograms are listed in a single command line interface based on the java package specified to instanceMain.
  *
- * If you want your own single command line program, extend this class and override {@link #getPackageList()}
- * to return a list of java packages in which to search for classes that extend CommandLineProgram
- * and/or {@link #getClassList()} to return a list of classes to include.
+ * If you want your own single command line program, extend this class and override if required:
+ *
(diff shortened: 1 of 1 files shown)


magicDGS:
Can you have a look to this @lbergelson? I'm asking you because you reviewed a previous PR to improve the usage of `Main` (like #2204 and #2226). Thanks in advance!

lbergelson:
@magicDGS This is a big improvement. Thank you for being the alpha tester for all of this.

magicDGS:
Thanks to you for reviewing my changes and include them into the framework!

PR #2247, opened 2016-11-02
Update htsjdk to 2.7.0

magicDGS:
* Update htsjdk to 2.7.0
* Remove usage of deprecated `SAMRecordUtil`
* Remove usage of deprecated `IndexFactory.writeIndex(index, indexFile)`
* Cleaning `IndexFeatureFile`

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/picard/sam/RevertSam.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/picard/sam/RevertSam.java
@@ -330,7 +330,7 @@ public final class RevertSam extends PicardCommandLineProgram {
 
         if (REMOVE_ALIGNMENT_INFORMATION) {
             if (rec.getReadNegativeStrandFlag()) {
-                rec.reverseComplement();
+                rec.reverseComplement(true);
                 rec.setReadNegativeStrandFlag(false);
             }
 
--- a/src/main/java/org/broadinstitute/hellbender/tools/picard/sam/RevertSam.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/picard/sam/RevertSam.java
@@ -6,7 +6,6 @@ import htsjdk.samtools.SAMFileWriter;
 import htsjdk.samtools.SAMReadGroupRecord;
 import htsjdk.samtools.SAMRecord;
 import htsjdk.samtools.SAMRecordQueryNameComparator;
-import htsjdk.samtools.SAMRecordUtil;
 import htsjdk.samtools.SAMTag;
 import htsjdk.samtools.SamReader;
 import htsjdk.samtools.SamReaderFactory;
@@ -331,7 +330,7 @@ public final class RevertSam extends PicardCommandLineProgram {
 
         if (REMOVE_ALIGNMENT_INFORMATION) {
             if (rec.getReadNegativeStrandFlag()) {
-                SAMRecordUtil.reverseComplement(rec);
+                rec.reverseComplement();
                 rec.setReadNegativeStrandFlag(false);
             }
 
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ repositories {
 
 }
 
-final htsjdkVersion = System.getProperty('htsjdk.version','2.6.1')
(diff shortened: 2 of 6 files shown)


lbergelson:
@droazen @magicDGS There's a test failure because the ADAM project relies on deprecated code in HTSJDK that was removed in 2.7.0.  We'll either need to disable the ADAM functionality or get a patched version of ADAM.  I have a [patch for ADAM](https://github.com/bigdatagenomics/adam/pull/1235), but I don't think they release very often.  We could host a patched artifact until then.

lbergelson:
@meganshand I wasn't sure if you saw this branch here. I know you had a similar one in progress.

src/main/java/org/broadinstitute/hellbender/tools/picard/sam/RevertSam.java
```diff
@@ -331,7 +330,7 @@ public void revertSamRecord(final SAMRecord rec) {
 
         if (REMOVE_ALIGNMENT_INFORMATION) {
             if (rec.getReadNegativeStrandFlag()) {
-                SAMRecordUtil.reverseComplement(rec);
+                rec.reverseComplement();
```
meganshand: In order to retain the same behavior here this should be rec.reverseComplement(true) so that it will make the changes in place instead of making a copy. Same thing in AbstractAlignmentMerger.

meganshand:
Thanks @lbergelson and @magicDGS, this looks far more thorough than I would have been :)

magicDGS:
Thanks for the review! @meganshand I adressed your comments, I saw that the patch for ADAM was accepted, @lbergelson, so I don't know if they will release soon a minor version for the bug fix. What should I do, @droazen?

droazen:
@magicDGS @lbergelson I vote that we temporarily disable ADAM support (and the associated test), as we shouldn't let this issue block our ability to update to a newer htsjdk.

droazen:
Changes here look good, let's merge after addressing the failing ADAM tests.

magicDGS:
How should I disable the ADAM support @droazen? Just disabling the ADAM tests? Or something else should be changed?

magicDGS:
Sorry, I overlooked the tests failing with `IndexFeatureFile`. Now the checking of a correct index name is made before the creation of the index itself (early failure) and some tests with a block-gzipped extension were failing. I think that now they are solved, @droazen.

droazen:
@magicDGS Was discussing this today with @lbergelson, and Louis thinks that we should temporarily depend on a snapshot of ADAM to work around the problem, rather than disabling ADAM support. The one disadvantage of that approach is that we'd be prevented from releasing GATK to maven central until we move off of the snapshot. I recommend that we reach out to the ADAM project to try to find out when their next release might be, so that we have a sense of the time scale involved.

magicDGS:
Ok,could we ask for a minor release? I will wait for your decision to continue my work on this PR...

droazen:
@lbergelson I think there might be another option we hadn't considered: couldn't we just suppress the deprecation warning for now using `@SuppressWarnings`?

lbergelson:
@droazen No, it's not a deprecation warning,  ADAM was actually using the deleted classes.

lbergelson:
It sounds like they're planning a release at the end of december, and would be able to do a point release for us if we needed it.  My vote is we put up a snapshot for now and ask them to do a november point release if they're able.  

https://github.com/bigdatagenomics/adam/issues/1248

lbergelson:
I misread what they said, I blame sleep deprivation.  They're releasing THIS month.  Probably on the 15th, so I think if we can wait until then to incorporate these changes we can avoid any problems.  Are other PR's blocked on this one?

magicDGS:
I have a couple of things that I want to use from htsjdk 2.7.0, but I can wait til the 15th (and actually I can disable in my work the ADAM support).

meganshand:
#2133 is also blocked by this PR, but I have no problem with waiting.

droazen:
Merging this now despite the ADAM situation, as we need to update to htsjdk 2.8.0 for a critical NIO bug fix.

PR #2248, opened 2016-11-02
updating gcloud command on travis

lbergelson:
they seem to have changed the gcloud cli to use --key-file instead of
--key in authentication

this was causing travis to fail on the cloud tests


droazen:
:+1: merge when ready

Issue #2245, opened 2016-11-02
FragmentCollection.create from a multi-sample ReadPileup throws an exception


magicDGS:
While using the implemented `ReadPileup.fixOverlaps()` with an input file for a `LocusWalker` which includes multiple samples, it throws the following exception:

```
IllegalArgumentException: FragmentUtils.create assumes that the incoming objects are ordered by SAMRecord alignment start, but saw a read FCC577TACXX:2:1102:5313:100586#CAGATCAT with alignment start 901 before the previous start 1000
```

This is because the `FragmentCollection.create` assumes ordered reads, but the iterator over a `ReadPileup` iterates over samples instead of position.

magicDGS:
This is already fixed in #2249, although it is sorting every time. Closing in favor of #2309 for efficiency related with ReadPileup.

PR #2249, opened 2016-11-02
Fixing ReadPileup constructor for multiple samples

magicDGS:
`FragmentCollection.create(ReadPileup)` assumes a `ReadPileup` sorted by start position and thus it throws an exception (#2245) while trying to fix a multi-sample `ReadPileup` constructed with an stratified one (like in LIBS).

For solving the issue, I just implemented a simple fix from the constructor (sorting the stratified pielup) and a simple test for check if #2245 is solved.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
@@ -44,7 +44,9 @@ public final class ReadPileup implements Iterable<PileupElement> {
      * Note: the current implementation of ReadPileup does not efficiently retrieve the stratified pileup
      */
     public ReadPileup(final Locatable loc, final Map<String, ReadPileup> stratifiedPileup) {
-        this(loc, stratifiedPileup.values().stream().flatMap(ReadPileup::getElementStream).collect(Collectors.toList()));
+        this(loc, stratifiedPileup.values().stream().flatMap(ReadPileup::getElementStream)
+                .sorted((l, r) -> Integer.compare(l.getRead().getStart(), r.getRead().getStart()))
+                .collect(Collectors.toList()));
     }
 
     /**
--- a/src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
@@ -44,9 +44,7 @@ public final class ReadPileup implements Iterable<PileupElement> {
      * Note: the current implementation of ReadPileup does not efficiently retrieve the stratified pileup
      */
     public ReadPileup(final Locatable loc, final Map<String, ReadPileup> stratifiedPileup) {
-        this(loc, stratifiedPileup.values().stream().flatMap(ReadPileup::getElementStream)
-                .sorted((l, r) -> Integer.compare(l.getRead().getStart(), r.getRead().getStart()))
-                .collect(Collectors.toList()));
+        this(loc, stratifiedPileup.values().stream().flatMap(ReadPileup::getElementStream).collect(Collectors.toList()));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -44,7 +44,9 @@ public ReadPileup(final Locatable loc, final List<PileupElement> pileup) {
      * Note: the current implementation of ReadPileup does not efficiently retrieve the stratified pileup
      */
     public ReadPileup(final Locatable loc, final Map<String, ReadPileup> stratifiedPileup) {
-        this(loc, stratifiedPileup.values().stream().flatMap(ReadPileup::getElementStream).collect(Collectors.toList()));
+        this(loc, stratifiedPileup.values().stream().flatMap(ReadPileup::getElementStream)
+                .sorted((l, r) -> Integer.compare(l.getRead().getStart(), r.getRead().getStart()))
+                .collect(Collectors.toList()));
```
droazen: Can we do the sort in `FragmentCollection.create()` rather than in the constructor for `ReadPileup`? Does any other code in `ReadPileup` require that the pileup elements be sorted by read start position? If not, I'd prefer not to pay the cost of sorting in the method where it's needed rather than for all `ReadPileups` constructed in this way.

src/test/java/org/broadinstitute/hellbender/utils/fragments/FragmentCollectionUnitTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private static final SAMFileHeader header = ArtificialReadUtils.createArtificialSamHeader();
+    private static final GATKRead read1 = ArtificialReadUtils.createArtificialRead(header, "10M");
+    private static final GATKRead read2 = ArtificialReadUtils.createArtificialRead(header, "10M");
+    static {
+        read1.setPosition(new SimpleInterval("22", 200, 210));
+        read2.setPosition(new SimpleInterval("22", 208, 218));
+        read1.setMatePosition(read2);
+        read2.setMatePosition(read1);
+    }
```
droazen: Can you make these local variables in your test case instead of static variables?

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
magicDGS: The point is that if it is already sorted, we are paying the cost of traversal in `FragmentCollection.create()`. Nevertheless, I'm working in a better way of handling the list of pileup elements (#2309).

I think that in the meanwhile will be better create a method for traversal of pileup elements in sort order, and use that one in the `FragmentCollection.create()` method.

src/test/java/org/broadinstitute/hellbender/utils/fragments/FragmentCollectionUnitTest.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
magicDGS: Done

magicDGS:
Thanks for the review. I rebased and made a small change. Back to you @droazen!

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -201,6 +201,15 @@ public ReadPileup getPileupForSample(final String sample, final SAMFileHeader he
     }
 
     /**
+     * Iterator over sorted by read start PileupElements.
+     */
+    public Iterator<PileupElement> sortedIterator() {
+        return getElementStream()
+                .sorted((l, r) -> Integer.compare(l.getRead().getStart(), r.getRead().getStart()))
+                .iterator();
+    }
```
droazen: Looks good now -- if you can just add a direct unit test in `ReadPileupUnitTest` for the new `sortedIterator()` method, then we can merge this.

magicDGS:
Added a very simple test for sorting iterator. Back to you @droazen.

PR #2257, opened 2016-11-08
removing GenomicsDB dependencies from travis.yml

lbergelson:
these should no longer be necessary and were causing problems because
one of the URL's no longer exists


lbergelson:
@droazen This fixes the build break on master.

Issue #1946, opened 2016-06-27
Connect HaplotypeCaller args to PairHMM


gspowley:
- Create `PairHMMNativeArguments` in HaplotypeCaller and pass to `VectorLoglessPairHMM` in `PairHMM.java`.
- Supply GATK temp directory in `VectorLoglessPairHMM.java`. Currently passing `null`, which uses the system temp directory.

```
final boolean isSupported = new IntelPairHmm().load(null);
```

gspowley:
@droazen @lbergelson 
We have a new version of GKL that resolves this issue and #1819 . I need to pass `PairHMMNativeArguments` from the HaplotypeCaller/MuTect2 command line to the GKL implementation.

I found the `pair_hmm_implementation` option in `UnifiedArgumentCollection` in gatk, but it looks like `HaplotypeCaller` uses `pair_hmm_implementation` in `LikelihoodEngineArgumentCollection` in gatk-protected. Is that right?

Can you provide some help/advice on how to hook up these new command line args?

droazen:
@gspowley This is still a TODO, correct?

gspowley:
Yes, @lbergelson and I worked on this a while ago. We need to revive the code.

gspowley:
@erniebrau please look at #2259 for the code changes required.

erniebrau:
I have made the changes to connect the PairHMM parameters [here](https://github.com/erniebrau/gatk-protected); it is based on [this branch](https://github.com/broadinstitute/gatk/tree/lb_connect_pairhmmargs), which @gspowley pointed me to. I am submitting a PR now...

gspowley:
Closed with https://github.com/broadinstitute/gatk/pull/2574 and https://github.com/broadinstitute/gatk-protected/pull/998.

PR #2259, opened 2016-11-10
Update to GKL 0.3.1

### LINKED ISSUE #1819: Ensure that multi-threaded AVX PairHMM falls back to single-threaded mode when OpenMP is not available or is the wrong version

gspowley:
Update to GKL 0.3.1 which provides the following features compared to GKL 0.1.2 currently used by GATK:

1. Improved performance for Level 1 compression.
1. OpenMP AVX PairHMM with fallback to non-OpenMP support logic in GATK.
1. Individual shared library objects for each native binding.
1. Flush-to-zero get/set support.

Updated `VectorPairHMMUnitTest` to test all supported implementations of `VectorLoglessPairHMM` and skip the test if no implementations are supported.

**Note**
The default number of threads used by OpenMP AVX PairHMM is set to 1, and the command line argument to change the number of threads is not connected yet (see #1946). So, the OpenMP and non-OpenMP PairHMM currently have the same performance.

Resolves #1819
First step for #1946

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
@@ -4,6 +4,7 @@ import com.google.common.annotations.VisibleForTesting;
 import htsjdk.variant.variantcontext.Allele;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.broadinstitute.gatk.nativebindings.pairhmm.PairHMMNativeArguments;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.MathUtils;
 import org.broadinstitute.hellbender.utils.Utils;
@@ -16,7 +17,7 @@ import java.io.Closeable;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Supplier;
+import java.util.function.Function;
 
 /**
  * Class for performing the pair HMM for local alignment. Figure 4.3 in Durbin 1998 book.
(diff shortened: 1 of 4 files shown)


lbergelson:
@gspowley https://github.com/broadinstitute/gatk/tree/lb_connect_pairhmmargs and https://github.com/broadinstitute/gatk-protected/tree/lb_connect_pairhmmargs

gspowley:
Thanks @lbergelson. I merged your changes from `lb_connect_pairhmmargs` into this PR.

This works with the `gatk-protected` changes in https://github.com/broadinstitute/gatk-protected/tree/lb_connect_pairhmmargs

Here's are idea of the performance improvement when running HaplotypeCaller (HC) on 4 CPUs.

PairHMM Threads | PairHMM Time (sec) | HC Time (sec) | HC Speedup
---------|------------------|-------------------------|-----------
1 | 50.7 | 96.6 | 1x
2 | 28.0 | 73.8 | 1.31x
3 | 21.6 | 67.2 | 1.44x
4 | 18.8 | 65.4 | 1.48x

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
```diff
@@ -26,24 +27,42 @@
...
     private HaplotypeDataHolder[] mHaplotypeDataArray;
 
-    public VectorLoglessPairHMM(PairHMMNativeArguments args) throws UserException.HardwareFeatureException {
-        // TODO: connect GATK temp directory
-        final boolean isSupported = new IntelPairHmm().load(null);
-
-        if (!isSupported) {
-            throw new UserException.HardwareFeatureException("Machine does not support AVX PairHMM.");
+    public VectorLoglessPairHMM(Implementation implementation,
+                                PairHMMNativeArguments pairHmmArgs) throws UserException.HardwareFeatureException {
```
lbergelson: Could you add a Utils.nonNull() check for implementation here.  Maybe don't allow null arguments here either?  I think it's not totally clear what should happen if you pass in null arguments. (I assume everything defaults to 0?)

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
```diff
@@ -26,24 +27,42 @@
...
+                pairHmm = new IntelPairHmm();
+                break;
+
+            case OMP:
+                isSupported = new IntelPairHmmOMP().load();
+                if (!isSupported) {
+                    throw new UserException.HardwareFeatureException("Machine does not support OpenMP AVX PairHMM.");
+                }
+                pairHmm = new IntelPairHmmOMP();
+                logger.info("Maximum number of OpenMP threads for AVX : " + pairHmmArgs.maxNumberOfThreads);
```
lbergelson: Won't this cause a null pointer if you pass in null arguments here?

src/test/java/org/broadinstitute/hellbender/utils/pairhmm/VectorPairHMMUnitTest.java
```diff
@@ -23,22 +24,24 @@
...
-
+    // Return a list of supported VectorLoglessPairHMM implementations, skip the test if none are supported
     private List<N2MemoryPairHMM> getHMMs() {
-        final PairHMMNativeArguments args = new PairHMMNativeArguments();
-        args.useDoublePrecision = false;
-        args.maxNumberOfThreads = 1;
-        final N2MemoryPairHMM avxPairHMM = new VectorLoglessPairHMM(args);
-        avxPairHMM.doNotUseTristateCorrection();
-        return Collections.singletonList(avxPairHMM);
+        List<N2MemoryPairHMM> list = new ArrayList<>();
```
lbergelson: I think this should be restructured so that skipped configurations always produce a skip.   
Right now if you skip some of the implementations than you don't know it because it never produces those tests.  You only get a skip if you can't run any of them.

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
lbergelson: Actually, I'm curious why this didn't cause an NPR in the tests.  From the line coverage it looks like `isSupported` was true, but then `new IntelPairHmmOMP()` threw an exception and it never reached the line below.  Is it possible that it could claim to load and then throw a hardware feature exception on instantiation?

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
```diff
@@ -26,24 +27,42 @@
...
-
-        if (!isSupported) {
-            throw new UserException.HardwareFeatureException("Machine does not support AVX PairHMM.");
+    public VectorLoglessPairHMM(Implementation implementation,
+                                PairHMMNativeArguments pairHmmArgs) throws UserException.HardwareFeatureException {
+        final boolean isSupported;
+
+        switch (implementation) {
+            case AVX:
+                isSupported = new IntelPairHmm().load();
```
lbergelson: Nitpicky, but I would generally avoid reusing `isSupported` in two cases like this since they really have nothing to do with each other.  Could you just declare it separately in each block, maybe give them distinct names, ie isAVXSupported and isOMPSupported.

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
```diff
@@ -33,36 +34,54 @@
 
     public enum Implementation {
         /* Very slow implementation which uses very accurate log10 sum functions. Only meant to be used as a reference test implementation */
-        EXACT(() -> {
+        EXACT(args -> {
             final Log10PairHMM hmm = new Log10PairHMM(true);
             logger.info("Using the non-hardware accelerated Java EXACT PairHMM implementation");
```
gspowley: done

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
```diff
@@ -33,36 +34,54 @@
...
             return hmm;
         }),
         /* PairHMM as implemented for the UnifiedGenotyper. Uses log10 sum functions accurate to only 1E-4 */
-        ORIGINAL(() -> {
```
gspowley: done

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
```diff
@@ -33,36 +34,54 @@
...
             logger.info("Using the AVX-accelerated native PairHMM implementation");
             return hmm;
         }),
-        /* Uses the fastest available PairHMM implementation (AVX if AVX is available, otherwise LOGLESS_CACHING */
-        FASTEST_AVAILABLE(() -> {
+        /* OpenMP Multi-threaded AVX implementation of LOGLESS_CACHING called through JNI. Throws if OpenMP AVX is not available */
+        AVX_LOGLESS_CACHING_OMP(args -> {
+            // Constructor will throw a UserException if OpenMP AVX is not available
+            final VectorLoglessPairHMM hmm = new VectorLoglessPairHMM(VectorLoglessPairHMM.Implementation.OMP, args);
+            logger.info("Using the OpenMP multi-threaded AVX-accelerated native PairHMM implementation");
```
gspowley: done

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
gspowley: Added nonNull checks. GKL uses 1 thread and single-precision if args is null, but I agree it's better to avoid null args.

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
gspowley: Yes, the null pointer in the unit test was an issue. The load passed, but trying to print the null arg threw an exception and skipped the test.

gspowley:
@lbergelson please let me know if we need any more changes.

src/test/java/org/broadinstitute/hellbender/utils/pairhmm/VectorPairHMMUnitTest.java
lbergelson: Isn't this going to skip all or none of them?  When this function is called it will either pass and return a list of Hmms or it will throw a skip exception.  I think it needs to be re-written as a dataprovider that provides each of `Implementation.values()`.  Then move the code to construct the hmm given an implementation into the test so that when it runs it generates one test for each implementation and then those tests are run or skipped individually.

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
```diff
@@ -26,24 +28,44 @@
...
+            case OMP:
+                final boolean isOmpSupported = new IntelPairHmmOMP().load();
+                if (!isOmpSupported) {
+                    throw new UserException.HardwareFeatureException("Machine does not support OpenMP AVX PairHMM.");
+                }
+                pairHmm = new IntelPairHmmOMP();
+                break;
+
+            default:
+                throw new UserException.HardwareFeatureException("Unknown PairHMM implementation.");
```
lbergelson: This should probably be a `GATKException` since it's a bug in the gatk rather than a hardware issue.  Also, could you include the implementation in the error message?

lbergelson:
@gspowley 2 minor comments, I think the test still has funny skipping behavior at the moment, and an exception nitpick that I missed before.

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
gspowley: done

lbergelson:
@gspowley 👍 Looks good to me now.  Ready to merge when tests pass.

PR #2260, opened 2016-11-11
Argument names for AddOrReplaceReadGroup to static

magicDGS:
I would like to keep in some of my tools the read group arguments in sync with the `AddOrReplaceReadGroup` in picard, but currently there is no way of access them.

This is a very simple and trivial patch to extract the short/long names to a static String variable to be able to use them. In addition, I refactored the variable names to the camel-case java convention.


magicDGS:
This is a very simple path, @lbergelson. Could you have a look?

lbergelson:
@magicDGS 👍 Looks reasonable to me.  

As a heads up, we're going to be deleting the gatk4 version of the picard tools in ~january and incorporating the picard versions directly as a library.  Which unfortunately means that modifications to them will have to be ported to the picard project directly.

magicDGS:
Thanks for the info, @lbergelson! I will contribute to Picard for this one too!

PR #2261, opened 2016-11-11
Allow custom exception handling to API users

magicDGS:
Simple patch to improve the `Main` usage in the same direction as previous PRs to finer control by API user:

* Added `handleNonUserException(final Exception exception)` to handle custom exceptions.


magicDGS:
Maybe this one is also for you @lbergelson. Sorry for split PRs in this way, but I implement them as far as I realize that there is some way of improving them for my toolkits.

Thanks for having a look to this PRs!

lbergelson:
@magicDGS 👍

PR #2262, opened 2016-11-14
remove bad test that was causing tests to abort

lbergelson:
removing MainTest.printReadsVersionTest because it was causing the test
suite to exit early.

it wasn't a very good test to begin with so it's not much of a loss


cmnbroad:
The change looks good once the tests run - for some (other) reason all of the travis builds are failing at the moment.

PR #2263, opened 2016-11-14
updating gatk-launch for spark 2.0 on gcloud

lbergelson:
2 changes to gatk-launch
- gcloud now requires arguments to the spark job to be separated from arguments to gcloud by `--`
- `--sparkMaster yarn-client` has been replaced with `--sparkMaster yarn --deploy-mode client`
  this only requires the sparkMaster to be changed in gatk-launch gcs because the client mode is set by gcloud


droazen:
:+1: merge when ready

PR #2271, opened 2016-11-17
making kryo respect our log level

lbergelson:
kryo uses com.esotericsoft.MinLog which is a yet another logger
making LoggingUtils.setLoggingLevel() set MinLog's level


magicDGS:
Another reason for the change to SLF4J (#2176), because there exists a [MINLOG-SLF4J](https://github.com/jdanbrown/minlog-slf4j) could be used to solve this.

droazen:
:+1: merge when ready

PR #2272, opened 2016-11-17
changing nCopies registration to use FieldSerializer

lbergelson:
this seems to fix an issue with serializing Genotypes in haplotype caller
updating kryo-serializers:0.37 -> 0.41 just in case

I couldn't come up with a test case that fails when run locally.  It only seems to fail when run with spark-submit.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/GATKRegistrator.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/GATKRegistrator.java
@@ -6,6 +6,10 @@ import com.esotericsoftware.kryo.serializers.JavaSerializer;
 import com.google.api.services.genomics.model.Read;
 import de.javakaffee.kryoserializers.UnmodifiableCollectionsSerializer;
 import htsjdk.samtools.SAMRecord;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/engine/spark/GATKRegistrator.java
```diff
@@ -52,6 +56,9 @@ public void registerClasses(Kryo kryo) {
         //register to avoid writing the full name of this class over and over
         kryo.register(PairedEnds.class, new FieldSerializer<>(kryo, PairedEnds.class));
 
+        kryo.register(Allele.class, new FieldSerializer<>(kryo, Allele.class));
+        kryo.register(FastGenotype.class, new FieldSerializer<>(kryo, FastGenotype.class));
+        kryo.register(VariantContext.class, new FieldSerializer<>(kryo, VariantContext.class));
```
droazen: If these 3 registrations are not strictly necessary, can we drop them? According to Kryo docs, `FieldSerializer` is already the default.

droazen:
One comment to address, then merge when ready.

Issue #2274, opened 2016-11-18
Collection arguments should replace the default value instead of appending


LeeTL1220:
```java

@Argument(
            doc = "PreAdapter Detail artifacts of interest on the forward strand.  'C/A' for a single artifact.  Separated by commas to assume multiple artifacts at the same time:  'C/A,T/G'  Artifacts must be one base to one base (e.g. 'CC/CA' is illegal).  C>A is OxoG.",
            shortName = ARTIFACT_MODES_SHORT_NAME, // -A
            fullName = ARTIFACT_MODES_FULL_NAME,
            optional = true
    )
    protected List<String> artifactModes = new ArrayList<>(Collections.singletonList("C/A"));

```
Using ``... -A C/T -A G/A ....``

I would expect to see a list ``["C/T", "G/A"]``

but instead the artifactModes argument is populated as follows:
```
0 = {String@2400} "C/A"
1 = {String@2401} "C/T"
2 = {String@2402} "G/A"
```

i.e. ``["C/A", "C/T", "G/A"]``

PR #2275, opened 2016-11-21
Change command line collection argument behavior from "append" to "replace". 

cmnbroad:
Also reject 'null' for collection arguments when used as one of several values for the same option.

Fixes https://github.com/broadinstitute/gatk/issues/2274.


droazen:
@lbergelson please review

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineParserTest.java
```diff
@@ -539,15 +539,14 @@ public void testClearDefaultValuesFromListArgumentAndAddNew() {
     }
 
     @Test
-    public void testDefaultValuesListArgument() {
```
cmnbroad: Not a bad idea - don't see one so I added it.

lbergelson:
@cmnbroad 👍

PR #2283, opened 2016-11-23
Extracting some args names in GATKTool to static values

magicDGS:
Very simple patch to extract arguments that may be useful for other toolkits into `StandardArgumentDefinitions` and some in-class changes.


lbergelson:
👍

PR #2285, opened 2016-11-29
Sh insdel call exp

SHuang-Broad:
A PR into a branch for myself to check test coverage (cannot be into the master because of the Spark bug).


SHuang-Broad:
Test coverage seem ok.

Issue #2276, opened 2016-11-21
VQSR sometimes outputs repeats of tranches


cmnbroad:
GATK4 in some cases outputs the first and last tranches twice in its tranche file and this leads to an extra line in the output file as well. This was reported as part of the tieout, i.e. for this test case:

gatk-launch VariantRecalibrator \
 --variant /seq/dax/ExAc2/Exome/v2/sites_only_combined/chr1.sites_only.unfiltered.vcf.gz \
 --resource mills,known=false,training=true,truth=true,prior=12:/seq/references/Homo_sapiens_assembly19/
 --resource axiom,known=false,training=true,truth=false,prior=10:/seq/references/Homo_sapiens_assembly19
 --resource dbsnp138,known=true,training=false,truth=false,prior=2:/seq/references/Homo_sapiens_assembly
 -an QD -an MQRankSum -an ReadPosRankSum -an FS -an SOR \
 -tranche 100.0 -tranche 99.95 -tranche 99.9 -tranche 99.8 -tranche 99.6 -tranche 99.5 -tranche 99.4 -tr
 -trustAllPolymorphic \
 --output results/recal4_indel_exome.vcf \
 -mode INDEL \
 -tranchesFile results/recal4_indel_exome.txt \
 --gatk3Compatibility

lucidtronix:
I think the tranches arguments got truncated in the command line above.  Here is one that causes the issue:

gatk-launch VariantRecalibrator \
--variant /seq/dax/ExAc2/Exome/v2/sites_only_combined/chr1.sites_only.unfiltered.vcf.gz \
--resource mills,known=false,training=true,truth=true,prior=12:/seq/references/Homo_sapiens_assembly19/v1/variant_calling/vqsr_resources/WGS/v2/Mills_and_1000G_gold_standard.indels.b37.vcf.gz \
--resource axiom,known=false,training=true,truth=false,prior=10:/seq/references/Homo_sapiens_assembly19/v1/variant_calling/vqsr_resources/WGS/v2/Axiom_Exome_Plus.genotypes.all_populations.poly.vcf.gz \
--resource dbsnp138,known=true,training=false,truth=false,prior=2:/seq/references/Homo_sapiens_assembly19/v1/variant_calling/vqsr_resources/WGS/v2/dbsnp_138.b37.vcf.gz \
-an QD -an MQRankSum -an ReadPosRankSum -an FS -an SOR \
[...]

PR #2286, opened 2016-11-30
Add a VQSR multi-tranche test.

cmnbroad:
Add a VQSR test that includes user-specified tranches (inspired by https://github.com/broadinstitute/gatk/issues/2276).


src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibratorIntegrationTest.java
```diff
@@ -98,7 +97,36 @@ public void testVariantRecalibratorSNP(final String[] params) throws IOException
...
+                        "-tranche", "99.5",
+                        "-tranche", "99.4",
+                        "-tranche", "99.3",
+                        "-tranche", "99.0",
+                        "-tranche", "98.0",
+                        "-tranche", "97.0",
+                        "-tranche", "90.0"
+                )
+        );
+        doSNPTest(args.toArray(new String[args.size()]), getLargeVQSRTestDataDir() + "expected/SNPWithTranches.txt");
```
droazen: The expected output here is named `SNPWithTranches.txt`, which is fine -- but the expected output for the test without explicit tranches is named `SNPTranches.txt`, which is confusing.

droazen:
One comment to address then merge when ready @cmnbroad

Issue #2292, opened 2016-12-05
Fisher Exact Test Edge Case


meganshand:
Fisher's Exact Test fails for the following two unit tests (that should be added to the data provider in `FisherExactTestUnitTest.java`):

```
tests.add(new Object[]{0, 0, 0, 3,                     1.0});
tests.add(new Object[]{9, 0, 0, 0,                     1.0});
```

R's implementation results in 1.0 for both of these tables, but the current GATK implementation results in 0.0 for both of them. They are examples of an edge case when the observed table is the only possibility given the marginal distributions (so the p-value should be 1).

lbergelson:
@meganshand Would you be willing to patch this one?

meganshand:
Yeah, I'll try to look at it on Friday.

lbergelson:
No rush!  Just wondering if you could take a look.

PR #2307, opened 2016-12-09
Handles edge case of Fisher Exact Test

meganshand:
Even though the Fisher Exact Test here is based on the R implementation, there was an edge case when the hypergeometric distribution we're getting the probabilities from only has one possible value. The difference in resulting p-value was coming from the fact that the apache Math3 HypergeometricDistribution behaves differently at this edge case than dhyper in R. 

Adresses #2292

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/FisherExactTest.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/FisherExactTest.java
@@ -4,15 +4,7 @@ import org.apache.commons.math3.distribution.AbstractIntegerDistribution;
 import org.apache.commons.math3.distribution.HypergeometricDistribution;
 
 import java.util.Arrays;
-import java.util.function.DoubleUnaryOperator;
 import java.util.stream.DoubleStream;
-import java.util.stream.IntStream;
-
-import static java.lang.Math.*;
-import static org.apache.commons.math3.util.MathArrays.ebeAdd;
-import static org.broadinstitute.hellbender.utils.MathUtils.arrayMax;
-import static org.broadinstitute.hellbender.utils.MathUtils.promote;
-import static org.broadinstitute.hellbender.utils.MathUtils.sum;
 
 /**
  * Implements the Fisher's exact test for 2x2 tables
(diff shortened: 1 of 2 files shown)


meganshand:
@lbergelson Can you please suggest a reviewer?

lbergelson:
@davidbenjamin Could you take a look at this one?

davidbenjamin:
Sure!

davidbenjamin:
@meganshand Thanks for making the code better!  If and only if you care to beautify it further while you're at it, I noticed a few other things:

* Using `addExact` instead of `+` is overkill, since the overflow it protects against can only occur if we have a read depth of 2,147,483,648.  In any case, it just throws an error.  If we wanted to be super-scrupulous, we would put an `if` statement in the `FisherExactTest` code to switch to an asymptotic approximation, but, like I said, overkill.

* This class has it's own `apply` method which replicates `MathUtils::applyToArray`.

* Similarly, it's `range` method should be deleted, and its use `support = range(lo, hi)` should become `support = new IndexRange(lo, hi)`.  Then `IntStream.of(support).mapToDouble(___).toArray()` becomes `range.mapToDouble(___)` and `apply(promote(support), ___)` becomes `support.mapToDouble(___)`.

* `final double relErr = 1 + pow(10, -7)` should become a `static` constant.

* The steps
```java
final double maxD1 = arrayMax(d1);
final double[] d2 = apply(apply(d1, d -> d - maxD1), d -> exp(d));
final double sumD2 = sum(d2);
return apply(d2,  d -> d/sumD2);
```
inside `dnHyper` are just a home-brewed way to normalize the log-space array `d1`.  Let's go throught the steps: 1) find the max.  2) subtract the max -- subtracting a constant in log-space is equivalent to dividing by a constant in real space, and since we're normalizing in the end this constant is arbitrary.  It's done for numerical stability.  3) exponentiate to get an unnormalized real-space array. 4) find the sum. 5) divide by the sum to get the normalized result.

The log-10 version of this is `MathUtils::normalizeFromLog10ToLinearSpace(d1)`.  You could either calculate `d1` in log-10 space or convert it, replacing the above line with `return MathUtils.normalizeFromLog10ToLinearSpace(MathUtils.applyToArrayInPlace(d1, MathUtils::logToLog10))`.  The latter option is simpler.

* Import static should be avoided except to escape horrible clutter, which is not the case here.

* The second argument of `Utils.validateArg(condition, calculated string expression)` should become `Utils.validateArg(condition, () -> calculated string expression)`.  In the first version, the expression is calculated *even if* the condition is satisfied, whereas in the second it is only calculated as needed.  It's not critical here but it's a good habit to get into.

PS I am a zealot of the Boy Scout Rule (always leave the camp site cleaner than you found it).  It is a great way to get familiar with a large code base and a great way to make the code more readable for the next person.

lh3:
I had a look at the source code of [HypergeometricDistribution](HypergeometricDistribution). If I am right, we are doing the following. We are invoking `logProbability()` for all possible `x[0][0]`. For a table with large numbers, we have to compute logBinomial for many iterations (see line 202–222 in the HypergeometricDistribution source code). Typically logBinomial calls logGamma and logGamma calls `log()` twice. This involves lots of computation and is not the fastest way to implement Fisher's exact test.

A faster way to implement the test takes the advantage of two observations. 1) When carrying the test, we are calling hypergeo(i,m+n,m,k), hypergeo(i+1,m+n,m,k), ... in order, and we can derive hypergeo(i+1,m+n,m,k) from hypergeo(i,m+n,m,k) by simply multiplying a number. This will be much faster than doing the full hypergeo->logBionomial->logGamma->log computation for each `i`. 2) For a large table, often when `i` is sufficiently smaller or larger than `x[0][0]`, the hypergeo probability is small enough to be ignored from the sum. It is not necessary to calculate hypergeo for the full range of `lo<=i<hi`. This trick can also dramatically reduce the number of iterations for large tables. htslib has a [exact test implementation](https://github.com/samtools/htslib/blob/bf753361dab9b1640cf64f7886dbfe35357a43c5/kfunc.c#L201) that considers the two observations.

I understand that the time spent on the `FisherExactTest` class probably won't show up at all in a profiler. I am not requesting to improve the implementation now. Just let you know the tricks. In addition, when we use this class for other purposes, a fast exact test may become a good thing.

meganshand:
@davidbenjamin I made the changes you requested, plus some additional cleanup. Since this function takes a `normalizedTable` it only ever actually sees tables whose sums are less than 400. The smallest p-value we'd expect given that we can't have entries that are larger than 400 is around 1e-120. Therefore we don't actually have to take the log of the probability and normalize it, we can just take the probability straight from `HypergeometricDistribution`. We also don't need `relErr`. 

Also given this I didn't make the changes @lh3 described, although this would clearly be a good way to reduce the computation needed for calculating the p-value with larger tables. 

If you think it would be useful to keep these numerical stability features, I can add them back in, but this removing them feels more readable to me given that we are only calculating small tables.

src/main/java/org/broadinstitute/hellbender/utils/FisherExactTest.java
```diff
@@ -21,59 +13,35 @@
...
-        final int[] support = range(lo, hi);
         final AbstractIntegerDistribution dist = new HypergeometricDistribution(null, m+n, m, k);
-        final double[] logdc = IntStream.of(support).mapToDouble(i -> dist.logProbability(i)).toArray();
-        final double oddsRatioAtNull = 1.0;
-        final double[] ds = dnHyper(oddsRatioAtNull, logdc, support);
-        final double relErr = 1 + pow(10, -7);
+        final double[] ds = support.mapToDouble(i -> dist.probability(i));
 
-        final double threshold= ds[x[0][0] - lo] * relErr;
+        final double threshold = ds[x[0][0] - lo];
```
davidbenjamin: @meganshand I read your comment about using normalized tables and while I agree with you I don't want this code to rely on that fact.  Fortunately, I think you can get numerical stability with negligible effort:

```java
final double[] logds = support.mapToDouble(i -> dist.logProbability(i));  // probability -> logProbability
final double threshold = ds[x[0][0] - lo];   // no change needed since log is monotonic
final double[] log10ds = DoubleStream.of(logds).filter(d -> d <= threshold).map(MathUtils::logToLog10).toArray();
final double pValue = MathUtils.sumLog10(log10ds);
```

meganshand:
Sorry, I didn't add the correct test. This one is closer to the edge case of when you have a large table that is still less than or equal to a total of 400. Back to you @davidbenjamin. Thanks!

meganshand:
Ah, I didn't hit refresh before fixing that test :) Thanks for the comments. I'll make those changes.

davidbenjamin:
@meganshand  back to you -- just a suggestion for getting numerical stability while keeping it readable.

meganshand:
Thanks @davidbenjamin, back to you!

davidbenjamin:
@meganshand Looks good!  You can squash and merge once tests pass.

lh3:
> Therefore we don't actually have to take the log of the probability and normalize it, we can just take the probability straight from HypergeometricDistribution.

The problem is mostly inside `HypergeometricDistribution`. A better implementation of this class should cache the last value, such that computing `hypergeo(i)` and then `hypergeo(i+1)` consecutively does not unnecessarily trigger full computation, which is quite expensive. Anyway, the current implementation is fine given that exact test is unlikely to be a bottleneck in its current use.

magicDGS:
Could I have some feedback about the efficiency of the Fisher's Exact Test implemented here? I'm planning to use it in other context where the performance could be reduced and I think that this is a good opportunity to have some information about it.

Thanks in advance!

meganshand:
@magicDGS I'm not sure what you're planning on using it for. This class is only tested for tables with small values (such that the sum of the table is less than 400). And as @lh3 said above, the way the HypergeometricDistribution is implemented it can be slow for tables with large values.

magicDGS:
Thanks for the information about this @meganshand. I will use this class in my use case and if it is not efficient enough I will try another solution.

Issue #2099, opened 2016-08-17
Remove -stand_emit_conf argument


ronlevine:
Port https://github.com/broadinstitute/gsa-unstable/pull/1454.

ronlevine:
We actually want to remove the argument, not deprecate it.

PR #2314, opened 2016-12-14
Remove -stand_emit_conf argument

ronlevine:
Implements https://github.com/broadinstitute/gatk/issues/2099.
Remove the `-stand_emit_conf` argument from `GenotypeCalculationArgumentCollection.java` and set as a static value in `GenotypingEngine.java`

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeCalculationArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeCalculationArgumentCollection.java
@@ -31,7 +31,6 @@ public final class GenotypeCalculationArgumentCollection implements Serializable
         this.snpHeterozygosity = other.snpHeterozygosity;
         this.indelHeterozygosity = other.indelHeterozygosity;
         this.STANDARD_CONFIDENCE_FOR_CALLING = other.STANDARD_CONFIDENCE_FOR_CALLING;
-        this.STANDARD_CONFIDENCE_FOR_EMITTING = other.STANDARD_CONFIDENCE_FOR_EMITTING;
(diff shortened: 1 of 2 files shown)


ronlevine:
@vdauwera Please review.

vdauwera:
Actually we don't just want to remove the argument, we want to rip out the STANDARD_CONFIDENCE_FOR_EMITTING  parameter entirely. We no longer need/want to be able to differentiate emit vs call.

droazen:
@davidbenjamin Can you review this one?

droazen:
@vdauwera That sounds like a pretty risky change to be making right when we're trying to tie-out our version of HaplotypeCaller. Perhaps this would be better done incrementally.

ronlevine:
@vdauwera Please take another look. Let me know if this implements the desired behavior.

vdauwera:
@droazen We can keep this on ice until after the tie outs are complete, it's not a problem. 

@ronlevine This looks good to me but we'll hold off of merging until @droazen gives us the all-clear.

droazen:
Just rebased this branch. Turns out we do need to merge it now in order to tie out against GATK 3.7

droazen:
Found a serious bug in this PR: `passesEmitThreshold()` was calling `passesCallThreshold(configuration.genotypeArgs.STANDARD_CONFIDENCE_FOR_CALLING)` instead of `passesCallThreshold(conf)`, which caused `STANDARD_CONFIDENCE_FOR_CALLING` to get compared against itself! Pushing a fix for this now.

ronlevine:
Just confirmed that the `passesEmitThreshold` and `passesCallThreshold` are implemented correctly in GATK3.

lbergelson:
@droazen Lets change the default to 10 and merge this.

droazen:
@lbergelson We'll change the default to 10 in a separate PR, since that change could wreak havoc in the tests.

Issue #2147, opened 2016-09-06
Make ReadPosRankSumTest.isUsableRead() account for deletions


ronlevine:
Port https://github.com/broadinstitute/gsa-unstable/pull/1467

ronlevine:
Closing via #2318.

PR #2318, opened 2016-12-15
Make ReadPosRankSumTest.isUsableRead() account for deletions

ronlevine:
Implement https://github.com/broadinstitute/gatk/issues/2147.

```
Change ReadPosRankSumTest.isUsableRead to take deletions into account. Previously, reads were skipped because the variant location was considered downstream from the read. 
```
The same fix was done to `AS_ReadPosRankSumTest`.

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReadPosRankSumTestUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReadPosRankSumTestUnitTest.java
@@ -8,6 +8,7 @@ import org.broadinstitute.hellbender.engine.ReferenceContext;
 import org.broadinstitute.hellbender.tools.walkers.annotator.allelespecific.AS_RankSumTest;
 import org.broadinstitute.hellbender.tools.walkers.annotator.allelespecific.AS_ReadPosRankSumTest;
 import org.broadinstitute.hellbender.utils.MannWhitneyU;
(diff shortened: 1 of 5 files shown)


ronlevine:
@ldgauthier Please review.

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReadPosRankSumTestUnitTest.java
```diff
@@ -158,4 +148,28 @@ public void testReadPos_Raw(){
         Assert.assertTrue(annotatePastEndRaw.isEmpty());
         Assert.assertTrue(annotatePastEndNonRaw.isEmpty());
     }
+
+    @DataProvider(name = "dataIsUsableRead")
+    private Object[][] dataIsUsableRead(){
+        return new Object[][]{
+                {"20M6D2M", 10, 1, 0, true},
+                {"20M6D2M", 10, 1, 29, false},
```
ldgauthier: Can you add one more test case where the read is usable to cover a position after the deletion?  I think  {"20M6D2M", 10, 1, 27, true} would do it, right?  That should look like our test case from GATK3:
image

ronlevine: Done.

ronlevine:
@ldgauthier If this is good to go, can you approve it.

PR #2258, opened 2016-11-09
Refactor `CallVariantsFromAlignedContigsSpark` to prep for calling SV.INS & SV.DEL

SHuang-Broad:
Refactoring of the structs and utilities involved calling simple inversions.
This helps preparing for calling simple insertions and deletions in SV.
Most changes are simple changes, no changes are made to the algorithm itself.
A simple fix of orginal code in `AlignmentRegion` was put in.
Major re-engineering of `getVariantContextForBreakpointAlleleAlignmentList()` in caller was done and explained in the temporary comment that will be removed after review is done.

@cwhelan would you please review? Thanks!

PR #2320, opened 2016-12-18
Initial implementation of Insertion/Deletion caller in SV pipeline.

SHuang-Broad:
What this PR does: 
* implemented alignment breaker based on CIGAR gaps; 
* calls insertion, deletion and naive tandem repeat annotation in addition to inversion; 
* some further refactoring of code related to sv caller; 
* tests updated accordingly

The PR is into another branch but NOT `master` because it is based on a continuation effort, which is already reviewed in #2258 , therefore has all the changes already made there (a little Spark 2 phobia caused this derail; once changes are reviewed, will see if tool is runnable and "pipelineable" with previous stages in the SV pipeline under the new Spark version).

@cwhelan please review.
Thanks!


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+                    final String assemblyId = assemblyIdAndContigsCollection._1;
+                    final ContigsCollection contigsCollection = assemblyIdAndContigsCollection._2;
+                    return contigsCollection.getContents().stream()
+                            .map(pair -> new Tuple2<>(new Tuple2<>(assemblyId, pair._1.toString()), pair._2.toString().getBytes()))
+                            .collect(Collectors.toList());
+                });
+    }
+
+    // TODO: 11/26/16 assuming 'I'/'D' cannot be the beginning or last CIGAR operator AND the following operator cannot be a clipping
+    /**
```
SHuang-Broad: While reviewing some of the calls, there was a region that seem to invalidate this assumption, the particular region is `chr21:44,372,670-44,372,939` in the benchmarking BAM we are using.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
```diff
@@ -14,76 +14,92 @@
...
 /**
  * A wrapper that wraps around the {@link AlnRgn} type returned by jBWA to represent one (out of the potentially multiple)
  * alignment record(s) of an locally-assembled contig.
  */
 @DefaultSerializer(AlignmentRegion.Serializer.class)
 class AlignmentRegion {
 
-    final String contigId;
+    static final String STRING_REP_SEPARATOR= "\t";
+    static final String DUMMY_ASM_ID = "ASSEMBLY";
```
SHuang-Broad: The reason was that
1. IDE keeps reminding me accessing the `assemblyId` field may produce NPE (though in practice it will not because the only case null assembly id was generated was with the `CallVariantsFromAlignedContigsSAMSpark`)
2. Various other accessors in this class `toString()` etc. will produce NPE if we are not careful.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
```diff
@@ -14,76 +14,92 @@
...
     public AlignmentRegion(final GATKRead read) {
-        this.assemblyId = null;
+        this.assemblyId = DUMMY_ASM_ID;
         this.contigId = read.getName();
+        this.referenceInterval = new SimpleInterval(read);
         this.forwardStrand = ! read.isReverseStrand();
         this.forwardStrandCigar = forwardStrand ? read.getCigar() : CigarUtils.invertCigar(read.getCigar());
-        this.referenceInterval = new SimpleInterval(read);
+        this.mapQual = read.getMappingQuality();
+        this.mismatches = read.hasAttribute("NM") ? read.getAttributeAsInteger("NM") : 0;
```
SHuang-Broad: Right, it was something I forgot to change. Will go with the special-value case.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
```diff
@@ -96,28 +112,37 @@ int endOfAlignmentInContig() {
         return assembledContigLength - SVVariantCallerUtils.getNumClippedBases(false, forwardStrandCigar);
     }
 
+    // TODO: 11/27/16 test
+    /**
+     * @return  A packed String representation of this AlignmentRegion, noticeably the fields are not separated by tab.
+     *          Note that the format is NOT the same as that used in {@link #toString()}.
```
SHuang-Broad: Nope. It was there and I didn't remove it, and called it while outputting debugging information. Will remove.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+        final int gapCount = (int) cigarElements.stream().filter(ce -> ce.getOperator().isIndel() && ce.getLength() >= sensitivity).count();
+        if (gapCount==0) return new ArrayList<>( Collections.singletonList(oneRegion) );
+
+        // get necessary information for computing
+        final String asmId = oneRegion.assemblyId;
+        final String contigId = oneRegion.contigId;
+        final String chr = oneRegion.referenceInterval.getContig();
+        final Cigar originalForwardStrandCigar = oneRegion.forwardStrandCigar;
+        final boolean isForwardStrand = oneRegion.forwardStrand;
+        final int newMapQual = oneRegion.mapQual;
```
SHuang-Broad: It was my undecidedness on what to do with mapping qualities when breaking the alignment. Technically I think (might be wrong) the mapping quality, since some `M` part of the original alignment is removed, is possibly affected, the mapping quality is not necessarily going to be the same as when the `M` part was there.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+        for(int i=0; i<cigarElements.size(); ++i){ // considers only 'M', 'I', 'D', 'S', and 'H'
+            final CigarElement current = cigarElements.get(i);
+            final CigarOperator op = current.getOperator();
+            final int operatorLen = current.getLength();
+            switch (op){
+                case M: case S: case H:
+                    cigarMemoryList.add(current);
+                    break;
+                case I: case D:
+                    if (operatorLen<sensitivity) {
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+import static org.broadinstitute.hellbender.tools.spark.sv.BreakpointPair.EvidenceFlankSide.*;
+
+/**
+ * This class represents inferred a pair of breakpoints of a simple SV event
+ * (a simple rearrangement between two genomic locations)
+ * that is suggested by the input {@link ChimericAlignment},
+ * and complications in pinning down the breakpoints's location to exact base pair resolution.
+ */
+@DefaultSerializer(BreakpointPair.Serializer.class)
+class BreakpointPair {
```
SHuang-Broad: Maybe we are defining breakpoints differently. I was referring to breakpoints as two genomic locations that are rearranged together in the sample that is different from the reference's arrangement, e.g. for simple inversion the left and right end of the inverted sequence. Hence for simple variants that we are currently dealing with, it always comes in pairs. For insertions, the `first` and `second` of the pair point to the same location.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+
+    static final String EMPTY_STRING = "";
```
SHuang-Broad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+    private static final int MAX_LOWER_CN = 10; // in case of CNV events involving tandem duplications, the maximum value of the lower copy number the algorithm will try
```
SHuang-Broad: updated comments

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+    /**
+     * Represents the strand of evidence that was used in computing this breakpoint pair.
+     */
+    enum EvidenceFlankSide {
+        BOTH, THREE_PRIME, FIVE_PRIME
+    }
+
+    /**
+     * A helper struct for annotation breakpoints represented by this BreakpointPair.
+     * To be updated as more types of complications can be mentally processed and
```
SHuang-Broad: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+     * handled by {@link #resolveComplications(ChimericAlignment)}.
+     */
+    @DefaultSerializer(BreakpointComplication.Serializer.class)
+    static final class BreakpointComplication {
```
SHuang-Broad: I was trying to be future-proof (maybe over-shooting), because I believe we might need to update this in more complicated calling.
But changed for now as you suggested.

SHuang-Broad: I was trying to be future-proof (maybe over-shooting), because I believe we might need to update this in more complicated calling, like in this case it is already incorporating the tandem duplication (albeit vey naively) annotations that must be included to tell downstream steps that it is not as simple as a simple insertion or deletion.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+        final String homologyForwardStrandRep;
+        final String insertedSequenceForwardStrandRep;
+        final String dupSeqForwardStrandRep;
+        final int refDup;
```
SHuang-Broad: Changed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+                breakpointComplication.serialize(kryo, output);
+            }
+
+            @Override
+            public BreakpointComplication read(final Kryo kryo, final Input input, final Class<BreakpointComplication> klass ) {
+                return new BreakpointComplication(kryo, input);
+            }
+        }
+    }
+
```
SHuang-Broad: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+            return BOTH;
+        } else {
+            return chimericAlignment.regionWithLowerCoordOnContig.forwardStrand ? FIVE_PRIME : THREE_PRIME;
+        }
+    }
+
+    // TODO: 12/12/16 the most complicated tandem duplication case is not exactly reproducible (e.g. '+' and '-' strand may give slightly different results by this treatment)
+    /**
+     * Given an ChimericAlignment representing two reference intervals rearranged as two intervals on the locally-assembled contig,
+     * identify potential complications such as homologyForwardStrandRep and duplication on the reference or the contig.
```
SHuang-Broad: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+     * Note: assuming input alignment regions are NOT completely enclosed in the other.
```
SHuang-Broad: Done, but the checking is in the constructor of `ChimericAlignment` now.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+        final byte[] contigSeq = chimericAlignment.contigSeq;
+
+        // a segment with lower coordinate on the locally-assembled contig could map to a higher reference coordinate region
+        // under two basic types of SV's: inversion (strand switch necessary) and translocation (no strand switch necessary)
+        if (chimericAlignment.strandSwitch!= ChimericAlignment.StrandSwitch.NO_SWITCH) { // the case involves an inversion
+            // TODO: 12/5/16 duplication detection to be done for inversion alleles
+            return new BreakpointComplication(getHomology(firstContigRegion, secondContigRegion, contigSeq), getInsertedSequence(firstContigRegion, secondContigRegion, contigSeq), EMPTY_STRING, 0, 0);
+        } else if (chimericAlignment.isNotSimpleTranslocation) { // no translocation
+
+            final int d1 = r2b - r1e - 1, // distance-1 between the two regions on reference
```
SHuang-Broad: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+                      d2 = c2b - c1e - 1; // distance-1 between the two regions on contig
+
+            String homologyForwardStrandRepresentation = EMPTY_STRING, insertedSeqForwardStrandRepresentation=EMPTY_STRING, dupSeqForwardStrandRepresentation=EMPTY_STRING;
+            int refDup=0, ctgDup=0;
+            // when d1<d2 (but not when both are negative), it always indicate an insertion, and vice versa
+            // when d2<0, there's always homologyForwardStrandRep
+            if (d1>0 && d2==0) {        // DEL: simple deletion, deleted sequence is [r1e+1, r2b-1] on the reference
+                homologyForwardStrandRepresentation    = EMPTY_STRING;
```
SHuang-Broad: The intent was to be more explicit in re-initializing (or rather assignment). Leaving them un-inilialized makes IDE unhappy.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+                insertedSeqForwardStrandRepresentation = EMPTY_STRING;
+                dupSeqForwardStrandRepresentation      = EMPTY_STRING;
+                refDup = ctgDup = 0;
+            } else if (d1==0 && d2>0) { // INS: simple insertion, inserted sequence is the sequence [c1e+1, c2b-1] on the contig
+                homologyForwardStrandRepresentation    = EMPTY_STRING;
+                insertedSeqForwardStrandRepresentation = getInsertedSequence(firstContigRegion, secondContigRegion, contigSeq);
+                dupSeqForwardStrandRepresentation      = EMPTY_STRING;
+                refDup = ctgDup = 0;
+            } else if (d1==0 && d2<0) { // DEL: tandem duplication on reference (CN = 2) but one copy was deleted on the contig; duplicated sequence on reference are [r1e-|d2|+1, r1e] and [r2b, r2b+|d2|-1]
```
SHuang-Broad: Done. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+                ctgDup      = 1;
+            } else if (d1<0 && d2==0) { // INS: tandem duplication of reference bases [r1e-|d1|+1, r1e] to contig bases [c1e-|d1|+1, c1e] and [c2b, c2b+|d1|-1]
+                final byte[] insertedSeq = Arrays.copyOfRange(contigSeq, c1e-Math.abs(d1), c1e);
+                if (!chimericAlignment.isForwardStrandRepresentation) SequenceUtil.reverseComplement(insertedSeq);
+                homologyForwardStrandRepresentation    = EMPTY_STRING;
+                insertedSeqForwardStrandRepresentation = EMPTY_STRING;
+                dupSeqForwardStrandRepresentation      = new String(insertedSeq);
+                refDup      = 1;
+                ctgDup      = 2;
+            } else if (d1<0 && d2>0) {  // INS: tandem duplication of reference bases [r1e-|d1|+1, r1e] to contig bases [c1e-|d1|+1, c1e] and [c2b, c2b+|d1|-1] and an inserted sequence [c1e+1, c2b-1]
```
SHuang-Broad: If I understand correctly, you mean the type of this variant should be `GATKSVVCFHeaderLines.SVTYPES.DUP` but not `GATKSVVCFHeaderLines.SVTYPES.INS`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+                final byte[] insertedSeq = Arrays.copyOfRange(contigSeq, c1e-Math.abs(d1), c1e);
+                if (!chimericAlignment.isForwardStrandRepresentation) SequenceUtil.reverseComplement(insertedSeq);
+                homologyForwardStrandRepresentation    = EMPTY_STRING;
+                insertedSeqForwardStrandRepresentation = getInsertedSequence(firstContigRegion, secondContigRegion, contigSeq); // note this does not incorporate the duplicated reference sequence
+                dupSeqForwardStrandRepresentation      = new String(insertedSeq);
+                refDup      = 1;
+                ctgDup      = 2;
+            } else if (d1>0 && d2>0) {  // DEL: (arbitrarily chosen SV type) with inserted sequence, i.e. large non-conserved substitution, reference bases [r1e+1, r2b-1] is substituted with contig bases [c1e+1, c2b-1]
```
SHuang-Broad: Done with updated comments. Thanks.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+                    SequenceUtil.reverseComplement(dupSeq);
+                }
+
+                homologyForwardStrandRepresentation    = getHomology(firstContigRegion, secondContigRegion, contigSeq);
+                dupSeqForwardStrandRepresentation      = new String(dupSeq);
+                insertedSeqForwardStrandRepresentation = EMPTY_STRING; //isExpansion ? StringUtils.repeat(dupSeqForwardStrandRepresentation, duplicationComplication[1]-duplicationComplication[0]) : EMPTY_STRING;
+                refDup      = duplicationComplication[isExpansion?0:1];
+                ctgDup      = duplicationComplication[isExpansion?1:0];
+            } else if (d1==0 && d2==0) {// SNP & indel
+                throw new IllegalStateException("Detected badly parsed chimeric alignment for identifying SV breakpoints; no rearrangement found: " + chimericAlignment.toString());
```
SHuang-Broad: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+     * attempts to find--naively and slowly--the copy numbers on the reference and the contig of tandem repeats,
+     * as well as the pseudo-homologyForwardStrandRep between the duplicated sequence and the right flanking region.
+     * The returned array of length 4 have entries in the following order:
+     *  estimate of copy number on the sequence where {@code overlapOnHigherCNSequence} was computed,
+     *  estimate of copy number on the sequence where {@code overlapOnLowerCNSequence} was computed,
+     *  duplicated unit length,
+     *  pseudo-homologyForwardStrandRep length
+     */
+    @VisibleForTesting
+    static int[] tandemDuplicationComplication(final int overlapOnHigherCNSequence, final int overlapOnLowerCNSequence) {
```
SHuang-Broad: I agree.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+
+        int higherCnEst=0, lowerCnEst=0, unitLen=0, pseudoHomLen=0;
```
SHuang-Broad: In the inner most loop in the if statement, `unitLen` which is the length of the repeated sequence, is updated.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+     * with higher reference coordinates.
+     */
+    @VisibleForTesting
+    static Tuple2<SimpleInterval, SimpleInterval> leftJustifyBreakpoints(final ChimericAlignment ca, final BreakpointComplication complication) {
+
+        final int homologyLen = complication.homologyForwardStrandRep.length();
+
+        final String leftBreakpointRefContig, rightBreakpointRefContig;
+        final int leftBreakpointCoord, rightBreakpointCoord;
+        if(complication.refDup!=0 && complication.contigDup!=0) { // todo : development artifact-- assuming tandem duplication is not co-existing with inversion
```
SHuang-Broad: Yes. If you look at how the complications are resolved in `resolveComplications()` the distance between alignment regions logic is NOT triggered if there's strand switch, which is a must for inversion. So in inversion calling we are assuming insertions and deletions are no co-existing (inversion are still going to be annotated with inserted sequence and homology).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+            leftBreakpointRefContig = rightBreakpointRefContig = contig;
+            final SimpleInterval leftReferenceInterval, rightReferenceInterval;
+            if (ca.isForwardStrandRepresentation) {
+                leftReferenceInterval = ca.regionWithLowerCoordOnContig.referenceInterval;
+                rightReferenceInterval = ca.regionWithHigherCoordOnContig.referenceInterval;
+            } else {
+                leftReferenceInterval = ca.regionWithHigherCoordOnContig.referenceInterval;
+                rightReferenceInterval = ca.regionWithLowerCoordOnContig.referenceInterval;
+            }
+            leftBreakpointCoord = leftReferenceInterval.getEnd() - homologyLen - Math.max(0, complication.contigDup-complication.refDup)*dupLen;
```
SHuang-Broad: Yes, let's talk in person as I spent sometime to figure this out and should explain to you in person.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ChimericAlignment.java
```diff
@@ -59,89 +56,231 @@
...
+        // fast forward to the first alignment region with high MapQ
+        AlignmentRegion current = iterator.next();
+        while (mapQualTooLow(current) && iterator.hasNext()) {
+            current = iterator.next();
+        }
+
+        // then iterate over the AR's in pair to identify CA's.
+        while ( iterator.hasNext() ) {
+            final AlignmentRegion next = iterator.next();
+            if (firstAlignmentIsTooShort(current, next, SVConstants.DEFAULT_MIN_ALIGNMENT_LENGTH)) {
```
SHuang-Broad: It is actually filtered out later (the code was there, I changed a bit of how the code was written in the refactor) in `nextAlignmentMayBeNovelInsertion()`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ChimericAlignment.java
```diff
@@ -59,89 +56,231 @@
...
-        this.contigId = contigId;
         this.regionWithLowerCoordOnContig = regionWithLowerCoordOnContig;
         this.regionWithHigherCoordOnContig = regionWithHigherCoordOnContig;
-        this.homology = homology;
-        this.insertedSequence = insertedSequence;
+
+        this.strandSwitch = determineStrandSwitch(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig);
+        this.involvesReferenceIntervalSwitch = involvesRefPositionSwitch(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig, samSequenceDictionary);
+        this.isForwardStrandRepresentation = isForwardStrandRepresentation(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig, this.strandSwitch, this.involvesReferenceIntervalSwitch);
+        this.isNotSimpleTranslocation = isNotSimpleTranslocation(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig, this.strandSwitch, this.involvesReferenceIntervalSwitch);
```
SHuang-Broad: Not quite understanding the concerns, maybe in person explanation will be easier.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ChimericAlignment.java
```diff
@@ -59,89 +56,231 @@
...
-     *  alignmentRegion2.toString()
-     *  insertedSequence
-     *  homology
-     *
+     * Determine if the region that maps to a lower coordinate on the contig also maps to a lower coordinate on the reference.
+     */
+    @VisibleForTesting
+    static boolean involvesRefPositionSwitch(final AlignmentRegion regionWithLowerCoordOnContig,
+                                             final AlignmentRegion regionWithHigherCoordOnContig,
+                                             final SAMSequenceDictionary samSequenceDictionary) {
```
SHuang-Broad: The no translocations restriction is handled later (currently in the code, if you think we can enforce it here, we can do it, but once we want to handle simple translocations, we must use it in someway, either the dictionary of `SVIntervals`), so that assumption is not guaranteed here.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/CallVariantsFromAlignedContigsSpark.java
```diff
@@ -36,7 +37,7 @@
 
     @Argument(doc = "output file name for called variants", shortName = "outputName",
             fullName = "outputName", optional = true)
-    private String outputName = SVConstants.INVERSIONS_OUTPUT_VCF;
+    private String outputName = SVConstants.INS_DEL_OUTPUT_VCF;
```
SHuang-Broad: I forget to update this name... Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ChimericAlignment.java
```diff
@@ -59,89 +56,231 @@
...
+    //                to get the 2nd case right, we need evidence flanking both sides of the inversion, and that could be difficult
+    /**
+     * Determine if the two regions indicate a inter-chromosomal translocation or intra-chromosomal translocation that
+     * DOES NOT involve a strand switch.
      */
-    protected ChimericAlignment fromString(String assembledBreakpointLine) {
-        final String[] fields = assembledBreakpointLine.split("\t");
-        return fromFields(fields);
+    @VisibleForTesting
+    static boolean isNotSimpleTranslocation(final AlignmentRegion regionWithLowerCoordOnContig,
```
SHuang-Broad: Roger. Thanks for pointing this out. I didn't think of this case.

SHuang-Broad: See ticket #2347

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/CallVariantsFromAlignedContigsSpark.java
```diff
@@ -61,65 +65,68 @@ public boolean requiresReference() {
...
+                .groupByKey()                                                                                               // 3. {consensus BP}
+                .map(tuple2 -> SVVariantConsensusCall.callVariantsFromConsensus(tuple2, broadcastReference));               // BP annotated with list of CA's
+    }
 
+    private static void debugStats(JavaPairRDD<Iterable<AlignmentRegion>, byte[]> alignmentRegionsWithContigSequences, final String outPrefix) {
+        log.info(alignmentRegionsWithContigSequences.count() + " contigs");
+        final long noARs = alignmentRegionsWithContigSequences.filter(tuple2 -> Iterables.size(tuple2._1())==0).count();
+        log.info(noARs + " contigs have no alignment regions");
+        final long oneARs = alignmentRegionsWithContigSequences.filter(tuple2 -> Iterables.size(tuple2._1())==1).count();
+        log.info(oneARs + " contigs have only one alignment regions");
```
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/CallVariantsFromAlignedContigsSpark.java
```diff
@@ -61,65 +65,68 @@ public boolean requiresReference() {
...
+
+        final JavaPairRDD<String, List<Tuple2<Integer, Integer>>> x = alignmentRegionsWithContigSequences.filter(tuple2 -> Iterables.size(tuple2._1())==2).mapToPair(tuple2 -> {
+            final Iterator<AlignmentRegion> it = tuple2._1().iterator();
+            final AlignmentRegion region1 = it.next(), region2 = it.next();
+            return new Tuple2<>(region1.assemblyId+":"+region1.contigId, Arrays.asList(new Tuple2<>(region1.mapQual, region1.referenceInterval.size()), new Tuple2<>(region2.mapQual, region2.referenceInterval.size())));
+        });
+        x.coalesce(1).saveAsTextFile(outPrefix+"_2AR_MQs");
```
SHuang-Broad: Updated with (maybe?) better names.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVConstants.java
```diff
@@ -16,12 +16,29 @@ private SVConstants() {}
...
      * Minimum flanking alignment length used in calling variants when going through contig alignments.
      */
     static final Integer DEFAULT_MIN_ALIGNMENT_LENGTH = 50;
 
-    static final String VARIANT_ID_FIELD_SEPARATER = ";";
+    static final String VARIANT_ID_FIELD_SEPARATOR = ";";
 
     static final int CHIMERIC_ALIGNMENTS_HIGHMQ_THRESHOLD = 60;
+
+    static final int GAPPED_ALIGNMENT_BREAK_DEFAULT_SENSITIVITY = 50; // alignment with gap of size >= 50 will be break apart.
```
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantConsensusCall.java
```diff
@@ -0,0 +1,242 @@
...
+
+        final Allele refAllele = Allele.create(new String(reference.getReferenceBases(null, new SimpleInterval(contig, start, start)).getBases()), true);
+
+        final Allele altAllele;
+        if (breakpointPair.evidenceFlankSide != BOTH) {
+            altAllele = Allele.create(SVConstants.VCF_ALT_ALLELE_STRING_INV);
+        } else {
+            if (start==end) { // expected to be insertion
+                if (breakpointPair.complication.insertedSequenceForwardStrandRep.isEmpty() && breakpointPair.complication.contigDup<breakpointPair.complication.refDup)
+                    throw new IllegalStateException("An identified breakpoint pair seem to suggest insertion but the inserted sequence is empty: " + breakpointPair.toString());
```
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantConsensusCall.java
```diff
@@ -0,0 +1,242 @@
...
+            throws IOException {
+
+        final String contig = breakpointPair.leftJustifiedLeftBreakpoint.getContig();
+        final int start = breakpointPair.leftJustifiedLeftBreakpoint.getStart();
+        final int end = breakpointPair.leftJustifiedRightBreakpoint.getStart();
+
+        final Allele refAllele = Allele.create(new String(reference.getReferenceBases(null, new SimpleInterval(contig, start, start)).getBases()), true);
+
+        final Allele altAllele;
+        if (breakpointPair.evidenceFlankSide != BOTH) {
```
SHuang-Broad: Let's talk in person before I put in changes so I don't do something worse...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantConsensusCall.java
```diff
@@ -0,0 +1,242 @@
...
+            attributeMap.put( (evidenceFlankSide == FIVE_PRIME) ? GATKSVVCFHeaderLines.INV_5_TO_3 : GATKSVVCFHeaderLines.INV_3_TO_5, "");
+        }
+        return attributeMap;
+    }
+
+    /**
+     * Not testing this because the complications are already tested in the BreakpointPair class' own test,
+     * more testing here would be actually testing VCBuilder.
+     */
+    private static Map<String, Object> getComplications(final BreakpointPair breakpointPair) {
```
SHuang-Broad: Replaced with `parseComplicationsAndMakeThemAttributeMap()`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
cwhelan: I just meant that the word 'tandem' as used in your comment doesn't really apply, I think, since there's an intervening inserted sequence if I'm understanding things correctly.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+import static org.broadinstitute.hellbender.tools.spark.sv.ContigsCollection.loadContigsCollectionKeyedByAssemblyId;
+
+/**
+ * Implements a parser for parsing alignments of locally assembled contigs, and
+ * generating {@link ChimericAlignment}'s from the alignments, if possible.
+ */
+class AssemblyAlignmentParser implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    static final int ARTIFICIAL_CIGAR_MISMATCH = -1;
```
SHuang-Broad: Will go with the special value here as well, just like the case in `AlignmentRegion`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+    @VisibleForTesting
+    static Iterable<AlignmentRegion> breakGappedAlignment(final AlignmentRegion oneRegion, final int sensitivity) {
+
+        final List<CigarElement> cigarElements = oneRegion.forwardStrandCigar.getCigarElements();
+        if(cigarElements.size()==1) return new ArrayList<>( Collections.singletonList(oneRegion) );
+        final int gapCount = (int) cigarElements.stream().filter(ce -> ce.getOperator().isIndel() && ce.getLength() >= sensitivity).count();
+        if (gapCount==0) return new ArrayList<>( Collections.singletonList(oneRegion) );
+
+        // get necessary information for computing
+        final String asmId = oneRegion.assemblyId;
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+        final String contigId = oneRegion.contigId;
+        final String chr = oneRegion.referenceInterval.getContig();
+        final Cigar originalForwardStrandCigar = oneRegion.forwardStrandCigar;
+        final boolean isForwardStrand = oneRegion.forwardStrand;
+        final int newMapQual = oneRegion.mapQual;
+        final int refStart = oneRegion.referenceInterval.getStart();
+        final int refEnd   = oneRegion.referenceInterval.getEnd();
+        final int contigLength = oneRegion.assembledContigLength;
+
+        // needs to accomplish three key functionality: 1. generate appropriate cigar, 2. infer reference coordinates, 3. infer contig coordinates, (cannot do yet: 4. mismatches and mapping quality)
```
SHuang-Broad: Done by added documentation for the whole method (essentially copying test cases from unit tests).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+        final List<AlignmentRegion> result = new ArrayList<>(gapCount+1);
+
+        final List<CigarElement> cigarMemoryList = new ArrayList<>();
+        final int clippedNBases = SVVariantCallerUtils.getNumClippedBases(true, originalForwardStrandCigar);
+        final int hardClippingAtEnd = (cigarElements.get(cigarElements.size()-1).getOperator()== CigarOperator.H)? cigarElements.get(cigarElements.size()-1).getLength() : 0;
+        int refIntervalStart = refStart, contigIntervalStart = 1 + clippedNBases;
+        int gapIndex=0;
+        for(int i=0; i<cigarElements.size(); ++i){ // considers only 'M', 'I', 'D', 'S', and 'H'
```
SHuang-Broad: done for 'EQ', 'X'. Need to run on actual data to decide if 'N' or 'P' is a serious issue.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+        final boolean isForwardStrand = oneRegion.forwardStrand;
+        final int newMapQual = oneRegion.mapQual;
+        final int refStart = oneRegion.referenceInterval.getStart();
+        final int refEnd   = oneRegion.referenceInterval.getEnd();
+        final int contigLength = oneRegion.assembledContigLength;
+
+        // needs to accomplish three key functionality: 1. generate appropriate cigar, 2. infer reference coordinates, 3. infer contig coordinates, (cannot do yet: 4. mismatches and mapping quality)
+        final List<AlignmentRegion> result = new ArrayList<>(gapCount+1);
+
+        final List<CigarElement> cigarMemoryList = new ArrayList<>();
```
SHuang-Broad: I chose the name because the implicit assumption made here that the function walks along the CIGAR assuming it is forward strand (always true for the input `AlignmentRegion`). So the intention was to signify that this is the memory of the CIGAR operators I have seen walking along the way. `newSubCigar` would lost this directioness, and it is a list not a CIGAR, IMO.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+                    cigarMemoryList.add(current);
+                    break;
+                case I: case D:
+                    if (operatorLen<sensitivity) {
+                        cigarMemoryList.add(current);
+                        break;
+                    }
+
+                    // collapse cigar memory list into a single cigar
+                    final List<CigarElement> cigarMemoryListModifiable = new ArrayList<>(cigarMemoryList);
```
SHuang-Broad: The intension was to make a local variable (the `cigarMemoryListModifiable`) to compute what CIGAR should be output for the newly generated alignments, i.e. increase readability.
In terms of performance, the original `cigarMemoryList` is going to be cleared anyway, so from that perspective it makes sense not to create the copy.
Am I understanding the issue correctly?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+                    final Cigar memoryCigar = new Cigar(cigarMemoryListModifiable);
+                    final int effectiveReadLen = memoryCigar.getReadLength() + SVVariantCallerUtils.getTotalHardClipping(memoryCigar) - SVVariantCallerUtils.getNumClippedBases(true, memoryCigar);
```
SHuang-Broad: Yes, thanks for pointing this out. Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+                    cigarMemoryListModifiable.add(new CigarElement(contigLength-contigIntervalEnd-hardClippingAtEnd, CigarOperator.S));
+                    if(hardClippingAtEnd!=0) cigarMemoryListModifiable.add(new CigarElement(hardClippingAtEnd, CigarOperator.H));
+                    final Cigar forwardStrandCigar = new Cigar(cigarMemoryListModifiable);
+
+                    final AlignmentRegion split = new AlignmentRegion(asmId, contigId, referenceInterval, forwardStrandCigar, isForwardStrand, newMapQual, ARTIFICIAL_CIGAR_MISMATCH, contigIntervalStart, contigIntervalEnd);
+
+                    result.add(split);
+
+                    // update pointers into reference and contig
+                    final boolean isInsertion = op==CigarOperator.I;
```
SHuang-Broad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+                    refIntervalStart     += isInsertion ? memoryCigar.getReferenceLength()  : memoryCigar.getReferenceLength() + operatorLen;
+                    contigIntervalStart  += isInsertion ? effectiveReadLen + operatorLen    : effectiveReadLen;
+                    cigarMemoryList.clear();    // update cigar memory (note: might not be an efficient frequent operation on linked-list, at least in theory, in practice, might be ok)
```
SHuang-Broad: done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
```diff
@@ -0,0 +1,476 @@
...
+    /**
+     * Given an ChimericAlignment representing two reference intervals rearranged as two intervals on the locally-assembled contig,
+     * identify potential complications such as homologyForwardStrandRep and duplication on the reference or the contig.
+     * Note: assuming input alignment regions are NOT completely enclosed in the other.
+     */
+    @VisibleForTesting
+    static BreakpointComplication resolveComplications(final ChimericAlignment chimericAlignment)
+            throws IOException {
+
+        final SimpleInterval leftReferenceInterval, rightReferenceInterval;
```
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantConsensusCall.java
```diff
@@ -0,0 +1,242 @@
...
+        final int start = breakpointPair.leftJustifiedLeftBreakpoint.getEnd();
+        final int end = breakpointPair.leftJustifiedRightBreakpoint.getStart();
+        final BreakpointPair.EvidenceFlankSide evidenceFlankSide = breakpointPair.evidenceFlankSide;
+
+        final Map<String, Object> attributeMap = new HashMap<>();
+
+        final GATKSVVCFHeaderLines.SVTYPES type;
+        if (evidenceFlankSide == BOTH) {
+            if (start==end) { // expected to be insertion
+                if (breakpointPair.complication.insertedSequenceForwardStrandRep.isEmpty() && breakpointPair.complication.contigDup<breakpointPair.complication.refDup)
```
SHuang-Broad: I believe this is related to the comment you have about how inversions are identified in this class. It would be better if we can talk in person as I had the same thought but decided to push the type identification into this class rather than the `BreakpointPair` class.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+     * Break a gapped alignment into multiple alignment regions, based on input sensitivity: i.e. if the gap(s) in the input alignment
+     * is equal to or larger than the size specified, two alignment regions will be generated.
+     * @return an iterable of size >= 1. if size==1, the returned iterable contains only the input (i.e. either no gap or hasn't reached sensitivity)
+     */
+    @VisibleForTesting
+    static Iterable<AlignmentRegion> breakGappedAlignment(final AlignmentRegion oneRegion, final int sensitivity) {
+
+        final List<CigarElement> cigarElements = oneRegion.forwardStrandCigar.getCigarElements();
+        if(cigarElements.size()==1) return new ArrayList<>( Collections.singletonList(oneRegion) );
+        final int gapCount = (int) cigarElements.stream().filter(ce -> ce.getOperator().isIndel() && ce.getLength() >= sensitivity).count();
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -0,0 +1,174 @@
...
+                    // update pointers into reference and contig
+                    final boolean isInsertion = op==CigarOperator.I;
+                    refIntervalStart     += isInsertion ? memoryCigar.getReferenceLength()  : memoryCigar.getReferenceLength() + operatorLen;
+                    contigIntervalStart  += isInsertion ? effectiveReadLen + operatorLen    : effectiveReadLen;
+                    cigarMemoryList.clear();    // update cigar memory (note: might not be an efficient frequent operation on linked-list, at least in theory, in practice, might be ok)
+                    final int hardClippingAtBeginning = SVVariantCallerUtils.getTotalHardClipping(memoryCigar);
+                    if (hardClippingAtBeginning!=0) cigarMemoryList.add(new CigarElement(hardClippingAtBeginning, CigarOperator.H));
+                    cigarMemoryList.add(new CigarElement(contigIntervalEnd - hardClippingAtBeginning + (isInsertion ? operatorLen : 0), CigarOperator.S));
+
+                    if (gapCount==++gapIndex) { // last gap
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ChimericAlignment.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ChimericAlignment.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
SHuang-Broad: The new variable is now removed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantConsensusCall.java
SHuang-Broad: done

SHuang-Broad:
@cwhelan , I've made the changes requested and update the tests accordingly. Would you please take a look again? Thanks!

SHuang-Broad:
The count of change of line is high because I did the suggested the name change `class BreakpointPair` -> `class NovelAdjacencyReferenceLocations`, just so that this makes reviewing the changes a bit easier.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
cwhelan: I'd at least make the dummy value be "NA" or something that indicates that `assemblyId` is not applicable when it's in use, then.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
cwhelan: Would 'previouslySeenCigarElements' capture what you're trying to say? Somehow the word 'Memory' was confusing to me at first (ie. does this have something to do with a special data structure?) Not a big deal if you want to leave it as is -- your choice.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointPair.java
cwhelan: Since you changed the naming to be about 'repeat numbers' rather than 'copy numbers' due to my comments, can you replace 'CN' with 'REPEAT_NUMBER' or something similar?

cwhelan:
Thanks for making these changes, this looks good now.

Issue #2309, opened 2016-12-12
Multi-sample ReadPileup could be more efficient


magicDGS:
In the previous version of GATK, a multi-sample pileup (already stratified by sample) was handled with a different class to be more efficient while performing operations by sample or re-splitting. This was done in a very complicated way, using [`PileupElementTracker`](https://github.com/broadgsa/gatk/blob/0b73e380436aaa5a41fb3aab97ab651207669f47/public/gatk-utils/src/main/java/org/broadinstitute/gatk/utils/pileup/PileupElementTracker.java) with different implementations of a `ReadBackedPileup` interface.

Instead of using a `List<PileupElement>` internally, `ReadPileup` could have an `ElementTracker` field, that could implement splitting, sorting by position and other operations to improve the efficiency.

This may solve issues like https://github.com/broadinstitute/gatk/issues/2245. In addition, it may improve the performance for a `LocusWalker` that needs the reads by sample, because the `LocusIteratorByState` already split by sample name the reads and pass them with a map.

PR #2321, opened 2016-12-20
Performance improvements for stratified/sorted ReadPileup

magicDGS:
This PR includes the following improvements for `ReadPileup` by using a new class for store `PileupElement` (`PileupElementTracker`):

* If sorting by alignment start is necessary, maintain a sorted `PileupElementTracker`
* If splitting by sample is already done, keep the stratified `PileupElementTracker`

Because pre-sorting and stratification is implemented in `LocusIteratorByState`, this will improve performance in `LocusWalker` and other tools using it. Other methods, `ReadPileup.fixOverlaps()` will benefit for these improvements.

This closes https://github.com/broadinstitute/gatk/issues/2309 and https://github.com/broadinstitute/gatk/issues/2245 (although the latest was already fixed, but it is rather inefficient).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
@@ -342,7 +342,8 @@ public final class LocusIteratorByState implements Iterator<AlignmentContext> {
                 }
 
                 if (!pile.isEmpty()) { // if this pileup added at least one base, add it to the full pileup
-                    fullPileupPerSample.put(sample, new ReadPileup(location, pile));
+                    // the pileup for LIBS is sorted by coordinate, because it comes from a sorted BAM file
+                    fullPileupPerSample.put(sample, new ReadPileup(location, pile, true));
                 }
             }
 
--- a/src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
@@ -1,35 +1,25 @@
 package org.broadinstitute.hellbender.utils.pileup;
 
-import com.google.api.services.genomics.model.Read;
 import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.util.Locatable;
-import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.exceptions.UserException;
-import org.broadinstitute.hellbender.tools.walkers.qc.Pileup;
 import org.broadinstitute.hellbender.utils.BaseUtils;
 import org.broadinstitute.hellbender.utils.QualityUtils;
 import org.broadinstitute.hellbender.utils.Utils;
-import org.broadinstitute.hellbender.utils.fragments.FragmentCollection;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
-import org.broadinstitute.hellbender.utils.read.ReadUtils;
-import org.broadinstitute.hellbender.engine.AlignmentContext;
 
 import java.util.*;
 import java.util.function.Predicate;
 import java.util.function.ToIntFunction;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 /**
  * Represents a pileup of reads at a given position.
  */
 public class ReadPileup implements Iterable<PileupElement> {
     private final Locatable loc;
-    private final List<PileupElement> pileupElements;
-
-    /** Constant used by samtools to downgrade a quality for overlapping reads that disagrees in their base. */
-    public static final double SAMTOOLS_OVERLAP_LOW_CONFIDENCE = 0.8;
+    private PileupElementTracker pileupElements;
 
     /**
      * Create a new pileup at loc, using the reads and their corresponding
@@ -37,25 +27,35 @@ public class ReadPileup implements Iterable<PileupElement> {
      * Note: This constructor keeps an alias to the given list.
      */
     public ReadPileup(final Locatable loc, final List<PileupElement> pileup) {
+        this(loc, pileup, false);
+    }
+
+    /**
+     * Create a new pileup at loc, using the reads and their corresponding
+     * offsets.
+     * Note: This constructor keeps an alias to the given list.
+     */
+    public ReadPileup(final Locatable loc, final List<PileupElement> pileup, final boolean preSorted) {
         Utils.nonNull(loc, "loc is null");
         Utils.nonNull(pileup, "element list is null");
         this.loc = loc;
-        this.pileupElements = pileup;
+        this.pileupElements = new UnifiedPileupElementTracker(pileup, preSorted);
     }
 
     /**
      * Create a new pileup at loc, using an stratified pileup
-     * Note: the current implementation of ReadPileup does not efficiently retrieve the stratified pileup
      */
     public ReadPileup(final Locatable loc, final Map<String, ReadPileup> stratifiedPileup) {
-        this(loc, stratifiedPileup.values().stream().flatMap(ReadPileup::getElementStream).collect(Collectors.toList()));
+        this.loc = loc;
+        this.pileupElements = new PerSamplePileupElementTracker(stratifiedPileup.entrySet().stream()
+                .collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue().pileupElements)));
     }
 
     /**
      * Create a new pileup without any aligned reads
      */
     public ReadPileup(final Locatable loc) {
-        this(loc, Collections.emptyList());
+        this(loc, Collections.emptyList(), true);
     }
 
     /**
@@ -72,6 +72,12 @@ public class ReadPileup implements Iterable<PileupElement> {
         this(loc, readsOffsetsToPileup(reads, offsets));
     }
 
+    // helper constructor
+    private ReadPileup(final Locatable loc, final PileupElementTracker tracker) {
+        this.loc = loc;
+        this.pileupElements = tracker;
+    }
+
     /**
      * Returns the first element corresponding to the given read or null there is no such element.
      *
@@ -80,7 +86,7 @@ public class ReadPileup implements Iterable<PileupElement> {
      */
     @VisibleForTesting
     PileupElement getElementForRead(final GATKRead read) {
-        return getElementStream().filter(el -> Objects.equals(el.getRead(), read)).findAny().orElse(null);
+        return pileupElements.getElementStream().filter(el -> Objects.equals(el.getRead(), read)).findAny().orElse(null);
     }
 
     /**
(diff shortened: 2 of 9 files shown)


magicDGS:
Hello @droazen, could you review this one? It is related with my work on the improvements for LIBS and `ReadPileup`, and it will make more efficient a lot of computation for splitting by sample the pileup object.

Thanks in advance!

magicDGS:
Friendly ping here, @droazen.

droazen:
@LeeTL1220 Since this PR possibly affects you more than anything else, would you have the time/interest to do a review pass on it?

src/main/java/org/broadinstitute/hellbender/utils/pileup/SingleSamplePileupElementTracker.java
```diff
@@ -0,0 +1,52 @@
...
+import htsjdk.samtools.SAMFileHeader;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
```
LeeTL1220: Can you put documentation here describing when this would be instantiated vs. the other pileup element trackers?

src/main/java/org/broadinstitute/hellbender/utils/pileup/UnifiedPileupElementTracker.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
```
LeeTL1220: Ditto about docs, because I don't understand why the SingleSamplePileupElementTracker inherits from Unified and PerSample inherits from PileupElementTracker.  If you could just address when a developer would instantiate each.

src/main/java/org/broadinstitute/hellbender/utils/pileup/SingleSamplePileupElementTracker.java
magicDGS: Added detailed documentation everywhere.

src/main/java/org/broadinstitute/hellbender/utils/pileup/UnifiedPileupElementTracker.java
magicDGS: Added a note for the inheritance and when they are supposed to be used.

magicDGS:
Thanks for comments about the documentation @LeeTL1220 - I fixed them (I hope) trying to explain the logic behind the element tracker. In principle a developer shouldn't care about when to use them, because they should come from a `LocusIteratorByState` or from inside a previous tracker. In addition, the implementation should be (most of the cases) hidden from the API user, which should use `ReadPileup`.

The idea of the trackers come from GATK3, so this is a custom port with some design differences. The basic idea is to cache some operations that may be time consuming for large pileups (sorting, split by sample, extract a single sample).

I actually haven't test the performance in a proper way, just running some tools in development with the branch and it feels like is faster - in my case I use all the features that are cache: split by sample, retrieving several times single-samples and also calling `fixOverlaps()` (which uses using sorted pileups).

I think that because the `LocusIteratorByState` is already splitting by sample, that can improve even more performance, because it will come directly in the state where it can be used by-sample in an efficient way. And maybe, if the tool does not require to split by sample at all, we can add an option to disable that behavior while creating the tracker.

Looking forward for your comments and ideas about this...

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -1,61 +1,61 @@
...
+    public ReadPileup(final Locatable loc, final List<PileupElement> pileup, final boolean preSorted) {
         Utils.nonNull(loc, "loc is null");
         Utils.nonNull(pileup, "element list is null");
         this.loc = loc;
-        this.pileupElements = pileup;
+        this.pileupElements = new UnifiedPileupElementTracker(pileup, preSorted);
     }
 
     /**
      * Create a new pileup at loc, using an stratified pileup
```
LeeTL1220: typo "an" --> "a"

LeeTL1220:
@magicDGS Overall, the documentation changes are good.  A few more minor comments:  
- If these are classes that are really pegged to the usage of LIBS, then you may want to mention that in the javadocs for the class.  A comment along the lines of "these classes are fairly low-level, developers should probably confirm that their changes do not belong in a higher-level calss such as LIBS".  
- I am okay with an informal test of the speed, even if you just look at some logs.  From the review, it looks like behavior of higher-level API calls will be unchanged.

>I think that because the LocusIteratorByState is already splitting by sample, that can improve even more performance, because it will come directly in the state where it can be used by-sample in an efficient way. And maybe, if the tool does not require to split by sample at all, we can add an option to disable that behavior while creating the tracker.

Agreed.  I do not think you need to worry about the flag, for now.  If you'd like, file an issue, but I think it is low priority for us.

I tend to be worried about new developers or contributors getting lost in the codebase.  And many do not have experience w/ GATK3.  Hence, the documentation about when to use the class and why it exists.

A couple additional things:
- I found a typo (see my comment)
- Can you document the `presorted` parameter?  Make sure to mention that if a developer specifies false, and sorting is needed, it will be done under-the-hood.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
magicDGS: Done.

magicDGS:
@LeeTL1220 - I addressed in the next commit the documentation issues; let me know if you have more suggestions on that.

If you have any suggestion for an already implemented tool in GATK for the performance and some data for profiling, I can do a couple of runs in my local computer for look at them. If not, I can implement an small example tool for profiling when splitting by sample is needed (I can also modify `ExampleLocusWalker` for that in a different PR to being able to profile this after this PR too).

src/main/java/org/broadinstitute/hellbender/utils/pileup/SingleSamplePileupElementTracker.java
```diff
@@ -23,23 +23,47 @@
...
+     */
     SingleSamplePileupElementTracker(final String sampleName) {
         this(sampleName, Collections.emptyList(), true);
     }
 
-    /** Instantiates an unsorted element tracker for the sample. */
+    /**
+     * Instantiates an unsorted element tracker for the sample.
+     *
+     * <p>Note: if a {@link #sortedIterator()} is requested, sorting would be performed an cached into the tracker.
```
LeeTL1220: typo "and" --> "and"

LeeTL1220: ... sorting will be performed ...

src/main/java/org/broadinstitute/hellbender/utils/pileup/SingleSamplePileupElementTracker.java
```diff
@@ -23,23 +23,47 @@
...
+     */
     SingleSamplePileupElementTracker(final String sampleName, final List<PileupElement> sampleElements) {
         this(sampleName, sampleElements, false);
     }
 
-    /** Instantiates a sorted/unsorted element tracker for them sample. The sample may be {@code null}. */
+    /**
+     * Instantiates a sorted/unsorted element tracker for them sample. The sample may be {@code null}.
+     *
+     * <p>Note: if {@code preSorted=true} and a {@link #sortedIterator()} is requested, sorting would be performed
```
LeeTL1220: ditto about typos

src/main/java/org/broadinstitute/hellbender/utils/pileup/UnifiedPileupElementTracker.java
```diff
@@ -43,12 +47,25 @@
...
+     */
     UnifiedPileupElementTracker(final List<PileupElement> pileup) {
         this(pileup, false);
     }
 
-    /** Instantiates a sorted/unsorted element tracker. */
+    /**
+     * Instantiates a sorted/unsorted element tracker.
+     *
+     * <p>Note: if {@code preSorted=true} and a {@link #sortedIterator()} is requested, sorting would be performed
```
LeeTL1220: Ditto on typos

src/main/java/org/broadinstitute/hellbender/utils/pileup/SingleSamplePileupElementTracker.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/utils/pileup/SingleSamplePileupElementTracker.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/utils/pileup/UnifiedPileupElementTracker.java
magicDGS: Done.

magicDGS:
Fixed typos @LeeTL1220 - maybe we can profile this with https://github.com/broadinstitute/gatk/pull/3614, which is using splitting by sample?

LeeTL1220:
@magicDGS That sounds perfectly fine to use #3614 .  Regardless, I think you are safe to merge unless that quick test is worrisome.

droazen:
I'd like to have a quick look at this before merge, to make sure it won't negatively affect the `HaplotypeCaller`

magicDGS:
Waiting for your review, @droazen, for making the profiling with #3614 (maybe it would be already accepted).

src/test/java/org/broadinstitute/hellbender/utils/pileup/PileupElementTrackerUnitTest.java
```diff
@@ -0,0 +1,144 @@
...
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class PileupElementTrackerUnitTest extends BaseTest {
```
magicDGS: Self-reminder: extend `GATKBaseTest` (requires rebase)

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -1,61 +1,61 @@
...
 
     /**
-     * Create a new pileup at loc, using an stratified pileup
-     * Note: the current implementation of ReadPileup does not efficiently retrieve the stratified pileup
+     * Create a new pileup at loc, using a stratified pileup.
      */
     public ReadPileup(final Locatable loc, final Map<String, ReadPileup> stratifiedPileup) {
-        this(loc, stratifiedPileup.values().stream().flatMap(ReadPileup::getElementStream).collect(Collectors.toList()));
+        this.loc = loc;
+        this.pileupElements = new PerSamplePileupElementTracker(stratifiedPileup.entrySet().stream()
```
magicDGS: Self note: this was changed in #4031 for a performance improvement for stratified pileups. We should check if there is any regression...

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
```diff
@@ -342,7 +342,8 @@ private void lazyLoadNextAlignmentContext() {
                 }
 
                 if (!pile.isEmpty()) { // if this pileup added at least one base, add it to the full pileup
-                    fullPileupPerSample.put(sample, new ReadPileup(location, pile));
+                    // the pileup for LIBS is sorted by coordinate, because it comes from a sorted BAM file
```
magicDGS: This was changed by the performance improvement for HC in #4031 - we should check if maintaining the stratified pileup here cause a regression of that improvement.

droazen:
Re-assigning to @jonn-smith and @jamesemery to assess whether this is worth resurrecting as part of the current HaplotypeCaller performance work.

jonn-smith:
Closing issue due to age and code conflicts.  If this issue becomes a blocker in the future we can reinvestigate then.

Issue #2297, opened 2016-12-06
Evaluate commons-configuration for use as an overridable master GATK configuration mechanism


droazen:
PRs like https://github.com/broadinstitute/gatk/pull/2156 make it clear that we need some master configuration mechanism in the GATK that can be overridden by clients/downstream projects. 

One promising option is `commons-configuration` (https://commons.apache.org/proper/commons-configuration/userguide/user_guide.html) using properties files -- we should look into this to see whether it does what we want.

droazen:
@tomwhite Can you look at this and make a proposal?

droazen:
Dupe of https://github.com/broadinstitute/gatk/issues/2368

PR #2322, opened 2016-12-21
Simple GATKConf (NO MERGE)

magicDGS:
Very simple implementation of #2297 using a custom `GATKConf` class to allow both promatically (`GATKConfBuilder`) and by commons-configuration API (constructor). Only includes:

* Packages/Classes to include in the CLP on startup.
* Packages to look for codecs on startup.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -1,11 +1,16 @@
 package org.broadinstitute.hellbender;
 
 import htsjdk.samtools.util.StringUtil;
+import org.apache.commons.configuration.ConfigurationException;
 import org.broadinstitute.hellbender.cmdline.ClassFinder;
 import org.broadinstitute.barclay.argparser.CommandLineException;
 import org.broadinstitute.barclay.argparser.CommandLineProgramGroup;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.broadinstitute.hellbender.cmdline.CommandLineProgram;
+import org.broadinstitute.hellbender.conf.GATKConf;
+import org.broadinstitute.hellbender.conf.GATKConfBuilder;
+import org.broadinstitute.hellbender.engine.FeatureManager;
+import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.ClassUtils;
 
@@ -19,8 +24,7 @@ import java.util.*;
  *
  * If you want your own single command line program, extend this class and override if required:
  *
- * - {@link #getPackageList()} to return a list of java packages in which to search for classes that extend CommandLineProgram.
- * - {@link #getClassList()} to return a list of single classes to include (e.g. required input pre-processing tools).
+ * - {@link #getConfiguration()} to setup the configuration on startup.
  * - {@link #getCommandLineName()} for the name of the toolkit.
  * - {@link #handleResult(Object)} for handle the result of the tool.
  * - {@link #handleNonUserException(Exception)} for handle non {@link UserException}.
@@ -57,20 +61,8 @@ public class Main {
     private static final int ANY_OTHER_EXCEPTION_EXIT_VALUE = 3;
     private static final String STACK_TRACE_ON_USER_EXCEPTION_PROPERTY = "GATK_STACKTRACE_ON_USER_EXCEPTION";
 
-    /**
-     * The packages we wish to include in our command line.
-     */
-    protected List<String> getPackageList() {
-        final List<String> packageList = new ArrayList<>();
-        packageList.addAll(Arrays.asList("org.broadinstitute.hellbender"));
-        return packageList;
-    }
-
-    /**
-     * The single classes we wish to include in our command line.
-     */
-    protected List<Class<? extends CommandLineProgram>> getClassList() {
-        return Collections.emptyList();
+    protected GATKConf getConfiguration() {
+        return GATKConfBuilder.getDefaultConfiguration();
     }
 
     /** Returns the command line that will appear in the usage. */
@@ -100,7 +92,9 @@ public class Main {
      * This method is not intended to be used outside of the GATK framework and tests.
      */
     public Object instanceMain(final String[] args) {
-        return instanceMain(args, getPackageList(), getClassList(), getCommandLineName());
+        final GATKConf configuration = getConfiguration();
+        FeatureManager.setConfiguration(configuration);
+        return instanceMain(args, configuration.getToolPackages(), configuration.getToolClasses(), getCommandLineName());
     }
 
     /**
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -1,7 +1,6 @@
 package org.broadinstitute.hellbender;
 
 import htsjdk.samtools.util.StringUtil;
-import org.apache.commons.configuration.ConfigurationException;
 import org.broadinstitute.hellbender.cmdline.ClassFinder;
 import org.broadinstitute.barclay.argparser.CommandLineException;
 import org.broadinstitute.barclay.argparser.CommandLineProgramGroup;
@@ -16,6 +15,7 @@ import org.broadinstitute.hellbender.utils.ClassUtils;
 
 import java.io.Serializable;
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * This is the main class of Hellbender and is the way of executing individual command line programs.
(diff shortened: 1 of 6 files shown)


magicDGS:
could some of you have a look to this, @droazen and/or @lbergelson? We started the discussion in #2156, so maybe this implementation is a good starting point.

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -100,7 +92,9 @@ public Object instanceMain(final String[] args, final List<String> packageList,
      * This method is not intended to be used outside of the GATK framework and tests.
      */
     public Object instanceMain(final String[] args) {
-        return instanceMain(args, getPackageList(), getClassList(), getCommandLineName());
+        final GATKConf configuration = getConfiguration();
+        FeatureManager.setConfiguration(configuration);
```
droazen: Rather than pass the full configuration object to engine classes like `FeatureManager`, I'd prefer to pass just the `List<String>` of packages to use for codec discovery.

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -57,20 +61,8 @@
...
-        return packageList;
-    }
-
-    /**
-     * The single classes we wish to include in our command line.
-     */
-    protected List<Class<? extends CommandLineProgram>> getClassList() {
-        return Collections.emptyList();
+    protected GATKConf getConfiguration() {
+        return GATKConfBuilder.getDefaultConfiguration();
```
droazen: I think it would be better if the master configuration object lived as a field in `CommandLineProgram` accessible by subclasses via an inherited method. This would simplify passing property-based information around the engine in the future.

src/main/java/org/broadinstitute/hellbender/conf/GATKConf.java
```diff
@@ -0,0 +1,58 @@
...
+    @SuppressWarnings("unchecked")
+    public List<String> getToolPackages() {
+        return configuration.getList(GATK_PROPERTY_PREFIX +  TOOLS_PACKAGES_KEY);
+    }
+
+    /**
+     * The single classes we wish to include in our command line.
+     */
+    @SuppressWarnings("unchecked")
+    public List<Class<? extends CommandLineProgram>> getToolClasses() {
```
droazen: I think configuration properties should always be typed as Strings. Otherwise we have to worry about typing issues / casting, etc. Even in this simple branch, there are already inconsistencies in typing (eg., depending on whether you call `setClpClasses()` or `addClpClasses()`, the clpClasses might be Strings or Classes

src/main/java/org/broadinstitute/hellbender/conf/GATKConf.java
```diff
@@ -0,0 +1,58 @@
...
+    GATKConf(final Configuration configuration) {
+        Utils.nonNull(configuration);
+        this.configuration = configuration;
+    }
+
+    /**
+     * The packages we wish to include in our command line.
+     */
+    @SuppressWarnings("unchecked")
+    public List<String> getToolPackages() {
```
droazen: I'm not sure we want to have to define a separate method for each property we add. Perhaps we could just have a bunch of constants with the property names, as you do above, and have the client just pass in one of those constants when accessing properties.

src/main/java/org/broadinstitute/hellbender/conf/GATKConfBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public static final GATKConf useConfiguration(final Configuration configuration) {
+        return new GATKConf(configuration);
+    }
+
+    public GATKConfBuilder setDefaults() {
+        return setToolsPackages(new ArrayList<>(DEFAULT_TOOLS_PACKAGES))
+                .setClpClasses(new ArrayList<>())
+                .setCodecPackages(new ArrayList<>(DEFAULT_CODEC_PACKAGES));
```
droazen: Can we check in a properties file with the default settings that gets loaded on startup, instead of hardcoding them here?

src/main/java/org/broadinstitute/hellbender/engine/FeatureManager.java
```diff
@@ -42,34 +45,33 @@
...
     /**
-     * At startup, walk through the packages in CODEC_PACKAGES, and save any (concrete) FeatureCodecs discovered
-     * in DISCOVERED_CODECS
+     * Walk through the packages in {@link GATKConf#getCodecPackages()}, and save any (concrete) FeatureCodecs discovered
+     * in DISCOVERED_CODECS.
+     *
+     * @throws IllegalStateException if codecs were alredy initialized.
      */
-    static {
+    public synchronized static void setConfiguration(final GATKConf configuration) {
```
droazen: Replace with `setCodecPackages()`, called from `CommandLineProgram` using the information from the master configuration object.

src/main/java/org/broadinstitute/hellbender/Main.java
magicDGS: I agree partially on this. It is better to have `GATKConf` in `CommandLineProgram`, but if this method remains, to allow extending `Main` classes to set their own configuration on startup. This is important for one reason: if an API client wants its own configuration and it only lives in `CommandLineProgram` accessible by subclasses, it requires that every time a tool is implemented this method should be overridden. Thus, it makes more dirty to implement walker types which only requires an apply, for example.

src/main/java/org/broadinstitute/hellbender/conf/GATKConf.java
magicDGS: I did it to cast the list properly in the case of the tool classes, for instance. It is true that as the required configuration increases, this class will grow a lot.

src/main/java/org/broadinstitute/hellbender/conf/GATKConf.java
magicDGS: The inconsistency is my fault, because I haven't paid too much attention in the builder (I'm sorry, I detected other inconsistencies).

I think that typing as String is a good idea, even more if the `GATKConf` use constants to access the properties.

src/main/java/org/broadinstitute/hellbender/conf/GATKConfBuilder.java
magicDGS: A properties file will be great, but I decided to hardcode these properties because how is it going to be handled this file in project dependencies? I'm not sure how to do this in a proper way, because I'm not coming from the IT background...I may need some help on this.

src/main/java/org/broadinstitute/hellbender/engine/FeatureManager.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/Main.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/Main.java
magicDGS: I added a settable field in `CommandLineProgram` in a similar way as in `FeatureManager`.

src/main/java/org/broadinstitute/hellbender/Main.java
magicDGS: I also move this part of the code to the other `instanceMain()` method, passing just the configuration object generated here instead of the list to set the `CommandLineProgram` configuration properly.

src/main/java/org/broadinstitute/hellbender/conf/GATKConf.java
magicDGS: I removed these methods and create `GATKConf.getGATKProperty(String key)` instead.

src/main/java/org/broadinstitute/hellbender/conf/GATKConf.java
magicDGS: As before, I removed this method.

src/main/java/org/broadinstitute/hellbender/engine/FeatureManager.java
magicDGS: But I call it in `instanceMain(final String[] args, final GATKConf configuration, final String commandLineName)`.

magicDGS:
I addressed some of your comments, @droazen. If you would like to have a properties file for the configuration, I will need some help on setting it up (although I will try by my own too).

Although I still set up some of the environment in `Main`, now the `CommandLineProgram` class have the same instance passed by `Main`.

Looking forward for your comments on the updates.

src/main/java/org/broadinstitute/hellbender/conf/GATKConf.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * The single classes we wish to include in our command line.
+     */
+    public final static String TOOL_CLASS_LIST_KEY = "toolClasses";
+
+    /**
+     * We will search these packages at startup to look for FeatureCodecs.
+     */
+    public final static String CODEC_PACKAGE_LIST_KEY = "codecPackages";
```
magicDGS: Another package list required is the one to look for read filters in `GATKReadFilterPluginDescriptor`.

cmnbroad:
One proposal for moving forward would be to have a default properties file with a known name/location that is included in the gatk jar (say, "gatk.default.properties"), which is always loaded and populates the initial configuration, and then use the classloader getResources method to also load all resources with some other known name (say, "gatk.properties"). That way any properties files on the classpath with the known name would be automatically discovered and loaded. The apache commons API allows looks like it has good support for handling this using a [composite](http://commons.apache.org/proper/commons-configuration/userguide/howto_compositeconfiguration.html#Composite_Configuration_Details) configuration. We would have to define some rules around override semantics, but it looks like the api provides a lot of control over that as well.

magicDGS:
Where should be sited "gatk.default.properties"? I'm still investigating a bit on how to do this in the proper way...

droazen:
Closing -- this is supplanted by the work being done in https://github.com/broadinstitute/gatk/issues/2368. @jonn-smith You should continue to use this closed PR as a reference for understanding @magicDGS 's use case.

Issue #2265, opened 2016-11-15
Port ASEReadCounter to GATK4 and make it ignore homozygous sites


vdauwera:
@chandrans commented on [Mon May 23 2016](https://github.com/broadinstitute/gsa-unstable/issues/1382)

## Feature request

A user has reported ASEReadCounter counts homozygous sites as having potential ASE. This should not be the case, as only heterozygous sites should be counted. The user should input a VCF sites file with only heterozygous sites, but sometimes that doesn't happen, so there should be an error message when he or she does not. 
### Tool(s) involved

ASEReadCounter
### Description

There needs to be an error message telling the user he/she needs to subset the sites VCF to only heterozygous sites.

---

@vdauwera commented on [Mon Nov 14 2016](https://github.com/broadinstitute/gsa-unstable/issues/1382#issuecomment-260490723)

May want to move this to GATK4, waiting for @meganshand to opine. 

---

@meganshand commented on [Tue Nov 15 2016](https://github.com/broadinstitute/gsa-unstable/issues/1382#issuecomment-260648100)

Yeah, let's move it to GATK4 and incorporate porting it in the first place.

vdauwera:
For @meganshand 

I'm assuming this tool can be open-sourced so I'm putting this in `gatk`, but feel free to move to protected if technically necessary.

PR #2325, opened 2016-12-22
Port of ASEReadCounter from GATK3

meganshand:
Addresses #2265 by porting the tool and adds skipping (with a warning) non het sites.


src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
```diff
@@ -1,254 +1,245 @@
...
         else
             throw new UserException("Must use valid CountPileupType");
 
-        return includeDeletions ? pileupWithDeletions: pileupWithDeletions.getPileupWithoutDeletions();
+        return pileupWithDeletions.makeFilteredPileup(p -> !p.isDeletion());
 
     }
 
-    protected String calculateLineForSite(final ReadBackedPileup pileup, final String contig, final long position, final String siteID, final char refAllele, final char altAllele){
+    protected void calculateLineForSite(final ReadPileup pileup, final String contig, final long position, final String siteID, final char refAllele, final char altAllele) {
```
lbergelson: Do you ever want to be able to pass in a separate location from where the read pileup is?  I think you can just pass in readpileup and drop the contig and position for a few less parameters

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -317,73 +321,59 @@ static void fixPairOverlappingQualities(final PileupElement firstElement,
...
-            for (final String sample : tracker.getSamples()) {
-                PileupElementTracker<PileupElement> perSampleElements = tracker.getElements(sample);
-                ReadBackedPileupImpl pileup = createNewPileup(loc, perSampleElements).getOverlappingFragmentFilteredPileup(discardDiscordant, baseQualNotMapQual);
-                filteredTracker.addElements(sample, pileup.pileupElementTracker);
-            }
-            return createNewPileup(loc, filteredTracker);
-        } else {
-            Map<String, PileupElement> filteredPileup = new HashMap<String, PileupElement>();
+    public ReadPileup getOverlappingFragmentFilteredPileup(boolean discardDiscordant, boolean baseQualNotMapQual) {
+        Map<String, PileupElement> filteredPileup = new HashMap<String, PileupElement>();
```
lbergelson: The old version seems to distinguish reads between samples.  It's fine to drop that functionality and just assume everything is from the same sample?

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -317,73 +321,59 @@ static void fixPairOverlappingQualities(final PileupElement firstElement,
...
+            // if we've never seen this read before, life is good
+            if (!filteredPileup.containsKey(readName)) {
+                filteredPileup.put(readName, p);
+            } else {
+                PileupElement existing = filteredPileup.get(readName);
 
-                // if we've never seen this read before, life is good
-                if (!filteredPileup.containsKey(readName)) {
-                    filteredPileup.put(readName, p);
+                // if the reads disagree at this position, throw them both out.  Otherwise
```
lbergelson: Do you need handling for things like supplementary alignments where you'll have more than 2 reads with the same name? Currently, if you have 3 reads and they disagree, you'll throw out the first two and then keep the last one...

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -317,73 +321,59 @@ static void fixPairOverlappingQualities(final PileupElement firstElement,
...
-
-            for (final String sample : tracker.getSamples()) {
-                PileupElementTracker<PileupElement> perSampleElements = tracker.getElements(sample);
-                ReadBackedPileupImpl pileup = createNewPileup(loc, perSampleElements).getOverlappingFragmentFilteredPileup(discardDiscordant, baseQualNotMapQual);
-                filteredTracker.addElements(sample, pileup.pileupElementTracker);
-            }
-            return createNewPileup(loc, filteredTracker);
-        } else {
-            Map<String, PileupElement> filteredPileup = new HashMap<String, PileupElement>();
+    public ReadPileup getOverlappingFragmentFilteredPileup(boolean discardDiscordant, boolean baseQualNotMapQual) {
```
lbergelson: It might be nicer to have this take a function for getting the tiebreaker instead of the boolean.  I.e. 

public ReadPileup getOverlappingFragmentFilteredPileup(boolean discardDiscordant, Function<ReadPileup, Integer> tieBreaker)

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -317,73 +321,59 @@ static void fixPairOverlappingQualities(final PileupElement firstElement,
...
                 }
             }
+        }
 
-            UnifiedPileupElementTracker<PileupElement> filteredTracker = new UnifiedPileupElementTracker<PileupElement>();
-            for (PileupElement filteredElement : filteredPileup.values())
-                filteredTracker.add(filteredElement);
+        List<PileupElement> filteredPileupList = new ArrayList<PileupElement>();
+        for (PileupElement filteredElement : filteredPileup.values())
+            filteredPileupList.add(filteredElement);
```
lbergelson: I think you can simplify this loop away with  `List<PileupElement> filteredPileupList = new ArrayList<>(filteredPileup.values())`

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounterIntegrationTest.java
```diff
@@ -1,121 +1,75 @@
...
     @Test
-    public void testASEReadCounterWithLowMQ() {
-        WalkerTest.WalkerTestSpec spec = new WalkerTest.WalkerTestSpec(
-                "-T ASEReadCounter -R " + b37KGReference + " -I " + privateTestDir + "NA12878.RNAseq.bam -sites "+privateTestDir +"NA12878.chr20_2444518_2637800.RNAseq.SYNONYMOUS_CODING.vcf -mmq 1 -o %s -U ALLOW_N_CIGAR_READS", 1,
-                Arrays.asList("4f1144be0bb2c4adeae00625afd04cda"));
-        executeTest("test high mq with no read passing", spec);
+    public void testASEReadCounterWithHighMQ() throws Exception {
+        IntegrationTestSpec spec = new IntegrationTestSpec(
+                "-R " + b37_reference_20_21 + " -I " + largeFileTestDir + "NA12878.RNAseq.bam -V " + publicTestDir + "NA12878.chr20_2444518_2637800.RNAseq.SYNONYMOUS_CODING.vcf -O %s -mmq 60 ",
+                Arrays.asList(publicTestDir + "expected.ASEReadCount.WithHighMQ.table"));
```
lbergelson: Could you move the test files into `src/test/resources/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter/`? 

It keep things a bit cleaner and easier to understand what a file is used for.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
```diff
@@ -0,0 +1,286 @@
...
+    @Override
+    public void apply(AlignmentContext alignmentContext, ReferenceContext referenceContext, FeatureContext featureContext) {
+        final String contig = alignmentContext.getContig();
+        final long position = alignmentContext.getPosition();
+
+        final char refAllele = (char) referenceContext.getBase();
+
+        final List<VariantContext> VCs = featureContext.getValues(variants);
+        if (VCs != null && VCs.size() > 1)
+            throw new UserException("More then one variant context at position: " + contig + ":" + position);
```
lbergelson: Could you add a test case that hits this case?

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
```diff
@@ -0,0 +1,286 @@
...
+            return;
+        }
+
+        if (vc.getHetCount() < 1) {
+            logger.warn("Ignoring site: variant is not het at postion: " + contig + ":" + position);
+            return;
+        }
+
+        if (vc.getNAlleles() == 1 || vc.getAlternateAllele(0).getBases().length == 0)
+            throw new UserException("The file of variant sites must contain heterozygous sites and cannot be a GVCF file containing <NON_REF> alleles.");
```
lbergelson: This case should have a test too.  You might want ones for the warning cases that just check that they don't crash, but I think those can slide if you don't want to bother.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
```diff
@@ -0,0 +1,286 @@
...
+
+    protected void calculateLineForSite(final ReadPileup pileup, final String contig, final long position, final String siteID, final char refAllele, final char altAllele) {
+
+        int rawDepth = 0, lowBaseQDepth = 0, lowMAPQDepth = 0, refCount = 0, altCount = 0, totalNonFilteredCount = 0, otherBasesCount = 0, improperPairsCount = 0;
+
+        for (final PileupElement base : pileup) {
+            rawDepth++;
+
+            if (base.getRead().isPaired() && (base.getRead().mateIsUnmapped() || !base.getRead().isProperlyPaired())) {
+                improperPairsCount++;
```
lbergelson: This line never gets executed, maybe you can add an improper pair to one of your test cases?

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -316,6 +320,62 @@ static void fixPairOverlappingQualities(final PileupElement firstElement,
...
+
+    /**
+     * Returns a new ReadPileup where only one read from an overlapping read
+     * pair is retained.  If discardDiscordant and the two reads in question disagree to their basecall,
+     * neither read is retained.  Otherwise, the read with the higher
+     * quality (base or mapping, depending on baseQualNotMapQual) observation is retained
+     *
+     * @return the newly filtered pileup
+     */
+    public ReadPileup getOverlappingFragmentFilteredPileup(boolean discardDiscordant, boolean baseQualNotMapQual) {
```
lbergelson: This needs a unit test in `ReadPileupUnitTest`, especially since it's a bit complicated and has a number of different potential cases.  (likewise, it's no argument overload should probably have a unit test too).

meganshand: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
meganshand: Because there's one case that needs to not output anything (including a new line) I'm having `calculateLineForSite` return `null` in that case. Do you have a cleaner solution than having the function be void and output the line or return a `StringBuilder` that's null in that one case?

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
```diff
@@ -0,0 +1,301 @@
...
+            case COUNT_FRAGMENTS: pileupWithDeletions = originalPileup.getOverlappingFragmentFilteredPileup(false, true);
+                break;
+            default: throw new UserException("Must use valid CountPileupType");
+        }
+
+        return pileupWithDeletions.makeFilteredPileup(p -> !p.isDeletion());
+
+    }
+
+    private StringBuilder calculateLineForSite(final ReadPileup pileup, final String siteID, final char refAllele, final char altAllele) {
```
lbergelson: I would return the String that gets build and not the builder itself.  Returning a `StringBuilder` sort of implies that the builder is intended to be used in some way, i.e. to continue building a line.  

In the no output case, you could return an empty string, and then print that instead of println if the new line is the issue.  Or just check the null case and not do anything like you're doing now.  

In general, my feeling is that void methods are the devil and should be as absolutely clear in the function as possible.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
meganshand: I made it a comparator.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
meganshand: I'm not quite sure if this is a problem (in my use case I would only submit one sample per BAM), but even if I wanted to make it robust to multiple sample BAMs, I'm not sure how to do it. `ReadPileup.getSamples` takes in a SAMFileHeader. I'm not sure how to get the BAM header when I'm inside of the `apply` function in a `LocusWalker`.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
meganshand: Presumably supplementary alignments should happen at some distant location that isn't in the same pileup as the primary alignments, but I changed it to work for 2+ reads (and added a test) anyway.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
meganshand: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounterIntegrationTest.java
meganshand: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
meganshand: fixed

meganshand:
@lbergelson Thanks for the review! I think I've caught all of your comments (except for handling multi-sample BAMs). Sorry for the delayed response.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -340,38 +354,35 @@ public ReadPileup getOverlappingFragmentFilteredPileup() {
      *
      * @return the newly filtered pileup
      */
-    public ReadPileup getOverlappingFragmentFilteredPileup(boolean discardDiscordant, boolean baseQualNotMapQual) {
+    public ReadPileup getOverlappingFragmentFilteredPileup(boolean discardDiscordant, Comparator<PileupElement> tieBreaker) {
```
lbergelson: 👍

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
```diff
@@ -253,34 +251,50 @@ protected void calculateLineForSite(final ReadPileup pileup, final String contig
...
-                totalNonFilteredCount + separator +
-                lowMAPQDepth + separator +
-                lowBaseQDepth + separator +
-                rawDepth + separator +
-                otherBasesCount + separator +
-                improperPairsCount;
-
-        outputStream.println(line);
+        final StringBuilder line = new StringBuilder();
+        line.append(pileup.getLocation().getContig());
```
lbergelson: Sorry about this, I know it's ugly, but I think it will save quite a bit of garbage collection churn.

There are a few ways to improve readability and reduce boiler plate here that you might want to think about in the future, 

1.  when it's something like this you can put two statements on one line, ( know, it violates everything you've ever been taught to believer in, I'm sure I've complained about that very thing before... but there are always things so ugly they require exceptions....)

ie.
```
line.append(pileup.getLocation().getContig()); line.append(separator);
line.append(pileup.getLocation().getStart()); line.append(separator);
```

2.  StringBuilders have a "fluent" interface, so you can chain calls to them together.

```
line.append(pileUp.getLocation().getContig()).append(separator)
    .append(pileup.getLocation().getStart()).append(separator)
```

src/test/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounterIntegrationTest.java
```diff
@@ -1,75 +1,136 @@
...
+                "-R " + b37_reference_20_21 + " -I " + largeFileTestDir + "NA12878.RNAseq.bam -V " + aseTestDir + "NA12878.chr20_2444518_2637800.RNAseq.SYNONYMOUS_CODING.vcf -mmq 60 -mbq 10 -O %s --outputFormat CSV",
+                Arrays.asList(aseTestDir + "expected.ASEReadCount.FileFormat.table"));
         spec.executeTest("test high mq with no read passing", this);
     }
+
+    @Test(expectedExceptions = UserException.class)
+    public void testASEReadCounterMultipleContexts() {
+        ArgumentsBuilder args = new ArgumentsBuilder();
+
+        args.add("-R");
```
lbergelson: ack, we really need shortcuts in argsbuilder for this

lbergelson:
@meganshand If you don't think multisample bams are a use case than I think it's probably fine to ignore them.  I was just wondering because it seemed like the old version had some support for it.  

I think you can get the bam header by calling `getHeaderForReads()` if you needed it.

Issue #2193, opened 2016-09-29
Remove spanning deletions if the deletion was removed


ronlevine:
@ronlevine commented on [Thu Sep 29 2016](https://github.com/broadinstitute/gatk-protected/issues/725)

@ronlevine commented on [Tue Jul 19 2016](https://github.com/broadinstitute/gatk/issues/2017)

Port https://github.com/broadinstitute/gsa-unstable/pull/1417.

---

@ronlevine commented on [Wed Jul 20 2016](https://github.com/broadinstitute/gatk/issues/2017#issuecomment-233990695)

Need to port `GenotypeGVCFs` before testing the proposed changes in `GenotypingEngine.java`.

---

@ronlevine commented on [Thu Sep 29 2016](https://github.com/broadinstitute/gatk/issues/2017#issuecomment-250468170)

@droazen Should this be a gatk-protected ticket instead since it depends on GenotypeGVCFs?

ronlevine:
Closed since #2326 was merged.

PR #2326, opened 2016-12-22
Remove spanning deletions if the deletion was removed

ronlevine:
Implements #2193.
This is a port of https://github.com/broadinstitute/gsa-unstable/pull/1417. Unit tests were added since the GATK3 version was tested with GenotypeGVCFs, which is in `gatk-protected`. There is a new internal class, `GenomeLocation` is used instead of `GenomeLoc` to store the deletions. This is because `GenomeLocParser`, which is used to create  `GenomeLoc` is not passed into `GenotypingEngine` in GATK4. I could not find a clean way to add it.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
@@ -9,7 +9,6 @@ import org.broadinstitute.barclay.argparser.CommandLineException;
 import org.broadinstitute.hellbender.engine.AlignmentContext;
 import org.broadinstitute.hellbender.engine.FeatureContext;
 import org.broadinstitute.hellbender.engine.ReferenceContext;
-import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.tools.walkers.annotator.VariantAnnotatorEngine;
 import org.broadinstitute.hellbender.tools.walkers.genotyper.afcalc.*;
 import org.broadinstitute.hellbender.utils.*;
@@ -31,7 +30,7 @@ public abstract class GenotypingEngine<Config extends StandardCallerArgumentColl
 
     protected final AFCalculator newAFCalculator;
 
-    protected final AFCalculatorProvider afCalculatorProvider   ;
+    protected final AFCalculatorProvider afCalculatorProvider;
 
     protected final Config configuration;
 
@@ -46,6 +45,8 @@ public abstract class GenotypingEngine<Config extends StandardCallerArgumentColl
     private final AFPriorProvider log10AlleleFrequencyPriorsSNPs;
 
     private final AFPriorProvider log10AlleleFrequencyPriorsIndels;
+
+    private final List<SimpleInterval> upstreamDeletionsLoc = new LinkedList<>();
 
     /**
      * Construct a new genotyper engine, on a specific subset of samples.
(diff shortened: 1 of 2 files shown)


ronlevine:
@ldgauthier Please review.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
```diff
@@ -29,9 +28,24 @@
  */
 public abstract class GenotypingEngine<Config extends StandardCallerArgumentCollection> {
 
+    /**
+     * Genome location
+     */
+    static class GenomeLocation {
```
droazen: Please port this class to use `SimpleInterval` instead, if possible.

ronlevine: 👍

ldgauthier:
Looks good to me. :+1: I'm glad I know about this code because we've also been pondering what should happen to spanning deletions when the parent deletion is filtered by VQSR.

ldgauthier:
@vdauwera As Louis just pointed out to me, if GenotypeGVCFs is scattered such that it breaks active regions this fix won't work.  In production (GATK3) I'm pretty sure we make sure we don't break exome intervals and for genomes we won't break active regions either.  In GATK4 the WDL scatter task will have this responsibility.  Probably worth pointing out in release notes or somewhere.

vdauwera:
Actually it looks like this fix broke something in 3.7, we're getting bug reports already. Might be breaking specifically when using multithreading, which wouldn't apply here, but we haven't got a reproducible report yet so it could be worse than we think. Will chase it down over the next few days. And will definitely keep this on the "must document" list...

ronlevine:
@ldgauthier `in production (GATK3)I'm pretty sure we make sure we don't break exome intervals and for genomes we won't break active regions either`, is this documented anywhere? I am reading the scheduling code in GATK3 and am trying to understand how intervals are broken up are processed per thread.

ldgauthier:
For production we use a different execution engine (Zamboni, not Queue).  The number of scatters is a static part of the workflow and Zamboni has the responsibility of splitting the interval list per scatter.

vdauwera:
Yeah that's one of the reasons we didn't catch this (and one other thread safety bug), we don't run multithreaded in prod and don't test for it systematically.

ronlevine:
OK. In GATK3, the sharding size is calculated in `GenomeAnalysisEngine.getShardStrategy`. Since `GenotypeGVCFs` is a RodWalker and does not use input reads (BAM file), the shard size is `1,000,000`. 
This might be more than a thread safety bug (which is easy to fix). What worries me is If the cache of upstream deletions spans intervals, this code will not work since the processing is asynchronous. For example, if there are 2 threads and the removed deletion crosses the shard barrier and the downstream interval thread is first to process, it will not see the upstream removed deletion.

vdauwera:
Ah, crap. If it's really bad (ie not easily fixable) we may be limited to disabling multithreading for GenotypeGVCFs. We can't devote much time to fixing up GATK3 anymore. @droazen would you care to opine?

droazen:
@vdauwera Does anyone use `-nt` with `GenotypeGVCFs` in production or in the wild? If not, we could make it no longer implement `TreeReducible`. Agree that non-trivial multithreading-related fixes to the GATK3 engine are probably not worth it at this point.

vdauwera:
Lots of people do, unfortunately. I'm reluctant to just disable multithreading entirely, even though it kind of is the right thing to do. 

Another option, potentially, is to allow users to disable the function that strips out the no-longer-needed spanning deletions, which is what is breaking here afaiu. It would just let through a handful of records that are unnecessary but they're not wrong. So it's the least-harm solution.

Issue #2211, opened 2016-10-13
Hook up GATKDocs system to GATK4 via a build target that builds documentation


droazen:

cmnbroad:
Closed via #2327.

PR #2327, opened 2016-12-22
Add documentation generation

cmnbroad:
Adds support for online documentation generation through Barclay (https://github.com/broadinstitute/gatk/issues/2211) via the gatkDoc gradle task. The first commit contains only annotation updates (to target selected classes as documentation targets). A more complete audit/update pass will need to be done; but we need some for now in order to be able to exercise the doc generation process. The second commit contains that actual code and templates for documentation, and the final one upgrades to a Barclay snapshot that has the necessary dependent classes.


build.gradle
```diff
@@ -94,7 +95,9 @@ compileTestJava {
 }
 
 dependencies {
-    compile 'org.broadinstitute:barclay:1.0.0'
+    compile 'org.broadinstitute:barclay:1.0.0-2-gb718440-SNAPSHOT'
+    // javadoc utilities
+    compile(files(((URLClassLoader) ToolProvider.getSystemToolClassLoader()).getURLs()))
```
lbergelson: Should this be `compileOnly`?

build.gradle
```diff
@@ -372,13 +375,43 @@ task sourcesJar(type: Jar) {
...
-        options.addStringOption('Xdoclint:none', '-quiet')
+        // do this for all javadoc tasks, inlcuding gatkDocs
+        options.addStringOption('Xdoclint:none')
     }
 }
 
+javadoc {
+    // This is a hack to disable the java 8 default javadoc lint until we fix the html formatting
+    // We only want to do this for the javadoc task, not gatkDoc
+    options.addStringOption('Xdoclint:none', '-quiet')
```
lbergelson: should this just be `-quite` here since `Xdoclint:non` is already included for all javadoc tasks?

build.gradle
```diff
@@ -372,13 +375,43 @@ task sourcesJar(type: Jar) {
...
+
+// Generate GATK Online Doc
+task gatkDoc(type: Javadoc, dependsOn: shadowJar) {
+    source = sourceSets.main.allJava
+    classpath = sourceSets.main.runtimeClasspath
+
+    // Since the custom doclet we run is basically another entry point for gatk, it needs all of the gatk
+    // dependencies on the docletpath (i.e., Barclay , etc), so we depend on and use the shadowJar. Also,
+    // add the resources folder to ensure log4j is configured.
+    options.docletpath = [file(shadowJar.archivePath)] + files("src/main/resources")
```
lbergelson: I don't think I understand why docletpath can't take the compileTime classpath as it's path.  Doesn't that include all the dependencies?

build.gradle
```diff
@@ -372,13 +375,43 @@ task sourcesJar(type: Jar) {
...
+    options.doclet = "org.broadinstitute.hellbender.utils.help.GATKHelpDoclet"
+
+    // ensure the target folder already exists
+    File outDir = new File("./gatkDoc")
+    outDir.mkdirs()
```
lbergelson: should we handle the case where this fails and returns false?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSparkTool.java
```diff
@@ -27,11 +31,12 @@
...
  */
 public abstract class MetricsCollectorSparkTool<T extends MetricsArgumentCollection>
-        extends GATKSparkTool {
+        extends GATKSparkTool
+        implements CommandLinePluginProvider {
 
     private static final long serialVersionUID = 1l;
 
-    /**
+    /**nR
```
lbergelson: There are some random letters here.

src/main/java/org/broadinstitute/hellbender/cmdline/SpecialArgumentsCollection.java
```diff
@@ -17,13 +17,13 @@
     public static final String ARGUMENTS_FILE_FULLNAME = "arguments_file";
```
lbergelson: It's weird that this is in barclay and gatk.  I think this should either be deleted in gatk, or it should be deleted in barclay and this should be refactored to use special argument names that are exposed in barclay.

src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
```diff
@@ -8,4 +8,22 @@ public static String forumPost(String post) {
         return GATK_FORUM_URL + post;
     }
 
+    /**
+     * Definition of the group names / categories of tools.
+     * The names get parsed to make supercategories in the doc index,
+     * so be careful when making big changes -- see GATKDoclet.java toMap()
```
lbergelson: That's kind of gross...  maybe this should be an enum instead?
Relatedly, these categories are weirdly redundant with but not concordant with ProgramGroups.

vdauwera: This was tacked-on for the purpose of organizing docs on the website. I don't remember having ProgramGroups in the old GATK system -- where are those defined?

src/main/java/org/broadinstitute/hellbender/tools/spark/CreateHadoopBamSplittingIndex.java
```diff
@@ -32,6 +34,7 @@
 @CommandLineProgramProperties(summary = "Create a HadoopBam splitting index as well as a bam index from a bam file.",
         oneLineSummary = "create a hadoop-bam splitting index" ,
         programGroup = SparkProgramGroup.class)
+@DocumentedFeature(groupName= HelpConstants.DOCS_CAT_DATA)
```
lbergelson: Should this be DOCS_CAT_SPARK?  How do we decide what goes in SPARK and what goes in DATA?

src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
```diff
@@ -2,21 +2,29 @@
...
+import org.broadinstitute.hellbender.utils.help.HelpConstants;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
 import org.broadinstitute.hellbender.utils.read.SAMFileGATKReadWriter;
 
 import java.io.File;
 
+/**
+ * Print reads from a SAM/BAM/CRAM file.
+ *
+ * {@GATK.walkertype ReadWalker}
```
lbergelson: Is this part of the doc system or just a comment?

src/main/java/org/broadinstitute/hellbender/tools/spark/CreateHadoopBamSplittingIndex.java
vdauwera: Needs review to update the categorization logic. Would recommend leaving that for a separate issue/PR.

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
```diff
@@ -21,7 +22,7 @@
  * ReadWalker authors must implement the apply() method to process each read, and may optionally implement
  * onTraversalStart() and/or onTraversalSuccess(). See the PrintReadsWithReference walker for an example.
  */
-public abstract class ReadWalker extends GATKTool {
+public abstract class ReadWalker extends GATKTool implements CommandLinePluginProvider {
```
lbergelson: Is there a reason `CommandLineProgram` doesn't implement this?  It seems like it should since it implements `getPluginDescriptors()`?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSparkTool.java
```diff
@@ -52,6 +57,15 @@
     public final boolean requiresReads(){ return true; }
 
     /**
+     * Return the list of GATKCommandLinePluginDescriptors to be used for this tool.
+     * Uses the read filter plugin.
+     */
+    @Override
+    public List<? extends CommandLinePluginDescriptor<?>> getPluginDescriptors() {
```
lbergelson: I think this is redundant with the definition in `GATKSparkTool`

src/main/java/org/broadinstitute/hellbender/utils/help/GATKHelpDoclet.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    @Override
+    protected final Map<String, String> getGroupMap(final DocumentedFeatureObject documentedFeatureObject) {
+        final Map<String, String> root = super.getGroupMap(documentedFeatureObject);
+
+        /**
+         * Add-on super-category definitions. The assignments depend on parsing the names
+         * defined in HelpConstants.java so be careful of changing anything.  The super-category
+         * value strings need to be the same as used in the Freemarker template.
+         */
```
lbergelson: This seems very brittle.

vdauwera: It is indeed -- it's also a duct tape patch I added years ago for organizing content on the website. Would be delighted to have something better, so suggestions >> welcome.

src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
lbergelson: They came from the picard help system.  Maybe we can unify the two concepts.

vdauwera: That would make a lot of sense, I agree

build.gradle
cmnbroad: Yes, thats a good idea. I realized today that now its including tools.jar in the uber jar, which we can't have and don't want.

build.gradle
cmnbroad: I originally tried setting the docletpath the way [Picard](https://github.com/broadinstitute/picard/blob/master/build.gradle#L232) does, which I think is similar to what you're suggesting, and had other problems. compileTime classpath seems to have all of the *dependencies*, but the doclet class itself isn't found that way. However, I just tried reverting to runtimeClassPath and that seems to work, so I'll just need to update this to manually include the tools.jar file.

build.gradle
cmnbroad: I would think so, and I originally had it that way; the gatkDoc task worked like that, but it breaks the javadoc task (it behaves as if lint is on). I'm not at all clear why.

build.gradle
```diff
@@ -372,13 +375,43 @@ task sourcesJar(type: Jar) {
     classifier = 'sources'
 }
 
-// This is a hack to disable the java 8 default javadoc lint until we fix the html formatting
 if (JavaVersion.current().isJava8Compatible()) {
```
cmnbroad: BTW, what do you think about removing this line. It seems unnecessary.

lbergelson: Remove!

build.gradle
cmnbroad: Makes sense. I added an existence test, and then throw if we can't create it. I also had to move the creation code to a taskGraph whenReady event, since otherwise the folder is created every time gradle runs the script.

lbergelson: Oh, yeah, that's a good point, you don't want it running on initialization.  You could also move it to a doFirst.  When does a whenReady fire?

cmnbroad: It fires after the configuration phase is complete and the entire task graph is determined, so you can test if a particular task will actually execute:

```
gradle.taskGraph.whenReady { graph ->
    if (graph.hasTask(gatkDoc) && !gatkDocDir.exists() && !gatkDocDir.mkdirs()) {
        throw new GradleException("Can't create output folder for gatkDoc task: " + gatkDocDir.absolutePath)
    }
}
```
I think doFirst makes more sense though, even if it is a little weird that the result is a closure with a doFirst block that executes AFTER the part of the closure that is NOT in the doFirst block!):

```
task gatkDoc {
    doFirst {
        System.out.println("This doesn't happen first")
    }
    System.out.println("This happens first")
}
```

lbergelson: Heh, it is a bit weird indeed.

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
cmnbroad: You're right. Its going to wind up there anyway when the PR to move up read filters plugin goes in, so I'll move it there now.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSparkTool.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/SpecialArgumentsCollection.java
cmnbroad: Yes, this is an oversight. This should  just be in Barclay.

src/main/java/org/broadinstitute/hellbender/tools/spark/CreateHadoopBamSplittingIndex.java
cmnbroad: I made up all these changes (the annotations, groups, etc.) and kept them in a separate commit since I needed a range of interesting cases (i.e., Spark tools, Walkers, Picard tools, ReadFilters) to use to exercise the code, but thinking we might not actually take them in this PR. We could leave them out, but without them, the doc process doesn't output anything, and they at least provide a jump start.

The question of how we decide what goes in Spark is an interesting one. We really have only one dimension of partitioning (if you don't count supercat - I'll save that for the other comment), but we might want more than one dimension so we can have Spark/NonSpark, and within those, the normal categories. As @vdauwera said, I think its worth doing thinking a bit about how we want to organize the doc before we go too much further.

src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
cmnbroad: Its part of the doc system. The doc system queries the doclet for a custom prefix ("GATK" in this case) and then pulls any embedded custom javadoc tags with that prefix up into the Freemarker map. It has some advantages and some disadvantages, but it allows the doc to be changed without modifying or depending on code. For now, its how we're adding the "walkertype" category to the doc.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/MetricsCollectorSparkTool.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/CreateHadoopBamSplittingIndex.java
vdauwera: You make a good point on dimensions, @cmnbroad. I wonder if perhaps the spark/non-spark nature can be captured in the walkertype? If so that would give us all we need to enable listing things based on arbitrary logic in webpages. E.g. we can have additional templates that organize tools differently and produce corresponding webpages; then we can present these organization options to users through clickable widgets.

src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
cmnbroad: Although there is some redundancy, I don't think we can completely unify them, since they don't completely intersect. There are documented features that are not CommandLinePrograms (like ReadFilters, and perhaps others such as transformers, annotations, etc), as well as vice versa. In order to eliminate much of the redundancy, the doc system uses the DocumentedFeature annotation as a tag interface to tag a class as a documentation target. When it creates the doc, it prefers to use the DocumentedFeature annotation values, but if they're not present, it will fall back to the  CommandLineProgramGroup ones if they're present. So most CommandLinePrograms just need to have an empty DocumentedFeature tag annotation to enable the doc. ReadFilters would want a fully populated DocumentedFeature annotation, since there will be no fallback.

cmnbroad: Having said all that, I just noticed a bug in Barclay where it doesn't always do this properly.

cmnbroad: BTW, there is a [doc](https://github.com/broadinstitute/barclay/wiki/Using-the-Documentation-Generation-utilities-in-Barclay) (under construction) in the Barclay repo that describes some of this.

src/main/java/org/broadinstitute/hellbender/tools/spark/CreateHadoopBamSplittingIndex.java
cmnbroad: @vdauwera Yes, we could use the custom tag (or another custom tag) for that, and it would work for online doc. But we'd still have the same problem with command line help output, which (currently) organizes the output by CommandLineProgramGroup and doesn't know anything about Javadoc tags or DocumentedFeature annotations. Of course, we could change that.

src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
vdauwera: Oh right, I think you explained this to me at the retreat but it didn't stick, sorry.

src/main/java/org/broadinstitute/hellbender/tools/spark/CreateHadoopBamSplittingIndex.java
vdauwera: Ah, to be frank I'm less concerned about the appearance of the cli help output, as long as everything is listed there that should be. If people want more elaborately organized content they should see the online docs, that's what they're for.

cmnbroad: @vdauwera @lbergelson Ok then, I think we can pretty easily achieve what we want by just consolidating ALL of the group names and descriptions (including the ones that are currently embedded in CommandLineProgramGroup derived classes) into HelpConstants, and then just writing a method there that maps a program group string to a "sparky" property and a "supercat" property. Then all the mapping code will be in one place, and we can remove the gross code in the custom doclet that currently parses the names to get a supercat, and replace it with a call to the new method. That, in combination with a change to Barclay to fallback to get the group name from CommandLineProgramProperties (I did only did this for summary but should do it for group name as well), I think will get us what we want.

If we agree on that, I'll make those changes.

vdauwera: That sounds reasonable to me

cmnbroad:
This last set of changes is dependent on https://github.com/broadinstitute/barclay/pull/17. Once that's reviewed we'll need to snapshot or release and upgrade to Barclay.

cmnbroad:
I separated out the code review updates from the annotation updates for the code review round. So the last 3 commits represent the code review changes, the corresponding annotation changes, and the Barclay snapshot upgrade, respectively).

cmnbroad:
Also, the results of generating doc using "gradle gatkDoc" on this build can be seen at ~cnorman/gatkDoc/index.html.

lbergelson:
@cmnbroad This looks good to me now.  Are we waiting on barclay release to merge this or should it go in now?

cmnbroad:
@lbergelson These changes are dependent on a Barclay [PR](https://github.com/broadinstitute/barclay/pull/17) that is not merged yet, so we should at least wait for that snapshot. Ideally, the other two (tiny) Barclay PRs in the queue should also be reviewed/merged, then we could do a release and upgrade to that.

cmnbroad:
@lbergelson When you get a chance. I think it was your last review that pushed me over the edge to refactor the barclay docgen code. So this now has additional commits that reflect that. One has the changes based on your general code review comments; one has the substantial changes due to refactoring (small changes to just 3 files I think); one has the annotation changes based on that; and one has the change to build.gradle for the barclay snapshot, which I need on this branch but will remove once I rebase. My apologies if its confusing. Le me know if you have questions.

lbergelson:
@cmnbroad The changes look good to me.  👍

lbergelson:
Is this waiting on a barclay proper release or should we merge with the snapshot?

cmnbroad:
No, gatk is already ahead of the Barclay snapshot this branch uses, so I need to remove the barclay snapshot referenced in build.gradle in this branch; then rebase and squash and resolve conflicts, then we can merge. Thanks!

PR #2329, opened 2017-01-03
Making public Transition

LeeTL1220:
Would like to make the Transition enum a public enum.


lbergelson:
@LeeTL1220 👍 

You're going to need to do the same pr against picard in the near future because this copy of transition is going away once we switch.

PR #2330, opened 2017-01-03
Utils methods to stream Iterable/Iterator

davidbenjamin:
Because calling `StreamSupport.stream(iterable.spliterator(false))` is really ugly.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/FeatureWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/FeatureWalker.java
@@ -4,6 +4,7 @@ import htsjdk.tribble.Feature;
 import htsjdk.tribble.FeatureCodec;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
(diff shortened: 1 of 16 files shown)


davidbenjamin:
@SHuang-Broad could you take a look at this?

SHuang-Broad:
Sorry for the late noticing. Will work on it now.

SHuang-Broad:
Looks good to me, except one concern about the use of `0` as the characteristics when streaming iterators (unset all bits in the characteristics of the returned spliterator). What about doing it in the way as suggested by accepted answer [here](http://stackoverflow.com/questions/24511052/)?

davidbenjamin:
@SHuang-Broad done (assuming tests pass).  Is this what you meant?

SHuang-Broad:
Sorry for the confusion, but assuming the iterator is iterating through an ordered container is not safe, I think. What about this:

```java
public static <T> Stream<T> stream(final Iterator<T> iterator) {
    return stream(() -> iterator);
}
```

It is reusing the version working with `Iterable`'s.

davidbenjamin:
@SHuang-Broad done!

SHuang-Broad:
👍

PR #2331, opened 2017-01-06
Add cloud prefetching to GATK

jean-philippe-martin:
via htsjdk's new wrapper feature.
Also provide a command-line switch to tune or disable it if necessary.

A test with CountReads on a ~900MB input shows a 40MB buffer
gives over 5x speedup.

DO NOT SUBMIT until htsjsk's new version is released
that incorporates the [wrapper feature](https://github.com/samtools/htsjdk/pull/775).
Then, update the build file before submitting.

Sample run:

$ ./gatk-launch CountReads -I "gs://${INPUTFOLDER}/CEUTrio.HiSeq.WGS.b37.ch20.4m-12m.NA12878.bam" --cloudPrefetchBuffer=0
(...)
org.broadinstitute.hellbender.tools.CountReads done. Elapsed time: 2.82 minutes.
$ ./gatk-launch CountReads -I "gs://${INPUTFOLDER}/CEUTrio.HiSeq.WGS.b37.ch20.4m-12m.NA12878.bam" --cloudPrefetchBuffer=40
(...)
org.broadinstitute.hellbender.tools.CountReads done. Elapsed time: 0.49 minutes.

cc: @lbergelson @droazen


jean-philippe-martin:
p.s. it's normal this doesn't build until we have the new htsjdk and I point the build file to it.

lbergelson:
@jean-philippe-martin There are snapshot releases of htsjdk published automatically at artifactory.broadinstitute.org ! 

It should work with version: 2.8.1-6-g2acb88b-20170106.162419-1

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -129,6 +137,14 @@ void initializeReference() {
         reference = referenceArguments.getReferenceFile() != null ? ReferenceDataSource.of(referenceArguments.getReferenceFile()) : null;
     }
 
+    private static SeekableByteChannel addPrefetcher(int cloudPrefetchBuffer, SeekableByteChannel x) {
+        try {
+            return new SeekableByteChannelPrefetcher(x, cloudPrefetchBuffer * 1024 * 1024);
+        } catch (IOException ex) {
+            throw new RuntimeException(ex);
```
droazen: `GATKException`

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -137,7 +153,14 @@ void initializeReference() {
      */
     void initializeReads() {
         if (! readArguments.getReadFiles().isEmpty()) {
-            SamReaderFactory factory = SamReaderFactory.makeDefault().validationStringency(readArguments.getReadValidationStringency());
+            // Prefetcher is useful for cloud files because of the latencies involved.
+            boolean gsPathPresent = readArguments.getReadPaths().stream()
+                .filter(p->p.toUri().getScheme().equals("gs")).findAny().isPresent();
```
droazen: What happens if there's a mix of GCS and non-GCS inputs? Can we avoid wrapping the non-GCS inputs? I think this wrapping would be better done in `ReadsDataSource`, which could explicitly call `SamInputResource.of(path, wrapper)` as appropriate on each input.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -167,7 +170,7 @@ public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices,
                 reader = samReaderFactory.open(samPath);
             }
             else {
-                final SamInputResource samResource = SamInputResource.of(samPath);
+                final SamInputResource samResource = SamInputResource.of(samPath, Function.identity());
```
droazen: This change is unnecessary, since the overload that takes just a `Path` does the same thing. Can you remove?

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jean-philippe-martin: In the case of no index, though, we'd need a SamReaderFactory.open(path, wrapper), which doesn't currently exist.

I need to investigate a bit more, but for some reason my IDE is suddenly unable to compile the code. I'm getting "Unable to resolve com.github.samtools:htsjdk:2.8.1-9-ged15dc0-20170110.192756-1", as if it's unable to find the artifactory.

It compiles fine from the command line, which suggests this may be a bug in the IDE.

jean-philippe-martin: While we're at it we may want to handle the non-index case, though that requires a change to htsjdk (samReaderFactory.open(samPath, wrapper))

jean-philippe-martin:
OK, back to you! 

I've fixed the issues you pointed out and added optional prefetching for index data too. I think this fixes the problem you saw with slow printreads.

This needs the corresponding changes in htsjdk though.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -137,6 +157,16 @@ void initializeReference() {
...
+            // Prefetcher is useful for cloud files because of the latencies involved.
+            Function<SeekableByteChannel, SeekableByteChannel> wrapper = Function.identity();
+            Function<SeekableByteChannel, SeekableByteChannel> indexWrapper = Function.identity();
+
+            if (cloudPrefetchBuffer > 0) {
+                wrapper = is -> GATKTool.addPrefetcher(cloudPrefetchBuffer, is);
+            }
+            if (cloudIndexPrefetchBuffer > 0) {
+                indexWrapper = is -> GATKTool.addPrefetcher(cloudIndexPrefetchBuffer, is);
+            }
```
droazen: Let's move all this code to construct the prefetchers (including `addPrefetcher()` above and this code snippet here) down into `ReadsDataSource`, and just pass the buffer sizes as ints into the `ReadsDataSource` constructor.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -79,6 +84,13 @@
     @Argument(fullName = StandardArgumentDefinitions.ADD_OUTPUT_SAM_PROGRAM_RECORD, shortName = StandardArgumentDefinitions.ADD_OUTPUT_SAM_PROGRAM_RECORD, doc = "If true, adds a PG tag to created SAM/BAM/CRAM files.", optional=true)
     public boolean addOutputSAMProgramRecord = true;
 
+    // default value of 40MB based on a test with CountReads (it's 5x faster than no prefetching)
+    @Argument(fullName = StandardArgumentDefinitions.CLOUD_PREFETCH_BUFFER_LONG_NAME, shortName = StandardArgumentDefinitions.CLOUD_PREFETCH_BUFFER_SHORT_NAME, doc = "Size of the cloud-only prefetch buffer (in MB; 0 to disable).", optional=true)
+    public int cloudPrefetchBuffer = 40;
+
+    @Argument(fullName = StandardArgumentDefinitions.CLOUD_INDEX_PREFETCH_BUFFER_LONG_NAME, shortName = StandardArgumentDefinitions.CLOUD_INDEX_PREFETCH_BUFFER_SHORT_NAME, doc = "Size of the cloud-only prefetch buffer (in MB; 0 to disable).", optional=true)
+    public int cloudIndexPrefetchBuffer = 5;
```
droazen: Does the index require a different/lower default for performance reasons? Did you test with both 40 and 5 here? It would be nice to combine these arguments into one, unless the performance is dependent on these values being different.

build.gradle
```diff
@@ -54,7 +54,12 @@ repositories {
 
 }
 
-final htsjdkVersion = System.getProperty('htsjdk.version','2.8.0')
+// TODO: switch to a non-snapshot version
+//final htsjdkVersion = System.getProperty('htsjdk.version','2.8.1-6-g2acb88b-20170106.162419-1')
+///final htsjdkVersion = System.getProperty('htsjdk.version','2.8.1-9-ged15dc0-20170110.192756-1')
+// the version I have locally because intellij is having difficulty with the artifactory
+final htsjdkVersion = '2.8.1-4-g087f83d-SNAPSHOT'
```
droazen: I've reviewed the htsjdk piece of this.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -145,7 +175,8 @@ else if (hasCramInput()) {
                 throw new UserException.MissingReference("A reference file is required when using CRAM files.");
             }
 
-            reads = new ReadsDataSource(readArguments.getReadPaths(), readArguments.getReadIndexPaths(), factory);
+            // TODO: consider using a smaller buffer when wrapping the index
```
droazen: Seems like you are already doing this?

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -134,7 +137,10 @@ public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices )
      * @param customSamReaderFactory SamReaderFactory to use, if null a default factory with no reference and validation
      *                               stringency SILENT is used.
      */
-    public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices, SamReaderFactory customSamReaderFactory ) {
+    public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices,
+            SamReaderFactory customSamReaderFactory,
+            Function<SeekableByteChannel, SeekableByteChannel> gcsWrapper,
+            Function<SeekableByteChannel, SeekableByteChannel> cloudIndexWrapper) {
```
droazen: javadoc for the new constructor args

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -155,6 +161,16 @@ public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices,
         int samCount = 0;
         for ( final Path samPath : samPaths ) {
             // Ensure each file can be read
+            Function<SeekableByteChannel, SeekableByteChannel> wrapper = Function.identity();
+            Function<SeekableByteChannel, SeekableByteChannel> indexWrapper = Function.identity();
+            if (samPath.toUri().getScheme().equals("gs")) {
```
droazen: Use `BucketUtils.isCloudStorageUrl()` instead (add a new overload of it that takes a `Path`, plus a unit test)

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -164,11 +180,16 @@ public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices,
...
             if ( samIndices == null ) {
-                reader = samReaderFactory.open(samPath);
+                reader = samReaderFactory.open(samPath, wrapper, indexWrapper);
             }
             else {
-                final SamInputResource samResource = SamInputResource.of(samPath);
-                samResource.index(samIndices.get(samCount));
+                final SamInputResource samResource = SamInputResource.of(samPath, wrapper);
+                Path indexPath = samIndices.get(samCount);
+                if (!samPath.toUri().getScheme().equals("gs")) {
```
droazen: `BucketUtils.isCloudStorageUrl()`

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -155,6 +161,16 @@ public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices,
...
+            Function<SeekableByteChannel, SeekableByteChannel> wrapper = Function.identity();
+            Function<SeekableByteChannel, SeekableByteChannel> indexWrapper = Function.identity();
+            if (samPath.toUri().getScheme().equals("gs")) {
+                if (gcsWrapper != null) {
+                    wrapper = gcsWrapper;
+                }
+                if (cloudIndexWrapper != null) {
+                    indexWrapper = cloudIndexWrapper;
+                }
+            }
```
droazen: Lines 164-173 could be collapsed into two lines using `? :` statements, which I think would help readability. Eg., 

    wrapper = BucketUtils.isCloudStorageUrl(samPath) && gcsWrapper != null ? gcsWrapper : Function.identity();

And similarly for the index.

src/test/java/org/broadinstitute/hellbender/engine/ReadsDataSourceUnitTest.java
```diff
@@ -573,7 +573,7 @@ public void testManuallySpecifiedIndices( final List<Path> bams, final List<Path
     public void testManuallySpecifiedIndicesWithCustomReaderFactory( final List<Path> bams, final List<Path> indices ) {
         final SamReaderFactory customFactory = SamReaderFactory.makeDefault().validationStringency(ValidationStringency.STRICT);
 
-        try ( final ReadsDataSource readsSource = new ReadsDataSource(bams, indices, customFactory) ) {
+        try ( final ReadsDataSource readsSource = new ReadsDataSource(bams, indices, customFactory, null, null) ) {
```
droazen: Revert this change, and add a new, separate unit test for the five-arg constructor that passes in dummy wrappers of some kind and assert that they've been applied.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jean-philippe-martin: Ah yes, deleting comment.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jean-philippe-martin: I haven't run tests yet, so this value is just a guess based on the fact that the indices I have are small.

You make a good point that having a single argument is easier. So let's allow the user to either only set cloudPrefetchBuffer (in which case this value is used for both the data and index), or to set both (if they want to optimize further; after all it would be a great concidence if the best value were the same for both data types). As a side effect this currently makes the default value be the same for both (we can change this later if needed).

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jean-philippe-martin: This would make it little bit harder to experiment with different types of wrapper (like adding one that caches more but does less prefetching). Are you sure? There is no other usage of the ReadsDataSource constructor aside from tests so there aren't many clients to benefit from that simplification.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
jean-philippe-martin: done

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
jean-philippe-martin: done

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
jean-philippe-martin: done

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
jean-philippe-martin: done

src/test/java/org/broadinstitute/hellbender/engine/ReadsDataSourceUnitTest.java
jean-philippe-martin: It doesn't compile without this change. I'm guessing you mean you want me to add a 3-arg constructor?

The inputs here are not on the Google Cloud, so the wrapper should be ignored. To check this, my new test passes a wrapper that returns null (if the code tries to use that, it'll blow up). The test passes.

jean-philippe-martin: To be thorough I have also added a test with files on the Google Cloud (I've added them both there and on the repo) where the files are XOR'd with a constant, and the wrapper undoes this. This detects if we ignore the wrapper or use the wrong one.

jean-philippe-martin:
Ready for you! I made it depend on the updated htsjdk (snapshot) and Travis is now happy.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
droazen: Yes, please move it all down into `ReadsDataSource`, and have the `ReadsDataSource` constructors take the buffer sizes as ints. There are clients in gatk-protected that use `ReadsDataSource` directly and would benefit, and in general we want to encapsulate these sort of low-level details in the data source classes as much as possible, and not bleed into the tool hierarchy.

build.gradle
```diff
@@ -54,7 +54,11 @@ repositories {
 
 }
 
-final htsjdkVersion = System.getProperty('htsjdk.version','2.8.0')
+// TODO: switch to a non-snapshot version
+// (it's snapshot now because we rely on not-yet-released changes to htsjdk)
+//final htsjdkVersion = System.getProperty('htsjdk.version','2.8.1-6-g2acb88b-20170106.162419-1')
```
droazen: No need to add these comments -- we are often on an htsjdk snapshot as a project.

build.gradle
```diff
@@ -86,11 +90,14 @@ jacocoTestReport {
 }
 
 //NOTE: we ignore contracts for now
+// TEMPORARY HACK: JP removed -Werror because
+// I'm using the latest version of htsjdk and get a depreciation warning
+// (presumably the latest version of gatk fixes this)
```
droazen: Let's not do this -- I'll open a separate PR against master to fix any deprecation issues now and move master to the same snapshot you're using, and when you rebase be sure to restore `-Werror`

droazen: PR is merged, hack can be removed once you rebase.

jean-philippe-martin:
Squashed, rebased. Let's see if anything broke in the process.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -183,6 +195,14 @@ void initializeReference() {
         reference = referenceArguments.getReferenceFile() != null ? ReferenceDataSource.of(referenceArguments.getReferenceFile()) : null;
     }
 
+    private static SeekableByteChannel addPrefetcher(int cloudPrefetchBuffer, SeekableByteChannel x) {
+        try {
+            return new SeekableByteChannelPrefetcher(x, cloudPrefetchBuffer * 1024 * 1024);
+        } catch (IOException ex) {
+            throw new GATKException("Unable to initialize the prefetcher: " + ex);
+        }
+    }
```
droazen: As mentioned in another comment that might have gotten buried/rebased away, please do move all this code in `GATKTool` down into `ReadsDataSource`, and have the `ReadsDataSource` constructors take the buffer sizes as ints. There are clients in gatk-protected that use `ReadsDataSource` directly and would benefit, and in general we want to encapsulate these sort of low-level details in the data source classes as much as possible, and not bleed into the tool hierarchy.

build.gradle
```diff
@@ -86,11 +86,14 @@ jacocoTestReport {
 }
 
 //NOTE: we ignore contracts for now
+// TEMPORARY HACK: JP removed -Werror because
```
droazen: Hack still appears to be present even after the rebase?

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -84,6 +89,13 @@
     @Argument(fullName = StandardArgumentDefinitions.ADD_OUTPUT_SAM_PROGRAM_RECORD, shortName = StandardArgumentDefinitions.ADD_OUTPUT_SAM_PROGRAM_RECORD, doc = "If true, adds a PG tag to created SAM/BAM/CRAM files.", optional=true)
     public boolean addOutputSAMProgramRecord = true;
 
+    // default value of 40MB based on a test with CountReads (it's 5x faster than no prefetching)
+    @Argument(fullName = StandardArgumentDefinitions.CLOUD_PREFETCH_BUFFER_LONG_NAME, shortName = StandardArgumentDefinitions.CLOUD_PREFETCH_BUFFER_SHORT_NAME, doc = "Size of the cloud-only prefetch buffer (in MB; 0 to disable).", optional=true)
+    public int cloudPrefetchBuffer = 40;
+
+    @Argument(fullName = StandardArgumentDefinitions.CLOUD_INDEX_PREFETCH_BUFFER_LONG_NAME, shortName = StandardArgumentDefinitions.CLOUD_INDEX_PREFETCH_BUFFER_SHORT_NAME, doc = "Size of the cloud-only prefetch buffer (in MB; 0 to disable). Defaults to cloudPrefetchBuffer if unset.", optional=true)
+    private int cloudIndexPrefetchBuffer = -1;
```
droazen: Should we be more specific and use "gcs" instead of "cloud" in these argument names?

jean-philippe-martin:
I'm trying this, but it breaks the test, because the test relies on being
able to set a not-actually-buffering wrapper. As a compromise I'd adding
both methods, so the convenience method is available but the tests can also
do their thing.

build.gradle
jean-philippe-martin: I know, sorry, fixed that since but haven't pushed yet because I'm moving code around.

droazen:
You mean you're adding both methods to `ReadsDataSource` @jean-philippe-martin? That seems fine -- I mainly care about keeping `GATKTool` clean.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jean-philippe-martin: I was thinking it's better to avoid acronyms. And the beauty of this thing is that should we ever decide to extend the code to be able to use a different cloud provider (not that Google isn't awesome, of course), the wrappers and everything will still work, we just need to update BucketUtils.isCloudStorageURL to the new provider and *bam* we get prefetching and caching for this other one. 

Prefetching is useful here because of the round-trip latencies involved since the data is not on the same computer as the processing. It's likely to remain useful when running on a different cloud provider (though of course the exact optimal value may change a bit).

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -164,11 +192,16 @@ public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices,
...
             if ( samIndices == null ) {
-                reader = samReaderFactory.open(samPath);
+                reader = samReaderFactory.open(samPath, wrapper, indexWrapper);
             }
             else {
-                final SamInputResource samResource = SamInputResource.of(samPath);
-                samResource.index(samIndices.get(samCount));
+                final SamInputResource samResource = SamInputResource.of(samPath, wrapper);
+                Path indexPath = samIndices.get(samCount);
+                if (!BucketUtils.isCloudStorageUrl(samPath)) {
```
droazen: So if the `indexPath` is on GCS and the `samPath` is not, we don't want to use an index wrapper? Can you explain why?

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -54,6 +54,10 @@ public static boolean isCloudStorageUrl(final String path) {
         return path.startsWith(GCS_PREFIX);
     }
 
+    public static boolean isCloudStorageUrl(final java.nio.file.Path path) {
+        return (path.toUri().getScheme() + "://").equals(GCS_PREFIX);
```
droazen: Seems like `getScheme()` can return null, so we should guard against that possibility.

build.gradle
jean-philippe-martin: Thank you!

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jean-philippe-martin: yes I know, as I said this required a bit more work than expected because doing it this way broke the tests. It's in now.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
droazen: Good points, let's keep the existing names then.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -191,15 +203,16 @@ void initializeReference() {
      */
     void initializeReads() {
         if (! readArguments.getReadFiles().isEmpty()) {
+            // Prefetcher is useful for cloud files because of the latencies involved.
+
```
droazen: Stray comment left over from before the rebase should be deleted or moved to an appropriate place.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
jean-philippe-martin: I'm pretty sure this is a relic of the intermediate state when we had a single wrapper function for both. I'm deleting these lines.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -191,15 +203,16 @@ void initializeReference() {
...
             SamReaderFactory factory = SamReaderFactory.makeDefault().validationStringency(readArguments.getReadValidationStringency());
             if (hasReference()) { // pass in reference if available, because CRAM files need it
                 factory = factory.referenceSequence(referenceArguments.getReferenceFile());
             }
             else if (hasCramInput()) {
                 throw new UserException.MissingReference("A reference file is required when using CRAM files.");
             }
-
-            reads = new ReadsDataSource(readArguments.getReadPaths(), readArguments.getReadIndexPaths(), factory);
+            reads = new ReadsDataSource(readArguments.getReadPaths(), readArguments.getReadIndexPaths(), factory, cloudPrefetchBuffer, getCloudIndexPrefetchBuffer());
```
droazen: Replace `getCloudIndexPrefetchBuffer()` call with inline `cloudIndexPrefetchBuffer < 0 ? cloudPrefetchBuffer : cloudIndexPrefetchBuffer`, and delete that method.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
jean-philippe-martin: added "" +

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jean-philippe-martin: removed

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -134,7 +138,49 @@ public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices )
...
+         * and a custom SamReaderFactory.
+         *
+         * @param samPaths paths to SAM/BAM/CRAM files, not null
+         * @param samIndices indices for all of the SAM/BAM/CRAM files, in the same order as samPaths. May be null,
+         *                   in which case index paths are inferred automatically.
+         * @param customSamReaderFactory SamReaderFactory to use, if null a default factory with no reference and validation
+         *                               stringency SILENT is used.
+         * @param cloudWrapper caching/prefetching wrapper for the data, if on Google Cloud.
+         * @param cloudIndexWrapper caching/prefetching wrapper for the index, if on Google Cloud.
+         */
```
droazen: Fix indent on comment block

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jean-philippe-martin: As you wish.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -134,7 +138,49 @@ public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices )
...
+     * @param cloudPrefetchBuffer MB size of caching/prefetching wrapper for the data, if on Google Cloud (0 to disable).
+     * @param cloudIndexPrefetchBuffer MB size of caching/prefetching wrapper for the index, if on Google Cloud (0 to disable).
+     */
+    public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices,
+            SamReaderFactory customSamReaderFactory,
+            int cloudPrefetchBuffer, int cloudIndexPrefetchBuffer) {
+        this(samPaths, samIndices, customSamReaderFactory,
+            (cloudPrefetchBuffer > 0 ? is -> ReadsDataSource.addPrefetcher(cloudPrefetchBuffer, is)
+                                     : Function.identity()),
+            (cloudIndexPrefetchBuffer > 0 ? is -> ReadsDataSource.addPrefetcher(cloudIndexPrefetchBuffer, is)
```
droazen: `ReadsDataSource.addPrefetcher` -> `addPrefetcher`, since we're now inside `ReadsDataSource`

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
jean-philippe-martin: Done, and applied boot to IDE's backside.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
jean-philippe-martin: OK.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -660,4 +672,12 @@ protected final Object doWork() {
...
     public void closeTool(){
     }
+
+    private int getCloudIndexPrefetchBuffer() {
+        if (cloudIndexPrefetchBuffer<0) {
+            // negative value means the default of "use cloudPrefetchBuffer"
+            return cloudPrefetchBuffer;
+        }
+        return cloudIndexPrefetchBuffer;
+    }
```
droazen: Now-unused method still needs deleting

jean-philippe-martin: Removed, sorry. That's what I get for drinking tea instead of coffee.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -164,11 +218,12 @@ public ReadsDataSource( final List<Path> samPaths, final List<Path> samIndices,
 
             SamReader reader;
             if ( samIndices == null ) {
-                reader = samReaderFactory.open(samPath);
+                reader = samReaderFactory.open(samPath, wrapper, indexWrapper);
```
droazen: So, when we hit this case (`samIndices == null`), the name of the index file will be inferred automatically using the name of the bam. In this case, we want to wrap the index, but I think we're passing in `Function.identity()` instead. This is due to the way `indexWrapper` is initialized:

```
            Function<SeekableByteChannel, SeekableByteChannel> indexWrapper =
                (samIndices != null && BucketUtils.isCloudStorageUrl(samIndices.get(samCount))
                    ? cloudIndexWrapper
                    : Function.identity());
```

I think when `samIndices == null` and `samPath` is a cloud storage url, we should set `indexWrapper` to `cloudIndexWrapper` so that the auto-discovered index file will be wrapped. Do you agree?

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -54,6 +54,11 @@ public static boolean isCloudStorageUrl(final String path) {
         return path.startsWith(GCS_PREFIX);
     }
 
+    public static boolean isCloudStorageUrl(final java.nio.file.Path path) {
+        // the initial "" protects us against a null scheme
+        return ("" + path.toUri().getScheme() + "://").equals(GCS_PREFIX);
```
droazen: I think this will still explode when `getScheme()` returns null, since the concatenation of a String with null causes a `NullPointerException` to be thrown. Need something like:

```
final String scheme = path.toUri().getScheme();
return scheme == null ? false : scheme + "://".equals(GCS_PREFIX);
```

droazen: Should probably also add a quick unit test that passes in a `Path` with a null scheme to this method.

src/test/java/org/broadinstitute/hellbender/engine/ReadsDataSourceUnitTest.java
```diff
@@ -587,6 +610,63 @@ public void testManuallySpecifiedIndicesWithCustomReaderFactory( final List<Path
...
+        try ( final ReadsDataSource readsSource = new ReadsDataSource(bams, indices, customFactory, xorData, xorIndex) ) {
+            Assert.assertTrue(readsSource.indicesAvailable(), "Explicitly-provided indices not detected for bams: " + bams);
+
+            final Iterator<GATKRead> queryReads = readsSource.query(new SimpleInterval("1", 1, 300));
+            int queryCount = 0;
+            while ( queryReads.hasNext() ) {
+                ++queryCount;
+                queryReads.next();
+            }
+            Assert.assertEquals(queryCount, 2, "Wrong number of reads returned in query");
```
droazen: Are there actually only 2 reads overlapping the interval 1:1-300 in the XOR'd version of this file? If it's a copy of `reads_data_source_test1.bam`, then I think the answer should be 5.

src/test/java/org/broadinstitute/hellbender/engine/ReadsDataSourceUnitTest.java
```diff
@@ -554,6 +563,20 @@ private File getFileWithAddedContig(
         };
     }
 
+    @DataProvider(name="cloudXorTestData")
+    public Object[][] cloudXorTestData() {
```
droazen: Move this `DataProvider` next to the `testCloudBamWithCustomReaderFactoryAndWrappers()` test case

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
jean-philippe-martin: That sounds reasonable, yes. I'll make the change.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
jean-philippe-martin: I respectfully disagree, the concatenation of a String with null returns the string "null" due to the language's [rules for string conversion](https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.1.11). I added a test for everyone's peace of mind and it confirms my understanding.

droazen: I stand corrected! Thanks for the clarification.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
droazen: Is there also a way we can add a test to check that when indices are not explicitly specified and the bam is on GCS, index wrappers get applied?

src/test/java/org/broadinstitute/hellbender/engine/ReadsDataSourceUnitTest.java
jean-philippe-martin: Yes, there are only 2 reads. I confirmed by running the same test without the wrapper.

It is indeed a copy of one of the other two (don't remember which), but we only get 5 reads when we read both `reads_data_source_test1.bam` and `reads_data_source_test2.bam` together, as happens when using _manuallySpecifiedIndexTestData_.

droazen: Ok, thanks for double-checking!

src/test/java/org/broadinstitute/hellbender/engine/ReadsDataSourceUnitTest.java
```diff
@@ -587,6 +610,63 @@ public void testManuallySpecifiedIndicesWithCustomReaderFactory( final List<Path
...
+            }
+            Assert.assertEquals(queryCount, 2, "Wrong number of reads returned in query");
+        }
+    }
+
+    // this code generates the xor'd inputs, in case we need to do it again.
+    private void createXorInput() throws IOException {
+        String prefix = "src/test/resources/org/broadinstitute/hellbender/engine/";
+        Path input = Paths.get(prefix + "reads_data_source_test2.bam");
+        Path output = Paths.get(prefix + "reads_data_source_test4.xor.bam");
```
droazen: Can you write the output to a temporary file (as returned by the inherited method `createTempFile()`), instead of writing to a file in `src/test/resources`?

src/test/resources/org/broadinstitute/hellbender/engine/reads_data_source_test4.xor.bam.bai
```diff
@@ -0,0 +1 @@
+QTPPPRPPPBPPQPPP�PPPPPPPSPPPPPP�PPRPPP�PPPPPPPSPPPPPPTPPPPPPPPPPPPPPPQPPP�PPPPPPPRPPPBPPQPPPSPPPPPP�UPPPPPP�PPRPPPSPPPPPP�UPPPPPPTPPPPPPPPPPPPPPPQPPPSPPPPPPRPPPBPPQPPP�UPPPPPPPP�QPPPP�PPRPPP�UPPPPPPPP�QPPPPSPPPPPPPPPPPPPPPQPPP�UPPPPPPPPPPPPPPPPPPPPPP
```
droazen: Write these to temp files deleted on exit instead of checking them in.

src/test/java/org/broadinstitute/hellbender/utils/gcs/BucketUtilsTest.java
```diff
@@ -16,21 +19,27 @@
 
     @Test
     public void testIsCloudStorageURL(){
-        Assert.assertTrue(BucketUtils.isCloudStorageUrl("gs://a_bucket/bucket"));
+        Assert.assertTrue(BucketUtils.isCloudStorageUrl("gs://abucket/bucket"));
```
droazen: Is `_` not allowed in GCS bucket names?

src/test/java/org/broadinstitute/hellbender/engine/ReadsDataSourceUnitTest.java
jean-philippe-martin: done.

src/test/java/org/broadinstitute/hellbender/utils/nio/XorWrapper.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.NonWritableChannelException;
+import java.nio.channels.SeekableByteChannel;
+import java.util.function.Function;
+
+/**
+ * A SeekableByteChannel wrapper for testing.
+ */
+public class XorWrapper implements SeekableByteChannel {
```
droazen: We generally put test-related utility classes in `src/main/java/org/broadinstitute/hellbender/utils/test/`, rather than under `src/test/java`, which we generally reserve for tests themselves.

droazen:
Third-pass review complete -- we should be able to merge once last few comments are addressed. Let me know when it's ready for a final look.

src/test/java/org/broadinstitute/hellbender/utils/gcs/BucketUtilsTest.java
jean-philippe-martin: That's an interesting story. Underscore is [legal](https://cloud.google.com/storage/docs/naming) in bucket names, but the same documentation strongly discourages using them because they are not legal in host names. The second element of URIs is normally a host name, so URI.create will fail for a URI like "gs://a_bucket/file". It [appears](http://www.zytrax.com/books/dns/apa/names.html) that underscores are now legal in host names, but they are still not accepted by the URI constructor.

src/test/resources/org/broadinstitute/hellbender/engine/reads_data_source_test4.xor.bam.bai
jean-philippe-martin: We'd have to write them to a bucket. This will create problems if more than one person/tool is running the test simultaneously.

src/test/java/org/broadinstitute/hellbender/engine/ReadsDataSourceUnitTest.java
jean-philippe-martin: This code is here for documentation purposes only. If it is a problem, should I remove it instead?

src/test/java/org/broadinstitute/hellbender/utils/nio/XorWrapper.java
jean-philippe-martin: I moved it, but as a result I fear it'll be included in the jar and loaded into every worker every time we run a job.

jean-philippe-martin:
OK, I _think_ I got everything (knock on wood).

droazen:
:+1: Latest version looks good -- merging!

jean-philippe-martin:
Great, thanks!

Issue #2287, opened 2016-11-30
Possible regression in merging bams on GCS in Spark


droazen:
@vdauwera reports that the following `ApplyBQSRSpark` command fails with an NIO-related error on dataproc:

```
time ./gatk-launch ApplyBQSRSpark \
    -I gs://hellbender/test/resources/benchmark/CEUTrio.HiSeq.WEx.b37.NA12892.bam \
    -R gs://gatk-legacy-bundles/b37/human_g1k_v37.2bit \
    -O gs://gatk-demo/TEST/gatk4-spark/recalibrated.bam \
    -bqsr gs://gatk-demo/TEST/gatk4-spark/recalibration.table \
    -apiKey $APIKEY_ORTMP \
    -- \
    --sparkRunner GCS \
    --cluster gvda-test-bqsr \
    --num-executors 40 \
    --executor-cores 4 \
    --executor-memory 20g
```

```
java.nio.file.FileSystemNotFoundException: Provider "gs" not installed
    at java.nio.file.Paths.get(Paths.java:147)
    at org.seqdoop.hadoop_bam.util.NIOFileUtil.asPath(NIOFileUtil.java:30)
    at org.seqdoop.hadoop_bam.util.SAMFileMerger.mergeParts(SAMFileMerger.java:51)
    at org.broadinstitute.hellbender.engine.spark.datasources.ReadsSparkSink.writeReadsSingle(ReadsSparkSink.java:231)
    at org.broadinstitute.hellbender.engine.spark.datasources.ReadsSparkSink.writeReads(ReadsSparkSink.java:153)
    at org.broadinstitute.hellbender.engine.spark.GATKSparkTool.writeReads(GATKSparkTool.java:247)
    at org.broadinstitute.hellbender.tools.spark.ApplyBQSRSpark.runTool(ApplyBQSRSpark.java:49)
    at org.broadinstitute.hellbender.engine.spark.GATKSparkTool.runPipeline(GATKSparkTool.java:348)
    at org.broadinstitute.hellbender.engine.spark.SparkCommandLineProgram.doWork(SparkCommandLineProgram.java:38)
    at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:109)
    at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:167)
[...]

PR #2333, opened 2017-01-10
revert nio sam file merger

lbergelson:
@tomwhite @droazen This fixes the problem I was seeing, but it's insanely slow for some reason.  A 30gb file on a heavy duty cluster was taking between 30-50 minutes depending on how I sharded it.  Most of that time is spent on the merging step where the master concatenates the chunks. 

I think the refactoring of the writing that I've done isn't a bad thing to have.  I added some extra log statements that it much easier to understand what's going on when it's slow.


droazen:
I think we should wait for the NIO-based fix rather than merging this, which would be useless for our performance work.

tomwhite:
@droazen I added a comment on https://github.com/broadinstitute/gatk/issues/2287 about what's needed from the GCS NIO side (see also https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1519).

lbergelson:
closing this then in favor of waiting for nio updates

PR #2335, opened 2017-01-11
deleting GATKException.CommandLineParserInternalException

lbergelson:
this seems to be outdated now that we have barclay


cmnbroad:
Look good. Thanks!

PR #2282, opened 2016-11-23
More consistent UserExceptions

magicDGS:
When I was trying to use user exceptions in a consistent way independently of the constructor (mostly related with files), I found very weird behaviour with the messages. Here I try to fix some of the things that I was struggling with:

* Support for path in constructors for `CouldNotReadInputFile`, `CouldNotCreateOutputFile`, `MalformedFile` and `MalformedBAM`, in addition to some missing constructors to have the same structure for all of them (with `File` and/or `String`).
* ~~Updated javadoc in `CommandLineException`, including extending classes to make clear that in the GATK framework is not printed out if it is thrown out of parameter validation.~~ __Edited__: this is not longer required, because `CommandLineException` is decoupled from `UserException` through barclay.
* Added a TODO into the `MalformedBAM` constructor that includes a `GATKRead` that is not used.
* __Edited__: added final to constructors.

PR #2336, opened 2017-01-11
improving error message in GATKSparkTool.writeReads

lbergelson:
changing from GATKException to the more appropriate UserException.CouldNotCreateOutputFile
updating user exceptions that take an exception to properly pass the exception on as their cause so they don't show incomplete stack traces


lbergelson:
@droazen I noticed that exception causes were getting lost while looking into the gcs issue, this should fix some of that.

magicDGS:
Maybe the part of stack trace should be included in #2282 too? Because I'm working on every `UserException`, I think that it will be important for all the new constructors...

lbergelson:
@magicDGS I think if you're adding new constructors that take an exception as a parameter they should probably propagate it to the super constructor.  I

droazen:
:+1: merge when ready

PR #2339, opened 2017-01-13
Handling samples and cleaning in ASEReadCounter and ReadPileup

meganshand:
@lbergelson Sorry, you were too fast for me! This now handles multiple samples in the ReadPileup filter and cleans up that StringBuilder.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
@@ -2,7 +2,7 @@ package org.broadinstitute.hellbender.tools.walkers.rnaseq;
 
 
 import htsjdk.variant.variantcontext.VariantContext;
-import htsjdk.samtools.util.Locatable;
+import htsjdk.samtools.SAMFileHeader;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.broadinstitute.hellbender.cmdline.programgroups.ReadProgramGroup;
 import org.broadinstitute.hellbender.engine.AlignmentContext;
@@ -214,14 +214,15 @@ public class ASEReadCounter extends LocusWalker {
     }
 
     private ReadPileup filterPileup(final ReadPileup originalPileup, final CountPileupType countType) {
+        SAMFileHeader header = getHeaderForReads();
 
         final ReadPileup pileupWithDeletions;
         switch (countType) {
-            case COUNT_FRAGMENTS_REQUIRE_SAME_BASE: pileupWithDeletions = originalPileup.getOverlappingFragmentFilteredPileup(true, ReadPileup.baseQualTieBreaker);
+            case COUNT_FRAGMENTS_REQUIRE_SAME_BASE: pileupWithDeletions = originalPileup.getOverlappingFragmentFilteredPileup(true, ReadPileup.baseQualTieBreaker, header);
                 break;
             case COUNT_READS: pileupWithDeletions = originalPileup;
                 break;
-            case COUNT_FRAGMENTS: pileupWithDeletions = originalPileup.getOverlappingFragmentFilteredPileup(false, ReadPileup.baseQualTieBreaker);
+            case COUNT_FRAGMENTS: pileupWithDeletions = originalPileup.getOverlappingFragmentFilteredPileup(false, ReadPileup.baseQualTieBreaker, header);
                 break;
             default: throw new UserException("Must use valid CountPileupType");
         }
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
```diff
@@ -269,30 +270,18 @@ else if (base.getBase() == altAllele) {
...
-        line.append(separator);
-        line.append(lowMAPQDepth);
-        line.append(separator);
-        line.append(lowBaseQDepth);
-        line.append(separator);
-        line.append(rawDepth);
-        line.append(separator);
-        line.append(otherBasesCount);
-        line.append(separator);
+        line.append(pileup.getLocation().getContig()).append(separator);
```
lbergelson: So...  I think I gave you bad advice. It's very likely that the compiler would optimize the `+` version of this into exactly the same as the `StringBuilder`.  It wouldn't necessarily do so in a more complex expression, but it's very likely to do so in this case.  

http://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -354,36 +354,39 @@ public ReadPileup getOverlappingFragmentFilteredPileup() {
...
+                } else {
+                    PileupElement existing = filteredPileup.get(readName);
+
+                    // if the reads disagree at this position, throw them all out.  Otherwise
+                    // keep the element with the highest quality score
+                    if (discardDiscordant && existing.getBase() != p.getBase()) {
+                        filteredPileup.remove(readName);
+                        readNamesDeleted.add(readName);
+                    } else {
+                        if (tieBreaker.compare(existing, p) < 0) {
```
lbergelson: I think this if statement should be moved up into the else as an `else if(..)`.  Less nesting is usually better if it's equivalent which this is.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -354,36 +354,39 @@ public ReadPileup getOverlappingFragmentFilteredPileup() {
...
-    public ReadPileup getOverlappingFragmentFilteredPileup(boolean discardDiscordant, Comparator<PileupElement> tieBreaker) {
-        Map<String, PileupElement> filteredPileup = new HashMap<String, PileupElement>();
-        Set<String> readNamesDeleted = new HashSet<>();
+    public ReadPileup getOverlappingFragmentFilteredPileup(boolean discardDiscordant, Comparator<PileupElement> tieBreaker, SAMFileHeader header) {
+        List<PileupElement> filteredPileupList = new ArrayList<PileupElement>();
 
-        for (PileupElement p : this) {
-            String readName = p.getRead().getName();
+        for( Map.Entry<String, ReadPileup> pileup : this.splitBySample(header, null).entrySet()) {
+            Map<String, PileupElement> filteredPileup = new HashMap<String, PileupElement>();
```
lbergelson: This code looks right.  It's a little complicated though, so to help improve readability I might extract everything in the outer loop into a private function that returns the collection of PileupElements.  

So it could read like:
```
 for (ReadPileup pileup : this.splitBySample(header, null).values()) {
            Collection<PileupElement> elements = filterSingleSampleForOverlaps(pileup, tieBreaker, discardDiscordant);
            filteredPileupList.addAll(elements);
}
```

There's a bunch of extra work being done here to support the multisample case, if it is slowing down the tool too much we could definitely optimize it.  But I've already violated the prime directive of software engineering and told you to do a silly optimization that probably does nothing, so... lets wait and see what shows up on the profile...

lbergelson:
@meganshand Looks good.  I have two trivial readability comments and one apology for my sins.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
meganshand: done

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
meganshand: Done. I'm not particularly worried about speed here. We probably won't use this tool except with a single sample, and if it's a problem for users I'd propose that we can deal with it then.

src/main/java/org/broadinstitute/hellbender/tools/walkers/rnaseq/ASEReadCounter.java
meganshand: I don't think the StringBuilder makes it that messy, so I'm going to leave it as is for now unless you have a strong opinion.

meganshand:
Back to you @lbergelson

Issue #2324, opened 2016-12-21
CommandLineException/CommandLineException.BadArgumentValue should be a UserException (or treated as one)


droazen:
Providing a bad value for an argument is the definition of a "user exception" (an error that is the user's fault). If we can't make it a `UserException` directly because of the gatk/barclay split, then we should at least wrap these exceptions in `UserException` at the top level catch, and ensure that a full error message gets printed for the user for these exceptions. 

Currently I think `CommandLineException` can be silently caught and ignored with no output due to a bug in `Main.mainEntry()`:

```
   protected final void mainEntry(final String[] args) {
        try {
            final Object result = instanceMain(args);
            handleResult(result);
            System.exit(0);
        } catch (final CommandLineException e){
            //the usage has already been printed so don't print it here.
            if(printStackTraceOnUserExceptions()) {
                e.printStackTrace();
            }
            System.exit(COMMANDLINE_EXCEPTION_EXIT_VALUE);
        } catch (final UserException e){
            CommandLineProgram.printDecoratedUserExceptionMessage(System.err, e);

            if(printStackTraceOnUserExceptions()) {
                e.printStackTrace();
            }
```

We should fix this as part of this ticket.

magicDGS:
The `CommandLineException` is not failing silently except it is thrown out of the command line parsing and/or in the tool `customCommandLineValidation()`. This handling part of the code is in `CommandLineProgram`:

```java
    /**
    *
    * @return true if command line is valid
    */
    protected boolean parseArgs(final String[] argv) {

        commandLineParser = new CommandLineArgumentParser(this, getPluginDescriptors());
        try{
            final boolean ret = commandLineParser.parseArguments(System.err, argv);
            commandLine = commandLineParser.getCommandLine();
            if (!ret) {
                return false;
            }
            final String[] customErrorMessages = customCommandLineValidation();
            if (customErrorMessages != null) {
                for (final String msg : customErrorMessages) {
                    System.err.println(msg);
                }
                commandLineParser.usage(System.err, false);
                return false;
            }
            return true;
        } catch (final CommandLineException e){
            //The CommandLineException is treated specially - we display help and no blow up
            commandLineParser.usage(System.err, true);
            printDecoratedUserExceptionMessage(System.err, e);
            //rethrow e - this will be caught upstream and the right exit code will be used.
            //we don't exit here though - only Main.main is allowed to call System.exit.
            throw e;
        }
    }
```

[...]

PR #2340, opened 2017-01-13
CommandLineException is no longer silent #2324

lbergelson:
fixes the issue where CommandLineExceptions produced no error message

added a public getUsage() method to CommandLineProgram
added a catch for these in instanceMain(), where the CommandLineProgram is in scope for printing the usage message
added a protected accessor getCommandLineParser to CommandLineProgram which guards against having an uninitialized CommandLineParser
moved the "A USER ERROR HAS OCCURRED" text out of the actual user exception and into the pretty printing

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -62,12 +62,12 @@ public class Main {
     /**
      * Prints the given message (may be null) to the provided stream, adding adornments and formatting.
      */
-    public static void printDecoratedUserExceptionMessage(final PrintStream ps, final Exception e){
+    protected static void printDecoratedExceptionMessage(final PrintStream ps, final Exception e, String prefix){
         Utils.nonNull(ps, "stream");
         Utils.nonNull(e, "exception");
         ps.println("***********************************************************************");
         ps.println();
-        ps.println("A USER ERROR has occurred: " + e.getMessage());
+        ps.println(prefix + e.getMessage());
         ps.println();
         ps.println("***********************************************************************") ;
     }
@@ -104,17 +104,20 @@ public class Main {
      */
     public Object instanceMain(final String[] args, final List<String> packageList, final List<Class<? extends CommandLineProgram>> classList, final String commandLineName) {
         final CommandLineProgram program = extractCommandLineProgram(args, packageList, classList, commandLineName);
-        if (null == program) return null; // no program found!
+        return runCommandLineProgram(program, args);
+    }
+
+    /**
+     * Run the given command line program with the raw arguments from the command line
+     * @param rawArgs thes are the raw arguments from the command line, the first will be stripped off
+     * @return the result of running  {program} with the given args, possibly null
+     */
+    protected static Object runCommandLineProgram(CommandLineProgram program, String[] rawArgs) {
+        if (null == program) return null; // no program found!  This will happen if help was specified with no other arguments
         // we can lop off the first two arguments but it requires an array copy or alternatively we could update CLP to remove them
         // in the constructor do the former in this implementation.
-        final String[] mainArgs = Arrays.copyOfRange(args, 1, args.length);
-        try {
-            return program.instanceMain(mainArgs);
-        } catch (final CommandLineException e){
-            //Print usage in the case of a CommandLineException
-            System.err.println(program.getUsage());
-            throw e;
-        }
+        final String[] mainArgs = Arrays.copyOfRange(rawArgs, 1, rawArgs.length);
+        return program.instanceMain(mainArgs);
     }
 
     /**
@@ -133,11 +136,15 @@ public class Main {
      * Note: this is the only method that is allowed to call System.exit (because gatk tools may be run from test harness etc)
      */
     protected final void mainEntry(final String[] args) {
+        final CommandLineProgram program = extractCommandLineProgram(args, getPackageList(), getClassList(), getCommandLineName());
         try {
-            final Object result = instanceMain(args);
+            final Object result = runCommandLineProgram(program, args);
             handleResult(result);
             System.exit(0);
         } catch (final CommandLineException e){
+            if (program != null) { //program can be null if th
+                System.err.println(program.getUsage());
+            }
             handleUserException(e);
             System.exit(COMMANDLINE_EXCEPTION_EXIT_VALUE);
         } catch (final UserException e){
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -117,24 +138,19 @@ protected final void mainEntry(final String[] args) {
             handleResult(result);
             System.exit(0);
         } catch (final CommandLineException e){
-            //the usage has already been printed so don't print it here.
-            if(printStackTraceOnUserExceptions()) {
-                e.printStackTrace();
-            }
+            handleUserException(e);
```
magicDGS: This will print also internal exceptions from Barclay, no?

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -93,7 +108,13 @@ public Object instanceMain(final String[] args, final List<String> packageList,
         // we can lop off the first two arguments but it requires an array copy or alternatively we could update CLP to remove them
         // in the constructor do the former in this implementation.
         final String[] mainArgs = Arrays.copyOfRange(args, 1, args.length);
-        return program.instanceMain(mainArgs);
+        try {
+            return program.instanceMain(mainArgs);
+        } catch (final CommandLineException e){
+            //Print usage in the case of a CommandLineException
+            System.err.println(program.getUsage());
```
magicDGS: Why not doing this in `mainEntry()`. I think that it is more configurable like that...

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -58,6 +60,19 @@
...
 
     /**
+     * Prints the given message (may be null) to the provided stream, adding adornments and formatting.
+     */
+    public static void printDecoratedUserExceptionMessage(final PrintStream ps, final Exception e){
+        Utils.nonNull(ps, "stream");
+        Utils.nonNull(e, "exception");
+        ps.println("***********************************************************************");
+        ps.println();
+        ps.println("A USER ERROR has occurred: " + e.getMessage());
```
magicDGS: Maybe the preamble ("A USER ERROR has ocurred") should be a parameter to allow other decorated exceptions being printed in a different way?

src/main/java/org/broadinstitute/hellbender/Main.java
lbergelson: Yes, that's currently the case, but it's going to be fixed soon I believe, with those exceptions being moved to a different exception type.

src/main/java/org/broadinstitute/hellbender/Main.java
lbergelson: I put it here because `mainEntry()` doesn't have access to the program itself, so it can't print the usage.  Will this cause problems for you?  I could pull up the call to `extractCommandLineProgram` into `mainEntry()` if having this here would cause you issues.

magicDGS: If it is complicated, I think that it could be maintained here. I'm just thinking in putative usages in the future for me or other framework users...

src/main/java/org/broadinstitute/hellbender/Main.java
cmnbroad: Right - Barclay will be changed soon so that CommandLineException can always be treated as a user error, and internal exception will be separate. Once thats done I think this will do exactly the right thing.

src/main/java/org/broadinstitute/hellbender/Main.java
cmnbroad: Does this need to be public ?

magicDGS: If someone extends main and want to handle some exceptions in a different way, decorating the exception may be useful as public or protected. And if the preamble is configurable, it may help to handle other exceptions differently for other toolkit.

lbergelson: made private and extracted the a prefix argument

magicDGS: I would rather prefer protected for usage in classes extending `Main`...

cmnbroad: Protected is what I had in mind when originally made my comment. It just seemed unnecessary for it to be public.

cmnbroad:
A couple of minor comments, otherwise LGTM.

src/main/java/org/broadinstitute/hellbender/Main.java
lbergelson: made protected

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -204,34 +200,25 @@ public Object instanceMain(final String[] argv) {
 
     /**
     *
-    * @return true if command line is valid
+    * @return true if program should be executed, false if an information only argument like {@link SpecialArgumentsCollection#HELP_FULLNAME} was specified
```
lbergelson: I've changed the behavior of this method.  It now always throws if validation, including custom validation was unsuccessful.  I believe this is much more consistent behavior with the rest of the codebase.

lbergelson:
@cmnbroad Could you take a quick look at this again when you get a chance?  I changed a few things in the untested methods to respond to @magicDGS's comments, but since they're so important and basically untested I think it would be good for someone to scan them.  I change the behavior of failing customCommandLineValidation to always throw, so it's consistent failing regular command line validation as well.   I also fixed some comment formatting.

cmnbroad:
Sure.

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -112,29 +136,28 @@ public Object instanceMain(final String[] args) {
...
-            final Object result = instanceMain(args);
+            final Object result = runCommandLineProgram(program, args);
             handleResult(result);
             System.exit(0);
         } catch (final CommandLineException e){
-            //the usage has already been printed so don't print it here.
-            if(printStackTraceOnUserExceptions()) {
-                e.printStackTrace();
+            if (program != null) { //program can be null if th
+                System.err.println(program.getUsage());
```
cmnbroad: The comment trails off a bit at the end... Having said that I don't see a code path that gets here with program==null, since runCommandLineProgram just returns if it doesn't have a program. I suppose that could change in the future though.

lbergelson: Hmn, that's a good point.  I was thinking  that since program could be null and we should guard against that case, but you're right that a null program should never throw a `CommandLineException`.   I'll drop it.

PR #2342, opened 2017-01-16
Change the name of the MappingQualityReadFilter command line argument

cmnbroad:
Opportunistic change from the ambiguous "mappingQuality" to "minimumMappingQuality".


droazen:
:+1: merge when ready

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
```diff
@@ -28,7 +28,7 @@
                 { LibraryReadFilter.class.getSimpleName(),
                         (Consumer<SetupTest>) this::setupLibraryTest, "--library", "Foo" },
                 { MappingQualityReadFilter.class.getSimpleName(),
-                        (Consumer<SetupTest>) this::setupMappingQualityTest, "--mappingQuality", "255" },
+                        (Consumer<SetupTest>) this::setupMappingQualityTest, "--minimumMappingQuality", "255" },
```
cmnbroad: We should add a test case here that uses a better example for mapping quality (not 255).

PR #2343, opened 2017-01-17
Minor fixes for CountingReadFilter

magicDGS:
Found in https://github.com/broadinstitute/gatk/pull/2195:

* `getSummaryLineForLevel` fix for indentLevel = 0 (including tests)
* `resetFilterCount` now is propagated to lhs/rhs


magicDGS:
@cmnbroad, I think that it is better to fix in a separate commit, instead of #2195. Can you review?

droazen:
:+1: looks good, merging

Issue #2270, opened 2016-11-17
Fix adapter boundary for positive strand


ronlevine:
Port https://github.com/broadinstitute/gsa-unstable/pull/1517

ronlevine:
Closing since #2346 is merged.

PR #2346, opened 2017-01-17
Fix adapter bounday for positive strand

ronlevine:
Implement #2270.


ronlevine:
@davidbenjamin Please review.

Issue #2349, opened 2017-01-18
ExampleAssemblyRegionWalkerSparkIntegrationTest fails intermittently


cmnbroad:
The Spark version of testExampleAssemblyRegionWalker fails reliably when I run all of the tests locally through Gradle. When I run the test by itself, either through Gradle or through IntelliJ, it always passes. 

Most of the time when it fails, the test finishes, but the output doesn't match the expected output. The test appears to assume a single output part file; however when it fails there are many part files (about 70), and you get the (second) stack below.

There may be a deeper problem when there are multiple partitions though, since sometimes the test fails mid-run with a ConcurrentModificationException:

org.apache.spark.SparkException: Job aborted due to stage failure: Task 16 in stage 1251.0 failed 1 times, most recent failure: Lost task 16.0 in stage 1251.0 (TID 2169, localhost): java.util.ConcurrentModificationException
	at java.util.ArrayDeque$DeqIterator.next(ArrayDeque.java:643)
	at org.broadinstitute.hellbender.engine.FeatureCache.getCachedFeaturesUpToStopPosition(FeatureCache.java:216)
	at org.broadinstitute.hellbender.engine.FeatureDataSource.queryAndPrefetch(FeatureDataSource.java:393)
	at org.broadinstitute.hellbender.engine.FeatureManager.getFeatures(FeatureManager.java:264)
	at org.broadinstitute.hellbender.engine.FeatureContext.getValues(FeatureContext.java:163)
	at org.broadinstitute.hellbender.engine.FeatureContext.getValues(FeatureContext.java:115)
	at org.broadinstitute.hellbender.tools.examples.ExampleAssemblyRegionWalkerSpark.lambda$assemblyFunction$213c9289$1(ExampleAssemblyRegionWalkerSpark.java:95)
	at org.apache.spark.api.java.JavaPairRDD$$anonfun$toScalaFunction$1.apply(JavaPairRDD.scala:1028)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$saveAsHadoopDataset$1$$anonfun$13$$anonfun$apply$7.apply$mcV$sp(PairRDDFunctions.scala:1204)
[...]

PR #2350, opened 2017-01-19
Optimization of intervals on Spark

tomwhite:
Optimization of intervals on Spark, so that specifying intervals is faster than not using intervals.
    
See https://github.com/HadoopGenomics/Hadoop-BAM/issues/91.
    
* With this change Hadoop-BAM no longer supports the
KEEP_PAIRED_READS_TOGETHER_PROPERTY property, so this is now
implemented in BwaSparkEngine (which is the only place that uses it.)
* Remove use of IntervalUtils.getAllIntervalsForReference in
ReadsSparkSource, since not specifying intervals now returns all
reads.
* Add coalesce(1) to ensure that example Spark walkers produce one
output file (some failed with this change).


src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
```diff
@@ -105,18 +113,19 @@ public ReadsSparkSource(final JavaSparkContext ctx, final ValidationStringency v
...
             }
             return null;
         }).filter(v1 -> v1 != null);
+        return putPairsInSamePartition(header, reads);
     }
 
     /**
      * Loads Reads using Hadoop-BAM. For local files, readFileName must have the fully-qualified path,
-     * i.e., file:///path/to/bam.bam. This excludes unmapped reads.
+     * i.e., file:///path/to/bam.bam.
```
droazen: Is it true that this method won't exclude unmapped reads?

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
```diff
@@ -135,9 +144,7 @@ public ReadsSparkSource(final JavaSparkContext ctx, final ValidationStringency v
      * @return RDD of (SAMRecord-backed) GATKReads from the file.
      */
     public JavaRDD<GATKRead> getParallelReads(final String readFileName, final String referencePath, int splitSize) {
```
droazen: Javadoc for this method still says " This excludes unmapped reads." -- is this still true?

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
```diff
@@ -135,9 +144,7 @@ public ReadsSparkSource(final JavaSparkContext ctx, final ValidationStringency v
...
-        final SAMFileHeader readsHeader = getHeader(readFileName, referencePath, null);
-        List<SimpleInterval> intervals = IntervalUtils.getAllIntervalsForReference(readsHeader.getSequenceDictionary());
-        return getParallelReads(readFileName, referencePath, intervals, splitSize);
+        return getParallelReads(readFileName, referencePath, null, splitSize);
```
droazen: Document effect of passing in null, both here and in the javadoc for the four-argument overload. (Eg., is it the same as passing in `IntervalUtils.getAllIntervalsForReference()` as previously, or is the behavior different?

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
```diff
@@ -208,6 +215,40 @@ public boolean accept(Path path) {
     }
 
     /**
+     * Ensure reads in a pair fall in the same partition (input split), if the reads are queryname-sorted,
+     * so they are processed together. No shuffle is needed.
+     */
+    JavaRDD<GATKRead> putPairsInSamePartition(final SAMFileHeader header, final JavaRDD<GATKRead> reads) {
```
lbergelson: It still seems like this logic would still be better in hadoop bam.  I don't know if there is any time when you would want to read a query name sorted bam and not keep the partitions together.

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
```diff
@@ -208,6 +215,40 @@ public boolean accept(Path path) {
...
+            PeekingIterator<GATKRead> current = Iterators.peekingIterator(it1);
+            // skip the first read in the _current_ partition if it is the second in a pair since it will be handled in the previous partition
+            if (current.hasNext() && current.peek() != null && current.peek().isSecondOfPair()) {
+                current.next();
+            }
+            // append the first read in the _next_ partition to the _current_ partition if it is the second in a pair
+            PeekingIterator<GATKRead> next = Iterators.peekingIterator(it2);
+            if (next.hasNext() && next.peek() != null && next.peek().isSecondOfPair()) {
+                return Iterators.concat(current, next);
+            }
```
droazen: It seems to me that we need to iterate until we find a read with a different name, since 1. the reads with the same name are not guaranteed to be sorted by "first of pair"/"second of pair" status, and 2. there can be supplementary alignments in addition to the primary alignments.

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
droazen: Can you explain again why we're moving this logic into GATK4? It really would be better to have it part of Hadoop-BAM, since it's unlikely a client would want to load queryname-sorted reads without this invariant.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleAssemblyRegionWalkerSpark.java
```diff
@@ -77,7 +77,7 @@ public ActivityProfileState isActive(AlignmentContext locusPileup, ReferenceCont
 
     @Override
     protected void processAssemblyRegions(JavaRDD<AssemblyRegionWalkerContext> rdd, JavaSparkContext ctx) {
-        rdd.map(assemblyFunction(knownVariants)).saveAsTextFile(outputFile);
+        rdd.map(assemblyFunction(knownVariants)).coalesce(1).saveAsTextFile(outputFile);
```
droazen: Add a comment explaining why this is necessary.

src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSourceUnitTest.java
```diff
@@ -239,11 +246,54 @@ public void testIntervals() throws IOException {
...
         for ( GATKRead read : bam2 ) {
             records.add(read);
         }
         return ctx.parallelize(records);
     }
+
+    @DataProvider(name="readPairsAndPartitions")
+    public Object[][] readPairsAndPartitions() {
+        return new Object[][] {
+                { 1, 1, new int[] {2} },
```
droazen: Add a one-line comment documenting what each value in the array represents.

src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSourceUnitTest.java
```diff
@@ -239,11 +246,54 @@ public void testIntervals() throws IOException {
...
+    private JavaRDD<GATKRead> createReads(JavaSparkContext ctx, SAMFileHeader header, int numPairs, int numPartitions) {
+        final int readSize = 151;
+        final int fragmentLen = 400;
+        final String templateName = "readpair";
+        int leftStart = 10000;
+        List<GATKRead> reads = new ArrayList<>();
+        for (int i = 0; i < numPairs;i++) {
+            leftStart += readSize * 2;
+            int rightStart = leftStart + fragmentLen - readSize;
+            reads.addAll(ArtificialReadUtils.createPair(header, templateName + i, readSize, leftStart, rightStart, true, false));
```
droazen: Would be good to add test cases with: 1. secondary alignments in addition to primary alignments  2. supplementary alignments in addition to primary alignments, and possibly 3. the case of more than two reads in a fragment (in which case both "first of pair" and "second of pair" can be set).

src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSourceUnitTest.java
```diff
@@ -239,11 +246,54 @@ public void testIntervals() throws IOException {
...
+        List<List<GATKRead>> partitions = pairedReads.mapPartitions((FlatMapFunction<Iterator<GATKRead>, List<GATKRead>>) it ->
+                Iterators.singletonIterator(Lists.newArrayList(it))).collect();
+        assertEquals(partitions.size(), numPartitions);
+        for (int i = 0; i < numPartitions; i++) {
+            assertEquals(partitions.get(i).size(), expectedReadsPerPartition[i]);
+        }
+        assertEquals(Arrays.stream(expectedReadsPerPartition).sum(), numPairs * 2);
+    }
+
+    private JavaRDD<GATKRead> createReads(JavaSparkContext ctx, SAMFileHeader header, int numPairs, int numPartitions) {
```
droazen: Should probably be named `createPairedReads()`

droazen:
First-pass review complete -- back to @tomwhite

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
magicDGS: Actually this is something that I would like to have for unmapped reads sorted by name....

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
tomwhite: Yes, previously it used `IntervalUtils. getAllIntervalsForReference()` which excluded unmapped reads, and which is inconsistent with the non-Spark version. Unmapped reads should be filtered out if they are not needed. Note that in `GATKSparkTool.getReads()` it says "If no intervals were specified, returns all the reads (both mapped and unmapped).".

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
tomwhite: No - fixed.

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
tomwhite: The trouble is that it is very difficult to include in Hadoop-BAM, since with the new optimized interval filtering in https://github.com/HadoopGenomics/Hadoop-BAM/pull/122 it's not possible to continue getting reads from the BAMFileReader when it's past the end of a chunk. The alternative - putting this RDD code into Hadoop-BAM - is a bigger change as Hadoop-BAM is an MR-level library so it doesn't have any Spark code in it. 

I would prefer to keep the code here so we can iterate on it (David has some suggested improvements). Perhaps we can add it to Hadoop-BAM in time.

tomwhite: See comment above to Louis.

@magicDGS do you mean put all of the unmapped reads in a single partition? That could be a future enhancement.

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
tomwhite: These are good suggestions. I'd like to address them in a separate issue if that's OK though, since the current logic replicates what we had in Hadoop-BAM.

src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSourceUnitTest.java
tomwhite: Done

src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSourceUnitTest.java
tomwhite: Done

src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSourceUnitTest.java
tomwhite: Opened https://github.com/broadinstitute/gatk/issues/2418

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java
magicDGS: Sorry, I didn't explained it well. I would like to have in the same partition unmapped reads with the same read name if they are sorted by queryname. What I meant is that could be a public static method here or somewhere as an API to use outside `ReadSparkSource`...

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleAssemblyRegionWalkerSpark.java
tomwhite: Good point. I've reverted this since it is tracked by https://github.com/broadinstitute/gatk/issues/2349, and is unrelated to this change.

tomwhite:
Thanks for all the comments! I've made fixes for the minor ones. I'd like to address the secondary and supplemental reads improvement separately (https://github.com/broadinstitute/gatk/issues/2418), so that it doesn't block progress on testing the intervals optimization. I hope this PR can go in once we have a Hadoop-BAM release (which I'll do soon).

(Regarding the "keep paired reads together" code - as I've explained in a comment, it's pragmatic to have this code in GATK.)

tomwhite:
I released a new version of Hadoop-BAM (7.8.0) and updated this PR to use it. All tests are passing now.

lbergelson:
yay!

PR #2352, opened 2017-01-23
Extract disableAllReadFilters to StandardArgumentDefinitions

magicDGS:
To allow implementation of other `ReadFilter` plugins with the same parameter names.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/StandardArgumentDefinitions.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/StandardArgumentDefinitions.java
@@ -20,6 +20,7 @@ public final class StandardArgumentDefinitions {
     public static final String ASSUME_SORTED_LONG_NAME = "assumeSorted";
     public static final String READ_FILTER_LONG_NAME = "readFilter";
     public static final String DISABLE_READ_FILTER_LONG_NAME = "disableReadFilter";
+    public static final String DISABLE_ALL_READ_FILTERS_NAME = "disableAllReadFilters";
     public static final String CREATE_OUTPUT_BAM_INDEX_LONG_NAME = "createOutputBamIndex";
     public static final String CREATE_OUTPUT_BAM_MD5_LONG_NAME = "createOutputBamMD5";
     public static final String CREATE_OUTPUT_VARIANT_INDEX_LONG_NAME = "createOutputVariantIndex";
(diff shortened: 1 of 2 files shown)


magicDGS:
This is a very simple PR. Maybe @droazen or @lbergelson could have a quick look? Thank you very much.

src/main/java/org/broadinstitute/hellbender/cmdline/StandardArgumentDefinitions.java
```diff
@@ -20,6 +20,7 @@ private StandardArgumentDefinitions(){}
     public static final String ASSUME_SORTED_LONG_NAME = "assumeSorted";
     public static final String READ_FILTER_LONG_NAME = "readFilter";
     public static final String DISABLE_READ_FILTER_LONG_NAME = "disableReadFilter";
+    public static final String DISABLE_ALL_READ_FILTER_NAME = "disableAllReadFilters";
```
cmnbroad: Minor typo - should be DISABLE_ALL_READ_FILTERS_NAME (plural).

magicDGS:
I have several use cases for this:
1. I didn't mention it in the first comment, but if a tool does not want to provide custom `ReadFilter`from the user, but allow to disable the ones applied by the tool, this will keep in sync the parameter name.
2.  I would like to have a `ReadFilter` plugin that is applied "after analysis", but the parameters here said that they are applied "before analysis". This will be misleading for my users. For this case I also opened #2353, to allow custom parameters based on the same read filter plugin descriptor implemented in GATK.

src/main/java/org/broadinstitute/hellbender/cmdline/StandardArgumentDefinitions.java
magicDGS: Done.

magicDGS:
I fixed the typo. Back to you @cmnbroad.

cmnbroad:
@magicDGS Looks good - can you squash and then I can merge? Thx.

magicDGS:
Sorry, I though that you are using the new method for merging a PR. Rebased and squashed now, so ready to go. Thanks @cmnbroad!

magicDGS:
I think that this does not make sense if the name is changed in #2365. Closing in favor of that one (I will reopen if the other is not accepted).

PR #2353, opened 2017-01-23
Extract an AbstractReadFilterPluginDescriptor

magicDGS:
Extract an `AbstracReadFilterPluginDescriptor` to allow different arguments to handle the CLP. This may be useful for:

- Include other package to look for `ReadFilter` (although this may be solved by #2297)
- Do not provide an argument to disable all default read filters. For example, useful if the defaults should be always applied.
- Different documentation for the arguments. For example, if the read filters are applied after processing instead of before analysis.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/AbstractReadFilterPluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/AbstractReadFilterPluginDescriptor.java
@@ -1,0 +1,330 @@
+package org.broadinstitute.hellbender.cmdline.GATKPlugin;
+
+import htsjdk.samtools.SAMFileHeader;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.broadinstitute.barclay.argparser.CommandLineException;
+import org.broadinstitute.barclay.argparser.CommandLinePluginDescriptor;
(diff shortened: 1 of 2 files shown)


magicDGS:
Because you implemented the original plugin descriptor, can you have a look @cmnbroad?

cmnbroad:
@magicDGS I'd strongly prefer not to introduce a read filter descriptor hierarchy if we can avoid it, as it will be tricky to get right, and add complexity. We definitely need to be able to extend the package list used by the descriptor to find plugins, but as you point out we'll be able to use the configuration mechanism for that. For before/after-analysis filters, I expect that we'll just add that directly to the existing plugin once we resolve https://github.com/broadinstitute/gatk/pull/2085 (which I hope to get to this week). I think the rest of the cases can be addressed by overriding makeReadFilter and providing custom behavior of filter merging. If this turns out to be something truly common, we could consider allowing the tool to inject an argument collection into the plugin.

magicDGS:
I would like to specify what passing a `ReadFilter` to some of my tools means, so maybe passing an `ArgumentCollection` will be simpler than this one, I agree. Although #2085 may solve the issue regarding the `ReadTransformer`/`ReadFilter` ordering, I would like to have in the plugin a way to specify different parameters (maybe some of then hidden before expose to users or advanced in the case of disabling).

I will open a new PR for that change, but I will really appreciate if I can get something like that in this and other plugins (if implemented).

PR #2355, opened 2017-01-25
Make GATKReadFilterPluginDescriptor extensible

magicDGS:
I would like to use the implementation of `GATKReadFilterPluginDescriptor` with some minor changes, but the current design of the class does not allow it. In this PR I include the following changes (divided in commits):

- Argument collection to apply to `GATKReadFilterPluginDescriptor` added on construction. This allow to modify the argument descriptions/behaviour depending on the usages.
-  Make fields protected, and a couple of final keywords to methods that are not expected to be modified.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
@@ -1,0 +1,52 @@
+package org.broadinstitute.hellbender.cmdline.GATKPlugin;
+
+import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+
+import java.util.*;
+
+/**
+ * {@link GATKReadFilterArgumentCollection} for optional read filters in the command line. It allows:
+ *
+ * - Provide a list of read filters to apply.
+ * - Disable some and/or all read filters.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class DefaultGATKReadFilterArgumentCollection extends GATKReadFilterArgumentCollection {
+    private static final long serialVersionUID = 1L;
+
+    @Argument(fullName = StandardArgumentDefinitions.READ_FILTER_LONG_NAME,
+            shortName = StandardArgumentDefinitions.READ_FILTER_SHORT_NAME,
+            doc="Read filters to be applied before analysis", optional=true, common = true)
+    public final List<String> userEnabledReadFilterNames = new ArrayList<>(); // preserve order
+
+    @Argument(fullName = StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME,
+            shortName = StandardArgumentDefinitions.DISABLE_READ_FILTER_SHORT_NAME,
+            doc="Read filters to be disabled before analysis", optional=true, common = true)
+    public final List<String> userDisabledReadFilterNames = new ArrayList<>();
+
+    @Argument(fullName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
+            shortName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
+            doc = "Disable all tool default read filters", common = true, optional = true)
+    public boolean disableToolDefaultReadFilters = false;
+
+    /** Returns a list with the read filters provided by the user, preserving the order. */
+    @Override
+    public Collection<String> getUserEnabledReadFilterNames() {
+        return userEnabledReadFilterNames;
+    }
+
+    /** Returns a set of filters disabled by the user. */
+    @Override
+    public Collection<String> getUserDisabledReadFilterNames() {
+        return userDisabledReadFilterNames;
+    }
+
+    /** {@inheritDoc}. */
+    @Override
+    public boolean disableToolDefaultReadFilters() {
+        return disableToolDefaultReadFilters;
+    }
+
+}
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
@@ -6,7 +6,8 @@ import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 import java.util.*;
 
 /**
- * {@link GATKReadFilterArgumentCollection} for optional read filters in the command line. It allows:
+ * Default {@link GATKReadFilterArgumentCollection} applied in GATK for optional read filters in the command line.
+ * It contains arguments that allow the user to:
  *
  * - Provide a list of read filters to apply.
  * - Disable some and/or all read filters.
@@ -31,21 +32,21 @@ public class DefaultGATKReadFilterArgumentCollection extends GATKReadFilterArgum
             doc = "Disable all tool default read filters", common = true, optional = true)
     public boolean disableToolDefaultReadFilters = false;
 
-    /** Returns a list with the read filters provided by the user, preserving the order. */
+    /** Returns the list with the read filters provided by the user, preserving the order. */
     @Override
-    public Collection<String> getUserEnabledReadFilterNames() {
+    public List<String> getUserEnabledReadFilterNames() {
         return userEnabledReadFilterNames;
     }
 
-    /** Returns a set of filters disabled by the user. */
+    /** Returns the set of filters disabled by the user. */
     @Override
-    public Collection<String> getUserDisabledReadFilterNames() {
+    public List<String> getUserDisabledReadFilterNames() {
         return userDisabledReadFilterNames;
     }
 
     /** {@inheritDoc}. */
     @Override
-    public boolean disableToolDefaultReadFilters() {
+    public boolean getDisableToolDefaultReadFilters() {
         return disableToolDefaultReadFilters;
     }
 
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
@@ -1,8 +1,7 @@
 package org.broadinstitute.hellbender.cmdline.GATKPlugin;
 
 import java.io.Serializable;
-import java.util.Collection;
-import java.util.Collections;
+import java.util.List;
 
 /**
  * An abstract ArgumentCollection for defining the set of read filter descriptor plugin arguments that are exposed to the user on the command line.
@@ -17,25 +16,17 @@ public abstract class GATKReadFilterArgumentCollection implements Serializable {
 
     /**
      * Returns the enabled filter names. Order should be honored.
-     *
-     * Default implementation returns an empty list.
      */
-    public Collection<String> getUserEnabledReadFilterNames() {
-        return Collections.emptyList();
-    }
+    public abstract List<String> getUserEnabledReadFilterNames();
 
     /**
      * Returns the disabled read filter names. Order should be honored.
-     *
-     * Default implementation returns an empty list.
      */
-    public Collection<String> getUserDisabledReadFilterNames() {
-        return Collections.emptyList();
-    }
+    public abstract List<String> getUserDisabledReadFilterNames();
 
     /**
      * Returns {@code true} if all default filters are disabled; {@code false} otherwise.
      */
-    public abstract boolean disableToolDefaultReadFilters();
+    public abstract boolean getDisableToolDefaultReadFilters();
 
 }
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
@@ -1,0 +1,41 @@
+package org.broadinstitute.hellbender.cmdline.GATKPlugin;
+
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * An abstract ArgumentCollection for defining the set of read filter descriptor plugin arguments that are exposed to the user on the command line.
+ *
+ * Subclasses should provide {@link org.broadinstitute.barclay.argparser.Argument} annotations for the arguments that should be exposed.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ * @see org.broadinstitute.hellbender.cmdline.GATKPlugin.GATKReadFilterPluginDescriptor
+ */
+public abstract class GATKReadFilterArgumentCollection implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Returns the enabled filter names. Order should be honored.
+     *
+     * Default implementation returns an empty list.
+     */
+    public Collection<String> getUserEnabledReadFilterNames() {
+        return Collections.emptyList();
+    }
+
+    /**
+     * Returns the disabled read filter names. Order should be honored.
+     *
+     * Default implementation returns an empty list.
+     */
+    public Collection<String> getUserDisabledReadFilterNames() {
+        return Collections.emptyList();
+    }
+
+    /**
+     * Returns {@code true} if all default filters are disabled; {@code false} otherwise.
+     */
+    public abstract boolean disableToolDefaultReadFilters();
+
+}
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
@@ -1,9 +1,10 @@
 package org.broadinstitute.hellbender.cmdline.GATKPlugin;
 
+import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.SAMFileHeader;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.barclay.argparser.ArgumentCollection;
 import org.broadinstitute.barclay.argparser.CommandLineException;
 import org.broadinstitute.barclay.argparser.CommandLinePluginDescriptor;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
@@ -33,20 +34,10 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
     private static final String pluginPackageName = "org.broadinstitute.hellbender.engine.filters";
     private static final Class<?> pluginBaseClass = org.broadinstitute.hellbender.engine.filters.ReadFilter.class;
 
-    @Argument(fullName = StandardArgumentDefinitions.READ_FILTER_LONG_NAME,
-            shortName = StandardArgumentDefinitions.READ_FILTER_SHORT_NAME,
-            doc="Read filters to be applied before analysis", optional=true, common = true)
-    public final List<String> userEnabledReadFilterNames = new ArrayList<>(); // preserve order
-
-    @Argument(fullName = StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME,
-            shortName = StandardArgumentDefinitions.DISABLE_READ_FILTER_SHORT_NAME,
-            doc="Read filters to be disabled before analysis", optional=true, common = true)
-    public final List<String> userDisabledReadFilterNames = new ArrayList<>();
-
-    @Argument(fullName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
-            shortName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
-            doc = "Disable all tool default read filters", common = true, optional = true)
-    public boolean disableToolDefaultReadFilters = false;
+    // the purpose of this argument collection is to allow the caller to control the exposure of the command line arguments
+    @VisibleForTesting
+    @ArgumentCollection
+    public GATKReadFilterArgumentCollection userArgs;
 
     // Map of read filter (simple) class names to the corresponding discovered plugin instance
     private final Map<String, ReadFilter> allDiscoveredReadFilters = new HashMap<>();
@@ -61,10 +52,12 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
     private final Set<String> requiredPredecessors = new HashSet<>();
 
     /**
+     * @param userArgs           Argument collection to control the exposure of the command line arguments.
      * @param toolDefaultFilters Default filters that may be supplied with arguments
      *                           on the command line. May be null.
      */
-    public GATKReadFilterPluginDescriptor(final List<ReadFilter> toolDefaultFilters) {
+    public GATKReadFilterPluginDescriptor(final GATKReadFilterArgumentCollection userArgs, final List<ReadFilter> toolDefaultFilters) {
+        this.userArgs = userArgs;
         if (null != toolDefaultFilters) {
             toolDefaultFilters.forEach(f -> {
                 final Class<? extends ReadFilter> rfClass = f.getClass();
@@ -79,6 +72,14 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
                 toolDefaultReadFilters.put(className, f);
             });
         }
+    }
+
+    /**
+     * @param toolDefaultFilters Default filters that may be supplied with arguments
+     *                           on the command line. May be null.
+     */
+    public GATKReadFilterPluginDescriptor(final List<ReadFilter> toolDefaultFilters) {
+        this(new DefaultGATKReadFilterArgumentCollection(), toolDefaultFilters);
     }
 
     /////////////////////////////////////////////////////////
@@ -155,7 +156,7 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
         // for the user to subsequently disable the required predecessor. That case is caught during final
         // validation done by the validateArguments method.
         String predecessorName = predecessorClass.getSimpleName();
-        boolean isAllowed = userEnabledReadFilterNames.contains(predecessorName)
+        boolean isAllowed = userArgs.getUserEnabledReadFilterNames().contains(predecessorName)
                 || (toolDefaultReadFilters.get(predecessorName) != null);
         if (isAllowed) {
             // Keep track of the ones we allow so we can validate later that they weren't subsequently disabled
@@ -175,8 +176,8 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
      */
     @Override
     public List<ReadFilter> getAllInstances() {
-        final ArrayList<ReadFilter> filters = new ArrayList<>(userEnabledReadFilterNames.size());
-        userEnabledReadFilterNames.forEach(s -> {
+        final ArrayList<ReadFilter> filters = new ArrayList<>(userArgs.getUserEnabledReadFilterNames().size());
+        userArgs.getUserEnabledReadFilterNames().forEach(s -> {
             ReadFilter rf = allDiscoveredReadFilters.get(s);
             filters.add(rf);
         });
@@ -227,7 +228,7 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
     @Override
     public void validateArguments() {
         // throw if a filter is *enabled* more than once by the user
-        final Set<String> duplicateUserEnabledFilterNames = Utils.getDuplicatedItems(userEnabledReadFilterNames);
+        final Set<String> duplicateUserEnabledFilterNames = Utils.getDuplicatedItems(userArgs.getUserEnabledReadFilterNames());
         if (!duplicateUserEnabledFilterNames.isEmpty()) {
             throw new CommandLineException.BadArgumentValue(
                     String.format("The read filter(s) are enabled more than once: %s",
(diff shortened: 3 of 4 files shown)


magicDGS:
Hi @cmnbroad, could you please have a look to this? It is the substitution for #2353. Thank you very much for this!

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/GATKReadFilterPluginArgumentCollection.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * An abstract ArgumentCollection for specifying parameters for {@link org.broadinstitute.hellbender.cmdline.GATKPlugin.GATKReadFilterPluginDescriptor}.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public abstract class GATKReadFilterPluginArgumentCollection implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    /** Returns the enabled filter names. */
+    public abstract Collection<String> getEnabledReadFilters();
```
cmnbroad: This should to by typed as a List to maintain ordering (Collection doesn't necessarily preserve order), and add a comment saying that order should be honored.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/OptionalGATKReadFilterPluginArgumentCollection.java
```diff
@@ -0,0 +1,49 @@
...
+            doc="Read filters to be disabled before analysis", optional=true)
+    public final Set<String> disableFilters = new HashSet<>();
+
+    @Argument(fullName = "disableAllReadFilters",
+            shortName = "disableAllReadFilters",
+            doc = "Disable all read filters", common = false, optional = true)
+    public boolean disableAllReadFilters = false;
+
+    /** Returns a list with the read filters provided by the user, preserving the order. */
+    public Collection<String> getEnabledReadFilters() {
```
cmnbroad: Also should be List.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/OptionalGATKReadFilterPluginArgumentCollection.java
```diff
@@ -0,0 +1,49 @@
...
+ * - Disable some and/or all read filters.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class OptionalGATKReadFilterPluginArgumentCollection extends GATKReadFilterPluginArgumentCollection {
+    private static final long serialVersionUID = 1L;
+
+    @Argument(fullName = StandardArgumentDefinitions.READ_FILTER_LONG_NAME,
+            shortName = StandardArgumentDefinitions.READ_FILTER_SHORT_NAME,
+            doc="Read filters to be applied before analysis", optional=true)
```
cmnbroad: Although I think the common attribute will go away in the near future, for now we should make all of these args consistent and have common=true.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -33,38 +35,28 @@ private void readObject(java.io.ObjectInputStream in)
...
-            shortName = StandardArgumentDefinitions.DISABLE_READ_FILTER_SHORT_NAME,
-            doc="Read filters to be disabled before analysis", optional=true)
-    public final Set<String> disableFilters = new HashSet<>();
-
-    @Argument(fullName = "disableAllReadFilters",
-            shortName = "disableAllReadFilters",
-            doc = "Disable all read filters", common = false, optional = true)
-    public boolean disableAllReadFilters = false;
+    @ArgumentCollection
+    protected final GATKReadFilterPluginArgumentCollection args;
```
cmnbroad: I think this should be private (and perhaps the whole class should be final). Also, use a more descriptive name for this, maybe "userArgs", and add a comment saying what the purpose of this is (to allow the caller to control the exposure of the command line arguments).

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -33,38 +35,28 @@ private void readObject(java.io.ObjectInputStream in)
...
 
     // Map of read filter (simple) class names to the corresponding discovered plugin instance
-    private Map<String, ReadFilter> readFilters = new HashMap<>();
+    protected Map<String, ReadFilter> readFilters = new HashMap<>();
```
cmnbroad: Not sure why the access level of these changed - can they not be private ?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -33,38 +35,28 @@ private void readObject(java.io.ObjectInputStream in)
...
     // Map of read filter (simple) class names to the corresponding default plugin instance
-    private Map<String, ReadFilter> toolDefaultReadFilters = new HashMap<>();
+    protected final Map<String, ReadFilter> toolDefaultReadFilters = new HashMap<>();
 
     // Set of dependent args for which we've seen values (requires predecessor)
-    private Set<String> requiredPredecessors = new HashSet<>();
+    protected final Set<String> requiredPredecessors = new HashSet<>();
 
     /**
+     * @param args               Argument collection for the pluging.
```
cmnbroad: Also here we should add more specific name/comment stating the purpose.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -33,38 +35,28 @@ private void readObject(java.io.ObjectInputStream in)
...
      * @param toolDefaultFilters Default filters that may be supplied with arguments
      *                           on the command line. May be null.
      */
```
cmnbroad: I know you didn't write this comment (I think I did), but can you change it so say "Filters that are enabled by default for this tool. May be null."

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/OptionalGATKReadFilterPluginArgumentCollection.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * {@link GATKReadFilterPluginArgumentCollection} for optional read filters in the command line. It allows:
+ *
+ * - Provide a list of read filters to apply.
+ * - Disable some and/or all read filters.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class OptionalGATKReadFilterPluginArgumentCollection extends GATKReadFilterPluginArgumentCollection {
```
cmnbroad: I think a better name for this would be DefaultGATKReadFilterArugmentCollection.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/GATKReadFilterPluginArgumentCollection.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.io.Serializable;
+import java.util.Collection;
+
+/**
+ * An abstract ArgumentCollection for specifying parameters for {@link org.broadinstitute.hellbender.cmdline.GATKPlugin.GATKReadFilterPluginDescriptor}.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public abstract class GATKReadFilterPluginArgumentCollection implements Serializable {
```
cmnbroad: Maybe just call this GATKReadFilterArgumentCollection.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/GATKReadFilterPluginArgumentCollection.java
```diff
@@ -0,0 +1,23 @@
+package org.broadinstitute.hellbender.cmdline.argumentcollections;
+
+import java.io.Serializable;
+import java.util.Collection;
+
+/**
+ * An abstract ArgumentCollection for specifying parameters for {@link org.broadinstitute.hellbender.cmdline.GATKPlugin.GATKReadFilterPluginDescriptor}.
```
cmnbroad: I think we need more javadoc here. At a  minimum something like, "An abstract ArgumentCollection for defining the set of read filter descriptor plugin arguments that are exposed to the user on the command line. Subclasses should provide @ Argument annotations for the arguments that should be exposed."

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/GATKReadFilterPluginArgumentCollection.java
cmnbroad: Also, I think all of these could have default implementations.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/GATKReadFilterPluginArgumentCollection.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/GATKReadFilterPluginArgumentCollection.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/OptionalGATKReadFilterPluginArgumentCollection.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/GATKReadFilterPluginArgumentCollection.java
magicDGS: I don't agree with the typing as a `List`, because some implementations may have a `Set` argument to do not repeat filters.

I agree that they should maintain the ordering, but maybe some clients do not require that in their descriptor because they are not using a composed filter or they are independent of the order.

I added a default implementation returning an empty list.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/OptionalGATKReadFilterPluginArgumentCollection.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/OptionalGATKReadFilterPluginArgumentCollection.java
magicDGS: The same as before, I think that this may depend on the implementation.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: I changed the access level in the second commit of some parts of the code to make the descriptor extensible. There are some design issues for my personal tools and usage that I would like to solve in my project without affecting GATK, like:
* I would like to have a more verbose help in the plugin, and the returned sets of `getAllowedValuesForDescriptorArgument` should output in my case 1) the values of default tool filters for disable then and 2) all the possible filter names (found by reflection) for the user to know the possible values.
* The treatment of `disableAllReadFilters` to disable just the defaults but not the provided by the user.
* I need the returned filters of `getAllInstances` to return also the tool defaults.

This was the main point of my work in this plugin, because I don't want to copy everything and modify a couple of methods, but just extend the class to provide my own method. In addition, this will be for me also a temporary solution for the issue regarding the plugin package, where I should provide my own one.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: The same as in the access level changes, I would like to extend this class to don't "reinvent the wheel".

I changed the name and added the comment.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: I realized that at least the argument collection should be package-protected for testing.

magicDGS:
I addressed partially your comments (and fixed a compilation error due to the tests using the previous arguments). One of the major points of discussion are the following:

* `Collection` instead of `List`: I think that the first is more flexible, because a client maybe wants to have a `LinkedHashSet` as the argument to avoid repetition of the same filter. I agree that the abstract class should discourage not honoring the user order.
* Access to methods/fields: I think that the plugin could be used outside GATK in a different way by extending it. I explained some of my usage cases in one of the comments in the code, but just by overriding a simple method the whole plugin could be used very nicely in some of them. I would prefer to do that than copy your code and re-implement the bits that I would like to change.

Back to you for your ideas on this, @cmnbroad!

magicDGS:
I'm adding some issues and PRs for make the plugin usable in other cases too, @cmnbroad. Maybe you prefer that solution instead of make it extensible. Just let me know.

magicDGS:
Probably this should wait til #2365.

magicDGS:
Because you already did a lot of reviewing here @cmnbroad, should I close as in #2359 for make a single PR with all the workaround?

cmnbroad:
@magiDGS My preference would be to do a single PR with all of the fixes for the enable/disable validation rules and allowed values changes, and only those changes. Then we can do a second one with the extensibility changes. You can decide if you want to close this PR, or use it as the basis for the second one - either way is fine with me. Thanks!

magicDGS:
I prefer to keep this one open, it will be simpler to rebase and include the changes. Should the `common = true` changes being included in the other or here?

cmnbroad:
I'd say here.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -33,20 +34,10 @@ private void readObject(java.io.ObjectInputStream in)
...
-    public final List<String> userDisabledReadFilterNames = new ArrayList<>();
-
-    @Argument(fullName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
-            shortName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
-            doc = "Disable all tool default read filters", common = true, optional = true)
-    public boolean disableToolDefaultReadFilters = false;
+    // the purpose of this argument collection is to allow the caller to control the exposure of the command line arguments
+    @VisibleForTesting
+    @ArgumentCollection
+    public GATKReadFilterArgumentCollection userArgs;
```
magicDGS: This is visible for testing but public because the test is in a different package, which is quite annoying. What do you think about moving the tests to the same package?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * An abstract ArgumentCollection for defining the set of read filter descriptor plugin arguments that are exposed to the user on the command line.
+ *
+ * Subclasses should provide {@link org.broadinstitute.barclay.argparser.Argument} annotations for the arguments that should be exposed.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ * @see org.broadinstitute.hellbender.cmdline.GATKPlugin.GATKReadFilterPluginDescriptor
+ */
+public abstract class GATKReadFilterArgumentCollection implements Serializable {
```
magicDGS: Maybe this class could keep the default read filters instead of the plugin itself, what do you think?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -318,8 +319,8 @@ public void validateArguments() {
      * the user and all tool defaults are disabled; {@code false} otherwise.
      */
     public boolean isDisabledFilter(final String filterName) {
-        return userDisabledReadFilterNames.contains(filterName)
-                || (disableToolDefaultReadFilters && !userEnabledReadFilterNames.contains(filterName));
+        return userArgs.getUserDisabledReadFilterNames().contains(filterName)
+                || (userArgs.disableToolDefaultReadFilters() && !userArgs.getUserEnabledReadFilterNames().contains(filterName));
```
magicDGS: Maybe this logic will be better as an abstract method in the argument collection, called here.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -227,32 +228,32 @@ public boolean isDependentArgumentAllowed(final Class<?> predecessorClass) {
     @Override
     public void validateArguments() {
```
magicDGS: Maybe this could be encapsulated in the argument collection as the default implementation, allowing user-specific arguments to contain extra-validation if required...

cmnbroad:
@magicDGS I haven't forgotten about this - I will get to it this week.

magicDGS:
Thanks @cmnbroad!

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
```diff
@@ -0,0 +1,52 @@
...
+            doc = "Disable all tool default read filters", common = true, optional = true)
+    public boolean disableToolDefaultReadFilters = false;
+
+    /** Returns a list with the read filters provided by the user, preserving the order. */
+    @Override
+    public Collection<String> getUserEnabledReadFilterNames() {
+        return userEnabledReadFilterNames;
+    }
+
+    /** Returns a set of filters disabled by the user. */
```
cmnbroad: "Returns a" -> "Returns the"

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
```diff
@@ -0,0 +1,52 @@
...
+    public final List<String> userDisabledReadFilterNames = new ArrayList<>();
+
+    @Argument(fullName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
+            shortName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
+            doc = "Disable all tool default read filters", common = true, optional = true)
+    public boolean disableToolDefaultReadFilters = false;
+
+    /** Returns a list with the read filters provided by the user, preserving the order. */
+    @Override
+    public Collection<String> getUserEnabledReadFilterNames() {
```
cmnbroad: It seems like this should return List to reinforce that its ordered.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
```diff
@@ -0,0 +1,52 @@
...
+        return userEnabledReadFilterNames;
+    }
+
+    /** Returns a set of filters disabled by the user. */
+    @Override
+    public Collection<String> getUserDisabledReadFilterNames() {
```
cmnbroad: List here too.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
```diff
@@ -0,0 +1,52 @@
...
+        return userDisabledReadFilterNames;
+    }
+
+    /** {@inheritDoc}. */
+    @Override
+    public boolean disableToolDefaultReadFilters() {
```
cmnbroad: This is a getter so it would make more sense to call it getDisableToolDefaultReadFilters; also it will then be consistent with the other getters.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
```diff
@@ -0,0 +1,41 @@
...
+     * Default implementation returns an empty list.
+     */
+    public Collection<String> getUserDisabledReadFilterNames() {
+        return Collections.emptyList();
+    }
+
+    /**
+     * Returns {@code true} if all default filters are disabled; {@code false} otherwise.
+     */
+    public abstract boolean disableToolDefaultReadFilters();
```
cmnbroad: Is there a reason why you provided default implementations for some but not all of the methods? Actually, I think I would err on the side of making them all abstract, or at least all consistent.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Sounds reasonable.

cmnbroad: Also please make this variable final.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: My preference would be to leave it here and keep the argument collection as just an argument container.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public abstract class GATKReadFilterArgumentCollection implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Returns the enabled filter names. Order should be honored.
+     *
+     * Default implementation returns an empty list.
+     */
+    public Collection<String> getUserEnabledReadFilterNames() {
```
cmnbroad: List ? See my comment above.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
```diff
@@ -0,0 +1,41 @@
...
+        return Collections.emptyList();
+    }
+
+    /**
+     * Returns the disabled read filter names. Order should be honored.
+     *
+     * Default implementation returns an empty list.
+     */
+    public Collection<String> getUserDisabledReadFilterNames() {
```
cmnbroad: List ? See my comment above.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: I would strongly prefer to keep it here - I think we want to discourage the proliferation of implementations that have a different set of rules.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: Ok.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: Ok, but what's about a method in the argument collection for extra-validation of arguments, and call it at the end of this method?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
magicDGS: Ok.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
magicDGS: This was before I did the changes in the validation, to allow custom implementations using a set. I changed it, but it will give more flexibility to API users with a plain collection...

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
magicDGS: I made them all abstract. My first thought was to make defaults the one with collections, and enforce the one for disabling because it is the only one affecting filters that are in the plugin and not in the actual collection. Anyway, it makes more sense to have all of them abstract....

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
```diff
@@ -0,0 +1,52 @@
+package org.broadinstitute.hellbender.cmdline.GATKPlugin;
+
+import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+
+import java.util.*;
+
+/**
+ * {@link GATKReadFilterArgumentCollection} for optional read filters in the command line. It allows:
```
magicDGS: Rephrased to: `Default {@link GATKReadFilterArgumentCollection} applied in GATK for optional read filters in the command line. It contains arguments that allow the user to:`

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterArgumentCollection.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/DefaultGATKReadFilterArgumentCollection.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: I made it final, but for changing the tests to a different package I open a different issue.

magicDGS:
Adressed all comments, @cmnbroad. Still the question on how to allow the user to provide custom validation. Back to you and thanks for reviewing!

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: @magicDGS Can you remind me if there is a use-case you have in mind for that ? I seem to remember thinking you had cases where you wanted to *reduce* visibility of builtin arguments. Is there a case where you need to add additional arguments that need validation ? Just want to understand the context.

magicDGS: For example, if a plugin requires one of two `ReadFilter` to be applied, this method will be useful to check if both of them are disabled (or in case one or the other should be applied, check if they are mutually exclusive).

Another example is a constraint in some `ReadFilter` parameters, that are not included in the original filter. For instance, if `ReadLenghtReadFilter` should have always a minimum length of 0.

I think that this will be easier at some point with Barclay anyway if https://github.com/broadinstitute/barclay/issues/31 is implemented (the later comment about the `Validatable` interface).

cmnbroad: Right - we need a more general solution like broadinstitute/barclay#31, so lets not invent something specific to read filter plugin.

magicDGS: Ok.

magicDGS:
So is it ready to merge, @cmnbroad? Thank you in advance!

magicDGS:
Thanks @cmnbroad!

Issue #2357, opened 2017-01-26
Disable default read filter with arguments blows up as a CommandLineException


magicDGS:
The following test class may fail with the error: "Values were supplied for (ReadLengthReadFilter) that is also disabled":

```java
/**
 * @author Daniel Gomez-Sanchez (magicDGS)
 */
public class GATKReadFilterPluginDescriptorUnitTest extends BaseTest {

    @CommandLineProgramProperties(summary = "Test read filter plugin with default arguments",
            oneLineSummary = "Test read filter plugin with default arguments",
            programGroup = TestProgramGroup.class)
    private static class TestWithDefaultReadFilters extends CommandLineProgram {

        private final List<ReadFilter> defaultFilters;

        public TestWithDefaultReadFilters(final List<ReadFilter> defaultFilters) {
            this.defaultFilters = defaultFilters;
        }

        protected List<? extends CommandLinePluginDescriptor<?>> getPluginDescriptors() {
            return Collections.singletonList(new GATKReadFilterPluginDescriptor(defaultFilters));
        }

        @Override
        protected Object doWork() {
            return null;
        }
    }


    @Test
    public void testWithDefaultReadFiltersWithParams() throws Exception {
        // this ReadFilter have parameters --maxReadLength/--minReadLength, that are set because of the default filter
        final CommandLineProgram clp = new TestWithDefaultReadFilters(Collections.singletonList(new ReadLengthReadFilter(10, 50)));
        // disable the read filter should not blow up because of that parameters, because they are not provided by the user
        clp.instanceMain(new String[]{"--" + StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME, "ReadLengthReadFilter"});

    }
}
```

I don't know if this may be solved in GATK or in Barclay, but at least a workaround for this logging a warning instead of blowing up will be better than throwing, because that means that default filters with parameters cannot be disabled.

magicDGS:
@droazen I think that this should have the label bug, and actually I think that is quite important because it does not allow to use properly the `--disableReadFiter` argument....

PR #2359, opened 2017-01-26
Fix ReadFilter plugin descriptor blowing up for default arguments when not provided

magicDGS:
Solves #2357.


magicDGS:
Can you have a look to this one, @cmnbroad? This is quite important because it does not allow to disable default filters like `MappingQualityReadFilter` because of the parameter.

magicDGS:
Closing. I will open a PR between today and tomorrow with all the workaround for this plugin (before the change in Barclay). We can discuss in the new one the details in a better way.

Issue #2358, opened 2017-01-26
Allowed values for --disableReadFilter should only include default filters


magicDGS:
Because it does not make sense to disable a filter that is not enabled. The only withdraw is that if no default filter is provided, this argument does not have possible values.

If #2355 is accepted, at least the part for the argument collection, this could be solved on the construction for the plugin by providing a plugin argument collection without the `--disableReadFilter` argument.

PR #2360, opened 2017-01-26
Allowed values for --disableReadFilter should only include default filters

magicDGS:
Solves #2358

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
@@ -190,9 +190,11 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
     // Return the allowable values for readFilterNames/disableReadFilter
     @Override
     public Set<String> getAllowedValuesForDescriptorArgument(final String longArgName) {
-        if (longArgName.equals(StandardArgumentDefinitions.READ_FILTER_LONG_NAME) ||
-                longArgName.equals(StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME)) {
+        if (longArgName.equals(StandardArgumentDefinitions.READ_FILTER_LONG_NAME)) {
             return readFilters.keySet();
+        }
+        if (longArgName.equals(StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME)) {
+            return toolDefaultReadFilters.keySet();
         }
         throw new IllegalArgumentException("Allowed values request for unrecognized string argument: " + longArgName);
     }
--- a/src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
@@ -14,12 +14,33 @@ import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 
 public class ReadFilterPluginUnitTest {
+
+    @DataProvider(name = "defaultFiltersForAllowedValues")
+    public Object[][] defaultFiltersForAllowedValues() {
+        return new Object[][] {
+                {Collections.emptyList()},
+                {Collections.singletonList(ReadFilterLibrary.GOOD_CIGAR)},
+                {Arrays.asList(ReadFilterLibrary.MAPPED, ReadFilterLibrary.GOOD_CIGAR)},
+                {Arrays.asList(ReadFilterLibrary.GOOD_CIGAR, ReadFilterLibrary.MAPPED)}
+        };
+    }
+
+    @Test(dataProvider = "defaultFiltersForAllowedValues")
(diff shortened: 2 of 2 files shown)


droazen:
I think that the "disable all read filters" argument should do exactly that -- disable them all. If you want an argument to disable only tool default read filters, I'd propose adding a separate `--disableToolDefaultReadFilters` argument.

magicDGS:
I see your point, I was also thinking about the same. But why should an user provide a `--disableReadFilters` and a `--readFilter` argument? If they are not going to be applied anyway, I don't see what is the advantage of having that argument...

But I would be ok for me if a different option is provided instead.

droazen:
Perhaps this is a case where we could use a mutex, and mark `--disableAllReadFilters` and `--readFilter` as mutually exclusive? I agree the combination is not useful. Or we could just rename `--disableAllReadFilters` to `--disableToolDefaultReadFilters`, if it's not useful to have both.

magicDGS:
I think that they should be definitely mutex, but also it should be a way to easily disable the defaults and get the ones requested by the user alone. Renaming the argument sounds good to me, but I don't know if it could have downstream problems for Broad or other users...

droazen:
I realized I've been commenting on the wrong PR (you opened so many at once! :) ). My comment above are meant to apply to https://github.com/broadinstitute/gatk/pull/2365

droazen:
As for this one, with this change if a user were specify something crazy like `--readFilter X --disableReadFilter X`, would it not disable read filter X? That seems wrong to me..

magicDGS:
This is only changing codepaths related with the help. So this change will change the usage to say that all filters are valid for disabeFilter to say that only the available ones are valid. The only problem could be in the docgen code, but not in the behavior of the plugin.

cmnbroad:
Change looks good - but can you please add a test, preferably for both longArgNames since it looks like there is no test for this method at all? Thx.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -190,10 +190,12 @@ public boolean isDependentArgumentAllowed(final Class<?> dependentClass) {
...
     @Override
     public Set<String> getAllowedValuesForDescriptorArgument(final String longArgName) {
-        if (longArgName.equals(StandardArgumentDefinitions.READ_FILTER_LONG_NAME) ||
-                longArgName.equals(StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME)) {
+        if (longArgName.equals(StandardArgumentDefinitions.READ_FILTER_LONG_NAME)) {
             return readFilters.keySet();
         }
+        if (longArgName.equals(StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME)) {
+            return toolDefaultReadFilters.keySet();
+        }
```
cmnbroad: Looks like a good change, but needs a test. Thx.

magicDGS: For disable filters the test is easy and I will do it. For the other one, it is a bit tricky: it returns all the read filters implemented and it could break whenever a new one is included into the package. Any idea on how to do it without the need of change every time a new `ReadFilter` in the package is added?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -190,10 +190,12 @@ public boolean isDependentArgumentAllowed(final Class<?> dependentClass) {
     // Return the allowable values for readFilterNames/disableReadFilter
     @Override
     public Set<String> getAllowedValuesForDescriptorArgument(final String longArgName) {
-        if (longArgName.equals(StandardArgumentDefinitions.READ_FILTER_LONG_NAME) ||
-                longArgName.equals(StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME)) {
+        if (longArgName.equals(StandardArgumentDefinitions.READ_FILTER_LONG_NAME)) {
             return readFilters.keySet();
         }
+        if (longArgName.equals(StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME)) {
+            return toolDefaultReadFilters.keySet();
```
droazen: I think we should allow the user to disable filters that are already disabled, and enable filters that are already enabled, with just a warning. This allows pipeline authors to guarantee that a certain filter will be on/off, regardless of how the tool defaults change over time. I think the only time we should blow up is if the user both enables and disables the same filter on the command line.

Is this consistent with current behavior @cmnbroad? It's not clear to me that this method is only generating allowed values for display purposes (as opposed to validation purposes) -- can we make this clearer, if it's the case?

magicDGS: Last commit adds a test for this. Disabling a filter that is not enabled just log a warning. Anyway, the tool display in the help that the "allowed values" are the default ones. This is something quite nice to do not output all the filters twice in the help (both in `--readFilter` and `--disableReadFilter`) and give to the user the default filters that are enabled to make clear what can be disabled without a warning.

magicDGS:
I added a test for @droazen question. @cmnbroad, if you find a way to test the return values for the long name, will be nice; if not, I think that this is ready to go.

magicDGS:
Friendly reminder here: even if the interface is going to change, this method only needs to be renamed.

magicDGS:
Or should I close like in #2359?

cmnbroad:
@magicDGS I would say close and integrate into the single new PR.

magicDGS:
Ok, closing.

Issue #2362, opened 2017-01-26
Request: read filter plugin method for get a list with default filters and CMD instances


magicDGS:
`GATKReadFilterPluginDescriptor.getAllInstances()` returns only the filters provided by the user, but I expect it to return the default ones.

I know that they are added to the merged filter in `getMergedReadFilter`, but this does not allow to retrieve it as a list. In addition, there is no way to access the default filters provided.

I suggest to move the code to merge into the same list the default and the user-provided filters to `getAllInstances()` and use that list in the `getMergedReadFilter`.

## EDITED:
The contract of `getAllInstances()` says that it should not include the default ones, so I propose a new specific method for retrieval.

cmnbroad:
@magicDGS A method like this should probably go into the base CommandLinePluginDescriptor class in Barclay first. 

Having said that, I'm going to close this issue here, since I've recently added a new method in Barclay called [getDefaultInstances](https://github.com/broadinstitute/barclay/blob/976fafea23216cf0577b8e3fe635b115c7975a76/src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java#L170) (not available until GATK upgrades, which I think will be soon).  That, in combination with getAllInstances, might be sufficient for what you need. If not, please propose the new method in Barclay (or reopen this one here if I'm missing something).

Either way, I think your original point about getAllInstances is valid, since the name doesn't reflect the contract, but for that I think we should just change the name of it in Barclay to something like "getUserInstances" to reflects what it actually does.

magicDGS:
Sounds like a good idea, thanks! I will use a copy of the current plugin in my project til all these changes of the plugin descriptor are in (both Barclay and my proposals here).

magicDGS:
@cmnbroad I have a question regarding `getDefaultInstances()`. Why the return type is a `List<Object>` and not a `List<T>`. That will make simpler for me using it, if not maybe we require the changes in my PR.

cmnbroad:
@magicDGS I made the return type "List<Object>" because the existing consumers that I know of (clp and docgen) have no explicit knowledge of the actual type used for any given plugin descriptor. The proper way to handle it would be to use the bounded type "List<? super T>"; if you just make it "List<T>", then casual consumers like the clp would get a compile error for this (its not covariant):

    List<Object> plugins = descriptor.getDefaultInstances();

[...]

PR #2364, opened 2017-01-26
Add getReadFiltersToUse() to read filter plugin

magicDGS:
Solves to some extend #2362, by adding a new getter.


magicDGS:
This should wait til #2365 is in.

cmnbroad:
Close per comments in #2362.

Issue #2363, opened 2017-01-26
Read filter descriptor isDisabledFilter method should honor --disableAllReadFilters


magicDGS:
Now if the user arguments provide the option `--disableAllReadFilters` and the `isDisableFilter` pass any of the default/user filters, it returns false. If some filter is required for the tool or a warning should be log, this does not allow to do it in the proper way.

I propose to honor this option in the current implementation.

magicDGS:
Closed by #2401

Issue #2361, opened 2017-01-26
--disableAllReadFilters should disable only defaults


magicDGS:
In the `ReadFilter` plugin, the `--disableAllReadFilters` option disable all read filters that are provided by the tool and by the user. From my point of view, it does not make sense to disable all the filters and provide some, but I think that the option may be a shortcut to disable all the provided ones and allow the user to set their own filters (or none, if they prefer it).

PR #2365, opened 2017-01-26
--disableAllReadFilters should allow selective read filter enabling

magicDGS:
Solves #2361.

## EDITED:
Solves also #2363

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
@@ -35,18 +35,20 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
 
     @Argument(fullName = StandardArgumentDefinitions.READ_FILTER_LONG_NAME,
             shortName = StandardArgumentDefinitions.READ_FILTER_SHORT_NAME,
-            doc="Read filters to be applied before analysis", optional=true)
+            doc="Read filters to be applied before analysis", optional = true, common = true)
     public final List<String> userReadFilterNames = new ArrayList<>(); // preserve order
 
     @Argument(fullName = StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME,
             shortName = StandardArgumentDefinitions.DISABLE_READ_FILTER_SHORT_NAME,
-            doc="Read filters to be disabled before analysis", optional=true)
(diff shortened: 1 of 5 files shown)


magicDGS:
Sorry for the confusion, @droazen. I was also thinking that you were commenting in this one. I will do the changes that you told me soon, but I didn't want to change an argument name without asking first.

magicDGS:
Some changes after the comments from @droazen (in commit #2360):
* Change the argument name from `--disableAllReadFilters` to `--disableToolDefaultReadFilters`
* Make `--disableToolDefaultReadFilters` mutex w.r.t. `--disableReadFilter`
* As pointed out in a different PR (#2355), make all the plugin arguments common.
* Make `isDisabled()` and `getAllInstances()` honor the `--disableToolDefaultReadFilters` argument (solves #2363)

I think that this makes more sense than disable absolutely all the read filters, including the ones provided by the user. The cases where this is more useful are:
- Process the data without filters: provide just `--disableToolDefaultReadFilters`
- Process the data without default filters and add any other: provide `--disableToolDefaultReadFilters` and `--readFilter` with the rest of filters
- Process the data with the default filters and include more: as in the previous behaviour, provide `--readFilter` with the rest of filters
- Process the data with some default filters but change the order: provide `--disableToolDefaultReadFilters` and the list of filters in the new order. 

Can you have a look to this one, @cmnbroad and/or @droazen?

cmnbroad:
@magicDGS Thanks for finding and fixing these. It looks like this is on the right track, but a couple of requests before I review the code in detail:

- I find it much easier to review PRs that have a single initial commit, and then one additional commit per round of review (especially since I often pull the branch into the IDE to inspect code, etc.). I realize there are times when you may want to keep some things separate, but generally one commit per round is easier. 
- Also, its easier to keep each PR independent and not duplicate changes (i.e., there two PRs now that have the same common attribute changes that I asked you to make). Again, there may be exceptions, but generally its easier to minimize duplication.

Can you squash this down to one commit, and then I can review ? Also please see my comments in https://github.com/broadinstitute/gatk/issues/2362, since they affect these changes. Thanks!

magicDGS:
I squash my commits into one, and rebase into the latest master (to be sure that the branch does not have conflicts). Here I'm not adding any method as in #2362, and it is different from the commit that have the change in the argument names (this does not include any ArgumentCollection).

Sorry for the inconvenience of opening too many PRs at the same time.

magicDGS:
Closing. I will open a PR between today and tomorrow with all the workaround for this plugin (before the change in Barclay). We can discuss in the new one the details in a better way.

Issue #2356, opened 2017-01-25
LocusWalker very slow when using a large interval file


LeeTL1220:
Included is a JProfiler screenshot of where it is spending all of its time.

As you can see below, it has not even run any of the ```LocusWalker.apply(...)``` code.
```
lichtens@OncobuntuMk3:~/IdeaProjects/hellbender-protected$ java -jar build/libs/gatk-protected.jar Pileup -I ~/broad_oncotator_configs/hcc_purity/SM-74NEG.bam -O ~/broad_oncotator_configs/pileup.out -L /home/lichtens/broad_oncotator_configs/allchr.1kg.phase3.v5a.snp.maf10.biallelic.recode.fixed.prune5.trim1M.test.interval_list
15:04:36.262 INFO  IntelGKLUtils - Trying to load Intel GKL library from:
	jar:file:/home/lichtens/IdeaProjects/hellbender-protected/build/libs/gatk-protected-all-0556d5b-SNAPSHOT-spark_standalone.jar!/com/intel/gkl/native/libIntelGKL.so
15:04:36.344 INFO  IntelGKLUtils - Intel GKL library loaded from classpath.
[January 25, 2017 3:04:36 PM EST] org.broadinstitute.hellbender.tools.walkers.qc.Pileup  --output /home/lichtens/broad_oncotator_configs/pileup.out --intervals /home/lichtens/broad_oncotator_configs/allchr.1kg.phase3.v5a.snp.maf10.biallelic.recode.fixed.prune5.trim1M.test.interval_list --input /home/lichtens/broad_oncotator_configs/hcc_purity/SM-74NEG.bam  --showVerbose false --outputInsertLength false --maxDepthPerSample 0 --interval_set_rule UNION --interval_padding 0 --interval_exclusion_padding 0 --readValidationStringency SILENT --secondsBetweenProgressUpdates 10.0 --disableSequenceDictionaryValidation false --createOutputBamIndex true --createOutputBamMD5 false --createOutputVariantIndex true --createOutputVariantMD5 false --lenient false --addOutputSAMProgramRecord true --help false --version false --verbosity INFO --QUIET false --use_jdk_deflater false --disableAllReadFilters false
[January 25, 2017 3:04:36 PM EST] Executing as lichtens@OncobuntuMk3 on Linux 3.19.0-30-generic amd64; Java HotSpot(TM) 64-Bit Server VM 1.8.0_51-b16; Version: Version:0556d5b-SNAPSHOT
[...]

PR #2366, opened 2017-01-26
enabling bam index caching in memory for GATKTool

lbergelson:
this seems to resolve the extremely slow runtimes when using large interval lists #2356

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
@@ -8,9 +8,6 @@ import htsjdk.samtools.util.Locatable;
 import htsjdk.tribble.Feature;
 import htsjdk.variant.variantcontext.writer.Options;
 import htsjdk.variant.variantcontext.writer.VariantContextWriter;
-import java.io.IOException;
-import java.nio.channels.SeekableByteChannel;
-import java.util.function.Function;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.ArgumentCollection;
 import org.broadinstitute.barclay.argparser.CommandLinePluginDescriptor;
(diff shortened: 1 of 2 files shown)


lbergelson:
@LeeTL1220 4.alpha.2-146-gd9046b4-SNAPSHOT is uploading now.  Give it a shot.

LeeTL1220:
@lbergelson Trying it now...

droazen:
If we go with this solution we'll need to carefully study the effect on memory usage -- BQSR in particular is run in a lower-memory environment than previously.

lbergelson:
@droazen I thought you might say that.  We might be able to come up with an intermediate caching solution where not the entire index is cached.  Although the index isn't really that big, so I'm not sure if it's a big deal.

LeeTL1220:
@lbergelson @droazen How about a flag to turn it off and on?

droazen:
Let's just measure the effect on memory usage with a single bam. Can use something like `PrintReads` to do the test. Then we can make a decision based on the result.

lbergelson:
@droazen I'm not sure what sort of test I can have for this change.  Any thoughts? 

It's good that I looked at this again, it had a stupid bug in it the way I did it at first.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -203,13 +204,18 @@ void initializeReference() {
      */
     void initializeReads() {
         if (! readArguments.getReadFiles().isEmpty()) {
-            SamReaderFactory factory = SamReaderFactory.makeDefault().validationStringency(readArguments.getReadValidationStringency());
+            final SamReaderFactory factory = SamReaderFactory.makeDefault().validationStringency(readArguments.getReadValidationStringency());
             if (hasReference()) { // pass in reference if available, because CRAM files need it
-                factory = factory.referenceSequence(referenceArguments.getReferenceFile());
+                factory.referenceSequence(referenceArguments.getReferenceFile());
```
droazen: Recommend reverting to `factory = factory.referenceSequence()` out of paranoia that the return value might not be `this`

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -96,6 +91,12 @@
     @Argument(fullName = StandardArgumentDefinitions.CLOUD_INDEX_PREFETCH_BUFFER_LONG_NAME, shortName = StandardArgumentDefinitions.CLOUD_INDEX_PREFETCH_BUFFER_SHORT_NAME, doc = "Size of the cloud-only prefetch buffer (in MB; 0 to disable). Defaults to cloudPrefetchBuffer if unset.", optional=true)
     private int cloudIndexPrefetchBuffer = -1;
 
+    @Argument(fullName = StandardArgumentDefinitions.DISABLE_BAM_INDEX_CACHING_LONG_NAME,
+            shortName = StandardArgumentDefinitions.DISABLE_BAM_INDEX_CACHING_SHORT_NAME,
+            doc = "If true, don't cache bam indexes, this will save memory but may decrease performance if many intervals are specified.",
```
droazen: Add note that this argument has no effect if intervals are not specified.

droazen:
Merge after addressing final comment :+1:

PR #2367, opened 2017-01-26
new JNI for bwamem

tedsharpe:
This is ready to go except that I've stuffed the compiled dynamic libraries into src/main/resources.  I'll create a maven central artifact instead, but that will take a little time, and we can start reviewing the code now to save time.
The C code that implements the native methods is in project broadinstitute/gatk-jni.


cwhelan:
@lbergelson or @droazen would either of you want to take a look at this too?

lbergelson:
@tedsharpe How's the performance of this new binding?  I'm assuming it's better and less broken than the old one.  Do you have any numbers or even just anecdotes?

tedsharpe:
@lbergelson Sample size = 1 anecdote:  Measured as 20x faster.
That's the ratio of **total Spark engine run time** for the old code relative to the new code, run locally on my Mac on an unaligned bam that's about 10% of the normal whole-genome size.  If we substracted out the (fixed) time of doing initialization and teardown, and everything but the actual aligning, the number would look better.  My guess is that we're about 1.5 orders of magnitude faster.  Not bad.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
```diff
@@ -1,206 +1,113 @@
...
-                final BwaIndex index = new BwaIndex(localRef);
-                final BwaMem mem = new BwaMem(index);
-                return Utils.concatIterators(alignChunks(mem, iter));
-            } catch (final IOException e) {
-                throw new GATKException("Cannot run BWA-MEM", e);
-            }
-        });
+    @Override
+    public void close() {
+        broadcastHeader.destroy();
```
lbergelson: Is it necessary to manually destroy broadcasts?  Did you find that spark doesn't clean them up in a reasonable time frame?

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
```diff
@@ -1,206 +1,113 @@
...
                 }
-                int size = 0;
-                final List<Tuple2<ShortRead, ShortRead>> list = new ArrayList<>();
-                while (iterator.hasNext() && size < fixedChunkSize) {
-                    final Tuple2<ShortRead, ShortRead> pair = iterator.next();
-                    list.add(pair);
-                    size += pair._1.getBases().length;
-                    size += pair._2.getBases().length;
+                final BwaMemIndex.BwaMemAligner aligner = bwaMemIndex.createAligner();
+                aligner.setFlagOption(aligner.getFlagOption()|BwaMemIndex.BwaMemAligner.MEM_F_PE);
```
lbergelson: could you add a comment about what this flag is?

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
```diff
@@ -1,206 +1,113 @@
...
-        this.numThreads = numThreads;
-        this.fixedChunkSize = fixedChunkSize;
-        this.referenceFileName = referenceFileName;
-    }
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+public final class BwaSparkEngine implements AutoCloseable {
```
lbergelson: This comment should probably be updated instead of removed.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
```diff
@@ -1,206 +1,113 @@
...
-     * @param iterator the read pairs
-     * @return an {@link Iterator} of chunks of read pairs
-     */
-    private Iterator<List<Tuple2<ShortRead, ShortRead>>> chunk(final Iterator<Tuple2<ShortRead, ShortRead>> iterator) {
-        return new UnmodifiableIterator<List<Tuple2<ShortRead, ShortRead>>>() {
-            @Override
-            public boolean hasNext() {
-                return iterator.hasNext();
+            final int nReads = inputReads.size();
+            if ( (nReads & 1) != 0 ) {
```
lbergelson: I would have gone with `%` but  this is fine too...

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
tedsharpe: I'm under the impression that it is necessary.  This guy thinks so, too:
http://apache-spark-developers-list.1001551.n3.nabble.com/Broadcast-Variable-Life-Cycle-td18683.html
But in any case, it just seems cleaner to tell Spark when you know you're done with it, rather than letting it hang around indefinitely.  Even if the finalizer on the driver side calls destroy, the driver VM isn't the VM that's going to be starved for memory and running the GC aggressively (typically).  I want to free resources on the executors as soon as I possibly can.
That's my reasoning, anyway.  I may be all wet.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
```diff
@@ -0,0 +1,349 @@
...
+                if ( result != null ) {
+                    globalInstance = null;
+                    result.close();
+                }
+            }
+        }
+    }
+
+    public static void closeAllDistributedInstances( final JavaSparkContext ctx ) {
+        int nJobs = ctx.defaultParallelism();
```
lbergelson: Does this actually guarantee that it will run on every executor?  It seems plausible that it may end up running several partitions on the same executor, especially if it's a quick operation.   Do you know?

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
```diff
@@ -0,0 +1,349 @@
...
+        final List<Integer> jobList = new ArrayList<>(nJobs);
+        for ( int idx = 0; idx != nJobs; ++idx ) jobList.add(idx);
+        ctx.parallelize(jobList, nJobs).foreach(idx -> BwaMemIndex.closeInstance());
+    }
+
+    static {
+        final String libName;
```
lbergelson: I think this should probably be a method called by the constructor instead of being initialized statically.  Static initializers that throw are not very nice.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
tedsharpe: Yup.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
```diff
@@ -0,0 +1,349 @@
...
+        if (NativeUtils.runningOnLinux()) libName = "/libbwa.Linux.so";
+        else if (NativeUtils.runningOnMac()) libName = "/libbwa.Darwin.dylib";
+        else libName = null;
+        if ( libName == null ) {
```
lbergelson: There should be a way to override the library path in case you have a different architecture (ppc for example).  Most libraries have used an environment variable or a system property to allow it to be injected.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
```diff
@@ -0,0 +1,349 @@
...
+    public static void closeAllDistributedInstances( final JavaSparkContext ctx ) {
+        int nJobs = ctx.defaultParallelism();
+        final List<Integer> jobList = new ArrayList<>(nJobs);
+        for ( int idx = 0; idx != nJobs; ++idx ) jobList.add(idx);
+        ctx.parallelize(jobList, nJobs).foreach(idx -> BwaMemIndex.closeInstance());
+    }
+
+    static {
+        final String libName;
+        if (NativeUtils.runningOnLinux()) libName = "/libbwa.Linux.so";
```
lbergelson: This is going to match for power but the library will fail to load.  Could you preempt the IBM complaint and add a custom exception for NativeUtils.runningOnPPCArchitecture().

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
```diff
@@ -0,0 +1,349 @@
...
+        else libName = null;
+        if ( libName == null ) {
+            throw new UserException.HardwareFeatureException("We have a JNI binding for bwa-mem only for Linux and Mac.");
+        }
+        if ( !NativeUtils.loadLibraryFromClasspath(libName) ) {
+            throw new UserException.HardwareFeatureException("Misconfiguration: Unable to load bwa-mem native library "+libName);
+        }
+    }
+
+    public BwaMemIndex( final String indexImageFile ) {
```
lbergelson: Is there ever a reason to call this over `getInstance`?  Is it safe to have 2 of them?  It seems like it should be private.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/BwaAndMarkDuplicatesPipelineSpark.java
```diff
@@ -35,29 +37,32 @@
...
     public boolean requiresReference() { return true; }
 
-    /**
-     *  command-line arguments to specify the BWA behavior
-     */
-    @ArgumentCollection
-    private BwaArgumentCollection bwaArgs = new BwaArgumentCollection();
+    @Argument(doc = "the bwa mem index image file name that you've distributed to each executor",
+            fullName = "bwamemIndexImage")
+    private String indexImageFile;
```
lbergelson: Is there a suggestion for distributing it?  Use --files?

tedsharpe: Yes.  That's exactly it.

lbergelson:
@tedsharpe 1.5 orders of magnitude faster sounds pretty damned good.

src/main/java/org/broadinstitute/hellbender/utils/bwa/Alignment.java
```diff
@@ -0,0 +1,142 @@
...
+import org.broadinstitute.hellbender.exceptions.GATKException;
+import org.broadinstitute.hellbender.utils.BaseUtils;
+import org.broadinstitute.hellbender.utils.QualityUtils;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import org.broadinstitute.hellbender.utils.read.SAMRecordToGATKReadAdapter;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class Alignment {
```
tedsharpe: Yes.  I punted on this.  What you'd ideally want to do is to prepare a SAMFileHeader in which the the @SQ tags match the reference against which we aligned.  I can't figure out how to get that info.  (The index information is incomplete.)
Next best might be to check that the reference contig names in the BWA index match the ones in the SAMFileHeader.  Then we might guess that we have a compatible coordinate system.  But then I still don't know where to get complete @SQ info in the case of an unaligned BAM, which won't have @SQ records in its header.
This is a bit of a mess.
For now, I'll just document it as you suggest.

src/main/java/org/broadinstitute/hellbender/utils/bwa/Alignment.java
```diff
@@ -0,0 +1,142 @@
...
+    private final int samFlag;     // flag bits per SAM format standard
+    private final int refId;       // index into reference dictionary (-1 if unmapped)
+    private final int refStart;    // 0-based coordinate, inclusive (-1 if unmapped)
+    private final int refEnd;      // 0-based coordinate, exclusive (-1 if unmapped)
+    private final int tigStart;    // 0-based coordinate, inclusive (-1 if unmapped)
```
tedsharpe: Ha, ha.  **Con**tig.  Sigh.  I'll figure out a better name.

src/main/java/org/broadinstitute/hellbender/utils/bwa/Alignment.java
```diff
@@ -0,0 +1,142 @@
...
+    public int getNMismatches() { return nMismatches; }
+    public int getAlignerScore() { return alignerScore; }
+    public int getSuboptimalScore() { return suboptimalScore; }
+    public Cigar getCigar() { return cigar; }
+    public String getMDTag() { return mdTag; }
+    public int getMateRefId() { return mateRefId; }
+    public int getMateRefStart() { return mateRefStart; }
+    public int getTemplateLen() { return templateLen; }
+
+    public GATKRead apply( final GATKRead originalRead, final List<String> refNames, final SAMFileHeader header,
```
tedsharpe: OK

src/main/java/org/broadinstitute/hellbender/utils/bwa/Alignment.java
```diff
@@ -0,0 +1,142 @@
...
+                           final boolean softclipAlts, final boolean justLikeBWA ) {
+        final SAMRecord samRecord;
+        try {
+            samRecord = (SAMRecord)originalRead.convertToSAMRecord(header).clone();
+        } catch ( CloneNotSupportedException x ) {
+            throw new GATKException("can't clone SAMRecord", x);
+        }
+        samRecord.setFlags(samFlag);
+        if ( refId >= 0 ) samRecord.setReferenceName(refNames.get(refId));
```
tedsharpe: I added documentation indicating that realigning isn't reliable.  But this is a mess, it's true.  Maybe I'll try to work out how to create an unaligned SAMRecord from a previously aligned GATKRead.  This big problem is that I want to copy tags that pertain to the unaligned read (like read group), but not those that pertain to the old alignment.  No idea how to do that.

src/main/java/org/broadinstitute/hellbender/utils/bwa/Alignment.java
```diff
@@ -0,0 +1,142 @@
...
+        else if ( mateRefId >= 0 ) samRecord.setReferenceName(refNames.get(mateRefId));
+        if ( refStart >= 0 ) samRecord.setAlignmentStart(refStart+1);
+        else if ( mateRefStart >= 0 ) samRecord.setAlignmentStart(mateRefStart+1);
+        if ( mapQual >= 0 ) samRecord.setMappingQuality(mapQual);
+        byte[] seq = samRecord.getReadBases();
+        byte[] quals = samRecord.getBaseQualities();
+        if ( (samFlag & SAMFlag.READ_REVERSE_STRAND.intValue()) != 0 &&
```
tedsharpe: OK

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
tedsharpe: No, I don't really know.  There's no locality info associated with the parallelized array, so you might hope that it gets divvied out "fairly".  I haven't been able to come up with anything better.  Ideas?  Anyone?  Anyone?

src/main/java/org/broadinstitute/hellbender/utils/bwa/Alignment.java
mwalker174: This may be getting beyond the scope of this PR, but there is a standard set of SAM tags [here](https://github.com/samtools/hts-specs/blob/master/SAMtags.pdf). Maybe we should sort out which tags are alignment-specific and provide a standard set of definitions programmatically in the engine somewhere (e.g. the SAMRecord adapter class). Thoughts?

src/main/java/org/broadinstitute/hellbender/utils/bwa/Alignment.java
tedsharpe: I'm going to declare that creating an appropriate SAMFileHeader for the BAM that's written by BwaSpark after doing the aligning is out of scope for this PR.  It's that program that's broken, it's not broken any worse than it was, and the breakage doesn't really have anything to do with alignment.  Reviewers OK with that?

src/main/java/org/broadinstitute/hellbender/utils/bwa/Alignment.java
tedsharpe: After some thought, I've decided to do this:
I'm creating a brand-spanking new SAMRecord and setting all of its fixed fields appropriately using the original read or the alignment info, as appropriate.  I'm also copying over the RG tag, if it's present.  I'm happy to copy a few more tags if there are such tags that typically appear in our unaligned BAMs.  BC?  OQ?

Rationale:  We're trying to replace the "bwa mem" command line, and if I've interpreted the code correctly, RG is the only tag that can be attached (by setting a single static @RG to use).  This is an improvement over that.

@mwalker174 I started to review the doc you referenced, but it just seemed too daunting.  Besides, there are all the non-defined "custom" tags (XS, for example, which is, in fact, produced by bwa).

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
tedsharpe: I've moved the 3 static methods that deal with a singleton instance into a separate class, BwaMemIndexSingleton.  Now BwaMemIndex knows nothing about Spark, and can also be used locally.

I've added a check in BwaMemIndexSingleton.getInstance() that we don't already have a different index loaded.  This will at least make us aware of abandonment in some cases (when we're switching indexes), and will also prevent people from accidentally writing one pipeline that tries to use two different indexes.

Best I can do, for now, I think.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
tedsharpe: Done.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
tedsharpe: Fixed.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
tedsharpe: I've added a system property to allow overriding.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
tedsharpe: Yes, it's safe to have more than one.  Hard to imagine a use-case, but there's nothing wrong about it.
Especially now that I've factored out BwaMemIndexSingleton, I'm inclined to leave BwaMemIndex as an ordinary object.  OK with you?

tedsharpe:
Seems like it might be time to check in the new, improved code that responds to the reviewers' comments.  Without objection, I'll do so in a few minutes.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndex.java
mwalker174: I couldn't find a much better solution, but you could use the number of partitions instead of number of workers. 

I thought maybe Spark wouldn't distribute this evenly to executors evenly in the case there are still tasks being performed. Also is it possible, based on what I [read](https://spark.apache.org/docs/latest/job-scheduling.html#remove-policy) here, that executors can be decommissioned based on timeout, meaning some might get shut down without closing it properly? Not sure if that would ever be the case here.

tedsharpe: I've tested this:  It appears that the message gets out to all threads on all executor Java VMs.  I'm pretty well convinced that this will work fine.
@mwalker174 It doesn't matter whether decommissioned executors close the index or not.  The process will die, and the memory mapping for the index will go poof with it.  We're just trying to reduce memory usage in the executors that are hanging around and may be about to start doing other tasks.

tedsharpe:
Reviewers?  Anything more to say?  Are we ready to rebase/squash/merge?

cwhelan:
I don't have anything else beyond what other reviewers have said, so it's fine with me.

mwalker174:
No further comments here. Looking forward to integrating this into PathSeq.

src/main/java/org/broadinstitute/hellbender/utils/bwa/Alignment.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Transforms an unaligned GATKRead into an aligned one, as specified by this Alignment.
+     * This method will not work reliably on a previously aligned read:  stale tag information about the previous
+     * alignment might get left behind.
+     *
+     * Also, BWA does this odd thing, supplying AS and XS tags for unaligned reads.
+     * To produce SAM records that are byte-identical to the ones created by the bwa mem command line,
+     * call this method with justLikeBWA=true.
+     */
+    public GATKRead apply( final GATKRead originalRead, final List<String> refNames, final SAMFileHeader header,
```
SHuang-Broad: name change suggestion to match the doc: `originalRead` -> `unalignedRead`

SHuang-Broad: another name change suggestion: the argument name `justLikeBWA` isn't quite communicating what it might do IMO, so what about changing it to `generateDefaultASandXS` or something similar?

SHuang-Broad:
@tedsharpe , it seems that the travis error may (or may not) be fixed by an re-fresh of gradle in IDE. It solved a similar problem I run into this morning.

droazen:
One comment on a naming issue while it's still not too late to change: I think `gatk-jni` is too generic a name for the project -- should ideally have `bwa` in its name.

tedsharpe:
I envisioned putting all our native code into this one artifact. Within 
a day of merging the PR for the BWA binding, I'm going to have a PR for 
adding a FermiLite binding.  Happy to call it gatk-bwafermi, or 
gatk-lh3, or whatever you'd recommend, but we really only need a single 
artifact IMHO.  Let me know.

cmnbroad:
One question/observation: Does the JNI layer create a copy of the read base byte arrays ? We're calling GATKRead.getBases, which itself returns a defensive copy of the entire array, so we may have 3 copies of all the read bases in memory at once, plus the ones we create on output for the alignments. It would be easy to bypass the first copy by calling getReadBases on the underlying SAMRecord (which does not make a copy) if the JNI layer is also creating a copy.

tedsharpe:
This is true, and it bothers me to be creating this superfluous, 
short-lived copy.  But getting the "underlying SAMRecord" isn't always a 
trivial operation.  (It is a trivial operation on a 
SAMRecordToGATKReadAdaptor, but not on other GATKRead types.)  I decided 
to try not to lose sleep over it, but your point is well taken.

droazen:
@tedsharpe Well, we already have a bunch of native/JNI code in Intel's GKL (Genomics Kernel Library) project, so calling it `gatk-jni` seems a bit too all-inclusive to me. I think multiple bindings in the same project is fine, we just need the right name (and, erm, I'll get back to you once I have a suggestion on that :)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
```diff
@@ -36,28 +32,21 @@
...
-        this.mapQual = alnRgn.getMQual();
-        this.mismatches = alnRgn.getNm();
-
-        this.assembledContigLength = forwardStrandCigar.getReadLength() + SVVariantCallerUtils.getTotalHardClipping(forwardStrandCigar);
-        this.startInAssembledContig = startOfAlignmentInContig();
-        this.endInAssembledContig = endOfAlignmentInContig();
+        this.assemblyId = assemblyId;
+        this.forwardStrand = (alignment.getSamFlag()& SAMFlag.READ_REVERSE_STRAND.intValue())==0;
+        this.forwardStrandCigar = alignment.getCigar();
+        this.referenceInterval = new SimpleInterval(refNames.get(alignment.getRefId()), alignment.getRefStart()+1, alignment.getRefEnd());
```
tedsharpe: Done

tedsharpe:
Ha, ha.  That's Intel-authored native code.  This artifact is for native 
code authored by the GATK team!
Let me know what name you'd prefer and I'll change it.  Shouldn't hold 
up this PR, though.  I promise to make another tiny PR on gatk to make 
use of the new artifact as soon as it's available.  Which will be soon, 
I hope.

mwalker174:
Going to need this to get this merged soon. If anyone has further comments please try to make it a priority. I hate to be pushy but this will probably become a blocker for me in a few days. Thanks!

tedsharpe:
Reviewers:  are we ready to rebase, squash, and merge?

PR #2369, opened 2017-01-27
Utils method to map a function over a list

davidbenjamin:
@lbergelson Here's another attempt to write functional code without all that Java boilerplate.  Instead of `list.stream().map(func).collect(Collectors.toList())` it's `Utils.map(list, func)`.  This saves 30 characters in a very common pattern and renders a lot of crowded one-liners much more readable.  If you approve, can you review or suggest someone trustworthy?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReadInputArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReadInputArgumentCollection.java
@@ -3,6 +3,7 @@ package org.broadinstitute.hellbender.cmdline.argumentcollections;
 import htsjdk.samtools.ValidationStringency;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.io.IOUtils;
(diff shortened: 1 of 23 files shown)


lbergelson:
So I'm conflicted about this.  It gets rid of boilerplate, but it also creates an extra thing people have to think about, and it's not symmetric with other collections unless we also add it for them to.

It's especially unfortunate because to me `map( list, function)` reads much more poorly either `list.map(function)`.  If java allowed infix operators it would be much nicer...    

If this was scala and we could add extension methods it would be an obvious addition, but I'm not quite sure how I feel about it here.  I sort of lean against it because it's taking a standard javaism and replacing it with a new non-idiomatic thing to save some ( admittedly annoying) characters.  If only stream had a `.toList()` function...  The case of Utils.stream(iterator)  was a much more obvious win for me than this one.  

Lets get a third opinion, @cmnbroad what do you think?

cmnbroad:
I love the improved economy of expression, but I'm not a fan of this kind of thing myself. In addition to the reasons @lbergelson mentioned, its hard to implement these without introducing subtle shape/type-variance issues, i.e., this silently turns a Set into a List, which can manifest downstream in surprising ways.

It would be nice though. If this were Scala, we wouldn't have to do this at all...

lbergelson:
Oh, yeah, I didn't even look at the implementation.  I assumed it was List -> List, not Collection -> List. If this was scala we could implement Collection -> same Collection, but alas, it's not scala.

davidbenjamin:
Fair enough, I trust you guys on this sort of thing.

PR #2371, opened 2017-01-27
Add ShardingArgumentCollection for shard intervals in differtent tools

magicDGS:
As part of my work to do a walker for sliding-window processing, I implemented a very simple argument collection for sharding intervals. As an example, I included in the `AssemblyRegionWalker`to allow users to also set the shard step in case they want to apply it.

It also allows to specify a window-step to `AssemblyRegionWalker`.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
@@ -2,7 +2,10 @@ package org.broadinstitute.hellbender.engine;
 
 import org.broadinstitute.barclay.argparser.Advanced;
 import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.barclay.argparser.ArgumentCollection;
 import org.broadinstitute.barclay.argparser.CommandLineException;
+import org.broadinstitute.hellbender.cmdline.argumentcollections.AssemblyRegionWalkerShardingArgumentCollection;
+import org.broadinstitute.hellbender.cmdline.argumentcollections.ShardingArgumentCollection;
 import org.broadinstitute.hellbender.engine.filters.CountingReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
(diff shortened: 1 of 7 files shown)


magicDGS:
Can you have a look to this, @droazen? I'm splitting the long-old PR for `SlidingWindowWalker` (#1708) into smaller pieces to get then into soon, and actually it solves partially https://github.com/broadinstitute/gatk-protected/issues/626. I think that the one implemented for `HaplotypeCaller`should be an `AsemblyRegionArgumentCollection` with a `ShardingArgumentCollection`inside.

magicDGS:
Will you have time to have a look to this in the following weeks, @droazen? Because the sliding-window walker is going to take more time, I would like to explore the `AssemblyRegionWalker` for window-traversal of reads instead of the sliding-window variant to see if it is enough for my purpose. But I require to tune the arguments for the shards, and with argument collections is the best way.

Thank you in advance. I rebased to the latest master to skip the tests in the cloud and have the proper status for the checks.

magicDGS:
Friendly ping to @droazen here. Because it is not a disruptive or complicated patch, will you have time to review it by next week? Thank you very much in advance!

I need to start using `AssemblyRegionWalker` with custom size, step and padding but with more descriptive parameters for my use case, and at the same time this will be also useful work for #1198 (even if it is not ported to GATK, but I will have equivalent classes if not).

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -45,11 +48,8 @@
  */
 public abstract class AssemblyRegionWalker extends GATKTool {
 
-    @Argument(fullName="readShardSize", shortName="readShardSize", doc = "Maximum size of each read shard, in bases. For good performance, this should be much larger than the maximum assembly region size.", optional = true)
-    protected int readShardSize = defaultReadShardSize();
-
-    @Argument(fullName="readShardPadding", shortName="readShardPadding", doc = "Each read shard has this many bases of extra context on each side. Read shards must have as much or more padding than assembly regions.", optional = true)
-    protected int readShardPadding = defaultReadShardPadding();
+    @ArgumentCollection
+    protected ShardingArgumentCollection shardingArgumentCollection = defaultShardingArgumentCollection();
```
magicDGS: This is the less disruptive way of doing this, but maybe will be useful to have abstract methods for getting size/step/padding and let tool developers decide if they expose them to the user or not. But that requires to remove the "default size/padding" getters and modify the tools implementing them.

I know that this is more complicated because one of the key tools in GATK will be affected by this: HaplotypeCaller. But  that is a good way to have finer control on what will be exposed to the final user, whenever the optimization for intervals is done.

magicDGS:
Thanks for the update, @droazen!

magicDGS:
Closing in favor of a new PR for sliding-window walker.

PR #2372, opened 2017-01-27
Add LocalVariantShard use in the same way as LocalReadShard

magicDGS:
The `SlidingWindowWalker` variant implementation (#1198) requires a similar class to `LocalReadShard` to perform sharding over variants. Here is a simple implementation based on the read one, which also includes:

* Move up to `Shard<T>` some methods in the read shard implementation.
* Extract a common `FilteringIterator` from `ReadFilteringIterator`

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/utils/iterators/FilteringIteratorUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/iterators/FilteringIteratorUnitTest.java
@@ -1,0 +1,77 @@
+package org.broadinstitute.hellbender.utils.iterators;
+
+import htsjdk.samtools.TextCigarCodec;
+import htsjdk.variant.variantcontext.Allele;
+import htsjdk.variant.variantcontext.VariantContext;
+import htsjdk.variant.variantcontext.VariantContextBuilder;
+import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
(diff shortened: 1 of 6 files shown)


magicDGS:
Here there is another part of the #1708 commit, @droazen. Could you have a look too?

magicDGS:
Hey @droazen, could you have a look to this one? It is for my work of doing variant shards in the same as read shards.

magicDGS:
Now that the two repos are merged, this shows that there is no side-effect in the previously protected code...

magicDGS:
Can you have a look to this one, @droazen?

src/test/java/org/broadinstitute/hellbender/utils/iterators/FilteringIteratorUnitTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.function.Predicate;
+
+public class FilteringIteratorUnitTest extends BaseTest {
```
magicDGS: Self-reminder: extend `GATKBaseTest` (requires rebase)

magicDGS:
Closing due to:
*  `LocalReadShard` does not exists anymore
* `LocalVariantShard` might not be useful

PR #2373, opened 2017-01-27
Throw CommandLineException for enabling twice the same read filter

magicDGS:
Only for no default filters. If not, the composed filter will contain twice the filter one after the other.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
@@ -204,6 +204,18 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
      */
     @Override
     public void validateArguments() {
+        // throw if any filter is specified twice
+        final List<String> moreThanOnce = userReadFilterNames.stream()
+                .collect(Collectors.groupingBy(e -> e, Collectors.counting()))
+                .entrySet().stream().filter(e -> e.getValue() != 1)
+                .map(e -> e.getKey() + " (" + e.getValue() + ")")
+                .collect(Collectors.toList());
+        if (!moreThanOnce.isEmpty()) {
+            throw new CommandLineException.BadArgumentValue(
+                    String.format("The read filter(s) are specified more than once: %s",
+                            Utils.join(", ", moreThanOnce)));
+        }
+
         // throw if any filter is both enabled *and* disabled by the user
         final Set<String> enabledAndDisabled = new HashSet<>(userReadFilterNames);
         enabledAndDisabled.retainAll(disableFilters);
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
@@ -204,16 +204,15 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
      */
     @Override
     public void validateArguments() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -204,6 +204,18 @@ public boolean isDependentArgumentAllowed(final Class<?> dependentClass) {
      */
     @Override
     public void validateArguments() {
+        // throw if any filter is specified twice
+        final List<String> moreThanOnce = userReadFilterNames.stream()
+                .collect(Collectors.groupingBy(e -> e, Collectors.counting()))
+                .entrySet().stream().filter(e -> e.getValue() != 1)
+                .map(e -> e.getKey() + " (" + e.getValue() + ")")
+                .collect(Collectors.toList());
```
cmnbroad: I think the counts are overkill; this could be much simpler:

```
final Set<String> uniqueUserFiltersNames = new HashSet<>();
final Set<String> duplicateUserFilterNames = userReadFilterNames.stream().filter(name -> !uniqueUserFiltersNames.add(name)).collect(Collectors.toSet());
```

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
```diff
@@ -164,6 +164,15 @@ public void testEnableNonExistentFilter() {
         clp.parseArguments(System.out, new String[] {"--RF", "fakeFilter"});
     }
 
+    @Test(expectedExceptions = CommandLineException.BadArgumentValue.class)
+    public void testEnableTwiceFilter() {
```
cmnbroad: Maybe "testEnableDuplicateFilter".

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
```diff
@@ -164,6 +164,15 @@ public void testEnableNonExistentFilter() {
...
+        CommandLineParser clp = new CommandLineArgumentParser(new Object(),
+                Collections.singletonList(new GATKReadFilterPluginDescriptor(null)));
+        clp.parseArguments(System.out, new String[] {
+                "--RF", ReadLengthReadFilter.class.getSimpleName(),
+                "--RF", ReadLengthReadFilter.class.getSimpleName()});
```
cmnbroad: Can you add another test case that includes both a duplicate and a non duplicate (through a data provider). thx.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: I did that for warn the user how many times they specified the filter. With a long command line, it could happen that it was more than two and the user will need to check the command line too much. I will change it anyway, but it will be nicer to have the counts.

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
magicDGS: Done.

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java

magicDGS:
Back to you @cmnbroad. Thanks for reviewing.

PR #2374, opened 2017-01-27
MultiPassLocusWalker (PROTOTYPE - DO NOT MERGE)

droazen:
Rough prototype for @LeeTL1220 to test. If it satisfies your use case, we can polish this branch up, add some tests, and merge.


droazen:
Discussed with @LeeTL1220 just now, and we decided that an approach based on the vanilla `LocusWalker` with calls to `splitContextBySampleName()` would result in a cleaner and more easily-parallelizable implementation, without the need for maintaining state across calls. So this branch is getting put on hold for now until we see whether the overhead of having to merge bams and then pull apart the pileups in the single-pass implementation is a significant performance issue.

@LeeTL1220 Let me know the outcome of that evaluation, and we'll decide whether to resurrect or kill this PR.

droazen:
Closing, as @LeeTL1220 ended up not needing this.

PR #2376, opened 2017-01-31
terminal 'I' to 'S' in gapped alignment breaker in SV-calling step

SHuang-Broad:
@cwhelan , this is the code for turning the edge case where an operator 'I' is the beginning or ending operator in the CIGAR of the contig alignment.
Please kindly review. Thanks!

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
@@ -11,6 +11,7 @@ import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.Utils;
 import scala.Tuple2;
 
 import java.io.Serializable;
@@ -42,7 +43,9 @@ class AssemblyAlignmentParser implements Serializable {
 
         final JavaPairRDD<Tuple2<String, String>, Iterable<AlignmentRegion>> alignmentRegionsKeyedByAssemblyAndContigId
                 = parseAlignments(ctx, pathToInputAlignments);
-        if (DEBUG_STATS) AssemblyAlignmentParser.debugStats(alignmentRegionsKeyedByAssemblyAndContigId, pathToInputAlignments);
+        if (DEBUG_STATS) {
+            AssemblyAlignmentParser.debugStats(alignmentRegionsKeyedByAssemblyAndContigId, pathToInputAlignments);
+        }
 
         final JavaPairRDD<Tuple2<String, String>, byte[]> contigSequences
                 = loadContigSequence(ctx, pathToInputAssemblies);
@@ -75,12 +78,10 @@ class AssemblyAlignmentParser implements Serializable {
                 });
     }
 
-    // TODO: 11/26/16 assuming 'I'/'D' cannot be the beginning or last CIGAR operator AND the following operator cannot be a clipping
-    // TODO: 1/18/17 'I' as first/last operator has been seen in actual data, for the purpose of this function, changing it to soft clipping is fine, but needs to be done later (should be very soon)
     // TODO: 1/18/17 handle mapping quality and mismatches of the new alignment better rather than artificial ones. this would ultimately require a re-alignment done right.
     /**
      * Break a gapped alignment into multiple alignment regions, based on input sensitivity: i.e. if the gap(s) in the input alignment
-     * is equal to or larger than the size specified, two alignment regions will be generated.
+     * is equal to or larger than the size specified by {@code sensitivity}, two alignment regions will be generated.
      *
      * To fulfill this functionality, needs to accomplish three key tasks correctly:
      * <ul>
@@ -114,24 +115,25 @@ class AssemblyAlignmentParser implements Serializable {
     @VisibleForTesting
     static Iterable<AlignmentRegion> breakGappedAlignment(final AlignmentRegion oneRegion, final int sensitivity) {
 
-        final List<CigarElement> cigarElements = oneRegion.forwardStrandCigar.getCigarElements();
+        final List<CigarElement> cigarElements = checkCigarAndConvertTerminalInsertionToSoftClip(oneRegion.cigarAlong5to3DirectionOfContig);
         if (cigarElements.size() == 1) return new ArrayList<>( Collections.singletonList(oneRegion) );
 
-        final List<AlignmentRegion> result = new ArrayList<>(10); // blunt guess
+        final List<AlignmentRegion> result = new ArrayList<>(3); // blunt guess
         final int originalMapQ = oneRegion.mapQual;
 
         final List<CigarElement> cigarMemoryList = new ArrayList<>();
-        final int clippedNBases = SVVariantCallerUtils.getNumClippedBases(true, oneRegion.forwardStrandCigar);
+        final int clippedNBasesFromStart = SVVariantCallerUtils.getNumClippedBases(true, cigarElements);
+
         final int hardClippingAtBeginning = cigarElements.get(0).getOperator()==CigarOperator.H ? cigarElements.get(0).getLength() : 0;
         final int hardClippingAtEnd = (cigarElements.get(cigarElements.size()-1).getOperator()== CigarOperator.H)? cigarElements.get(cigarElements.size()-1).getLength() : 0;
         final CigarElement hardClippingAtBeginningMaybeNull = hardClippingAtBeginning==0 ? null : new CigarElement(hardClippingAtBeginning, CigarOperator.H);
-        int     refIntervalStart    = oneRegion.referenceInterval.getStart(),
-                contigIntervalStart = 1 + clippedNBases;
-        int gapCount = 0;
+        int contigIntervalStart = 1 + clippedNBasesFromStart;
+        // we are walking along the contig following the cigar, which indicates that we might be walking backwards on the reference if oneRegion.forwardStrand==false
+        int refBoundary1stInTheDirectionOfContig = oneRegion.forwardStrand ? oneRegion.referenceInterval.getStart() : oneRegion.referenceInterval.getEnd();
         for (final CigarElement cigarElement : cigarElements) {
             final CigarOperator op = cigarElement.getOperator();
             final int operatorLen = cigarElement.getLength();
-            switch (op){
+            switch (op) {
                 case M: case EQ: case X: case S: case H:
                     cigarMemoryList.add(cigarElement);
                     break;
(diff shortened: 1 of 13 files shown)


SHuang-Broad:
@cwhelan , turns out this also corrects some mistakes made in mis-interpreting the old code. Just updated.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -145,8 +149,14 @@
                     final Cigar memoryCigar = new Cigar(cigarMemoryList);
                     final int effectiveReadLen = memoryCigar.getReadLength() + SVVariantCallerUtils.getTotalHardClipping(memoryCigar) - SVVariantCallerUtils.getNumClippedBases(true, memoryCigar);
 
-                    // infer reference interval
-                    final SimpleInterval referenceInterval = new SimpleInterval(oneRegion.referenceInterval.getContig(), refIntervalStart, refIntervalStart + memoryCigar.getReferenceLength()-1);
+                    // infer reference interval taking into account of strand
+                    final int refBoundary2ndInTheDirectionOfContig = refBoundary1stInTheDirectionOfContig + (oneRegion.forwardStrand ? memoryCigar.getReferenceLength()-1 : -memoryCigar.getReferenceLength()+1) ;
```
cwhelan: Can you make the last part of the ternary be explicitly `-1 * memoryCigar.getReferenceLength() + 1`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -168,8 +178,11 @@
                     cigarMemoryList.add(new CigarElement(contigIntervalEnd - hardClippingAtBeginning + (op.consumesReadBases() ? operatorLen : 0), CigarOperator.S));
 
                     // update pointers into reference and contig
-                    refIntervalStart     += op.consumesReadBases() ? memoryCigar.getReferenceLength()  : memoryCigar.getReferenceLength() + operatorLen;
-                    contigIntervalStart  += op.consumesReadBases() ? effectiveReadLen + operatorLen    : effectiveReadLen;
+                    if (oneRegion.forwardStrand)
```
cwhelan: Please add braces here (and to the other if statements in this method)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -114,20 +113,25 @@
...
+        }
         final int hardClippingAtBeginning = cigarElements.get(0).getOperator()==CigarOperator.H ? cigarElements.get(0).getLength() : 0;
         final int hardClippingAtEnd = (cigarElements.get(cigarElements.size()-1).getOperator()== CigarOperator.H)? cigarElements.get(cigarElements.size()-1).getLength() : 0;
         final CigarElement hardClippingAtBeginningMaybeNull = hardClippingAtBeginning==0 ? null : new CigarElement(hardClippingAtBeginning, CigarOperator.H);
-        int     refIntervalStart    = oneRegion.referenceInterval.getStart(),
-                contigIntervalStart = 1 + clippedNBases;
+        int contigIntervalStart = 1 + clippedNBasesFromStart;
         int gapCount = 0;
+        // we are walking along the contig following the cigar, which indicates that we might be walking backwards on the reference if oneRegion.forwardStrand==false
+        int refBoundary1stInTheDirectionOfContig = oneRegion.forwardStrand ? oneRegion.referenceInterval.getStart() : oneRegion.referenceInterval.getEnd();
```
cwhelan: If we reversed the cigar here first, would we be able to get rid of all of the if statements in the body of this loop? It seems like it might make the code a lot cleaner. But I guess you'd still need to keep track of it for the on-contig coordinates. Just something to think about.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -182,15 +195,53 @@
...
+        }
         result.add(new AlignmentRegion(oneRegion.assemblyId, oneRegion.contigId, lastReferenceInterval, lastForwardStrandCigar,
                 oneRegion.forwardStrand, originalMapQ, SVConstants.CallingStepConstants.ARTIFICIAL_MISMATCH,
-                contigIntervalStart, oneRegion.assembledContigLength-SVVariantCallerUtils.getNumClippedBases(false, oneRegion.forwardStrandCigar)));
+                contigIntervalStart, oneRegion.assembledContigLength-clippedNBasesFromEnd));
 
         return result;
     }
 
+    private static boolean parseCigarAndConvertTrailingInsertionToSoftClip(final Cigar cigarAlongInput5to3Direction,
```
cwhelan: It's not just 'trailing', right, but either the first or the last operator?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -182,15 +195,53 @@
...
+            cigarHasInsertionAsTrailingOperator = true;
+            emptyTargetCollection.set(0, new CigarElement(element.getLength(), CigarOperator.S));
+        }
+        element = emptyTargetCollection.get(emptyTargetCollection.size()-1);
+        if (element.getOperator() == CigarOperator.I) {
+            cigarHasInsertionAsTrailingOperator = true;
+            emptyTargetCollection.set(emptyTargetCollection.size()-1, new CigarElement(element.getLength(), CigarOperator.S));
+        }
+
+        return cigarHasInsertionAsTrailingOperator;
```
cwhelan: See my comment earlier asking why we can't just return the new cigar list and get rid of this boolean return value.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantConsensusCall.java
```diff
@@ -133,15 +133,20 @@ static String produceVariantId(final NovelAdjacencyReferenceLocations novelAdjac
...
+    static int getSvLength(final NovelAdjacencyReferenceLocations breakpoints, GATKSVVCFHeaderLines.SVTYPES svtype) {
         final int start = breakpoints.leftJustifiedLeftRefLoc.getEnd();
         final int end = breakpoints.leftJustifiedRightRefLoc.getStart();
 
-        if (start != end) {
-            return end - start;
-        } else {
-            return breakpoints.complication.insertedSequenceForwardStrandRep.length()
-                    + (breakpoints.complication.dupSeqRepeatNumOnCtg - breakpoints.complication.dupSeqRepeatNumOnRef)*breakpoints.complication.dupSeqForwardStrandRep.length();
+        switch (svtype) {
```
cwhelan: Rather than have switch statements on SVTYPES scattered around the code, it might be nice to create abstract methods on the SVType enum and override them when necessary. That way, when we add a new SV type, we don't have to search for all switch statements and make sure they support the new type correctly. Is that a refactoring you'd be comfortable doing as part of this PR? Other places to modify would include the `produceAlleles` and `produceVariantIds` methods.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -27,7 +27,7 @@
 class AssemblyAlignmentParser implements Serializable {
     private static final long serialVersionUID = 1L;
     private static final Logger log = LogManager.getLogger(AssemblyAlignmentParser.class);
-    private static final boolean DEBUG_STATS = false;
+    private static final boolean DEBUG_STATS = true;
```
SHuang-Broad: Nope, that was for running with debugging. Will switch back.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -75,8 +75,7 @@
                 });
     }
 
-    // TODO: 11/26/16 assuming 'I'/'D' cannot be the beginning or last CIGAR operator AND the following operator cannot be a clipping
-    // TODO: 1/18/17 'I' as first/last operator has been seen in actual data, for the purpose of this function, changing it to soft clipping is fine, but needs to be done later (should be very soon)
+    // TODO: 1/31/17 assuming 'I'/'D' cannot be the operator adjacent to clipping operators 'S'/'H', and 'D' cannot be the beginning or last operator
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -114,20 +113,25 @@
     @VisibleForTesting
     static Iterable<AlignmentRegion> breakGappedAlignment(final AlignmentRegion oneRegion, final int sensitivity) {
 
-        final List<CigarElement> cigarElements = oneRegion.forwardStrandCigar.getCigarElements();
+        final List<CigarElement> cigarElements = new ArrayList<>();
+        final boolean cigarHasInsertionAsTrailingOperator = parseCigarAndConvertTrailingInsertionToSoftClip(oneRegion.cigarAlong5to3DirectionOfContig, cigarElements);
```
SHuang-Broad: Probably a bad name. By "trailing" I mean the first non-clipping operator from either end.

SHuang-Broad: And because of the assumption that's now enforced in the argument checking step on gap operators not neighboring the clipping operators, "trailing" actually means not the first or the last operator.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
SHuang-Broad: Yep, that's what I mean by "trailing". Any better naming suggestion?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
SHuang-Broad: The two `if` statements regarding the computation for reference regions would be gone, but the computation for the corresponding contig regions would be made complicated by necessary `if` statements.
inherently, we are dealing with two sets of coordinates which would have opposite directions if the alignment between them is the reverse-strand. So walking along either one following the appropriate representation of the CIGAR would require walking backwards on the other. We might be able to think of a way to code this smartly.

cwhelan: If you can't think of a better way now, that's OK. It just feels like there might be a way to make this simpler eventually.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
SHuang-Broad: I've updated the logic to be permissive about 'I' neighboring 'S'/'H' given that we've seen a SAM record doing it, so the "first non-clipping operator from either end" is what I mean now.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -114,20 +113,25 @@
...
         if (cigarElements.size() == 1) return new ArrayList<>( Collections.singletonList(oneRegion) );
 
         final List<AlignmentRegion> result = new ArrayList<>(10); // blunt guess
         final int originalMapQ = oneRegion.mapQual;
 
         final List<CigarElement> cigarMemoryList = new ArrayList<>();
-        final int clippedNBases = SVVariantCallerUtils.getNumClippedBases(true, oneRegion.forwardStrandCigar);
+        int clippedNBasesFromStart = SVVariantCallerUtils.getNumClippedBases(true, oneRegion.cigarAlong5to3DirectionOfContig);
+        if (cigarHasInsertionAsTrailingOperator && cigarElements.get(0).getOperator()==CigarOperator.S) {
```
SHuang-Broad: Done with a new overload of `SVVariantCallerUtils.getNumClippedBases()`. Was trying to avoid new CIGAR construction as much as possible.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantConsensusCall.java
SHuang-Broad: Done with a new `SVTYPE` class. It is better now. Thanks!

SHuang-Broad:
@cwhelan , I've addressed the comments in this commit. It's not very small due to the refactoring of `SVTYPE` and dealing with the new edge case cigars like `10S10I10M`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantCallerUtils.java
```diff
@@ -38,25 +39,106 @@ static int getTotalHardClipping(final Cigar cigar) {
...
+     * <ul>
+     *     <li>empty input list;</li>
+     *     <li>when having only 1 operations, it must be an alignment;</li>
+     *     <li>deletion operation cannot neighbor clipping operations;</li>
+     *     <li>there must be at least one alignment operation in the list</li>
+     * </ul>
+     * @param cigarElements
+     */
+    @VisibleForTesting
+    static void checkCigarElements(final List<CigarElement> cigarElements) {
```
cwhelan: `validateCigar` might be a better name?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantConsensusCall.java
```diff
@@ -172,41 +98,55 @@ static int getSvLength(final NovelAdjacencyReferenceLocations breakpoints, GATKS
...
+        try {
+            final SVTYPE.RECOGNIZED_SVTYPES x = SVTYPE.RECOGNIZED_SVTYPES.valueOf(type.toString());
+        } catch (final IllegalArgumentException ex) {
+            throw new GATKException.ShouldNeverReachHereException("Inferred type is not known yet: " + type.toString(), ex);
+        }
+        return type;
+    }
+
+    @VisibleForTesting
+    static Map<String, String> setFlags(final NovelAdjacencyReferenceLocations novelAdjacencyReferenceLocations, final SVTYPE svtype) {
```
cwhelan: It seems like this could also be moved to the SVTYPE subclasses -- what do you think?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVTYPE.java
```diff
@@ -0,0 +1,169 @@
...
+        DEL(final NovelAdjacencyReferenceLocations novelAdjacencyReferenceLocations) {
+            final String contig = novelAdjacencyReferenceLocations.leftJustifiedLeftRefLoc.getContig();
+            final int start = novelAdjacencyReferenceLocations.leftJustifiedLeftRefLoc.getEnd();
+            final int end = novelAdjacencyReferenceLocations.leftJustifiedRightRefLoc.getStart();
+
+            final String startString;
+            if (novelAdjacencyReferenceLocations.complication.dupSeqForwardStrandRep.isEmpty()) {
+                startString = toString();
+            } else {
+                startString = "DUP" +
```
cwhelan: As discussed in person, let's start the id with "DEL" or "INS" and then have the repeat expansion or contraction information afterwards.

cwhelan: Also, as noted in person, DEL shouldn't have an expansion check.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVTYPE.java
```diff
@@ -0,0 +1,169 @@
...
+                        ((novelAdjacencyReferenceLocations.complication.dupSeqRepeatNumOnRef < novelAdjacencyReferenceLocations.complication.dupSeqRepeatNumOnCtg) ? SVConstants.CallingStepConstants.TANDUP_EXPANSION_STRING : SVConstants.CallingStepConstants.TANDUP_CONTRACTION_STRING);
```
cwhelan: As noted in person, INS shouldn't have a contraction check.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -145,101 +144,152 @@
...
 
-                    // infer contig interval
+                    // task 2: infer contig interval
                     final int contigIntervalEnd = contigIntervalStart + effectiveReadLen - 1;
 
-                    // now add trailing cigar element and create the real cigar
+                    // task 3: now add trailing cigar element and create the real cigar for the to-be-returned AR
                     cigarMemoryList.add(new CigarElement(oneRegion.assembledContigLength-contigIntervalEnd-hardClippingAtEnd, CigarOperator.S));
-                    if (hardClippingAtEnd != 0)
+                    if (hardClippingAtEnd != 0) { // be faithful to hard clipping (as the accompanying bases have be hard-clipped away)
```
SHuang-Broad: done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -145,101 +144,152 @@
...
-                    contigIntervalStart += op.consumesReadBases() ? effectiveReadLen + operatorLen    : effectiveReadLen;
-
-                    ++gapCount;
+                    final int refBoundaryAdvance = op.consumesReadBases() ? memoryCigar.getReferenceLength() : memoryCigar.getReferenceLength() + operatorLen;
+                    refBoundary1stInTheDirectionOfContig += oneRegion.forwardStrand ? refBoundaryAdvance : -refBoundaryAdvance;
+                    contigIntervalStart += op.consumesReadBases() ? effectiveReadLen + operatorLen : effectiveReadLen;
 
                     break;
                 default:
                     throw new GATKException("Alignment have N or P in it: " + oneRegion.toString()); // TODO: 1/20/17 still not quite sure if this is quite right, it doesn't blow up on NA12878 WGS, but who knows what happens in the future
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -145,101 +144,152 @@
...
-        boolean cigarHasInsertionAsTrailingOperator = false;
-        if (element.getOperator() == CigarOperator.I) {
-            cigarHasInsertionAsTrailingOperator = true;
-            emptyTargetCollection.set(0, new CigarElement(element.getLength(), CigarOperator.S));
+        SVVariantCallerUtils.checkCigarElements(cigarElements);
+
+        final List<CigarElement> convertedList = convertInsToSoftClipFromOneEnd(cigarElements, true);
+        final List<CigarElement> result = convertInsToSoftClipFromOneEnd(convertedList, false);
+        Utils.validateArg(cigarElements.stream().anyMatch(ele -> ele.getOperator().isAlignment()),
+                "No alignment between ref and ctg after 'I' -> 'S' conversion, this suggest new edge case CIGAR record: " + TextCigarCodec.encode(cigarAlongInput5to3Direction));
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -145,101 +144,152 @@
...
+     *
+     * @return the converted and compactified list of cigar elements
+     */
+    @VisibleForTesting
+    static List<CigarElement> convertInsToSoftClipFromOneEnd(final List<CigarElement> cigarElements,
+                                                             final boolean fromStart) {
+        final int numHardClippingBasesFromOneEnd = SVVariantCallerUtils.getNumHardClippingBases(fromStart, cigarElements);
+        final int numSoftClippingBasesFromOneEnd = SVVariantCallerUtils.getNumSoftClippingBases(fromStart, cigarElements);
+
+        final int indexForFirstMaybeInsertion;
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantCallerUtils.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantCallerUtils.java
```diff
@@ -38,25 +39,106 @@ static int getTotalHardClipping(final Cigar cigar) {
...
+        Utils.validateArg(!cigarElements.isEmpty(), "Cannot parse empty list cigarElements");
+        Utils.validateArg(cigarElements.size()>1 || (cigarElements.size()==1 && cigarElements.get(0).getOperator().isAlignment()),
```
SHuang-Broad: Yes, you are right in that the 4th check covers the 2nd and the 2nd is not a good check. 
The case of `10H10S1D10M` is now caught as well, with new test coverage. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantCallerUtils.java
```diff
@@ -38,25 +39,106 @@ static int getTotalHardClipping(final Cigar cigar) {
...
+                "invalid input cigar elements " + cigarElements.toString());
```
SHuang-Broad: case covered by another check now

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantCallerUtils.java
```diff
@@ -38,25 +39,106 @@ static int getTotalHardClipping(final Cigar cigar) {
...
+        Utils.validateArg ( !(cigarElements.get(0).getOperator().isClipping() && cigarElements.get(1).getOperator()==CigarOperator.D)
```
SHuang-Broad: case covered by new check now

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVTYPE.java
```diff
@@ -0,0 +1,169 @@
+package org.broadinstitute.hellbender.tools.spark.sv;
+
+import htsjdk.variant.variantcontext.Allele;
+
+import static org.broadinstitute.hellbender.tools.spark.sv.NovelAdjacencyReferenceLocations.EndConnectionType.FIVE_TO_FIVE;
+
+/**
+ * Created by shuang on 2/7/17.
+ */
+abstract class SVTYPE {
```
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVTYPE.java
```diff
@@ -0,0 +1,169 @@
...
+
+    enum RECOGNIZED_SVTYPES {
```
SHuang-Broad: Changed to TYPES

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVTYPE.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVTYPE.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantConsensusCall.java
SHuang-Broad: You are right! Now I can get rid of the reflections that I disliked. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVTYPE.java
```diff
@@ -0,0 +1,169 @@
...
+        INV, DEL, INS, DUP;
+    }
+
+    abstract String getVariantId();
```
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -145,101 +144,152 @@
...
-        element = emptyTargetCollection.get(emptyTargetCollection.size()-1);
+
+        final CigarElement element = cigarElements.get(indexForFirstMaybeInsertion);
         if (element.getOperator() == CigarOperator.I) {
-            cigarHasInsertionAsTrailingOperator = true;
-            emptyTargetCollection.set(emptyTargetCollection.size()-1, new CigarElement(element.getLength(), CigarOperator.S));
-        }
 
-        return cigarHasInsertionAsTrailingOperator;
+            cigarElements.set(indexForFirstMaybeInsertion, new CigarElement(element.getLength(), CigarOperator.S));
```

SHuang-Broad:
@cwhelan , addressed all comments for the 2nd round. Please take another look.
Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -251,47 +247,63 @@
...
-                }
-                return result;
-            } else {
-                return cigarElements;
-            }
+            cigarElements.set(indexOfFirstNonClippingOperation, new CigarElement(element.getLength(), CigarOperator.S));
+
+            return compactifyNeighboringSoftClippings(cigarElements);
+
+//            if (numSoftClippingBasesFromOneEnd!=0) { // compactify with possible neighboring clipping
```
cwhelan: Please remove commented out code before checking in.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -251,47 +247,63 @@
...
         }
     }
 
+    @VisibleForTesting
+    static List<CigarElement> compactifyNeighboringSoftClippings(final List<CigarElement> cigarElements) {
+        final List<CigarElement> result = new ArrayList<>(cigarElements.size());
+        for (final CigarElement element : cigarElements) {
+            if (result.isEmpty() || result.get(result.size()-1).getOperator()!=element.getOperator()) {
+                result.add(element);
+            } else { // no additional checking because this is possible only for two neighboring soft clippings, one of which is converted from an insertion operation
```
cwhelan: Are you sure about that (that this will never happen except for in the case of a converted soft clip)? We could be a little more conservative here and only do the merge if the adjacent elements are in fact soft clips. I just worry that there's some BWA edge case that could give two adjacent "M" operators or something, and this would squish them incorrectly into a new "S" element. Also the name of the method is specific to soft clips.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/GATKSVVCFHeaderLines.java
```diff
@@ -16,7 +16,7 @@
 
     // VCF standard SV header lines
     // todo: add these and the other standard SV info fields from the VCF spec to htsjdk VCFStandardHeaderLines
-    public static final String SVTYPE = "SVTYPE";
+    public static final String SVTYPE = "SvType";
```
cwhelan: I think this still needs to be all uppercase.

cwhelan: Maybe this was an IDE search/replace overstep.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVConstants.java
```diff
@@ -19,10 +19,10 @@ private SVConstants() {
 
         static final String INVERSIONS_OUTPUT_VCF = "inversions.vcf";
         static final String CURRENTLY_CAPABLE_VARIANTS_VCF = "inv_del_ins.vcf";
-        static final String VCF_ALT_ALLELE_STRING_INV = "<INV>";
-        static final String VCF_ALT_ALLELE_STRING_INS = "<INS>";
-        static final String VCF_ALT_ALLELE_STRING_DEL = "<DEL>";
-        static final String VCF_ALT_ALLELE_STRING_DUP = "<DUP>";
+        static final String VCF_ALT_ALLELE_STRING_INV = "<Inversion>";
```
cwhelan: These need to be INS, DEL, etc. Maybe more search/replace issues?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantCallerUtils.java
```diff
@@ -130,15 +130,37 @@ static int getNumSoftClippingBases(final boolean fromStart, final List<CigarElem
      * @param cigarElements
      */
     @VisibleForTesting
-    static void checkCigarElements(final List<CigarElement> cigarElements) {
+    static void validateCigar(final List<CigarElement> cigarElements) {
```
cwhelan: You could update the comment for this method, which still describes the redundant check you removed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SvType.java
```diff
@@ -0,0 +1,155 @@
...
+        private static String getIDString(final NovelAdjacencyReferenceLocations novelAdjacencyReferenceLocations) {
+            final String contig = novelAdjacencyReferenceLocations.leftJustifiedLeftRefLoc.getContig();
+            final int start = novelAdjacencyReferenceLocations.leftJustifiedLeftRefLoc.getEnd();
+            final int end = novelAdjacencyReferenceLocations.leftJustifiedRightRefLoc.getStart();
+
+            final String startString;
+            if (novelAdjacencyReferenceLocations.complication.dupSeqForwardStrandRep.isEmpty()) {
+                startString = TYPES.INS.name();
+            } else {
+                startString = "DuplicationTandem" +
```
cwhelan: Should this be "TANDEMUPLICATION_EXPANSION_ID_START_STRING"?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
SHuang-Broad: Sorry for not catching this...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
SHuang-Broad: Done with new exception-throwing check.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/GATKSVVCFHeaderLines.java
SHuang-Broad: right, IDE overkill...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVConstants.java
SHuang-Broad: yep

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVVariantCallerUtils.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SvType.java
SHuang-Broad: not sure why i didn't catch this. it is now changed so that tandem duplication is handled completely in the `DuplicationTandem` class.

SHuang-Broad:
Holding off because after rebase the aligner is complaining for a null pointer exception. Need to investigate why.

SHuang-Broad:
Whole pipeline working now with updated way of running the alignment step, which is reflected in the PR #2435

PR #2379, opened 2017-02-01
Trivial formatting update for skip interval file parser

SHuang-Broad:
@tedsharpe please kindly review.
The skip interval that has the corresponding format lives here:
gs://sv-data-dsde-dev/reference/GRCh38.kill.intervals

and was generated by 
gs://sv-data-dsde-dev/extra_input_generator_scripts

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVUtils.java
@@ -21,6 +21,8 @@ import java.util.stream.Collectors;
  */
 public final class SVUtils {
 
+    private static final String REFERENCE_GAP_INTERVAL_FILE_COMMENT_LINE_PROMPT = "#";
+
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVUtils.java
```diff
@@ -21,6 +21,8 @@
  */
 public final class SVUtils {
 
+    private static final String REFERENCE_GAP_INTERVAL_FILE_COMMENT_LINE_PROMPT = "##";
```
tedsharpe: Let's make this a single '#'.  User can still start lines with '##' if he or she prefers.
Otherwise, thumbs up!

SHuang-Broad: Done with updated script and kill interval list in the buckets. Thanks!

Issue #2383, opened 2017-02-02
Request: rename SplitReads argument names  


magicDGS:
For some tools that I'm developing, I want to keep as much as possible the nomenclature for some common parameters in Picard/GATK to integrate my custom tools into a framework that is consistent regarding the argument names. Because of that, I'm having several issues with the `SplitReads`arguments while using splitting by these parameters in my toolkit:

- The current `SplitReads` argument names are in capital, but the rest of the framework use a camel case formatting.
- The short names clash with the `AddOrReplaceReadGroups` from Picard.

Because I have a tool that adds read groups with the possibility of splitting in a similar way as `SplitReads` I would like to integrate these parameters in the same tool and I can't with the current implementation. I propose to change the `SplitReads` names to a camel case format different from Picard and pointing out that they are related with splitting.

magicDGS:
Although it is not completely solved because it is still clashing for the short names, I'm closing here since #2384 is merged and we discussed there that this is not going to happen.

PR #2384, opened 2017-02-02
Modify SplitReads argument names

magicDGS:
As described in #2383, to avoid clash with Picard arguments for add read groups and point out that they are meant to split.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/SplitReads.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/SplitReads.java
@@ -36,9 +36,12 @@ import java.util.stream.Collectors;
 )
 public final class SplitReads extends ReadWalker {
 
-    public static final String SAMPLE_SHORT_NAME = "SM";
-    public static final String READ_GROUP_SHORT_NAME = "RG";
-    public static final String LIBRARY_NAME_SHORT_NAME = "LB";
+    public static final String SAMPLE_LONG_NAME = "splitSample";
+    public static final String SAMPLE_SHORT_NAME = "splitSM";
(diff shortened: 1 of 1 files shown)


lbergelson:
@vdauwera Could you chime in here as it's a UI question and relates to the upcoming picard/gatk UIpocalypse.

magicDGS:
Just some information: there is not real clash between arguments within Picard/GATK, but I would like to use the arguments from both to keep consistency with your tools.

droazen:
As a compromise, I'd vote to keep the existing short argument names, but add new long argument names (since there weren't long names previously).

vdauwera:
I agree with @droazen, let's keep the short names as they are (short) and add the long names.

magicDGS:
@droazen, @vdauwera, this does not solve the clash while using both splitting and RG arguments (my usage case in #2383). The sort names are the ones that are the same as in Picard `AddOrReplaceReadGroups`; the long ones where assumed to be the field names (`SAMPLE`, `READ_GROUP`, `LIBRARY_NAME`) and these ones are not clashing with anything...

droazen:
@magicDGS You could always rename the arguments within your tool, rather than changing the interface to our tool.

magicDGS:
It is what I will do if this change is not accepted, @droazen. It was just a request to keep the arguments in my toolkit equals to Picard/GATK to easier integration of pipelines with my custom tools and yours.

Anyway, thanks for taking in account my proposals. I added a new commit removing the change for the short arguments. Back to you!

droazen:
Latest version (just adding the long names and leaving the short names unchanged) is acceptable to us, so we'll merge this.

magicDGS:
Thnaks @droazen!

PR #2387, opened 2017-02-02
Fix EOF handling in prefetcher

jean-philippe-martin:
- add missing null check for reads a whole block past EOF
- correct handling for seek just past EOF + read
- fix reuse of caching buffer when near EOF
- add tests that reproduce(d) the problems

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
@@ -1,7 +1,6 @@
 package org.broadinstitute.hellbender.utils.nio;
 
 import com.google.common.base.Stopwatch;
-
 import java.io.Closeable;
 import java.io.IOException;
 import java.nio.ByteBuffer;
@@ -10,6 +9,7 @@ import java.nio.channels.NonWritableChannelException;
 import java.nio.channels.SeekableByteChannel;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.UnknownFormatConversionException;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -61,6 +61,8 @@ public final class SeekableByteChannelPrefetcher implements SeekableByteChannel 
     // number of times the user asks for data with a lower index than what we already have
     // (so they're not following the expected pattern of increasing indexes)
     public long nbGoingBack = 0;
+    // number of times the user asks for data past the end of the file
+    public long nbReadsPastEnd = 0;
 
     /**
      * WorkUnit holds a buffer and the instructions for what to put in it.
@@ -96,7 +98,7 @@ public final class SeekableByteChannelPrefetcher implements SeekableByteChannel 
             }
             chan.position(pos);
             // read until buffer is full, or EOF
-            while (chan.read(buf) > 0) {}
+            while (chan.read(buf) >= 0 && buf.hasRemaining()) {}
             return buf;
         }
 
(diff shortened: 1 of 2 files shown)


jean-philippe-martin:
note: this is a small PR.

droazen:
@jean-philippe-martin Yep, either @lbergelson or myself will review later today

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -227,17 +248,38 @@ public int read(ByteBuffer dst) throws IOException {
...
         int offset = (int)(position - (blockIndex * bufSize));
+        // src |==============---------------------|
+        //     :<---src.pos-->------src.limit----->:
+        // |---:--position->
+        //     :<--offset-->
+        //     ^ blockIndex*bufSize
         int availableToCopy = src.position() - offset;
+        if (availableToCopy < 0) {
+            // the caller is asking to read past the end of the file
+            nbReadsPastEnd++;
```
lbergelson: I'd extract this as a little function since it's used twice.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -227,17 +248,38 @@ public int read(ByteBuffer dst) throws IOException {
             src = fetch(position);
             msWaitingForData += waitingForData.elapsed(TimeUnit.MILLISECONDS);
         } catch (InterruptedException e) {
+            // Restore interrupted status
+            Thread.currentThread().interrupt();
```
lbergelson: Could you explain this?  I'm not very familiar with how interruptions work.  Wouldn't being interrupted have set the interrupt status?  Why do we want to be set as interrupted if we are going to ignore the interrupt and return 0?

jean-philippe-martin: The interrupt status was reset when the exception was thrown (it's there to indicate someone's trying to kill this thread). We cannot add a checked exception to the signature since this method is overridden, so instead we return as soon as possible and ask the caller to not block (calling interrupt() will cause most blocking functions to throw instead). This brings us closer to the goal, which is to unwind the thread's whole stack so it can be interrupted safely.

lbergelson:
@jean-philippe-martin I have a question about this because I don't understand why you're having calling `Thread.interrupt()`  The EOF bug fixes look good though and we're working on throwing together a demonstration of the whole GCS process for monday so I'm going to merge now and then be educated about interrupts later. 

👍  to fixing annoying edge cases.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
lbergelson: Oh, you responded while I was righting back.  Thanks.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
jean-philippe-martin: An even better way to factor this out is with a try-finally since we need betweenCallsToRead to be reset when the method returns, regardless of how that happens. I'll put that in.

jean-philippe-martin:
ah, well, ok, I could have pushed the updated version now though. Now that the branch's deleted I guess I'll have to redo that in a different PR.

lbergelson:
Sorry, I didn't know that you would be around this afternoon!

jean-philippe-martin:
No problem. Yes, I'm not sick anymore, yay! :-)

jean-philippe-martin:
I've rolled this change into #2391.

PR #2388, opened 2017-02-03
Upgrade to Barclay snapshot for tagged arguments.

cmnbroad:
Awaiting the Barclay snapshot with broadinstitute/barclay#33, which is ready, but I'll sanity check again in AM before I merge.

NOTE: All of the GATK tests have passed locally with this branch. However, I had to make one temporary change because SelectVariants has a feature that clashes with the collection list file feature in https://github.com/broadinstitute/barclay/pull/28. SelectVariants currently has two arguments that are defined as `List<File>`, that are each intended to take a list of file names, each of which in turn contains a list of sample names. SelectVariants manually loads all of the samples from all of the files in the list, and creates a list of unique sample names.

With the https://github.com/broadinstitute/barclay/pull/28, the CLP now loads the list directly, and hands SelectVariants a list of sample names rather than the list of file names, which breaks one test. I think both features are working as intended, but collide when a .list file is used.

I temporarily renamed the test file to not have a .list extension (so it won't trigger the CLP file loading), but we'll have to decide how to properly reconcile these two features.


cmnbroad:
Awaiting the Barclay snapshot with https://github.com/broadinstitute/barclay/pull/33, which is ready, but I'll review again in AM before I merge.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -187,6 +187,23 @@ public boolean isDependentArgumentAllowed(final Class<?> dependentClass) {
         return filters;
     }
 
+    /**
+     * Get the list of default plugins used for this instance of this descriptor. Used for help/doc generation.
+     * @return List of T
+     */
+    @Override
+    public List<Object> getDefaultInstances() { return new ArrayList<>(toolDefaultReadFilters.values()); }
```
magicDGS: Should not filter also the disabled ones? For documentation, there are no disabled ones and it could be reused to get the ones disables too for other purposes. I would appreciate a behaviour like that.

cmnbroad: @magicDGS I see the distinction you're making, but I think thats a different method. The intent of this method is that it returns the "list of defaults used by the tool", not the "list of defaults that are left after arguments are parsed". It should return the same list before and after parsing.

I know you're anxious to get your changes into the code base, but I think the next step will be for me to create a Barclay PR that gets the interface shaped approximately the way we want it based on all of the discussion we've had (we already have two proposed name changes to implement). I hope that helps.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -187,6 +187,23 @@ public boolean isDependentArgumentAllowed(final Class<?> dependentClass) {
...
+
+    /**
+     * Pass back the list of ReadFilter instances that were actually seen on the
+     * command line in the same order they were specified. This list does not
+     * include the tool defaults.
```
droazen: Method comment makes no sense -- this method returns a single `Class` object, not a `List`

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariantsIntegrationTest.java
```diff
@@ -328,7 +328,7 @@ public void testNoGTs() throws IOException {
     @Test
     public void testSelectFromMultiAllelic() throws IOException {
         final String testFile = getToolTestDataDir() + "multi-allelic.bi-allelicInGIH.vcf";
-        final String samplesFile = getToolTestDataDir() + "GIH.samples.list";
+        final String samplesFile = getToolTestDataDir() + "GIH.samples.files";
```
droazen: Fine for now -- create a ticket to discuss whether anything else needs to be done in `SelectVariants` in response to this new feature.

droazen:
One comment to address, then merge when ready 👍

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: I think that what you say is quite reasonable, @cmnbroad. I will wait til the decision is made.

PR #2391, opened 2017-02-03
Turn off fine-grained timing of prefetching by default

jean-philippe-martin:
These measurement are useful when tuning performance (or hunting down performance anomalies), but they have a measurable overhead (10% difference on a test with 1000 intervals, 5x the standard deviation on 10 runs). So turn them off by default.

Also refactor a few of those into a try-finally to avoid repetition and its associated risks on correctness.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
@@ -63,6 +63,9 @@ public final class SeekableByteChannelPrefetcher implements SeekableByteChannel 
     public long nbGoingBack = 0;
     // number of times the user asks for data past the end of the file
     public long nbReadsPastEnd = 0;
+    // timing statistics have an overhead, so only turn them on when debugging performance
+    // issues.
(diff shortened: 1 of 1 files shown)


droazen:
Tests failed on travis?

jean-philippe-martin:
I see `Timeout (30 minutes) reached. Terminating "./gradlew jacocoTestReport"`. It's not clear to me how my changes could have introduced a deadlock or similar problem. 

I ran the full test suite (`./gradlew test`) locally to take a look and it passes. Took 20min.

Running `SeekableByteChannelPrefetcherTest` by itself also passes, unsurprisingly.

droazen:
:+1: looks good -- merging

droazen:
Since you said in an email you had some failures with this version, but not the version before this one, I'll wait until you give an explicit ok on this branch @jean-philippe-martin before I merge.

droazen:
@jean-philippe-martin Is this good to go, or are you still seeing failures with this version?

jean-philippe-martin:
I tried this again carefully (12 runs) and only one failed, due to a "remote server unavailable" error. This is the same we see before this PR (we need to retry more aggressively) so I think we can merge safely.

PR #2393, opened 2017-02-03
hooking up tools so they take path wrappers for Features

lbergelson:
this duplicates work done for the Reads

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
@@ -16,6 +16,7 @@ import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
 import org.broadinstitute.hellbender.utils.io.IOUtils;
+import org.broadinstitute.hellbender.utils.nio.SeekableByteChannelPrefetcher;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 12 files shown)


src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
```diff
@@ -222,17 +242,22 @@ public static boolean isGenomicsDBPath(final String path) {
...
             final String absolutePath = IOUtils.getPath(featureInput.getFeaturePath()).toAbsolutePath().toUri().toString();
             // Instruct the reader factory to not require an index. We will require one ourselves as soon as
             // a query by interval is attempted.
-            return AbstractFeatureReader.getFeatureReader(absolutePath, codec, false);
+
+            if (BucketUtils.isCloudStorageUrl(absolutePath)){
+                return AbstractFeatureReader.getFeatureReader(absolutePath, null, codec, false, Function.identity(), Function.identity());
+            } else {
+                return AbstractFeatureReader.getFeatureReader(absolutePath, null, codec, false, cloudWrapper, cloudIndexWrapper);
+            }
```
droazen: Hmm, this seems backwards -- if it's a cloud storage url, you want to apply the wrappers, not the identity function. I'll fix via a push into this branch when I add the tests.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -37,7 +37,7 @@
     void initializeFeatures() {
         //Note: we override this method because we don't want to set feature manager to null if there are no FeatureInputs.
         //This is because we have at least 1 source of features (namely the driving dataset).
-        features = new FeatureManager(this, FEATURE_CACHE_LOOKAHEAD);
+        features = new FeatureManager(this, FEATURE_CACHE_LOOKAHEAD, cloudPrefetchBuffer, cloudIndexPrefetchBuffer);
```
droazen: Subclasses of `VariantWalkerBase` instantiate their own `FeatureDataSources` -- I think construction of the prefetching functions is going to have to be moved down there from `FeatureManager`.

droazen: Also need to take care of `MultiVariantDataSource`

droazen:
I've pushed up an initial patch to this branch to fix the issues I found. I'll push up tests later tonight after I grab some dinner :)

droazen:
Merging this now to have usable VCF NIO support in master -- continuous tests to prove that the wrapper is applied will be added in a separate PR, but my ad-hoc tests on the latest version of this branch suggest it's working fine.

PR #2396, opened 2017-02-07
Add integration tests for GCS support in VariantWalkers

droazen:


droazen:
@lbergelson please review

src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
```diff
@@ -62,10 +62,10 @@ public void setTestVerbosity(){
     public static final String publicTestDir = new File(gatkDirectory, publicTestDirRelative).getAbsolutePath() + "/";
     public static final String publicTestDirRoot = publicTestDir.replace(publicTestDirRelative, "");
 
-    public static final String GSC_GATK_TEST_RESOURCES = "gs://hellbender/test/resources/";
+    public static final String GCS_GATK_TEST_RESOURCES = "gs://hellbender/test/resources/";
```
jean-philippe-martin: Thanks for fixing the spelling. I'd suggest building those paths from command-line arguments, though, so we don't need to tell the whole world about this particular path.

jean-philippe-martin:
Looks good to me otherwise!

src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
lbergelson: Oh, yeah, we have an environment variable for this so people can redirect to their own copy of the test resources if they want.  HELLBENDER_TEST_INPUTS.  We should use that instead.

lbergelson: `Basetest.getGCPTestInputPath()` returns the appropriate test path.  (Maybe it should be changed to GCS?  I can't remember what gcp stood for?

droazen: Done

PR #2399, opened 2017-02-07
splitting tests on travis

lbergelson:
tests will now run as cloud, integration, and unit on travis
this reduces our wallclock time from 30ish -> 20ish minutes

cleaned up some wierdness in the way things were specified as well

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,9 @@ jdk:
 - oraclejdk8
 env:
   matrix:
-  - CLOUD=mandatory
-  - CLOUD=false
+  - TEST_TYPE=cloud
+  - TEST_TYPE=integration TEST_VERBOSITY=minimal
+  - TEST_TYPE=unit TEST_VERBOSITY=minimal
   global:
   #for genomics db
   - LD_LIBRARY_PATH=$TRAVIS_BUILD_DIR/dependencies/libcsv/.libs
@@ -36,8 +37,6 @@ env:
   - secure: "E0LWXgX3aWSE/DWHXXDx4vrAq4uX6vKg402wToaZ5otbHQ/UP0H7/FA5jomavAXoC46oMVHZcEltZ5OVhuJ0NW8yYxUCecJ1D/YvVQmnfFABcV/qLM+k4e2rYQOKVw/pejB2gG8XdTA+XE2WyTeENbmIkputS8f1ndKWCmZxuuk="
 matrix:
   fast_finish: true
-  allow_failures:
-    - env: CLOUD=todo
 before_cache:
   - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
 cache:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -61,26 +60,25 @@ before_install:
...
-- if [[ $TRAVIS_SECURE_ENV_VARS == false && $CLOUD == mandatory ]]; then
+- if [[ $TRAVIS_SECURE_ENV_VARS == false && $TEST_TYPE == cloud ]]; then
     echo "Can't run cloud tests without keys so don't bother building";
   else
     ./gradlew assemble;
     ./gradlew installDist;
   fi
 script:
-#basic sanity test to check that gatk-launch actually runs something
-- ./gatk-launch PrintReads -I src/test/resources/NA12878.chr17_69k_70k.dictFix.bam -O output.bam
```
droazen: Why remove this sanity check? No longer needed?

build.gradle
```diff
@@ -210,33 +210,25 @@ tasks.withType(Jar) {
...
             includeGroups 'cloud', 'bucket'
-        } else if (CLOUD == "todo") {
-            // run only the in-development cloud tests
-            includeGroups 'cloud_todo', 'bucket_todo'
-        } else if (CLOUD == "together") {
-            // run both local tests and mandatory cloud tests, together.
-            // This is good when e.g. you are done for the day and want to run tests on your machine overnight.
-            excludeGroups 'cloud_todo', 'bucket_todo'
+        } else if (TEST_TYPE == "integration"){
+            include "**/*IntegrationTest*"
```
droazen: Think you might want to remove the last * so that it's strictly a class name suffix?

build.gradle
```diff
@@ -210,33 +210,25 @@ tasks.withType(Jar) {
...
+            excludeGroups "cloud", "bucket"
+        } else if (TEST_TYPE == "unit") {
+            exclude "**/*IntegrationTest*"
+            excludeGroups "cloud", "bucket"
+        } else if (TEST_TYPE == "all"){
```
droazen: Document the different possible values for `TEST_TYPE` in a comment

build.gradle
```diff
@@ -248,8 +240,7 @@ test {
     minHeapSize = "1G"
     maxHeapSize = "4G"
 
-
-    if (CI == "true" && CLOUD == "false" ) {
+    if (TEST_VERBOSITY == "minimal") {
```
droazen: Can you explain this change? It's not intuitive to me

.travis.yml
```diff
@@ -61,26 +60,25 @@ before_install:
...
 #run tests
 #install git-lfs
-- if [[ $TRAVIS_SECURE_ENV_VARS == false && $CLOUD == mandatory ]]; then
+- if [[ $TRAVIS_SECURE_ENV_VARS == false && $TEST_TYPE == cloud ]]; then
     echo "Can't run cloud tests without keys so don't run tests";
   else
+    ./gatk-launch PrintReads -I src/test/resources/NA12878.chr17_69k_70k.dictFix.bam -O output.bam;
```
droazen: Preserve comment explaining why we do this step.

.travis.yml
```diff
@@ -61,26 +60,25 @@ before_install:
...
     bash scripts/install_git_lfs.sh;
     travis_wait 30 ./gradlew jacocoTestReport;
   fi
 after_success:
 - bash <(curl -s https://codecov.io/bash)
-- if [[ $TRAVIS_BRANCH == master && $CLOUD == false ]]; then ./gradlew uploadArchives; fi
+- if [[ $TRAVIS_BRANCH == master && $TEST_TYPE == integration ]]; then ./gradlew uploadArchives; fi
```
droazen: Why do we only do this for integration tests?

droazen:
Review complete -- back to @lbergelson

build.gradle
lbergelson: It seemed better to control it explicitly rather than being derived from some combination of CI and CLOUD.  I'm happy to change the names though.

droazen: What I meant was that it's not clear to me how `CI == "true" && CLOUD == "false"` in the previous version maps to `TEST_VERBOSITY == "minimal"` in the current version, so I was just curious.

.travis.yml
lbergelson: We want to upload exactly 1 artifact.  So we have to choose one of the 3 test suites to do so.  I chose integration vs unit pretty arbitrarily.

build.gradle
lbergelson: This toggle is between showing all test output and showing abbreviated test output.  Previously we showed abbreviated test output when running on travis and NOT running cloud tests, because there were only a few cloud tests so there wasn't any point in minimizing their output and it was more useful to be able to see them as they ran.  I figured it would make more sense to set the test output verbosity explicitly instead of having it depend on some combination of other confusing factors.  I'm not sure that TEST_VERBOSTY = minimal is the right UI choice though, any suggestion on names?  `ABBREVIATE_TEST_OUTPUT = true`?

build.gradle
lbergelson: done

build.gradle
lbergelson: done, and updated the readme

.travis.yml
lbergelson: Including the comment here broke travis linting.  I haven't figured out how to put comment lines in the middle of a bash statement in the travis script yet.  If you can figure it out, I'm happy to move it.  For now I'll just keep if above the block.

lbergelson:
@droazen I responded to your comments.  I've additionally some changes to the readme to include information about the test environment variables.  Let me know if there are horrible

build.gradle
droazen: Minimal is fine, I think

droazen:
:+1: looks good, merge when ready

Issue #2398, opened 2017-02-07
default read filters are excluded from the help output for --disableReadFilters


lbergelson:
Tool default ReadFilters are not included in the help output as possible values for either `--readFilter` or `--disableReadFilter`.

It makes some sense to not include default read filters in the list of read filters to enable, but we definitely need them in the list to disable.

magicDGS:
I think that https://github.com/broadinstitute/gatk/pull/2360 fix it.

magicDGS:
This is kind of a duplicate of #2358

cmnbroad:
Closed via #2401.

Issue #2377, opened 2017-01-31
Allow sane combinations of enable/disable plugin arguments, disallow insane combinations


droazen:
We should audit the plugin system (and add tests) to ensure that crazy combinations of enable/disable arguments (like `--readFilter` and `--disableReadFilter`) are disallowed, while useful combinations are permitted. Here's my attempt at an initial proposal:

`--enable X --disable X`: crazy, should be an error

`--enable X --enable X`: error

`--disable X --disable X`: error

`--enable X when X is already on by default in the tool`: warning, but should be allowed -- this is useful for pipeline authors to guarantee that a particular filter will be on, even if tool defaults change over time. We should make sure that the filter is only actually applied ONCE, however.

`--disable X when X is not enabled by default in the tool`: warning, but should be allowed -- this is useful for pipeline authors to guarantee that a particular filter will be off, even if tool defaults change over time.

droazen:
For @cmnbroad. The current behavior may or may not be consistent with the above -- even if it is, we should add tests to ensure it stays that way.

magicDGS:
This is the purpose of all my work here. The current behavior is the following (if already there, links point to the code):

- `--enable X --disable X` [throws `CommandLineException`](https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java#L208-L214) as expected. There is also a test covering this in [`ReadFilterPluginUnitTest`](https://github.com/broadinstitute/gatk/blob/35b1253510c5a63c90fd28aff4f2753cdf9e5071/src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java#L267-L274))
- `--enable X --enable X` does not throw anything yet. I did a PR last week for fix (test included): #2373 
- `--disable X --disable X` does not throw anything yet. I can add this behavior to the PR for enabling twice the same filter
[...]

PR #2385, opened 2017-02-02
Repeated disabled filters throws CommandLineException

magicDGS:
Fixes one of the cases described in #2377.

PR #2401, opened 2017-02-10
GATKReadFilterPluginDescriptor fixes

### LINKED ISSUE #2357: Disable default read filter with arguments blows up as a CommandLineException

### LINKED ISSUE #2358: Allowed values for --disableReadFilter should only include default filters

### LINKED ISSUE #2361: --disableAllReadFilters should disable only defaults

### LINKED ISSUE #2397: disabling a non-existent read filter gives a warning instead of an exception
`--disableReadFilter` should throw if a value that is not a valid ReadFilter is given

magicDGS:
- Disable default read filter with arguments does not longer blows up  but log a warning (fixes #2357)
- Improved help message for read filter (fixes #2358 and #2398)
- `--disableAllReadFilters` changed to `--disableToolDefaultReadFilters` and `isDisabledFilter()` honor this behavior (fixes #2361 and #2363)
- Add test for exercise the path of logging a warning when disabling a  not enabled filter (as proposed in #2377)
- Disabling a non-existent read filter throws a `CommandLineException`  (fixes #2397)

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
@@ -43,13 +43,17 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
             doc="Read filters to be disabled before analysis", optional=true, common = true)
     public final List<String> disableFilters = new ArrayList<>();
 
-    @Argument(fullName = "disableAllReadFilters",
-            shortName = "disableAllReadFilters",
-            doc = "Disable all read filters", common = true, optional = true)
-    public boolean disableAllReadFilters = false;
+    @Argument(fullName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
+            shortName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
+            doc = "Disable all tool default read filters", common = true, optional = true)
+    public boolean disableToolDefaultReadFilters = false;
 
     // Map of read filter (simple) class names to the corresponding discovered plugin instance
     private Map<String, ReadFilter> readFilters = new HashMap<>();
+    // readFilters map is overridden in validateArguments
+    // this set is for keeping the names for the help in alphabetical order
+    // and to use them for checking if an argument exists or not
+    private final Set<String> readFilterNames = new TreeSet<>();
 
     // List of default filters in the order they were specified by the tool
     private List<String> toolDefaultReadFilterNamesInOrder = new ArrayList<>();
@@ -141,6 +145,7 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
             readFilter = (ReadFilter) pluggableClass.newInstance();
             readFilters.put(simpleName, readFilter);
         }
+        readFilterNames.add(simpleName);
         return readFilter;
     }
 
@@ -182,6 +187,9 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
             ReadFilter rf = readFilters.get(s);
             if (rf != null) {
                 filters.add(rf);
+            } else if (disableToolDefaultReadFilters) {
+                // include in the read instances since the tool one is disabled, but it is provided by the user
+                filters.add(toolDefaultReadFilters.get(s));
             }
         });
         return filters;
@@ -207,9 +215,11 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
     // Return the allowable values for readFilterNames/disableReadFilter
     @Override
     public Set<String> getAllowedValuesForDescriptorArgument(final String longArgName) {
-        if (longArgName.equals(StandardArgumentDefinitions.READ_FILTER_LONG_NAME) ||
-                longArgName.equals(StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME)) {
-            return readFilters.keySet();
+        if (longArgName.equals(StandardArgumentDefinitions.READ_FILTER_LONG_NAME)) {
+            return readFilterNames;
+        }
+        if (longArgName.equals(StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME)) {
+            return toolDefaultReadFilters.keySet();
         }
         throw new IllegalArgumentException("Allowed values request for unrecognized string argument: " + longArgName);
     }
@@ -221,6 +231,15 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
      */
     @Override
     public void validateArguments() {
+        // throw if any filter does not exists
+        final Set<String> nonExistant = new HashSet<>(disableFilters);
+        nonExistant.removeAll(readFilterNames);
+        if (!nonExistant.isEmpty()) {
+            throw new CommandLineException.BadArgumentValue(
+                    String.format("The read filter(s) specified do not exists: ",
+                            Utils.join(", ", nonExistant)));
+        }
+
         // throw if any filter is duplicated
         final Set<String> duplicateUserFilterNames = Utils.getDuplicatedItems(userReadFilterNames);
         if (!duplicateUserFilterNames.isEmpty()) {
@@ -263,9 +282,11 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
 
         // throw if args were specified for a filter that was also disabled
         disableFilters.forEach(s -> {
-            if (requiredPredecessors.contains(s)) {
+            if (requiredPredecessors.contains(s) && !toolDefaultReadFilters.containsKey(s)) {
                 throw new CommandLineException(
                         String.format("Values were supplied for (%s) that is also disabled", s));
+            } else {
+                logger.warn("Values were supplied for (%s) that is also disabled", s);
             }
         });
 
@@ -293,11 +314,12 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
 
     /**
      * Determine if a particular ReadFilter was disabled on the command line.
-     * @param filterName name of the filter to query
-     * @return true if the name appears in the list of disabled filters
+     * @param filterName name of the filter to query.
+     * @return {@code true} if the name appears in the list of disabled filters, or is a tool default not provided by the user and all tool defaults are disabled; {@code false} otherwise.
      */
     public boolean isDisabledFilter(final String filterName) {
-        return disableFilters.contains(filterName);
+        return disableFilters.contains(filterName)
+                || (disableToolDefaultReadFilters && !userReadFilterNames.contains(filterName));
     }
 
     /**
@@ -348,13 +370,12 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
         Utils.nonNull(samHeader);
         Utils.nonNull(aggregateFunction);
 
-        if (disableAllReadFilters) {
-            return aggregateFunction.apply(null, samHeader);
-        }
-
         // start with the tool's default filters in the order they were specified, and remove any that were disabled
         // on the command line
-        final List<ReadFilter> finalFilters = toolDefaultReadFilterNamesInOrder
+        // if --disableToolDefaultReadFilters is specified, just initialize an empty list with initial capacity of user filters
+        final List<ReadFilter> finalFilters = (disableToolDefaultReadFilters)
+                ? new ArrayList<>(userReadFilterNames.size())
+                : toolDefaultReadFilterNamesInOrder
                 .stream()
                 .filter(s -> !isDisabledFilter(s))
                 .map(s -> toolDefaultReadFilters.get(s))
@@ -366,7 +387,7 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
             clFilters.forEach(f -> finalFilters.add(f));
         }
 
-        return aggregateFunction.apply(finalFilters, samHeader);
+        return (finalFilters.isEmpty()) ? aggregateFunction.apply(null, samHeader) : aggregateFunction.apply(finalFilters, samHeader);
     }
 
 }
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
@@ -49,16 +49,16 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
     public boolean disableToolDefaultReadFilters = false;
 
     // Map of read filter (simple) class names to the corresponding discovered plugin instance
-    private Map<String, ReadFilter> allDiscoveredReadFilters = new HashMap<>();
-
-    // List of default filters in the order they were specified by the tool
-    private List<String> toolDefaultReadFilterNamesInOrder = new ArrayList<>();
+    private final Map<String, ReadFilter> allDiscoveredReadFilters = new HashMap<>();
 
     // Map of read filter (simple) class names to the corresponding default plugin instance
-    private Map<String, ReadFilter> toolDefaultReadFilters = new HashMap<>();
+    // it is a LinkedHashMap because we want to remember the order in which these were provided, and also keep the
+    // actual instances in case they have any additional state provided by the tool
+    // when they were created
+    private final Map<String, ReadFilter> toolDefaultReadFilters = new LinkedHashMap<>();
 
     // Set of dependent args for which we've seen values (requires predecessor)
-    private Set<String> requiredPredecessors = new HashSet<>();
+    private final Set<String> requiredPredecessors = new HashSet<>();
 
     /**
      * @param toolDefaultFilters Default filters that may be supplied with arguments
@@ -76,10 +76,6 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
                 if (className.length() == 0) {
                     className = rfClass.getName();
                 }
-                // we want to remember the order in which these were provided, and also keep the
-                // actual instances in case they have any additional state provided by the tool
-                // when they were created
-                toolDefaultReadFilterNamesInOrder.add(className);
                 toolDefaultReadFilters.put(className, f);
             });
         }
@@ -380,10 +376,10 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
         final List<ReadFilter> finalFilters =
                 disableToolDefaultReadFilters ?
                         new ArrayList<>(userEnabledReadFilterNames.size()) :
-                        toolDefaultReadFilterNamesInOrder
+                        toolDefaultReadFilters.entrySet()
                                 .stream()
-                                .filter(s -> !isDisabledFilter(s))
-                                .map(s -> toolDefaultReadFilters.get(s))
+                                .filter(e -> !isDisabledFilter(e.getKey()))
+                                .map(e -> e.getValue())
                                 .collect(Collectors.toList());
 
         // now add in any additional filters enabled on the command line (preserving order)
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
@@ -36,12 +36,12 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
     @Argument(fullName = StandardArgumentDefinitions.READ_FILTER_LONG_NAME,
             shortName = StandardArgumentDefinitions.READ_FILTER_SHORT_NAME,
             doc="Read filters to be applied before analysis", optional=true, common = true)
-    public final List<String> userReadFilterNames = new ArrayList<>(); // preserve order
+    public final List<String> userEnabledReadFilterNames = new ArrayList<>(); // preserve order
 
     @Argument(fullName = StandardArgumentDefinitions.DISABLE_READ_FILTER_LONG_NAME,
             shortName = StandardArgumentDefinitions.DISABLE_READ_FILTER_SHORT_NAME,
             doc="Read filters to be disabled before analysis", optional=true, common = true)
-    public final List<String> disableFilters = new ArrayList<>();
+    public final List<String> userDisabledReadFilterNames = new ArrayList<>();
 
     @Argument(fullName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
             shortName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
@@ -49,11 +49,7 @@ public class GATKReadFilterPluginDescriptor extends CommandLinePluginDescriptor<
     public boolean disableToolDefaultReadFilters = false;
 
     // Map of read filter (simple) class names to the corresponding discovered plugin instance
-    private Map<String, ReadFilter> readFilters = new HashMap<>();
-    // readFilters map is overridden in validateArguments
-    // this set is for keeping the names for the help in alphabetical order
-    // and to use them for checking if an argument exists or not
-    private final Set<String> readFilterNames = new TreeSet<>();
+    private Map<String, ReadFilter> allDiscoveredReadFilters = new HashMap<>();
 
     // List of default filters in the order they were specified by the tool
     private List<String> toolDefaultReadFilterNamesInOrder = new ArrayList<>();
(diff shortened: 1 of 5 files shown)


magicDGS:
This is the set of fixes for the filter plugin, @cmnbroad. The first commit is the change introduced in #2385 for less verbose test output, so it should be drop once it is accepted.

magicDGS:
I rebased into the latest changes to include the changes in #2385. Now this should be prepard for review, @cmnbroad! Thanks for everything!

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -221,6 +231,15 @@ public boolean isDependentArgumentAllowed(final Class<?> dependentClass) {
...
     @Override
     public void validateArguments() {
+        // throw if any filter does not exists
+        final Set<String> nonExistant = new HashSet<>(disableFilters);
+        nonExistant.removeAll(readFilterNames);
+        if (!nonExistant.isEmpty()) {
+            throw new CommandLineException.BadArgumentValue(
+                    String.format("The read filter(s) specified do not exists: ",
+                            Utils.join(", ", nonExistant)));
+        }
```
cmnbroad: I'm assuming this is intended to be the fix for https://github.com/broadinstitute/gatk/issues/2397). I think it would be much cleaner and simpler to just integrate this check with the existing code below that already warns if a disabled filter is not enabled by the tool (I think you just need to add an if stmt there). Also, the error message (and comment, if you retain it) should specifically mention that its referring to a disabled filter.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -263,9 +282,11 @@ public void validateArguments() {
 
         // throw if args were specified for a filter that was also disabled
         disableFilters.forEach(s -> {
-            if (requiredPredecessors.contains(s)) {
+            if (requiredPredecessors.contains(s) && !toolDefaultReadFilters.containsKey(s)) {
                 throw new CommandLineException(
                         String.format("Values were supplied for (%s) that is also disabled", s));
```
cmnbroad: There doesn't appear to be a test that executes this throw (I'm not sure if it was hit before, but since the if stmt is changed we should have tests that hit both clauses).

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -293,11 +314,12 @@ public void validateArguments() {
...
      * Determine if a particular ReadFilter was disabled on the command line.
-     * @param filterName name of the filter to query
-     * @return true if the name appears in the list of disabled filters
+     * @param filterName name of the filter to query.
+     * @return {@code true} if the name appears in the list of disabled filters, or is a tool default not provided by the user and all tool defaults are disabled; {@code false} otherwise.
      */
     public boolean isDisabledFilter(final String filterName) {
-        return disableFilters.contains(filterName);
+        return disableFilters.contains(filterName)
+                || (disableToolDefaultReadFilters && !userReadFilterNames.contains(filterName));
```
cmnbroad: I'm not sure why this change is needed  ? If it is, then the first line of the javadoc needs to change to reflect what it actually does.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -348,13 +370,12 @@ public CountingReadFilter getMergedCountingReadFilter(final SAMFileHeader samHea
...
         // start with the tool's default filters in the order they were specified, and remove any that were disabled
         // on the command line
-        final List<ReadFilter> finalFilters = toolDefaultReadFilterNamesInOrder
+        // if --disableToolDefaultReadFilters is specified, just initialize an empty list with initial capacity of user filters
+        final List<ReadFilter> finalFilters = (disableToolDefaultReadFilters)
+                ? new ArrayList<>(userReadFilterNames.size())
+                : toolDefaultReadFilterNamesInOrder
                 .stream()
                 .filter(s -> !isDisabledFilter(s))
                 .map(s -> toolDefaultReadFilters.get(s))
```
cmnbroad: I think this was much clearer before (I realize it had to change to account for the new meaning of disableToolDefaultReadFilters). I guess its partly a style thing, but the extra parens and lack of indentation makes it hard to read; I think its much easier to parse like this:

```
        final List<ReadFilter> finalFilters = 
                disableToolDefaultReadFilters ?
                        new ArrayList<>(userReadFilterNames.size()) :
                        toolDefaultReadFilterNamesInOrder
                            .stream()
                            .filter(s -> !isDisabledFilter(s))
                            .map(s -> toolDefaultReadFilters.get(s))
```

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -43,13 +43,17 @@ private void readObject(java.io.ObjectInputStream in)
...
+            shortName = StandardArgumentDefinitions.DISABLE_TOOL_DEFAULT_READ_FILTERS,
+            doc = "Disable all tool default read filters", common = true, optional = true)
+    public boolean disableToolDefaultReadFilters = false;
 
     // Map of read filter (simple) class names to the corresponding discovered plugin instance
     private Map<String, ReadFilter> readFilters = new HashMap<>();
+    // readFilters map is overridden in validateArguments
+    // this set is for keeping the names for the help in alphabetical order
+    // and to use them for checking if an argument exists or not
+    private final Set<String> readFilterNames = new TreeSet<>();
```
magicDGS: Do you mean to remove this set or just use a `HashSet` instead? There are two reasons for introducing this field:

* Previously, the help was different if command line validation was before or after the plugin validation. As I point out in my comments, the `readFilters` map is overriden in `validateArguments` for the ones that will be used. But the parsing system also perform custom validation of the tools after the validation of plugin arguments, so if something fails after that is called, the help will be incorrect. In addition, I did not explore other validation paths in Barclay, but it could happen that other validation is included after the plugin one and this may fail again to print all the filters.
* I also used this variable to check for existence of the plugin in the names, because relying in the `readFilters` name seems like a bad idea if the state is not maintained.

I will wait for the next round of comments for modifying this, just to clarify if I should remove it or change it for a hash-based set.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -182,6 +187,9 @@ public boolean isDependentArgumentAllowed(final Class<?> dependentClass) {
             ReadFilter rf = readFilters.get(s);
             if (rf != null) {
                 filters.add(rf);
+            } else if (disableToolDefaultReadFilters) {
+                // include in the read instances since the tool one is disabled, but it is provided by the user
+                filters.add(toolDefaultReadFilters.get(s));
```
magicDGS: In this PR the meaning of disable "all" the filters changed to disable only the defaults. I tried to solve the following problems by this change:

* What happen if `--disableDefaultReadFilters --readFilter DEFAULT_FILTER_NAME` is specified? I think that the user provided filters must be applied independently of the default set.
* How can an user specify always the same parameters independently of any change in the tool default filters? I think that one of the points of the new behavior for this parameter is to provide a solution to that question. Disabling default filters one by one is not possible if they change over time, because that requires changes in the command line too; thus, this is a shortcut to disabling all the default read filters without knowing them, and provide always the same set of filters to apply. 
* How can an user reorganize default filters to apply? The order of filtering may be important in some case, for instance if one of them is computationally expensive with respect to the others. As an example: there is only one tool default filter, which is expensive and the user wants to pre-filter by a less demanding one. Previously it was not possible, but with this change the command line will provide the following: `--disableToolDefaultReadFilters --readFilter CUSTOM --readFilter DEFAULT_EXPENSIVE`

I think that this proposal is worthy from the user point of view, even if the new contract is not yet applied. Either that or this change should go to the `getMergedReadFilter()`to allow the behavior explained above.

It is true that there is no test case for this, I've just implemented one.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: True, it is much cleaner than this implementation. I've done this to point out all the non-existent in the error, as a good way to remove in the same batch all possible invalid filters. Anyway, I simplified as you suggested and changed the error message.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: There is a test case for throwing that is disabled because it fails after this change (I think that this is something related with Barclay, I added a `TODO` there for more information) and a new one covering the bug that I reported in #2357 (just logging). Thus, the only case where this could be exercised is when a non-default filter is disabled and arguments provided. Next commit will include that test.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: This is a fix for #2363. The new behavior of `--disableToolDefaultReadFilters` remove all the defaults filters only if the user one is not enable them again (if that change in `getAllInstances` is accepted). If the new behavior is removing all the default ones without allowing to re-enabled them, this line should be `(disableToolDefaultReadFilters && toolDefaultReadFilterNamesInOrder.contains(readFilter))`.

If someone is using this method to check if a required filter is disabled for a tool even if it is a default one, this should honor also the `--disableToolDefaultReadFilters`. Because the tool defaults were disabled in the command line if that parameter is provided, I think that it is still reflected in the first sentence. Anyway, I added in next commit a note saying that include disabled defaults if `--disableToolDefaultReadFilters` is provided in the command line.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
magicDGS: Reformat done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -366,7 +387,7 @@ public CountingReadFilter getMergedCountingReadFilter(final SAMFileHeader samHea
             clFilters.forEach(f -> finalFilters.add(f));
         }
 
-        return aggregateFunction.apply(finalFilters, samHeader);
+        return (finalFilters.isEmpty()) ? aggregateFunction.apply(null, samHeader) : aggregateFunction.apply(finalFilters, samHeader);
```
magicDGS: Good catch! It is something that I changed for whatever reason in one of my multiple PRs and branches fixing the plugin and kept here because I didn't realize...

Simplified in next commit.

magicDGS: In addition, I added to the javadoc that the aggregateFunction should handle the empty list in the same way as `null`.

magicDGS:
Addressed most of the issues in the first pass. I have some questions/explanations for few of your comments, so back to you @cmnbroad. And thanks for discussing these issues!

lbergelson:
@magicDGS The cloud tests are skipping now for your builds so they shouldn't all be marked as failures incorrectly.  I was hoping that was fixed now.

magicDGS:
Awesome @lbergelson! Now the tests are passing here, so I will probably rebase all of my PRs soon to avoid the annoying "red cross of death". Thanks!

magicDGS:
What is the progress on this @cmnbroad? Is this waiting for the new Barclay plugin interface?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: I meant the field should be removed altogether. readFilters is overwritten as you point out, but it doesn't need to be. So I've removed it in my set of changes.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: The use cases you mention are reasonable, but I think just making this change will result in duplicates in the merged filter. I think my updated changes handle this (I also added a test for the first bullet point).

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Right, I update these code paths a bit and updated the tests for this so that there are test that execute all the code paths. The test you mentioned is still disabled, but I added comments about why we need to work around the Barclay behavior.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Sounds good.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -80,6 +76,9 @@ public GATKReadFilterPluginDescriptor(final List<ReadFilter> toolDefaultFilters)
                 if (className.length() == 0) {
                     className = rfClass.getName();
                 }
+                // we want to remember the order in which these were provided, and also keep the
+                // actual instances in case they have any additiona any state provided by the tool
```
magicDGS: "any additiona any state" -> "any additional state"

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -127,70 +126,63 @@ public Object getInstance(final Class<?> pluggableClass) throws IllegalAccessExc
...
 
     /**
-     * Pass back the list of ReadFilter instances that were actually seen on the
-     * command line in the same order they were specified. This list does not
-     * include the tool defaults.
+     * Pass back the list of ReadFilter instances that were actually seen on the command line in the same
+     * order they were specified. Its possible for this to return a filter that was originally included
+     * in the list of tool defaults only in the case where the user also specifies it on the command line.
+     *
+     * NOTE: this method is somewhat misnamed in that it doesn't return ALL instances since it leaves out
```
magicDGS: We should add a note similar to this in `getDefaultInstances`

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -231,62 +227,77 @@ public boolean isDependentArgumentAllowed(final Class<?> dependentClass) {
...
+        // throw if a disabled filter doesn't exist; warn if it wasn't enabled by the tool in the first place
+        userDisabledReadFilterNames.forEach(s -> {
+            if (!allDiscoveredReadFilters.containsKey(s)) {
+                throw new CommandLineException.BadArgumentValue(String.format("Disabled filter (%s) does not exist", s));
             } else if (!toolDefaultReadFilters.containsKey(s)) {
                 logger.warn(String.format("Disabled filter (%s) is not enabled by this tool", s));
             }
         });
 
         // warn on redundant enabling of filters already enabled by default
```
magicDGS: This will read better like "warn if a filter is both default and enabled by the user"

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
```diff
@@ -380,17 +384,15 @@ public void testDisabledDefaultWithArgsProvided() {
...
+        GATKReadFilterPluginDescriptor rfDesc = new GATKReadFilterPluginDescriptor(
+                Collections.singletonList(new ReadLengthReadFilter(1, 10)));
+        CommandLineParser clp = new CommandLineArgumentParser(new Object(), Collections.singletonList(rfDesc));
         clp.parseArguments(nullMessageStream, new String[] {
                 "-disableReadFilter", ReadLengthReadFilter.class.getSimpleName()});
 
-        // get the command line read filters
-        clp.getPluginDescriptor(GATKReadFilterPluginDescriptor.class);
+        // Make sure ReadLengthReadFilter got disabled without an exception
+        Assert.assertTrue(rfDesc.userDisabledReadFilterNames.contains(ReadLengthReadFilter.class.getSimpleName()));
```
magicDGS: I will add also a test for `isDisabled`.

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
```diff
@@ -259,20 +259,23 @@ public void testDisableOneFilter() {
...
+        GATKReadFilterPluginDescriptor rfDesc = new GATKReadFilterPluginDescriptor(null);
         final CommandLineParser clp = new CommandLineArgumentParser(new Object(),
-                Collections.singletonList(new GATKReadFilterPluginDescriptor(null)));
-        clp.parseArguments(nullMessageStream, new String[] {
+                Collections.singletonList(rfDesc));
+        clp.parseArguments(nullMessageStream, new String[]{
                 "-disableReadFilter", ReadFilterLibrary.MAPPED.getClass().getSimpleName()
         });
+        // Make sure mapped filter got disabled with no exception
+        Assert.assertTrue(rfDesc.userDisabledReadFilterNames.contains(ReadFilterLibrary.MAPPED.getClass().getSimpleName()));
```
magicDGS: I will add also a test for isDisabled.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -80,6 +76,9 @@ public GATKReadFilterPluginDescriptor(final List<ReadFilter> toolDefaultFilters)
                 if (className.length() == 0) {
                     className = rfClass.getName();
                 }
+                // we want to remember the order in which these were provided, and also keep the
+                // actual instances in case they have any additiona any state provided by the tool
+                // when they were created
                 toolDefaultReadFilterNamesInOrder.add(className);
                 toolDefaultReadFilters.put(className, f);
```
magicDGS: What do you think about using a `LinkedHashMap` instead of a list and a `HashMap`, @cmnbroad?

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -36,24 +36,20 @@ private void readObject(java.io.ObjectInputStream in)
...
             doc = "Disable all tool default read filters", common = true, optional = true)
     public boolean disableToolDefaultReadFilters = false;
 
     // Map of read filter (simple) class names to the corresponding discovered plugin instance
-    private Map<String, ReadFilter> readFilters = new HashMap<>();
-    // readFilters map is overridden in validateArguments
-    // this set is for keeping the names for the help in alphabetical order
-    // and to use them for checking if an argument exists or not
-    private final Set<String> readFilterNames = new TreeSet<>();
+    private Map<String, ReadFilter> allDiscoveredReadFilters = new HashMap<>();
```
magicDGS: What do you think if we make all these fileds "final" to avoid overriding as previously, @cmnbroad?

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
```diff
@@ -406,6 +408,22 @@ public void testDisabledNonDefaultFilterWithArgsProvided() {
         clp.getPluginDescriptor(GATKReadFilterPluginDescriptor.class);
     }
 
+    @Test
+    public void testDisableDefaultsAndReenableWithDifferentArgs() {
```
magicDGS: I will add also a test of disable all the defaults and re-enable with an user filter before it.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
magicDGS: Done.

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
magicDGS: Done.

src/test/java/org/broadinstitute/hellbender/engine/filters/ReadFilterPluginUnitTest.java
magicDGS: Done.

magicDGS:
Back to you @cmnbroad. Your commit and some minor changes are included. There are still two questions:

* Why not using a `LinkedHashMap` instead of a `HashMap`/`ArrayList` for the default filters?
* Maybe it is better to make the fields final to do not override them in future changes or being aware of the change of state.

cmnbroad:
@magicDGS Both of those proposed changes (LinkedHashMap/final) sound right to me.

magicDGS:
Perfect, I include that changes in the latest commit. Back to you @cmnbroad.

magicDGS:
Thanks @cmnbroad!

cmnbroad:
@magicDGS Thanks for working through this!

PR #2403, opened 2017-02-10
publishing a master-snapshot update on every push to master

### LINKED ISSUE #1995: Snapshot for master
It will be nice to have in the Broad Institute artifactory a `master-SNAPSHOT` in addition to the commit-specific SNAPSHOT to use in developmental projects which use the latest commit.

lbergelson:
This would publish master-SNAPSHOT on every commit to master.  It would give people a single moving snapshot to build off of if they wanted to be on the bleeding edge.

Fixes #1995 

It's hard to test that it pushes only on commits to master without committing to master... but I think it should work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,11 @@ script:
 after_success:
 - bash <(curl -s https://codecov.io/bash)
 - if [[ $TRAVIS_BRANCH == master && $TEST_TYPE == integration ]]; then ./gradlew uploadArchives; fi
+# if this is actually a commit to master and not a pull request build into master, then publish master-snapshot
+- if [[ $TRAVIS_BRANCH == master && $TRAVIS_PULL_REQUEST == false && $TEST_TYPE == integration ]]; then 
+    git tag master;
+    ./gradlew uploadArchives; 
+  fi
 after_failure:
 - dmesg | tail -100
 after_script:


lbergelson:
@cmnbroad Could you take a look at this?

.travis.yml
```diff
@@ -81,6 +81,11 @@ script:
 after_success:
 - bash <(curl -s https://codecov.io/bash)
 - if [[ $TRAVIS_BRANCH == master && $TEST_TYPE == integration ]]; then ./gradlew uploadArchives; fi
+# if this is actually a commit to master and not a pull request build into master, then publish master-snapshot
+- if [[ $TRAVIS_BRANCH == master && $TRAVIS_PULL_REQUEST == false && $TEST_TYPE == integration ]]; then 
+    git tag master;
```
lbergelson: tagging here to set the version to master

cmnbroad: So this is just for some local side effect ? We don't need to push this tag ?

cmnbroad: @lbergelson Is it the case that line 83 already does this - it's just that it uses the commit-derived name ? So will this double upload for every merged PR ? Trying to understand the conditions.

lbergelson: The version is based off the tag name, so it's a cheap way to set the version to master-snapshot.  We definitely DO NOT want to push a tag named master, that would cause much havoc I think interacting in confusing ways with the branch named master.

The conditions are a bit confusing and set up as follows:

Travis builds every commit, if it's building a commit that isn't in a pull request then `TRAVIS_BRANCH` is set to the branch that commit comes from and `TRAVIS_PULL_REQUEST=false`.  Travis does a second build for every commit in a pull request, which is that commit rebased onto master.  In that case, `TRAVIS_BRANCH=master` and `TRAVIS_PULL_REQUEST=true`

We want to upload a snapshot of every pull request commit so people can test against those, as well as a permanent snapshot of every commit to master, that's what the top case does, upload a uniquely numbered snapshot for each pull request commit and each commit to master.  The new case uploads a floating snapshot that's called `master-SNAPSHOT` only on proper commits to master, not on pull requests.  So in the case of a commit to master we will upload twice in a row, one a commit with the githash in it's name and one moving snapshot called master.

cmnbroad: I see.

PR #2404, opened 2017-02-10
Allow tests to fail gracefully if HELLBENDER_TEST_INPUTS isn't set.

cmnbroad:
The entire test suite aborts if HELLBENDER_TEST_INPUTS isn't set because an exception is thrown when loading the VariantWalkerGCSSupportIntegrationTest class.

With this change, the tests will still fail, but the rest of the test suite will run. Depending on what the intent for these tests is, another possibility would be to add a dependsOn method with a hard dependency so the tests would be skipped in the case of no env variable.

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerGCSSupportIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerGCSSupportIntegrationTest.java
@@ -17,8 +17,8 @@ import java.util.List;
  */
 public class VariantWalkerGCSSupportIntegrationTest extends CommandLineProgramTest {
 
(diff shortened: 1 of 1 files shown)


lbergelson:
ack, there's an unexpected consequence.  👍  When tests pass.

droazen:
How did travis not catch this?

lbergelson:
those values are set on travis

droazen:
Also, the cloud tests should be skipped by default (ie., in the case of no environment variables set).

droazen:
@lbergelson They are set even when running the non-cloud part of the test matrix?

lbergelson:
Yeah, they're always set.  We could change that though.

PR #2407, opened 2017-02-14
Tests to prove that we can access and query bams in GCS from ReadWalkers

### LINKED ISSUE #2406: Implement good tests for GCS NIO support in ReadWalkers

droazen:
Resolves #2406


droazen:
@jean-philippe-martin or @cmnbroad please review

src/test/java/org/broadinstitute/hellbender/engine/ReadWalkerGCSSupportIntegrationTest.java
```diff
@@ -0,0 +1,71 @@
...
+                { TEST_BAM_ON_GCS, Collections.singletonList(singleInterval), false, EXPECTED_SINGLE_INTERVAL_RESULTS },
+                { TEST_BAM_ON_GCS, multipleIntervals, false, EXPECTED_MULTIPLE_INTERVALS_RESULTS },
+                { TEST_BAM_ON_GCS, multipleIntervals, true, EXPECTED_MULTIPLE_INTERVALS_WITH_UNMAPPED_RESULTS },
+                { TEST_BAM_ON_GCS, null, true, EXPECTED_UNMAPPED_ONLY_RESULTS }
+        };
+    }
+
+    @Test(dataProvider = "GCSTestCases", groups = {"bucket"})
+    public void testReadBAMOnGCS( final String bam, final List<SimpleInterval> intervals, final boolean includeUnmapped, final String expectedOutput ) throws IOException {
+        final StringBuilder intervalArgBuilder = new StringBuilder("");
```
cmnbroad: Minor nit - at first glance the "" looks weird since its unnecessary.

src/test/java/org/broadinstitute/hellbender/engine/ReadWalkerGCSSupportIntegrationTest.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Tests to prove that we can access and query inputs on Google Cloud Storage (GCS) in ReadWalkers.
+ */
+public class ReadWalkerGCSSupportIntegrationTest extends CommandLineProgramTest {
+
+    private static final String TEST_BAM_ON_GCS = "org/broadinstitute/hellbender/engine/CEUTrio.HiSeq.WGS.b37.NA12878.20.21.10000000-10000020.with.unmapped.bam";
```
jean-philippe-martin: We may consider using a larger file so we can exercise NIO's prefetcher and caching a bit more. We don't need to read more bytes, just a few spread across a large file (so the prefetcher actually has to fetch more than one 30MB block).

droazen: I wanted these basic tests to be fast, so I used a small file containing representative snippets of reads. But we can expand these tests over time -- eg., if you want to add a separate test case to `ReadWalkerGCSSupportIntegrationTest` that uses a larger file to fully exercise the prefetcher, be my guest!

droazen:
@cmnbroad I suspect CRAM will not work, and will require further changes to htsjdk. I'll capture these in a ticket.

src/test/java/org/broadinstitute/hellbender/engine/ReadWalkerGCSSupportIntegrationTest.java
jean-philippe-martin: Fair enough! We should be able to exercise the prefetcher without taking too long, though, by reading only a small part of the large file and keeping the prefetch size small.

PR #2411, opened 2017-02-17
Rev htsjdk to version 2.9.0-1-g55bf01b-20170217.164935-1

droazen:


droazen:
@cmnbroad please review

droazen:
@tomwhite This includes your patches to `BAMFileReader`, among other things.

build.gradle
```diff
@@ -55,7 +55,7 @@ repositories {
 
 }
 
-final htsjdkVersion = System.getProperty('htsjdk.version','2.8.1-22-gb05d1f2-SNAPSHOT')
+final htsjdkVersion = System.getProperty('htsjdk.version','2.9.0-1-g55bf01b-20170217.164935-1')
```
cmnbroad: Is there some difference between using this name and using 2.9.0-1-g55bf01b-SNAPSHOT ? Otherwise this looks fine - just wondering if there is a reason for this.

droazen: There can be more than one snapshot for a given commit hash -- by including the timestamp, we can select a specific snapshot. Not strictly necessary, but I think it's nice having the snapshot date right in the version number, for informational purposes.

Issue #2413, opened 2017-02-21
Total ploidy does not have to equal to the number of priors


ronlevine:
Port https://github.com/broadinstitute/gsa-unstable/pull/1557

ronlevine:
Closed since #2416 is merged.

PR #2416, opened 2017-02-21
Total ploidy does not have to equal to the number of priors

ronlevine:
Implement https://github.com/broadinstitute/gatk/issues/2413.


ronlevine:
@davidbenjamin Please review.

PR #2417, opened 2017-02-21
Improve error message for no-access and disabled-account cases

jean-philippe-martin:
*before*: GATK crashes with a stack trace. The stack trace contains useful
info, but:
 * it's hard to read
 * it doesn't include the name of the file we cannot access

*now*, a better message that addresses both issues:
> A USER ERROR has occurred: Couldn't read file gs://(...). Error was:
> 401: Anonymous users does not have storage.objects.get access to object (...).

Additional information (including a stack trace) is displayed if the
user specifies `--verbosity=DEBUG`

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
@@ -577,16 +577,6 @@ public final class IOUtils {
             }
         } catch (com.google.cloud.storage.StorageException cloudBoom) {
             // probably a permissions problem, or perhaps a disabled account.
-            logger.debug(cloudBoom.toString());
-            logger.debug("code: " + cloudBoom.getCode());
-            logger.debug("location: " + cloudBoom.getLocation());
-            logger.debug("reason: " + cloudBoom.getReason());
-            logger.debug("message: " + cloudBoom.getMessage());
-            String msg = "";
-            for (StackTraceElement x : cloudBoom.getStackTrace()) {
-                msg += x.toString() + "\n";
-            }
-            logger.debug("stack trace:\n" + msg);
             // Looks like this for a disabled bucket error:
             //   A USER ERROR has occurred: Couldn't read file gs://foo/bar. Error was:
             //   403: The account for bucket "foo" has been disabled.
@@ -594,6 +584,8 @@ public final class IOUtils {
             // (use `gcloud auth application-default revoke` to forget the default credentials)
             //   A USER ERROR has occurred: Couldn't read file gs://(...). Error was:
             //   401: Anonymous users does not have storage.objects.get access to object (...).
+            // The user can see the underlying exception by passing
+            // -DGATK_STACKTRACE_ON_USER_EXCEPTION=true
             throw new UserException.CouldNotReadInputFile(path, cloudBoom.getCode() + ": " + cloudBoom.getMessage(), cloudBoom);
         }
     }
--- a/src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.hellbender.utils.io;
 
+import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.contrib.nio.CloudStorageFileSystem;
 import htsjdk.samtools.BamFileIoUtils;
 import htsjdk.samtools.cram.build.CramIO;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -564,14 +565,36 @@ public static void assertFileIsReadable(final File file) {
...
+            }
+        } catch (com.google.cloud.storage.StorageException cloudBoom) {
+            // probably a permissions problem, or perhaps a disabled account.
+            logger.debug(cloudBoom.toString());
+            logger.debug("code: " + cloudBoom.getCode());
+            logger.debug("location: " + cloudBoom.getLocation());
+            logger.debug("reason: " + cloudBoom.getReason());
+            logger.debug("message: " + cloudBoom.getMessage());
+            String msg = "";
+            for (StackTraceElement x : cloudBoom.getStackTrace()) {
```
lbergelson: We don't need to manually print the stack trace here.  There's an option to print stack traces for user exceptions already.  you can set the environment variable -DGATK_STACKTRACE_ON_USER_EXCEPTION=true

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -564,14 +565,36 @@ public static void assertFileIsReadable(final File file) {
...
+            }
+            if ( ! Files.isReadable(path) ) {
+                throw new UserException.CouldNotReadInputFile(path, "It is not readable, check the file permissions");
+            }
+        } catch (com.google.cloud.storage.StorageException cloudBoom) {
+            // probably a permissions problem, or perhaps a disabled account.
+            logger.debug(cloudBoom.toString());
+            logger.debug("code: " + cloudBoom.getCode());
+            logger.debug("location: " + cloudBoom.getLocation());
+            logger.debug("reason: " + cloudBoom.getReason());
```
lbergelson: What's the difference between reason, and message?  What's the difference between location and path?

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -564,14 +565,36 @@ public static void assertFileIsReadable(final File file) {
...
 
-        if ( ! Files.exists(path) ) {
-            throw new UserException.CouldNotReadInputFile(path, "It doesn't exist.");
-        }
-        if ( ! Files.isRegularFile(path) ) {
-            throw new UserException.CouldNotReadInputFile(path, "It isn't a regular file");
-        }
-        if ( ! Files.isReadable(path) ) {
-            throw new UserException.CouldNotReadInputFile(path, "It is not readable, check the file permissions");
+        try {
```
lbergelson: @jean-philippe-martin It's a good idea to translate this into a clearer exception message.  I think most of the logging is unnecessary because you can already get the stack trace, which I'm assuming will include the code, location, reason, and message in it?  

Do you think we need a top-level catch for StorageException in Main in order to translate them if their thrown in other places?  Does the storage exception include the information about the path that was being accessed?  I assume that's very similar information to the location that it includes. ?

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
jean-philippe-martin: Oh thank you, I didn't know about that option! I have removed the whole printing code then since it's unnecessary.

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
jean-philippe-martin: [Here is the documentation](https://cloud.google.com/storage/docs/json_api/v1/status-codes). 
`reason` is a short code like "invalidParameter" and `message` is meant for human consumption. `location` is the name of the parameter that caused the error. `path` of course is the argument to the method, the file we're trying to make sure is readable.

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
jean-philippe-martin: I don't think we need anything special at the top beyond what's already there. Those exceptions do not includes the path, no, except possibly in the message itself (and we already print that).

jean-philippe-martin:
I already addressed the reviewer's feedback, so this should be assigned to @droazen or @lbergelson.

droazen:
Right you are -- assigning to @lbergelson for another look.

lbergelson:
@jean-philippe-martin One more comment and a whine about StorageExceptions not including all the information I want them to have.  👍 After the comment's addressed.

jean-philippe-martin:
@lbergelson I *think* I hear you insist that we print out all the fields if the StorageException bubbles up to the top, and say I can merge after that. I've added code to print those fields.

lbergelson:
@jean-philippe-martin Sorry, That baby was not very asleep last night so I may be slightly less coherent than usual...   

I see how you heard that, but it wasn't what I meant.  What I mean was that we should eventually move the information about how to set -DSTACK_TRACE_ON_USEREXCEPTION into the top level UserException message to make it discoverable, and remove it from the comment it's in now.  Lets do that in a different PR though since it's sort of orthogonal and the best thing to do might be to integrate it as a regular commandline option instead of an environment variable.

Separately from that, I was wondering if we should catch StorageExceptions at the top level and handle them specially.  If we're going to do that I think we could just add them to the UserException catch block and have them be treated the same way, no need for special handling.

jean-philippe-martin:
Ah, ok. I've filed a new issue for the documentation of -DSTACK_TRACE_ON_USEREXCEPTION (#2245). 

StorageException is different from UserException in that it doesn't have the user-relevant context. Like  IOExceptions, we normally catch StorageException and transform them into user-friendly UserExceptions (as this PR does). Because of this lack of context, I don't think there's much "special" we can do about them (printing them out fully, as we do now, is the best I can think of). If you have an idea though I'm open to suggestions.

jean-philippe-martin:
Trying out the shiny new "Squash and merge" button.

PR #2419, opened 2017-02-22
Ensure reads are coordinate sorted before using the overlaps partitioner join strategy

tomwhite:


src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
```diff
@@ -111,7 +114,18 @@ protected void runTool(final JavaSparkContext ctx) {
...
+        if (joinStrategy.equals(JoinStrategy.OVERLAPS_PARTITIONER)) {
+            // the overlaps partitioner requires that reads are coordinate-sorted
+            final SAMFileHeader readsHeader = getHeaderForReads().clone();
+            readsHeader.setSortOrder(SAMFileHeader.SortOrder.coordinate);
+            ReadCoordinateComparator comparator = new ReadCoordinateComparator(readsHeader);
+            markedFilteredReadsForBQSR = markedFilteredReadsForBQSR
+                    .mapToPair(read -> new Tuple2<>(read, null))
+                    .sortByKey(comparator, true, numReducers)
+                    .keys();
+        }
```
tomwhite: Note that the reads are unsorted after MD, so we need to make them coordinate-sorted for the overlaps partitioner. This incurs another shuffle, so we'll see how this performs (vs broadcast for example), when we benchmark longer pipelines.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
```diff
@@ -111,7 +114,18 @@ protected void runTool(final JavaSparkContext ctx) {
...
-        final JavaRDD<GATKRead> markedFilteredReadsForBQSR = markedReads.filter(read -> bqsrReadFilter.test(read));
+        JavaRDD<GATKRead> markedFilteredReadsForBQSR = markedReads.filter(read -> bqsrReadFilter.test(read));
+
+        if (joinStrategy.equals(JoinStrategy.OVERLAPS_PARTITIONER)) {
+            // the overlaps partitioner requires that reads are coordinate-sorted
+            final SAMFileHeader readsHeader = getHeaderForReads().clone();
+            readsHeader.setSortOrder(SAMFileHeader.SortOrder.coordinate);
+            ReadCoordinateComparator comparator = new ReadCoordinateComparator(readsHeader);
+            markedFilteredReadsForBQSR = markedFilteredReadsForBQSR
+                    .mapToPair(read -> new Tuple2<>(read, null))
```
droazen: Do we have to create a dummy pair like this just to sort?

droazen: Also, can you extract a utility method in `SparkUtils` that sorts an RDD of reads by coordinate? This seems like something that would be generally useful.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
```diff
@@ -111,7 +114,18 @@ protected void runTool(final JavaSparkContext ctx) {
...
+                    .sortByKey(comparator, true, numReducers)
+                    .keys();
+        }
 
         VariantsSparkSource variantsSparkSource = new VariantsSparkSource(ctx);
         JavaRDD<GATKVariant> bqsrKnownVariants = variantsSparkSource.getParallelVariants(baseRecalibrationKnownVariants, getIntervals());
```
droazen: Will the final call to `writeReads()` do yet another shuffle to sort the already-sorted output, when we're running with `shardedOutput = false` and using the overlaps partitioner? If yes, can we avoid this?

src/main/java/org/broadinstitute/hellbender/tools/spark/BaseRecalibratorSpark.java
```diff
@@ -84,6 +85,10 @@ protected void runTool( JavaSparkContext ctx ) {
             throw new UserException.Require2BitReferenceForBroadcast();
         }
 
+        if (joinStrategy == JoinStrategy.OVERLAPS_PARTITIONER && getHeaderForReads().getSortOrder() != SAMFileHeader.SortOrder.coordinate) {
+            throw new UserException.BadInput("Reads must be coordinate sorted when using the overlaps partitioner join strategy.");
+        }
```
droazen: Do we need an equivalent check in `BQSRPipelineSpark`? Should this check be pushed down into `AddContextDataToReadSpark` so that you don't have to duplicate it in both places?

droazen:
First-pass review complete, back to @tomwhite

droazen:
@tomwhite Would you have time to address the review comments on this branch? It would be good to get this merged.

droazen:
@tomwhite Pinging you on this one

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
tomwhite: Yes, we have to extract a dummy pair to do this. The RDD#sortBy method takes a function to extract a key by, but it only sorts by natural ordering, and we need a custom comparator.

tomwhite: I've extracted the utility methods to `SparkUtils`.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
tomwhite: Yes, it will. To avoid it we'll need to create our own RDD, e.g. SortedRDD, which remembers the "sortedness". I think this is a bigger refactoring that deserves its own issue.

src/main/java/org/broadinstitute/hellbender/tools/spark/BaseRecalibratorSpark.java
tomwhite: BQSR does not shuffle the reads so they will have the same sort order as the input, so I don't think a check is needed.

tomwhite:
@droazen back to you.

src/main/java/org/broadinstitute/hellbender/tools/spark/BaseRecalibratorSpark.java
tomwhite: On second thoughts, we do need an equivalent check.

tomwhite:
@droazen another one that's ready

PR #2427, opened 2017-02-27
fix bug in soft-clip evaluation, and relax quality test a little

tedsharpe:
This is a very small commit.  Didn't want to forget about this bug I discovered.

1) found a bug (line 62 of ReadClassifier -- we used to always start at 0 which isn't at all good for reversed reads)
2) want to relax the strictness of the quality check a bit -- manual review showed a substantial number of plausible split read candidates that were being discarded


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadClassifier.java
```diff
@@ -55,9 +56,10 @@ private static boolean hasInitialSoftClip( final List<CigarElement> cigarElement
         if ( firstEle.getOperator() == CigarOperator.HARD_CLIP && itr.hasNext() ) {
             firstEle = itr.next();
         }
+        final int clipStart = firstEle.getLength() - MIN_SOFT_CLIP_LEN;
         return firstEle.getOperator() == CigarOperator.SOFT_CLIP &&
-                firstEle.getLength() >= MIN_SOFT_CLIP_LEN &&
-                isHighQualityRegion(read.getBaseQualities(), 0);
+                clipStart >= 0 &&
+                isHighQualityRegion(read.getBaseQualities(), clipStart);
```
cwhelan: Just to rephrase to verify that I understand what is happening: now we only check for low quality bases in the first "MIN_SOFT_CLIP_LEN" clipped bases as opposed to the whole soft-clipped region, and now we allow up to 3 low-quality bases in that region instead of zero. Is that correct? Is that the full substance of this change?

tedsharpe: Almost.  We've always checked just MIN_SOFT_CLIP_LEN base qualities.  It's just the position of this window that changes for left soft-clips.  I had intended to check the quals adjacent to the clip -- as I do for the right soft-clips -- but was instead checking from the beginning of the read for left soft-clips.
<---------soft clip--------><--------aligned part of read------>
<----check these--->  (old, wrong way)
            <---check these---> (new, right way)
This was particularly bad for negative strand reads, because I was checking where you'd expect bad bases to lurk (at the beginning of the RC'd read).
Also, I now allow 3 low quality bases in the checked window rather than 0, as you observed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadClassifier.java
```diff
@@ -74,8 +76,9 @@ private static boolean hasFinalSoftClip( final List<CigarElement> cigarElements,
     }
 
     private static boolean isHighQualityRegion( final byte[] quals, int idx ) {
+        int lowQuals = 0;
         for ( final int end = idx+MIN_SOFT_CLIP_LEN; idx != end; ++idx ) {
-            if ( quals[idx] < MIN_QUALITY ) return false;
+            if ( quals[idx] < MIN_QUALITY && ++lowQuals > MAX_LOW_QUALITY_SCORES ) return false;
```
cwhelan: Stylistic point, but I really don't like having a ++ as a prefix operator nested within a clause of a condition.. would you mind blowing this out? eg

'''
if ( quals[idx] < MIN_QUALITY ) {
   lowQuals = lowQuals + 1;
   if ( lowQuals > MAX_LOW_QUALITY_SCORES ) {
           return false;
   }
}
'''

I know it's more lines of code but I find it much easier to read..

PR #2433, opened 2017-03-02
Add --maximumMappingQuality to MappingQualityReadFilter

### LINKED ISSUE #2432: Enable PrintReads to exclude reads above a certain MQ threshold
@chandrans commented on [Wed Sep 14 2016](https://github.com/broadinstitute/gsa-unstable/issues/1476)

### Tool(s) involved

PrintReads
### Description

Right now, we have a flag that outputs reads with a MQ greater than the set threshold. Users may want to only output reads that have a MQ below a certain threshold. There should be a flag (something like -maximumMappingQuality) that only emits reads that have less than the set Mapping Quality. 

For example, a user needs this to compare low mapQ reads from different samples, blast search and/or perform de novo assembly.

---

This Issue was generated from your [forums](http://gatkforums.broadinstitute.org/gatk/discussion/comment/32943#Comment_32943) 

---

@vdauwera commented on [Mon Nov 14 2016](https://github.com/broadinstitute/gsa-unstable/issues/1476#issuecomment-260498511)

Might end up being a feature request for bidirectional read filtering functionality in GATK4. 

---

@chandrans commented on [Tue Feb 28 2017](https://github.com/broadinstitute/gsa-unstable/issues/1476#issuecomment-283141618)

Okay. Should I keep this open or close it/move to GATK4?

---

@vdauwera commented on [Tue Feb 28 2017](https://github.com/broadinstitute/gsa-unstable/issues/1476#issuecomment-283150007)

Yeah move to 4 please

magicDGS:
Fixes #2432

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/filters/MappingQualityReadFilter.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/filters/MappingQualityReadFilter.java
@@ -8,7 +8,10 @@ import org.broadinstitute.hellbender.utils.read.GATKRead;
 import java.io.Serializable;
 
 /**
- * Keep reads with mapping qualities above a specified threshold.
+ * Keep reads with mapping qualities within a specified range.
+ *
+ * Note: this filter does not handle specially the unavailable mapping quality ({@link org.broadinstitute.hellbender.utils.QualityUtils#MAPPING_QUALITY_UNAVAILABLE}).
+ * Use {@link org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary.MappingQualityAvailableReadFilter} to explicitly filter out reads with unavailable quality.
  */
 @DocumentedFeature(groupName= HelpConstants.DOC_CAT_READFILTERS, groupSummary=HelpConstants.DOC_CAT_READFILTERS_SUMMARY)
 public final class MappingQualityReadFilter extends ReadFilter implements Serializable {
@@ -17,7 +20,7 @@ public final class MappingQualityReadFilter extends ReadFilter implements Serial
     @Argument(fullName="minimumMappingQuality", shortName="minimumMappingQuality", doc = "Minimum mapping quality to keep (inclusive)", optional=true)
     public int minMappingQualityScore = 10;
 
-    @Argument(fullName="maximumMappingQuality", shortName="maximumMappingQuality", doc = "Maximum mapping quality to keep (exclusive)", optional=true)
+    @Argument(fullName="maximumMappingQuality", shortName="maximumMappingQuality", doc = "Maximum mapping quality to keep (inclusive)", optional=true)
     public Integer maxMappingQualityScore = null;
 
     // Command line parser requires a no-arg constructor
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/engine/filters/MappingQualityReadFilter.java
```diff
@@ -14,18 +14,28 @@
 public final class MappingQualityReadFilter extends ReadFilter implements Serializable {
     private static final long serialVersionUID = 1L;
 
-    @Argument(fullName="minimumMappingQuality", shortName="minimumMappingQuality", optional=true)
+    @Argument(fullName="minimumMappingQuality", shortName="minimumMappingQuality", doc = "Minimum mapping quality to keep (inclusive)", optional=true)
     public int minMappingQualityScore = 10;
 
+    @Argument(fullName="maximumMappingQuality", shortName="maximumMappingQuality", doc = "Maximum mapping quality to keep (exclusive)", optional=true)
```
droazen: Looks like the maximum is actually *inclusive* (`<=`), not exclusive (`<`) as the doc says. Can you update the doc here? I think we should keep the comparison inclusive, as an exclusive comparison would be too confusing.

src/main/java/org/broadinstitute/hellbender/engine/filters/MappingQualityReadFilter.java
```diff
@@ -14,18 +14,28 @@
 public final class MappingQualityReadFilter extends ReadFilter implements Serializable {
```
droazen: Need to update the class-level docs, which currently say "Keep reads with mapping qualities above a specified threshold."

src/main/java/org/broadinstitute/hellbender/engine/filters/MappingQualityReadFilter.java
```diff
@@ -14,18 +14,28 @@
...
+        this.minMappingQualityScore = minMappingQualityScore;
+        this.maxMappingQualityScore = maxMappingQualityScore;
+    }
+
     @Override
     public boolean test( final GATKRead read ) {
-        return read.getMappingQuality() >= minMappingQualityScore;
+        final int mq = read.getMappingQuality();
+        return  mq >= minMappingQualityScore
+                && (maxMappingQualityScore == null || mq <= maxMappingQualityScore);
```
droazen: How should we deal with the special meaning of mapping quality 255 here?

src/main/java/org/broadinstitute/hellbender/engine/filters/MappingQualityReadFilter.java
magicDGS: Sorry, I think that I changed the implementation because it was confusing and I forgot to update the text.

droazen:
Review complete, back to @magicDGS. Just a few minor doc changes, and then this should be good to go.

src/main/java/org/broadinstitute/hellbender/engine/filters/MappingQualityReadFilter.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/engine/filters/MappingQualityReadFilter.java
magicDGS: Done.

magicDGS:
Back to you @droazen. Please, review the javadoc changes (my English can be clunky sometimes).

PR #2435, opened 2017-03-06
Scripts running the whole sv-pipeline as exists now from birth to termination

SHuang-Broad:
@cwhelan @tedsharpe @vruano @mwalker174 , I've organized the scripts for running the whole sv pipeline as it exists right now. There used to be a PR (if I recall correctly) but since that's outdated, why not an up-to-date one.

Here's how to run it

1. To create the cluster

```
./create_cluster.sh broad-dsde-methods sv-methods-1 broad-dsde-methods/sv
```
where the 1st argument is the project name, 2nd argument is the cluster name, and the 3rd name is the place where input data lives.

2. To run the whole pipeline

```
./svCall.sh /Users/shuang/GATK/gatk sv-methods-1 /user/shuang/NA12878_PCR-_30X
```
where the 1st argument is the location of my GATK directory and the 3rd argument is the location of all outputs on the cluster. So change them as necessary. The different stages called by the master script `svCall.sh` are (in order) `scanBam.sh` -> `assembly.sh` -> `alignAssembly.sh` -> `callVariants.sh`, which all take the same arguments.

3. To delete the whole cluster

```
./delete_cluster.sh sv-methods-1
```
This avoids having to wait for the web-based Console's confirmation.

One thing to note though, is that I've copied everything to a bucket at
```
gs://broad-dsde-methods/sv/
```
under a different project. We used to be developing under the project "broad-dsde-dev", but we are asked to move to project "broad-dsde-methods". So to run these scripts, you might need to switch to a different project via

```
gcloud config set project broad-dsde-methods
```

### DIFF
--- a/scripts/sv/alignAssembly.sh
+++ b/scripts/sv/alignAssembly.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -eu
+
+if [[ -z ${GATK_DIR+x} || -z ${CLUSTER_NAME+x} || -z ${MASTER_NODE+x} || -z ${PROJECT_OUTPUT_DIR+x} ]]; then 
+    if [[ "$#" -ne 3 ]]; then
+        echo "Please provide: local directory of GATK build, cluster name, output dir on the cluster"
+    exit 1
+    fi
+    GATK_DIR=$1
+    CLUSTER_NAME=$2
+    OUTPUT_DIR=$3
+    MASTER_NODE="hdfs://""$CLUSTER_NAME""-m:8020"
+    PROJECT_OUTPUT_DIR="$MASTER_NODE"/"$OUTPUT_DIR"
+fi
+
+cd "$GATK_DIR" 
+
+./gatk-launch AlignAssembledContigsSpark \
+    -R "$MASTER_NODE"/reference/Homo_sapiens_assembly19.fasta \
+    --inputFile "$PROJECT_OUTPUT_DIR"/assembly_0 \
+    -O "$PROJECT_OUTPUT_DIR"/aligned_assemblies \
+    -- \
+    --sparkRunner GCS \
+    --cluster "$CLUSTER_NAME" \
+    --driver-memory 30G \
+    --executor-memory 30G \
+    --num-executors 20 \
+    --conf spark.yarn.executor.memoryOverhead=16000
--- a/scripts/sv/alignAssembly.sh
+++ b/scripts/sv/alignAssembly.sh
@@ -14,7 +14,7 @@ if [[ -z ${GATK_DIR+x} || -z ${CLUSTER_NAME+x} || -z ${MASTER_NODE+x} || -z ${PR
     PROJECT_OUTPUT_DIR="$MASTER_NODE"/"$OUTPUT_DIR"
 fi
 
-REFERENCE_IMG_LOCATION="$MASTER_NODE"/reference/Homo_sapiens_assembly19.fasta.64.img
+REFERENCE_IMG_LOCATION=/reference/Homo_sapiens_assembly19.fasta.64.img
 
 echo "Assuming reference image: " "$REFERENCE_IMG_LOCATION"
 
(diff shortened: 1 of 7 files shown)


scripts/sv/delete_cluster.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -eu
+if [ "$#" -ne 1 ]; then
+    echo "Please provide name of cluster to be shutdown"
+    exit 1
+fi
+
+CLUSTER_NAME=$1
+
+gcloud dataproc clusters delete "$CLUSTER_NAME" --async
```
cwhelan: Does this need project information?

scripts/sv/alignAssembly.sh
```diff
@@ -0,0 +1,29 @@
...
+    CLUSTER_NAME=$2
+    OUTPUT_DIR=$3
+    MASTER_NODE="hdfs://""$CLUSTER_NAME""-m:8020"
+    PROJECT_OUTPUT_DIR="$MASTER_NODE"/"$OUTPUT_DIR"
+fi
+
+cd "$GATK_DIR" 
+
+./gatk-launch AlignAssembledContigsSpark \
+    -R "$MASTER_NODE"/reference/Homo_sapiens_assembly19.fasta \
```
cwhelan: Could you add top-level comments to scripts that are dependent on using hg19/b37 versions of the reference that highlights that fact? Just want to avoid any future pain on the part of someone who doesn't realize that. In a future PR we could figure out how to parameterize that (or hopefully just move away from needing these scripts altogether).

scripts/sv/scanBam.sh
```diff
@@ -0,0 +1,37 @@
...
+    OUTPUT_DIR=$3
+    MASTER_NODE="hdfs://""$CLUSTER_NAME""-m:8020"
+    PROJECT_OUTPUT_DIR="$MASTER_NODE"/"$OUTPUT_DIR"
+fi
+
+cd "$GATK_DIR" 
+
+./gatk-launch FindBreakpointEvidenceSpark \
+    -R "$MASTER_NODE"/reference/Homo_sapiens_assembly19.fasta \
+    -I "$MASTER_NODE"/data/NA12878_PCR-_30X.bam \
```
cwhelan: It might be nice to parameterize the bam name out of this script so it could be reused on other hg19 data.

scripts/sv/delete_cluster.sh
SHuang-Broad: Doesn't seem like it according to [this](https://cloud.google.com/dataproc/docs/guides/manage-cluster#delete_a_cluster)

scripts/sv/alignAssembly.sh
SHuang-Broad: Yep. Also working on generating a PR for a integration test that would take this into account.

scripts/sv/scanBam.sh
SHuang-Broad: Yep

SHuang-Broad:
@cwhelan Updated, please take a look again. Thanks!

SHuang-Broad:
@cwhelan I was actually debating with myself about whether to include the initialization script here, as it was living in the bucket referred to in the creation script.
So we could do this:
always store the initialization script locally with the creation script instead of referring to a script living remotely, and makes that a required argument. The good: this makes it easier to track changes; The bad: initialization script must be removed from the bucket to avoid tracking possible different versions.

A non-technical issue: we are "delivering" SGA in the initialization script, if that comes in to this repo, legal might have a problem with it. On the other hand, it the initialization script lives in a place only we can access, we are "installing SGA for our own use", which is not a problem with the GPL license.

cwhelan:
Oh, that's right, I'd forgotten about the SGA license issue. Since we're
about to move to fermi-lite (hopefully), let's just hold off on checking in
the initialization script until that's done, keeping it in the known bucket
location.

PR #2444, opened 2017-03-08
assemble and align contigs in FindBreakpointEvidenceSpark

tedsharpe:


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -370,62 +586,100 @@ private static void writeMetadata( final ReadMetadata readMetadata,
         return new Tuple2<>(kmerIntervalsAndDispositions._1(), kmerMultiMap);
     }
 
+    @VisibleForTesting interface FASTQHandler
+            extends Serializable, Function<Tuple2<Integer,List<FastqRead>>, AssemblyResults> {
+    }
+
     /**
      * Transform all the reads for a supplied set of template names in each inverval into FASTQ records
```
SHuang-Broad: typo `interval` -> `interval`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -370,62 +586,100 @@ private static void writeMetadata( final ReadMetadata readMetadata,
         return new Tuple2<>(kmerIntervalsAndDispositions._1(), kmerMultiMap);
     }
 
+    @VisibleForTesting interface FASTQHandler
```
SHuang-Broad: What about moving it&mdash;and its implementations&mdash;to `SVFastqUtils`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -181,36 +197,78 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+                                      final SAMFileHeader header, final List<AssemblyResults> assemblyResultsList ) {
+        try ( final OutputStream os = BucketUtils.createFile(samFile, pipelineOptions) ) {
+            final SAMTextWriter writer = new SAMTextWriter(os);
+            writer.setSortOrder(SAMFileHeader.SortOrder.queryname, true);
+            writer.setHeader(header);
+            final List<String> refNames =
+                header.getSequenceDictionary().getSequences().stream()
+                        .map(SAMSequenceRecord::getSequenceName).collect(Collectors.toList());
+            for ( final AssemblyResults assemblyResults : assemblyResultsList ) {
+                if ( assemblyResults.getErrorMessage() != null ) continue;
```
SHuang-Broad: Reading more, I see the error message is most likely generated not because of failure from fermi lite but too big an assembly file?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
```diff
@@ -1,68 +1,162 @@
...
+            }
+
+            @Override
+            public FastqRead read( final Kryo kryo, final Input input, final Class<FastqRead> type ) {
+                return new FastqRead(kryo, input);
+            }
+        }
+    }
+
+    public static List<FastqRead> readFastqFile( final String fileName, final PipelineOptions options ) {
```
SHuang-Broad: IDE tells me it is not used anywhere. Any future uses intended?

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSparkUnitTest.java
```diff
@@ -116,38 +117,40 @@ public void generateFastqsTest() {
...
-        final ByteArrayInputStream actualStream = new ByteArrayInputStream(concatenatedFastqs);
-        try( InputStream expectedStream = new BufferedInputStream(new FileInputStream(expectedFile)) ) {
-            int val;
-            while ( (val = expectedStream.read()) != -1 )
-                Assert.assertEquals(actualStream.read(), val);
-            Assert.assertEquals(actualStream.read(), -1);
-        }
-        catch ( final IOException ioe ) {
-            throw new GATKException("can't read expected values", ioe);
+    private static final class FastqComparator implements FindBreakpointEvidenceSpark.FASTQHandler {
```
SHuang-Broad: I believe this "complicated" code is indicating that the FASTQHandler interface may deserve a 1st class citizen treatment, so that if this test tries to access the `FASTQHandler` in `FindBreakpointEvidenceSpark`, the checks wouldn't complain.

tedsharpe:
Thanks for your comments, Steve.  I'll address them.
I don't know why I made a PR:  this code is untested.  Let's let it languish for a bit.

scripts/sv/callVariants.sh
```diff
@@ -21,12 +21,11 @@ echo "Assuming 2-bit reference: " "$TWOBIT_REFERENCE_LOCATION"
 
 cd "$GATK_DIR" 
 
-./gatk-launch CallVariantsFromAlignedContigsSpark \
-    --inputAlignments "$PROJECT_OUTPUT_DIR"/aligned_assemblies \
-    --inputAssemblies "$PROJECT_OUTPUT_DIR"/assembly_0 \
+./gatk-launch CallVariantsFromAlignedContigsSAMSpark \
```
tedsharpe: It was my impression that CallVariantsFromAlignedContigsSpark and CallVariantsFromAlignedContigsSAMSpark were identical except for input format.  It seemed to me that the SAM version was simpler and cleaner, because all of the necessary data is in one place in a single file.  Therefore, I changed FindBreakpoints to write a SAM file.  Am I mixed up?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: There are a couple of reasons that intervals don't get assembled (too big is one, too few kmers is another), and the idea of the AssemblyResults class is to present either the assembly, or the reason why there isn't one.  Though we can't write contigs that we didn't generate, the errors aren't silent.  They'll be output in the intervalFile.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -290,26 +351,181 @@ private static void writeMetadata( final ReadMetadata readMetadata,
...
             }
         }
-        catch ( IOException ioe ) {
+        catch ( final IOException ioe ) {
             throw new GATKException("Can't write metadata file.", ioe);
         }
     }
 
+    @DefaultSerializer(AssemblyResults.Serializer.class)
+    @VisibleForTesting static final class AssemblyResults {
```
tedsharpe: How about AlignedAssemblyOrExcuse?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
tedsharpe: Used to be used, but no longer necessary.  I'm leaving it for symmetry.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSparkUnitTest.java
tedsharpe: Sorry.  I don't understand this comment.

tedsharpe:
On the general comment of this class being too big:  I totally agree.
I haven't figured out how to make Java in this semi-functional style pretty.  I could certainly pull a mess of nested classes out into top-level classes just to make the file smaller.  But most of them are so specific to their use in this program that they wouldn't really be useful outside this context.  I'll probably do just that, but it's not clear to me that it makes the program more comprehensible.

scripts/sv/callVariants.sh
SHuang-Broad: Unfortunately, yes... Not your fault. When I picked up the generalization to expand to calling insertions and deletions, both `CallVariantsFromAlignedContigsSAMSpark` and `CallVariantsFromAlignedContigsSpark` already exist. The `SAM` version, if I recall correctly, was used for 10X data, and the shorter one was used for the short reads.
The logic and parsing becomes separate somehow when I started inputing more functionalities into the parsing step used in `CallVariantsFromAlignedContigsSpark` (like breaking gapped alignment and dealing with edge case CIGARs). 

That said, it should not be a problem too difficult to fix. What I can do is to branch out from your branch, put in the fix in `AssemblyAlignmentParser` just for expected input, and PR into your branch (to relieve you from the pain of digging into the details of the file parsing as I have fresh memory of the details). What do you think?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
SHuang-Broad: Got it. Is there a way to make it more apparent? When I wrote the link to SGA, Chris and I were discussing what to do in face of assembly failure, and it was decided to throw an exception is OK. Maybe outputting a warning?

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSparkUnitTest.java
SHuang-Broad: Sorry for the confusion. The intension was to not put too much complexity in the test code so that they are unit enough. If this concern is valid, then the `FastqComparator` would preferably live in main but not test, and most likely, like its parent `FastqHandler` live in `FindBreakpointEvidence`. But, I recall that there's some check in GATK that if other classes access inner classes of this class, there will be an error (maybe under some special conditions that I misunderstood), therefore to make this `FastqHandler` accessible to other classes, it probably be a class by itself. Am I still confusing?

scripts/sv/callVariants.sh
tedsharpe: Ah.  So the right action is to put SAM parsing into CallVariantsFromAlignedContigsSpark?
I can probably do that.  So far I've been careful not to disrupt the old SGA pipeline, but this would break that.  Might it be better for one of us to move your indel parsing code into ...SAMSpark?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Yes.  As it stands there will be a warning when assemblies are not done.  It will be written to the intervals file.  I've beefed up the comments.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSparkUnitTest.java
tedsharpe: I think maybe what you're saying is that the FastqComparator class that lives in FindBreakpointEvidenceSparkUnitTest seems like a lot of code for a unit test.  Have I grasped your issue?  If so, I could work on making it smaller, but to tell you the truth, it doesn't seem like a lot of code to me.

tedsharpe:
Sorry to outdate the comments.  This code will be ready to go, I believe, when we have either moved the up-to-date back end code that calls indels into CallVariantsFromAlignedContigsSAMSpark, or we have moved SAM input parsing into CallVariantsFromAlignedContigsSpark.
I have now tested it successfully on NA12878 on our cluster.

scripts/sv/callVariants.sh
SHuang-Broad: I'll do it.

SHuang-Broad:
@tedsharpe just a ping that #2453 does what we discussed here.

scripts/sv/callVariants.sh
cwhelan: The SAM version was just there so that I could harness the call variants step using a BAM file as input as opposed to our internal AlignmentRegion representation. If we're moving to a SAM representation for everything, we should probably just be able to have one tool.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -105,9 +107,12 @@
...
     // --------- locations ----------
 
-    @Argument(doc = "directory for fastq output", shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+    @Argument(doc = "sam file for aligned contigs", shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
             fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME)
-    private String outputDir;
+    private String outputSAM;
+
+    @Argument(doc = "bwa-mem index image file", fullName = "alignerIndexImage")
+    private String alignerIndexFile;
```
SHuang-Broad: Name change suggestion: `alignerIndexImageFile`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -774,7 +768,7 @@ public Params() {
             maxQNamesPerKmer = 500;                 // maximum template names for an assembly kmer
             assemblyKmerMapSize = 250000;           // guess for unique, error-free, scrubbed kmers per assembly partition
             assemblyToMappedSizeRatioGuess = 7;     // guess for ratio of total reads in assembly to evidentiary reads in interval
-            maxFASTQSize = 10000000;                // maximum FASTQ size
+            maxFASTQSize = 3000000;                 // maximum assembly size (total input bases)
```
SHuang-Broad: Why the reduction in size?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/QNameFinder.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.function.Function;
+
+/**
+ * Class to find the template names associated with reads in specified intervals.
+ */
+public final class QNameFinder implements Function<GATKRead, Iterator<QNameAndInterval>> {
+    private final ReadMetadata metadata;
+    private final List<SVInterval> intervals;
+    private final Iterator<QNameAndInterval> noName = Collections.emptyIterator();
+    private int intervalsIndex = 0;
```
tedsharpe: Each instance of QNameFinder is processing one partition of the BAM.  The reads are known to be coordinate ordered, as are the intervals.  We don't want to search from the beginning of the intervals list with each applied read -- especially for the later partitions we'd be walking a long way to find our spot.  Instead, we save our location from the last read, and increment when necessary.  Point is that because the data is ordered, we'll never need to back up in the intervals list, we'll only occasionally need to read further into the list.
It would work as an instance variable, but it would be much, much slower.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/QNameFinder.java
```diff
@@ -0,0 +1,43 @@
...
+    public Iterator<QNameAndInterval> apply( final GATKRead read ) {
+        final int readContigId = metadata.getContigID(read.getContig());
+        final int readStart = read.getUnclippedStart();
+        final int intervalsSize = intervals.size();
+        while ( intervalsIndex < intervalsSize ) {
+            final SVInterval interval = intervals.get(intervalsIndex);
+            if ( interval.getContig() > readContigId ) break;
+            if ( interval.getContig() == readContigId && interval.getEnd() > read.getStart() ) break;
+            intervalsIndex += 1;
+        }
```
tedsharpe: Intervals are disjoint.  So, no.  It's not possible.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadsForQNamesFinder.java
```diff
@@ -0,0 +1,61 @@
...
+                                 final int nIntervals, final boolean includeMappingLocation, final boolean dumpFASTQs ) {
+        this.qNamesMultiMap = qNamesMultiMap;
+        this.nIntervals = nIntervals;
+        this.nReadsPerInterval = 2 * qNamesMultiMap.size() / nIntervals;
+        this.includeMappingLocation = includeMappingLocation;
+        this.dumpFASTQs = dumpFASTQs;
+    }
+
+    public Iterable<Tuple2<Integer, List<SVFastqUtils.FastqRead>>> call( final Iterator<GATKRead> readsItr ) {
+        @SuppressWarnings({ "unchecked", "rawtypes" })
```
tedsharpe: No.  It's necessary because I'm downcasting from a raw type.  (Note that it's not possible in Java to write new List<>[nIntervals].  You can't make arrays of a generic type.  I had to write new List[nIntervals] instead.)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/IntervalCoverageFinder.java
```diff
@@ -0,0 +1,47 @@
...
+        int intervalsIndex = 0;
+        while ( readItr.hasNext() ) {
+            final GATKRead read = readItr.next();
+            final int readContigId = metadata.getContigID(read.getContig());
+            final int readStart = read.getUnclippedStart();
+            final int intervalsSize = intervals.size();
+            while ( intervalsIndex < intervalsSize ) {
+                final SVInterval interval = intervals.get(intervalsIndex);
+                if ( interval.getContig() > readContigId ) break;
+                if ( interval.getContig() == readContigId && interval.getEnd() > read.getStart() ) break;
```
tedsharpe: overlapLen will be 0, in this case.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSparkUnitTest.java
```diff
@@ -55,7 +55,7 @@ public void getQNamesTest() {
         final Set<String> actualQNames = new HashSet<>();
         FindBreakpointEvidenceSpark.getQNames(params, ctx, broadcastMetadata, expectedIntervalList, mappedReads)
                 .stream()
-                .map(FindBreakpointEvidenceSpark.QNameAndInterval::getKey)
+                .map(QNameAndInterval::getKey)
```
tedsharpe: Set comparison is implemented in a rather silly way in the java library.  We need to make sure we have an O(1) get function.  Collectors.toSet makes no guarantees what kind of set you'll get.  Assert.assertEquals makes no guarantee about which set will be used for iteration and which will be used for lookup.  Only safe thing to do is to pin down the types.
This is a trifling worry for this little unit test, but it's just my habit.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Counting bases rather than bytes of FASTQ.

tedsharpe:
Back to you @SHuang-Broad

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/QNameFinder.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    public Iterator<QNameAndInterval> apply( final GATKRead read ) {
+        final int readContigId = metadata.getContigID(read.getContig());
+        final int readStart = read.getUnclippedStart();
+        final int intervalsSize = intervals.size();
+        while ( intervalsIndex < intervalsSize ) {
+            final SVInterval interval = intervals.get(intervalsIndex);
+            if ( interval.getContig() > readContigId ) break;
+            if ( interval.getContig() == readContigId && interval.getEnd() > read.getStart() ) break;
```
SHuang-Broad: Question solved after in person discussion.

SHuang-Broad:
👍

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -371,61 +340,104 @@ private static void writeMetadata( final ReadMetadata readMetadata,
...
-
-        final int fastqSize = fastqsList.stream().mapToInt(fastqRec -> fastqRec.length).sum();
-        final String disposition;
-        if ( fastqSize > maxFastqSize ) disposition = "FASTQ not written -- too big (" + fastqSize + " bytes).";
-        else {
-            final String fileName = outputDir + "/assembly" + intervalAndFastqs._1() + ".fastq";
-            SVFastqUtils.writeFastqFile(fileName, null, fastqsList);
-            disposition = fileName;
+    /** This FASTQHandler aligns assembly contigs with BWA, along with some optional writing of intermediate results. */
+    private static final class FastqWriter implements FASTQHandler {
```
cwhelan: It doesn't seem right to have a the actual assemblies being performed by a class called `FastqWriter`. I see that the `FASTQHandler` interface is there to help you with testing. Suggested renames:

```
FASTQHandler -> FASTQAssemblyProducer
FastqWriter -> FASTQFermiAssembler
```

What do you think?

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
```diff
@@ -15,34 +13,38 @@
...
+     * readGroup can be null.
      *
      * Also, BWA does this odd thing, supplying AS and XS tags for unaligned reads.
      * To produce SAM records that are byte-identical to the ones created by the bwa mem command line,
      * call this method with alwaysGenerateASandXS=true.
      */
-    public static GATKRead applyAlignment(final GATKRead unalignedRead, final BwaMemAlignment alignment,
-                                            final List<String> refNames, final SAMFileHeader header,
-                                            final boolean softclipAlts, final boolean alwaysGenerateASandXS ) {
+    public static SAMRecord applyAlignment( final String readName, final byte[] basesArg, final byte[] qualsArg,
```
cwhelan: How come this now returns a SAMRecord instead of a GATKRead? The comment "Transforms an unaligned GATKRead into an aligned one, as specified by the supplied BwaMemAlignment" implies to me that it would return another GATKRead. I'd at least update the comment and maybe add an explanation.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
tedsharpe: Agreed.  Name could be better.
Doesn't necessarily have anything to do with FASTQ, though.  That was just a means to an end when we were using SGA.
Maybe it's a LocalAssemblyHandler interface, and a FermiLiteAssemblyHandler implementation?

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
tedsharpe: Comment is out of date, and I'll fix it.  The reason it's a SAMRecord that I'm returning is that I'm writing a SAM file for the variant caller to use, so this is maximally convenient for me.  It's a one-liner to boost a SAMRecord into a GATKRead (by wrapping it with an adapter), so other users of the utility shouldn't find it too burdensome.  Currently there's only one other user, the BwaSparkEngine.

Issue #2414, opened 2017-02-21
How to generate  bwamemIndexImage


tushu1232:
Bwa Spark needs a required argument  bwamemIndexImage.How to generate it.
Are there any defined steps

lbergelson:
Hi @tushu1232.  The index image is a new feature that's not fully exposed in a friendly.  It's the 5 indexes that bwa requires baked into a single file for easy distribution. ( .amb, .ann, .bwt, .pac, .sa )  We're going to add an official tool to generate it in the near future, but at present the only way to do so is to write a tool yourself that calls into org.broadinstitute.hellbender.utils.bwa.BwaMemIndex.createIndexImage();

rtemanni:
Hi @lbergelson, would it be possible to have the indeximage for hs37d5 and hs38DH available for download ?
Thanks in advance.

lbergelson:
@rtemanni That sounds like a good idea to host them.  Maybe we could include them in the GATKBundle in the future.  They're pretty large files though, and can be generated from the reference directly, so I'm not sure if it makes sense to download them or just create the index yourself from your own reference.  We will definitely consider it though.

SHuang-Broad:
Did something [here](https://github.com/SHuang-Broad/bwamem-index-image) with a "shameless" copy of Ted's original code. It takes a very short time to create the image, so downloading the big image file (~5GB) might not worth it.

lbergelson:
@SHuang-Broad Why not just add that as a gatktool?

SHuang-Broad:
Laziness... I just did that project today for fun & practice. It shouldn't be difficult to do it in Java. I'll do it later.

PR #2452, opened 2017-03-10
Simple cmd line tool for creating index image file

SHuang-Broad:
@lbergelson here you go for #2414


src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreator.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+              shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+              doc = "Output file (if not provided, defaults to STDOUT)",
+              optional = false)
+    private String referenceIndexImageOutputLoc = null;
+
+    @Override
+    protected final Object doWork() {
+        BwaMemIndex.createIndexImage(referenceFastaLoc, referenceIndexImageOutputLoc);
```
lbergelson: @SHuang-Broad @tedsharpe As far as I can see it looks like `createIndexImage` requires that they already ran the `bwa index` command.  

Would it be possible for us to expose that `index` to the jni so that you wouldn't need a separate copy of bwa-mem to create the index.  (It could cause a lot of confusion if there are version differences between the packaged bwa mem and the one that people are indexing with.)

src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
```diff
@@ -31,6 +31,10 @@ public void unloadIndex() {
 
     @Test
     public void testAlignSingleRead() throws Exception {
+        singleReadAlingment(index);
+    }
+
+    public static void singleReadAlingment(final BwaMemIndex index) {
```
lbergelson: typo: Ali**ng**ment -> Ali**gn**ment

src/test/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreatorTest.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Test if image can be successfully created.
+ */
+public class BwaMemIndexImageCreatorTest extends CommandLineProgramTest {
```
lbergelson: This should be named `BwaMemIndexImageCreatorIntegrationTest`

src/test/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreatorTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private static final File TEST_DATA_DIR = new File("src/test/resources/large/");
+    private static final File testReferenceFasta = new File(TEST_DATA_DIR, "human_g1k_v37.20.21.fasta");
+
+    @Override
+    public String getTestedClassName() {
```
lbergelson: if you name it <ClassName>IntegrationTest then you don't need to override this.

src/test/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreatorTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+        final File tempImage = BaseTest.createTempFile("tempBwaMemIndexImage", ".img");
+        final List<String> args = new ArrayList<>(Arrays.asList(
+                "--input", testReferenceFasta.getAbsolutePath(),
+                "--output", tempImage.getAbsolutePath()));
+        runCommandLine(args);
+
+        // pigg-backing on the existing integration test
+        final BwaMemIndex index = new BwaMemIndex(tempImage.getAbsolutePath());
+        BwaMemIntegrationTest.singleReadAlingment(index);
```
lbergelson: could you move these two methods into their own class called `BwaMemTestUtils`.  It can stay in src/test.  We prefer to not have actual tests call into eachother because it can make it confusing when someone modifies a test class and other break.

src/test/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreatorTest.java
```diff
@@ -0,0 +1,47 @@
...
+    public void testImageFileGeneration() throws Exception {
+
+        final File tempImage = BaseTest.createTempFile("tempBwaMemIndexImage", ".img");
+        final List<String> args = new ArrayList<>(Arrays.asList(
+                "--input", testReferenceFasta.getAbsolutePath(),
+                "--output", tempImage.getAbsolutePath()));
+        runCommandLine(args);
+
+        // pigg-backing on the existing integration test
+        final BwaMemIndex index = new BwaMemIndex(tempImage.getAbsolutePath());
```
lbergelson: This should be in a try() with resources so it's auto closed.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreator.java
```diff
@@ -0,0 +1,38 @@
...
+@CommandLineProgramProperties(
+        summary = "Creates the image file for use by BwaMemAligner",
+        oneLineSummary = "Creates the image file for use by BwaMemAligner",
+        programGroup = BwaMemUtilitiesProgramGroup.class
+)
+public final class BwaMemIndexImageCreator extends CommandLineProgram {
+
+    @Argument(fullName = StandardArgumentDefinitions.INPUT_LONG_NAME,
+              shortName = StandardArgumentDefinitions.INPUT_SHORT_NAME,
+              doc = "Output file (if not provided, defaults to STDOUT)",
```
lbergelson: It should also mention that you need the five indexes pre-computed amb, awm, etc.

src/test/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreatorTest.java
lbergelson: also, I would rename them so they have `assert` in their name, i.e. `assertCorrectSingleReadAlignment`

src/main/java/org/broadinstitute/hellbender/cmdline/programgroups/BwaMemUtilitiesProgramGroup.java
```diff
@@ -0,0 +1,16 @@
+package org.broadinstitute.hellbender.cmdline.programgroups;
+
+import org.broadinstitute.barclay.argparser.CommandLineProgramGroup;
+import org.broadinstitute.hellbender.utils.help.HelpConstants;
+
+/**
+ * Program group for using non-essential part of JNI binding to Bwa Mem
```
SHuang-Broad: My impression is that the binding is for alignment only, and the distribution of images of those indices is more desirable in the Spark context. I can certainly remove that if you feel that would be more appropriate.

src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
SHuang-Broad: done

src/test/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreatorTest.java
SHuang-Broad: I extended the `CommandLineProgramTest` to piggy-back on its `runCommandLine()`

src/test/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreatorTest.java
SHuang-Broad: done

src/test/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreatorTest.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/cmdline/programgroups/BwaMemUtilitiesProgramGroup.java
SHuang-Broad: changed to "misc"

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreator.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreator.java
```diff
@@ -0,0 +1,38 @@
...
+              optional = false)
+    private String referenceFastaLoc = null;
+
+    @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+              shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+              doc = "Output file (if not provided, defaults to STDOUT)",
```
SHuang-Broad: done

SHuang-Broad:
@lbergelson Thanks for the review and sorry for the late reply.

src/test/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexImageCreatorTest.java
lbergelson: CommandLineProgramTest has a bit of magic that determines the testedClassName automatically if it's name <NameOfTool>IntegrationTest.  I think this should be renamed still.

lbergelson:
@SHuang-Broad looks good.  I still think you should rename that integration test though.

lbergelson:
👍

droazen:
The new tool should be moved from the `utils` package to the `tools` package @SHuang-Broad

PR #2453, opened 2017-03-10
Reconcile the two alignment formats expected in SV calling stage

SHuang-Broad:
@tedsharpe 
Here you go. I hope this does not generate too much trouble for you.


tedsharpe:
thanks, steve

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -31,7 +35,32 @@
...
+        if (! primaryRead.isPresent()) {
+            throw new GATKException("no primary alignment for read " + gatkReads.get(0).getName());
+        }
+
+        final byte[] bases = primaryRead.get().getBases();
+        if (primaryRead.get().isReverseStrand()) {
+            SequenceUtil.reverseComplement(bases, 0, bases.length);
+        }
+
+        final Iterable<AlignmentRegion> alignmentRegionIterable = IteratorUtils.asIterable(gatkReads.stream()
```
tedsharpe: This will be a single-use Iterable (i.e., one that can only iterate once, and is always empty thereafter).  That's probably OK, but it's certainly not clear that it will be without examining the calling code.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/CallVariantsFromAlignedContigsSAMSpark.java
```diff
@@ -72,7 +64,7 @@ public ReadFilter makeReadFilter() {
     protected void runTool(final JavaSparkContext ctx) {
 
         final JavaRDD<Iterable<GATKRead>> alignmentsGroupedByName = getReads().mapToPair(r -> new Tuple2<>(new Tuple2<>(r.getName(), r.getName()), r)).groupByKey().map(Tuple2::_2);
```
tedsharpe: Actually, we could just insist that the input SAM be queryname sorted -- then we don't have to grouping at all.

SHuang-Broad: nope. you are correct. done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java
```diff
@@ -31,7 +35,32 @@
     private static final boolean DEBUG_STATS = false;
 
     /**
-     * Loads the alignment regions and sequence of all locally-assembled contigs from the text file they are in;
+     * Converts a {@link GATKRead}, particularly the alignment and sequence information in it to the custom {@link AlignmentRegion} format.
+     */
+    @VisibleForTesting
+    static Tuple2<Iterable<AlignmentRegion>, byte[]> convertToAlignmentRegions(final Iterable<GATKRead> reads) {
```
SHuang-Broad: done with 2 pass streaming, one is essentially for cautiously checking that there's a primary alignment in the input.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AssemblyAlignmentParser.java

PR #2455, opened 2017-03-11
Update to Barclay 1.0.0-24-g87c3fa2-SNAPSHOT

### LINKED ISSUE #2454: Request: update Barclay dependency
Please, I would like to update my GATK dependency to the latest master. Nevertheless, the Barclay version included contains a bug that is solved in https://github.com/broadinstitute/barclay/commit/87c3fa228fa96d11b7177f8bfdfca4801c83a068, but a change in the usage method of Barclay before the fix to print hidden arguments broke the GATK command line class.

Could it be possible to update to the latest Barclay master, please?

magicDGS:
Fixes #2454

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -104,7 +104,7 @@ dependencies {
     compileOnly(javadocJDKFiles)
     testCompile(javadocJDKFiles)
 
-    compile 'org.broadinstitute:barclay:1.0.0-17-g30db73c-SNAPSHOT'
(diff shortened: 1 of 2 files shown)


magicDGS:
Can you have a look to this one, @cmnbroad? It is just a simple change for let me upgrade my dependencies and do not include the NPE in not bounded arguments...

cmnbroad:
We're already on a Barclay snapshot, so this should be fine. I'm going to manually run tests with this merged with the filter descriptor changes, and generate doc manually just to make sure nothing has regressed, then will merge.

PR #2456, opened 2017-03-12
updating sv cluster creation scripts for partial data copy customization

SHuang-Broad:
@cwhelan @tedsharpe @vruano please review.
It is intended for copying the right data for the "correctly"-named cluster for performing intended analysis.
Cluster provisioning time is now ~25 minutes.

### DIFF
--- a/scripts/sv/create_cluster.sh
+++ b/scripts/sv/create_cluster.sh
@@ -5,12 +5,32 @@
 set -eu
 if [[ "$#" -ne 3 ]]; then
     echo "Please provide the project, the name of cluster, and the bucket where the data lives"
+    echo "For the name of the cluster, if running on [1] NA12878 sample, [2] CHM mix sample, "
+    echo "  please indicate in the name of the cluster by including a string [1] 'na12878', [2]'chmmix', "
+    echo "  so that the correct data is copied directly for you."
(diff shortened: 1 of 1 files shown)


cwhelan:
@SHuang-Broad 

Nice hack using the cluster name. I don't see any other way to pass an arg to an initialization action.

I have one suggestion to consider, but if you think it's too much work or not worth it feel free to skip: what if we separated out the reference bundle to copy from the data by specifying them both in the cluster name? That way we could, say, load either the hg19 or hg38 reference depending on the data we might be working with. So you could say "cluster-hg38" or "cluster-hg19" or "cluster-hg19-na12878". 

Carrying it further, if we had a special convention for specifying data, like "data-$SAMPLE", we could just map $SAMPLE to a subdirectory on the bucket. That would provide a ton of flexibility.

One minor note while you are messing with these scripts: the createCluster.sh script comment says "This script deletes a Google Dataproc cluster used for running the GATK-SV pipeline." Could you change to say it creates rather than deletes a cluster?

vruano:
I don't think I'm yet in a position to review this... as far as I can see it does not raises flags.

SHuang-Broad:
Now replaced by #2468

Issue #2310, opened 2016-12-12
FeatureWalker intervals are ignored


magicDGS:
As an example, the following test is failing in [`ExampleFeatureWalkerIntegrationTest`](https://github.com/broadinstitute/gatk/blob/master/src/test/java/org/broadinstitute/hellbender/tools/examples/ExampleFeatureWalkerIntegrationTest.java):

```java
...
    @Test
    public void testExampleFeatureWalkerWithIntervals() throws IOException {
        IntegrationTestSpec testSpec = new IntegrationTestSpec(
                " -R " + hg19MiniReference +
                        " -I " + TEST_DATA_DIRECTORY + "reads_data_source_test1.bam" +
                        " -F " + TEST_DATA_DIRECTORY + "example_features.bed" +
                        " -L 1 " +
                        " -auxiliaryVariants " + TEST_DATA_DIRECTORY + "feature_data_source_test.vcf" +
                        " -O %s",
                Arrays.asList(TEST_OUTPUT_DIRECTORY + "expected_ExampleFeatureWalkerIntegrationTestWithIntervals_output.txt")
        );
        testSpec.executeTest("testExampleIntervalWalker", this);
    }
```

Where `expected_ExampleFeatureWalkerIntegrationTestWithIntervals_output.txt` is just the first feature from the [`expected_ExampleFeatureWalkerIntegrationTest_output.txt`](https://github.com/broadinstitute/gatk/blob/master/src/test/resources/org/broadinstitute/hellbender/tools/examples/expected_ExampleFeatureWalkerIntegrationTest_output.txt):

```
Current feature: htsjdk.tribble.bed.FullBEDFeature:1:11-300
	Overlapping reference bases: NNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNN

	Overlapping read at 1:200-275
[...]

PR #2467, opened 2017-03-17
Fix FeatureWalker intervals

magicDGS:
Currently the interval arguments are ignored in `FeatureWalker` (#2310) because the initialization of intervals is done after the ones for features. This PR includes the following changes:
- Failing test (as described in #2310)
- Fix by moving the code for set the intervals for traversal to `onStartup()`, as in the case of the reads in `ReadWalker`.


magicDGS:
This is a small patch which solves a bug in `FeatureWalker` which was driving me crazy with some of my custom tools. Please, could you have a quick look, @droazen? Thank you in advance!

droazen:
:+1: thanks for this fix @magicDGS !  Looks good, merging.

magicDGS:
Thanks @droazen!

PR #2468, opened 2017-03-19
Updated scripts for sv cluster creation, initialization and pipeline execution

SHuang-Broad:
Replaces #2456 as recommended by @cwhelan and helped by @lbergelson.


scripts/sv/svDiscover.sh
```diff
@@ -0,0 +1,46 @@
...
+
+set -eu
+
+if [[ "$#" -lt 7 ]]; then
+    echo -e "Please provide:"
+    echo -e "  [1] local directory of GATK build (required)"
+    echo -e "  [2] cluster name (required)"
+    echo -e "  [3] output directory on the cluster (required)"
+    echo -e "  [4] absolute path to the BAM on the cluster (index assumed accompanying the bam) (required)"
+    echo -e "  [5] absolute path to the reference fasta on the cluster (required)"
```
cwhelan: For these and other usage comments, could you just specify that the path should be on HDFS? This is always something that's handy to see at a glance.

scripts/sv/svDiscover.sh
```diff
@@ -0,0 +1,46 @@
...
+    echo -e "  [6] absolute path to the 2-bit reference on the cluster (required)"
```
cwhelan: Would it be possible to have a naming convention or something that would allow us to only have _one_ location for the reference, and then it could infer the associated files? Ie. if you knew the reference base name, we could assume that you could access $basename.fasta, $basename.2bit, $basename.indeximage, and $basename.skiplist? This would make things a lot more streamlined.

scripts/sv/scanBam.sh
```diff
@@ -2,30 +2,33 @@
...
-echo "Assuming reference skip list: " "$SKIP_INTERVAL_LIST"
-
 cd "$GATK_DIR" 
 
 ./gatk-launch FindBreakpointEvidenceSpark \
-    -R "$REFERENCE_LOCATION" \
+    -R "$REF_FASTA" \
     -I "$INPUT_BAM" \
     -O "$PROJECT_OUTPUT_DIR"/fastq \
     --exclusionIntervals "$SKIP_INTERVAL_LIST" \
```
cwhelan: One more comment -- the 'dups' file on the next line hard-codes a path to /reference that's not referred to anywhere else.

scripts/sv/create_cluster.sh
```diff
@@ -1,27 +1,64 @@
...
 
 set -eu
-if [[ "$#" -ne 3 ]]; then
-    echo "Please provide the project, the name of cluster, and the bucket where the data lives"
+
+if [[ "$#" -lt 5 ]]; then
+    echo -e "Please provide:"
+    echo -e "  [1] project name (required)"
+    echo -e "  [2] cluster name (required)"
+    echo -e "  [3] GCS directory containing the correct referencei & indices (required)"
```
SHuang-Broad: done

scripts/sv/create_cluster.sh
```diff
@@ -1,27 +1,64 @@
...
+    echo -e "  [5] local initialization script"
+    echo -e "  [6] GCS path to upload initialization script to (required as per Google)"
+    echo -e "or when the initialization script is already in the cloud: "
+    echo -e "  [5] path to initialization script on GCS"
+    echo -e "\nExample when you have a local custom initialization script:"
+    echo -e "  bash create_cluster.sh \\"
+    echo -e "       <PROJECT-NAME> \\"
+    echo -e "       <CLUSTER-NAME> \\"
+    echo -e "       <BUCKET-DIR-STORING-REFERENCE> \\"
+    echo -e "       <BUCKET-DIR-STROING-BAM> \\"
```
SHuang-Broad: done

scripts/sv/default_init.sh
```diff
@@ -0,0 +1,26 @@
...
+# On the worker nodes we copy the bwa index image file 
+# from a bucket to each node's local disk.
+
+NAME=$(/usr/share/google/get_metadata_value hostname)
+ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
+REFLOC=$(/usr/share/google/get_metadata_value attributes/reference)
+SAMPLE=$(/usr/share/google/get_metadata_value attributes/sample)
+
+if [[ "${ROLE}" == 'Master' ]]; then
+    hadoop distcp "$REFLOC"/* hdfs://"$NAME":8020/reference/ &
```
SHuang-Broad: Not sure, but this 10G directory is not the one lagging the initialization (the 100G bam is), so I'm not too concerned about it here.

scripts/sv/scanBam.sh
SHuang-Broad: Yes, we need to make a choice,
1. ask Ted to generate a dups for hg19, and put it in the correct place then we can deal with it like the 2bit references. 
2. we "lie" in its name, pretend it is hg19 with the file name and use the same contents from GRCh38-versioned file's content.

scripts/sv/svDiscover.sh
SHuang-Broad: done, except for the index image fie because it lives on local file systems rather than hdfs, and except for the "....dups" file.

scripts/sv/svDiscover.sh
SHuang-Broad: done, but not sure if it fits what you suggested exactly.

SHuang-Broad:
Addressed comments. 
Note to self: when this is in, the two "....kill.intervals" files must be renamed to have the same base name as the reference FASTA file in our bucket.

scripts/sv/scanBam.sh
cwhelan: Just call it a "kmer_kill_list" or something similar, it doesn't matter how it was generated as long as it's still appropriate to use on hg19, which we think it is.

cwhelan:
This looks good now. Feel free to merge when ready.

scripts/sv/scanBam.sh
SHuang-Broad: Done

PR #2488, opened 2017-03-20
Move to the latest version of google-cloud-java

jean-philippe-martin:
Main motivation: it includes fixes we need for writing to
Google Cloud Storage buckets via NIO.

However this comes with some API changes too so we must agree
on what to do there.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -112,7 +112,7 @@ dependencies {
     // Using the shaded version to avoid conflicts between its protobuf dependency
     // and that of Hadoop/Spark (either the one we reference explicitly, or the one
     // provided by dataproc).
-    compile 'com.google.cloud:google-cloud-nio:0.5.1:shaded'
+    compile 'com.google.cloud:google-cloud-nio:0.10.0-alpha:shaded'
     compile 'com.google.cloud.genomics:google-genomics-dataflow:v1beta2-0.15'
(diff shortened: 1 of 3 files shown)


jean-philippe-martin:
It looks like this version sadly doesn't include the [package rename](https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1767).

jean-philippe-martin:
A workaround is to comment out the one method that needs it until the next release (no one inside of gatk calls this method). Would that be OK?

droazen:
Which method is that @jean-philippe-martin ?

jean-philippe-martin:
It's `BucketUtils.getAuthenticatedGcs`

droazen:
That method is used in tests, isn't it? 

I'd be ok with commenting it out for now for the sake of getting this merged, provided that you open a ticket to uncomment the method and re-enable the relevant tests once the package rename is available.

jean-philippe-martin:
Yes you're right, it's only used in tests. Adding a ticket for it.

jean-philippe-martin:
@droazen Was that a thumbs up, or would you like to do more checking before I squash and merge?

droazen:
That was not a thumbs up, but this is ----> 👍   merge when ready!

PR #2491, opened 2017-03-20
allow null-likelihoods in  RankSumTest annotate

lbergelson:
previously RankSumTest disallowed null likelihoods in it's annotate method
this was different from GATK3 and causes problems in GenotypeGVCFs

relaxing this constraint so annotate will now return Collections.emptyMap() if likelihoods is null


droazen:
:+1: merge when ready

Issue #2120, opened 2016-08-25
BwaSpark filters reads incorrectly


lbergelson:
BwaSpark is using WELLFORMED as it's read filter.  This includes a number of checks that the reads validate against the header.  However, most unaligned sam files are likely to have no or incomplete headers.  This causes many reads to be unexpectedly filtered.

We should define a better filter criteria BwaSpark.

droazen:
@tedsharpe Would you have time to have a quick look at this to see if it's still an issue for the current version of `BwaSpark`?

lbergelson:
this is the cause of #2038

lbergelson:
I have a fix for this in lb_fixing_read_filter_for_bwa but I still need to write a test.

tedsharpe:
Sure.  I can take a look.  Seems like partly a matter of taste, however.  The only real requirement for a read that is to be aligned is that it have some sequence.  Shall we back the filter off to that bare minimum?

droazen:
Closing -- this is fixed.

Issue #2165, opened 2016-09-14
PrimaryAlignmentReadFilter doesn't filter supplementary reads


cmnbroad:
PrimaryAlignmentReadFilter filters out secondary alignments, but not supplementary alignments (this is what GATK3 does as well). This doesn't match the SAM spec (though it does match the language used by htsjdk); is redundant given that in GATK4 we have a NotSecondaryAlignmentReadFilter for use with InsertSizeMetricsCollector; and is likely confusing.

Proposed change is to visit the usages of the PrimaryAlignmentReadFilter (BaseRecalibrator, Pileup, CalculateTargetBaseCallCoverage and HaplotypCaller) and for any that don't want supplementary alignments filtered, replace the filter with NotSecondary, and for those that do, change Primary to also reject supplementary reads.

cmnbroad:
Related to GATK3 discussion https://github.com/broadinstitute/gsa-unstable/issues/1472.

magicDGS:
I know the usage in the `Pileup`, because I added it when I ported it: I've just tried to mimic the samtools mpileup default filter flags. In samtools, the help of mpileup said regarding the filters: 
```
--ff, --excl-flags STR|INT  filter flags: skip reads with mask bits set
                                            [UNMAP,SECONDARY,QCFAIL,DUP]
```
Thus I guess that in the `Pileup` the usage is correct.

cmnbroad:
Closed via https://github.com/broadinstitute/gatk/pull/3195.

PR #2494, opened 2017-03-20
Improve BwaSpark by setting more reasonable default set of read filters

tedsharpe:
Also added pairedMode flag
Begins to address issue #2120


src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -145,14 +145,12 @@ private ReadFilterLibrary(){ /*no instance*/ }
         @Override public boolean test(final GATKRead read) {
             return read.isProperlyPaired();}}
 
-    //TODO: should this reject supplementary alignments ? If not it should be
-    //removed since its redundant with NotSecondaryAlignmentReadFilter
-    //https://github.com/broadinstitute/gatk/issues/2165
+    //* this now tests for unmapped or neither secondary nor supplementary as its name suggests */
```
lbergelson: So I'm not sure this PR is the right place to address this issue/ I'm not sure we're in agreement about changing PrimaryAlignmentFilter.  It's confusing because the sam spec refers to `primary` as meaning `not secondary`, which doesn't quite jibe with common usage as `not secondary AND not supplementary`.  I think it's best if we at the very least address this in a separate PR (which needs a matching PR in protected to fix tools that intended to use the original meaning.)

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -145,14 +145,12 @@ private ReadFilterLibrary(){ /*no instance*/ }
...
     @DocumentedFeature(groupName=HelpConstants.DOC_CAT_READFILTERS, groupSummary=HelpConstants.DOC_CAT_READFILTERS_SUMMARY)
     public static class PrimaryAlignmentReadFilter extends ReadFilter {
         private static final long serialVersionUID = 1L;
         @Override public boolean test(final GATKRead read) {
-            return ! read.isSecondaryAlignment();}}
+            return read.isUnmapped() || !(read.isSecondaryAlignment()||read.isSupplementaryAlignment());}}
```
lbergelson: I don't think people would expect this to accept unmapped reads.  You might need to define a different filter just for the bwa use that has this behavior.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -29,6 +33,9 @@
             fullName = "bwamemIndexImage")
     private String indexImageFile;
 
+    @Argument(doc = "do pairwise alignment", fullName = "pairedMode", optional = true)
```
lbergelson: This should be inverted to do single ended alignment.  We typically make flags default to false since it's a bit weird to have to specify a value for the flag.   It's fine to keep the option in all the other code as `pairedMode`, just invert this argument.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
```diff
@@ -31,29 +28,45 @@
...
+
+    public BwaSparkEngine( final JavaSparkContext ctx,
+                           final String indexFileName,
+                           SAMFileHeader inputHeader,
+                           final SAMSequenceDictionary refDictionary,
+                           final boolean pairedMode ) {
         Utils.nonNull(indexFileName);
         Utils.nonNull(inputHeader);
+        if ( pairedMode && inputHeader.getSortOrder() != SAMFileHeader.SortOrder.queryname ) {
+            throw new GATKException("Input must be queryname sorted to use paired alignment mode.");
```
lbergelson: I think this should be a `UserException` since the user provided a bad bam file.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
```diff
@@ -80,8 +95,17 @@ public void close() {
...
-            if ( (nReads & 1) != 0 ) {
-                throw new GATKException("We're supposed to be aligning paired reads, but there are an odd number of them.");
+            if ( pairedMode ) {
+                if ( (nReads & 1) != 0 ) {
+                    throw new GATKException("We're supposed to be aligning paired reads, but there are an odd number of them.");
+                }
+                for ( int idx = 0; idx != nReads; idx += 2 ) {
+                    final String readName1 = inputReads.get(idx).getName();
+                    final String readName2 = inputReads.get(idx+1).getName();
+                    if ( !Objects.equals(readName1,readName2) ) {
```
lbergelson: good idea to add this check

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -40,15 +47,25 @@ public boolean requiresReads() {
     }
 
     @Override
-    protected void runTool(final JavaSparkContext ctx) {
+    public List<ReadFilter> getDefaultReadFilters() {
+        // 1) unmapped or neither secondary nor supplementary and 2) has some sequence
+        return Collections.singletonList(ReadFilterLibrary.PRIMARY_ALIGNMENT.and(ReadFilterLibrary.SEQ_IS_STORED));
```
lbergelson: I think it's preferred to have the filters be separate entries in a list rather than chaining them.  If they're separated out individual then they can be toggled on/off on the command line and will show up in help output appropriately.  If they're chained into a single filter then I don't think they can be inspected properly.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
lbergelson: I think it's fine to keep the changes in the other tools to use `NOT_SECONDARY_ALIGNMENT` though even though it's really part of #2165.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
lbergelson: We might also want to include `ReadFilterLibrary.HAS_MATCHING_BASES_AND_QUALS` since we can detect that here and we would screen those reads out downstream.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
tedsharpe: I was imitating what I saw in other filters that were part of Wellformed.  Cf., ValidAlignmentStartReadFilter, ValidAlignmentEndReadFilter, ReadUtils.alignmentAgreesWithHeader, ReadLengthEqualsCigarLengthReadFilter.
Are you sure you don't want me to accept unmapped reads?
It looks to me as if most every filter that tests alignment criteria accepts them.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
tedsharpe: OK.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
tedsharpe: Just imitating what I saw in WellformedReadFilter.  Is that one immune to inspection?  NB, if I don't allow unmapped reads though the PrimaryLineReadFilter I'll definitely need an OR node.  So if we can't read through AND and OR nodes, we'll still be stuck.

I'm not sure I buy all this command line control of filters, anyway.  The filters specified for the tool are the minimal set, and if you go futzing with them, you'll likely cause an unpleasant exception.  I suspect that this might actually be the case for quite a few tools.  But that's just an aside.

I disagree about adding HAS_MATCHING_BASES_AND_QUALS:  It's not necessary for this tool, and the set of downstream tools that might or might not be run is a mystery.  This tool is a stand-alone from my point of view.  But I defer to your judgement if you'd like to insist.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
tedsharpe: OK.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
tedsharpe: yeah.  especially because the filters could garf up the pairing.

lbergelson:
closing this because I handled the rebase in #2494

PR #2495, opened 2017-03-20
Document how to authenticate to Google Cloud Storage

jean-philippe-martin:
Via a VM, your account, or a service account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,12 +116,12 @@ If you are looking for the codebase of the current production version of GATK, p
   ```
   ./gatk-launch PrintReads -I gs://mybucket/path/to/my.bam -L 1:10000-20000 -O output.bam
   ```
-* For public datasets, that will just work. For private datasets you must set up authentication.
+* You must set up your credentials first. There are three options:
 * Option (a): run in a Google Cloud Engine VM
   * If you are running in a Google VM then your credentials are already in the VM and will be picked up by GATK, you don't need to do anything special.
 * Option (b): use your own account
   * Install [Google Cloud SDK](https://cloud.google.com/sdk/)
-  * Log into you account:
+  * Log into your account:
   ```
   gcloud auth application-default login
   ```
--- a/README.md
+++ b/README.md
@@ -110,6 +110,35 @@ If you are looking for the codebase of the current production version of GATK, p
   ./gatk-launch PrintReadsSpark -I input.bam -O output.bam
   ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -110,6 +110,35 @@ If you are looking for the codebase of the current production version of GATK, p
   ./gatk-launch PrintReadsSpark -I input.bam -O output.bam
   ```
 
+####Running GATK4 with inputs on Google Cloud Storage:
+
+* GATK can read BAM or VCF inputs from a Google Cloud Storage bucket. Just use the "gs://" prefix:
+  ```
+  ./gatk-launch PrintReads -I gs://mybucket/path/to/my.bam -L 1:10000-20000 -O output.bam
+  ```
+* For public datasets, that will just work. For private datasets you must set up authentication.
```
lbergelson: Is this true?  Who gets charged for the data transfer in this case?  The owner of the data?

README.md
```diff
@@ -110,6 +110,35 @@ If you are looking for the codebase of the current production version of GATK, p
...
+* Option (a): run in a Google Cloud Engine VM
+  * If you are running in a Google VM then your credentials are already in the VM and will be picked up by GATK, you don't need to do anything special.
+* Option (b): use your own account
+  * Install [Google Cloud SDK](https://cloud.google.com/sdk/)
+  * Log into you account:
```
droazen: you -> your

README.md
```diff
@@ -110,6 +110,35 @@ If you are looking for the codebase of the current production version of GATK, p
...
+  ```
+  gcloud auth application-default login
```
droazen: If you type `gcloud auth login` (without the 'application-default`), will it still work?

README.md
droazen: Is this true? For public datasets you don't need to be authenticated at all?

droazen:
Back to @jean-philippe-martin , merge after addressing comments

README.md
jean-philippe-martin: done

README.md
jean-philippe-martin: no, it won't work.

README.md
jean-philippe-martin: ah you may be right, let me double-check.

jean-philippe-martin: I was wrong. Technically the user doesn't need to be authenticated if you have an API key, but we aren't using one so the user needs to authenticate.

PR #2500, opened 2017-03-21
gatk-launch launch now sets properties for local jars correctly

### LINKED ISSUE #2316: system properties not applied when running gatk-launch with jar explicitly specified
If gatk-launch is run so that it invokes a local jar directly without the generated launch script, it fails to properly pass system properties.

This will cause confusing bugs and performance issues.  It was discovered in due to #2300.  This effects all users using our packaged jars available on the website since they don't include the gradle generated wrapper script.

lbergelson:
previously system properties were set by the wrapper script but not by gatk-launch when running with a local jar
now both local jar and wrapper script get the same properties
more properties are now explicitely set for spark tools as well
fixes #2316

### DIFF
--- a/gatk-launch
+++ b/gatk-launch
@@ -43,15 +43,26 @@ GATK_LOCAL_JAR_ENV_VARIABLE = "GATK_LOCAL_JAR"
 GATK_SPARK_JAR_ENV_VARIABLE = "GATK_SPARK_JAR"
 BIN_PATH = script + "/build/libs"
 
-EXTRA_JAVA_OPTIONS="-Dsamjdk.compression_level=1 -DGATK_STACKTRACE_ON_USER_EXCEPTION=true "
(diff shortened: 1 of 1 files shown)


gatk-launch
```diff
@@ -145,13 +156,17 @@ def getSparkSubmitCommand(sparkSubmitCommand):
...
     wrapperScript = getGatkWrapperScript(throwIfNotFound=False)
     if wrapperScript is not None:
         return [wrapperScript]
 
-    return ["java", "-jar", getLocalJar()]  # will throw if local jar not found
+    return formatLocalJarCommand(getLocalJar())  # will throw if local jar not found
+
+
+def formatLocalJarCommand(localJarFromEnv):
+    return ["java"] + PACKAGED_LOCAL_JAR_OPTIONS  + [ "-jar", localJarFromEnv]
```
droazen: `localJarFromEnv` should be renamed, since it's not necessarily from the environment. Just `localJar` is probably fine.

droazen:
:+1: merge after addressing comments

PR #2506, opened 2017-03-21
Set channel reopen option and retry options for robustness

jean-philippe-martin:
(these are all cloud-specific)

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
@@ -44,6 +44,9 @@ public final class BucketUtils {
     // slashes omitted since hdfs paths seem to only have 1 slash which would be weirder to include than no slashes
     public static final String FILE_PREFIX = "file:";
 
+    // if the channel errors out, re-open up to this many times
+    public static final int NIO_MAX_REOPENS = 3;
+
 
     public static final Logger logger = LogManager.getLogger("org.broadinstitute.hellbender.utils.gcs");
 
@@ -393,7 +396,7 @@ public final class BucketUtils {
         final String pathWithoutBucket = String.join("/", Arrays.copyOfRange(split, 3, split.length));
         CloudStorageConfiguration cloudConfig = CloudStorageConfiguration.builder()
             // if the channel errors out, re-open up to this many times
-            .maxChannelReopens(3)
+            .maxChannelReopens(NIO_MAX_REOPENS)
             .build();
         StorageOptions sopt = setGenerousTimeouts(StorageOptions.newBuilder()).build();
         return CloudStorageFileSystem.forBucket(BUCKET, cloudConfig, sopt).getPath(pathWithoutBucket);
--- a/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
@@ -35,7 +35,7 @@ import java.util.List;
 import java.util.UUID;
 
 /**
- * Utilities for dealing with google buckets
+ * Utilities for dealing with google buckets.
  */
 public final class BucketUtils {
     public static final String GCS_PREFIX = "gs://";
--- a/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
@@ -1,8 +1,6 @@
 package org.broadinstitute.hellbender.utils.gcs;
 
 import com.google.api.services.storage.Storage;
-import com.google.auth.oauth2.GoogleCredentials;
-import com.google.auth.oauth2.ServiceAccountCredentials;
 import com.google.cloud.HttpTransportOptions;
 import com.google.cloud.RetryParams;
 import com.google.cloud.dataflow.sdk.options.GcsOptions;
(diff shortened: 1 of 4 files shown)


jean-philippe-martin:
The failure is: "ERROR: (gcloud.components.update) The component [bq-nix] failed to download." This looks like a transient issue related to gcloud components update.

FWIW I was able to update my desktop's installation with no issue.

jean-philippe-martin:
I knew it, this just needed a retry.

jean-philippe-martin:
It'd be nice if we could please have a review of this code. The retries fix a real problem, and there isn't much code so it shouldn't be too hard to review.

droazen:
@jean-philippe-martin I'm concerned that this fix does not set retries for all pathways into the NIO code. In particular, what about classes like `ReadsDataSource` that add a prefetcher via `SeekableByteChannelPrefetcher.addPrefetcher()`, or classes in htsjdk that create a `Path` from a raw `String`? Could you comment on this issue before I review? Thanks!

jean-philippe-martin:
@droazen sure. This PR updates both `BucketUtils.getPathOnGcs` and `IOUtils.getPath` (the latter indirectly) to create a Google Cloud Storage filesystem with a default reopen set to 3. Anyone opening the given `Path` (or even a `Path` derived from it, e.g. via `subpath`) will have the retries enabled.

`addPrefetcher` wraps an existing `Path`, so it inherits the underlying `Path`'s retry behavior.

Classes within htsjdk that create a `Path` from a String, without using our utility code, would indeed not set retries and errors on those files would not get the benefit of our extra retry. The exception to that is the Spark code path with `NioBam`: this goes via `ReadsIterable`, which also sets the retries.

My understanding is that the normal way to open BAMs for tools always creates the `Path` object when parsing the command line, as in e.g. `ReadInputArgumentCollection::GetReadIndexPaths()`. That code path that calls `IOUtils.getPath` (and thus sets the retries). I would expect that the added support for VCF follows the same style, though it looks like it doesn't.

If there's a code path that I missed where you think we need retries then please let me know so I can add it!

jean-philippe-martin:
Since the code isn't reviewed yet I took the liberty of adding one much push with a single change: adding a "synchronized" to protect against a potential data race in `SeekableByteChannelPrefetcher`. The contract for `ReadableByteChannel` (which this implements) requires `read` to be thread-safe.

I don't know whether this was the cause of #2516. I haven't been able to reproduce it since, but then again even before this change it wasn't easy to trigger.

jean-philippe-martin:
Edit: that wasn't it, I saw #2516 again last night.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -393,9 +391,28 @@ public static String getPathWithoutBucket(String path) {
         final String[] split = gcsUrl.split("/");
         final String BUCKET = split[2];
         final String pathWithoutBucket = String.join("/", Arrays.copyOfRange(split, 3, split.length));
-        return CloudStorageFileSystem.forBucket(BUCKET).getPath(pathWithoutBucket);
+        CloudStorageConfiguration cloudConfig = CloudStorageConfiguration.builder()
+            // if the channel errors out, re-open up to this many times
+            .maxChannelReopens(3)
```
droazen: Can you extract the number of retries as a `public static final` constant `BucketUtils.GCS_NIO_RETRIES`, and use that constant both here and in `ReadsIterable` below?

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -393,9 +391,28 @@ public static String getPathWithoutBucket(String path) {
...
+                .build())
+            .setRetryParams(RetryParams.newBuilder()
+                .setRetryMaxAttempts(10)
+                .setRetryMinAttempts(6)
+                .setMaxRetryDelayMillis(30000)
+                .setTotalRetryPeriodMillis(120000)
+                .setInitialRetryDelayMillis(250)
+                .build());
+    }
     // TODO(jpmartin): uncomment once shaded.cloud_nio.com.google.auth.Credentials is visible.
```
droazen: The package name fix is still not released, correct?

droazen:
Before merging this, could you have a look at `FeatureDataSource` (in particular, `getFeatureReader()`) to make sure it will also benefit from the new retry settings? This is the code path hit when reading VCFs.

droazen:
Review complete, back to @jean-philippe-martin , squash and merge after addressing comments.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
jean-philippe-martin: Actually there was a new release just two days ago. Updating to the newest gcloud should probably be a separate PR, though.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
jean-philippe-martin: Certainly! I named it NIO_MAX_REOPENS since it also controls how many times we re-open the channel (and normal retries are handled differently).

jean-philippe-martin:
`getFeatureReader` uses `IOUtils.getPath` so it's fine, this sets the retries.

jean-philippe-martin:
The Travis error is:

    The log length has exceeded the limit of 4 MB (this usually means that the test suite is
    raising the same exception over and over).

    The job has been terminated."

Looks like we'll have to make our tests less chatty.

lbergelson:
We're having issues with all of our tests today.  Github is refusing our git-lfs requests because they're over quota and we need to figure out how to either authenticate our requests in a safe way from travis or figure out why we're suddenly going over quota.  It happened very suddenly and I suspect there might be an issue on github's end..

droazen:
Unfortunately the only thing to do for now is to keep hitting "restart" on all the travis builds until they pass..

jean-philippe-martin:
One problem I've noticed is that the log is full of notices from tests we're skipping. The limit on Travis log length is new I think. Removing the mention of each individual skipped may help us have logs that fall below Travis' maximum.

droazen:
The only reason the logs are so long right now is because git-lfs is failing intermittently to download the large files. Under normal circumstances we definitely want to see in the logs when a test is skipped, as there should not be many (or any) skipped tests.

jean-philippe-martin:
I can put this back, but I do think that many tests are skipped by one or the other of our 3 runs. Consider this line from the [log](https://s3.amazonaws.com/archive.travis-ci.org/jobs/219089434/log.txt):

     194479 tests completed, 7 failed, 9729 skipped

droazen:
That is not normal @jean-philippe-martin -- eg., a recent passing build of the unit tests on master (https://travis-ci.org/broadinstitute/gatk/builds/219053956) has just 5 out of 419980 tests skipped:

```
Results: SUCCESS (419980 tests, 419975 successes, 0 failures, 5 skipped)
```

The ~9000+ skips you're seeing are almost certainly due to the intermittent git lfs quota issues -- we've contacted travis (and github) support about this. If you keep re-running the tests they should pass eventually.

jean-philippe-martin:
I see, I was mistaken. Thank you @droazen! I'll retry instead, then.

jean-philippe-martin:
There we go! Thank you!

PR #2510, opened 2017-03-21
adding large example g.vcfs to large files

lbergelson:
* 3 single sample gvcfs covering parts of chr20-21
* combined (but not genotyped) gvcf that's the result of running gatk3 CombineGVCFs on the 3 single sample files
* chr20 / 21 from hg38 to match the gvcfs


lbergelson:
@droazen I had a confusing git-lfs issue caused by the fact that I had added my own pre-commit hook.  This is just adding the files for @kgururaj and @kdatta.

kdatta:
@lbergelson can the git-lfs issue you saw create any problems for us?

lbergelson:
@kdatta It shouldn't cause any problems for you.  It was a local configuration issue.  If your already set up to download the lfs files these should just work.

src/test/resources/large/.gitattributes
```diff
@@ -31,3 +31,8 @@ dbsnp_138.b37.20.21.vcf.blockgz.gz.tbi filter=lfs diff=lfs merge=lfs -text
 tiledb_bin filter=lfs diff=lfs merge=lfs -text
 VQSR/** filter=lfs diff=lfs merge=lfs -text
 human_g1k_v37.20.21.fasta.img filter=lfs diff=lfs merge=lfs -text
+Homo_sapiens_assembly38.20.21.fasta filter=lfs diff=lfs merge=lfs -text
+gvcfs/HG00096.g.vcf.gz filter=lfs diff=lfs merge=lfs -text
+gvcfs/HG00268.g.vcf.gz filter=lfs diff=lfs merge=lfs -text
+gvcfs/NA19625.g.vcf.gz filter=lfs diff=lfs merge=lfs -text
+gvcfs/expected.combined.g.vcf.gz filter=lfs diff=lfs merge=lfs -text
```
droazen: Can you include the origin in the name? (Eg., if it comes from GATK3 include the version of GATK3 that generated it in the name.

droazen:
One comment to address, then merge when ready :+1:

Issue #2502, opened 2017-03-21
SV called variants.vcf ASSEMBLY_IDS and CONTIG_IDS uses % rather than , to separate elements.


vruano:
VCF's standard is to use , to separate elements in INFO and FORMAT annotations. 

The output of the SV variant caller instead of , uses % to separate elements.

cwhelan:
@SHuang-Broad was there a reason you used '%'?

vruano:
This seems to come from this constant in GATKSVVCFHeaderLines.java line 15.

    GATKSVVCFHeaderLines.FORMAT_FIELD_SEPARATOR= "%"

vruano:
The constant name is rather deceptive as it would be strictly speaking ':' I guess it should be called something like FIELD_ELEMENT_SEPARATOR and I guess there is VCF Consants class constant for that anyway.

SHuang-Broad:
Nope, simply an ignorance of that part of the VCF spec, and the fact that `INSERTED_SEQUENCE_MAPPINGS` is using `,` for separating fields of a single mapping.

SHuang-Broad:
ping @vruano for reviewing #2512 that fixes this issue

SHuang-Broad:
closed by #2512

PR #2512, opened 2017-03-22
fix for issue 2502 "SV called variants.vcf ASSEMBLY_IDS and CONTIG_IDS uses % rather than , to separate elements"

SHuang-Broad:
#2502

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
@@ -28,6 +28,7 @@ import java.util.Objects;
 class AlignmentRegion {
 
     static final String STRING_REP_SEPARATOR= "\t";
+    static final String PACKED_STRING_REP_SEPARATOR= "_";
     static final String DUMMY_ASM_ID = "ASSEMBLY";
 
     final String assemblyId;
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
```diff
@@ -120,7 +121,10 @@ int endOfAlignmentInContig() {
      *          Note that the format is NOT the same as that used in {@link #toString()}.
      */
     String toPackedString() {
-        return assemblyId + "-" + contigId + ":" + startInAssembledContig + "-" + endInAssembledContig + ":" + referenceInterval.getContig() + ',' + referenceInterval.getStart() + ',' + (forwardStrand ? '+' : '-') + ',' + TextCigarCodec.encode(cigarAlong5to3DirectionOfContig) + ',' + mapQual + ',' + mismatches;
+        return assemblyId + PACKED_STRING_REP_SEPARATOR + contigId + PACKED_STRING_REP_SEPARATOR +
```
vruano: Use this instead:

```
String.join(PACKED_STRING_REP_SEPARATOR, assemblyId, contigId, startInAssembledContig, ... mapQual, mismatches);
```
For those elements that are not ```CharSequence```s you might need to prepend ```"" + ``` or call ```String.valueOf(x)```.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/NovelAdjacencyReferenceLocationsUnitTest.java
```diff
@@ -158,7 +158,7 @@ public void testGetAssembledBreakpointsFromAlignmentRegionsWithOverlappingAlignm
         Assert.assertEquals(chimericAlignment.regionWithLowerCoordOnContig, region1);
         Assert.assertEquals(chimericAlignment.regionWithHigherCoordOnContig, region3);
         Assert.assertEquals(chimericAlignment.insertionMappings.size(), 1);
-        Assert.assertEquals(chimericAlignment.insertionMappings.get(0), "1-contig-1:484-525:20,23103196,-,483S42M968S,60,2");
+        Assert.assertEquals(chimericAlignment.insertionMappings.get(0), "1_contig-1_484_525_20_23103196_-_483S42M968S_60_2");
```
vruano: Perhaps you should construct the "actual" string using the constant PACKED_STRING_REP_SEPARATOR so that this test don't fail if you happen to change that constant value.

vruano:
Review done. Please change the commit/pull-request description so that it explains what has been changed; pointing to the issue is good but one should be able to see what has been done just from doing a 'git log' on it.
Back to @SHuang-Broad.

SHuang-Broad:
Addressed the comments. Please take a second look @vruano

vruano:
Ok good to go.

PR #2515, opened 2017-03-22
Don't count overlapping read pairs as discordant read pairs

cwhelan:
The current FindBreakpointEvidence code is classifying reads pairs that overlap such that the start position of the reverse read is before the start position of the forward read as "OutiesPair" discordant read pair evidence. However, these are likely due to sequencing of very short inserts that causes some of the adapter to be sequenced and potentially aligned. This change requires a read pair to not be overlapping to be counted as an 'OutiesPair'.

On the CHM dataset this causes the number of intervals discovered to drop from 23152 to 21633, and the number of called variants to drop from 3467 to 3366. 

@tedsharpe could you review?


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadClassifier.java
```diff
@@ -113,8 +113,9 @@ private void checkDiscordantPair( final GATKRead read, final List<BreakpointEvid
             evidenceList.add(new BreakpointEvidence.InterContigPair(read, readMetadata));
         } else if ( read.isReverseStrand() == read.mateIsReverseStrand() ) {
             evidenceList.add(new BreakpointEvidence.SameStrandPair(read, readMetadata));
-        } else if ( (read.getStart() < read.getMateStart() && read.isReverseStrand()) ||
-                    (read.getStart() > read.getMateStart() && !read.isReverseStrand()) ) {
+        } else if ( (read.getEnd() < read.getMateStart() && read.isReverseStrand()) ||
+                    (read.getStart() > read.getMateStart() + read.getLength() && !read.isReverseStrand()) ) {
```
tedsharpe: I understand the problem (lack of availability of the mate's cigar for the downstream read), but the asymmetry of the two cases bugs me a little.

In the case of templates shorter than a read I'd expect the mates' starting positions to be very nearly identical (maybe not exactly, because the adapter sequence might accidentally match reference for a base or two).

We don't actually need disjoint, we just need one read to be *clearly* upstream or *clearly* downstream of the other one.  (If they're close, it's probably just a short template.)

I think we could eliminate the short template case by just coding a fudge factor:
```   } else if ( (read.getStart()+10 < read.getMateStart() && read.isReverseStrand()) ||
                   (read.getStart()-10 > read.getMateStart() && !read.isReverseStrand()) ) {```

If you don't like that idea, how about using the reference length of the (wrong) cigar, rather than entire read length?
```(read.getStart() > read.getMateStart()+read.getCigar().getReferenceLength() && !read.isReverseStrand())```
The total read length seems too strict to me.

If you don't like that idea either, just merge it.  I'm fine with it.  I think we might occasionally miss evidence of small inversions this way, but to tell you the truth, I'm struggling to figure out whether that's really true or not.

cwhelan: @tedsharpe  I'd been loath to do that (add some sort of short allowable overhang between pairs) since I didn't feel like adding yet another hard-coded magic number. But, after further consideration, I think you're right, and setting it to 10 seems to get close to, but not quite to, the same effect on the number of intervals called, so we're likely still filtering out the bad stuff but maybe keeping a little good stuff that we weren't before. I've committed the change and will merge when tests pass.

PR #2518, opened 2017-03-22
Fixed javadoc. Closes #1945.

### LINKED ISSUE #1945: Documentation and parameter name for AFCalculatorProvider disagree
For abstract method

```
/**
 * Returns a AF calculator given the required homogeneous ploidy and allele count (including the reference).
 * @param ploidy the required ploidy.
 * @param maximumAltAlleles the allele count.
 * @return never {@code null}
 */
 public abstract AFCalculator getInstance(final int ploidy, final int maximumAltAlleles);
```

which is the intended? should the parameter name be changed or the other way?

davidbenjamin:
@droazen here you go.


lbergelson:
👍

PR #2527, opened 2017-03-23
Refer alt contigs to original molecule when finding cross-contig pairs.

tedsharpe:


cwhelan:
I'd like to test this out. @tedsharpe could you add your "contigNameToMoleculeName" file to the /broad-dsde-methods/sv/reference/GRCh38 bucket, perhaps along with a little README note on how you created it? It'd also be nice if you could update the scripts in scripts/sv in github (I think you'd have to change svDiscover.sh and scanBam.sh) to correctly pass this along.

tedsharpe:
@cwhelan  Ready to test.  I copied the file to GCS, and forced-pushed an update to the branch to fix the scripts and to allow reading from GCS.

cwhelan:
My initial test with this doesn't seem to show much of an effect. Do you have some example locations of places where we're getting these bad clusters of Main<->Alt contig read pairs?

Issue #1730, opened 2016-04-19
GenotypeAnnotations require GT, but some fields could be computed without that information


magicDGS:
When I'm trying to use some `GenotypeAnnotation`s, like `DepthPerAlleleBySample`, I found that the method does not return the AD field if the provided genotype (not the builder) is null or not called. But this annotation does not requires that the genotype is called for being computed (if some reads spans the variant, it does not need that the genotype is called to compute the depths).

I would like to have these annotations that does not required the genotype to be computed to annotate genotypes without the GT field, and it will be great if I can re-use your code instead of implement `GenotypeAnnotation` by my own.

Could it be possible to change the `GenotypeAnnotation` implementations to handle these cases? If not, it will be very helpful to have some documentation in the `annotate` method for each implementation that describe what happen when the GT is not called, or the `g` parameter is set to `null`.

Thank you in advance for your help.

PR #2534, opened 2017-03-27
GenotypeAnnotations without called genotype constraint

magicDGS:
This is a fix for #1730, which includes:

* New method param in` VariantAnnotatorEngine` to filter genotypes to annotate (first commit). Previous method signature is maintained using a filter for called genotypes. 
* Remove the `!g.isCalled()` check in all the `GenotypeAnnotation` from GATK (second commit), to allow annotation of non-called genotypes. All of the genotype annotations does not require this constraint, and in my personal usage I would need to use annotations from GATK with non-called genotypes.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
@@ -163,18 +163,20 @@ public final class VariantAnnotatorEngine {
      * @param ref the reference context of the variant to annotate or null if there is none
      * @param likelihoods likelihoods indexed by sample, allele, and read within sample. May be null
      * @param addAnnot function that indicates if the given annotation type should be added to the variant
+     * @param annotGenotype function that indicates if the given genotype should be annotated
      */
     public VariantContext annotateContext(final VariantContext vc,
                                           final FeatureContext features,
                                           final ReferenceContext ref,
                                           final ReadLikelihoods<Allele> likelihoods,
-                                          final Predicate<VariantAnnotation> addAnnot) {
+                                          final Predicate<VariantAnnotation> addAnnot,
+                                          final Predicate<Genotype> annotGenotype) {
         Utils.nonNull(vc, "vc cannot be null");
         Utils.nonNull(features, "features cannot be null");
 
         // annotate genotypes, creating another new VC in the process
         final VariantContextBuilder builder = new VariantContextBuilder(vc);
-        builder.genotypes(annotateGenotypes(ref, vc, likelihoods, addAnnot));
+        builder.genotypes(annotateGenotypes(ref, vc, likelihoods, addAnnot, annotGenotype));
         final VariantContext newGenotypeAnnotatedVC = builder.make();
 
         final Map<String, Object> infoAnnotMap = new LinkedHashMap<>(newGenotypeAnnotatedVC.getAttributes());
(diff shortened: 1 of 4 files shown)


droazen:
@magicDGS Since we're currently trying to get our annotations to match GATK3 as closely as possible, this is not a change I'd be comfortable making at this time. We can revisit after our beta release in May and consider this for merge at that time.

magicDGS:
Thanks for your feedback, @droazen. I think that it will be nice to have a better annotator engine for handling what should be on/off in which cases instead of hardcoded them when it is necessary.

But I can wait til the release.

droazen:
Assigning to @ldgauthier for an eventual review (although we should not merge this until the `HaplotypeCaller` tie-out is complete.

ldgauthier:
@magicDGS Thanks for the change.  You make a good point, especially since we genotype samples with ambiguous PLs as no-calls when there is data to annotate.

This looks good to me, but it seems that it didn't change any of the tests.  Can you add a test somewhere that shows no-call output with annotations?  For example, from some real data I've seen a genotype such that the VCF has
`1       13418   .       G       A,<NON_REF>     6402.35 . [INFO annotations] GT:AD:DP:GQ:PGT:PID:PL   ./.:9,0,0:9:.:.:.:0,0,0,0,0,0`, which has 9 reference reads you could output strand orientation for.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
```diff
@@ -172,10 +174,24 @@ public VariantContext annotateContext(final VariantContext vc,
...
+     * Annotates the given variant context for called Genotypes - adds all annotations that satisfy the predicate.
+     *
+     * @see #annotateContext(VariantContext, FeatureContext, ReferenceContext, ReadLikelihoods, Predicate, Predicate)
+     */
+    public VariantContext annotateContext(final VariantContext vc,
+                                          final FeatureContext features,
+                                          final ReferenceContext ref,
+                                          final ReadLikelihoods<Allele> likelihoods,
+                                          final Predicate<VariantAnnotation> addAnnot) {
+        return annotateContext(vc, features, ref, likelihoods, addAnnot, Genotype::isCalled);
```
magicDGS: The `Genotype::isCalled` is removing the annotation for no calls by default. I kept the method with the same signature to do not change the behaviour of the tools such as HC, which might prevent the change to get in soon.

If you are ok with the change for all the tools, I can remove this method and use the one with a predicate for the genotypes that does not filter anything.

ldgauthier:
I don't understand -- if the behavior doesn't change then how is this a fix for issue #1730?  Are you using the GenotypeAnnotations outside of the VariantAnnotatorEngine?

magicDGS:
Yes, I am using it in the development of a downstream project - to implement my own annotations and also use the GATK ones to annotate a different set of variants (without called genotypes). That's why I though that it is maybe better if there is a finer control.

Of course, GATK can benefit from this by adding an option to annotate uncalled genotypes instead of the current behaviour. I always try to minimize the impact on the current behaviour to let the GATK team decide what is better for your purposes...

ldgauthier:
Given that this helps you out, but doesn't change the behavior of our tools, it's pretty much a refactor from our end and I'm happy to give it a 👍

magicDGS:
Thanks @ldgauthier - I rebased into the latest master to solve the conflicts. Back to you!

ldgauthier:
@droazen I'm good with this. Are you okay merging this given that it won't change anything for us?

magicDGS:
Friendly ping here, @droazen!

droazen:
Closing this one as abandoned.

PR #2540, opened 2017-03-29
Tool to read data in parallel from GCS to HDFS with Spark

cwhelan:
This tool copies single large files or directories from GCS into HDFS using Spark. Spark parallelization allows each task to copy a chunk in the size of the blocks of the target HDFS system simultaneously.

When copying a directory containing a 120GB WGS bam and its index, this takes approximately 1 minute on a 10 worker / 160 core cluster, as opposed to approximately 20 minutes using Hadoop distcp.

This may eventually be superseded by the NIO GCS integration work if that ends up performing comparably.

@lbergelson would you like to review? Or feel free to nominate someone else.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
@@ -1,0 +1,204 @@
+package org.broadinstitute.hellbender.tools.spark;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.spark.api.java.JavaPairRDD;
+import org.apache.spark.api.java.JavaSparkContext;
+import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
+import org.broadinstitute.hellbender.cmdline.programgroups.SparkProgramGroup;
+import org.broadinstitute.hellbender.engine.spark.GATKSparkTool;
+import org.broadinstitute.hellbender.exceptions.GATKException;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
+import org.broadinstitute.hellbender.utils.io.IOUtils;
+import scala.Tuple2;
+
+import java.io.*;
+import java.nio.ByteBuffer;
+import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * This tool uses Spark to do a parallel copy of either a file or a directory from GCS into HDFS.
+ * Files are divided into chunks of size equal to the HDFS block size (with the exception of the final
+ * chunk) and each Spark task is responsible for copying one chunk. To copy all of the files in a GCS directory,
+ * provide the GCS directory path, including the trailing slash. Directory copies are non-recursive so
+ * subdirectories will be skipped. Within directories each file is divided into chunks independently (so this will be
+ * inefficient if you have lots of files smaller than the block size). After all chunks are copied, the HDFS
+ * concat method is used to stitch together chunks into single files without re-copying them.
+ */
+@CommandLineProgramProperties(summary="Parallel copy a file or directory (non-recursive) from GCS into HDFS",
+        oneLineSummary="Parallel copy a file or directory from GCS into HDFS.",
+        programGroup = SparkProgramGroup.class)
+public class ParallelCopyGCSDirectoryIntoHDFSSpark extends GATKSparkTool {
+    private static final long serialVersionUID = 1L;
+
+    // default buffer size for reading chunks is 64MiB based on performance profiling and what appears to be conventional
+    // wisdom to use a power of two for byte buffer sizes
+    public static final int SIXTY_FOUR_MIB = 67108864;
+
+    @Argument(doc = "input GCS file path (add trailing slash when specifying a directory)", fullName = "inputGCSPath")
+    private String inputGCSPath = null;
+
+    @Argument(doc = "output directory on HDFS", shortName = "outputHDFSDirectory",
+            fullName = "outputHDFSDirectory")
+    private String outputHDFSDirectory;
+
+    @Override
+    protected void runTool(final JavaSparkContext ctx) {
+
+        if (! BucketUtils.isCloudStorageUrl(inputGCSPath)) {
+            throw new UserException("Input path "+ inputGCSPath + " is not a GCS URI");
+        }
+
+        if (! BucketUtils.isHadoopUrl(outputHDFSDirectory)) {
+            throw new UserException("Output directory " + outputHDFSDirectory + " is not an HDFS URI");
+        }
+
+        final String inputGCSPathFinal = inputGCSPath;
+        final String outputDirectoryFinal = outputHDFSDirectory;
+
+        org.apache.hadoop.fs.Path outputHdfsDirectoryPath = new org.apache.hadoop.fs.Path(outputHDFSDirectory);
+
+        try(FileSystem fs = outputHdfsDirectoryPath.getFileSystem(new Configuration())) {
+
+            if (fs.exists(outputHdfsDirectoryPath)) {
+                throw new UserException("Specified output directory " + outputHdfsDirectoryPath + " already exists. Please specify a new directory name.");
+            }
+            fs.mkdirs(outputHdfsDirectoryPath);
+
+            final long chunkSize = getChunkSize(fs);
+
+            final List<Path> gcsNIOPaths = getGCSFilePathsToCopy(inputGCSPathFinal);
+
+            List<Tuple2<String, Integer>> chunkList = setupChunks(chunkSize, gcsNIOPaths);
+
+            if (chunkList.size() == 0) {
+                logger.info("no files found to copy");
+                return;
+            }
+
+            final JavaPairRDD<String, Integer> chunkRDD = ctx.parallelizePairs(chunkList, chunkList.size());
+            final JavaPairRDD<String, Tuple2<Integer, String>> chunkMappingRDD =
+                    chunkRDD.mapToPair(p -> new Tuple2<>(p._1(), readChunkToHdfs(p._1(), chunkSize, p._2(), outputDirectoryFinal)));
+            final Map<String, Iterable<Tuple2<Integer, String>>> chunksByFilePath = chunkMappingRDD.groupByKey().collectAsMap();
+
+            concatenateChunks(outputDirectoryFinal, fs, gcsNIOPaths, chunksByFilePath);
+
+        } catch (NoSuchFileException e) {
+            throw new UserException("Could not locate input path " + e.getFile() + ". If you are trying to copy an entire directory, please include a trailing slash on your path.");
+
+        } catch (IOException e) {
+            throw new GATKException(e.getMessage(), e);
+        }
+
+    }
+
+    private void concatenateChunks(final String outputDirectoryFinal, final FileSystem fs, final List<Path> gcsNIOPaths, final Map<String, Iterable<Tuple2<Integer, String>>> chunksByFilePath) throws IOException {
+        for (Path path : gcsNIOPaths) {
+
+            if (Files.isDirectory(path)) {
+                continue;
+            }
+
+            final String filePath = path.toUri().toString();
+            final Iterable<Tuple2<Integer, String>> chunkListForFile = chunksByFilePath.get(filePath);
+            final String basename = path.getName(path.getNameCount() - 1).toString();
+            final org.apache.hadoop.fs.Path outFilePath = new org.apache.hadoop.fs.Path(outputDirectoryFinal + "/" + basename);
+            fs.createNewFile(outFilePath);
+
+            SortedMap<Integer, String> chunkMap = new TreeMap<>();
+            for (Tuple2<Integer, String> entry : chunkListForFile) {
+                chunkMap.put(entry._1(), entry._2());
+            }
+
+            org.apache.hadoop.fs.Path[] chunkPaths = new org.apache.hadoop.fs.Path[chunkMap.size()];
+
+            final Iterator<Integer> iterator = chunkMap.keySet().iterator();
+            while (iterator.hasNext()) {
+                final Integer next = iterator.next();
+                final String chunkPath = chunkMap.get(next);
+                chunkPaths[next] = new org.apache.hadoop.fs.Path(chunkPath);
+            }
+
+            fs.concat(outFilePath, chunkPaths);
+        }
+    }
+
+    private List<Tuple2<String, Integer>> setupChunks(final long chunkSize, final List<Path> gcsNIOPaths) throws IOException {
+        List<Tuple2<String, Integer>> chunkList = new ArrayList<>();
+        for (Path path : gcsNIOPaths) {
+
+            if (Files.isDirectory(path)) {
+                logger.info("skipping directory " + path);
+                continue;
+            }
+
+            final long fileSize = Files.size(path);
+            final long chunks = fileSize / chunkSize + (fileSize % chunkSize == 0 ? 0 : 1);
+            logger.info("processing path " + path + ", size = " + fileSize + ", chunks = " + chunks);
+
+            for (int i = 0; i < chunks; i++) {
+                chunkList.add(new Tuple2<>(path.toUri().toString(), i));
+            }
+        }
+        return chunkList;
+    }
+
+    private List<Path> getGCSFilePathsToCopy(final String inputGCSPathFinal) throws IOException {
+        final List<Path> gcsNIOPaths;
+        final Path inputGCSNIOPath = IOUtils.getPath(inputGCSPathFinal);
+        if (Files.isDirectory(inputGCSNIOPath)) {
+            logger.info("transferring input directory: " + inputGCSPathFinal);
+            gcsNIOPaths = Files.list(inputGCSNIOPath).collect(Collectors.toList());
+        } else {
+            logger.info("transferring single file: " + inputGCSNIOPath);
+            gcsNIOPaths = Collections.singletonList(inputGCSNIOPath);
+        }
+        return gcsNIOPaths;
+    }
+
+    static long getChunkSize(final FileSystem fs) {
+        return Long.parseLong(fs.getConf().get("dfs.blocksize"));
+    }
+
+    private static final Tuple2<Integer, String> readChunkToHdfs(final String inputGCSPathFinal, final long chunkSize, final Integer chunkNum, final String outputDirectory) {
+        final Path gcsPath = IOUtils.getPath(inputGCSPathFinal);
+        final String basename = gcsPath.getName(gcsPath.getNameCount() - 1).toString();
+        org.apache.hadoop.fs.Path outputPath = new org.apache.hadoop.fs.Path(outputDirectory);
+        final String chunkPath = outputPath + "/" + basename + ".chunk." + chunkNum;
+
+        try (SeekableByteChannel channel = Files.newByteChannel(gcsPath);
+             final OutputStream outputStream = new BufferedOutputStream(BucketUtils.createNonGCSFile(chunkPath))){
+
+            final long start = chunkSize * (long) chunkNum;
+            channel.position(start);
+            ByteBuffer byteBuffer = ByteBuffer.allocateDirect((int) Math.min(SIXTY_FOUR_MIB, chunkSize));
+            long bytesRead = 0;
+            while(channel.read(byteBuffer) > 0) {
+                byteBuffer.flip();
+                while (byteBuffer.hasRemaining() && bytesRead < chunkSize) {
+                    byte b = byteBuffer.get();
+                    outputStream.write(b);
+                    bytesRead++;
+                }
+                if (bytesRead == chunkSize) {
+                    break;
+                }
+                if (bytesRead > chunkSize) {
+                    throw new GATKException("Encountered an unknown error condition and read too many bytes; output file may be corrupt");
+                }
+                byteBuffer.clear();
+            }
+        } catch (IOException e) {
+            throw new GATKException(e.getMessage() + "; inputGCSPathFinal = " + inputGCSPathFinal, e);
+        }
+        return new Tuple2<>(chunkNum, chunkPath);
+    }
+
+}
--- a/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
@@ -390,7 +390,8 @@ public final class BucketUtils {
      * To transform other types of string URLs into Paths, use IOUtils.getPath instead.
      */
     public static java.nio.file.Path getPathOnGcs(String gcsUrl) {
-        final String[] split = gcsUrl.split("/");
+        // use a split limit of -1 to preserve empty split tokens, especially trailing slashes on directory names
+        final String[] split = gcsUrl.split("/", -1);
         final String BUCKET = split[2];
         final String pathWithoutBucket = String.join("/", Arrays.copyOfRange(split, 3, split.length));
         return CloudStorageFileSystem.forBucket(BUCKET).getPath(pathWithoutBucket);
--- a/src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
@@ -1,0 +1,121 @@
+package org.broadinstitute.hellbender.tools.spark;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.*;
+import org.apache.hadoop.hdfs.MiniDFSCluster;
+import org.broadinstitute.hellbender.CommandLineProgramTest;
+import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
+import org.broadinstitute.hellbender.utils.io.IOUtils;
+import org.broadinstitute.hellbender.utils.test.ArgumentsBuilder;
+import org.broadinstitute.hellbender.utils.test.IntegrationTestSpec;
+import org.broadinstitute.hellbender.utils.test.MiniClusterUtils;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.nio.file.Files;
+import java.util.Collections;
+
+
+public class ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest extends CommandLineProgramTest {
+
+    @Override
+    public String getTestedToolName() {
+        return ParallelCopyGCSDirectoryIntoHDFSSpark.class.getSimpleName();
+    }
+
+    @Test(groups = {"spark", "bucket"})
+    public void testCopyLargeFile() throws Exception {
+        MiniDFSCluster cluster = null;
+        try {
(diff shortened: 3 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -390,7 +390,7 @@ public static String getPathWithoutBucket(String path) {
      * To transform other types of string URLs into Paths, use IOUtils.getPath instead.
      */
     public static java.nio.file.Path getPathOnGcs(String gcsUrl) {
-        final String[] split = gcsUrl.split("/");
+        final String[] split = gcsUrl.split("/", 4);
```
droazen: So you decided to stick with `4` here rather than `-1`?

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+    private MiniDFSCluster cluster;
+
+    @BeforeClass(alwaysRun = true)
+    private void setupMiniCluster() throws IOException {
+        cluster = MiniClusterUtils.getMiniCluster();
+    }
+
+    @AfterClass(alwaysRun = true)
+    private void shutdownMiniCluster() {
+        MiniClusterUtils.stopCluster(cluster);
```
droazen: Can you create the mini cluster within your one test case, instead of using `@BeforeClass` and `@AfterClass` like this?

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public String getTestedToolName() {
+        return ParallelCopyGCSDirectoryIntoHDFSSpark.class.getSimpleName();
+    }
+
+    @Test(groups = {"spark", "bucket"})
+    public void testCopyFile() throws Exception {
+        // copy a multi-block file
+        final Path tempPath = MiniClusterUtils.getTempPath(cluster, "test", "dir");
+        final String gcpInputPath = getGCPTestInputPath() + "large/dbsnp_138.b37.1.1-65M.vcf";
```
droazen: This input is ~250 mb -- what's the runtime of this test?

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+        String args =
+                "--inputGCSPath " + gcpInputPath +
+                        " --apiKey " + getGCPTestApiKey() +
```
droazen: Should not need an `--apiKey` argument here after following my suggestions above.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+                        " --outputHDFSDirectory " + tempPath;
+        ArgumentsBuilder ab = new ArgumentsBuilder().add(args);
+        IntegrationTestSpec spec = new IntegrationTestSpec(
+                ab.getString(),
+                Collections.emptyList());
+        spec.executeTest("testCopyFile-" + args, this);
+
+        final String hdfsPath = tempPath + "/" + "dbsnp_138.b37.1.1-65M.vcf";
+        Assert.assertEquals(BucketUtils.fileSize(hdfsPath, getAuthenticatedPipelineOptions()),
+                BucketUtils.fileSize(gcpInputPath, getAuthenticatedPipelineOptions()));
```
droazen: Try to refactor out these usages of `getAuthenticatedPipelineOptions()` here and below if you can -- should be possible using a combination of NIO and, eg., `BucketUtils.fileSize()` with a null second argument.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+        BucketUtils.copyFile(hdfsPath, getAuthenticatedPipelineOptions(), publicTestDir + "fileFromHDFS.vcf");
+        BucketUtils.copyFile(gcpInputPath, getAuthenticatedPipelineOptions(), publicTestDir + "fileFromGCS.vcf");
```
droazen: Copy into temp locations marked for deletion on exit obtained via `BaseTest.createTempFile()`, not into `publicTestDir`. As it stands, this test doesn't clean up after itself.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+        MiniClusterUtils.stopCluster(cluster);
+    }
+
+    @Override
+    public String getTestedToolName() {
+        return ParallelCopyGCSDirectoryIntoHDFSSpark.class.getSimpleName();
+    }
+
+    @Test(groups = {"spark", "bucket"})
+    public void testCopyFile() throws Exception {
```
droazen: Also would be good to assert within the test that the file gets divided into more than 1 chunk (otherwise it's not much of a test!)

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Argument(doc = "output directory on HDFS", shortName = "outputHDFSDirectory",
+            fullName = "outputHDFSDirectory")
+    private String outputHDFSDirectory;
+
+
+    @Override
+    protected void runTool(final JavaSparkContext ctx) {
+
+        final AuthHolder auth = getAuthHolder();
```
cwhelan: Done.. I'd first written this code without fully using the NIO API and this was a holdover from that.

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+        final String inputGCSPathFinal = inputGCSPath;
```
cwhelan: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+        final String outputDirectoryFinal = outputHDFSDirectory;
```
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+
+        org.apache.hadoop.fs.Path file = new org.apache.hadoop.fs.Path(outputHDFSDirectory);
```
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+
+        try(FileSystem fs = file.getFileSystem(new Configuration())){
+
+            fs.mkdirs(file);
+            final long chunkSize = Long.parseLong(fs.getConf().get("dfs.blocksize"));
+
+            final List<Path> gcsPaths;
```
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+            final Path inputGCSNIOPath = IOUtils.getPath(inputGCSPathFinal);
+            if (Files.isDirectory(inputGCSNIOPath)) {
+                logger.warn("directory: " + inputGCSPathFinal);
```
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+                gcsPaths = Files.list(inputGCSNIOPath).collect(Collectors.toList());
+            } else {
+                logger.warn("single file");
```
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+            List<Tuple2<String, Integer>> chunkList = new ArrayList<>();
+            for (Path path : gcsPaths) {
+
+                if (Files.isDirectory(path)) {
+                    logger.info("skipping directory " + path);
+                    continue;
+                }
+
+                final long size = Files.size(path);
+                final long chunks = size / chunkSize + 1;
```
cwhelan: Nice catch, added an extra check to see if the extra block is actually needed.

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+
+            if (chunkList.size() == 0) {
+                logger.info("no files found to copy");
+                return;
+            }
+
+            final JavaPairRDD<String, Integer> chunkRDD = ctx.parallelizePairs(chunkList, chunkList.size());
+
+            final JavaPairRDD<String, Tuple2<Integer, String>> chunkMappingRDD =
+                    chunkRDD.mapToPair(p -> new Tuple2<>(p._1(), readChunkToHdfs(auth, p._1(), chunkSize, p._2(), outputDirectoryFinal)));
```
cwhelan: I'll add something, although you also get the Spark stage progress bar for chunk progress.

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    private static final Tuple2<Integer, String> readChunkToHdfs(final AuthHolder authHolder, final String inputGCSPathFinal, final long chunkSize, final Integer chunkNum, final String outputDirectory) {
+        final Path gcsPath = IOUtils.getPath(inputGCSPathFinal);
+        final String basename = gcsPath.getName(gcsPath.getNameCount() - 1).toString();
+        org.apache.hadoop.fs.Path outputPath = new org.apache.hadoop.fs.Path(outputDirectory);
+        final String chunkPath = outputPath + "/" + basename + ".chunk." + chunkNum;
+
+        try (SeekableByteChannel channel = Files.newByteChannel(gcsPath);
+             final OutputStream outputStream = new BufferedOutputStream(BucketUtils.createFile(chunkPath, authHolder))){
```
cwhelan: done

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
cwhelan: Oops, sorry. Will fix.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
cwhelan: done

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
cwhelan: Removed.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
cwhelan: It takes about a minute on my laptop. I wanted to get a file that's at least 2 chunks, which are 128MB on dataproc and our test clusters. I'm open to any suggestions about better ways to test this.

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+            ByteBuffer byteBuffer = ByteBuffer.allocate(bufferSize);
+            long bytesRead = 0;
+            while(channel.read(byteBuffer) > 0) {
+                byteBuffer.flip();
+                while (byteBuffer.hasRemaining() && bytesRead < chunkSize) {
+                    byte b = byteBuffer.get();
+                    outputStream.write(b);
+                    bytesRead++;
+                }
+                if (bytesRead >= chunkSize) {
```
cwhelan: Added a throw to guard against that case.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
cwhelan: Sorry, I just hadn't pushed that change from 4 to -1 up yet. I'll push it with the rest of the PR-comment-related-fixes.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
cwhelan: The longest part of the time spent is when I do the non-parallel copy from the bucket to the temp dir so that I can compare file contents. Maybe I could skip doing that.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
cwhelan: Added this assertion.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
cwhelan: Removed calls to `getAuthenticatedPipelineOptions` from the test.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
lbergelson: @cwhelan I'm pretty sure you can configure the mini-cluster to use smaller block sizes.  That might let you use a smaller file but still have multiple chunks.

lbergelson: Hmn, if you wanted to do that you might have to make a change to minicluster utils to expose the conf so you can change settings.

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+        final String basename = gcsPath.getName(gcsPath.getNameCount() - 1).toString();
+        org.apache.hadoop.fs.Path outputPath = new org.apache.hadoop.fs.Path(outputDirectory);
+        final String chunkPath = outputPath + "/" + basename + ".chunk." + chunkNum;
+
+        try (SeekableByteChannel channel = Files.newByteChannel(gcsPath);
+             final OutputStream outputStream = new BufferedOutputStream(BucketUtils.createFile(chunkPath, authHolder))){
+
+            final long start = chunkSize * (long) chunkNum;
+            channel.position(start);
+            final int bufferSize = 512;
```
cwhelan: I had no idea how to set the buffer size and was just following online examples which tended to have small buffer sizes. After some experimenting, it looks like setting the buffer size high, and (crucially) using `ByteBuffer.allocateDirect` results in the best performance. There's a lot of variability in timing from run to run as I keep trying it, so I'm just going to set it to the min of 64MB and the chunk size for now.

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+                    chunkPaths[next] = new org.apache.hadoop.fs.Path(chunkPath);
+                }
+
+                fs.concat(outFilePath, chunkPaths);
+            }
+
+
+
+        } catch (IOException e) {
+            throw new GATKException(e.getMessage(), e);
```
cwhelan: I added an extra catch for when the GCS input path can't be found, with a reminder to use a trailing slash if you want to copy a directory (which is how I seem to mess this up a lot).

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+
+        final AuthHolder auth = getAuthHolder();
+        final String inputGCSPathFinal = inputGCSPath;
+        final String outputDirectoryFinal = outputHDFSDirectory;
+
+        org.apache.hadoop.fs.Path file = new org.apache.hadoop.fs.Path(outputHDFSDirectory);
+
+        try(FileSystem fs = file.getFileSystem(new Configuration())){
+
+            fs.mkdirs(file);
```
cwhelan: I added a UserException if the directory already exists. Otherwise this tool would overwrite any files within the directory with the same name as an incoming file.

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
cwhelan: Actually, it's pretty hard to add a logger within a spark task that displays in a reliable way. I think I'm going to suggest we just rely on the Spark progress bar for now.

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+    private String inputGCSPath = null;
+
+
+    @Argument(doc = "output directory on HDFS", shortName = "outputHDFSDirectory",
+            fullName = "outputHDFSDirectory")
+    private String outputHDFSDirectory;
+
+
+    @Override
+    protected void runTool(final JavaSparkContext ctx) {
```
cwhelan: done, refactored into a few methods

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,171 @@
...
+            throw new GATKException(e.getMessage(), e);
+        }
+
+    }
+
+    private static final Tuple2<Integer, String> readChunkToHdfs(final AuthHolder authHolder, final String inputGCSPathFinal, final long chunkSize, final Integer chunkNum, final String outputDirectory) {
+        final Path gcsPath = IOUtils.getPath(inputGCSPathFinal);
+        final String basename = gcsPath.getName(gcsPath.getNameCount() - 1).toString();
+        org.apache.hadoop.fs.Path outputPath = new org.apache.hadoop.fs.Path(outputDirectory);
+        final String chunkPath = outputPath + "/" + basename + ".chunk." + chunkNum;
```
cwhelan: The Hadoop Path object seems to strip trailing slashes (see `org.apache.hadoop.fs.Path.normalizePath`). Running with a trailing slash doesn't seem to cause any problems.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
cwhelan: I modified `MiniClusterUtils.getMiniCluster` to optionally take a configuration as a parameter, which seems to work. The test now transfers a 2MB file in 2 chunks and runs in about 20 seconds on my laptop.

cwhelan:
@droazen thanks for the review. I've tried to address all of your comments. Back to you!

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
```diff
@@ -0,0 +1,200 @@
...
+
+        try (SeekableByteChannel channel = Files.newByteChannel(gcsPath);
+             final OutputStream outputStream = new BufferedOutputStream(BucketUtils.createNonGCSFile(chunkPath))){
+
+            final long start = chunkSize * (long) chunkNum;
+            channel.position(start);
+            ByteBuffer byteBuffer = ByteBuffer.allocateDirect((int) Math.min(67108864, chunkSize));
```
droazen: Document where this magic value `67108864` comes from, and extract as a named constant.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -390,7 +390,7 @@ public static String getPathWithoutBucket(String path) {
      * To transform other types of string URLs into Paths, use IOUtils.getPath instead.
      */
     public static java.nio.file.Path getPathOnGcs(String gcsUrl) {
-        final String[] split = gcsUrl.split("/");
+        final String[] split = gcsUrl.split("/", -1);
```
droazen: Add comment explaining why we pass in -1, just in case posterity is puzzled by this.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
...
+        MiniDFSCluster cluster = null;
+        try {
+            final Configuration conf = new Configuration();
+            // set the minicluster to have a very low block size so that we can test transfering a file in chunks without actually needing to move a big file
+            conf.set("dfs.blocksize", "1048576");
```
droazen: How many chunks does this result in? Ideally it would be at least 3 or 4 chunks for a good test.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+            final File tempDir = createTempDir("ParallelCopy");
+
+            BucketUtils.copyFile(hdfsPath, null, tempDir + "fileFromHDFS.vcf");
+            Assert.assertEquals(Utils.calculateFileMD5(new File(tempDir + "fileFromHDFS.vcf")), "7e0204f91fdafcd6b13b6907537f3045");
+        } finally {
+            MiniClusterUtils.stopCluster(cluster);
+        }
+    }
+}
```
droazen: Would be good to add two additional test cases if you have time: one that tests copying an entire directory, and a second that tests copying a file that is smaller than the chunk size. These additional tests can be very simple and use very small files, since you've already tested the case of a "large" file ("large" being relative to the chunk size) in your existing test.

droazen:
Final review complete, back to @cwhelan . Just had a few minor remaining comments/suggestions, then feel free to squash/merge when ready.

src/main/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSpark.java
cwhelan: 64MiB -- this was just based on figuring out that a large chunk size seemed to work best and reading that bytebuffers should be allocated in powers of two (although I never really tested that that's necessary). Moved to a constant and added comment.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
cwhelan: done

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
cwhelan: It was only two; switched to a different file that's three chunks.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
cwhelan: I added one more test case that copies a directory with two very small files in it -- hopefully that covers both of your cases.

src/test/java/org/broadinstitute/hellbender/tools/spark/ParallelCopyGCSDirectoryIntoHDFSSparkIntegrationTest.java
droazen: Instead of switching to a larger file, is it possible to just decrease the block size further? (thinking about test runtimes here)

cwhelan:
The file size only went from 1.9MB to 3MB -- there's no perceptible
difference in test runtime that I can see.

droazen:
Forgot to mention earlier that another good test case to add would be a file whose size exactly equals the chunk size, since that triggers a special case in the code. Up to you whether you can sleep at night without such a test in place, since your group will be the primary users of this tool :)

PR #2544, opened 2017-03-29
fix mistake in regex testing if a string ends with slash

SHuang-Broad:
in SV cluster creation script


cwhelan:
Looks good to me.

PR #2547, opened 2017-03-30
Demote scope of GenotypingEngine.upstreamDeletionsLoc

davidbenjamin:
@droazen here you go.  It turns out that this variable had no business having class scope, so I made it local to the one method that uses it.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
@@ -47,8 +47,6 @@ public abstract class GenotypingEngine<Config extends StandardCallerArgumentColl
 
     private final AFPriorProvider log10AlleleFrequencyPriorsIndels;
 
-    private final List<SimpleInterval> upstreamDeletionsLoc = new LinkedList<>();
-
     /**
      * Construct a new genotyper engine, on a specific subset of samples.
      *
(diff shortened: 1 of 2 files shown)


davidbenjamin:
Pinging @droazen

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
```diff
@@ -353,56 +351,38 @@ private OutputAlleleSubset(final List<Allele> alleles, final List<Integer> mleCo
      * @return information about the alternative allele subsetting {@code null}.
      */
     private OutputAlleleSubset calculateOutputAlleleSubset(final AFCalculationResult afCalculationResult, final VariantContext vc) {
```
droazen: As discussed in person, we fully support the mission of making this class less stateful (and particularly making it so that `isVcCoveredByDeletion()` does not modify state, which was the original ticket), but we're concerned that this method `calculateOutputAlleleSubset()` was intended to be stateful, and to accumulate deletions across calls. It's also not well covered by tests, so this change could easily break the method.

Can you think about this issue and reply here once you've looked into it, @davidbenjamin ? Also at-mentioning @ronlevine to solicit his comments, since he appears to have added `upstreamDeletionsLoc` to this class in the first place.

ronlevine: @droazen Is correct. It was intended to maintain state for upstream deletions. Make sure the upstream spanning deletion does not span shard barriers, then there will be problems with the book keeping. AFAIK, shards overlap genome locations but for a large spanning deletion, the size might not be sufficient.

ronlevine: @davidbenjamin Here is the motivation for the change: https://github.com/broadinstitute/gsa-unstable/issues/1188.

davidbenjamin:
Will try again later some time after the HC tie-out and repo merge.

Issue #2549, opened 2017-03-30
How Can I Recalibrate Indel Quality?


ShawnWilliams:
Hi,
It seems that gatk4.0 has removed indel quality recalibration function. 
I need indel quality to use Mutect2. Is there any tool I can use to recalibrate indel quality?

lbergelson:
Hi @ShawnWilliams, our analysis showed us that indel quality recalibration wasn't giving us very much of a benefit and it was quite expensive computationally, so we disabled it by default.  

There's a hidden argument `-indelBQSR` which will re-enable that functionality, but we don't think it's necessary for most recent sequencing data.  I think it was a mistake that that argument was made hidden, I'm going to label it `Advanced` instead...

ShawnWilliams:
Hi @lbergelson 
Thanks so much for your help!
And thanks @droazen for add "question" label to my question!

I find that I can run BaseRecalibrator with -indelBQSR to generate a .grp file contains insertion type and deletion type, and with -enableBAQ to turn on BAQ function.
But I can't run ApplyBQSR with -indelBQSR, and the error information is:
***********************************************************************

A USER ERROR has occurred: i is not a recognized option

***********************************************************************

And if I run ApplyBQSR with no optional arguments, the output .bam file doesn't contain inserntion quality and deletion quality. How can I get a output bam file with indel quality?

lbergelson:
@ShawnWilliams I'm sorry, I think I steered you wrong.  It looks like the ability to *apply* the indel recalibration was removed at some point.  Which is probably why the argument was hidden in the first place...  

We did analysis and found that it didn't made a significant difference in the final variant calls if you don't do it, so we dropped support for it.  I thought we'd just turned it off by default, but it looks like it's actually been removed from the apply functions. 

Could you tell me a bit about your data?  Unless there is something very strange about it I wouldn't  expect it to gain much benefit from indel recalibration.  (Base recalibration is more important  if you have low quality or older data, but even in those cases the indel recalibration doesn't make that much difference.)   

If your really need to run it it looks like your only option is going to be to use gatk3 for now.  Sorry to have mislead you.

ShawnWilliams:
@lbergelson Thanks for your explanation, it really eliminates my confusion!
[...]

PR #2550, opened 2017-03-30
made enableBAQ and indelBQSR arguments Advanced instead of Hidden

lbergelson:
@droazen easy one for you... we do have test for both of these.  In fact, I think we ONLY have tests that enable these...

Inspired by #2549

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/recalibration/RecalibrationArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/recalibration/RecalibrationArgumentCollection.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.hellbender.utils.recalibration;
 
+import org.broadinstitute.barclay.argparser.Advanced;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.Hidden;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/utils/recalibration/RecalibrationArgumentCollection.java
```diff
@@ -99,11 +100,11 @@
     @Argument(fullName = "preserve_qscores_less_than", shortName = "preserveQ", doc = "Don't recalibrate bases with quality scores less than this threshold (with -" + StandardArgumentDefinitions.BQSR_TABLE_SHORT_NAME + ")", optional = true)
     public int PRESERVE_QSCORES_LESS_THAN = QualityUtils.MIN_USABLE_Q_SCORE;
 
-    @Hidden
+    @Advanced
     @Argument(fullName = "enableBAQ", shortName = "enableBAQ", doc = "do BAQ correction")
     public boolean enableBAQ = false;
 
-    @Hidden
+    @Advanced
```
droazen: Add javadoc explaining why these two arguments are `@Advanced` and not turned on by default.

droazen:
:+1: merge after addressing my comment

droazen:
@lbergelson Did we decide that we actually don't want to make this change?

droazen:
Closing as per discussion with @lbergelson

PR #2552, opened 2017-03-30
fixes the error in computing contig span in BwaMemAlignment -> AlignmentRegion

SHuang-Broad:
which resulted in inversion call dropping
@tedsharpe , feel free to cherry pick this change in your PR #2444 
But it does indicate that the binding to bwa mem has an issue in template span inference, or there's mis-communication.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
@@ -59,8 +59,8 @@ class AlignmentRegion {
         this.mapQual = alignment.getMapQual();
         this.mismatches = alignment.getNMismatches();
         this.assembledContigLength = contigLen;
-        this.startInAssembledContig = alignment.getSeqStart()+1;
-        this.endInAssembledContig = alignment.getSeqEnd();
+        this.startInAssembledContig = startOfAlignmentInContig();
+        this.endInAssembledContig = endOfAlignmentInContig();
     }
 
     @VisibleForTesting


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignmentRegion.java
```diff
@@ -59,8 +59,8 @@ public AlignmentRegion(final String assemblyId, final String contigId, final int
         this.mapQual = alignment.getMapQual();
         this.mismatches = alignment.getNMismatches();
         this.assembledContigLength = contigLen;
-        this.startInAssembledContig = alignment.getSeqStart()+1;
-        this.endInAssembledContig = alignment.getSeqEnd();
+        this.startInAssembledContig = startOfAlignmentInContig();
+        this.endInAssembledContig = endOfAlignmentInContig();
```
tedsharpe: Yes, this is my bug.  This issue is that the seqStart() and seqEnd() functions return coordinates on the contig sequence *as reversed*, if the SAM flag indicates that the sequence has been reversed.  AlignmentRegion wants the coordinates on the original contig.
This code will fix the problem, but it would be a little faster to write:
```
        if ( forwardStrand ) {
            this.startInAssembledContig = alignment.getSeqStart() + 1;
            this.endInAssembledContig = alignment.getSeqEnd();
        } else {
            this.startInAssembledContig = contigLen - alignment.getSeqEnd() + 1;
            this.endInAssembledContig = contigLen - alignment.getSeqStart();
        }
```

SHuang-Broad: Thanks. This is unfortunately a mis-communication.

SHuang-Broad:
Closing because change is now picked up in #2444

PR #2556, opened 2017-03-31
Fixed a bug in TandemRepeat annotation

davidbenjamin:
It was assuming that the `ReferenceContext`'s interval has the same start as the `VariantContext` being annotated, which is often not true.  I tested this in Mutect and it seems to work correctly now.

@LeeTL1220 This will improve indel specificty in M2 by 75% or so.  Can you review?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/TandemRepeat.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/TandemRepeat.java
@@ -35,7 +35,7 @@ public final class TandemRepeat extends InfoFieldAnnotation {
             return Collections.emptyMap();
         }
 
-        final Pair<List<Integer>,byte[]> result = GATKVariantContextUtils.getNumTandemRepeatUnits(vc, getRefBases(ref, vc));
+        final Pair<List<Integer>,byte[]> result = GATKVariantContextUtils.getNumTandemRepeatUnits(vc, getRefBasesStartingAtVariantLocus(ref, vc));
         if (result == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/TandemRepeat.java
```diff
@@ -58,4 +58,10 @@
                 GATKVCFConstants.REPEATS_PER_ALLELE_KEY);
     }
 
+    private static byte[] getRefBases(final ReferenceContext ref, final VariantContext vc) {
```
LeeTL1220: I just want to confirm that there is a test that breaks given the bug that this branch fixes.  In other words, a test that would fail without this fix.  Unless you do not think that is useful.

LeeTL1220: `getRefBases` is a bit broad.  Can you rename to be a bit more specific?

davidbenjamin: there is now

davidbenjamin: done

davidbenjamin:
@LeeTL1220 your will has been done.

LeeTL1220:
I think that change is fine.  I expected you to add a test, rather than altering an existing one, but it looks like you are testing a superset of the typical conditions, anyway. If that is not the case, don't merge.  Otherwise, go for it!

Issue #2422, opened 2017-02-23
Get NIO writing working for both bams and vcfs in walkers


droazen:
Writing output via NIO is currently broken, as reported by @jean-philippe-martin:

```
 /home/jpmartin/build/install/gatk/bin/gatk PrintReads -I inputs/CEUTrio.HiSeq.WEx.b37.NA12892.bam -L 10:1000000-2000000 -O gs://jpmartin/staging/out_tmp.bam
htsjdk.samtools.util.RuntimeIOException: Error opening file: /home/jpmartin/gs:/jpmartin/staging/out_tmp.bam
        at htsjdk.samtools.SAMFileWriterFactory.makeBAMWriter(SAMFileWriterFactory.java:246)
```

Now that the read support is in good shape, let's get the writing end working -- should hopefully be an easy fix.

droazen:
For @jean-philippe-martin

jean-philippe-martin:
It looks like we may had to add `Path` support to htsjdk's `SAMFileWriterFactory` class. It's using `File` right now, which prevents it from using NIO.

jean-philippe-martin:
GCS output works now for PrintReads. Next step is to switch all the other walkers to use the Path overload, too.

jean-philippe-martin:
For what it's worth, this works but I should improve the error message. If you don't have the right permissions, currently you get an error that's something like this:

```
htsjdk.samtools.util.RuntimeIOException: Error opening file: file:///home/jpmartin//gatk/gs:/bobs-bucket-you-cant-write-to/test-output/printtogcs.bam
```

This is both
- confusing: it makes it look like the path is wrong even though the code actually tried the correct path
- improvable: it'd be nice if we gave the reason there was an error (403 Forbidden: user jpmartin doesn't have write access)

Issue #840, opened 2015-08-19
Configure travis to install/init git lfs


droazen:
Now that we're using git lfs to manage our large test resources, we need to configure travis to install/init git lfs before running the test suite.

droazen:
@lbergelson As the resident travis expert, feel like taking this one?

lbergelson:
Yeah, I can do this.

droazen:
Done!

PR #2558, opened 2017-03-31
NIO output support for PrintReads

jean-philippe-martin:
Add a version of `createSAMWriter` that takes in a `Path` so we can work with NIO for output. This allows PrintReads to write its output to a cloud bucket.

Tested like this:
```
$ ./gatk-launch PrintReads -I gs://$MYBUCKET/CEUTrio.HiSeq.WEx.b37.NA12892.bam -L 10:1000000-2000000  -O gs://$MYBUCKET/pr.bam
$ gsutil cp gs//$MYBUCKET/pr.bam .
$ samtools view pr.bam | md5sum
c7f41be91031bea6d28d59d40b54f304  -
```

This MD5 matches the expected output for this command.

Once we agree this is the correct way forward, I should go ahead and change all of the other tools similarly.

This PR is part of addressing #2422


jean-philippe-martin:
Hitting a snag: the md5 output option doesn't seem to work with streams, and so the simple approach of "just use Path everywhere" fails because makeSAMWriter in htsjdk doesn't behave identically between a file or an outputStream.

Now the question is: is there a fundamental reason for that, or can just add in the md5 feature? At first blush, it seems possible.

@droazen, what's your expert opinion?

jean-philippe-martin:
I got it to work after adding Path support to htsjdk; I'll send a pull request their way first.

jean-philippe-martin:
Still blocked on htsjdk review.

jean-philippe-martin:
I haven't forgotten about this it's just still blocked by [htsjdk #840](https://github.com/samtools/htsjdk/pull/840).

jean-philippe-martin:
htsjdk updated, so I changed this to refer to the snapshot and rebased.

Seems to work:

```
$ samtools view pr.bam | md5sum
c7f41be91031bea6d28d59d40b54f304  -
```

jean-philippe-martin:
Travis' failing, but is this really my code? It looks like the test file is corrupted:

````
[Lorg.broadinstitute.hellbender.tools.picard.sam.MergeBamAlignmentIntegrationTest$MultipleAlignmentSpec;@32a44fc3) FAILED
    htsjdk.samtools.SAMFormatException: Error parsing text SAM file. CIGAR covers 15 bases but the sequence is 16 read bases ; File /tmp/aligned.6084110718916383573.sam; Line 4
    Line: theRead	0	chr1	1	205	1S13M1S	*	0	0	ACGTACGTACGTACGT	5555555555555555	YB:i:1
````

I indeed count 16 reads in "ACGTACGTACGTACGT" where there should be 15. It looks like there are also 15 quality values next, so perhaps it's the CIGAR that's wrong?

**Comment by lbergelson** [comment]
@jean-philippe-martin Where are you seeing that test failure?  I'm seeing the travis failures as 

```
htsjdk.samtools.SAMException: Exception when processing alignment for BAM index i 1/2 76b aligned read.
	at htsjdk.samtools.BAMFileWriter.writeAlignment(BAMFileWriter.java:142)
	at htsjdk.samtools.SAMFileWriterImpl.addAlignment(SAMFileWriterImpl.java:201)
	at htsjdk.samtools.AsyncSAMFileWriter.synchronouslyWrite(AsyncSAMFileWriter.java:36)
	at htsjdk.samtools.AsyncSAMFileWriter.synchronouslyWrite(AsyncSAMFileWriter.java:16)
	at htsjdk.samtools.util.AbstractAsyncWriter$WriterRunnable.run(AbstractAsyncWriter.java:124)
	at java.lang.Thread.run(Thread.java:748)
Caused by: htsjdk.samtools.SAMException: Exception creating BAM index for record i 1/2 76b aligned read.
	at htsjdk.samtools.BAMIndexer.processAlignment(BAMIndexer.java:119)
	at htsjdk.samtools.BAMFileWriter.writeAlignment(BAMFileWriter.java:139)
	... 5 more
Caused by: htsjdk.samtools.SAMException: IOException in BinaryBAMIndexWriter reference 0
	at htsjdk.samtools.BinaryBAMIndexWriter.writeReference(BinaryBAMIndexWriter.java:151)
	at htsjdk.samtools.BAMIndexer.advanceToReference(BAMIndexer.java:138)
	at htsjdk.samtools.BAMIndexer.processAlignment(BAMIndexer.java:115)
	... 6 more
Caused by: java.nio.channels.ClosedByInterruptException
	at java.nio.channels.spi.AbstractInterruptibleChannel.end(AbstractInterruptibleChannel.java:202)
	at sun.nio.ch.FileChannelImpl.write(FileChannelImpl.java:216)
	at java.nio.channels.Channels.writeFullyImpl(Channels.java:78)
	at java.nio.channels.Channels.writeFully(Channels.java:101)
	at java.nio.channels.Channels.access$000(Channels.java:61)
	at java.nio.channels.Channels$1.write(Channels.java:174)
	at java.io.BufferedOutputStream.flushBuffer(BufferedOutputStream.java:82)
	at java.io.BufferedOutputStream.flush(BufferedOutputStream.java:140)
	at htsjdk.samtools.BinaryBAMIndexWriter.writeReference(BinaryBAMIndexWriter.java:149)
	... 8 more
```

One of the logs is here: https://storage.googleapis.com/hellbender-test-logs/build_reports/12617.7/tests/test/index.html

We saw a whole host of those cigar validation errors you're seeing when we updated htsdjk the last time, it's a new validation that wasn't previously checked in htsjdk, so a lot of the test files had errors in them that no one had ever noticed/bothered fixing.

**Comment by lbergelson** [comment]
Also, in case you didn't know, our travis builds should all publish test logs, you can find the link if you skip down to the very end of the travis log and look for a line like:

```
See the test report at https://storage.googleapis.com/hellbender/test/build_reports/12615.1/tests/test/index.html
```

**Comment by jean-philippe-martin** [comment]
Thanks @lbergelson. I'll check on that.

**Comment by jean-philippe-martin** [comment]
rebased, using master's (new) version of htsjdk.

**Comment by jean-philippe-martin** [comment]
All checks passed, yay! @lbergelson I don't think you've reviewed this code yet, but now it's ready for review!

File: build.gradle [hunk 1]
```diff
@@ -95,10 +95,12 @@ jacocoTestReport {
 
 //NOTE: we ignore contracts for now
 compileJava {
-  options.compilerArgs = ['-proc:none', '-Xlint:all','-Werror', '-Xdiags:verbose']
+  // removing -Werror temporarily
```
lbergelson: Could you put `-Werror` back?  Is there a warning you couldn't resolve?

src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
```diff
@@ -27,13 +31,17 @@
 public final class PrintReads extends ReadWalker {
 
     @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME, shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc="Write output to this file")
-    public File OUTPUT;
+    public String OUTPUT_STR;
+
+    public Path getOutputPath() {
```
lbergelson: I would just inline this method.

src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
```diff
@@ -27,13 +31,17 @@
 public final class PrintReads extends ReadWalker {
 
     @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME, shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc="Write output to this file")
-    public File OUTPUT;
+    public String OUTPUT_STR;
```
lbergelson: I think we should just rename this to just `output`.  The all caps is wrong since it's not really constant.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -676,6 +677,32 @@ public final SAMFileGATKReadWriter createSAMWriter(final File outputFile, final
...
+    /*
+     * Create a common SAMFileWriter using the reference and read header for this tool.
+     *
+     * @param outputFile    - if this file has a .cram extension then a reference is required. Can not be null.
+     * @param preSorted     - if true then the records must already be sorted to match the header sort order
+     *
+     * @throws UserException if outputFile ends with ".cram" and no reference is provided
+     * @return SAMFileWriter
+     */
+    public final SAMFileGATKReadWriter createSAMWriter(final Path outputPath, final boolean preSorted) {
```
lbergelson: I think the old file methods should delegate to these new ones.   Wrap the file in `Utils.nonNull(file).toPath()` to avoid introducing NPEs where we used to have `IllegalArgumentException`.

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
```diff
@@ -1054,6 +1090,43 @@ public static SAMFileWriter createCommonSAMWriterFromFactory(
...
+            }
+            // htsjdk doesn't currently support presorted & outputStream at the same time for CRAM :-(
+            return factory.makeCRAMWriter(header, outputPath, referenceFile);
+        } else if (outputPath.toString().endsWith(".sam")) {
+            return factory.makeSAMWriter(header, preSorted, outputPath);
+        } else {
+            // default to BAM, matching the logic in htsjdk
+            return factory.makeBAMWriter(header, preSorted, outputPath);
+        }
+    }
```
cmnbroad: I think this should just delegate to makeWriter like the File overload above does. That should do the right thing. Even better, can we consolidate the two methods so the original File one delegates to the this one. That would help with ensuring they're the same (i.e., we should preserve the existing behavior of cloning the header before passing it off to the writer).

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
```diff
@@ -1054,6 +1090,43 @@ public static SAMFileWriter createCommonSAMWriterFromFactory(
...
+        OpenOption... openOptions)
+    {
+        Utils.nonNull(outputPath);
+        Utils.nonNull(header);
+
+        if (outputPath.toString().endsWith(CramIO.CRAM_FILE_EXTENSION)) {
+            if (null == referenceFile ) {
+                throw new UserException("A reference file is required for writing CRAM files");
+            }
+            // htsjdk doesn't currently support presorted & outputStream at the same time for CRAM :-(
```
cmnbroad: Not sure what this refers to - but see my comment below.

build.gradle
jean-philippe-martin: I put -Werr back in, it works. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
jean-philippe-martin: done

src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
jean-philippe-martin: done

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jean-philippe-martin: The two methods are not exactly identical, sadly, so if we do this then we're changing the semantics of the old ones.

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
jean-philippe-martin: This explains why we are not forwarding the "preSorted" flag. Because there is no overload that will take it.

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
jean-philippe-martin: OK let's give this a try.

jean-philippe-martin:
Remains to write the extra tests

jean-philippe-martin:
I added a version with Path for CreateCommonSAMWriter. However for me to add the CRAM test, could you please tell me what I should pass for the reference? Without it I cannot write to that format, and I don't see an existing test here that writes a CRAM file.

jean-philippe-martin:
@lbergelson counter-proposal: since writing to a temp location in GCS would risk collisions if multiple people run the test, how about writing to JimFS instead? It's a RAM filesystem so each test machine gets its own, and it still requires the code to use the Path objects correctly since any conversion to File would fail. As a bonus, we do not incur Cloud charges and the test is much faster, so we can keep it as a unit test instead of an integration test.

jean-philippe-martin:
I upgraded both the File and the Path tests for `createCommonSAMWriter`, checking that the written file contents match the source file.

cmnbroad:
@jean-philippe-martin I had assumed that there were already CRAM tests in ReadUtilsUnitTest for the old createSAMWriter method, but there aren't (there are integration tests that exercise the cram code path through PrintReadsIntegrationTest.) In general you have to just know which reference goes with which cram. The canonical file we use is print_reads.cram/print_reads.fasta (in the tools test folder). You should probably copy them into the ReadUtils test folder to use it in those tests. If you do,you'll have to also copy print_reads.fasta.fai, or you'll get misleading CRAM errors. Also note that you won't be able to use iterator comparison when comparing an Iterator<CRAMRecord> with Iterator<BAMRecord>.

lbergelson:
@jean-philippe-martin I like your counter proposal in general for testing path integration.  I think writing to GCS over NIO is an important enough feature that we should have at least 1 test in gatk that actually writes to a real GCS bucket in case there's ever an issue specifically with GCS (authentication issues are one potential problem I can imagine).  

It seems like we should be able to design in a way that avoids collisions.  What does `Files.createTempFile()` do with gcs? My guess is that it probably doesn't do the right thing, but maybe we could fix it so it would? Or use some sort of scheme with random UUID's like the methods in BucketUtils that we have already.

jean-philippe-martin:
(Travis failure looks to be a transient issue getting https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot/com/github/samtools/htsjdk/2.11.0-18-ga35b420-SNAPSHOT/htsjdk-2.11.0-18-ga35b420-20170925.132250-2.jar)

lbergelson:
@jean-philippe-martin I restarted travis.

jean-philippe-martin:
I've added the CRAM tests. Remain to add an integration test that writes to GCS.

@cmnbroad can you explain what you mean by "Also note that you won't be able to use iterator comparison when comparing an Iterator with Iterator." ?

cmnbroad:
@jean-philippe-martin Sorry, I originally typed "you can't compare `Iterator<CRAMRecord>` with `Iterator<SAMRecord>`" , but I didn't quote it, so it displays as "you can't compare Iterator with Iterator". Anyway, it looks like you're not doing that.

Thanks for adding the CRAM tests. Rather than adding separate data providers and methods for them though, can you just change the existing providers and methods to have an output extension and a reference (null is OK), and then thread those through the test code. I made a quick branch with a commit [here](https://github.com/broadinstitute/gatk/commit/5e52fca813e57065713852d12f80a7599fcbc3ce) to show you what I mean - it would eliminate a lot of redundant code.

jean-philippe-martin:
@cmnbroad thank you for the branch, that made this very easy! I didn't know that the `SamReaderFactory` could also read CRAM files. Perhaps htsjdk should document this feature more prominently.

jean-philippe-martin:
@lbergelson I added an integration test that writes to GCS... it doesn't work for me ("com.google.cloud.storage.StorageException (...) does not have storage.objects.get access to (...)")

This may be due to a misconfiguration on my end. I wonder if it'll work with Travis.

jean-philippe-martin:
@lbergelson I was able to get the auth set up right for the integration test (locally). It fails in a few places because some combinations are not supported.

jean-philippe-martin:
Yay, Travis passes! OK @lbergelson it's ready for review and perhaps we can finally close this PR.

jean-philippe-martin:
@droazen you were asking for a check of the performance. I ran the following two commands:

```
$ ./gatk-launch PrintReads -I gs://$MYBUCKET/CEUTrio.HiSeq.WGS.b37.ch20.1m-2m.NA12878.bam  -O gs://$MYBUCKET/pr.bam
$ ./gatk-launch PrintReads -I gs://$MYBUCKET/CEUTrio.HiSeq.WGS.b37.ch20.1m-2m.NA12878.bam  -O /tmp/pr.bam
```

output to: | local disk | GCS
--|--|--
run 1 | 0.12 min | 0.68 min
run 2 | 0.12 min | 0.29 min
run 3 | 0.12 min | 0.28 min
**median** |**0.12 min** | **0.29 min**

So it looks like there's a significant performance difference. For what it's worth, copying the output file to GCS from my desktop takes 3.5s. The log when running PrintReads indicates:

```
11:06:13.011 INFO  PrintReads - HTSJDK Defaults.COMPRESSION_LEVEL : 1
11:06:13.011 INFO  PrintReads - HTSJDK Defaults.USE_ASYNC_IO_READ_FOR_SAMTOOLS : false
11:06:13.011 INFO  PrintReads - HTSJDK Defaults.USE_ASYNC_IO_WRITE_FOR_SAMTOOLS : true
11:06:13.011 INFO  PrintReads - HTSJDK Defaults.USE_ASYNC_IO_WRITE_FOR_TRIBBLE : false
```

jean-philippe-martin:
Biweekly reminder that a review would be nice when you have the chance.

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsIntegrationTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.SAMFileWriter;
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.SamFiles;
+import htsjdk.samtools.SamReader;
+import htsjdk.samtools.SamReaderFactory;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.FileSystem;
```
cmnbroad: Also unused.

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsUnitTest.java
```diff
@@ -3,8 +3,19 @@
...
+import com.google.common.jimfs.Configuration;
+import com.google.common.jimfs.Jimfs;
 import htsjdk.samtools.*;
+import htsjdk.samtools.SamReader.PrimitiveSamReader;
+import htsjdk.samtools.cram.ref.CRAMReferenceSource;
+import htsjdk.samtools.cram.ref.ReferenceSource;
 import htsjdk.samtools.reference.IndexedFastaSequenceFile;
+import htsjdk.samtools.util.CloseableIterator;
+import java.io.IOException;
+import java.io.InputStream;
```
cmnbroad: Quite a few of these imports are unused.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -676,6 +677,32 @@ public final SAMFileGATKReadWriter createSAMWriter(final File outputFile, final
         );
     }
 
+    /*
+     * Create a common SAMFileWriter using the reference and read header for this tool.
+     *
+     * @param outputFile    - if this file has a .cram extension then a reference is required. Can not be null.
```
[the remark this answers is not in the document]
jean-philippe-martin: done.

src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
```diff
@@ -2,6 +2,9 @@
 
 import htsjdk.samtools.*;
 import htsjdk.samtools.cram.build.CramIO;
+import java.nio.file.Files;
```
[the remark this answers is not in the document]
jean-philippe-martin: removed.

src/main/java/org/broadinstitute/hellbender/utils/test/BaseCloudTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.nio.file.Paths;
+import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
+
+/**
+ * Base class for tests that use cloud resources, we include
+ * a few helpful methods.
+ */
+public abstract class BaseCloudTest extends BaseTest {
+
+  protected FileSystem getAuthenticatedGcs(String bucket) throws IOException {
```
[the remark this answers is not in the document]
jean-philippe-martin: done

src/main/java/org/broadinstitute/hellbender/utils/test/BaseCloudTest.java
```diff
@@ -0,0 +1,52 @@
...
+    byte[] creds = Files.readAllBytes(Paths.get(getGoogleServiceAccountKeyPath()));
```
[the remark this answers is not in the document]
jean-philippe-martin: done

src/main/java/org/broadinstitute/hellbender/utils/test/BaseCloudTest.java
```diff
@@ -0,0 +1,52 @@
...
+  private void printKeyLines(Path path, String... keywords) throws IOException {
+    for (String line : Files.readAllLines(path)) {
+      for (String keyword : keywords) {
+        if (line.contains(keyword)) {
+          System.err.println(line);
+        }
+      }
+    }
+  }
+
```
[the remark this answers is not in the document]
jean-philippe-martin: helpDebugAuthError is generally useful when writing new tests, which is why I put it outside.
getAuthenticatedGcs could be useful for new tests as well but you are correct that currently it's only used in one.

We can move them up to BaseTest, up to you. I thought it'd be nice to separate out the cloud-specific code but that's not required.

I added more documentation.

src/test/java/org/broadinstitute/hellbender/utils/nio/GcsNioIntegrationTest.java
```diff
@@ -34,6 +38,27 @@ public void openPublicFile() throws IOException {
...
+    @Test(groups={"bucket"})
+    public void writePrivateFile() throws IOException {
+        try {
+        final String dest = getGCPTestStaging() + "GcsNioIntegrationTest-writePrivateFile-test" + new Random().nextInt();
+        final Path outputPath = BucketUtils.getPathOnGcs(dest);
+        System.out.println("Writing to " + dest);
+        try (OutputStream os = Files.newOutputStream(outputPath)) {
+            os.write(42);
+        }
+        Files.deleteIfExists(outputPath);
```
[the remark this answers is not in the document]
jean-philippe-martin: Fixed.

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsIntegrationTest.java
jean-philippe-martin: Removed all unused imports in this file.

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsUnitTest.java
jean-philippe-martin: Removed all unused imports in this file.

jean-philippe-martin:
@cmnbroad asking:

> I had the same thought as @lbergelson. How does obtaining a Path from the File and delegating to the > Path overload change the semantics ?

OK so I checked again (it's been a long time since I first checked) and now I see that they go to `createCommonSAMWriterFromFactory` which then go to the same place. So it appears this isn't true anymore and the semantics would be the same.

With this in mind, then, we can simplify createCommonSAMWriter as well as createSAMWriter.

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsUnitTest.java
jean-philippe-martin: Since unused imports are an issue, I ran the IDE tool to remove all unused imports. The source should be easier on the eyes now.

jean-philippe-martin: I've reverted all of the other files because this tool touched pretty much every single file in the project, that was crazy.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -660,19 +668,34 @@ private void checkToolRequirements() {
      * @return SAMFileWriter
      */
     public final SAMFileGATKReadWriter createSAMWriter(final File outputFile, final boolean preSorted) {
-        if (!hasReference() && IOUtils.isCramFile(outputFile)) {
+        return createSAMWriter(
+            (null == outputFile ? null : outputFile.toPath()),
```
cmnbroad: Per Louis' old suggestion below, you can just wrap the File arg as "Utils.nonNull(file).toPath()" so we don't get NPEs downstream.

src/main/java/org/broadinstitute/hellbender/utils/test/BaseCloudTest.java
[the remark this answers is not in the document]
cmnbroad: Yes, lets move them up to BaseTest please.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jean-philippe-martin: done.

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsIntegrationTest.java
```diff
@@ -0,0 +1,102 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Iterator;
+import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class ReadUtilsIntegrationTest extends BaseTest {
```
magicDGS: Should extend `GATKBaseTest` after #3475

jean-philippe-martin:
Travis is failing with:

```org.broadinstitute.hellbender.tools.copynumber.plotting.PlotDenoisedCopyRatiosIntegrationTest.testPlotting FAILED
org.broadinstitute.hellbender.utils.R.RScriptExecutorException: 
    Rscript exited with 1
    Command Line: Rscript -e tempLibDir = '/tmp/travis/Rlib.2269507098272465687';source('/tmp/CNVPlottingLibrary.8173385226459567897.R');source('/tmp/PlotDenoisedCopyRatios.718174557431831918.R'); --args --sample_name=test --standardized_copy_ratios_file=/home/travis/build/broadinstitute/gatk/src/test/resources/org/broadinstitute/hellbender/tools/copynumber/plotting/plotting-copy-ratios.tsv --denoised_copy_ratios_file=/home/travis/build/broadinstitute/gatk/src/test/resources/org/broadinstitute/hellbender/tools/copynumber/plotting/plotting-copy-ratios.tsv --contig_names=1CONTIG_DELIMITERX --contig_lengths=62393753CONTIG_DELIMITER48807114 --output_dir=/tmp/travis/testDir6678707894869143461/ --output_prefix=test
    Stdout: 
    Stderr: Error in library(data.table) : there is no package called ‘data.table’
```

This is outside of my area of expertise, I'm going to need a hand.

cmnbroad:
@jean-philippe-martin Its not just your branch. Looks like [this](https://github.com/broadinstitute/gatk/issues/3772). We're working on a fix.

cmnbroad:
@jean-philippe-martin If you rebase on master it should work now.

jean-philippe-martin:
@cmnbroad Travis is green, thank you!
OK for me to press the big green "Squash and merge" button?

cmnbroad:
@jean-philippe-martin The changes all look fine now, but it looks like you merged master into your branch (twice along the way?) instead of rebasing the branch onto master, so your branch commits are now distributed through time. I have no idea if the squash&merge does the right thing with these (without changing master history ?) @lbergelson do you know ?

lbergelson:
I'm not sure what will happen if you squash and merge.  It'll probably be fine?  You could do a manual merge though.

droazen:
Let's please repair the branch before merge, rather than risk clobbering master.

jean-philippe-martin:
Rebased to master (which put all the commits together), then squashed.

jean-philippe-martin:
@droazen with your blessing, we can merge.

cmnbroad:
@jean-philippe-martin The rebase looks fine, but now some new conflicts were introduced over the weekend. Please resolve once more and then, mercifully, we can merge this.

jean-philippe-martin:
Here's the new one. Once I see tests are green, I'll merge it in.

jean-philippe-martin:
Green, merging! Thank you everyone!

Issue #2402, opened 2017-02-10
Remove all code paths in engine (and tests) that use an explicit Google API key


droazen:
NIO uses the gcloud system authentication, so we shouldn't need this anymore.

droazen:
For @jean-philippe-martin

jean-philippe-martin:
Unless, that is, we want to use the API key to allow non-logged-in access to public data?

droazen:
I don't think so -- using the API key introduces all sorts of security issues with sanitizing command lines. I think we just want to rely on the default Google credentials.

lbergelson:
👍 to removing the api key.  We're leaking them all over the place...  We could optionally go back to accepting a key file as a form of authentication which is much less dangerous.

jean-philippe-martin:
OK. So my understanding of the plan here is that 
today we read BAM files 3 different ways:
1. Using hadoop when on hdfs and using Spark
2. google-genomic's `BAMIO` with Spark for BQSR and related code
3. `NIO` -> gatk `ReadsDataSource` -> htsjdk BAM reader, for non-Spark tools

The code in 2. uses `AuthHolder` which in turn takes in an API key. In order to get rid of the API key we're changing the code in 2. to use the method from 3. to read the BAM files on the cloud.

jean-philippe-martin:
In particular that means we're leaving in the code that talks to the Google Genomics Reference API (the `ReferenceAPISource` class), which also uses the API key. So at least in this stage we're not going to literally remove all instances of the API key (as the issue description suggests).

droazen:
@jean-philippe-martin Is this one done (modulo the `ReferenceAPISource class)?

jean-philippe-martin:
[I see dead code](http://www.imdb.com/title/tt0167404/quotes?item=qt0249072), still. There's more work to do. For example `MetricsUtils.saveMetrics` takes in an authHolder but then doesn't use it - I need to remove that and transitively remove it from all the functions that only take one in to pass it there.

PR #2559, opened 2017-03-31
Remove dependency on BAMIO

jean-philippe-martin:
Use NIO and htsjdk instead.

This belongs under issue #2402.

It's all gone:
```
$ grep -r BAMIO src/*
$
```


droazen:
:+1:

jean-philippe-martin:
Excellent, thank you!

PR #2568, opened 2017-04-04
changing condition in GenotypingEngine to match GATK3

lbergelson:


droazen:
:+1: merge once tests pass

PR #2569, opened 2017-04-04
(Do not merge) Create a snapshot of gatk-public based on 4.alpha.2-188-g7332d10 with TandemRepeat fix applied

droazen:
This branch takes the version of gatk-public that gatk-protected currently depends on (4.alpha.2-188-g7332d10) and applies @davidbenjamin 's fix to the `TandemRepeat` annotation to it. The only purpose of this PR is to cause a snapshot to be generated -- do not merge!

This is necessary to unblock @davidbenjamin 's work, because the `HaplotypeCaller` tests are failing if we update protected to the latest public head, and although we've fixed some of the issues there are some unexplained failures in the concordance tests.


droazen:
Realized I had to upload a snapshot manually, since the git history of this branch is not compatible with master -- closing

PR #2574, opened 2017-04-05
Added support for receiving parameters for GKL VectorLoglessPairHMM: implement…

erniebrau:
…ation (AVX vs AVX-OMP), max num threads, and toggle using double precision

The changes are based off of [the lb_connect_pairhmmargs branch](https://github.com/broadinstitute/gatk/tree/lb_connect_pairhmmargs), which, apparently, never got merged. The main difference between that branch and this fork is that the PairHMMNativeArguments object is passed all the way through to VectorLoglessPairHMM, which was not the case in the lb_connect_pairhmmargs branch.

There is a corresponding [fork of gatk-protected](https://github.com/erniebrau/gatk-protected-1) which defines the new CLI arguments that talk to the changes in this PR. There should be a PR from that fork to gatk-protected, if and when this one is merged.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
@@ -23,11 +23,23 @@ import java.util.Map;
  */
 public final class VectorLoglessPairHMM extends LoglessPairHMM {
 
+    /**
+     * Type for implementation of VectorLoglessPairHMM
+     */
+    public enum Implementation {
+        /**
+         * AVX-accelerated version of PairHMM
+         */
+        AVX,
+        /**
+         * OpenMP multi-threaded AVX-accelerated version of PairHMM
+         */
+        OMP
+    }
+
     private static final Logger logger = LogManager.getLogger(VectorLoglessPairHMM.class);
     private long threadLocalSetupTimeDiff = 0;
     private long pairHMMSetupTime = 0;
-
-    public enum Implementation {AVX, OMP}
 
     private final PairHMMNativeBinding pairHmm;
 
@@ -36,6 +48,12 @@ public final class VectorLoglessPairHMM extends LoglessPairHMM {
     private final Map<Haplotype, Integer> haplotypeToHaplotypeListIdxMap = new LinkedHashMap<>();
     private HaplotypeDataHolder[] mHaplotypeDataArray;
 
+    /**
+     * Create a VectorLoglessPairHMM
+     *
+     * @param implementation    which implementation to use (AVX or OMP)
+     * @param args              arguments to the native GKL implementation
+     */
     public VectorLoglessPairHMM(Implementation implementation, PairHMMNativeArguments args) throws UserException.HardwareFeatureException {
         final boolean isSupported;
 
--- a/src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.hellbender.utils.pairhmm;
 
 import com.intel.gkl.pairhmm.IntelPairHmm;
+import com.intel.gkl.pairhmm.IntelPairHmmOMP;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.gatk.nativebindings.pairhmm.HaplotypeDataHolder;
@@ -26,6 +27,8 @@ public final class VectorLoglessPairHMM extends LoglessPairHMM {
     private long threadLocalSetupTimeDiff = 0;
     private long pairHMMSetupTime = 0;
 
+    public enum Implementation {AVX, OMP}
+
     private final PairHMMNativeBinding pairHmm;
 
     //Hold the mapping between haplotype and index in the list of Haplotypes passed to initialize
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -176,7 +176,7 @@ dependencies {
     // Dependency change for including MLLib
     compile('com.esotericsoftware:reflectasm:1.10.0:shaded')
 
-    compile('com.intel.gkl:gkl:0.4.1') {
+    compile('com.intel.gkl:gkl:0.4.3') {
```
droazen: Can you rebase this branch onto the latest master before we review @erniebrau ? We are already on GKL 0.4.3 in this repo. Thanks!

src/test/java/org/broadinstitute/hellbender/utils/pairhmm/VectorPairHMMUnitTest.java
```diff
@@ -23,22 +24,27 @@
...
         args.useDoublePrecision = false;
         args.maxNumberOfThreads = 1;
-        final N2MemoryPairHMM avxPairHMM = new VectorLoglessPairHMM(args);
-        avxPairHMM.doNotUseTristateCorrection();
-        return Collections.singletonList(avxPairHMM);
+
+        for (VectorLoglessPairHMM.Implementation imp : VectorLoglessPairHMM.Implementation.values()) {
+            try {
+                final N2MemoryPairHMM avxPairHMM = new VectorLoglessPairHMM(imp, args);
+                avxPairHMM.doNotUseTristateCorrection();
```
droazen: Can you add a comment explaining why this line is needed?

src/test/java/org/broadinstitute/hellbender/utils/pairhmm/VectorPairHMMUnitTest.java
```diff
@@ -23,22 +24,27 @@
...
+                list.add(avxPairHMM);
+            }
+            catch (UserException.HardwareFeatureException e ) {}
```
droazen: I think instead of swallowing all exceptions here, we want the tests for `OMP`/`AVX` to be individually skipped if the PairHMM can't be loaded, so that it will be clear from the test logs what was tested.

src/test/java/org/broadinstitute/hellbender/utils/pairhmm/VectorPairHMMUnitTest.java
```diff
@@ -23,22 +24,27 @@
...
+        }
+
+        if (list.size() == 0) {
+            throw new SkipException("AVX PairHMM is not supported on this system or the library is not available");
```
droazen: Throwing an exception in a method called from a `DataProvider` is problematic -- I think it causes the test suite to fail in hard-to-interpret ways. Recommend moving the throwing of the `SkipException` to the `@Test` method itself.

src/test/java/org/broadinstitute/hellbender/utils/pairhmm/VectorPairHMMUnitTest.java
```diff
@@ -23,22 +24,27 @@
...
-       }
-    }
-
+    // Return a list of supported VectorLoglessPairHMM implementations, skip the test if none are supported
     private List<N2MemoryPairHMM> getHMMs() {
-        final PairHMMNativeArguments args = new PairHMMNativeArguments();
+        List<N2MemoryPairHMM> list = new ArrayList<>();
+        PairHMMNativeArguments args = new PairHMMNativeArguments();
         args.useDoublePrecision = false;
         args.maxNumberOfThreads = 1;
```
droazen: Are we properly exercising the `OMP` implementation with 1 thread?

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
```diff
@@ -26,23 +27,40 @@
...
     private final Map<Haplotype, Integer> haplotypeToHaplotypeListIdxMap = new LinkedHashMap<>();
     private HaplotypeDataHolder[] mHaplotypeDataArray;
 
-    public VectorLoglessPairHMM(PairHMMNativeArguments args) throws UserException.HardwareFeatureException {
-        // TODO: connect GATK temp directory
-        final boolean isSupported = new IntelPairHmm().load(null);
-
-        if (!isSupported) {
-            throw new UserException.HardwareFeatureException("Machine does not support AVX PairHMM.");
+    public VectorLoglessPairHMM(Implementation implementation, PairHMMNativeArguments args) throws UserException.HardwareFeatureException {
```
droazen: Add javadoc for public constructor

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
```diff
@@ -26,23 +27,40 @@
     private long threadLocalSetupTimeDiff = 0;
     private long pairHMMSetupTime = 0;
 
+    public enum Implementation {AVX, OMP}
```
droazen: Add documentation for `Implementation` enum

droazen:
First-pass review complete, back to @erniebrau to address comments.

src/test/java/org/broadinstitute/hellbender/utils/pairhmm/VectorPairHMMUnitTest.java
erniebrau: I am not sure why it's there. The line was present in the previous version of the test; I left it because I wasn't sure what it did. Looking at VectorLoglessPairHMM code, it doesn't look like it's needed. I will only comment it out in case I am wrong. (Keep in mind I'm still new to this codebase and project :-) )

The rest of your comments are addressed by a commit I am about to push.

droazen: I would leave this line for now then, out of an abundance of caution, but could you file a github issue asking for an explanation of the line, and assign to @gspowley (or whoever is git-blamed for the original line)?

gspowley: According to [this line](https://github.com/broadinstitute/gatk/blob/47d5416fe2590d02a3223d0e00c6d8a8b4e2b3d0/src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java#L92), `doNotUseTristateCorrection` is for debugging purposes. By default, we use the tristate correction. To disable the tristate correction, code must call `doNotUseTristateCorrection()`. The only code calling `doNotUseTristateCorrection()` is in `VectorPairHMMUnitTest` and `PairHMMUnitTest`. git-blame says Adam and @lbergelson added these lines of code. I would say these lines are not needed, since adding these lines means the test behaves differently than the default tool usage. 

BTW, AVX PairHMM in GKL includes the tristate correction and does not have an option to disable it.

src/test/java/org/broadinstitute/hellbender/utils/pairhmm/VectorPairHMMUnitTest.java
erniebrau: Do you mean because we're not testing with useDoublePrecision = true?

droazen: I meant that it's probably worth testing with numThreads != 1 to properly exercise the OMP implementation. Could you submit a separate PR to add a test case that uses 2 threads with the OMP implementation? I'm going to merge this PR now since it's blocking the PR against gatk-protected.

droazen:
:+1: Looks good, merging -- as I suggested above though, we should do a separate PR at some point to add a multithreaded test case for the OMP implementation.

PR #2576, opened 2017-04-06
Modify cluster creation scripts to use ParallelCopyGCSDirectoryIntoHDFSSpark

cwhelan:
This PR removes the use of hadoop distcp to copy reference and sample data into HDFS, and replaces it with the new ParallelCopyGCSDirectoryIntoHDFSSpark tool.

Based on initial tests, cluster creation with reference and a 110GB sample BAM file now takes ~8 minutes with SGA installation on the worker nodes and ~5 minutes without SGA installation.

@SHuang-Broad can you review?

### DIFF
--- a/scripts/sv/create_cluster.sh
+++ b/scripts/sv/create_cluster.sh
@@ -4,19 +4,21 @@
 
 set -eu
 
-if [[ "$#" -lt 5 ]]; then
+if [[ "$#" -lt 6 ]]; then
     echo -e "Please provide:"
-    echo -e "  [1] project name (required)"
-    echo -e "  [2] cluster name (required)"
-    echo -e "  [3] GCS directory containing the correct reference & indices (required)"
-    echo -e "  [4] GCS directory containing the BAM and its index (required), and"
+    echo -e "  [1] local directory of GATK build (required)"
+    echo -e "  [2] project name (required)"
+    echo -e "  [3] cluster name (required)"
+    echo -e "  [4] GCS directory containing the correct reference & indices (required)"
+    echo -e "  [5] GCS directory containing the BAM and its index (required), and"
     echo -e "either when a local initialization script is to be uploaded to a bucket:"
-    echo -e "  [5] local initialization script"
-    echo -e "  [6] GCS path to upload initialization script to (required as per Google)"
+    echo -e "  [6] local initialization script"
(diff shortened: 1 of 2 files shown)


scripts/sv/create_cluster.sh
```diff
@@ -6,17 +6,19 @@ set -eu
 
 if [[ "$#" -lt 5 ]]; then
     echo -e "Please provide:"
```
SHuang-Broad: I believe the total number of required arguments now is >=6.

scripts/sv/default_init.sh
```diff
@@ -18,10 +18,7 @@ ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
 REFLOC=$(/usr/share/google/get_metadata_value attributes/reference)
 SAMPLE=$(/usr/share/google/get_metadata_value attributes/sample)
```
SHuang-Broad: I believe the comment are outdated now, as don't do anything now for the master.
And it would be better if we remove the unused variables `NAME`, `SAMPLE`.

scripts/sv/default_init.sh
```diff
@@ -18,10 +18,7 @@ ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
...
 
-if [[ "${ROLE}" == 'Master' ]]; then
-    hadoop distcp "$REFLOC"/* hdfs://"$NAME":8020/reference/ &
-    hadoop distcp "$SAMPLE"/* hdfs://"$NAME":8020/data/
-else
+if [[ ! "${ROLE}" == 'Master' ]]; then
     if [[ "$REFLOC" =~ .+/$ ]]; then
```
SHuang-Broad: I prefer your new style (my old way was redundant...)

```
if  [[ ! "$REFLOC =~ .+/$ ]]; then
    REFLOC+="/"
fi
mkdir -p /reference && gsutil -m cp "$REFLOC"*.img /reference/
```

cwhelan: changed

scripts/sv/default_init.sh
cwhelan: done

scripts/sv/create_cluster.sh
cwhelan: yep, thanks for catching that

PR #2581, opened 2017-04-07
Update to latest gcloud, uncomment getAuthenticatedGcs

jean-philippe-martin:
Also uncomment getAuthenticatedGcs test and update
code to new way of setting retry params.

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/utils/nio/GcsNioIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/nio/GcsNioIntegrationTest.java
@@ -1,10 +1,6 @@
 package org.broadinstitute.hellbender.utils.nio;
 
-import com.google.cloud.RetryParams;
 import com.google.cloud.storage.StorageException;
-import com.google.cloud.storage.StorageOptions;
-import com.google.cloud.storage.contrib.nio.CloudStorageConfiguration;
-import com.google.cloud.storage.contrib.nio.CloudStorageFileSystem;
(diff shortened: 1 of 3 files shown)


src/test/java/org/broadinstitute/hellbender/utils/nio/GcsNioIntegrationTest.java
```diff
@@ -88,7 +83,7 @@ public void openPrivateFileWithExplicitCredentials() throws IOException {
      * That means that you must NOT set $GOOGLE_APPLICATION_CREDENTIALS
      * Yet you must set getGoogleServiceAccountKeyPath() (you may have to switch it to a different
      * environment variable).
-     *
+     **/
     @Test(enabled = false, groups = {"cloud"}, expectedExceptions = {StorageException.class})
```
lbergelson: Should this be enabled?  Currently it's still set to disabled.

jean-philippe-martin: This should not be enabled, because this is a manual test that relies on an environment that is *not* the test environment. In the test environment, GOOGLE_APPLICATION_CREDENTIALS is set (so we can run the test before this one). To run this test you have to not set GOOGLE_APPLICATION_CREDENTIALS. So we can only run one of these tests at a time, not both. It's there to make it easier to run the manual test if we want to.

droazen:
:+1: merging

jean-philippe-martin:
Great, thank you!

PR #2588, opened 2017-04-08
Exclude htsjdk and protobuf from genomicsdb dependency - Issue 2578

kdatta:
Exclude htsjdk and protobuf from genomicsdb dependency

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,7 +105,9 @@ dependencies {
     testCompile(javadocJDKFiles)
 
     compile 'org.broadinstitute:barclay:1.0.0-24-g87c3fa2-SNAPSHOT'
-    compile 'com.intel:genomicsdb:0.3.0'
+    compile ('com.intel:genomicsdb:0.3.0') {
+      exclude module: 'htsjdk'
+    }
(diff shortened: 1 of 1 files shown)


lbergelson:
@kdatta Is the version of protobuffs that gatk depends on compatible with the version that genomicsDB needs?

kdatta:
No, they are not compatible. We use the stable version 3.0.2 which uses a base class GeneratedMessageV3 whereas the 3.0.0-beta-1 version uses GeneratedMessage. Hence, we need to build separately for gatk. Having said that, our cmake-based build environment makes it very easy to change protobuf versions. So, this is not a problem.

kgururaj:
Adding to Kushal's comment: the Protobuf generated files distributed as part of the GenomicsDB repo (master branch) are produced using Protobuf 3.0.2. If you clone the source and build the GenomicsDB jar, it will require 3.0.2 and probably fail if you build GATK with the built GenomicsDB jar.

However, if you use the jar that we (Intel team) will upload to Maven central, there will be no issues since the uploaded jar was built using Protobuf v3.0.0-beta1.

droazen:
:+1:

PR #2593, opened 2017-04-10
ignore certain contigs when calling inter-contig pairs

tedsharpe:


tedsharpe:
Only cuts the runtime down by 5%, but it produces the same vcf.

scripts/sv/fermiLite/scanBam.sh
```diff
@@ -22,13 +22,15 @@ REF_FASTA="${MASTER_NODE}$5"
 REF_INDEX_IMAGE="$6"
 INTERVAL_KILL_LIST=$(echo "${REF_FASTA}" | sed 's/.fasta$/.kill.intervals/')
 KMER_KILL_LIST=$(echo "${REF_FASTA}" | sed 's/.fasta$/.kill.kmers/')
+ALTS_KILL_LIST=$(echo "${REF_FASTA}" | sed 's/.fasta$/.kill.alts/')
```
cwhelan: The file /seq/references/Homo_sapiens_assembly38/v0/Homo_sapiens_assembly38.fasta.64.alt is missing 170 alt contigs that are in your kill list, though.

tedsharpe: Just for the sake of documentation, even though we've already discussed it:  The missing contigs in the alt file stored in the reference directory do not list contigs that do not share substantial homology with one of chromosomal contigs.  Those aren't a problem for the mapper, so they're omitted.
We'll use the list of all non-chromosomal contigs for our SV cross-contig kill list.

tedsharpe:
@SHuang-Broad Do you want an opportunity to review, or shall I just merge?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -235,10 +249,13 @@ protected void runTool( final JavaSparkContext ctx ) {
             final Locations locations,
             final PipelineOptions pipelineOptions )
     {
+        Set<Integer> crossContigsToIgnoreSet;
+        if ( crossContigsToIgnoreFile != null ) crossContigsToIgnoreSet = Collections.emptySet();
```
tedsharpe: Oh, wow.  So this has never actually been tested.  The 5% variation I saw was just noise.
Thanks, Steve.  I'll make this fix and test before merging.  Thanks for catching this.

SHuang-Broad: I was wondering about this possibility actually :smirk:

SHuang-Broad: Just approved since you've noticed this.

cwhelan:
Yikes, sorry I didn't notice that in my review.

tedsharpe:
Decreases runtime from 55 mins to 47.  Produces about the same number of variant calls, though some of them are a bit different.
I've rebased, squashed, and will merge when Travis and local testing have completed unless someone objects.

cwhelan:
Sounds good to me.

PR #2594, opened 2017-04-12
default output name was not being set correctly

tedsharpe:
getFullPath is similar to dirname -- it's not the right method to call


SHuang-Broad:
👍 . Thanks for catching the bug, @tedsharpe !

PR #2595, opened 2017-04-13
a new tool to run the whole process of SV discovery, 

SHuang-Broad:
by delegating work to 
`FindBreakpointEvidenceSpark` and `DiscoverVariantsFromContigAlignmentsSpark`, 
so we have a single tool running the whole pipeline.

This PR also does:
* refactoring of `FindBreakpointEvidenceSpark` and `DiscoverVariantsFromContigAlignmentsSpark` to accommodate the new tool
* added three integration test (dummy in the sense that it only makes sure they run, and no correctness check on the output) for the three tools

Known differences:

* For NA12878 test sample: The `FindBreakpointEvidenceSpark`->`DiscoverVariantsFromContigAlignmentsSpark` generated VCF and `StructuralVariationDiscoveryPipelineSpark` generated VCF differ by how supplementary alignment's soft clipping is treated. The `FindBreakpointEvidenceSpark`->`DiscoverVariantsFromContigAlignmentsSpark` path has an optimization turned on that soft clipped bases for supplementary alignments are hard clipped away (no contig sequence is lost as it is always saved in the primary alignment), so the CIGARs are a little different. As a consequence, the SAM file generated by the two routes also differ in this CIGAR and sequence part.
* For CHM test sample: The differences are more delicate and even master version yields slightly different results from run to run. So I summarized them in the attached zip.

@tedsharpe and @cwhelan please take a look, as `FindBreakpointEvidenceSpark` is modified (no change of logic, but how code is called).
[chm.zip](https://github.com/broadinstitute/gatk/files/919925/chm.zip)


SHuang-Broad:
Also @vruano , if you feel like it, please see if the integration test is at least resembling what you would like to see.

cwhelan:
@SHuang-Broad can you create two new issues, for 1) the error you saw on one of the runs about the missing bwa index file -- maybe we could verify that it's there on all the nodes or do retries and 2) the variability in the number of kmers found and variants discovered? You can assign the latter one to @tedsharpe .

cwhelan:
@SHuang-Broad where does this optimization around soft clipping come from? How does it get turned on or off?

SHuang-Broad:
@cwhelan . The soft-clip -> hard-clip optimization is in lines 47-63 in `BwaMemAlignmentUtils.applyAlignment()`, which was called by `AlignedAssemblyOrExcuse.writeSAMFile()` in our discovery pipeline (BwaSparkEngine calls it as well but not affecting what we are talking about here).

SHuang-Broad:
Ticket #2600 created for randomness of kmers and templates

cwhelan:
@SHuang-Broad I see. So the conversion to SAM and back when we write the file actually changes the results (or at least their annotations). It makes me a little nervous that in one version of the pipeline the records go through `BwaMemAlignmentUtils.applyAlignment` and in the other they don't, since that method has some complex logic. Right now we have two possible paths:

`AlignedAssemblyOrExcuse -> SAMRecord -> writeToFile -> GATKRead -> AlignmentRegion`

or 

`AlignedAssemblyOrExcuse -> AlignmentRegion`

What if we always converted to `SAMRecord`? It's a little more expensive but it would cut down on alternate code paths and conversion code, and IMO would make the code a lot simpler to read if I didn't have to think about which code path I was in. I'm also worried that the different conversions could lead to bugs that will be hard to debug since you have to know the code path that generated them. @tedsharpe what do you think?

tedsharpe:
The decision about whether to hard clip or soft clip supplementary alignments is a flag to bwa mem.  All it does is to replace initial and final 'S' in the cigar with 'H' instead.  The code in applyAlignment respects that decision.  So if we don't want any hard clipping, that's easy enough to do -- we just turn off that flag.
That's probably the right thing to do since the code on line 89 of the AlignmentAssemblyParser always grabs the entire sequence, whether or not there's been hard clipping.  I'd guess it's likely that there a bugs lurking here.
I don't see any particular need to convert to SAM and then back into an AlignmentRegion.  We can eliminate the SAM writing entirely once we have a single tool, and then there will only be a single path.

scripts/sv/fermiLite/scanBam.sh
```diff
@@ -36,5 +36,5 @@ KMER_KILL_LIST=$(echo "${REF_FASTA}" | sed 's/.fasta$/.kill.kmers/')
     --cluster "${CLUSTER_NAME}" \
     --num-executors 20 \
     --driver-memory 30G \
-    --executor-memory 30G \
+    --executor-memory 40G \
```
tedsharpe: I believe that if you go over 32Gb you get 8-byte references rather than 4.  We should fight hard to stay below 32Gb, unless I'm wrong.

cwhelan:
It would be nice to keep the separate tool around so that it can be used
for stuff like calling variants from de novo assemblies, like I showed at
Tuesday's meeting.

But if both of you think it's not necessary to simplify the conversions
I'll defer to you.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignedAssemblyOrExcuse.java
```diff
@@ -131,7 +131,7 @@ public static void writeSAMFile( final String samFile, final PipelineOptions pOp
                     final List<BwaMemAlignment> alignments = allAlignments.get(contigIdx);
                     if ( alignments.isEmpty() ) continue;
                     final FermiLiteAssembly.Contig contig = assembly.getContig(contigIdx);
-                    final String readName = String.format("asm%06d:tig%05d", assemblyId, contigIdx);
+                    final String readName = formatAssemblyID(assemblyId) + ":" + formatContigID(contigIdx); //String.format("asm%06d:tig%05d", assemblyId, contigIdx);
```
tedsharpe: Vastly less efficient (but called infrequently enough that it doesn't matter).  But I also find the revision less clear, to tell you the truth.

tedsharpe:
I agree that we ought to retain a tool that takes SAM as input and 
produces a VCF.  At this point I'm less concerned with maintaining the 
SGA path, even though that seemed important to me a few days ago.
Maybe if we added the contig sequence to the AlignmentRegion the whole 
thing would be less messy?  We'd just need three converters: from SAM, 
from SGA, or from FermiLite.

scripts/sv/fermiLite/scanBam.sh
SHuang-Broad: Will change back. The reason for this change was to increase the memory so the 8-hanging jobs phenomenon could be temporarily solve by "brute force"

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignedAssemblyOrExcuse.java
SHuang-Broad: If we can be sure that the assembly and contig Ids are going be integers in nature in the foreseeable future, we can change the fields themselves in `AlignmentRegion` to integers. But the non-scientific part, i.e. the name of the reads produced in the (even when optional) SAM file and the annotations in the final VCF must have an agreement or at least have a protocol to understand each other.

SHuang-Broad:
For documentation, thought we have discussed in person:
we all agree that the class `AlignmentRegion` is becoming a bottle neck in communication between the evidence gathering stage and the variant discovery stage. It seems that the comments here are mostly concerned with how this particular class `AlignmentRegion` is created from the output of `FindBreakpointEvidenceSpark`. In memory, that is this PR, it is constructed directly from `BwaMemAlignemt`; in file, which is the way we currently have in master, there's the conversion to SAM and back. I've looked at what is stored in `AlignmentRegion` and what is in `AlignedAssemblyOrExcuse`, the only information missing is a directly stored `CIGAR` object. 

So it makes sense now to retire this `AlignmentRegion` class. Considering it is used extensively in the discovery stage, this is non-trivial work. But it is worth making things correct now. 

Hence closing this PR and the migration starts...

PR #2605, opened 2017-04-14
Replace MannWhitneyU with updated version from GATK3

### LINKED ISSUE #2604: Replace MannWhitneyU with updated version from GATK3
MannWhitneyU was re-written from scratch in 2016 in GATK3, but these changes never got ported to GATK4. This new version produces significantly different results from the version currently in GATK4, resulting in VERY different values for the `RankSumTest` annotations in `HaplotypeCaller` output.

droazen:
MannWhitneyU was re-written from scratch in 2016 in GATK3,
but these changes never got ported to GATK4. This new version
produces significantly different results from the version
currently in GATK4, resulting in VERY different values for the
RankSumTest annotations in HaplotypeCaller output.

@meganshand informs me that the updated GATK3 version has been
validated in R, and has much better tests than the old version.

This is a straightforward port of that version with minimal changes:

-Merged "MWUnitTest" and "RankSumUnitTest" from GATK3 into a single
 test class MannWhitneyUUnitTest
-Ported MathUtils.binomialCoefficient() and wrote new test for it
-Updated RankSumTest class and tests as appropriate

I've confirmed that with this change, the RankSum annotations produced
by the GATK4 HaplotypeCaller closely match those produced by the GATK3
HaplotypeCaller

Resolves #2604

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RankSumTest.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RankSumTest.java
@@ -1,8 +1,10 @@
 package org.broadinstitute.hellbender.tools.walkers.annotator;
 
+import com.google.common.primitives.Doubles;
 import htsjdk.variant.variantcontext.Allele;
 import htsjdk.variant.variantcontext.GenotypesContext;
 import htsjdk.variant.variantcontext.VariantContext;
+import org.apache.commons.lang.ArrayUtils;
 import org.broadinstitute.hellbender.engine.ReferenceContext;
 import org.broadinstitute.hellbender.utils.MannWhitneyU;
 import org.broadinstitute.hellbender.utils.QualityUtils;
@@ -67,12 +69,16 @@ public abstract class RankSumTest extends InfoFieldAnnotation {
             return Collections.emptyMap();
         }
 
+        final MannWhitneyU mannWhitneyU = new MannWhitneyU();
+
         // we are testing that set1 (the alt bases) have lower quality scores than set2 (the ref bases)
-        final double p = MannWhitneyU.runOneSidedTest(useDithering, altQuals, refQuals).getLeft();
-        if (Double.isNaN(p)) {
+        final MannWhitneyU.Result result = mannWhitneyU.test(Doubles.toArray(altQuals), Doubles.toArray(refQuals), MannWhitneyU.TestType.FIRST_DOMINATES);
+        final double zScore = result.getZ();
+
+        if (Double.isNaN(zScore)) {
             return Collections.emptyMap();
         } else {
-            return Collections.singletonMap(getKeyNames().get(0), String.format("%.3f", p));
+            return Collections.singletonMap(getKeyNames().get(0), String.format("%.3f", zScore));
         }
     }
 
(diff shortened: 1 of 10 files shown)


droazen:
@meganshand Could you review (or nominate a reviewer if you're busy)?

droazen:
This PR fixes about 60-70% of the annotation differences in the GATK4 `HaplotypeCaller` as compared to GATK3.

src/test/java/org/broadinstitute/hellbender/utils/MannWhitneyUUnitTest.java
```diff
@@ -1,50 +1,178 @@
...
+                counts++;
+                quals += dist2.remove(0);
+            }
+
+            final Double qual = quals / counts;
+            for ( int i = 0; i < numToReduceIn2; i++ )
+                dist2.add(qual);
+        }
+
+        final Double result = mannWhitneyU.test(Doubles.toArray(distribution1), Doubles.toArray(dist2), MannWhitneyU.TestType.TWO_SIDED).getP();
```
meganshand: It might be overkill, but could you also test the Z-score here, especially since that's what actually gets output for the RankSum annotations?

meganshand: Also, since the only MannWhitney we actually use is `TestType.FIRST_DOMINATES`, would it be worthwhile to test that case instead of `TWO_SIDED`?

lbergelson:
@droazen After discussion with @meganshand, it seems like it would be helpful if we merge this before she starts on the long running parts of the evaluation since it's likely to impact the results.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RankSumTest.java
```diff
@@ -67,12 +69,16 @@ public RankSumTest(){
...
-        final double p = MannWhitneyU.runOneSidedTest(useDithering, altQuals, refQuals).getLeft();
-        if (Double.isNaN(p)) {
+        final MannWhitneyU.Result result = mannWhitneyU.test(Doubles.toArray(altQuals), Doubles.toArray(refQuals), MannWhitneyU.TestType.FIRST_DOMINATES);
+        final double zScore = result.getZ();
+
+        if (Double.isNaN(zScore)) {
             return Collections.emptyMap();
         } else {
-            return Collections.singletonMap(getKeyNames().get(0), String.format("%.3f", p));
+            return Collections.singletonMap(getKeyNames().get(0), String.format("%.3f", zScore));
```
droazen: GATK4 definitely was not outputting p-values previously, since it was usually outputting negative values for the RankSum annotations :)

src/test/java/org/broadinstitute/hellbender/utils/MannWhitneyUUnitTest.java
droazen: Well, we already do have good test cases for the z-score in the `FIRST_DOMINATES` cases -- this is `testOnesidedZ()`. This additional test case in `testDistribution()` was salvaged from GATK3's `RankSumUnitTest` without any effort to generalize it.

Given the time constraints on this branch, and given that we do have test coverage on the combination of options that's actually used to generate the annotations, I'm inclined to leave `testDistribution()` as-is for now. If you feel strongly that it should be expanded, could you create a separate ticket in this repo for that task? Thanks!

droazen:
I've created https://github.com/broadinstitute/gatk/issues/2625 to separately address a possible performance issue introduced by this branch.

droazen:
Merging!

PR #2607, opened 2017-04-18
Includes a new tool to generated BAM extract with the reads that evid…

vruano:
…ence for a number of SV called variants.

It takes as input the variant call vcf, the interval where the variants of interest are, the bam file with the input reads and the location
of the assembly fastsq.

The output is a new BAM file only with the reads that are part of the assemblies that were use to call those variants. These include the original input alignment
data for each read including those pairs that were totally unmapped.

When the evidence reads have primary and secondary alignment there is a possibility that some alternative alingments that are redudant will be present in the output.


cwhelan:
@tedsharpe Would you be able to review this one? I'm not sure if someone from the engine team should also look at the non-SV related changes (cc @droazen )

droazen:
I'll have a look at the engine changes

tedsharpe:
Yes, I'll have a look.

tedsharpe:
I have no objection to this PR.  However, it might be simpler to modify the SV pipeline to optionally produce this data on the fly.

vruano:
@tedsharpe, this tool would normally be used on a handful of called variants, are you suggesting to create small bam files one for each variants with all its evidence instead? that would be much bigger than the input bam (as reads would appear in more than one bam file).

tedsharpe:
Perhaps I'm misunderstanding the intent of your new tool.
SV's are rare, and touch only a small fraction of the reads.  Occasionally nearby events might pull in the same reads, but it seldom happens.
The tool currently has the capability of writing a FASTQ for each assembly (as input to SGA: though we're now using FermiLite by default, the capability remains).  We could write a headerless SAM instead, if you'd like.

vruano:
@droazen, are you still reviewing this?

droazen:
@vruano Yes, sorry! I'll try to have a look at this later today.

vruano:
@droazen once upon a time...

src/main/java/org/broadinstitute/hellbender/utils/LocatableComparator.java
```diff
@@ -0,0 +1,92 @@
...
+ * </p>
+ * <p>
+ *     Also, the {@code null} contig is considered the last contig and so all instances with a non-null contig would come before
+ *     any instances with a {@code null} contig.
+ * </p>
+ * <p>
+ *     The relative order between contig is what is left to be determined by the implementing class by defining {@link #compareContigs(String,String)}.
+ * </p>
+ */
+public abstract class LocatableComparator implements Comparator<Locatable>, Serializable {
```
droazen: Can you explain why you need this new abstract comparator class? The previous lexicographic comparator was 3 lines of code and did the same thing:

```
    public static final Comparator<Locatable> LEXICOGRAPHICAL_ORDER_COMPARATOR =
            Comparator.comparing(Locatable::getContig,Comparator.nullsLast(String::compareTo))
                    .thenComparingInt(Locatable::getStart)
                    .thenComparingInt(Locatable::getEnd);
```

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
```diff
@@ -136,6 +141,76 @@ public void setPosition( final String contig, final int start ) {
...
+                        Array.set(newO, i, Array.get(o, length - i - 1));
+                    }
+                    genomicsRead.set(tag, newO);
+                } else if (o instanceof CharSequence) {
+                    genomicsRead.set(tag, StringUtils.reverse(o.toString()));
+                } else {
+                    throw new UnsupportedOperationException("cannot handle tag reverse on type: " + o.getClass());
+                }
+            }
+        }
```
droazen: Ouch -- you will definitely need unit tests in `GATKReadAdaptersUnitTest` for this...

droazen:
@vruano Sorry for the delay on this branch! I am away this week, but had time this morning to do a quick initial pass over your engine-related changes here. I have some concerns about your additions to the `GATKRead` interface in particular -- we should discuss how best to accomplish your goals there without making the interface confusing or incoherent.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/EvidenceSamForVariantsSparkIntegrationTest.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Created by valentin on 4/7/17.
+ */
+public class EvidenceSamForVariantsSparkIntegrationTest extends CommandLineProgramTest {
```
magicDGS: This should live in `src/test`

src/test/java/org/broadinstitute/hellbender/utils/CollectionUtilsTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.testng.annotations.Test;
+
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Unit tests for {@link CollectionUtils}.
+ */
+public class CollectionUtilsTest extends BaseTest {
```
magicDGS: Should extend `GATKBaseTest` after #3475

droazen:
@vruano Do you still want to get this ancient PR in?

PR #2611, opened 2017-04-21
Fix test assertion that fails on newer versions of testNG.

cmnbroad:
The version of testNG that we're using has some [inconsistencies](https://github.com/cbeust/testng/pull/790) in how nested arrays/collections are handled, and they've resolved them in newer versions by implementing changes that are not backward compatible. The AlleleFrequencyCalculatorUnitTest tests fail with newer versions of testNG that have these changes, since they were wrapping an array in a single element List (not sure if that was intentional or not).

The testNG issues seem to suggest using assertEqualsDeep for [some cases](https://github.com/cbeust/testng/issues/1342) but even the newest version doesn't seem to have an overload for Collections. For AlleleFrequencyCalculatorUnitTest though, the collection wrapping was unnecessary anyway, so the array comparisons can be made directly.


cmnbroad:
@lbergelson FYI - see the links to details of the testNG issues above.

Also, I reversed the actual/expected order in this call.

droazen:
:+1: looks fine, merge when tests pass @cmnbroad

lbergelson:
@cmnbroad Thanks for the links.  

The test as it existed before seems like a typo that happened to work due to coincidental weird behavior in testng.  It's really confusing that `Arrays.asList(int[])` is `List<int[]>` and not `List<Integer>`

PR #2612, opened 2017-04-21
Fix issue in GenomicsDBImportIntegrationTest where GCS test cases were not marked with correct test group

droazen:
Also fix an issue with git lfs and some tabix indices in src/test/resources/large/gvcfs


droazen:
@kdatta or @kgururaj please review

kdatta:
looks good to me.

PR #2617, opened 2017-04-24
Upgrade testNG to 6.11, use force dependency resolution for testNG

cmnbroad:
And remove unnecessary dependencies introduced for GenomicsDB.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,7 +58,7 @@ final htsjdkVersion = System.getProperty('htsjdk.version','2.9.1')
 final sparkVersion = System.getProperty('spark.version', '2.0.2')
 final hadoopBamVersion = System.getProperty('hadoopBam.version','7.8.0')
 final genomicsdbVersion = System.getProperty('genomicsdb.version','0.5.0-proto-3.0.0-beta-1')
-
+final testNGVersion = '6.11'
 
 configurations.all {
     resolutionStrategy {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -58,6 +58,7 @@ final htsjdkVersion = System.getProperty('htsjdk.version','2.9.1')
 final sparkVersion = System.getProperty('spark.version', '2.0.2')
 final hadoopBamVersion = System.getProperty('hadoopBam.version','7.8.0')
 final genomicsdbVersion = System.getProperty('genomicsdb.version','0.5.0-proto-3.0.0-beta-1')
+final testNGVersion = 'org.testng:testng:6.11'
```
droazen: Should be just the version, not the full artifact ID

build.gradle
```diff
@@ -74,6 +75,13 @@ configurations.all {
     all*.exclude group: 'com.google.guava', module: 'guava-jdk5'
 }
 
+configurations.compile {
+    resolutionStrategy {
+        // force testng dependency so we don't pick up a different version via GenomicsDB
+        force testNGVersion
+    }
+}
```
droazen: Move to `configurations.all`

droazen:
Merge after addressing comments

build.gradle
lbergelson: change to just the version here

build.gradle
```diff
@@ -74,6 +75,13 @@ configurations.all {
     all*.exclude group: 'com.google.guava', module: 'guava-jdk5'
 }
 
+configurations.compile {
+    resolutionStrategy {
```
lbergelson: I would move this into the configurations.all block

PR #2627, opened 2017-04-27
interval tree

tedsharpe:


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVIntervalTree.java
```diff
@@ -0,0 +1,904 @@
+package org.broadinstitute.hellbender.tools.spark.sv;
+
```
lbergelson: @tedsharpe I just want to check that you know that there is already an interval tree in htsjdk that you implemented that seems to have very similar functionality.

lbergelson: (there's actually 2 interval trees in htsjdk, one of which you wrote, as well as `OverlapDetector` which is built on your interval tree)

tedsharpe: @lbergelson Yes, very similar functionality, except...
The original is an interval tree on a single integer number line, we need an interval tree with a number line for each contig.  Also, the one in htsjdk has been hacked from the original to use fully closed intervals rather than half-open.  Since we sometimes need to mark locations as 0-length intervals, that's not helpful.
Too bad I didn't make the interval concept an interface when I first wrote it.  Probably no way to reassemble Humpty Dumpty at this point.

lbergelson: I'm curious what you need 0-length intervals for.  Indicating insertions?

tedsharpe: At the moment, the only use is as a sentinel (to mark the end of a stream of intervals).  But sure, your example is a good one.  I'll probably need them when I write the code that tries to piece together breakpoint evidence across partition boundaries, but maybe not.  I don't know.  Just seems useful and consistent to be able to shrink an interval to a point, I guess.

lbergelson: There was a year long argument about this and the gatk closed intervals....  @droazen still bears the scars of the interval wars.

tedsharpe: Yes.  It's surprising that the wrong answer won.  Ha, ha, just kidding.

lbergelson: It's unfortunate it was done this way to start.  Auditing the code for off by one errors if we converted was deemed impossible at this stage of things.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/SVIntervalTreeTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @Test(groups = "spark")
+    public void sizeAndClearTest() {
+        final SVIntervalTree<Integer> testTree = initTree();
+        Assert.assertEquals(testTree.size(), intervals.length);
+        testTree.clear();
+        Assert.assertEquals(testTree.size(), 0);
+    }
+
+    @Test(groups = "spark")
```
cwhelan: Should these be in the "spark" group or the "sv" group?

src/test/resources/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark/SVBreakpointsTest.kmers
```diff
@@ -1,356 +1,345 @@
-AATTATATTCCAGAGTACAAAAGAAATTTGCAGATAGAATCACAGAACCAC
```
cwhelan: Why did these results change? I thought this PR was pretty much just a performance improvement.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointClusterer.java
```diff
@@ -1,61 +1,70 @@
...
-import java.util.function.Function;
 
 /**
  * A class that acts as a filter for breakpoint evidence.
  * It passes only that evidence that is part of a putative cluster.
  */
-public final class BreakpointClusterer implements Function<BreakpointEvidence, Iterator<BreakpointEvidence>> {
+public final class BreakpointClusterer implements Iterator<BreakpointEvidence> {
     private final int minEvidenceCount;
     private final int staleEventDistance;
```
tedsharpe: Sure I worried about it, but I think it's ultimately a lot less data than the kmerization, so now I'm not as worried.  And it's so much simpler this way.  There are on the order of 10^6 reads per partition, a BreakpointEvidence object is on the order of 10^3 bytes, so that's only 10^9 bytes even if every read were evidence.  Don't think it'll blow up.
I've removed staleEventDistance to clean up.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointClusterer.java
```diff
@@ -1,61 +1,70 @@
...
-        this.reportableEntries = new ArrayList<>(2 * minEvidenceCount);
+        this.evidenceTree = new SVIntervalTree<>();
     }
 
-    @Override
-    public Iterator<BreakpointEvidence> apply( final BreakpointEvidence evidence ) {
-        if ( evidence.getContigIndex() != currentContig ) {
-            currentContig = evidence.getContigIndex();
-            locMap.clear();
+    public Iterator<BreakpointEvidence> apply( final Iterator<BreakpointEvidence> evidenceItr ) {
```
tedsharpe: Good point.  I've renamed 'apply' to 'buildTree', made it void and private, and called it from the constructor.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointClusterer.java
```diff
@@ -1,61 +1,70 @@
...
+        while ( evidenceItr.hasNext() ) {
+            final BreakpointEvidence evidence = evidenceItr.next();
+            final SVInterval location = evidence.getLocation();
+            final SVIntervalTree.Entry<List<BreakpointEvidence>> entry = evidenceTree.find(location);
+            if ( entry != null ) entry.getValue().add(evidence);
```
tedsharpe: Pretty sure that Louis said that this was one of our departures from Google style:  single statements following an "if", "else", or "else if" that fit comfortably on the same line are allowed (but not required) to be unbraced.
Since you prefer braces, I'll change these.
However, since you've thrown down the gauntlet, I'm going to start nailing you guys on very long lines (max line length is supposed to be 100 characters).  So there.

cwhelan:
If Louis says this is allowed in our style guide then you can leave them
in. I didn't realize that. Feel free to drop the hammer on us for any style
violations.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -101,71 +96,26 @@ protected BreakpointEvidence( final Kryo kryo, final Input input ) {
...
-        return result;
-    }
-
-    @Override
     public String toString() {
-        return contigIndex + "[" + eventStartPosition + ":" + getContigEnd() + "] " + templateName + templateEnd;
-    }
-
-    @Override
-    public boolean equals( final Object obj ) {
```
tedsharpe: It's very hard to write proper comparison/equality functions for hierarchies in which sub-classes add aspect.  Some argue that it's impossible.  I had been using BreakpointEvidence instances as keys in a sorted tree and needed to hack something that worked, but it wasn't pretty.
I now feel that the whole mess was created by a misanalysis:  By including genomic location fields directly in the class, I was, in effect, asserting that BreakpointEvidence IsA location.  Things have gotten cleaner because now BreakpointEvidence HasA location, and BreakpointEvidence is no longer used as a key.
I eliminated what I felt was an ugly hack.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVIntervalTree.java
```diff
@@ -0,0 +1,904 @@
...
+        }
+
+        private void setMaxEnd() {
+            maxEndInterval = interval;
+            if ( left != null ) maxEndInterval = laterEndingInterval(maxEndInterval, left.maxEndInterval);
+            if ( right != null ) maxEndInterval = laterEndingInterval(maxEndInterval, right.maxEndInterval);
+        }
+
+        private static SVInterval laterEndingInterval( final SVInterval interval1, final SVInterval interval2 ) {
+            final int contig1 = interval1.getContig();
```
tedsharpe: Unfortunately, no.  The natural order comparison is implemented on {contig, start, end} in that order.  Here I'm implementing the sort order {contig, end}.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVIntervalTree.java
```diff
@@ -0,0 +1,904 @@
...
+            while ( mom != null && !mom.isBlack ) {
+                final Node<V1> gramma = mom.parent;
+                Node<V1> auntie = gramma.left;
+                if ( auntie == mom ) {
+                    auntie = gramma.right;
+                    if ( auntie != null && !auntie.isBlack ) {
+                        mom.isBlack = true;
+                        auntie.isBlack = true;
+                        gramma.isBlack = false;
+                        daughter = gramma;
```
tedsharpe: Always makes me giggle a little when I read it.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/SVIntervalTreeTest.java
tedsharpe: I haven't quite figured out our policy.  I saw some other code that had introduced an "sv" group, and thought that made sense.  But I guess I haven't learned to use "sv" consistently.
We'd like to group all our test methods in the future under group "sv"?  Is that right?
I'll change these to "sv".

src/test/resources/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark/SVBreakpointsTest.kmers
tedsharpe: I found and fixed an off-by-one error in SVInterval::overlaps.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointClusterer.java
lbergelson: We've typically allowed single line if statements without braces if they are short and reasonable like this.  I personally prefer braces and try to always have them even on single liners.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVIntervalTree.java
```diff
@@ -0,0 +1,904 @@
...
+                        fixup(parent);
+
+                        if ( isBlack )
+                            root = removeFixup(parent, null, root);
+                    } else {
+                        parent.right = null;
+                        fixup(parent);
+
+                        if ( isBlack )
+                            root = removeFixup(parent, null, root);
```
lbergelson: This one needs braces.  We've had to many easy to miss bugs from someone coming in and adding another line in the indented block and forgetting to add braces.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointClusterer.java
cwhelan: OK, as long as you thought about it. Definitely makes things much simpler without it.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointClusterer.java
cwhelan: How about a compromise @tedsharpe -- single line if statements are fine, but it always looks a little funny to me if it's an if/else and the else has braces but the if doesn't, as in a couple of the other examples in this PR.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/SVIntervalTreeTest.java
lbergelson: The groups were well intentioned but not widely used.  We used to have a spark group because spark needed to run specially and we wanted to be able to sometimes allow spark failures on travis.  We no longer allow that though. The only thing we actually use test groups for right now is to separate tests that require login credentials from ones that don't.  (The `cloud` tests.)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
cwhelan: OK, makes sense.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointClusterer.java
tedsharpe: Seems like most people prefer braces.  I'll try to remember to stick to that style.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVIntervalTree.java
cwhelan: ah, sorry, i missed that

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointClusterer.java
tedsharpe: @cwhelan With regard to the compromise:  Maybe I'll do the unbraced thing in this type of circumstance alone, if everyone's jiggy with it:
```
for ( loop stuff ) {
    if ( some condition ) break;
    if ( some other condition ) return;
    if ( a whole nother condition ) continue;
}
```
I.e., one word, keyword statements for loop control.
P.S.:  I've always thought that this looked jazzy and cool:
```
if ( condition ) some_simple_statement;
else {
    something_more_complex;
}
```
Oh, well.

tedsharpe:
I think I've covered the PR comments.  Eager to hear how we stand now.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVIntervalTree.java
tedsharpe: Imported antique code.  Sorry I failed to revise to new style standards.

tedsharpe:
rebased and squashed.  will merge when Travis completes.

tedsharpe:
gonna have to assume that travis works.  gotten aced out twice already with other merges to master.

Issue #2586, opened 2017-04-07
fix FS annotation difference with GATK3


lbergelson:
The FS annotation is still different than GATK3 in some cases.  
The task is to:
1. identify which version of the annotation is correct, if either
2. fix it in either gatk4 or gatk3.

meganshand:
It seems that there are some small numerical differences in apache commons Math3 Hypergeometric distribution.

In GATK3 we call the Hypergeometric distribution to get the probability directly, in GATK4 we call the Hypergeometric distribution to get the log of the probability. In these cases there are sometimes some very small rounding errors in the 15th or so least significant digit. The problem is that we either add these numbers to the p-value or not by checking if they are less than or equal to some threshold. In this case these rounding errors can have a large impact (especially when the number of reads is relatively small) on the final p-value.

R seems to have more digits and therefore usually reports that they are the same values, so in this case neither GATK3 or GATK4 match R. I haven't tried other implementations.

meganshand:
On second thought there is an incredibly simple solution that R uses by including everything past 10e-7 as equal or more extreme.

PR #2629, opened 2017-04-27
Fixes rounding errors in FS using the same solution as R

meganshand:
Addresses #2586

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/FisherExactTest.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/FisherExactTest.java
@@ -11,6 +11,7 @@ import java.util.stream.DoubleStream;
  * assuming the null hypothesis of odd ratio of 1.
  */
 public final class FisherExactTest {
+    private static final double REL_ERR = 1 - 10e-7;
 
     /**
      * Computes the 2-sided pvalue of the Fisher's exact test on a normalized table that ensures that the sum of
@@ -37,7 +38,7 @@ public final class FisherExactTest {
 
         final AbstractIntegerDistribution dist = new HypergeometricDistribution(null, m+n, m, k);
         final double[] logds = support.mapToDouble(dist::logProbability);
-        final double threshold = logds[x[0][0] - lo];
+        final double threshold = logds[x[0][0] - lo] * REL_ERR;
         final double[] log10ds = DoubleStream.of(logds).filter(d -> d <= threshold).map(MathUtils::logToLog10).toArray();
         final double pValue = MathUtils.sumLog10(log10ds);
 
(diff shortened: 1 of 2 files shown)


meganshand:
@lbergelson Could you please take a look?

src/main/java/org/broadinstitute/hellbender/utils/FisherExactTest.java
```diff
@@ -37,7 +38,7 @@ public static double twoSidedPValue(final int[][] normalizedTable) {
 
         final AbstractIntegerDistribution dist = new HypergeometricDistribution(null, m+n, m, k);
         final double[] logds = support.mapToDouble(dist::logProbability);
-        final double threshold = logds[x[0][0] - lo];
+        final double threshold = logds[x[0][0] - lo] * REL_ERR;
```
lbergelson: So I have no idea if this is mathematically supportable, but if it makes us match with R, that sounds good...

lbergelson: I'm assuming that this means we will be different from gatk3 though?

src/test/java/org/broadinstitute/hellbender/utils/FisherExactTestUnitTest.java
```diff
@@ -26,6 +26,8 @@
         tests.add(new Object[]{1, 2, 3, 4,                     1.0});
         tests.add(new Object[]{0, 0, 100000, 100000,           1.0});
         tests.add(new Object[]{100000,100000,100000,0,         0.0});             //below R's or Java's precision
+        tests.add(new Object[]{12, 4, 26, 7,                   1.0});
```
lbergelson: Could you add comments next to these cases explaining what they're testing.

lbergelson:
@meganshand 1 very minor comment about the tests.  👍 After that.  This is awesome to find such a simple solution.

src/test/java/org/broadinstitute/hellbender/utils/FisherExactTestUnitTest.java
droazen: Also, do these new test cases fail before the fix?

droazen:
I suspect that this also explains the `FS` differences that palantir found in the `HaplotypeCaller`

src/test/java/org/broadinstitute/hellbender/utils/FisherExactTestUnitTest.java
meganshand: One of them does (the second one), I added both because the reversed one fails in GATK3.

src/main/java/org/broadinstitute/hellbender/utils/FisherExactTest.java
meganshand: Yes, different from GATK3 unless we add the fix there.

droazen:
@meganshand I assume you'll want an updated gatk-protected snapshot that incorporates this change (will require us to update some expected test outputs). Do you anticipate any other patches to public of this nature before you kick off the official runs of the tool?

meganshand:
@droazen I will want an updated gatk-protected snapshot in a docker image (I can use the private docker images in gatk-protected on dockerhub). I don't know of any other patches we'll need at this point.

lbergelson:
@meganshand Are you also making this change in gatk3?  
@droazen Do we want to update the expected test outputs if they will not match gatk3?

droazen:
Thanks @meganshand -- I'll work on updating gatk-protected.

PR #2630, opened 2017-04-27
minor fixes to address missed review comments from pr #2626

lbergelson:
since #2626 was merged before review comments were addressed, this fixes
the issues raised in those comments
closes #2628

PR #2626, opened 2017-04-26
updating GenomicsDB integration to match the changes in the importer

lbergelson:
* added a reference parameter to FeatureData source and FeatureManager methods
  * genomicsDB requires a reference, previously this was being passed
  through by hardcoding it in the required json files
  * json files are now autogenerated by the importer tool, but the
  reference wasn't being handled correctly

* updated the various walkers to pass the reference through if available

* gendb:// paths now point to the workspace directory instead of a
directory of jsons

* removed the ability to specify array, vidmap.json, and
callset.json paths in the importer tool since we now rely on the
structure and naming of the files when loading
moved some constants to GenomicsDBConstants

* updated GenomicsDBIntegration tests to use the new importer instead of a
prepackaged and very brittle set of json files

fixed a bug in GenomicsDBImporterIntegrationTests that made both tests
write to the same workspace

PR #2631, opened 2017-04-27
Minor fixes to address pr #2626 comments, and extract genomics db test utils again

lbergelson:
@droazen I realized I should probably have exposed these methods for protected since we'd need to duplicated them in order to do broadinstitute/gatk-protected#1013

this includes the changes in #2630


droazen:
Looks good, merging

PR #2634, opened 2017-04-28
Added batchsize and consolidation parameters to GenomicsDBImporter

kdatta:
build will break because genomicsdb-0.6.0 hasn't been released yet and dependence on protobuf-java-format needs to be fixed!

@droazen, remember now that I added the dependence on protobuf-java-format to use protobuf.JsonFormat.printToString() method which converts a protobuf structure to JSON string. I want to use import configuration protocol buffers in this code which means this dependence will be back to bite us! Need to fix this cause I don't want to break the Spark build again

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -1,6 +1,8 @@
 package org.broadinstitute.hellbender.tools.genomicsdb;
 
 import com.intel.genomicsdb.ChromosomeInterval;
+import com.intel.genomicsdb.GenomicsDBCallsetsMapProto;
+import com.intel.genomicsdb.GenomicsDBImportConfiguration;
 import com.intel.genomicsdb.GenomicsDBImporter;
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.tribble.AbstractFeatureReader;
@@ -24,9 +26,11 @@ import org.broadinstitute.hellbender.utils.nio.SeekableByteChannelPrefetcher;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.FileNotFoundException;
 import java.nio.channels.SeekableByteChannel;
 import java.util.*;
 import java.util.function.Function;
+
 
 /**
  * This tool imports GVCFs to GenomicsDB. To run this tool,
@@ -47,6 +51,7 @@ public final class GenomicsDBImport extends GATKTool {
 
     private static final long DEFAULT_VCF_BUFFER_SIZE_PER_SAMPLE = 16*1024L;
     private static final long DEFAULT_SEGMENT_SIZE = 1048576L;
+    private static final int DEFAULT_ZERO_BATCH_SIZE = 0;
 
     public static final String WORKSPACE_ARG_NAME = "genomicsDBWorkspace";
     private static final String SEGMENT_SIZE_ARG_NAME = "genomicsDBSegmentSize";
@@ -57,6 +62,8 @@ public final class GenomicsDBImport extends GATKTool {
     private static final String VID_MAP_FILE_ARG_NAME = "genomicsDBVidMapFile";
 
     private static final String CALLSET_MAP_FILE_ARG_NAME = "genomicsDBCallsetMapFile";
+    private static final String BATCHSIZE_ARG_NAME = "batchSize";
+    private static final String CONSOLIDATE_ARG_NAME = "consolidate";
 
     @Argument(fullName = WORKSPACE_ARG_NAME,
               shortName = WORKSPACE_ARG_NAME,
@@ -81,7 +88,7 @@ public final class GenomicsDBImport extends GATKTool {
               shortName = StandardArgumentDefinitions.VARIANT_SHORT_NAME,
               doc = "GVCF files to be imported to GenomicsDB. Each file must contain" +
                     "data for only a single sample")
-    private List<String> variantFilePaths;
+    private List<String> variantPaths;
 
     @Argument(fullName = VCF_BUFFER_SIZE_ARG_NAME,
               shortName = VCF_BUFFER_SIZE_ARG_NAME,
@@ -116,6 +123,28 @@ public final class GenomicsDBImport extends GATKTool {
               optional = true)
     private Boolean overwriteExistingWorkspace = false;
 
+    @Argument(fullName = BATCHSIZE_ARG_NAME,
+              shortName = BATCHSIZE_ARG_NAME,
+              doc = "Batch size controls the number of samples for which readers are open at once " +
+                    "and therefore provides a way to minimize memory consumption. However, it can take longer to complete. " +
+                    "Use the consolidate flag if more than a hundred batches were used. This will improve feature read time. " +
+                    "batchSize=0 means no batching (i.e. readers for all samples will be opened at once) " +
+                    "Defaults to " + DEFAULT_ZERO_BATCH_SIZE,
+              optional = true)
+    private int batchSize = DEFAULT_ZERO_BATCH_SIZE;
+
+    @Argument(fullName = CONSOLIDATE_ARG_NAME,
+              shortName = CONSOLIDATE_ARG_NAME,
+              doc = "Boolean flag to enable consolidation. If importing data in batches, a new fragment is created for " +
+                    "each batch. In case thousands of fragments are created, GenomicsDB feature readers will try " +
+                    "to open ~20x as many files. Also, internally GenomicsDB would consume more memory to maintain " +
+                    "bookkeeping data from all fragments. Use this flag to merge all fragments into one. " +
+                    "Merging can potentially improve read performance, however overall benefit might not be noticeable " +
+                    "as the top Java layers has significantly higher overheads. This flag have no effect if only one " +
+                    "batch is used. Defaults to false",
+              optional = true)
+    private Boolean doConsolidation = false;
+
     @Override
     public boolean requiresIntervals() { return true; }
 
@@ -127,19 +156,30 @@ public final class GenomicsDBImport extends GATKTool {
       return false;
     }
 
-    // The GenomicsDB importer object
-    private GenomicsDBImporter importer = null;
-
     // Intervals from command line (singleton for now)
     private List<ChromosomeInterval> intervals;
 
-    private final Map<String, FeatureReader<VariantContext>> sampleToVCFMap = new HashMap<>();
+    // List of all sample names
+    private List<String> sampleNames = new ArrayList<>();
+
+    // Linked hash map between sample names and corresponding GVCF file name
+    private Map<String, String> sampleNameToVcfUri = new LinkedHashMap<>();
 
     // Needed as smartMergeHeaders() returns a set of VCF header lines
     private Set<VCFHeaderLine> mergedHeader = null;
 
-    // VCFHeader object created from mergedHeader
-    private VCFHeader mergedVCFHeader = null;
+    // VCFHeader created from the merged header
+    private VCFHeader mergedVCFHeader;
+
+    // Path to vidmap file to be written by GenomicsDBImporter
+    private File vidMapJSONFile;
+
+    // Path to callsetmap file to be written by GenomicsDBImporter
+    private File callsetMapJSONFile;
+
+    // GenomicsDB callset map protobuf structure containing all callset names
+    // used to write the callset json file on traversal success
+    private GenomicsDBCallsetsMapProto.CallsetMappingPB callsetMappingPB;
 
     /**
      * Before traversal starts, create the feature readers
@@ -149,26 +189,38 @@ public final class GenomicsDBImport extends GATKTool {
     @Override
     public void onStartup() {
 
-        for (final String variantPath : variantFilePaths) {
+        List<VCFHeader> headers = new ArrayList<>(variantPaths.size());
 
-            final String variantURI = IOUtils.getPath(variantPath).toAbsolutePath().toUri().toString();
-            final Function<SeekableByteChannel, SeekableByteChannel> cloudWrapper = (cloudPrefetchBuffer > 0 ? is -> SeekableByteChannelPrefetcher.addPrefetcher(cloudPrefetchBuffer, is) : Function.identity());
-            final Function<SeekableByteChannel, SeekableByteChannel> cloudIndexWrapper = (cloudIndexPrefetchBuffer > 0 ? is -> SeekableByteChannelPrefetcher.addPrefetcher(cloudIndexPrefetchBuffer, is) : Function.identity());
-            final AbstractFeatureReader<VariantContext, LineIterator> reader = AbstractFeatureReader.getFeatureReader(variantURI, null, new VCFCodec(), true, cloudWrapper, cloudIndexWrapper);
+        for (String variantPath : variantPaths) {
+
+            final AbstractFeatureReader<VariantContext, LineIterator> reader = getReaderFromVCFUri(variantPath);
+            VCFHeader header = (VCFHeader)reader.getHeader();
 
             // A GVCF file must contain only one sample, throw an exception otherwise
-            if (((VCFHeader)reader.getHeader()).getNGenotypeSamples() != 1) {
+            if (header.getNGenotypeSamples() != 1) {
                 throw new UserException("Input GVCF: " + variantPath + " should contain data for one sample");
             }
+            String sampleName = header.getGenotypeSamples().get(0);
+            headers.add(header);
+            sampleNames.add(sampleName);
+            if (sampleNameToVcfUri.put(sampleName, variantPath) != null) {
+                throw new UserException("Duplicate sample " + sampleName);
+            }
 
-            final String sampleName = ((VCFHeader) reader.getHeader()).getGenotypeSamples().get(0);
-            sampleToVCFMap.put(sampleName, reader);
+            // Import occurs in batches. By closing all readers here we release resources
+            // and preserve memory for later stages where readers are allocated
+            // for samples in a given batch
+            try {
+                reader.close();
+            } catch (IOException e) {
+                throw new GATKException("FeatureReader close() failed for " + sampleName, e);
+            }
         }
 
-        createMergedHeader();
+        mergedHeader = VCFUtils.smartMergeHeaders(headers, true);
+        mergedVCFHeader = new VCFHeader(mergedHeader);
 
         initializeIntervals();
-
         super.onStartup();
     }
 
@@ -184,35 +236,26 @@ public final class GenomicsDBImport extends GATKTool {
         // If provided buffer size is less 1KB it is considered to be too small
         // Either use default 16KB or any value larger than 10KB
         if (vcfBufferSizePerSample < 1024L) {
-            logger.warn("Buffer size per column partition per sample is too small." +
-                " Either use default value " + DEFAULT_VCF_BUFFER_SIZE_PER_SAMPLE +
-                " or larger values than 10KB");
+            throw new UserException("Buffer size per column partition per sample is too small." +
+                    " Either use default value " + DEFAULT_VCF_BUFFER_SIZE_PER_SAMPLE +
+                    " or larger values than 10KB");
         }
 
-        final File vidMapJSONFile = (vidMapJSONFileName.equals(GenomicsDBConstants.DEFAULT_VIDMAP_FILE_NAME)) ?
+        vidMapJSONFile = (vidMapJSONFileName.equals(GenomicsDBConstants.DEFAULT_VIDMAP_FILE_NAME)) ?
             new File(workspaceDir + "/" + vidMapJSONFileName) :
             new File(vidMapJSONFileName);
 
-        final File callsetMapJSONFile = (callsetMapJSONFileName.equals(GenomicsDBConstants.DEFAULT_CALLSETMAP_FILE_NAME)) ?
+        callsetMapJSONFile = (callsetMapJSONFileName.equals(GenomicsDBConstants.DEFAULT_CALLSETMAP_FILE_NAME)) ?
             new File (workspaceDir + "/" + callsetMapJSONFileName) :
             new File(callsetMapJSONFileName);
 
         logger.info("Vid Map JSON file will be written to " + vidMapJSONFile);
         logger.info("Callset Map JSON file will be written to " + callsetMapJSONFile);
-        
-        final long variantContextBufferSize = vcfBufferSizePerSample * sampleToVCFMap.size();
+        logger.info("Importing to array - " + workspace + "/" + arrayName);
 
-        logger.info("Writing data to array - " + workspace + "/" + arrayName);
-
-        try {
-            logger.info("GenomicsDB import batch started");
-            importer = new GenomicsDBImporter(sampleToVCFMap, mergedHeader, intervals.get(0), workspace,
-                                              arrayName, variantContextBufferSize, segmentSize,
-                                              vidMapJSONFile.getAbsolutePath(),
-                                              callsetMapJSONFile.getAbsolutePath());
-        } catch (IOException e) {
-            throw new UserException("Error initializing GenomicsDBImporter", e);
-        }
+        // Passing in false here so that sample names will be sorted.
+        // This is needed for consistent ordering across partitions/machines
+        callsetMappingPB = GenomicsDBImporter.generateSortedCallSetMap(sampleNames, false);
     }
 
     /**
@@ -221,24 +264,172 @@ public final class GenomicsDBImport extends GATKTool {
      */
     @Override
     public void traverse() {
-        try {
-            importer.importBatch();
-        } catch (IOException e) {
-            throw new UserException("GenomicsDB importBatch raised exception", e);
+
+        int updatedBatchSize = (batchSize == DEFAULT_ZERO_BATCH_SIZE) ? sampleNames.size() : batchSize;
+        int sampleCount = sampleNames.size();
+        int totalBatchCount = (sampleCount/updatedBatchSize) + (sampleCount%updatedBatchSize==0 ? 0 : 1);
+
+        GenomicsDBImporter importer;
+
+        for (int i = 0, batchCount = 1; i < sampleCount; i += updatedBatchSize, ++batchCount) {
+
+            Map<String, FeatureReader<VariantContext>> sampleToReaderMap =
+                getFeatureReaders(sampleNames, sampleNameToVcfUri, updatedBatchSize, sampleCount, i);
+
+            logger.info("Importing batch " + batchCount + " with " + sampleToReaderMap.size() + " samples");
+            final long variantContextBufferSize = vcfBufferSizePerSample * sampleToReaderMap.size();
+
+            GenomicsDBImportConfiguration.ImportConfiguration importConfiguration =
+                    createImportConfiguration(workspace, arrayName,
+                            variantContextBufferSize, segmentSize,
+                            i, (i+updatedBatchSize-1));
+
+            try {
+                importer = new GenomicsDBImporter(sampleToReaderMap, mergedHeader, intervals.get(0), importConfiguration);
+            } catch (IOException e) {
+                throw new UserException("Error initializing GenomicsDBImporter in batch " + batchCount, e);
+            }
+
+            try {
+                importer.importBatch();
+            } catch (IOException e) {
+                throw new UserException("GenomicsDB import failed in batch " + batchCount, e);
+            }
+
+            closeReaders(sampleToReaderMap);
+            logger.info("Done importing batch " + batchCount + "/" + totalBatchCount);
         }
     }
 
     @Override
     public Object onTraversalSuccess() {
-        logger.info("Import completed!");
-        return importer.isDone();
+        if (batchSize==DEFAULT_ZERO_BATCH_SIZE) {
+            logger.info("Import completed!");
+        } else {
+            logger.info("Import of all batches to GenomicsDB completed!");
+        }
+
+        // Write the vid and callset map JSON files
+        try {
+            GenomicsDBImporter.writeVidMapJSONFile(vidMapJSONFile.getAbsolutePath(), mergedHeader);
+        } catch (FileNotFoundException fe) {
+            throw new UserException("Unable to write vid map JSON file " + vidMapJSONFile.getAbsolutePath(), fe);
+        }
+        try {
+            GenomicsDBImporter.writeCallsetMapJSONFile(callsetMapJSONFile.getAbsolutePath(), callsetMappingPB);
+        } catch (FileNotFoundException fe) {
+            throw new UserException("Unable to write callset map JSON file " + callsetMapJSONFile.getAbsolutePath(), fe);
+        }
+
+        if (doConsolidation) {
+            logger.info("GenomicsDB consolidation started");
+            GenomicsDBImporter.consolidateTileDBArray(workspace, arrayName);
+            logger.info("GenomicsDB consolidation completed");
+        }
+
+        return true;
     }
 
-    @Override
-    public void closeTool() {
-        for (Map.Entry<String, FeatureReader<VariantContext>> reader : sampleToVCFMap.entrySet()) {
+    /**
+     * Method to create feature readers for input files or GCS URLs
+     * in the current batch
+     *
+     * @param sampleNames  List of all sample names
+     * @param sampleNameToFileName  Sample name to file name mapping
+     * @param batchSize  Current batch size
+     * @param sampleCount  total samples in this import
+     * @param lowerSampleIndex  0-based Lower bound of sample index -- inclusive
+     * @return  Feature readers to be imported in the current batch
+     */
+    private Map<String, FeatureReader<VariantContext>> getFeatureReaders(List<String> sampleNames, Map<String, String> sampleNameToFileName, int batchSize, int sampleCount, int lowerSampleIndex) {
+        Map<String, FeatureReader<VariantContext>> sampleToReaderMap = new LinkedHashMap<>();
+
+        for(int i = lowerSampleIndex; i < sampleCount && i < lowerSampleIndex+batchSize; ++i) {
+            final String sampleName = sampleNames.get(i);
+            assert sampleNameToFileName.containsKey(sampleName);
+
+            final AbstractFeatureReader<VariantContext, LineIterator> reader = getReaderFromVCFUri(sampleNameToFileName.get(sampleName));
+
+            assert sampleName.equals(((VCFHeader) reader.getHeader()).getGenotypeSamples().get(0));
+            sampleToReaderMap.put(sampleName, reader);
+        }
+        return sampleToReaderMap;
+    }
+
+    /**
+     * Creates a feature reader object from a given VCF URI (can also be
+     * a local file path) and returns it
+     * @param variantPath  URI or file path
+     * @return  Feature reader
+     */
+    private AbstractFeatureReader<VariantContext, LineIterator> getReaderFromVCFUri(String variantPath) {
+        final String variantURI = IOUtils.getPath(variantPath).toAbsolutePath().toUri().toString();
+        final Function<SeekableByteChannel, SeekableByteChannel> cloudWrapper = (cloudPrefetchBuffer > 0 ? is -> SeekableByteChannelPrefetcher.addPrefetcher(cloudPrefetchBuffer, is) : Function.identity());
+        final Function<SeekableByteChannel, SeekableByteChannel> cloudIndexWrapper = (cloudIndexPrefetchBuffer > 0 ? is -> SeekableByteChannelPrefetcher.addPrefetcher(cloudIndexPrefetchBuffer, is) : Function.identity());
+        return AbstractFeatureReader.getFeatureReader(variantURI, null, new VCFCodec(), true, cloudWrapper, cloudIndexWrapper);
+    }
+
+    /**
+     * Creates a GenomicsDB configuration data structure
+     * instead of sending a long list of parameters to the constructor call
+     *
+     * @param workspace  GenomicsDB workspace
+     * @param arrayName  GenomicsDB array
+     * @param variantContextBufferSize  Buffer size to store VCF records for all samples
+     * @param segmentSize  Buffer size to store columnar data to be serialized to disk
+     * @param lbSampleIndex  Lower bound of sample index -- inclusive (0-based)
+     * @param ubSampleIndex  Upper bound of sample index -- inclusive (0-based)
+     * @return  GenomicsDB import configuration object
+     */
+    private GenomicsDBImportConfiguration.ImportConfiguration createImportConfiguration(
+        final String workspace,
+        final String arrayName,
+        final long variantContextBufferSize,
+        final long segmentSize,
+        final long lbSampleIndex,
+        final long ubSampleIndex) {
+
+        GenomicsDBImportConfiguration.Partition.Builder pBuilder =
+            GenomicsDBImportConfiguration.Partition.newBuilder();
+
+        // Since, there is one partition for this import, the
+        // begin column partition index is 0
+        GenomicsDBImportConfiguration.Partition partition =
+            pBuilder
+                .setWorkspace(workspace)
+                .setArray(arrayName)
+                .setBegin(0)
+                .build();
+
+        GenomicsDBImportConfiguration.GATK4Integration.Builder gBuilder =
+            GenomicsDBImportConfiguration.GATK4Integration.newBuilder();
+
+        GenomicsDBImportConfiguration.GATK4Integration gatk4Parameters =
+            gBuilder
+                .setLowerSampleIndex(lbSampleIndex)
+                .setUpperSampleIndex(ubSampleIndex)
+                .build();
+
+        GenomicsDBImportConfiguration.ImportConfiguration.Builder cBuilder =
+            GenomicsDBImportConfiguration.ImportConfiguration.newBuilder();
+
+        return cBuilder
+                .addColumnPartitions(0, partition)
+                .setGatk4IntegrationParameters(gatk4Parameters)
+                .setSizePerColumnPartition(variantContextBufferSize)
+                .setSegmentSize(segmentSize)
+                .build();
+    }
+
+    /**
+     * Close all readers in the current batch
+     *
+     * @param sampleToReaderMap  Map of sample names to readers
+     */
+    private void closeReaders(Map<String, FeatureReader<VariantContext>> sampleToReaderMap) {
+        for (Map.Entry<String, FeatureReader<VariantContext>> reader : sampleToReaderMap.entrySet()) {
             try {
-               reader.getValue().close();
+                reader.getValue().close();
             } catch (IOException e) {
                 throw new GATKException("FeatureReader close() failed for " + reader.getKey(), e);
             }
@@ -280,21 +471,6 @@ public final class GenomicsDBImport extends GATKTool {
         if (!tempFile.exists()) {
             throw new UserException(workspaceDir.getAbsolutePath() + " is not a valid GenomicsDB workspace");
         }
-    }
-
-    /**
-     * Create merged header from input GVCFs
-     */
-    private void createMergedHeader() {
-        List<VCFHeader> headers = new ArrayList<>();
-        for (Map.Entry<String, FeatureReader<VariantContext>> sampleToVCF : sampleToVCFMap.entrySet()) {
-            FeatureReader<VariantContext> reader = sampleToVCF.getValue();
-            VCFHeader vcfHeader = (VCFHeader)reader.getHeader();
-            headers.add((VCFHeader)reader.getHeader());
-        }
-
-        mergedHeader = VCFUtils.smartMergeHeaders(headers, true);
-        mergedVCFHeader = new VCFHeader(mergedHeader);
     }
 
     /**
--- a/src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
@@ -9,6 +9,7 @@ import htsjdk.variant.bcf2.BCF2Codec;
 import htsjdk.variant.variantcontext.VariantContext;
 import htsjdk.variant.vcf.VCFCodec;
 import org.broadinstitute.hellbender.CommandLineProgramTest;
+import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.tools.genomicsdb.GenomicsDBImport;
 import org.broadinstitute.hellbender.tools.genomicsdb.GenomicsDBConstants;
 import org.broadinstitute.hellbender.utils.IntervalUtils;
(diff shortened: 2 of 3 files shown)


droazen:
@kdatta The Spark build broke last time because of a jar signature file from the `gnu.getopt` dependency that made it into our final GATK jar, not because of `protobuf-java-format`. We've now excluded these signature files from our jar.

When will `genomicsdb-0.6.0` be released?

kdatta:
I see, then it seems safe to use the protobuf java format. We intend to release the 0.6.0 version asap, preferably today.

kdatta:
@kgururaj tagging for review

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -149,26 +170,46 @@ public boolean requiresReference() {
...
-        for (final String variantPath : variantFilePaths) {
+        List<VCFHeader> headers = new ArrayList<>(variantFilePaths.size());
 
-            final String variantURI = IOUtils.getPath(variantPath).toAbsolutePath().toUri().toString();
+        for (String variantFilePath : variantFilePaths) {
+
+            final String variantURI = IOUtils.getPath(variantFilePath).toAbsolutePath().toUri().toString();
             final Function<SeekableByteChannel, SeekableByteChannel> cloudWrapper = (cloudPrefetchBuffer > 0 ? is -> SeekableByteChannelPrefetcher.addPrefetcher(cloudPrefetchBuffer, is) : Function.identity());
             final Function<SeekableByteChannel, SeekableByteChannel> cloudIndexWrapper = (cloudIndexPrefetchBuffer > 0 ? is -> SeekableByteChannelPrefetcher.addPrefetcher(cloudIndexPrefetchBuffer, is) : Function.identity());
             final AbstractFeatureReader<VariantContext, LineIterator> reader = AbstractFeatureReader.getFeatureReader(variantURI, null, new VCFCodec(), true, cloudWrapper, cloudIndexWrapper);
```
kgururaj: There seems to be some code duplication above and in getFeatureReaders() - perhaps we can extract the functionality into a common function getFeatureReader(String variantFilePath) and call it from both places.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+
+
+                try {
+                    importer.importBatch();
+                    assert importer.isDone();
+                } catch (IOException e) {
+                    throw new UserException("GenomicsDB importBatch raised exception", e);
+                }
+
+                closeReaders(sampleToReaderMap);
```
kgururaj: Can we collapse the code segments in the if and else portions by setting batchSize equal to number of samples if DEFAULT_ZERO_BATCH_SIZE?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -149,26 +170,46 @@ public boolean requiresReference() {
     @Override
     public void onStartup() {
 
-        for (final String variantPath : variantFilePaths) {
+        List<VCFHeader> headers = new ArrayList<>(variantFilePaths.size());
 
-            final String variantURI = IOUtils.getPath(variantPath).toAbsolutePath().toUri().toString();
+        for (String variantFilePath : variantFilePaths) {
```
droazen: Name these `variantPath` and `variantPaths`, since they're not necessarily normal files (could be GCS URIs)

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -149,26 +170,46 @@ public boolean requiresReference() {
...
             // A GVCF file must contain only one sample, throw an exception otherwise
-            if (((VCFHeader)reader.getHeader()).getNGenotypeSamples() != 1) {
-                throw new UserException("Input GVCF: " + variantPath + " should contain data for one sample");
+            if (header.getNGenotypeSamples() != 1) {
+                throw new UserException("Input GVCF: " + variantFilePath + " should contain data for one sample");
+            }
+            String sampleName = header.getGenotypeSamples().get(0);
+            headers.add(header);
+            sampleNames.add(sampleName);
+            sampleNameToFileName.put(sampleName, variantFilePath);
```
droazen: Check the return value of this `put()` call -- if the key was already present (indicated by a non-null return value), throw a `UserException` saying that there were duplicate samples.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -133,13 +147,20 @@ public boolean requiresReference() {
     // Intervals from command line (singleton for now)
     private List<ChromosomeInterval> intervals;
 
-    private final Map<String, FeatureReader<VariantContext>> sampleToVCFMap = new HashMap<>();
+    // List of all sample names
+    private List<String> sampleNames = new ArrayList<>();
+
+    // Linked hash map between sample names and corresponding GVCF file name
+    private Map<String, String> sampleNameToFileName = new LinkedHashMap<>();
```
droazen: Name this `sampleNameToVcfUri`, since the String values here are produced by `IOUtils.getPath(variantFilePath).toAbsolutePath().toUri().toString();`

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -149,26 +170,46 @@ public boolean requiresReference() {
...
+            } catch (IOException e) {
+                throw new GATKException("FeatureReader close() failed for " + sampleName, e);
             }
-
-            final String sampleName = ((VCFHeader) reader.getHeader()).getGenotypeSamples().get(0);
-            sampleToVCFMap.put(sampleName, reader);
         }
 
-        createMergedHeader();
+        mergedHeader = VCFUtils.smartMergeHeaders(headers, true);
```
droazen: Reading all the headers up-front like this, accumulating them in a collection, and then merging them might prove way too expensive over GCS. In a future PR (does not need to happen in this PR!), we might want to change this. One possibility is that we could pass in (via a new argument) a single input whose header to treat as canonical. I'll file a ticket to explore this.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -149,26 +170,46 @@ public boolean requiresReference() {
...
 
-        initializeIntervals();
+        //This sorts the list sampleNames if !useSamplesInOrder
+        //Why you should use this? If you are writing multiple partitions in different machines,
+        //you must have consistent ordering of samples across partitions. If file order is different
+        //in different processes, then set useSamplesInOrder to false and let the sort in
+        //generateSortedCallSetMap ensure consistent ordering across samples
```
droazen: I'd replace this long-ish comment with the more compact: "Passing in false here so that sample names will be sorted. This is needed for consistent ordering across partitions/machines."

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
      */
     @Override
     public void traverse() {
-        try {
-            importer.importBatch();
-        } catch (IOException e) {
-            throw new UserException("GenomicsDB importBatch raised exception", e);
+
+        if (batchSize == DEFAULT_ZERO_BATCH_SIZE) {
```
droazen: Extract the body of the `if` here to a `doBatchlessImport()` method, and call it from here.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+            try {
+                importer.importBatch();
+                assert importer.isDone();
+            } catch (IOException e) {
+                throw new UserException("GenomicsDB importBatch raised exception", e);
+            }
+
+            closeReaders(sampleToReaderMap);
+
+        } else {
```
droazen: Extract the body of the `else` here to a `doBatchedImport()` method, and call it from here.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+                importer = new GenomicsDBImporter(sampleToReaderMap, mergedHeader, intervals.get(0), importConfiguration);
+            } catch (IOException e) {
+                throw new UserException("Error initializing GenomicsDBImporter", e);
+            }
+
+            try {
+                importer.importBatch();
+                assert importer.isDone();
+            } catch (IOException e) {
+                throw new UserException("GenomicsDB importBatch raised exception", e);
```
droazen: "GenomicsDB importBatch raised exception" -> "GenomicsDB import failed"

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+            }
+
+            closeReaders(sampleToReaderMap);
+
+        } else {
+            for (int i = 0; i < sampleNames.size(); i += batchSize) {
+
+                int upperBoundSampleIndex = min(sampleNames.size(), i+batchSize);
```
droazen: If this is an exclusive upper bound (as I think it must be, since there is no sample name at the actual index `sampleNames.size()`), can you mention this in a comment here?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+                Map<String, FeatureReader<VariantContext>> sampleToReaderMap =
+                    getFeatureReaders(sampleNames, sampleNameToFileName, i, upperBoundSampleIndex);
+
+                logger.info("Importing batch " + i + " with " + batchSize + " samples");
```
droazen: Don't use `i` as the batch number in the log message -- keep a counter of which batch we're on (starting with 1), and use that in the log message.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+                long variantContextBufferSize = vcfBufferSizePerSample * batchSize;
```
droazen: Make this `final`

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+
+                GenomicsDBImportConfiguration.ImportConfiguration importConfiguration =
+                        createImportConfiguration(workspace, arrayName,
+                                variantContextBufferSize, segmentSize,
+                                "", "",
```
droazen: Why aren't we passing in the vidmap/callset JSONs here? When do they get written in the batched case?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+                                i, upperBoundSampleIndex);
+
+                try {
+                    importer = new GenomicsDBImporter(sampleToReaderMap, mergedHeader, intervals.get(0), importConfiguration);
+                } catch (IOException e) {
+                    throw new UserException("Error initializing GenomicsDBImporter for batch " + i +
```
droazen: Replace `i` with the true batch number in this message, as above.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+                            " processing ", e);
+                }
+
+
+                try {
+                    importer.importBatch();
+                    assert importer.isDone();
+                } catch (IOException e) {
+                    throw new UserException("GenomicsDB importBatch raised exception", e);
+                }
```
droazen: These two try/catch blocks (starting at line 296) are identical in the batched and non-batched code -- can you extract a common `doImport(sampleToReaderMap, mergedHeader, interval, importConfiguration)` method, and call it from both the batched and non-batched code?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
droazen: If this change is possible, I'd support it.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+
+                closeReaders(sampleToReaderMap);
+            }
```
droazen: Add a "Done importing batch X" logger message here.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
 
-    @Override
-    public void closeTool() {
-        for (Map.Entry<String, FeatureReader<VariantContext>> reader : sampleToVCFMap.entrySet()) {
+    /**
+     * Method to create feature readers for input files or GCS URLs
+     * in the current batch
+     *
+     * @param sampleNames  List of all sample names
+     * @param sampleNameToFileName  Sample name to file name mapping
```
droazen: Document `lowerSampleIndex` and `upperSampleIndex` as well -- in particular, document whether they are inclusive or exclusive indices.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+            assert sampleName.equals(((VCFHeader) reader.getHeader()).getGenotypeSamples().get(0));
+            sampleToReaderMap.put(sampleName, reader);
+        }
+        return sampleToReaderMap;
+    }
+
+    /**
+     * Private method to create a GenomicsDB configuration data structure
+     * instead of sending a long list of parameters to the constructor call
+     *
```
droazen: Add documentation for the method arguments. In particular, document what happens if you pass empty Strings in for the JSON files, and document whether the indices are inclusive or exclusive.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +249,169 @@ public void onTraversalStart() {
...
+                .build();
+
+        GenomicsDBImportConfiguration.GATK4Integration.Builder gBuilder =
+            GenomicsDBImportConfiguration.GATK4Integration.newBuilder();
+
+        if (!vidMapJSONFile.isEmpty())
+            gBuilder.setOutputVidmapJsonFile(vidMapJSONFile);
+
+        if (!callsetMapJSONFile.isEmpty())
+            gBuilder.setOutputCallsetmapJsonFile(callsetMapJSONFile);
```
droazen: What happens when you don't set these on the builder?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -338,6 +496,7 @@ private void initializeIntervals() {
      */
     @Override
     public SAMSequenceDictionary getBestAvailableSequenceDictionary() {
+        VCFHeader mergedVCFHeader = new VCFHeader(mergedHeader);
```
droazen: I think it makes more sense to store the merged VCF header as a field, instead of re-creating it on every call to this method.

src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
```diff
@@ -55,39 +55,109 @@ public void testGenomicsDBImportGCSInputs() throws IOException {
...
+  }
+
+  @Test(groups = {"bucket"})
+  public void testGenomicsDBImportGCSInputsInBatches() throws IOException {
+    final String hg00096_cloud = getGCPTestInputPath() + "large/gvcfs/HG00096.g.vcf.gz";
+    final String hg00268_cloud = getGCPTestInputPath() + "large/gvcfs/HG00268.g.vcf.gz";
+    final String na19625_cloud = getGCPTestInputPath() + "large/gvcfs/NA19625.g.vcf.gz";
+    final String combined = largeFileTestDir + "gvcfs/combined.gatk3.7.g.vcf.gz";
+    final SimpleInterval interval = new SimpleInterval("chr20", 17960187, 17981445);
+    final int batchSize = 1;
```
droazen: This variable appears to be unused.

src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
```diff
@@ -55,39 +55,109 @@ public void testGenomicsDBImportGCSInputs() throws IOException {
...
+    final String hg00268 = largeFileTestDir + "gvcfs/HG00268.g.vcf.gz";
+    final String na19625 = largeFileTestDir + "gvcfs/NA19625.g.vcf.gz";
+    final String combined = largeFileTestDir + "gvcfs/combined.gatk3.7.g.vcf.gz";
+    final SimpleInterval interval = new SimpleInterval("chr20", 17960187, 17981445);
+
+    testGenomicsDBImporterWithBatchSize(Arrays.asList(hg00096, hg00268, na19625), interval, combined);
+  }
+
+  @Test(groups = {"bucket"})
+  public void testGenomicsDBImportGCSInputsInBatches() throws IOException {
```
droazen: Recommend that you test with several different batch sizes here (1, 2, 3, and I'd also include 4 to prove that the tool works even if you pass in a batch size > the number of samples). You can easily do this by adding a `DataProvider` for this method that provides the batch sizes, and then pass the batch size into `testGenomicsDBImporterWithBatchSize()` on each invocation.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -116,6 +121,15 @@
...
     private Boolean overwriteExistingWorkspace = false;
 
+    @Argument(fullName = BATCHSIZE_ARG_NAME,
+              shortName = BATCHSIZE_ARG_NAME,
+              doc = "Batch size to group samples in during import. " +
+                    "This will minimize memory consumption, but might take longer to complete." +
+                    "Also, if batching creates too many fragments in GenomicsDB, run consolidation" +
+                    " to improve feature read time. Defaults to " + DEFAULT_ZERO_BATCH_SIZE,
+              optional = true)
+    private int batchSize = DEFAULT_ZERO_BATCH_SIZE;
```
droazen: If the batchSize is greater than the number of samples, will the tool work / not blow up?

src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
```diff
@@ -55,39 +55,109 @@ public void testGenomicsDBImportGCSInputs() throws IOException {
...
+      BaseTest.assertCondition(actualVcs, expectedVcs, (a, e) -> {
+          // TODO: Temporary hacks to make this test pass. Must be removed later
+          if (// allele order
+            e.getStart() != 17967343 && e.getStart() != 17966384 &&
+            // split block
+            e.getEnd() != 17981447
+            ) {
+            VariantContextTestUtils.assertVariantContextsAreEqual(a, e, Collections.emptyList());
+          }
+        });
```
droazen: This code to create readers for the actual and expected files is duplicated in the batch and non-batch test cases -- recommend that you extract a common `checkGenomicsDBAgainstExpected()` method and call it from both test methods.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -28,6 +29,8 @@
 import java.util.*;
 import java.util.function.Function;
 
+import static java.lang.Math.min;
```
kdatta: Blessings of Intellij IDE

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kdatta: The tool will not blow up. The batch size is used to calculate lower and upper bound of sample indexes. If batch size is larger than number of samples, last sample index will be used. I will add a test with this.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kdatta: yeah, makes sense

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kdatta: They are not written.

droazen: Aren't they needed? If they don't get written, the next tool in the pipeline (`GenotypeGVCFs`) will blow up.

kdatta: Yes, they are needed. The first time we call the importer they must be passed. I have added this logic in the next push.

kdatta:
@droazen, genomicsdb-0.6.0-proto-3.0.0-beta-1 released.

kdatta:
Also, the test breaks with this messge: "Can't run cloud tests without keys so don't run tests." I'm guessing Travis is not able to see the Google project keys!

kdatta:
The test is failing due to a bug in my code. Need to look into this a bit.

droazen:
Pasting the relevant portion of the travis log here for reference. The test suite appears to be dying with error code 134 (process received `SIGABRT`), perhaps due to the following error: `[E::bcf_hdr_add_sample] Duplicated sample name 'HG00096'`

```
:test[M::bwa_idx_load_from_disk] read 0 ALT contigs
[M::mem_pestat] skip orientation FF as there are not enough pairs
[M::mem_pestat] analyzing insert size distribution for orientation FR...
[M::mem_pestat] (25, 50, 75) percentile: (387, 415, 430)
[M::mem_pestat] low and high boundaries for computing mean and std.dev: (301, 516)
[M::mem_pestat] mean and std.dev: (407.54, 32.35)
[M::mem_pestat] low and high boundaries for proper pairs: (258, 559)
[M::mem_pestat] skip orientation RF as there are not enough pairs
[M::mem_pestat] skip orientation RR as there are not enough pairs
[M::mem_pestat] skip orientation FF as there are not enough pairs
[M::mem_pestat] analyzing insert size distribution for orientation FR...
[M::mem_pestat] (25, 50, 75) percentile: (387, 415, 430)
[M::mem_pestat] low and high boundaries for computing mean and std.dev: (301, 516)
[M::mem_pestat] mean and std.dev: (407.54, 32.35)
[M::mem_pestat] low and high boundaries for proper pairs: (258, 559)
[M::mem_pestat] skip orientation RF as there are not enough pairs
[M::mem_pestat] skip orientation RR as there are not enough pairs
[M::mem_pestat] skip orientation FF as there are not enough pairs
[M::mem_pestat] analyzing insert size distribution for orientation FR...
[M::mem_pestat] (25, 50, 75) percentile: (470, 502, 533)
[M::mem_pestat] low and high boundaries for computing mean and std.dev: (344, 659)
[M::mem_pestat] mean and std.dev: (501.23, 49.14)
[M::mem_pestat] low and high boundaries for proper pairs: (281, 722)
[M::mem_pestat] skip orientation RF as there are not enough pairs
[M::mem_pestat] skip orientation RR as there are not enough pairs
Created workspace /tmp/travis/genomicsdb-tests-1057177121940476778/workspace
GENOMICSDB_TIMER,Fetch from VCF,Wall-clock time(s),0.081966,Cpu time(s),0.0818668,Critical path wall-clock time(s),0.081742,Cpu time(s),0.0816418,#critical path,63
GENOMICSDB_TIMER,Combining cells,Wall-clock time(s),0.014772,Cpu time(s),0.0147254,Critical path wall-clock time(s),0.000297,Cpu time(s),0.000297,#critical path,4
GENOMICSDB_TIMER,Flush output,Wall-clock time(s),0,Cpu time(s),0,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Sections time,Wall-clock time(s),0.096828,Cpu time(s),0.0967047,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Time in single thread phase(),Wall-clock time(s),7.3e-05,Cpu time(s),7.8149e-05,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Time in read_all(),Wall-clock time(s),0.096994,Cpu time(s),0.0968966,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
WARNING: No valid combination operation found for INFO field DS - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field HaplotypeScore - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field InbreedingCoeff - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field MLEAC - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field MLEAF - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field DS - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field HaplotypeScore - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field InbreedingCoeff - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field MLEAC - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field MLEAF - the field will NOT be part of INFO fields in the generated VCF records
Created workspace /tmp/travis/genomicsdb-batchsize-tests-311791747726779418/workspace-1
GENOMICSDB_TIMER,Fetch from VCF,Wall-clock time(s),0.018375,Cpu time(s),0.0183694,Critical path wall-clock time(s),0.017755,Cpu time(s),0.0177497,#critical path,19
GENOMICSDB_TIMER,Combining cells,Wall-clock time(s),0.003093,Cpu time(s),0.00308688,Critical path wall-clock time(s),0.001558,Cpu time(s),0.00155431,#critical path,12
GENOMICSDB_TIMER,Flush output,Wall-clock time(s),0,Cpu time(s),0,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Sections time,Wall-clock time(s),0.021501,Cpu time(s),0.0215038,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Time in single thread phase(),Wall-clock time(s),4e-05,Cpu time(s),3.435e-05,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Time in read_all(),Wall-clock time(s),0.021587,Cpu time(s),0.0215973,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Fetch from VCF,Wall-clock time(s),0.013039,Cpu time(s),0.0130251,Critical path wall-clock time(s),0.012809,Cpu time(s),0.0127959,#critical path,15
GENOMICSDB_TIMER,Combining cells,Wall-clock time(s),0.002354,Cpu time(s),0.00230224,Critical path wall-clock time(s),0.000915,Cpu time(s),0.000913849,#critical path,8
GENOMICSDB_TIMER,Flush output,Wall-clock time(s),0,Cpu time(s),0,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Sections time,Wall-clock time(s),0.015422,Cpu time(s),0.0153636,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Time in single thread phase(),Wall-clock time(s),2.6e-05,Cpu time(s),2.6756e-05,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Time in read_all(),Wall-clock time(s),0.015482,Cpu time(s),0.0154346,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Fetch from VCF,Wall-clock time(s),0.052394,Cpu time(s),0.0522175,Critical path wall-clock time(s),0.050241,Cpu time(s),0.0500711,#critical path,55
GENOMICSDB_TIMER,Combining cells,Wall-clock time(s),0.008567,Cpu time(s),0.00855882,Critical path wall-clock time(s),0.004002,Cpu time(s),0.00399618,#critical path,29
GENOMICSDB_TIMER,Flush output,Wall-clock time(s),0,Cpu time(s),0,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Sections time,Wall-clock time(s),0.061053,Cpu time(s),0.0609081,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Time in single thread phase(),Wall-clock time(s),0.000112,Cpu time(s),0.00010141,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
GENOMICSDB_TIMER,Time in read_all(),Wall-clock time(s),0.061276,Cpu time(s),0.0611596,Critical path wall-clock time(s),0,Cpu time(s),0,#critical path,0
WARNING: No valid combination operation found for INFO field DS - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field HaplotypeScore - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field InbreedingCoeff - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field MLEAC - the field will NOT be part of INFO fields in the generated VCF records
WARNING: No valid combination operation found for INFO field MLEAF - the field will NOT be part of INFO fields in the generated VCF records
[E::bcf_hdr_add_sample] Duplicated sample name 'HG00096'

Results: SUCCESS (0 tests, 0 successes, 0 failures, 0 skipped)
:test FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':test'.
> Process 'Gradle Test Executor 1' finished with non-zero exit value 134
```

kdatta:
@droazen, I already found the bug. Fixing it now.

kdatta:
@droazen, both batch size and consolidate arguments are pushed. all tests pass.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +262,176 @@ public void onTraversalStart() {
...
     @Override
     public void traverse() {
-        try {
-            importer.importBatch();
-        } catch (IOException e) {
-            throw new UserException("GenomicsDB importBatch raised exception", e);
+
+        int updatedBatchSize = (batchSize == DEFAULT_ZERO_BATCH_SIZE) ? sampleNames.size() : batchSize;
+        int sampleCount = sampleNames.size();
+        int totalBatchCount = (sampleCount/updatedBatchSize) + (sampleCount%updatedBatchSize);
```
droazen: This value seems wrong -- if, eg., `sampleCount` were 55 and `updatedBatchSize` were 10, then this would give `(55 / 10) + (55 % 10) = 5 + 5 = 10`, but the correct number of batches is 6 in this case (55 / 10 + 1)

kdatta: you're right. the logic should be:
totalBatchCount = (sampleCount/updatedBatchSize) + (sampleCount%updatedBatchSize==0 ? 0 : 1);

kdatta:
@droazen, I'll push a new commit once you complete reviewing

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -116,6 +122,27 @@
...
+    @Argument(fullName = CONSOLIDATE_ARG_NAME,
+              shortName = CONSOLIDATE_ARG_NAME,
+              doc = "Boolean flag to enable consolidation. If importing data in batches, a new fragment is created for " +
+                    "each batch. In case thousands of fragments are created, GenomicsDB feature readers will try " +
+                    "to open ~20x as many files. Also, internally GenomicsDB would consume more memory to maintain " +
+                    "bookkeeping data from all fragments. Use this flag to merge all fragments into one. " +
+                    "Merging can potentially improve read performance, however overall benefit might not be noticeable " +
+                    "as the top Java layers has significantly higher overheads. This flag has no effect if only one " +
+                    "batch is used. Defaults to false")
+    private Boolean doConsolidation = false;
```
droazen: This argument should be marked as optional

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +263,176 @@ public void onTraversalStart() {
...
+
+            try {
+                importer = new GenomicsDBImporter(sampleToReaderMap, mergedHeader, intervals.get(0), importConfiguration);
+            } catch (IOException e) {
+                throw new UserException("Error initializing GenomicsDBImporter in batch " + batchCount, e);
+            }
+
+            try {
+                importer.importBatch();
+                assert importer.isDone();
```
droazen: Throw a `GATKException` if the importer is not done, rather than making an `assert`, since asserts are not typically turned on at runtime.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +263,176 @@ public void onTraversalStart() {
...
+            Map<String, FeatureReader<VariantContext>> sampleToReaderMap =
+                getFeatureReaders(sampleNames, sampleNameToVcfUri, updatedBatchSize, sampleCount, i);
+
+            logger.info("Importing batch " + batchCount + " with " + sampleToReaderMap.size() + " samples");
+            final long variantContextBufferSize = vcfBufferSizePerSample * sampleToReaderMap.size();
+
+            GenomicsDBImportConfiguration.ImportConfiguration importConfiguration =
+                    createImportConfiguration(workspace, arrayName,
+                            variantContextBufferSize, segmentSize,
+                            "", "",
```
droazen: Do we still need these parameters to `createImportConfiguration()` that we're passing in empty Strings for? This is the only call to this method, so I think we could remove these two parameters.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +263,176 @@ public void onTraversalStart() {
...
-        return importer.isDone();
+        if (batchSize==DEFAULT_ZERO_BATCH_SIZE) {
+            logger.info("Import completed!");
+        } else {
+            logger.info("Import of all batches to GenomicsDB completed!");
+        }
+
+        // Write the vid and callset map JSON files
+        GenomicsDBImporter.writeVidMapJSONFile(vidMapJSONFile.getAbsolutePath(), mergedHeader);
+        GenomicsDBImporter.writeCallsetMapJSONFile(callsetMapJSONFile.getAbsolutePath(), callsetMappingPB);
```
droazen: Can these writes fail? If they do, can you ensure that either a `UserException` or `GATKException` gets thrown, depending on whether or not it's likely to be the user's fault?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +263,176 @@ public void onTraversalStart() {
...
+
+        if (doConsolidation) {
+            logger.info("GenomicsDB consolidation started");
+            GenomicsDBImporter.consolidateTileDBArray(workspace, arrayName);
```
droazen: Same question here -- can the consolidation fail? If yes, ensure that either a UserException or GATKException gets thrown.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +263,176 @@ public void onTraversalStart() {
...
-        for (Map.Entry<String, FeatureReader<VariantContext>> reader : sampleToVCFMap.entrySet()) {
+    /**
+     * Method to create feature readers for input files or GCS URLs
+     * in the current batch
+     *
+     * @param sampleNames  List of all sample names
+     * @param sampleNameToFileName  Sample name to file name mapping
+     * @param batchSize  Current batch size
+     * @param sampleCount  total samples in this import
+     * @param lowerSampleIndex  Lower bound of sample index -- inclusive
```
droazen: Document whether this is 0-based or 1-based.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -221,24 +263,176 @@ public void onTraversalStart() {
...
+     * instead of sending a long list of parameters to the constructor call
+     *
+     * @param workspace  GenomicsDB workspace
+     * @param arrayName  GenomicsDB array
+     * @param variantContextBufferSize  Buffer size to store VCF records for all samples
+     * @param segmentSize  Buffer size to store columnar data to be serialized to disk
+     * @param vidMapJSONFile  JSON File containing fields and contig information. File is not written if empty
+     * @param callsetMapJSONFile  JSON File containing callset information. File is not written if empty
+     * @param lbSampleIndex  Lower bound of sample index -- inclusive
+     * @param ubSampleIndex  Upper bound of sample index -- inclusive
```
droazen: Document whether these indices are 0-based or 1-based

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -116,6 +122,28 @@
               optional = true)
     private Boolean overwriteExistingWorkspace = false;
 
+    @Argument(fullName = BATCHSIZE_ARG_NAME,
+              shortName = BATCHSIZE_ARG_NAME,
+              doc = "Batch size controls the number of samples for which readers are open at once " +
+                    "and therefore provides a way to minimize memory consumption. However, it can take longer to complete. " +
+                    "Use consolidate flag if more than hundred batches were used. This will improve feature read time. " +
```
droazen: "Use consolidate flag" -> "Use the consolidate flag"

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -116,6 +122,28 @@
...
+                    "batchSize=0 means no batching. Readers for all samples will be opened at once. " +
```
droazen: "batchSize=0 means no batching. Readers for all samples will be opened at once. " -> "batchSize=0 means no batching (ie., readers for all samples will be opened at once)"

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -116,6 +122,28 @@
...
+    private int batchSize = DEFAULT_ZERO_BATCH_SIZE;
+
+    @Argument(fullName = CONSOLIDATE_ARG_NAME,
+              shortName = CONSOLIDATE_ARG_NAME,
+              doc = "Boolean flag to enable consolidation. If importing data in batches, a new fragment is created for " +
+                    "each batch. In case thousands of fragments are created, GenomicsDB feature readers will try " +
+                    "to open ~20x as many files. Also, internally GenomicsDB would consume more memory to maintain " +
+                    "bookkeeping data from all fragments. Use this flag to merge all fragments into one. " +
+                    "Merging can potentially improve read performance, however overall benefit might not be noticeable " +
+                    "as the top Java layers has significantly higher overheads. This flag has no effect if only one " +
```
droazen: has -> have

src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
```diff
@@ -26,68 +29,150 @@
...
+        writeToGenomicsDB(vcfInputs, interval, workspace, 0, false, 0);
+        checkJSONFilesAreWritten(workspace);
+        checkGenomicsDBAgainstExpected(workspace, interval, expectedCombinedVCF);
+    }
+
+    private void testGenomicsDBImporterWithBatchSize(final List<String> vcfInputs, final SimpleInterval interval, final String expectedCombinedVCF, int batchSize) throws IOException {
+        final String workspace = createTempDir("genomicsdb-batchsize-tests-").getAbsolutePath() + "/workspace-" + batchSize;
+
+        writeToGenomicsDB(vcfInputs, interval, workspace, batchSize, false, 0);
+        checkGenomicsDBAgainstExpected(workspace, interval, expectedCombinedVCF);
```
droazen: Why is there no `checkJSONFilesAreWritten()` call here? Don't we want to check that the JSONs are written when using batching?

src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
```diff
@@ -26,68 +29,150 @@
...
+
+        if (useBufferSize)
+            args.addArgument("genomicsDBVCFBufferSize", String.valueOf(bufferSizePerSample));
+
+        runCommandLine(args);
+    }
+
+    private void checkJSONFilesAreWritten(String workspace) {
+        Assert.assertTrue(new File(workspace, GenomicsDBConstants.DEFAULT_VIDMAP_FILE_NAME).exists());
+        Assert.assertTrue(new File(workspace, GenomicsDBConstants.DEFAULT_CALLSETMAP_FILE_NAME).exists());
```
droazen: Can you check that the JSONs are valid here as well?

src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
```diff
@@ -26,68 +29,150 @@
...
+        final String hg00268_cloud = getGCPTestInputPath() + "large/gvcfs/HG00268.g.vcf.gz";
+        final String na19625_cloud = getGCPTestInputPath() + "large/gvcfs/NA19625.g.vcf.gz";
+        final String combined = largeFileTestDir + "gvcfs/combined.gatk3.7.g.vcf.gz";
+        final SimpleInterval interval = new SimpleInterval("chr20", 17960187, 17981445);
+
+        testGenomicsDBImporterWithBatchSize(Arrays.asList(hg00096_cloud, hg00268_cloud, na19625_cloud), interval, combined, batchSize);
+    }
+
+    @Test(expectedExceptions = UserException.class)
+    public void testZeroVCFBufferSize() throws IOException {
```
droazen: Document in a one-line comment why this test case exists.

droazen:
Second-pass review complete -- back to @kdatta for a few final changes.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kdatta: removed this assertion. this is redundant cause all exceptions are thrown inside importBatch()

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kdatta: nope, removed

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kdatta: Yeah, these writes throw FileNotFoundException which is now propagated to this tool and thrown as an UserException

kdatta: the only reason it can fail if a correct path is not provided. seems like user's fault

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kdatta: If consolidation fails, the GenomicsDBImporter will throw an exception and die which is the expected behavior.

src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
kdatta: I found these checks are redundant! GenomicsDB feature reader has lots of checks and if the callset or vidmap files are not generated correctly it fails. Hence, removed

src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
kdatta: removed these checks

src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
droazen: It doesn't hurt to have these extra checks in our tests -- they would protect against the scenario where the checks in the feature reader were accidentally removed. Given that we already had a bug involving the JSON files not getting written in batch mode, I'd prefer to keep these.

src/test/java/org/broadinstitute/hellbender/tools/GenomicsDBImportIntegrationTest.java
kdatta: Inside GenomicsDB we validate the json files. I think checking the content and validity again will just add redundant code here. But, I will add the check back to see whether these files are written. Okay?

droazen: Yes, at least check explicitly that they are written.

kdatta:
@droazen, changed submitted

droazen:
:+1: merging!

PR #2637, opened 2017-04-28
add SA tags to sam output

tedsharpe:


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignedAssemblyOrExcuse.java
```diff
@@ -130,12 +131,17 @@ public static void writeSAMFile( final String samFile, final PipelineOptions pOp
...
                     final List<BwaMemAlignment> alignments = allAlignments.get(contigIdx);
                     if ( alignments.isEmpty() ) continue;
+                    final Map<BwaMemAlignment,String> saTagMap = BwaMemAlignmentUtils.createSATags(alignments,refNames);
                     final FermiLiteAssembly.Contig contig = assembly.getContig(contigIdx);
                     final String readName = String.format("asm%06d:tig%05d", assemblyId, contigIdx);
                     final byte[] calls = contig.getSequence();
                     for ( final BwaMemAlignment alignment : alignments ) {
-                        writer.addAlignment(BwaMemAlignmentUtils.applyAlignment(readName, calls, null, null, alignment,
-                                refNames, header, false, false));
+                        SAMRecord samRecord =
```
SHuang-Broad: `final`

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
```diff
@@ -91,11 +91,43 @@ else if ( alwaysGenerateASandXS ) {
...
+            }
+            saTags.put(alignment, saTag.toString());
+        }
+        return saTags;
+    }
+
+    /**
      * Describes an alignment as a string for use in an SA tag, for example.
      */
     public static String asTag( final BwaMemAlignment alignment, final List<String> refNames ) {
```
SHuang-Broad: This method seems to deal with SA tag only and is used only by `createSATags `. I would suggest making it private, and rename to something like `asSATag` (doesn't have to be, just for the name to reflect the intention a bit better, or the XA tag could be formatted like this as well?)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/AlignedAssemblyOrExcuse.java
tedsharpe: Hmmpf.  I wonder why IntelliJ didn't warn me about this one.  Fixed.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
tedsharpe: Yeah.  I don't see a need to make this private.  It's generally useful, even if it's only used to make SA tags right now.  The encoding used for SA and XA tags allows you to reference other alignments within an SAM record, and might see other uses.  So asTag pretty well describes turning an alignment into this kind of string, I think.
I think I'll duck your suggestion, if that's OK.

tedsharpe:
Rebased and squashed.  Is everyone happy?

PR #2643, opened 2017-05-01
Detect SeekableByteChannelPrefetcher double-wrapping

jean-philippe-martin:
Wrapping a prefetcher into another would be an error, check
for it and report it. The commit also includes a test.

Also shut down the executor when the channel is closed.


jean-philippe-martin:
I don't think that's the elusive bug we're looking for, but a bit more argument validation certainly cannot hurt.

The gradle files look like they're auto-generated, I certainly didn't change them myself. Let me know if you think we should delete them instead.

jean-philippe-martin:
I've ran this one in my test and it's clear the current code doesn't have the problem of prefetching a prefetcher.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Thu Jan 12 19:57:42 EST 2017
+#Mon May 01 15:13:13 PDT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.1-all.zip
```
lbergelson: This file gets updated like this accidentally a lot.  When it's generated it's created with `bin`, but Intellij always suggest changing it to `all` so people do that and then accidentally commit it.  It doesn't really matter either way, if it's set to bin then it only downloads the gradle binaries, with all it also downloads the source jars so it can be inspected locally more easily.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -441,6 +447,13 @@ public boolean isOpen() {
     @Override
     public void close() throws IOException {
         if (open) {
+            exec.shutdown();
+            try {
+                exec.awaitTermination(3, TimeUnit.MINUTES);
```
lbergelson: A 3 minute wait is potentially a long confusing stall.  Is there a reason we want to wait or should we just kill the thread immediately?

jean-philippe-martin: A zero timeout wouldn't kill the thread, it would just stop waiting for it to acknowledge the interruption. Then the thread would be reading on a closed channel and would fail. This isn't a huge deal it would just confuse matters.

How about a much shorter timeout? I expect that five seconds is still enough for most cases.

lbergelson: Oh, I see, I figured it would kill the thread if the timer ran out.  Should we just aggressively kill the thread here?  I'm assuming if we're closing we want to stop the prefetching immediately.  I think we don't want to close the channel and end up with surprising uncaught IOExceptions.

lbergelson: Maybe `shutdownNow` instead of await termination?  Is there any downside to murdering the prefetch thread if we're closing this stream?

jean-philippe-martin: Looking at the docs you're right - shutdown doesn't interrupt the thread so we need shutdownNow! But there's no way in Java to "murder" a thread, interrupting is the closest we can get. That means it's still worthwhile to wait for the worker to process its interrupted state (using awaitTermination).

lbergelson: David suggests `awaitTermination` first, and then if it timed out, calling `shutdownNow`.  Lets go with maybe 10 seconds?  Seems like a reasonable amount of time that no one will be sad waiting for it but most transfers should finish cleanly.

jean-philippe-martin: It's fine to call `shutdownNow` right away - no one's going to read the data being prefetched so there's no point in waiting 10s before interrupting the prefetching. Instead, let's interrupt at once and give the thread 10s to notice and stop cleanly.

I tried testing this, but the reads were too fast so I added a sleep in there. The output shows that the prefetching read gets interrupted as we expect (so we don't wait the full 10s), and the caller of close() gets no exception. 

> WorkUnit.call() starting
> WorkUnit.call() starting for real now
> WorkUnit.call() done
> isInterrupted: false
> WorkUnit.call() starting
> shutdownNow
> sleep was interrupted
> WorkUnit.call() starting for real now
> WorkUnit.call interrupted by: java.nio.channels.ClosedByInterruptException
> close
> Call to close() took: 0ms.

That said, the same test also shows that just closing the channel's fine too in the sense that the caller of close() sees no exception, so I guess this wasn't really necessary. I'll leave in the awaitTermination but cut the timeout to a single second since there's no penalty for timing out.

lbergelson:
@jean-philippe-martin Yup, reading hundreds of different files in the same tool.  We're trying to optimize the buffer size for that case, that's what we're looking at in that other thread you were commenting on.  It seems like we may have some memory leak somewhere else, you seem to need much more memory than we would expect if everything is working as expected.  

This branch should fix the thread leak we were seeing.

👍 Merge when tests pass.

jean-philippe-martin:
I added an integration test that tries the same thing with a bucket (it works fine). Merging as soon as green (do it for me if I miss it).

lbergelson:
thanks @jean-philippe-martin !

jean-philippe-martin:
You're welcome!

PR #2646, opened 2017-05-02
Implement remaining PathSeq tools

mwalker174:

The pipeline tools are:
1. PathSeqFilterSpark : quality/low-complexity/host read filtering
2. PathSeqPathogenAlignSpark : bwa-mem aligner
3. PathSeqClassifyReadsSpark : quantifies pathogen abundance

These are supported by utilities:
4. PathSeqKmerSpark - creates kmer library (either a Hopscotch set or Bloom filter) used by Filter tool
5. PathSeqBuildReferenceTaxonomy - creates a file containing taxonomic information for a given reference, required by ClassifyReads

tools.spark.pathseq package:
Contains all the tools. Static helper functions were put into PS<ToolName>Utils classes, eg PSFilterUtils contains functions used by PathSeqFilterSpark. PSUtils contains mostly functions that are used by more than one tool. There are a number of other "PS" classes for doing Bwa, taxonomy bookkeeping, and read classification. Also has the kmer and host alignment read filters.

tools.spark.sv package:
Added base masking to SVKmerShort class.

tools.spark.utils package:
Hopscotch set and Bloom filter for long primitives. Each type has a "Large" class for sets exceeding the maximum JVM array size (~2B).


lbergelson:
@mwalker174 This pull request is absolutely massive.  If it can be meaningfully decomposed into a few smaller pull requests to make reviews easier that might make more sense.

src/main/java/org/broadinstitute/hellbender/engine/filters/AmbiguousBaseReadFilter.java
```diff
@@ -14,23 +14,27 @@
 
     private static final long serialVersionUID = 1L;
 
-    @Argument(doc="Threshold fraction of non-regular bases (e.g. N) above which to filter",
-            fullName="ambigFilterFrac",
-            shortName="ambigFilterFrac", optional=true)
-    public float N_FRAC = 0.05f;
+    @Argument(fullName = "ambigFilterFrac", shortName = "ambigFilterFrac", optional = true)
+    public double N_FRAC = 0.05;
```
lbergelson: Arguments are not constants and shouldn't be in UPPERCASE.  There are places that have them like that that haven't been refactored, but we shouldn't be adding more.

lbergelson: Also, what happened to the dcoumentation string here?

src/main/java/org/broadinstitute/hellbender/engine/filters/MarkDuplicatesReadFilter.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.io.Serializable;
+
+/**
+ * Filters out reads marked as duplicates.
+ */
+public final class MarkDuplicatesReadFilter extends ReadFilter implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+    private boolean filterOpticalOnly = false;
```
lbergelson: shouldn't this be an argument?

src/main/java/org/broadinstitute/hellbender/engine/filters/MarkDuplicatesReadFilter.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.broadinstitute.hellbender.tools.spark.transforms.markduplicates.MarkDuplicatesSparkUtils;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+import java.io.Serializable;
+
+/**
+ * Filters out reads marked as duplicates.
+ */
+public final class MarkDuplicatesReadFilter extends ReadFilter implements Serializable {
```
lbergelson: I'm unclear about the need for this class vs two separate filters, `OpticalDuplicateReadFilter` and `DuplicateReadFilter`.  Seems confusing.

mwalker174:
@lbergelson @cwhelan @tedsharpe Going to break this PR down into smaller pieces. I will implement the feedback received thus far as well.

PR #2648, opened 2017-05-03
Cleaning code in ReadUtils

davidbenjamin:
@SHuang-Broad Would you mind reviewing this?  I admit it's not very entertaining.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
@@ -18,6 +18,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * A miscellaneous collection of utilities for working with reads, headers, etc.
@@ -25,8 +26,7 @@ import java.util.*;
  */
 public final class ReadUtils {
 
-    private ReadUtils() {
-    }
+    private ReadUtils() { }
 
     private static final Logger logger = LogManager.getLogger();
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/utils/read/ReadUtils.java
```diff
@@ -935,17 +872,12 @@ public static GATKRead resetOriginalBaseQualities(final GATKRead read) {
...
         final int referenceIndex = getReferenceIndex(read, header);
-        // Read is aligned to nonexistent contig
-        if( ! read.isUnmapped() && referenceIndex == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX ) {
-            return false;
-        }
-        final SAMSequenceRecord contigHeader = header.getSequence(referenceIndex);
-        // Read is aligned to a point after the end of the contig
-        return read.isUnmapped() || read.getStart() <= contigHeader.getSequenceLength();
+        return read.isUnmapped() || (referenceIndex != SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX
+                && read.getStart() <= header.getSequence(referenceIndex).getSequenceLength());
```
droazen: Hey @davidbenjamin, this kind of refactoring involving flipping boolean logic everywhere is actually really risky and error-prone in my experience, particular if there aren't comprehensive unit tests for all of these methods. I have a feeling if I spent an hour looking at this branch, I might be able to find a new bug or two. Given that we're trying to stabilize for a release at the moment, and the benefits of doing this refactoring are pretty slight, I recommend we hold off on these kinds of changes for now.

davidbenjamin: @droazen okay, I'll hold off!

SHuang-Broad: @davidbenjamin I'd be happy to review when things are stable.

PR #2651, opened 2017-05-03
unify two methods for checking readable files and let multiple files be passed. Closes #2145.

### LINKED ISSUE #2145: Allow Utils.regularReadableUserFile to take multiple files.

davidbenjamin:
@samuelklee This fixes another of your issues.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
@@ -704,27 +704,6 @@ public final class Utils {
 
 
     /**
-     * Checks that one or more user provided files are in fact regular (i.e. not a directory or a special device) readable files.
-     *
-     * @param files the input files to test.
-     * @throws IllegalArgumentException if any input file {@code file} is {@code null} or {@code files} is {@code null}.
-     * @throws UserException if any {@code file} is not a regular file or it cannot be read.
-     */
-    public static void canRead(final File... files) {
-        nonNull(files, "Unexpected null input.");
-        for (final File file : files) {
-            nonNull(file, "Unexpected null file reference.");
-            if (!file.exists()) {
-                throw new UserException.CouldNotReadInputFile(file, "The input file does not exist.");
-            } else if (!file.isFile()) {
-                throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(), "The input file is not a regular file");
-            } else if (!file.canRead()) {
-                throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(), "The input file cannot be read");
-            }
-        }
-    }
-
-    /**
      * Calculates the optimum initial size for a hash table given the maximum number
      * of elements it will need to hold. The optimum size is the smallest size that
      * is guaranteed not to result in any rehash/table-resize operations.
--- a/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
@@ -711,14 +711,20 @@ public final class Utils {
      * @throws UserException if {@code file} is not a regular file or it cannot be read.
      * @return the same as the input {@code file}.
      */
-    public static File regularReadableUserFile(final File file) {
+    public static void canRead(final File file) {
         nonNull(file, "unexpected null file reference");
-        if (!file.canRead()) {
-            throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(),"the input file does not exist or cannot be read");
+        if ( ! file.exists() ) {
+            throw new UserException.CouldNotReadInputFile(file, "It doesn't exist.");
         } else if (!file.isFile()) {
             throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(),"the input file is not a regular file");
-        } else {
-            return file;
+        } else if (!file.canRead()) {
+            throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(),"the input file does not exist or cannot be read");
+        }
+    }
+
+    public static void canRead(final File... files) {
+        for (final File file : files) {
+            canRead(file);
         }
     }
 
--- a/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
@@ -704,27 +704,23 @@ public final class Utils {
 
 
     /**
-     * Checks that a user provided file is in fact a regular (i.e. not a directory or a special device) readable file.
+     * Checks that one or more user provided files are in fact regular (i.e. not a directory or a special device) readable files.
      *
-     * @param file the input file to test.
-     * @throws IllegalArgumentException if {@code file} is {@code null} or {@code argName} is {@code null}.
-     * @throws UserException if {@code file} is not a regular file or it cannot be read.
-     * @return the same as the input {@code file}.
+     * @param files the input files to test.
+     * @throws IllegalArgumentException if any input file {@code file} is {@code null} or {@code files} is {@code null}.
+     * @throws UserException if any {@code file} is not a regular file or it cannot be read.
      */
-    public static void canRead(final File file) {
-        nonNull(file, "unexpected null file reference");
-        if ( ! file.exists() ) {
-            throw new UserException.CouldNotReadInputFile(file, "It doesn't exist.");
-        } else if (!file.isFile()) {
-            throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(),"the input file is not a regular file");
-        } else if (!file.canRead()) {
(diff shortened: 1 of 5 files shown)


src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -428,7 +433,7 @@ public void testUnsuccessfulRegularReadableFileCheck(final File file) {
         if (file == null){
             throw new SkipException("cannot make a file unreadable (maybe you're running as root)");
         }
-        Utils.regularReadableUserFile(file);
+        Utils.canRead(file);
```
samuelklee: Change the name of this test method and the associated data provider.

samuelklee: Note also the use of "RRFCD" in the data provider, which you should change as well (lest it become a mysterious, vestigial acronym).

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -405,10 +404,16 @@ public void testCheckForDuplicatesWithoutDuplicates() {
...
+    public void testSuccessfulCanReadFileCheck() {
         final File expectedFile = createTempFile("Utils-RRFC-test",".txt");
-        final File actualFile = Utils.regularReadableUserFile(expectedFile);
-        Assert.assertSame(actualFile, expectedFile);
+        Utils.canRead(expectedFile);
+    }
+
+    @Test
+    public void testSuccessfulCanReadFilesCheck() {
+        final File file1 = createTempFile("Utils-RRFC-test1",".txt");
```
samuelklee: I would change "RRFC" here and in the test above (see comment below).

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -711,14 +711,20 @@ public static void validate(final boolean condition, final Supplier<String> msg)
      * @throws UserException if {@code file} is not a regular file or it cannot be read.
      * @return the same as the input {@code file}.
      */
-    public static File regularReadableUserFile(final File file) {
+    public static void canRead(final File file) {
         nonNull(file, "unexpected null file reference");
-        if (!file.canRead()) {
-            throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(),"the input file does not exist or cannot be read");
+        if ( ! file.exists() ) {
```
samuelklee: White space.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -711,14 +711,20 @@ public static void validate(final boolean condition, final Supplier<String> msg)
...
+            throw new UserException.CouldNotReadInputFile(file, "It doesn't exist.");
```
samuelklee: Can you make the exception message analogous to those below (or vice versa)?

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -711,14 +711,20 @@ public static void validate(final boolean condition, final Supplier<String> msg)
      * @throws UserException if {@code file} is not a regular file or it cannot be read.
      * @return the same as the input {@code file}.
      */
-    public static File regularReadableUserFile(final File file) {
+    public static void canRead(final File file) {
```
samuelklee: Is there a need for both a single-argument method along with one with varargs?  I'd just put the loop in here.  Be sure to fix up the javadoc accordingly.  Also, might want to add an additional null check in case only a single null argument is passed, as well as define behavior when zero arguments are passed.

samuelklee:
Also, I think this is something that @asmirnov239 originally requested in a PR review long ago!

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
davidbenjamin: done

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
davidbenjamin: done + done

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
davidbenjamin: done

davidbenjamin: when zero arguments are passed nothing happens, which seems fine to me

davidbenjamin:
@samuelklee that's everything.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -704,21 +704,23 @@ public static void validate(final boolean condition, final Supplier<String> msg)
...
      */
-    public static File regularReadableUserFile(final File file) {
-        nonNull(file, "unexpected null file reference");
-        if (!file.canRead()) {
-            throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(),"the input file does not exist or cannot be read");
-        } else if (!file.isFile()) {
-            throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(),"the input file is not a regular file");
-        } else {
-            return file;
+    public static void canRead(final File... files) {
```
droazen: Erm, this method really **should** live in `IOUtils`, in my opinion.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -704,21 +704,23 @@ public static void validate(final boolean condition, final Supplier<String> msg)
...
+        nonNull(files, "Unexpected null input.");
+        for (final File file : files) {
+            nonNull(file, "Unexpected null file reference.");
+            if (!file.exists()) {
+                throw new UserException.CouldNotReadInputFile(file, "The input file does not exist.");
+            } else if (!file.isFile()) {
+                throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(), "The input file is not a regular file");
+            } else if (!file.canRead()) {
+                throw new UserException.CouldNotReadInputFile(file.getAbsolutePath(), "The input file cannot be read");
```
droazen: This error message should mention file permissions.

samuelklee:
Thanks for the changes!  :+1: to @droazen's comments as well.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
davidbenjamin: done

magicDGS:
Maybe late because it is already merged, but I think that the tests are misplaced now... shouldn't they live in `IOUtilsUnitTest`, @davidbenjamin?

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -589,4 +570,25 @@ public static void assertFileIsReadable(final Path path) {
...
     }
+
+    /**
+     * Checks that one or more user provided files are in fact regular (i.e. not a directory or a special device) readable files.
+     *
+     * @param files the input files to test.
+     * @throws IllegalArgumentException if any input file {@code file} is {@code null} or {@code files} is {@code null}.
+     * @throws UserException if any {@code file} is not a regular file or it cannot be read.
+     */
+    public static void canRead(final File... files) {
```
droazen: Looks like you already hit merge on this branch, but I neglected to mention that the new name for the method is also not ideal. "IOUtils.canRead(directory)" will fail on a readable directory, since a directory is not a regular file, but the method name implies that it should pass.

Also, as @magicDGS points out, the tests are now misplaced. I recommend that you do a second PR to fix these issues @davidbenjamin

davidbenjamin: @droazen Sorry about that. I thought you were just chiming in rather than volunteering as an official reviewer.  Should I restore the old method name that Louis objected to?

droazen: I think `IOUtils.fileIsReadable()` is fine and not too verbose (slightly less verbose than the name Louis objected to). `canRead()` is just confusing, since the method tests file vs. non-file in addition to readability.

Issue #2652, opened 2017-05-03
give prefetching threads useful names


lbergelson:
Currently the prefetcher threads all show up as something like "thread-pool-1-thread-1".  It would be nice if we could give them a name so it's immediately clear in the profiler what they are.  It looks like this can be done with a custom thread factory.  Guava has one that should make this easy, 
`new ThreadFactoryBuilder().setNameFormat("nio-prefetcher-thread-%d").build()`

lbergelson:
@jean-philippe-martin A really trivial one here if you get the chance.

lbergelson:
closed by #2654

PR #2654, opened 2017-05-03
Name prefetching thread

jean-philippe-martin:
Addresses issue #2652 with the added bonus that we distinguish between different prefetcher instances (they each only have one thread, so the thread index won't help all that much besides checking they don't have too many threads).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.hellbender.utils.nio;
 
 import com.google.common.base.Stopwatch;
+import java.util.concurrent.ThreadFactory;
 import org.broadinstitute.hellbender.exceptions.GATKException;
 
 import java.io.Closeable;
@@ -18,6 +19,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import shaded.cloud_nio.com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 /**
  * SeekableByteChannelPrefetcher wraps an existing SeekableByteChannel to add prefetching.
@@ -43,6 +45,8 @@ public final class SeekableByteChannelPrefetcher implements SeekableByteChannel 
     private long position = 0;
     private boolean open;
     private Stopwatch betweenCallsToRead = Stopwatch.createUnstarted();
+    private static int prefetcherCount = 0;
+    private final int prefetcherIndex;
 
     // statistics, for profiling
     // time spent blocking the user because we're waiting on the network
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -145,6 +149,7 @@ public void close() throws IOException {
     }
 
     public SeekableByteChannelPrefetcher(SeekableByteChannel chan, int bufSize) throws IOException {
+        this.prefetcherIndex = (prefetcherCount++);
```
lbergelson: Should we make prefetcher count volatile or atomic so that it's thread safe?

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -159,7 +164,11 @@ public SeekableByteChannelPrefetcher(SeekableByteChannel chan, int bufSize) thro
             this.bufSize = bufSize;
         }
         this.open = true;
-        exec = Executors.newFixedThreadPool(1);
+        // make sure the prefetching thread's name indicate what it is and
+        // which prefetcher it belongs to (for debugging purposes only, naturally).
+        String nameFormat = "nio-prefetcher-" + prefetcherIndex + "-thread-%d";
```
lbergelson: pull this string out as a constant

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -159,7 +164,11 @@ public SeekableByteChannelPrefetcher(SeekableByteChannel chan, int bufSize) thro
...
+        ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat(nameFormat).build();
```
lbergelson: This is a separate issue, but I think we should make the prefetcher threads into daemon threads.  
We're seeing an issue where the main thread dies with an exception but the prefetcher threads stay active and prevent the program from closing.  Making them daemons should cause them to die appropriately if the main thread dies I believe.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
jean-philippe-martin: Volatile wouldn't work (it just makes the individual operations atomic, not the read-assign-write as a whole as is needed). We could use atomics, or a lock, but that seems overkill for a debug variable. I don't think it's the end of the world if multiple threads end up with the same name, and I think that's preferable to adding complexity. I can be swayed, though.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
jean-philippe-martin: I'm not sure what you're asking or trying to accomplish. You mean to store it beyond the scope of the constructor?

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
jean-philippe-martin: Excellent idea, let's.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
lbergelson: ack, I wasn't paying enough attention, ignore that comment.

src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
lbergelson: As long as nothing will be broken by potentially having two threads with identical names than I'm fine letting through the rare case of having a race condition in our thread naming...

jean-philippe-martin: Yup, it's [no problem](http://docs.oracle.com/javase/6/docs/api/java/lang/Thread.html):

> More than one thread can have the same name

jean-philippe-martin:
Resolving conflict (and squashing while I'm at it)

PR #2655, opened 2017-05-04
unified check for current and deprecated spanning deletion allele

### LINKED ISSUE #2589: treat deprecated spanning deletion allele uniformly in all places
Currently some checks for spanning deletions check against `Alleles.SPAN_DEL`, others also check for `GATKVCFConstants.SPANNING_DELETION_SYMBOLIC_ALLELE_DEPRECATED`.  

If we want to be able to accept older style vcfs with spanning deletions we should check for both in all cases.  If we don't we should remove checks for the deprecated allele everywhere.  

There's a method `isSpanningDeletion` in GenotypeGVCFs that checks for both, this could be moved to `GATKVCFConstants`.

davidbenjamin:
Closes #2589.

@lbergelson since you raised the issue can you review?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
@@ -376,8 +376,10 @@ public abstract class GenotypingEngine<Config extends StandardCallerArgumentColl
                 final boolean isPlausible = afCalculationResult.isPolymorphicPhredScaledQual(allele, configuration.genotypeArgs.STANDARD_CONFIDENCE_FOR_CALLING);
 
                 siteIsMonomorphic &= !isPlausible;
-                final boolean isNonCoveringSpanningDeletion = GATKVCFConstants.isSpanningDeletion(allele) && !isVcCoveredByDeletion(vc);
-                final boolean toOutput = (isPlausible || forceKeepAllele(allele) || isNonRefWhichIsLoneAltAllele) && !isNonCoveringSpanningDeletion;
+
+                //it's possible that the upstream deletion that spanned this site was not emitted, mooting the symbolic spanning deletion allele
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingEngine.java
```diff
@@ -376,11 +376,9 @@ private OutputAlleleSubset calculateOutputAlleleSubset(final AFCalculationResult
                 final boolean isPlausible = afCalculationResult.isPolymorphicPhredScaledQual(allele, configuration.genotypeArgs.STANDARD_CONFIDENCE_FOR_CALLING);
 
                 siteIsMonomorphic &= !isPlausible;
-                boolean toOutput = (isPlausible || forceKeepAllele(allele) || isNonRefWhichIsLoneAltAllele);
-                if ( allele.equals(GATKVCFConstants.SPANNING_DELETION_SYMBOLIC_ALLELE_DEPRECATED) ||
-                        allele.equals(Allele.SPAN_DEL) ) {
-                    toOutput &= isVcCoveredByDeletion(vc);
-                }
+                final boolean isNonCoveringSpanningDeletion = GATKVCFConstants.isSpanningDeletion(allele) && !isVcCoveredByDeletion(vc);
+                final boolean toOutput = (isPlausible || forceKeepAllele(allele) || isNonRefWhichIsLoneAltAllele) && !isNonCoveringSpanningDeletion;
```
droazen: `!isNonCoveringSpanningDeletion` is a double negation, which makes the new version harder to understand than the old version.

davidbenjamin: Renamed to `isSpuriousSpanningDeletion`.

davidbenjamin:
@lbergelson Did you mean "request changes"?  I can't make the same change in protected without a snapshot of public that contains these changes.

davidbenjamin:
@lbergelson I'm waiting for your official git reviewer approval.

lbergelson:
@davidbenjamin ack!  I hit the wrong button.  I'm sorry!

PR #2659, opened 2017-05-04
Change the last GCVF partition band upper bound from strictly less than Integer.MAX_VALUE to MAX_GENOTYPE_QUAL

ronlevine:
Finish porting https://github.com/broadinstitute/gsa-unstable/pull/1440.
* Changed the last band from [lastValue,  Integer.MAX_VALUE) to [lastValue,  MAX_GENOTYPE_QUAL) .
* Changed the expected test data and compare against the output partition (value) rather than the input (key).
* Changed GVCFBlock in a test gvcf file.


ronlevine:
@samuelklee Please review.

samuelklee:
:+1: Looks good, thanks again!

PR #2661, opened 2017-05-04
Ll emit empty loci

LeeTL1220:
- Changes LocusWalker to emit uncovered loci if requested.  Default is previous behavior.
- Needed when there needs to be differentiation between "we do not care about this locus" versus "this locus has no coverage".

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
@@ -588,8 +588,7 @@ public final class AssemblyRegion implements Locatable {
 
         // Use the provided activity profile to determine the bounds of each assembly region:
         List<AssemblyRegion> assemblyRegions = new ArrayList<>();
-        final Iterable<AlignmentContext> locatableIterable = () -> locusIterator;
-        for ( final AlignmentContext pileup : locatableIterable ) {
+        locusIterator.forEachRemaining(pileup -> {
             if ( ! activityProfile.isEmpty() ) {
                 final boolean forceConversion = pileup.getLocation().getStart() != activityProfile.getEnd() + 1;
                 assemblyRegions.addAll(activityProfile.popReadyAssemblyRegions(assemblyRegionPadding, minRegionSize, maxRegionSize, forceConversion));
@@ -603,7 +602,8 @@ public final class AssemblyRegion implements Locatable {
                 final ActivityProfileState profile = evaluator.isActive(pileup, pileupRefContext, features);
                 activityProfile.add(profile);
             }
-        }
+        });
+
         assemblyRegions.addAll(activityProfile.popReadyAssemblyRegions(assemblyRegionPadding, minRegionSize, maxRegionSize, true));
 
         return assemblyRegions;
--- a/src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
@@ -588,7 +588,8 @@ public final class AssemblyRegion implements Locatable {
 
         // Use the provided activity profile to determine the bounds of each assembly region:
         List<AssemblyRegion> assemblyRegions = new ArrayList<>();
-        for ( final AlignmentContext pileup : locusIterator ) {
+        final Iterable<AlignmentContext> locatableIterable = () -> locusIterator;
+        for ( final AlignmentContext pileup : locatableIterable ) {
             if ( ! activityProfile.isEmpty() ) {
                 final boolean forceConversion = pileup.getLocation().getStart() != activityProfile.getEnd() + 1;
                 assemblyRegions.addAll(activityProfile.popReadyAssemblyRegions(assemblyRegionPadding, minRegionSize, maxRegionSize, forceConversion));
--- a/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
@@ -44,7 +44,7 @@ import java.util.stream.StreamSupport;
 public abstract class LocusWalker extends GATKTool {
 
     @Argument(fullName = "maxDepthPerSample", shortName = "maxDepthPerSample", doc = "Maximum number of reads to retain per sample per locus. Reads above this threshold will be downsampled. Set to 0 to disable.", optional = true)
-    protected int maxDepthPerSample = defaultMaxDepthPerSample();
+    protected long maxDepthPerSample = defaultMaxDepthPerSample();
 
     /**
      * Should the LIBS keep unique reads? Tools that do should override to return {@code true}.
@@ -98,7 +98,7 @@ public abstract class LocusWalker extends GATKTool {
      * Returns default value for the {@link #maxDepthPerSample} parameter, if none is provided on the command line.
      * Default implementation returns 0 (no downsampling by default).
      */
-    protected int defaultMaxDepthPerSample() {
+    protected long defaultMaxDepthPerSample() {
         return 0;
     }
 
@@ -161,6 +161,21 @@ public abstract class LocusWalker extends GATKTool {
         final Iterator<GATKRead> readIterator = getTransformedReadStream(countedFilter).iterator();
         // get the LIBS
         final LocusIteratorByState libs = new LocusIteratorByState(readIterator, getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
+        final Iterator<AlignmentContext> iterator = createAlignmentContextIterator(header, libs);
+
+        // iterate over each alignment, and apply the function
+        final Spliterator<AlignmentContext> spliterator = Spliterators.spliteratorUnknownSize(iterator, 0);
+        StreamSupport.stream(spliterator, false)
+            .forEach(alignmentContext -> {
+                        final SimpleInterval alignmentInterval = new SimpleInterval(alignmentContext);
+                        apply(alignmentContext, new ReferenceContext(reference, alignmentInterval), new FeatureContext(features, alignmentInterval));
+                        progressMeter.update(alignmentInterval);
+                }
+            );
+        logger.info(countedFilter.getSummaryLine());
+    }
+
+    private Iterator<AlignmentContext> createAlignmentContextIterator(SAMFileHeader header, LocusIteratorByState libs) {
         Iterator<AlignmentContext> iterator;
 
         validateEmitEmptyLociParameters();
@@ -177,17 +192,7 @@ public abstract class LocusWalker extends GATKTool {
             // prepare the iterator
             iterator = (hasIntervals()) ? new IntervalOverlappingIterator<>(libs, intervalsForTraversal, header.getSequenceDictionary()) : libs;
         }
-
-        // iterate over each alignment, and apply the function
-        final Spliterator<AlignmentContext> spliterator = Spliterators.spliteratorUnknownSize(iterator, 0);
-        StreamSupport.stream(spliterator, false)
-            .forEach(alignmentContext -> {
-                        final SimpleInterval alignmentInterval = new SimpleInterval(alignmentContext);
-                        apply(alignmentContext, new ReferenceContext(reference, alignmentInterval), new FeatureContext(features, alignmentInterval));
-                        progressMeter.update(alignmentInterval);
-                }
-            );
-        logger.info(countedFilter.getSummaryLine());
+        return iterator;
     }
 
     /**
--- a/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
@@ -44,7 +44,7 @@ import java.util.stream.StreamSupport;
 public abstract class LocusWalker extends GATKTool {
 
     @Argument(fullName = "maxDepthPerSample", shortName = "maxDepthPerSample", doc = "Maximum number of reads to retain per sample per locus. Reads above this threshold will be downsampled. Set to 0 to disable.", optional = true)
-    protected long maxDepthPerSample = defaultMaxDepthPerSample();
+    protected int maxDepthPerSample = defaultMaxDepthPerSample();
 
     /**
      * Should the LIBS keep unique reads? Tools that do should override to return {@code true}.
@@ -98,7 +98,7 @@ public abstract class LocusWalker extends GATKTool {
      * Returns default value for the {@link #maxDepthPerSample} parameter, if none is provided on the command line.
      * Default implementation returns 0 (no downsampling by default).
      */
-    protected long defaultMaxDepthPerSample() {
+    protected int defaultMaxDepthPerSample() {
         return 0;
     }
 
--- a/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
@@ -84,6 +84,8 @@ public abstract class LocusWalker extends GATKTool {
      *
      * NOTE:  Typically, this should only be used when intervals are specified.
      * NOTE:  If MappedReadFilter is removed, then emitting empty loci will fail.
+     * NOTE:  If there is no available sequence dictionary and this is set to true, there should be a failure.  Please
+     *  consider requiring reads and/or references for all tools that wish to set this to {@code true}.
      *
      * @return {@code true} if this tool requires uncovered loci information to be emitted, {@code false} otherwise
      */
@@ -224,11 +226,6 @@ public abstract class LocusWalker extends GATKTool {
             if (!hasReference() && !hasIntervals()) {
                 logger.warn("****************************************");
                 logger.warn("* Running this tool without a reference nor intervals can yield unexpected results, since it will emit results for loci with no reads.  A sequence dictionary has been found.  The easiest way avoid this message is to specify a reference.");
-                logger.warn("****************************************");
-            }
-            if (!hasIntervals() && hasReference()) {
-                logger.warn("****************************************");
-                logger.warn("* Running this tool without intervals will cause this tool to run on the entire genome, since it will emit results for loci with no reads.");
                 logger.warn("****************************************");
             }
         }
--- a/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
@@ -2,13 +2,19 @@ package org.broadinstitute.hellbender.engine;
 
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMReadGroupRecord;
+import htsjdk.samtools.util.SequenceUtil;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.CommandLineException;
 import org.broadinstitute.hellbender.engine.filters.CountingReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.filters.WellformedReadFilter;
+import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.transformers.ReadTransformer;
+import org.broadinstitute.hellbender.utils.IntervalUtils;
+import org.broadinstitute.hellbender.utils.SequenceDictionaryUtils;
+import org.broadinstitute.hellbender.utils.iterators.IntervalAlignmentContextIterator;
+import org.broadinstitute.hellbender.utils.iterators.IntervalLocusIterator;
 import org.broadinstitute.hellbender.utils.iterators.IntervalOverlappingIterator;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.locusiterator.LIBSDownsamplingInfo;
@@ -72,6 +78,19 @@ public abstract class LocusWalker extends GATKTool {
     public boolean includeNs() {
         return false;
     }
+
+    /**
+     * Does this tool emit information for uncovered loci? Tools that do should override to return {@code true}.
+     *
+     * NOTE:  Typically, this should only be used when intervals are specified.
+     * NOTE:  If MappedReadFilter is removed, then emitting empty loci will fail.
+     *
+     * @return {@code true} if this tool requires uncovered loci information to be emitted, {@code false} otherwise
+     */
+    public boolean emitEmptyLoci() {
+        return false;
+    }
+
 
     /**
      * Returns default value for the {@link #maxDepthPerSample} parameter, if none is provided on the command line.
@@ -140,8 +159,23 @@ public abstract class LocusWalker extends GATKTool {
         final Iterator<GATKRead> readIterator = getTransformedReadStream(countedFilter).iterator();
         // get the LIBS
         final LocusIteratorByState libs = new LocusIteratorByState(readIterator, getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
-        // prepare the iterator
-        final Spliterator<AlignmentContext> iterator = (hasIntervals()) ? new IntervalOverlappingIterator<>(libs, intervalsForTraversal, header.getSequenceDictionary()).spliterator() : libs.spliterator();
+        Spliterator<AlignmentContext> iterator;
+
+        validateEmitEmptyLociParameters();
+        if (emitEmptyLoci()) {
+
+            // If no intervals were specified, then use the entire reference (or best available sequence dictionary).
+            if (intervalsForTraversal == null) {
+                intervalsForTraversal = IntervalUtils.getAllIntervalsForReference(getBestAvailableSequenceDictionary());
+            }
+            final IntervalLocusIterator intervalLocusIterator = new IntervalLocusIterator(intervalsForTraversal.iterator());
+            iterator =  new IntervalAlignmentContextIterator(libs, intervalLocusIterator, header.getSequenceDictionary()).spliterator();
+
+        } else {
+            // prepare the iterator
+            iterator = (hasIntervals()) ? new IntervalOverlappingIterator<>(libs, intervalsForTraversal, header.getSequenceDictionary()).spliterator() : libs.spliterator();
+        }
+
         // iterate over each alignment, and apply the function
         StreamSupport.stream(iterator, false)
             .forEach(alignmentContext -> {
@@ -176,5 +210,27 @@ public abstract class LocusWalker extends GATKTool {
     protected final void onShutdown() {
         // Overridden only to make final so that concrete tool implementations don't override
         super.onShutdown();
+    }
+
+    /**
+     *  The emit empty loci parameter comes with several pitfalls when used incorrectly.  Here we check and either give
+     *   warnings or errors.
+     */
+    protected void validateEmitEmptyLociParameters() {
+        if (emitEmptyLoci()) {
+            if (getBestAvailableSequenceDictionary() == null) {
+                throw new UserException.MissingReference("Could not create a sequence dictionary nor find a reference.  Therefore, emitting empty loci is impossible and this tool cannot be run.  The easiest fix here is to specify a reference dictionary.");
+            }
+            if (!hasReference() && !hasIntervals()) {
+                logger.warn("****************************************");
+                logger.warn("* Running this tool without a reference nor intervals can yield unexpected results, since it will emit results for loci with no reads.  A sequence dictionary has been found.  The easiest way avoid this message is to specify a reference.");
+                logger.warn("****************************************");
+            }
+            if (!hasIntervals() && hasReference()) {
+                logger.warn("****************************************");
+                logger.warn("* Running this tool without intervals will cause this tool to run on the entire genome, since it will emit results for loci with no reads.");
+                logger.warn("****************************************");
+            }
+        }
     }
 }
--- a/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
@@ -161,7 +161,7 @@ public abstract class LocusWalker extends GATKTool {
         final Iterator<GATKRead> readIterator = getTransformedReadStream(countedFilter).iterator();
         // get the LIBS
         final LocusIteratorByState libs = new LocusIteratorByState(readIterator, getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
-        Spliterator<AlignmentContext> iterator;
+        Iterator<AlignmentContext> iterator;
 
         validateEmitEmptyLociParameters();
         if (emitEmptyLoci()) {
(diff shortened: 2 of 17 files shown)


src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+/**
+ * For special cases where we want to emit AlignmentContexts regardless of whether we have an overlap with a given interval.
+ *
+ * In other words, this will iterate over each base of each interval and emit an alignment context.  If no reads overlap,
+ *  it will emit an empty alignment context.  Each empty alignment context will be a new instance.
+ */
+public class IntervalAlignmentContextIterator implements Iterable<AlignmentContext>, Iterator<AlignmentContext> {
```
magicDGS: If the method in LIBS is implemented, this could live in the `locusiterator` package and reduce visibility, because I do not see any other usage that the one implemented here.

LeeTL1220:
Needed for:

https://github.com/broadinstitute/gatk-protected/issues/987

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -74,6 +80,19 @@ public boolean includeNs() {
...
 
     /**
+     * Does this tool emit information for uncovered loci? Tools that do should override to return {@code true}.
+     *
+     * NOTE:  Typically, this should only be used when intervals are specified.
+     * NOTE:  If MappedReadFilter is removed, then emitting empty loci will fail.
+     *
+     * @return {@code true} if this tool requires uncovered loci information to be emitted, {@code false} otherwise
+     */
+    public boolean emitEmptyLoci() {
```
LeeTL1220: Please see issues #2678 and #2679 

We will address this in a separate pull request that will possibly refactor the code changes in this PR.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -140,8 +159,23 @@ public void traverse() {
         final Iterator<GATKRead> readIterator = getTransformedReadStream(countedFilter).iterator();
         // get the LIBS
         final LocusIteratorByState libs = new LocusIteratorByState(readIterator, getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
-        // prepare the iterator
-        final Spliterator<AlignmentContext> iterator = (hasIntervals()) ? new IntervalOverlappingIterator<>(libs, intervalsForTraversal, header.getSequenceDictionary()).spliterator() : libs.spliterator();
+        Spliterator<AlignmentContext> iterator;
+
+        validateEmitEmptyLociParameters();
+        if (emitEmptyLoci()) {
```
LeeTL1220: Filed issues.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -177,4 +211,26 @@ protected final void onShutdown() {
...
     }
+
+    /**
+     *  The emit empty loci parameter comes with several pitfalls when used incorrectly.  Here we check and either give
+     *   warnings or errors.
+     */
+    protected void validateEmitEmptyLociParameters() {
+        if (emitEmptyLoci()) {
+            if (getBestAvailableSequenceDictionary() == null) {
+                throw new UserException.MissingReference("Could not create a sequence dictionary nor find a reference.  Therefore, emitting empty loci is impossible and this tool cannot be run.  The easiest fix here is to specify a reference dictionary.");
```
LeeTL1220: I am keeping this as a user exception as an extra level of protection.  However, I have also added comments to javadoc.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -177,4 +211,26 @@ protected final void onShutdown() {
...
+            }
+            if (!hasReference() && !hasIntervals()) {
+                logger.warn("****************************************");
+                logger.warn("* Running this tool without a reference nor intervals can yield unexpected results, since it will emit results for loci with no reads.  A sequence dictionary has been found.  The easiest way avoid this message is to specify a reference.");
+                logger.warn("****************************************");
+            }
+            if (!hasIntervals() && hasReference()) {
+                logger.warn("****************************************");
+                logger.warn("* Running this tool without intervals will cause this tool to run on the entire genome, since it will emit results for loci with no reads.");
```
LeeTL1220: I discussed with @lbergelson  ... I figured that most people would be using this to find coverage at certain loci.  And that if someone was doing this over the whole genome, it was probably an accident.  We do not have strong opinions on this, though.

For now, I am leaving it in.

magicDGS: I think that the correct place to log the warning should be in concrete implementations, because the abstraction itself is not aware of putative usages where may not be the case and this will show a warning to the user and may be scary for them...

Anyway, I'm fine with it but it will be nicer to do not warn always because it is something that may be a normal usage (at least for some of my tools).

LeeTL1220: You are a real user and you do not like it.  I have removed it.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalLocusIterator.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+
+    private Iterator<SimpleInterval> createBaseLocationIterator(final SimpleInterval fullInterval) {
+        if (fullInterval == null) {
+            return Collections.<SimpleInterval>emptyList().iterator();
+        }
+
+        List<SimpleInterval> dummyList = Collections.singletonList(fullInterval);
+        return IntervalUtils.cutToShards(dummyList, 1).iterator();
```
LeeTL1220: Fixed with ``ShardedIntervalIterator``... Good catch.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
LeeTL1220: Agreed.  The issues are filed for this work.  Not going to be handled in this PR, though.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
magicDGS: Thanks @LeeTL1220! :)

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
magicDGS: I can take care of that issues, because it will be useful for me in my personal work... Thanks for open the issues!

src/main/java/org/broadinstitute/hellbender/utils/iterators/ShardedIntervalIterator.java
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Iterator that will break up each input interval into shards.  The advantage of this iterator is that it is not RAM
+ *  intensive.
+ *
+ *  Empty iterator of intervals is supported.
+ *
+ * Developer note:  This class queues up the next interval to return and then has (minimal) code to detect when
+ *  an empty iterator was passed in for initialization.
+ */
+public class ShardedIntervalIterator implements Iterable<SimpleInterval>, Iterator<SimpleInterval> {
```
magicDGS: This functionality is cool and very usuful! Thanks for it!

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
```diff
@@ -0,0 +1,109 @@
...
+        while (IntervalUtils.compareLocatables(currentInterval, currentAlignmentContext, dictionary) > 0) {
+            advanceAlignmentContext();
+        }
+    }
+
+    private void advanceAlignmentContext() {
+        if (alignmentContextIterator.hasNext()) {
+            currentAlignmentContext = this.alignmentContextIterator.next();
+        } else {
+            currentAlignmentContext = createEmptyAlignmentContext(currentInterval);
```
vruano: If ```currentInterval``` is ```null``` is this going to blow?

For example if there is not locus in the locus iterator I think this will happen.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public IntervalAlignmentContextIterator(Iterator<AlignmentContext> alignmentContextIterator, IntervalLocusIterator intervalLocusIterator, SAMSequenceDictionary dictionary) {
+        this.alignmentContextIterator = alignmentContextIterator;
+        this.intervalLocusIterator = intervalLocusIterator;
+        this.dictionary = dictionary;
+
+        // Iterate once.
+        advanceIntervalLocus();
+        advanceAlignmentContext();
+        advanceAlignmentContextToCurrentInterval(this.currentInterval);
```
vruano: If you are mean to advance the alignmentContext to the currentInterval why bother to call ```advanceAlignmentContext()``` in between? of course need to look into the code to fully understand this but it looks a bit odd.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
```diff
@@ -0,0 +1,109 @@
...
+
+        return result;
+
+    }
+
+    private AlignmentContext createEmptyAlignmentContext(final SimpleInterval interval) {
+        return new AlignmentContext(interval, new ReadPileup(interval, new ArrayList<>()));
+    }
+
+    private void advanceAlignmentContextToCurrentInterval(final SimpleInterval interval) {
```
vruano: I would do some refactoring on advanceAlignmentContextToCurrentInterval and advanceCurrentInterval...

for example the later accesses the currentInterval field directly whereas the other take is as a parameters although it is always given the currentInterval, so what is the point?

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalOverlappingIterator.java
```diff
@@ -48,6 +48,8 @@ public IntervalOverlappingIterator(Iterator<T> iterator, List<SimpleInterval> in
         this.iterator = iterator;
         this.intervals = intervals.iterator();
         this.dictionary = dictionary;
+
+        //TODO: If the input intervals is empty, won't this throw an exception?  Is that planned?
```
vruano: Yes it would so please address.... try to fail early!

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
LeeTL1220: I agree with @vruano, Iterable means that it is a collection.  The interface has spliterator and forEach.

If LocusWalker (and LocusWalkerSpark) can handle doing an iterator and forEachRemaining call, instead (of spliterator and forEach).  Under the hood, the spliterator call here is just grabbing an iterator anyway.

The problem that I see is that we may require a spliterator in the future.  But to solve this, I would rather just instantiate the spliterator explicitly from the iterator (``Spliterators.spliteratorUnknownSize(iterator, Spliterator.ORDERED)``).  That way, @vruano is satisfied and we keep the flexibility of spliterator.  In another branch I have a Factory class, which I can leverage to encapsulate away this logic.

LeeTL1220: And ...
http://stackoverflow.com/questions/5836174/implement-java-iterator-and-iterable-in-same-class

LeeTL1220: Purged a lot of Iterable....

@vruano Are you okay with  ``Iterable<Blah> iterable = () -> iteratorBlah``?  Needs to be done in a couple of places to support the foreach.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
```diff
@@ -0,0 +1,109 @@
...
+        this.dictionary = dictionary;
+
+        // Iterate once.
+        advanceIntervalLocus();
+        advanceAlignmentContext();
+        advanceAlignmentContextToCurrentInterval(this.currentInterval);
+    }
+
+    @Override
+    public Iterator<AlignmentContext> iterator() {
```
LeeTL1220: This has been eliminated.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalLocusIterator.java
```diff
@@ -0,0 +1,78 @@
...
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.Utils;
+
+
+import java.util.*;
+
+/**
+ * Returns a SimpleInterval for each locus in a set of intervals.  I.e. returns each genomic point location in an interval list.
+ */
+public class IntervalLocusIterator implements Iterable<SimpleInterval>, Iterator<SimpleInterval> {
```
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -177,4 +211,26 @@ protected final void onShutdown() {
...
+     *  The emit empty loci parameter comes with several pitfalls when used incorrectly.  Here we check and either give
+     *   warnings or errors.
+     */
+    protected void validateEmitEmptyLociParameters() {
+        if (emitEmptyLoci()) {
+            if (getBestAvailableSequenceDictionary() == null) {
+                throw new UserException.MissingReference("Could not create a sequence dictionary nor find a reference.  Therefore, emitting empty loci is impossible and this tool cannot be run.  The easiest fix here is to specify a reference dictionary.");
+            }
+            if (!hasReference() && !hasIntervals()) {
+                logger.warn("****************************************");
```
LeeTL1220: I do not believe that the missing reference one is presumptuous.  If you cannot get a sequence dictionary of any kind and you have no intervals, then you can't determine the loci to traverse if you are emitting uncovered loci.

As for the second one (``!hasReference() && !hasIntervals()``) this means that the GATK created a dictionary from some other source.  The user would have to be pretty advanced to know how that would play out.  

We can discuss more, but I would like to leave this one in.  Especially, since I think most users using emit empty loci will have forgotten their interval file and/or reference if this warning gets triggered.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
LeeTL1220: It does look a bit odd.  I just wrote some comments to explain.  If this is not good enough to address, tell me.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
```diff
@@ -0,0 +1,109 @@
...
+        return this;
+    }
+
+    @Override
+    public boolean hasNext() {
+        return (intervalLocusIterator.hasNext()) || (currentInterval != null);
```
LeeTL1220: I think you are right.  

Done.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
LeeTL1220: Thought it would be more flexible, but then it had bugs.  I've refactored as you suggest.

Done.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
LeeTL1220: Added a test for empty interval list ``testIntervalIteratorIsEmpty()`` and it passes.

Technically, yes, if null it will throw an exception when trying to create the empty alignment context, but null is never passed in.

Done (unless you are not satisfied).

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -140,8 +159,23 @@ public void traverse() {
         final Iterator<GATKRead> readIterator = getTransformedReadStream(countedFilter).iterator();
         // get the LIBS
         final LocusIteratorByState libs = new LocusIteratorByState(readIterator, getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
-        // prepare the iterator
-        final Spliterator<AlignmentContext> iterator = (hasIntervals()) ? new IntervalOverlappingIterator<>(libs, intervalsForTraversal, header.getSequenceDictionary()).spliterator() : libs.spliterator();
+        Spliterator<AlignmentContext> iterator;
```
LeeTL1220: Made a private method.

Done.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -74,6 +80,19 @@ public boolean includeNs() {
...
+     * @return {@code true} if this tool requires uncovered loci information to be emitted, {@code false} otherwise
+     */
+    public boolean emitEmptyLoci() {
+        return false;
+    }
+
+
+    /**
      * Returns default value for the {@link #maxDepthPerSample} parameter, if none is provided on the command line.
      * Default implementation returns 0 (no downsampling by default).
```
LeeTL1220: Changed to long (had to propagate to several files)

Done

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalOverlappingIterator.java
LeeTL1220: It does.  

Done.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+public class IntervalAlignmentContextIterator implements Iterable<AlignmentContext>, Iterator<AlignmentContext> {
+    private Iterator<AlignmentContext> alignmentContextIterator;
+    private IntervalLocusIterator intervalLocusIterator;
+    private SimpleInterval currentInterval;
+    private AlignmentContext currentAlignmentContext;
+    private SAMSequenceDictionary dictionary;
+
+    public IntervalAlignmentContextIterator(Iterator<AlignmentContext> alignmentContextIterator, IntervalLocusIterator intervalLocusIterator, SAMSequenceDictionary dictionary) {
+        this.alignmentContextIterator = alignmentContextIterator;
```
LeeTL1220: Addressed.

LeeTL1220:
@vruano Comments addressed.  Back to you for another review.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
LeeTL1220: (the exception will be InvalidArgument, not NullPointer)

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
```diff
@@ -588,7 +588,8 @@ public boolean isFinalized() {
 
         // Use the provided activity profile to determine the bounds of each assembly region:
         List<AssemblyRegion> assemblyRegions = new ArrayList<>();
-        for ( final AlignmentContext pileup : locusIterator ) {
+        final Iterable<AlignmentContext> locatableIterable = () -> locusIterator;
+        for ( final AlignmentContext pileup : locatableIterable ) {
```
droazen: Can you explain why you decided to make `LocusIteratorByState` non-iterable? This does not strike me as an improvement from a client perspective...

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
droazen: Recommend keeping this as a simple boolean flag -- adding a secondary layer of intervals seems overly complex, and it's unclear that there would be a use case for it.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
LeeTL1220: Actually, going to long is not trivial.

Also, this is actually number of reads per sample-locus, which means we need to go to long if we have any loci w/ coverage > 2 billion...  Not only that, this would be post downsampling.  So, we could (possibly) still support > 2 billion so long as that was not downsampled.

Rolling this back to int.  @vruano, is this okay?  Lots of other stuff will break anyway, if we were to need a long here.  (Per conversation with @lbergelson )

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
droazen: Never mind, found the relevant discussion in the code review history.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
vruano: I don't think long or int makes much of a difference, whatever it is easier.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
vruano: As we discuss, I think you should be using Iterator's ```forEachRemaining```

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -161,7 +161,7 @@ public void traverse() {
         final Iterator<GATKRead> readIterator = getTransformedReadStream(countedFilter).iterator();
         // get the LIBS
         final LocusIteratorByState libs = new LocusIteratorByState(readIterator, getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
-        Spliterator<AlignmentContext> iterator;
+        Iterator<AlignmentContext> iterator;
```
vruano: Can this be final. Also did you were going to encapsulate the initialization in a private method?, perhaps it is in the next commit, disregard if that is the case.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -171,15 +171,16 @@ public void traverse() {
...
         } else {
             // prepare the iterator
-            iterator = (hasIntervals()) ? new IntervalOverlappingIterator<>(libs, intervalsForTraversal, header.getSequenceDictionary()).spliterator() : libs.spliterator();
+            iterator = (hasIntervals()) ? new IntervalOverlappingIterator<>(libs, intervalsForTraversal, header.getSequenceDictionary()) : libs;
         }
 
         // iterate over each alignment, and apply the function
-        StreamSupport.stream(iterator, false)
+        final Spliterator<AlignmentContext> spliterator = Spliterators.spliteratorUnknownSize(iterator, 0);
+        StreamSupport.stream(spliterator, false)
```
vruano: Why create the stream rather than use Iterator's ```.forEachRemaining```?

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalLocusIterator.java
```diff
@@ -26,11 +26,6 @@ public IntervalLocusIterator(final Iterator<SimpleInterval> intervalIterator) {
     }
 
     @Override
-    public Iterator<SimpleInterval> iterator() {
-        return this;
-    }
-
-    @Override
```
vruano: The override should stay.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalOverlappingIterator.java
```diff
@@ -53,11 +53,6 @@ public IntervalOverlappingIterator(Iterator<T> iterator, List<SimpleInterval> in
     }
 
     @Override
-    public Iterator<T> iterator() {
-        return this;
-    }
-
-    @Override
```
vruano: Same.

src/main/java/org/broadinstitute/hellbender/utils/iterators/ShardedIntervalIterator.java
```diff
@@ -70,11 +70,6 @@ private SimpleInterval calculateShardedInterval() {
     }
 
     @Override
-    public Iterator<SimpleInterval> iterator() {
-        return this;
-    }
-
-    @Override
```
vruano: Same.

vruano: Ha! sorry I realize now that you are just getting the override from the other method, so is fine, ok forget this one and the previous too.

src/test/java/org/broadinstitute/hellbender/utils/iterators/IntervalOverlappingIteratorUnitTest.java
```diff
@@ -58,7 +60,8 @@
     public void testIterator(List<SimpleInterval> intervals, SAMSequenceDictionary dictionary, Locatable[] records,	Locatable[] expected) {
         IntervalOverlappingIterator<Locatable> iterator = new IntervalOverlappingIterator<>(Arrays.asList(records).iterator(), intervals, dictionary);
         int index = 0;
-        for(Locatable loc: iterator) {
+        final Iterable<Locatable> iterable = () -> iterator;
+        for(Locatable loc: iterable) {
```
vruano: Use ```iterator.forEachRemaining(...)```

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -161,6 +161,21 @@ public void traverse() {
         final Iterator<GATKRead> readIterator = getTransformedReadStream(countedFilter).iterator();
         // get the LIBS
         final LocusIteratorByState libs = new LocusIteratorByState(readIterator, getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
+        final Iterator<AlignmentContext> iterator = createAlignmentContextIterator(header, libs);
```
vruano: Ok here it is.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -177,17 +192,7 @@ public void traverse() {
...
-        final Spliterator<AlignmentContext> spliterator = Spliterators.spliteratorUnknownSize(iterator, 0);
-        StreamSupport.stream(spliterator, false)
-            .forEach(alignmentContext -> {
-                        final SimpleInterval alignmentInterval = new SimpleInterval(alignmentContext);
-                        apply(alignmentContext, new ReferenceContext(reference, alignmentInterval), new FeatureContext(features, alignmentInterval));
-                        progressMeter.update(alignmentInterval);
-                }
-            );
-        logger.info(countedFilter.getSummaryLine());
+        return iterator;
```
vruano: It does not seem to be a need to keep the return statement at the end, you could have a return in conditional block. In any case if you keep it like this then make the iterator variable final.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
```diff
@@ -30,13 +30,17 @@ public IntervalAlignmentContextIterator(final Iterator<AlignmentContext> alignme
 
         // Iterate once.
         advanceIntervalLocus();
+
+        // Just to make sure that current alignment context is not null.  We only want to do this during the initialization.
         advanceAlignmentContext();
```
vruano: Is the comment correct? Isn't the case that ```advanceAlignmentContextToCurrentInterval()``` calls ```advanceAlignmentContext()```?

vruano:
Done with second round of reviews.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
LeeTL1220: @vruano I think he should not be an error.  An advanced user may know what they were doing -- especially if they know how getBestSequenceDictionary() works.  Have I convinced you?

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
LeeTL1220: No further action.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegion.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalLocusIterator.java
LeeTL1220: It did.  It's an artifact of how github rendered this change

No action.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalOverlappingIterator.java
LeeTL1220: Ditto.

No action.

src/test/java/org/broadinstitute/hellbender/utils/iterators/IntervalOverlappingIteratorUnitTest.java
LeeTL1220: Harder here, but done

LeeTL1220:
@vruano Addressed second round of comments.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
magicDGS: I support the warning (although I prefer not having it) because is an use case where you input the reads with header and then traverse everything that is there, including no covered regions...

src/test/java/org/broadinstitute/hellbender/utils/iterators/IntervalOverlappingIteratorUnitTest.java
```diff
@@ -59,13 +57,15 @@
     @Test(dataProvider = "data")
     public void testIterator(List<SimpleInterval> intervals, SAMSequenceDictionary dictionary, Locatable[] records,	Locatable[] expected) {
         IntervalOverlappingIterator<Locatable> iterator = new IntervalOverlappingIterator<>(Arrays.asList(records).iterator(), intervals, dictionary);
-        int index = 0;
-        final Iterable<Locatable> iterable = () -> iterator;
-        for(Locatable loc: iterable) {
+
```
vruano: Ok I see why it is more difficult, in truth this is just testing code and I normally I just don't care much about it unless the coverage is bad...

I guess it my review was an automatics response when I saw again the iterator to iterable transformation trick. 

I would say that it is sufficient to add al the elements returned by the iterator and compare the corresponding list directly with the expected list without bothering to transform the expected list into an iterator and compare them that way.... up to you, I will accept this as it is since, again, I don't care about test code that much.

LeeTL1220: I'm going to leave it as is.  I may make Lists in the future.

PR #2662, opened 2017-05-04
Add masking and spacing for SVKmers

mwalker174:
@tedsharpe please review.

- SVKmerizer takes in an integer specifying the spacing between between kmers. This is is an effective way to reduce the kmer set size without affecting sensitivity much.
- SVKmerShort - added masking function that returns a copy of the current kmer after deleting bases at the specified positions
- Reworded some error messages about kmer length
- Moved and added some hashing functions to SVUtils, which will be used in another PR for the long-typed set classes and de-duplication filter.


cwhelan:
@tedsharpe can you review this one?

tedsharpe:
@cwhelan Yes.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerLong.java
```diff
@@ -152,6 +150,10 @@ public final int compareTo( final SVKmerLong that ) {
         return result;
     }
 
+    public SVKmer mask(final byte[] mask, final int kSize) {
```
tedsharpe: This becomes:
```
@Override
public SVKmerLong mask( final SVKmerLong mask ) {
    valHigh &= mask.valHigh;
    valLow &= mask.valLow;
}
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerShort.java
```diff
@@ -106,40 +112,67 @@ public SVKmerShort canonical( final int kSize ) {
...
         return Long.compare(this.valLow, that.valLow);
     }
 
     /**
+     * Masks the current kmer by removing the specified bases.
+     * The mask array must be of even length, contain integers between 0 and kSize-1, and be in ascending order
+     * @param mask indices of masked bases
+     * @return kmer of length kSize - mask.length, with masked bases removed
+     */
+    public SVKmer mask(final byte[] mask, final int kSize) {
```
tedsharpe: This becomes:
```
@Override
public SVKmerShort mask( final SVKmerShort mask ) {
    valLow &= mask.valLow;
}
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizer.java
```diff
@@ -41,26 +56,25 @@ public boolean hasNext() {
...
-
-    public static Stream<SVKmer> stream( final CharSequence seq, final int kSize, final SVKmer kmer ) {
-        return Utils.stream(new SVKmerizer(seq, kSize, kmer));
+    public static Stream<SVKmer> stream(final CharSequence seq, final int kSize, final int kSpace, final SVKmer kmer) {
+        return StreamSupport.stream(((Iterable<SVKmer>)() -> new SVKmerizer(seq, kSize, kSpace, kmer)).spliterator(), false);
     }
 
-    public static Stream<SVKmer> stream( final byte[] seq, final int kSize, final SVKmer kmer ) {
-        return Utils.stream(new SVKmerizer(seq, kSize, kmer));
+    public static Stream<SVKmer> stream(final byte[] seq, final int kSize, final int kSpace, SVKmer kmer ) {
```
tedsharpe: Ditto, overload rather than replace.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVUtils.java
```diff
@@ -146,6 +146,53 @@ public static int hashMapCapacity( final int nElements )
     }
 
     /**
+     * 64-bit FNV-1a hash for long's
+     */
+    public static long fnvLong64( long start, final long toHash ) {
```
tedsharpe: Not sure why I didn't use the 64-bit FNV-1a in the first place.  Let's use this one exclusively and kill fnvLong and fnvInt.  But now that this has moved out into independent space (rather than being encapsulated in SVKmer), I think we need a 1-arg fnvLong64 that supplies the correct initial seed:
```
public static long fnvLong64( final long toHash ) {
    return fnvLong64(14695981039346656037L, toHash);
}
```

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerShortUnitTest.java
```diff
@@ -28,40 +30,152 @@ public void testDefaultConstructionWithTooSmallK() {
...
+        final Random rand = new Random(2738493L);
+        for (int i = 0; i < seqLen - K; i++) {
+
+            //Random k-mer
+            String substr = bases.substring(i, i + K);
+
+            //For generating the correct answer
+            final StringBuilder sb = new StringBuilder(substr);
+
+            //Generate random indices to mask, an even number of them so that the k-mer will still be odd
```
tedsharpe: I think there's going to be a problem with canonicalization if you happen to choose the middle base as one of the ones to be masked.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerShort.java
tedsharpe: My goof:  you can't have covariant args, just covariant return types.  You'll have to downcast an SVKmer.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizer.java
```diff
@@ -41,26 +56,25 @@ public boolean hasNext() {
     public SVKmer next() {
         if ( nextKmer == null ) throw new NoSuchElementException("Kmerization sequence exhausted.");
         final SVKmer result = nextKmer;
-        nextKmer = nextKmer(nextKmer, kSize-1);
+        nextKmer = nextKmer(nextKmer, kSize-kSpace);
```
tedsharpe: This is the only place you use kSpace.  (This is very slick, by the way.  Love the economy of it.)
So you might as well precompute this difference in the constructor.  (I.e., have a field called kAdvance, rather than kSpace, and initialize it to ```kSize-kSpace``` in the constructor.)
Tiny performance improvement, but this is likely to be called billions of times.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqKmerSpark.java
```diff
@@ -90,7 +90,7 @@ protected void runTool( final JavaSparkContext ctx ) {
      */
     private static List<SVKmer> processRefRDD(final int kSize, final JavaRDD<byte[]> refRDD ) {
         return refRDD.flatMap(seq ->
-                    SVKmerizer.stream(seq, kSize, new SVKmerShort(kSize))
+                    SVKmerizer.stream(seq, kSize, 1, new SVKmerShort(kSize))
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmer.java
```diff
@@ -35,8 +37,11 @@ public SVKmer() {}
     public abstract Base lastBase();
     public abstract int firstTrimer(final int kSize );
     public abstract int lastTrimer();
+    public abstract SVKmer mask(final byte[] mask, final int kSize);
```
mwalker174: Still kicking myself that I didn't just do this to begin with. Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmer.java
```diff
@@ -35,8 +37,11 @@ public SVKmer() {}
...
     public abstract String toString( final int kSize );
 
+    public long getLong() { throw new GATKException("getLong() not implemented for this SVKmer"); }
```
mwalker174: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerLong.java
mwalker174: Not going to implement for SVKmerLong, see above.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerShort.java
mwalker174: Not making this an SVKmer abstract method, see above.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizer.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizer.java
```diff
@@ -41,26 +56,25 @@ public boolean hasNext() {
...
         return sk.next();
     }
 
     public static SVKmer toKmer( final byte[] seq, final SVKmer kmer ) {
         return toKmer(new ASCIICharSequence(seq),kmer);
     }
-
-    public static Stream<SVKmer> stream( final CharSequence seq, final int kSize, final SVKmer kmer ) {
-        return Utils.stream(new SVKmerizer(seq, kSize, kmer));
+    public static Stream<SVKmer> stream(final CharSequence seq, final int kSize, final int kSpace, final SVKmer kmer) {
```
mwalker174: Done, also with the constructors.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerizer.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVUtils.java
mwalker174: Done

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/SVKmerShortUnitTest.java
mwalker174: You're probably right. With the new masking scheme I've reworked these tests.

PR #2663, opened 2017-05-04
PathSeq classes for taxonomy bookkeeping and abundance calcs

mwalker174:
@cwhelan please review.

Helper classes for the tools that use taxonomy data - namely one to build the taxonomic database and one to read it in and use it to compute microbial abundances from an aligned BAM. Most are just small data containers, with the exception of PSTree.


src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
```diff
@@ -0,0 +1,322 @@
...
+    /**
+     * Find all nodes reachable from the root
+     */
+    private Set<String> traverse() {
+        final Queue<String> queue = new LinkedList<>();
+        final Set<String> visited = new HashSet<>(tree.size());
+        queue.add(root);
+        while (!queue.isEmpty()) {
+            final String id = queue.poll();
+            if (!visited.contains(id)) { //checked visited in case there are cycles
```
cwhelan: Would it be better to throw an exception here? I'm worried that it could cause a loop in `getPathOf` for some reason the tree actually did contain cycles.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
```diff
@@ -0,0 +1,322 @@
...
+    /**
+     * Removes all nodes not in the given set
+     */
+    public void retainNodes(final Set<String> idsToKeep) {
+        Utils.validateArg(idsToKeep.contains(root), "Cannot remove root");
+        final HashMap<String, PSTreeNode> new_tree = new HashMap<>(idsToKeep.size());
+        for (final String id : tree.keySet()) {
+            if (idsToKeep.contains(id)) {
+                final PSTreeNode info = tree.get(id);
+                final PSTreeNode new_info = info.copy();
```
cwhelan: camelCase

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
```diff
@@ -0,0 +1,322 @@
...
+    public boolean hasNode(final String id) {
+        return tree.containsKey(id);
+    }
+
+    /**
+     * Removes all nodes not in the given set
+     */
+    public void retainNodes(final Set<String> idsToKeep) {
+        Utils.validateArg(idsToKeep.contains(root), "Cannot remove root");
+        final HashMap<String, PSTreeNode> new_tree = new HashMap<>(idsToKeep.size());
```
cwhelan: camelCase

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
```diff
@@ -0,0 +1,322 @@
...
+    public void setLengthOf(final String id, final long length) {
+        Utils.validateArg(tree.containsKey(id), "Could not set rank of node id " + id + " because it does not exist");
+        Utils.validateArg(!id.equals(root), "Cannot set rank of root");
+        tree.get(id).setLength(length);
+    }
+
+    /**
+     * Returns path of node id's from the input id to the root.
+     */
+    public List<String> getPathOf(String id) {
```
cwhelan: As I mentioned before, it looks like this will infinite loop if the tree has a cycle. Perhaps you want to check for that an throw an exception, or at least document why if you don't think that'll happen in practice.

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTaxInfoTest.java
```diff
@@ -0,0 +1,64 @@
...
+        final PSTaxInfo infoA2 = new PSTaxInfo();
+        infoA2.parent_tax = "1";
+        infoA2.length = 1000;
+        infoA2.name = "tax A";
+        infoA2.rank = "species";
+        infoA2.ref_names.add("ref|NC_A2");
+        infoA2.ref_names.add("ref|NC_A1");
+
+        Assert.assertEquals(infoA, infoA2);
+        Assert.assertNotEquals(infoA.hashCode(), infoA2.hashCode());
```
cwhelan: In general the contract for `hashCode` should be that if the objects are equal according to `equals`, their hash codes should be the same. If you ever try to put these objects into a set or as keys in a map you might run into trouble. I missed this when I looked at the class but your test here points it out!

Do you really need to preserve the order of the ref names in the PSTaxInfo object? You could get rid of this problem if you put them into a set. If you really need them to preserve the order but want equals not to care about the order, you could sort them or put them into a set when you compute the hash code, although that would be inefficient.

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTaxInfoTest.java
```diff
@@ -0,0 +1,64 @@
...
+        final PSTaxInfo infoA = new PSTaxInfo();
+        infoA.parent_tax = "1";
+        infoA.length = 1000;
+        infoA.name = "tax A";
+        infoA.rank = "species";
+        infoA.ref_names.add("ref|NC_A1");
+        infoA.ref_names.add("ref|NC_A2");
+
+        final String strA = infoA.toString();
+        Assert.assertNotNull(strA);
```
cwhelan: Why not just test the complete output of `toString` here?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSHitInfo.java
```diff
@@ -0,0 +1,17 @@
+package org.broadinstitute.hellbender.tools.spark.pathseq;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+/**
+ * Stores taxonomic IDs that were hits of a read pair. ID's can be repeated.
+ */
+public final class PSHitInfo {
+    public final ArrayList<String> taxIDs;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSScoreInfo.java
```diff
@@ -0,0 +1,19 @@
+package org.broadinstitute.hellbender.tools.spark.pathseq;
+
+/**
+ * Scores assigned to a taxonomic node, used in ClassifyReads
```
mwalker174: Added some clarification to what this class is for. I have added a comment referencing the ClassifyReads tool documentation (upcoming PR), since the specifics of the scoring are actually calculated there.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSScoreInfo.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public final class PSScoreInfo {
+    public double score = 0; //PathSeq score
+    public double score_normalized = 0; //Score normalized to percent of pathogen reads
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSScoreInfo.java
```diff
@@ -0,0 +1,19 @@
...
+    public int reads = 0; //Number of total reads mapped
+    public int unambiguous = 0; //Number of reads mapped unamibuously to this node
+    public long ref_length = 0; //Length of reference in bp
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTaxInfo.java
```diff
@@ -0,0 +1,52 @@
+package org.broadinstitute.hellbender.tools.spark.pathseq;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+/**
+ * Helper class for ClassifyReads that stores taxon data
+ */
+public final class PSTaxInfo {
```
mwalker174: Came up with PSPathogenReferenceTaxonProperties and added some explanation to the class header comment.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTaxInfo.java
```diff
@@ -0,0 +1,52 @@
...
+    public String name = null;
+    public String rank = null;
+    public String parent_tax = null;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTaxInfo.java
```diff
@@ -0,0 +1,52 @@
...
+    public long length = 0;
+    public final Collection<String> ref_names;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
```diff
@@ -0,0 +1,322 @@
...
+ */
+@DefaultSerializer(PSTree.Serializer.class)
+public class PSTree {
+
+    private final String root;
+    private Map<String, PSTreeNode> tree;
+
+    public PSTree(final String root_id) {
+        tree = new HashMap<>();
+        root = root_id;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
```diff
@@ -0,0 +1,322 @@
...
+
+        kryo.setReferences(oldReferences);
+    }
+
+    /**
+     * Returns short String representation of all nodes
+     */
+    private static String getAbbreviatedNodeListString(final Set<String> nodes) {
+        String nodesList = "[";
+        int count = 0;
```
mwalker174: I like the streaming style. Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
```diff
@@ -0,0 +1,322 @@
...
+     */
+    public void addNode(final String id, final String name, final String parent, final long length, final String rank) {
+        Utils.validateArg((id != null) && (name != null) && (parent != null) && (rank != null), "Passed a null argument to addNode()");
+        Utils.validateArg(!root.equals(id), "Tried to set root attributes using addNode()");
+
+        //If the node already exists, keep its current children and set everything else
+        if (!tree.containsKey(id)) {
+            tree.put(id, new PSTreeNode());
+        }
+        tree.get(id).setName(name);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
```diff
@@ -0,0 +1,322 @@
...
+                    throw new UserException.BadInput("Malformed tree detected. Node " + id + " has non-existent parent " + parent);
+                }
+                if (!tree.get(parent).getChildren().contains(id)) {
+                    throw new UserException.BadInput("Malformed tree detected. Node " + id + " has parent " + parent + ", which does not have child " + id);
+                }
+            }
+        }
+
+        //Check disconnected sets of nodes using a traversal from the root
+        final Set<String> reachable = traverse();
```
mwalker174: Yes - done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
mwalker174: Yes, for some reason I thought there was a cycle check in checkStructure() but apparently not. Added the exception here, where it will now be thrown during checkStructure() as well.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTree.java
mwalker174: Added the check

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTaxInfoTest.java
mwalker174: Actually I don't think equals and hashcode are used for anything (must have changed something that required them originally). Removing the functions and tests. The class will be covered by tests in a future PR.

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTaxInfoTest.java
mwalker174: Test removed, see above.

PR #2664, opened 2017-05-04
PathSeq common utilities class

mwalker174:

Miscellaneous utility functions used by two or more PathSeq tools.

Note pathseqGetRecommendedNumReducers is needed because some tools read 2 BAMs - one with paired reads one with unpaired reads - and the GATKTool's version will only let you call it on the BAM specified by --input.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSPairedUnpairedSplitterSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSPairedUnpairedSplitterSpark.java
@@ -1,0 +1,87 @@
+package org.broadinstitute.hellbender.tools.spark.pathseq;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.spark.HashPartitioner;
+import org.apache.spark.api.java.JavaRDD;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import scala.Tuple2;
+
+import java.util.*;
+
+/**
+ * Class for separating paired and unpaired reads in an RDD
+ */
+public final class PSPairedUnpairedSplitterSpark {
+
+    protected final Logger logger = LogManager.getLogger(this.getClass());
+    private final JavaRDD<Tuple2<List<GATKRead>, List<GATKRead>>> repartitionedReads;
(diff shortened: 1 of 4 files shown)


tedsharpe:
Let's talk about groupReadPairs, pairedReads, and unpairedReads.

The groupBy method called by groupReadPairs is very expensive both in time and in memory.  It's a full hash shuffle of GATKReads (time expensive), that results in a gazillion 1- and 2-element Lists (memory expensive).  So you certainly don't want to do it twice.  But the way pairedReads and unpairedReads is set up, you *will* do it twice if you want to process both paired and unpaired reads.  (And even if you aren't, someone else might try to use this code to do so.)

So my first suggestion is that you remove the call to groupReadPairs from pairedReads and unpairedReads, and let a user groupReadPairs once, and reuse the resulting JavaPairRDD to process paired and unpaired reads.

My second suggestion is quite a bit more complicated, but I think it would result in far better performance.  I'll sketch it out here, and then I can explain it further in person, if it's a direction you'd like to pursue.  The first step is to create a JavaRDD<GATKRead> in which all pairs sharing a template name are in the same partition (but without grouping them).  To do that, you temporarily boost the input JavaRDD into a JavaPairRDD<String,GATKRead> by extracting the read name as a key.  Then you repartition (to do the shuffle).  Then you map back to an ordinary JavaRDD<GATKRead> by keeping just the value.  (Note:  if the BAM has queryname sort order, you can just skip this step entirely.)
Now you can do a mapPartition operation to filter for paired or unpaired reads:  Iterate over the reads in the partition, and keep a hash map of [name -> read] of reads that have not yet found mates.  To filter for paired reads, whenever you find the name of the current read already in the table, just emit the current read and the read in the map as a pair, and delete the read from the map (you're done with that name -- this keeps the table smaller).  To filter for unpaired reads, just delete any map entry that you successfully look up, and insert any name that you don't.  What's left at the end of this process are all the unpaired reads.

mwalker174:
The functions for getting paired and unpaired reads now go with the second suggestion to partition the reads by read names (rather than use groupBy), then map the entire partitions to a list of paired reads and a list of unpaired reads. 

Lots of angle brackets but I think this approach minimizes the number of lines while yielding substantial efficiency gains over the original approach, mainly by avoiding a needless second shuffle.

PR #2665, opened 2017-05-04
Update PathSeq read filters and DUST transformer

mwalker174:
@tedsharpe @cwhelan please review.

- Adds MarkDuplicatesReadFilter (to replace MarkedOpticalDuplicateReadFilter). MarkedOpticalDuplicateReadFilter will be removed in a subsequent PR because the Filter tool currently uses it.
- Changed some types (short to int, float to double) in the DUST algorithm
- Adds HostAlignmentReadFilter for filtering sufficiently well-mapped host reads. The helper function is there to run the test on supplementary alignments. I chose not to expose this as a GATK filter because the definitions of coverage and identity used here could be different than what some users would expect.

@lbergelson Addressed your comments from the other branch:
- Added docstring to AmbiguousBaseReadFilter argument
- Made filterOpticalOnly an argument
- Argument variables changed from uppercase to lowercase
- See above regarding the duplicates filters


src/main/java/org/broadinstitute/hellbender/engine/filters/MarkDuplicatesReadFilter.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public MarkDuplicatesReadFilter() { }
+
+    public MarkDuplicatesReadFilter(final boolean filterOpticalOnly) {
+        this.filterOpticalOnly = filterOpticalOnly;
+    }
+
+    @Override
+    public boolean test(final GATKRead read) {
+        return !((!filterOpticalOnly && read.isDuplicate()) || (read.hasAttribute(MarkDuplicatesSparkUtils.OPTICAL_DUPLICATE_TOTAL_ATTRIBUTE_NAME)
```
tedsharpe: You're killin' me here.  How about something like:
```
if ( !read.isDuplicate() ) return true;
if ( !filterOpticalOnly ) return false;
final Integer attrVal = read.getAttributeAsInteger(MarkDuplicatesSparkUtils.OPTICAL_DUPLICATE_TOTAL_ATTRIBUTE_NAME);
return attrVal == null || attrVal.intValue() == 0;
```

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/HostAlignmentReadFilter.java
```diff
@@ -0,0 +1,49 @@
...
+    private final int minCoverage, minIdentity;
+
+    public HostAlignmentReadFilter(final int minCov, final int minIdent) {
+        this.minCoverage = minCov;
+        this.minIdentity = minIdent;
+    }
+
+    @Override
+    public boolean test(final GATKRead read) {
+        return read.isUnmapped() || !read.hasAttribute("NM")
```
tedsharpe: I got the logic of this one on first reading!  So that's good.  But if you break it up, you can avoid the double attribute lookup.
```
if ( read.isUnmapped() ) return true;
final Integer nmVal = read.getAttributeAsInteger("NM");
if ( nmVal == null ) return true;
return test(read.getCigar(), nmVal.intValue());
```
Up to you.  This is somewhat a matter of taste.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/HostAlignmentReadFilter.java
```diff
@@ -0,0 +1,49 @@
...
+                || test(read.getCigar(), read.getAttributeAsInteger("NM"));
+    }
+
+    public boolean test(final Cigar c, final int NM) {
```
tedsharpe: Louis is going to scold you for a variable called NM (upper case).  nMismatches?

src/main/java/org/broadinstitute/hellbender/engine/filters/MarkDuplicatesReadFilter.java
```diff
@@ -0,0 +1,32 @@
...
+import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.hellbender.tools.spark.transforms.markduplicates.MarkDuplicatesSparkUtils;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+import java.io.Serializable;
+
+/**
+ * Filters out reads marked as duplicates.
+ */
+public final class MarkDuplicatesReadFilter extends ReadFilter implements Serializable {
```
cwhelan: It seems confusing to me to have `MarkDuplicatesReadFilter` here and `ReadFilterLibrary.NotDuplicateReadFilter` in the same package. It might be more consistent to call this `NotOpticalDuplicateReadFilter` and chain it with `NotDuplicateReadFilter` if desired. Did @lbergelson already comment on this?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/HostAlignmentReadFilter.java
cwhelan: I also think single-variable names are frowned upon. `cigar`?

src/main/java/org/broadinstitute/hellbender/engine/filters/MarkDuplicatesReadFilter.java
mwalker174: Wasn't aware of NotDuplicateReadFilter. Changed the class to NotOpticalDuplicateReadFilter.

src/main/java/org/broadinstitute/hellbender/engine/filters/MarkDuplicatesReadFilter.java
mwalker174: Changed the class to NotOpticalDuplicatesReadFilter (see below), but point taken.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/HostAlignmentReadFilter.java
mwalker174: Didn't realize getAttributeAsInteger had such a long code path. Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/HostAlignmentReadFilter.java
mwalker174: Done

PR #2669, opened 2017-05-05
prevent GenotypeGVCFs crash due to RMSMappingQuality #2658

### LINKED ISSUE #2658: RMSMappingQuality explodes when there are 0 reads
@kcibul reports getting the following error while running `GenotypeGVCFs`:

```
New error coming out of GenotypeGVCFs when running 1000 shards:

[11:20] 
***********************************************************************

A USER ERROR has occurred: Bad input: Cannot calculate Root Mean Square Mapping Quality if there are 0 or less reads.
Number of reads recorded as :0
In VariantContext: [VC UG_call @ chr1:143249103 Q38.86 of type=SNP alleles=[C*, T] attr={DP=14, ExcessHet=3.0102999210357666, MLEAC=[2], MLEAF=[0.25], RAW_MQ=0.0} GT=[[09C81377 C*/C* GQ 6 DP 5 PL 0,6,90 {MIN_DP=4}],[09C83237 C*/C* GQ 3 DP 6 PL 0,3,45 {MIN_DP=6}],[09C97255 ./. DP 0 PL 0,0,0 {MIN_DP=0}],[09C98651 ./. DP 0 PL 0,0,0 {MIN_DP=0}],[09C99383 T/T GQ 3 PL 45,3,0 {PGT=0|1, PID=143249097_C_T, SB=[0, 0, 0, 0]}],[09C99677 ./. DP 0 PL 0,0,0 {MIN_DP=0}],[10C100868 C*/C* GQ 3 DP 7 PL 0,3,45 {MIN_DP=4}],[10C101312 ./. DP 0 PL 0,0,0 {MIN_DP=0}],[10C102545 ./. DP 1 PL 0,0,0 {MIN_DP=0}],[10C102782 ./. DP 0 PL 0,0,0 {MIN_DP=0}]]

[11:21] 
```

I think this might be a bug, or at least a case of over-aggressive error checking. We have an extra check in our version of `RMSMappingQuality` that is not present in the GATK3 version:

```
        if (numOfReads <= 0){
            throw new UserException.BadInput("Cannot calculate Root Mean Square Mapping Quality if there are 0 or less reads." +
                                                     "\nNumber of reads recorded as :" +numOfReads +
                                                     "\nIn VariantContext: "+ vc.toStringDecodeGenotypes());
        }
```

We should match the GATK3 behavior in this case, unless we can **prove** (and not merely infer) that GATK3 also explodes.

@lbergelson Since you are git blamed here, can you give your thoughts on this?

lbergelson:
Removing the error check in RMSMappingQuality.getNumReads which was causing a crash when depth was being reported as 0 or less.
It will now return -1 which should match gatk3 behavior.
Fixes #2658

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
@@ -158,9 +158,7 @@ public final class RMSMappingQuality extends InfoFieldAnnotation implements Stan
             }
         }
         if (numOfReads <= 0){
-            throw new UserException.BadInput("Cannot calculate Root Mean Square Mapping Quality if there are 0 or less reads." +
-                                                     "\nNumber of reads recorded as :" +numOfReads +
-                                                     "\nIn VariantContext: "+ vc.toStringDecodeGenotypes());
+            numOfReads = -1;  //return -1 to result in a NaN
         }
         return numOfReads;
     }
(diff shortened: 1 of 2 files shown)


droazen:
:+1: looks good to me -- merging

PR #2671, opened 2017-05-05
Make it possible for individual tools to override the default NIO buffer sizes

### LINKED ISSUE #2670: Make it possible for tools to override the default NIO buffer sizes

droazen:
GenomicsDBImport now specifies default NIO buffer sizes of 2 MB, rather than the
default 40 MB. The smaller size is due to the large number of inputs this tool
typically has.

Resolves #2670


droazen:
@lbergelson please review

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -241,6 +241,25 @@ public ReadTransformer makePostReadFilterTransformer(){
     }
 
     /**
+     * @return Default size in MB of the cloud prefetch buffer. May be overridden by individual tools.
+     *         The default implementation returns a value (40 MB) that is suitable for tools with a small
+     *         number of large cloud inputs. Tools with large numbers of cloud inputs will likely want to
+     *         override to specify a smaller size.
+     */
+    public int getDefaultCloudPrefetchBufferSize() {
```
lbergelson: Shouldn't these be protected since they're really internal methods designed for overriding?

lbergelson: requiresReference is also public so I guess these match class convention

PR #2672, opened 2017-05-05
Rename IOUtils.canRead() to IOUtils.canReadFile(), and move tests from UtilsUnitTest to IOUtilsUnitTest

droazen:


droazen:
@davidbenjamin please review. Since it was my miscommunication on the previous PR that led to this issue, thought I'd rectify things ;)

davidbenjamin:
@droazen If you can believe it I believe we had exactly the same idea at exactly the same time, though I think I am by far the guiltier one.  I will review this and close the other PR.

PR #2675, opened 2017-05-05
Allowing GenomicsDBImport to skip header merging

lbergelson:
Adding an advanced option --sampleNameMap to GenomicsDBImport which can be specified instead of -V
* This allows providing a file mapping samples to their vcf uri which is used instead of loading each header and checking the sample names.
* The first file in the mapping is chosen to act as the header template.

If the headers are not consistent there may be data corruption.

A mapping file is a tab seperated file with no header in the format

Sample1	file1
Sample2 file2
Sample3 file3
...

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -454,7 +454,7 @@ public final class GenomicsDBImport extends GATKTool {
         final String arrayName,
         final long variantContextBufferSize,
         final long segmentSize,
-        final  long lbSampleIndex,
+        final long lbSampleIndex,
         final long ubSampleIndex) {
 
         final GenomicsDBImportConfiguration.Partition.Builder pBuilder =
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -213,8 +213,9 @@ public final class GenomicsDBImport extends GATKTool {
     }
 
     private void initializeHeaderAndSampleMappings() {
-        //Only one of
+        // Only one of -V and -sampleNameToPathFile may be specified
         if (sampleNameToPathFile == null) {
+            // -V was specified
             final List<VCFHeader> headers = new ArrayList<>(variantPaths.size());
 
             for (final String variantPath : variantPaths) {
@@ -235,6 +236,7 @@ public final class GenomicsDBImport extends GATKTool {
             mergedHeaderLines = VCFUtils.smartMergeHeaders(headers, true);
             mergedHeaderSequenceDictionary = new VCFHeader(mergedHeaderLines).getSequenceDictionary();
         } else {
+            // -sampleNameMap was specified
             sampleNameToVcfUri.putAll(loadSampleNameMapFile(IOUtils.getPath(sampleNameToPathFile)));
             sampleNames.addAll(sampleNameToVcfUri.keySet());
 
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -24,6 +24,7 @@ import org.broadinstitute.hellbender.engine.GATKTool;
 import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.io.IOUtils;
 import org.broadinstitute.hellbender.utils.nio.SeekableByteChannelPrefetcher;
 
@@ -35,7 +36,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.*;
 import java.util.function.Function;
-import java.util.stream.Collectors;
 
 
 /**
@@ -95,9 +95,11 @@ public final class GenomicsDBImport extends GATKTool {
               shortName = VCF_BUFFER_SIZE_ARG_NAME,
               doc = "Buffer size in bytes to store variant contexts." +
                     " Larger values are better as smaller values cause frequent disk writes." +
-                    " Defaults to " + DEFAULT_VCF_BUFFER_SIZE_PER_SAMPLE,
+                    " Defaults to " + DEFAULT_VCF_BUFFER_SIZE_PER_SAMPLE + " which was empirically determined to work" +
+                    " well for many inputs.",
               optional = true,
-              minValue = 1024L)
+              minValue = 1024L,
+              minRecommendedValue = 10 * 1024)
     private long vcfBufferSizePerSample = DEFAULT_VCF_BUFFER_SIZE_PER_SAMPLE;
 
     @Argument(fullName = OVERWRITE_WORKSPACE_NAME,
@@ -133,14 +135,14 @@ public final class GenomicsDBImport extends GATKTool {
     @Advanced
     @Argument(fullName = SAMPLE_NAME_MAP_LONG_NAME,
             shortName = SAMPLE_NAME_MAP_LONG_NAME,
-            doc = "Path to file containing a mapping of sample name to file name in tab delimited format.  If this is " +
-                    "specified then the header from the first sample will be used as the header template rather than " +
+            doc = "Path to file containing a mapping of sample name to file uri in tab delimited format.  If this is " +
+                    "specified then the header from the first sample will be treated as the merged header rather than " +
                     "merging the headers, and the sample names will taken from this file.  This is a performance optimization " +
                     "that relaxes the normal checks for consistent headers.  Using vcfs with incompatible headers may result " +
                     "in silent data corruption.",
             optional = true,
             mutex = {StandardArgumentDefinitions.VARIANT_LONG_NAME})
-    private String sampleNameToPathFile;
+    private String sampleNameMapFile;
 
     @Override
     public boolean requiresIntervals() { return true; }
@@ -170,16 +172,13 @@ public final class GenomicsDBImport extends GATKTool {
     // Intervals from command line (singleton for now)
     private List<ChromosomeInterval> intervals;
 
-    // List of all sample names
-    private final Set<String> sampleNames = new LinkedHashSet<>();
-
     // Linked hash map between sample names and corresponding GVCF file name
-    private final Map<String, String> sampleNameToVcfUri = new LinkedHashMap<>();
+    private LinkedHashMap<String, String> sampleNameToVcfUri = new LinkedHashMap<>();
 
     // Needed as smartMergeHeaders() returns a set of VCF header lines
     private Set<VCFHeaderLine> mergedHeaderLines = null;
 
-    // VCFHeader created from the merged header
+    // sequence dictionary created from the merged header
     private SAMSequenceDictionary mergedHeaderSequenceDictionary;
 
     // Path to vidmap file to be written by GenomicsDBImporter
@@ -206,27 +205,27 @@ public final class GenomicsDBImport extends GATKTool {
     }
 
     private void assertVariantPathsOrSampleNameFileWasSpecified(){
-        if ( (variantPaths == null || variantPaths.isEmpty()) && sampleNameToPathFile == null) {
+        if ( (variantPaths == null || variantPaths.isEmpty()) && sampleNameMapFile == null) {
             throw new CommandLineException.MissingArgument(StandardArgumentDefinitions.VARIANT_LONG_NAME,
-                                                       "One of " + StandardArgumentDefinitions.VARIANT_LONG_NAME + " or " + SAMPLE_NAME_MAP_LONG_NAME + " must be specified" );
+                                                       "One of --" + StandardArgumentDefinitions.VARIANT_LONG_NAME + " or --" + SAMPLE_NAME_MAP_LONG_NAME + " must be specified" );
         }
     }
 
+    /**
+     * sets the values of mergedHeaderLines, mergedHeaderSequenceDictionary, and sampleNameToVcfUri
+     */
     private void initializeHeaderAndSampleMappings() {
-        // Only one of -V and -sampleNameToPathFile may be specified
-        if (sampleNameToPathFile == null) {
+        // Only one of -V and --sampleNameMapFile may be specified
+        if (sampleNameMapFile == null) {
             // -V was specified
             final List<VCFHeader> headers = new ArrayList<>(variantPaths.size());
-
             for (final String variantPath : variantPaths) {
                 final  VCFHeader header = loadHeaderFromVCFUri(variantPath);
-
-                // A GVCF file must contain only one sample, throw an exception otherwise
+                Utils.validate(header != null, "Null header was found in " + variantPath + ".");
                 assertGVCFHasOnlyOneSample(variantPath, header);
                 headers.add(header);
 
                 final String sampleName = header.getGenotypeSamples().get(0);
-                sampleNames.add(sampleName);
                 final String previousPath = sampleNameToVcfUri.put(sampleName, variantPath);
                 if (previousPath != null) {
                     throw new UserException("Duplicate sample: " + sampleName + ". Sample was found in both "
@@ -236,14 +235,16 @@ public final class GenomicsDBImport extends GATKTool {
             mergedHeaderLines = VCFUtils.smartMergeHeaders(headers, true);
             mergedHeaderSequenceDictionary = new VCFHeader(mergedHeaderLines).getSequenceDictionary();
         } else {
-            // -sampleNameMap was specified
-            sampleNameToVcfUri.putAll(loadSampleNameMapFile(IOUtils.getPath(sampleNameToPathFile)));
-            sampleNames.addAll(sampleNameToVcfUri.keySet());
-
+            // --sampleNameMap was specified
+            sampleNameToVcfUri = loadSampleNameMapFile(IOUtils.getPath(sampleNameMapFile));
             final String firstHeaderPath = sampleNameToVcfUri.entrySet().iterator().next().getValue();
             final VCFHeader header = loadHeaderFromVCFUri(firstHeaderPath);
             mergedHeaderLines = header.getMetaDataInInputOrder();
             mergedHeaderSequenceDictionary = header.getSequenceDictionary();
+        }
+
+        if ( mergedHeaderSequenceDictionary == null) {
+            throw new UserException("The merged vcf header has no sequence dictionary. Please provide a header that contains a sequence dictionary.");
         }
     }
 
@@ -255,34 +256,38 @@ public final class GenomicsDBImport extends GATKTool {
         }
     }
 
-    private static void assertGVCFHasOnlyOneSample(String variantPath, VCFHeader header) {
+    private static void assertGVCFHasOnlyOneSample(final String variantPath, final VCFHeader header) {
         // A GVCF file must contain only one sample, throw an exception otherwise
         final int numberOfSamples = header.getNGenotypeSamples();
         if (numberOfSamples != 1) {
-            throw new UserException("Input GVCF: " + variantPath + " was expected to contain a single sample but actually contained " + numberOfSamples + ".");
+            throw new UserException("Input GVCF: " + variantPath + " was expected to contain a single sample but actually contained " + numberOfSamples + " samples.");
         }
     }
 
     @VisibleForTesting
-    static Map<String, String> loadSampleNameMapFile(final Path sampleToFileMapPath) {
+    static LinkedHashMap<String, String> loadSampleNameMapFile(final Path sampleToFileMapPath) {
         try {
-            final Map<String, String> sampleToFilename = Files.lines(sampleToFileMapPath)
-                    .map(line -> {
-                        final String[] split = line.split("\\s");
-                        if (split.length != 2) {
-                            throw new UserException.BadInput(
-                                    "Expected a file of format\nSample\tFile\n but found line: " + line);
-                        }
-                        return split;
-                    })
-                    .collect(Collectors.toMap(split -> split[0],
-                                             split -> split[1],
-                                              (a, b) -> { throw new UserException.BadInput("Found two mappings for the same sample:\n" + a + "\n" + b); },
-                                              HashMap::new));
+            final List<String> lines = Files.readAllLines(sampleToFileMapPath);
+            if (lines.isEmpty()) {
+                throw new UserException.BadInput( "At least 1 sample is required but none were found in the sample mapping file");
+            }
 
-            if (sampleToFilename.entrySet().isEmpty()) {
-                throw new UserException.BadInput(
-                        "At least 1 sample is required but none were found in the sample mapping file");
+            final LinkedHashMap<String, String> sampleToFilename = new LinkedHashMap<>(lines.size());
+            for ( final String line : lines) {
+                final String[] split = line.split("\\t",-1);
+                if (split.length != 2
+                        || split[0].isEmpty() || containsWhitespace(split[0])
+                        || split[1].isEmpty() || containsWhitespace(split[1])) {
+                    throw new UserException.BadInput(
+                            "Expected a file of format\nSample\tFile\n but found line: " + line);
+                }
+
+                final String sample = split[0];
+                final String path = split[1];
+                final String oldPath = sampleToFilename.put(sample, path);
+                if (oldPath != null){
+                    throw new UserException.BadInput("Found two mappings for the same sample: " + sample + "\n" + path + "\n" + oldPath );
+                }
             }
             return sampleToFilename;
         } catch (final IOException e) {
@@ -290,6 +295,14 @@ public final class GenomicsDBImport extends GATKTool {
         }
     }
 
+    private static boolean containsWhitespace(final String s){
+        for ( char c : s.toCharArray()){
+            if (Character.isWhitespace(c)){
+                return true;
+            }
+        }
+        return false;
+    }
 
     /**
      * Before traversal, fix configuration parameters and initialize
@@ -298,7 +311,7 @@ public final class GenomicsDBImport extends GATKTool {
     @Override
     public void onTraversalStart() {
 
-        File workspaceDir = overwriteOrCreateWorkspace();
+        final File workspaceDir = overwriteOrCreateWorkspace();
 
         vidMapJSONFile = new File(workspaceDir + "/" + GenomicsDBConstants.DEFAULT_VIDMAP_FILE_NAME);
         callsetMapJSONFile = new File (workspaceDir + "/" + GenomicsDBConstants.DEFAULT_CALLSETMAP_FILE_NAME);
@@ -309,7 +322,7 @@ public final class GenomicsDBImport extends GATKTool {
 
         // Passing in false here so that sample names will be sorted.
         // This is needed for consistent ordering across partitions/machines
-        callsetMappingPB = GenomicsDBImporter.generateSortedCallSetMap(new ArrayList<>(sampleNames), false);
+        callsetMappingPB = GenomicsDBImporter.generateSortedCallSetMap(new ArrayList<>(sampleNameToVcfUri.keySet()), false);
     }
 
     /**
@@ -319,8 +332,8 @@ public final class GenomicsDBImport extends GATKTool {
     @Override
     public void traverse() {
 
-        final int updatedBatchSize = (batchSize == DEFAULT_ZERO_BATCH_SIZE) ? sampleNames.size() : batchSize;
-        final int sampleCount = sampleNames.size();
+        final int sampleCount = sampleNameToVcfUri.size();
+        final int updatedBatchSize = (batchSize == DEFAULT_ZERO_BATCH_SIZE) ? sampleCount : batchSize;
         final int totalBatchCount = (sampleCount/updatedBatchSize) + (sampleCount%updatedBatchSize==0 ? 0 : 1);
 
         GenomicsDBImporter importer;
@@ -328,7 +341,7 @@ public final class GenomicsDBImport extends GATKTool {
         for (int i = 0, batchCount = 1; i < sampleCount; i += updatedBatchSize, ++batchCount) {
 
             final Map<String, FeatureReader<VariantContext>> sampleToReaderMap =
-                getFeatureReaders(new ArrayList<>(sampleNames), sampleNameToVcfUri, updatedBatchSize, sampleCount, i);
+                getFeatureReaders(new ArrayList<>(sampleNameToVcfUri.keySet()), sampleNameToVcfUri, updatedBatchSize, sampleCount, i);
 
             logger.info("Importing batch " + batchCount + " with " + sampleToReaderMap.size() + " samples");
             final long variantContextBufferSize = vcfBufferSizePerSample * sampleToReaderMap.size();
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -140,11 +125,23 @@
...
-                    "as the top Java layers has significantly higher overheads. This flag have no effect if only one " +
+                    "as the top Java layers have significantly higher overheads. This flag has no effect if only one " +
                     "batch is used. Defaults to false",
               optional = true)
     private Boolean doConsolidation = false;
 
+    @Advanced
+    @Argument(fullName = SAMPLE_NAME_MAP_LONG_NAME,
+            shortName = SAMPLE_NAME_MAP_LONG_NAME,
+            doc = "Path to file containing a mapping of sample name to file name in tab delimited format.  If this is " +
```
droazen: file name -> file URI?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -140,11 +125,23 @@
...
+                    "specified then the header from the first sample will be used as the header template rather than " +
```
droazen: used as the header template -> treated as the merged header rather than actually merging the headers

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -87,34 +85,21 @@
     @Argument(fullName = StandardArgumentDefinitions.VARIANT_LONG_NAME,
               shortName = StandardArgumentDefinitions.VARIANT_SHORT_NAME,
               doc = "GVCF files to be imported to GenomicsDB. Each file must contain" +
-                    "data for only a single sample")
+                    "data for only a single sample. Either this or " + SAMPLE_NAME_MAP_LONG_NAME +
+                    " must be specified.",
+              optional = true,
```
droazen: We want this argument to be required though, if the sample name map is not provided

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -140,11 +125,23 @@
...
+    @Argument(fullName = SAMPLE_NAME_MAP_LONG_NAME,
+            shortName = SAMPLE_NAME_MAP_LONG_NAME,
+            doc = "Path to file containing a mapping of sample name to file name in tab delimited format.  If this is " +
+                    "specified then the header from the first sample will be used as the header template rather than " +
+                    "merging the headers, and the sample names will taken from this file.  This is a performance optimization " +
+                    "that relaxes the normal checks for consistent headers.  Using vcfs with incompatible headers may result " +
+                    "in silent data corruption.",
+            optional = true,
+            mutex = {StandardArgumentDefinitions.VARIANT_LONG_NAME})
+    private String sampleNameToPathFile;
```
droazen: Rename to `sampleNameMapFile`

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -174,16 +171,16 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
-    private Map<String, String> sampleNameToVcfUri = new LinkedHashMap<>();
+    private final Map<String, String> sampleNameToVcfUri = new LinkedHashMap<>();
 
     // Needed as smartMergeHeaders() returns a set of VCF header lines
-    private Set<VCFHeaderLine> mergedHeader = null;
+    private Set<VCFHeaderLine> mergedHeaderLines = null;
 
     // VCFHeader created from the merged header
-    private VCFHeader mergedVCFHeader;
+    private SAMSequenceDictionary mergedHeaderSequenceDictionary;
```
droazen: You forgot to update the comment for this field

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
-        List<VCFHeader> headers = new ArrayList<>(variantPaths.size());
-
-        for (String variantPath : variantPaths) {
-
-            final AbstractFeatureReader<VariantContext, LineIterator> reader = getReaderFromVCFUri(variantPath);
-            VCFHeader header = (VCFHeader)reader.getHeader();
+    private void assertVariantPathsOrSampleNameFileWasSpecified(){
+        if ( (variantPaths == null || variantPaths.isEmpty()) && sampleNameToPathFile == null) {
+            throw new CommandLineException.MissingArgument(StandardArgumentDefinitions.VARIANT_LONG_NAME,
+                                                       "One of " + StandardArgumentDefinitions.VARIANT_LONG_NAME + " or " + SAMPLE_NAME_MAP_LONG_NAME + " must be specified" );
```
droazen: Insert "--" before arg names in error message for clarity

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
+        super.onStartup();
+    }
 
-        List<VCFHeader> headers = new ArrayList<>(variantPaths.size());
-
-        for (String variantPath : variantPaths) {
-
-            final AbstractFeatureReader<VariantContext, LineIterator> reader = getReaderFromVCFUri(variantPath);
-            VCFHeader header = (VCFHeader)reader.getHeader();
+    private void assertVariantPathsOrSampleNameFileWasSpecified(){
```
droazen: It is unfortunate that our arg parsing system can't do this check for us.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
-            if (sampleNameToVcfUri.put(sampleName, variantPath) != null) {
-                throw new UserException("Duplicate sample " + sampleName);
+    private void initializeHeaderAndSampleMappings() {
+        // Only one of -V and -sampleNameToPathFile may be specified
+        if (sampleNameToPathFile == null) {
+            // -V was specified
+            final List<VCFHeader> headers = new ArrayList<>(variantPaths.size());
+
+            for (final String variantPath : variantPaths) {
+                final  VCFHeader header = loadHeaderFromVCFUri(variantPath);
```
droazen: Check that the header is non-null, throw if it is.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: it is unfortunate.  I filed a bug https://github.com/broadinstitute/barclay/issues/52

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
         }
+    }
 
-        mergedHeader = VCFUtils.smartMergeHeaders(headers, true);
-        mergedVCFHeader = new VCFHeader(mergedHeader);
+    private static void assertGVCFHasOnlyOneSample(String variantPath, VCFHeader header) {
+        // A GVCF file must contain only one sample, throw an exception otherwise
+        final int numberOfSamples = header.getNGenotypeSamples();
+        if (numberOfSamples != 1) {
+            throw new UserException("Input GVCF: " + variantPath + " was expected to contain a single sample but actually contained " + numberOfSamples + ".");
```
droazen: add "samples" at the end of this message

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
+                }
             }
+            mergedHeaderLines = VCFUtils.smartMergeHeaders(headers, true);
+            mergedHeaderSequenceDictionary = new VCFHeader(mergedHeaderLines).getSequenceDictionary();
+        } else {
+            // -sampleNameMap was specified
+            sampleNameToVcfUri.putAll(loadSampleNameMapFile(IOUtils.getPath(sampleNameToPathFile)));
+            sampleNames.addAll(sampleNameToVcfUri.keySet());
+
+            final String firstHeaderPath = sampleNameToVcfUri.entrySet().iterator().next().getValue();
```
droazen: Is it guaranteed that the first line in the sample name map file corresponds to the first entry in `sampleNameToVcfUri`? Do you have an explicit unit test for this?

droazen: I'd also add an explicit check here that the first entry matches the physical line in the map file, and throw a `GATKException` if it doesn't.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -174,16 +171,16 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
     private List<ChromosomeInterval> intervals;
 
     // List of all sample names
-    private List<String> sampleNames = new ArrayList<>();
+    private final Set<String> sampleNames = new LinkedHashSet<>();
```
droazen: Maybe we should just use `keySet()` from `sampleNameToVcfUri` instead of maintaining this `Set` separately

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
+    }
 
-        initializeIntervals();
-        super.onStartup();
+    @VisibleForTesting
+    static Map<String, String> loadSampleNameMapFile(final Path sampleToFileMapPath) {
+        try {
+            final Map<String, String> sampleToFilename = Files.lines(sampleToFileMapPath)
+                    .map(line -> {
+                        final String[] split = line.split("\\s");
```
droazen: Is it useful to require a single whitespace char of any type instead of a single tab? Also might have issues with newlines. Probably safest to just require a tab.

droazen: Also I think you might want to pass in -1 as a second arg to split, to catch stray tabs at the end of a line

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
+                        if (split.length != 2) {
+                            throw new UserException.BadInput(
+                                    "Expected a file of format\nSample\tFile\n but found line: " + line);
+                        }
+                        return split;
+                    })
+                    .collect(Collectors.toMap(split -> split[0],
+                                             split -> split[1],
+                                              (a, b) -> { throw new UserException.BadInput("Found two mappings for the same sample:\n" + a + "\n" + b); },
+                                              HashMap::new));
```
droazen: I assume this bug is fixed :)

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
+
+            if (sampleToFilename.entrySet().isEmpty()) {
```
droazen: Can't you just write `sampleToFilename.isEmpty()` here?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
 
-        initializeIntervals();
-        super.onStartup();
+    @VisibleForTesting
+    static Map<String, String> loadSampleNameMapFile(final Path sampleToFileMapPath) {
+        try {
+            final Map<String, String> sampleToFilename = Files.lines(sampleToFileMapPath)
+                    .map(line -> {
+                        final String[] split = line.split("\\s");
+                        if (split.length != 2) {
```
droazen: Also check that both tokens are non-empty, non-whitespace

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -247,29 +300,16 @@ public void onTraversalStart() {
 
         File workspaceDir = overwriteOrCreateWorkspace();
 
-        // If provided buffer size is less 1KB it is considered to be too small
-        // Either use default 16KB or any value larger than 10KB
```
droazen: Preserve this recommendation on the right value for `vcfBufferSizePerSample` in the docs for the argument above.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -355,8 +395,8 @@ public Object onTraversalSuccess() {
      * @param lowerSampleIndex  0-based Lower bound of sample index -- inclusive
      * @return  Feature readers to be imported in the current batch
      */
-    private Map<String, FeatureReader<VariantContext>> getFeatureReaders(List<String> sampleNames, Map<String, String> sampleNameToFileName, int batchSize, int sampleCount, int lowerSampleIndex) {
-        Map<String, FeatureReader<VariantContext>> sampleToReaderMap = new LinkedHashMap<>();
+    private Map<String, FeatureReader<VariantContext>> getFeatureReaders(final List<String> sampleNames, final Map<String, String> sampleNameToFileName, final int batchSize, int sampleCount, int lowerSampleIndex) {
```
droazen: Why are some of the int args not final?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -395,12 +435,12 @@ public Object onTraversalSuccess() {
...
      * @return  GenomicsDB import configuration object
      */
-    private GenomicsDBImportConfiguration.ImportConfiguration createImportConfiguration(
+    private static GenomicsDBImportConfiguration.ImportConfiguration createImportConfiguration(
         final String workspace,
         final String arrayName,
         final long variantContextBufferSize,
         final long segmentSize,
-        final long lbSampleIndex,
+        final  long lbSampleIndex,
```
droazen: Extra whitespace here

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
+            mergedHeaderSequenceDictionary = new VCFHeader(mergedHeaderLines).getSequenceDictionary();
+        } else {
+            // -sampleNameMap was specified
+            sampleNameToVcfUri.putAll(loadSampleNameMapFile(IOUtils.getPath(sampleNameToPathFile)));
+            sampleNames.addAll(sampleNameToVcfUri.keySet());
+
+            final String firstHeaderPath = sampleNameToVcfUri.entrySet().iterator().next().getValue();
+            final VCFHeader header = loadHeaderFromVCFUri(firstHeaderPath);
+            mergedHeaderLines = header.getMetaDataInInputOrder();
+            mergedHeaderSequenceDictionary = header.getSequenceDictionary();
```
droazen: Should we throw if this header has no sequence dictionary? (which IS a possibility!)

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -202,42 +199,98 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
...
+                final String sampleName = header.getGenotypeSamples().get(0);
+                sampleNames.add(sampleName);
+                final String previousPath = sampleNameToVcfUri.put(sampleName, variantPath);
+                if (previousPath != null) {
+                    throw new UserException("Duplicate sample: " + sampleName + ". Sample was found in both "
+                                                    + variantPath + " and " + previousPath + ".");
+                }
             }
+            mergedHeaderLines = VCFUtils.smartMergeHeaders(headers, true);
+            mergedHeaderSequenceDictionary = new VCFHeader(mergedHeaderLines).getSequenceDictionary();
```
droazen: Should we throw if the merged header has no sequence dictionary (theoretically possible)

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -29,6 +29,21 @@
...
+    private static final File TEST_REFERENCE_GENOME = new File(largeFileTestDir + "/Homo_sapiens_assembly38.20.21.fasta");
+
+    private static final String HG_00096 = largeFileTestDir + "gvcfs/HG00096.g.vcf.gz";
+    private static final String HG_00268 = largeFileTestDir + "gvcfs/HG00268.g.vcf.gz";
+    private static final String NA_19625 = largeFileTestDir + "gvcfs/NA19625.g.vcf.gz";
+    private static final List<String> LOCAL_GVCFS = Arrays.asList(HG_00096, HG_00268, NA_19625);
+
+    private static final String HG_00096_CLOUD = getGCPTestInputPath() + "large/gvcfs/HG00096.g.vcf.gz";
+    private static final String HG_00268_CLOUD = getGCPTestInputPath() + "large/gvcfs/HG00268.g.vcf.gz";
+    private static final String NA_19625_CLOUD = getGCPTestInputPath() + "large/gvcfs/NA19625.g.vcf.gz";
```
droazen: The tests were structured the way they were before precisely in order to avoid calling `getGCPTestInputPath()` outside of an `@Test` method (otherwise TestNG gives VERY hard-to-interpret error messages when it fails).

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -46,66 +61,73 @@ public String getTestedClassName() {
     }
 
     @Test
-    public void testGenomicsDBImportFileInputs() throws IOException {
-        final String hg00096 = largeFileTestDir + "gvcfs/HG00096.g.vcf.gz";
-        final String hg00268 = largeFileTestDir + "gvcfs/HG00268.g.vcf.gz";
-        final String na19625 = largeFileTestDir + "gvcfs/NA19625.g.vcf.gz";
-        final String combined = largeFileTestDir + "gvcfs/combined.gatk3.7.g.vcf.gz";
-        final SimpleInterval interval = new SimpleInterval("chr20", 17960187, 17981445);
+    public void testSampleMappingFileInsteadOfVCFs() throws IOException {
```
droazen: Move these tests after the "normal" tests for the tool

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -46,66 +61,73 @@ public String getTestedClassName() {
...
+        final String sampleFileContents = "HG00096\t" + HG_00096 +"\n" +
+                "HG00268\t"+ HG_00268 + "\n" +
+                "NA19625\t"+ NA_19625;
+        final File sampleNameFile = IOUtils.writeTempFile(sampleFileContents, "sampleNameMap", ".tx");
```
droazen: File extension here is weird

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -46,66 +61,73 @@ public String getTestedClassName() {
...
+        runCommandLine(args);
+        checkJSONFilesAreWritten(workspace);
+        checkGenomicsDBAgainstExpected(workspace, INTERVAL, COMBINED);
+    }
+
+    @Test(expectedExceptions = CommandLineException.class)
+    public void testCantSpecifyVCFAndSampleNameFile(){
+        final ArgumentsBuilder args = new ArgumentsBuilder()
+                .addArgument(GenomicsDBImport.SAMPLE_NAME_MAP_LONG_NAME, "someFile")
+                .addArgument(StandardArgumentDefinitions.VARIANT_LONG_NAME, "someotherfile")
```
droazen: Test would be more legit if these files actually existed (to prove that the error comes from specifying both args and not from file not found).

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -46,66 +61,73 @@ public String getTestedClassName() {
...
+    public void testGenomicsDBImportGCSInputsInBatches(final int batchSize) throws IOException {
+        testGenomicsDBImporterWithBatchSize(CLOUD_GVCFS, INTERVAL, COMBINED, batchSize);
     }
 
     /**
-     * Check whether user exception is thrown with vcf
-     * buffer size < 1024 bytes
      *
-     * @throws UserException  Value must be >1024 bytes
+     * @throws CommandLineException.OutOfRangeArgumentValue  Value must be >1024 bytes
```
droazen: > or >=?

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -131,7 +153,7 @@ private void testGenomicsDBImportWithZeroBufferSize(final List<String> vcfInputs
 
     }
 
-    private void writeToGenomicsDB(final List<String> vcfInputs, final SimpleInterval interval, final String workspace, int batchSize, final Boolean useBufferSize, int bufferSizePerSample) {
+    private void writeToGenomicsDB(final List<String> vcfInputs, final SimpleInterval interval, final String workspace, final int batchSize, final Boolean useBufferSize, int bufferSizePerSample) {
```
droazen: The last arg is not final?

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportUnitTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @DataProvider
+    public Object[][] getBadTestFiles(){
+        return new Object[][]{
+                {"Sample1\tsamplePath\n"
+                +"Sample1\tsamplePath"},  //duplicate sample name
+                {""}, // empty file
+                {"Sample1\tSample2\tFile"}, //various wrong numbers of inputs
+                {"Sample1\t"},
+                {"Sample1"}
```
droazen: Also test:

* Empty first token, non-empty second token
* First token only whitespace
* Extra tab at end of otherwise normal line
* Sample and samplePath separated by a newline instead of a tab
* A line with only tabs

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportUnitTest.java
```diff
@@ -0,0 +1,43 @@
...
+        GenomicsDBImport.loadSampleNameMapFile(sampleFile.toPath()  );
+    }
+
+    @Test
+    public void testSampleNameToFileMap(){
+        final File sampleFile = IOUtils.writeTempFile("Sample1\tfile1\nSample2\tfile2", "sampleMapping", ".txt");
+        final Map<String, String> expected = new HashMap<>();
+        expected.put("Sample1", "file1");
+        expected.put("Sample2", "file2");
+        Assert.assertEquals(GenomicsDBImport.loadSampleNameMapFile(sampleFile.toPath()), expected);
```
droazen: Also assert that the first entry in the map returned is for `Sample1`

droazen:
Review complete, back to @lbergelson

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: We have an integration test for that.  Limitation of the arg parser at the moment.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: updated

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: guarded against

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: I've swithed everything to linked maps and sets.  That should fix the issue.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: good idea, done

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: fixed

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: yes, I don't know why I didn't do that before...

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: done, and made it an ugly loop instead of a beautiful collections of millions of lambdas

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: because I had to scroll right to see them..

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: huh, i'm not sure what's going on here but I can't see the extra space in my editor, maybe intellij fixed it for me automagically

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: added a check

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: added a check

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
lbergelson: that's a very good point.  I will revert that and add a comment.

lbergelson: I changed it to build these paths off of the local list at each test.  It's concise and can't fail statically.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
lbergelson: moved

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
lbergelson: unweirded

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
lbergelson: @droazen  I don't think it's possible to get a file not found as a CommandLineParserException, but I see your point here.  I've updated to use the actual test files.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
lbergelson: >=

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
lbergelson: fixed

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportUnitTest.java
lbergelson: added, also added a test for whitespace within a token

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportUnitTest.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: nm, I was looking at the wrong whitespace

lbergelson:
@droazen I think I've addressed all your comments.  Going to merge when tests pass.

PR #2681, opened 2017-05-08
HomRefBlock: when doing bounds checking in add(), treat GQs > 99 as 99.

droazen:
This fixes an explosion in the HaplotypeCaller in protected caused by
the recent change in public to cap the last GVCF band.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/variant/writers/HomRefBlock.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/variant/writers/HomRefBlock.java
@@ -1,6 +1,5 @@
 package org.broadinstitute.hellbender.utils.variant.writers;
 
-import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.util.Locatable;
 import htsjdk.variant.variantcontext.Allele;
 import htsjdk.variant.variantcontext.Genotype;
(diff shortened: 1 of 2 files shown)


droazen:
@lbergelson please review

src/main/java/org/broadinstitute/hellbender/utils/variant/writers/HomRefBlock.java
```diff
@@ -94,15 +94,20 @@ private Genotype createHomRefGenotype(String sampleName) {
...
     public void add(final int pos, final Genotype genotype) {
         Utils.nonNull(genotype, "genotype cannot be null");
         if ( ! genotype.hasPL() ) { throw new IllegalArgumentException("genotype must have PL field");}
         if ( pos != end + 1 ) { throw new IllegalArgumentException("adding genotype at pos " + pos + " isn't contiguous with previous end " + end); }
         if ( genotype.getPloidy() != ploidy) { throw new IllegalArgumentException("cannot add a genotype with a different ploidy: " + genotype.getPloidy() + " != " + ploidy); }
-        if ( !withinBounds(genotype.getGQ())) {
+        // Make sure the GQ is within the bounds of this band. Treat GQs > 99 as 99.
+        if ( !withinBounds(Math.min(genotype.getGQ(), VCFConstants.MAX_GENOTYPE_QUAL))) {
             throw new IllegalArgumentException("cannot add a genotype with GQ=" + genotype.getGQ() + " because it's not within bounds ["
                     + this.getGQLowerBound() + ',' + this.getGQUpperBound() + ')');
```
lbergelson: I think we need to make the constructor cap the maximum upper bound at 99 as well.  Right now it's possible to make one with a higher upper bound and then this downconversion will cause issues.

droazen: Done (capped at 100 though, since the upper bound is exclusive rather than inclusive)

PR #2690, opened 2017-05-09
ProgressMeter: allow labels to be configurable per-traversal/tool, and hook up to GenomicsDBImport

### LINKED ISSUE #1943: customizable labels in progress monitor
current monitor always says "Records Processed   Records/Minute"
this is less good than 
"Reads Processed   Reads/Minute"
"Assembly Regions Processed   Assembly Regions/Minute"

### LINKED ISSUE #2683: GenomicsDBImport is not hooked up to its progress meter
This tool always emits "Processed 0 total records" at traversal end. We just need to hook it up to the standard progress meter.

droazen:
-Tools can now customize the progress meter to use a different word than
 "records" in its output (eg., "reads", "regions", etc.)

-Updated standard walker classes to specify appropriate labels

-Hooked up GenomicsDBImport to the progress meter (it was always reporting
 "Processed 0 records" at traversal end)

Resolves #1943
Resolves #2683

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/ProgressMeter.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/ProgressMeter.java
@@ -2,6 +2,7 @@ package org.broadinstitute.hellbender.engine;
 
 import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.util.Locatable;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.hellbender.utils.Utils;
@@ -50,6 +51,11 @@ public final class ProgressMeter {
      * Number of milliseconds in a minute
      */
     public static final long MILLISECONDS_PER_MINUTE = MILLISECONDS_PER_SECOND * 60L;
+
+    /**
+     * Default label for records in logger messages. For display purposes only.
+     */
+    public static final String DEFAULT_RECORD_LABEL = "records";
 
     /**
      * We output a line to the logger after this many seconds have elapsed
@@ -115,6 +121,11 @@ public final class ProgressMeter {
     private boolean stopped;
 
     /**
+     * Label for records in logger messages. For display purposes only.
+     */
+    private String recordLabel = DEFAULT_RECORD_LABEL;
+
+    /**
      * Create a progress meter with the default update interval of {@link #DEFAULT_SECONDS_BETWEEN_UPDATES} seconds
      * and the default time function {@link #DEFAULT_TIME_FUNCTION}.
      */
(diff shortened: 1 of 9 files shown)


droazen:
@cmnbroad please review

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -169,6 +169,9 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
         return 2;
     }
 
+    @Override
+    public String getProgressMeterRecordLabel() { return "batches"; }
```
cmnbroad: It seems awkward to have the default batchSize be 0, which means no batching, but to have the progress units be "batches".

src/main/java/org/broadinstitute/hellbender/engine/ProgressMeter.java
```diff
@@ -210,15 +230,17 @@ public void stop() {
         Utils.validate( !stopped, "the progress meter has been stopped already");
         this.stopped = true;
         currentTimeMs = timeFunction.getAsLong();
-        logger.info(String.format("Traversal complete. Processed %d total records in %.1f minutes.", numRecordsProcessed, elapsedTimeInMinutes()));
+        logger.info(String.format("Traversal complete. Processed %d total %s in %.1f minutes.", numRecordsProcessed, recordLabel, elapsedTimeInMinutes()));
```
magicDGS: Maybe now that you are here it may be useful to add a more verbose output by:;

1. Call `printProgress()` before logging the last line to see the last locus processed, and total processing rate.
2. It says "Traversal complete", but it may be used in other contexts. Maybe rephrase the message to "Processing complete: %s total %s in %.1f minute" may be useful for other use cases (and the "Traversal complete" could be moved to the `GATKTool` specific method...
3. Similarly to the progress logger in HTSJDK, it may be useful to customize "Processed" with tool-specific names ("Sorted", "Merged", "Traversed"...). This is just something that may be useful, but I really do not need it.

I'm very interested in have 1 to "profile" tools easier just by looking at the last row of the progress meter... but these are just suggestions...

src/main/java/org/broadinstitute/hellbender/engine/ProgressMeter.java
```diff
@@ -161,6 +172,15 @@ public void setRecordsBetweenTimeChecks( final long recordsBetweenTimeChecks ) {
     }
 
     /**
+     * Change the label used for records in logger messages. Default label is {@link #DEFAULT_RECORD_LABEL}
+     *
+     * @param label Label to use for records in logger messages.
+     */
+    public void setRecordLabel( final String label ) {
```
droazen: Added

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
droazen: Well, this is really just a matter of semantics. "No batching" is just another way of saying "1 batch", for the purposes of this tool :)

src/main/java/org/broadinstitute/hellbender/engine/ProgressMeter.java
droazen: I've added a final call to `printProgress()` in `stop()`, as requested.

Since the existing progress meter class has a notion of "current locus", I think in its current state it only makes sense to use it for traversals over the genome. We could consider generalizing it for other purposes in a future PR, however.

magicDGS: Thanks for taking into account my suggestions! :)

I'm using it also for a sorting iteration, but it is alright anyway.

PR #2695, opened 2017-05-10
adding openjdk to our tests

lbergelson:
Adding tests on openJDK since it seems like we're going to have to support it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,13 @@ git:
   depth: 9999999
 jdk:
 - oraclejdk8
+- openjdk8
 env:
   matrix:
-  - TEST_TYPE=cloud
+  - TEST_TYPE=cloud 
   - TEST_TYPE=integration TEST_VERBOSITY=minimal
   - TEST_TYPE=unit TEST_VERBOSITY=minimal
+
   global:
   #for genomics db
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -5,11 +5,13 @@ git:
   depth: 9999999
 jdk:
 - oraclejdk8
+- openjdk8
```
droazen: Can you explain whether this will cause us to run the test suite twice, once with Oracle JDK and once with OpenJDK? Will it affect our test suite runtime at all?

.travis.yml
```diff
@@ -5,11 +5,13 @@ git:
...
 env:
   matrix:
-  - TEST_TYPE=cloud
+  - TEST_TYPE=cloud
```
droazen: I assuming this invisible character is just innocent whitespace...

droazen:
:+1: merge after addressing comments @lbergelson

.travis.yml
lbergelson: oh, yeah, some whitespace there

.travis.yml
lbergelson: It means we run our tests twice.  It doubles the compute but only increases the wallclock time slightly due to the fact that we have twice as many chances for a weird slow outlier to happen.  We could just replace oracle with openjdk if we don't want to consume more compute.

droazen:
Merging this one

PR #2709, opened 2017-05-11
Issue 2457 docker updates

### LINKED ISSUE #2457: Script to create a GATK base image for docker
Then gatk-protected and gatk docker images can specify this one in ``USING``.

See gatk-protected scripts/docker for both a deployment script and a Dockerfile that (subset) can be used for a base image.

### LINKED ISSUE #2700: Can we stop relying on the working directory when we create the docker image?
(assumes that #2457 is already complete)
Currently, the docker image creation will grab all the local files, which is prone to error.  The github hash/tag parameters are simply to determine how to tag the image in docker hub.  In gatk-protected, we used to actually grab the source code from github in the Dockerfile.  I'm not suggesting we go back to this model.

What we might want to consider:
The build_docker.sh script could download the github hash/tag to a temp dir and then build the docker image from the temp dir.  This solution would be more robust and would decrease errors.

### LINKED ISSUE #2708: PSUtilsTest should be rewritten so it succeeds when it runs as root
PSUtilsTest should be rewritten so it succeeds when it runs as root.  It could possibly make use of  `BaseTest.getSafeNonExistentFile()`.

LeeTL1220:
- Create a GATK4 Dockerfile
- Added a script to create the docker image for the GATK4 Dockerfile.  This will create a staging directory if desired.
- Created a GATK Base Dockerfile that other GATK4 Dockerfiles can use in ``FROM``
- Added a script to create the GATK Base image
- Documentation
- Fixed unit tests that fail in the docker image.
- docker images based on OpenJDK 8

Closes #2457 
Closes #2708 
Closes #2700

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -60,6 +60,9 @@ final hadoopBamVersion = System.getProperty('hadoopBam.version','7.8.0')
 final genomicsdbVersion = System.getProperty('genomicsdb.version','0.6.2-proto-3.0.0-beta-1')
 final testNGVersion = '6.11'
 
+final baseJarName = 'gatk'
+final secondaryBaseJarName = 'hellbender'
+
 configurations.all {
     resolutionStrategy {
         force 'com.google.http-client:google-http-client:1.21.0'
@@ -218,14 +221,11 @@ def createSymlinks(archivePath, symlinkLocation) {
 }
 
 // Suffix is what will be added to the symlink
-def createAllSymlinks(destinationDir, archivePath, suffix) {
-    def finalSuffix = "-" + suffix
-    if (suffix == "") {
-        finalSuffix = ""
-    }
+def createGatkSymlinks(destinationDir, archivePath, suffix, baseJarName, secondaryBaseJarName) {
+    def finalSuffix = (suffix == "") ? "" : ("-" + suffix)
 
-    def symlinkLocation = destinationDir.toString() + "/hellbender-protected" + finalSuffix + ".jar"
-    def symlinkLocation2 = destinationDir.toString() + "/gatk" + finalSuffix + ".jar"
+    def symlinkLocation = destinationDir.toString() + "/" + baseJarName + finalSuffix + ".jar"
+    def symlinkLocation2 = destinationDir.toString() + "/" + secondaryBaseJarName + finalSuffix + ".jar"
 
     createSymlinks(archivePath.getAbsolutePath(), symlinkLocation)
     createSymlinks(archivePath.getAbsolutePath(), symlinkLocation2)
@@ -372,7 +372,7 @@ shadowJar {
         // Create a symlink to the newly created jar.  The name will be gatk.jar and
         //  it will be at the same level as the newly created jar.  (overwriting symlink, if it exists)
         // Please note that this will cause failures in Windows, which does not support symlinks.
-        createAllSymlinks(destinationDir.toString(), archivePath, "")
+        createGatkSymlinks(destinationDir.toString(), archivePath, "", baseJarName, secondaryBaseJarName)
     }
 }
 
@@ -387,7 +387,7 @@ task sparkJar(type: ShadowJar) {
         // Create a symlink to the newly created jar.  The name will be gatk.jar and
         //  it will be at the same level as the newly created jar.  (overwriting symlink, if it exists)
         // Please note that this will cause failures in Windows, which does not support symlinks.
-        createAllSymlinks(destinationDir.toString(), archivePath, classifier)
+        createGatkSymlinks(destinationDir.toString(), archivePath, classifier, baseJarName, secondaryBaseJarName)
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -210,6 +210,27 @@ processTestResources {
 sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
+def createSymlinks(archivePath, symlinkLocation) {
+    exec {
+        commandLine 'ln', '-fs', archivePath, symlinkLocation
+        ignoreExitValue = false
+    }
+}
+
+// Suffix is what will be added to the symlink
+def createAllSymlinks(destinationDir, archivePath, suffix) {
+    def finalSuffix = "-" + suffix
+    if (suffix == "") {
+        finalSuffix = ""
+    }
+
+    def symlinkLocation = destinationDir.toString() + "/hellbender-protected" + finalSuffix + ".jar"
+    def symlinkLocation2 = destinationDir.toString() + "/gatk" + finalSuffix + ".jar"
+
+    createSymlinks(archivePath.getAbsolutePath(), symlinkLocation)
+    createSymlinks(archivePath.getAbsolutePath(), symlinkLocation2)
+}
+
 final isRelease = Boolean.getBoolean("release")
 version = (isRelease ? gitVersion() : gitVersion() + "-SNAPSHOT").replaceAll(".dirty", "")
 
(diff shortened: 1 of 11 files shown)


build.gradle
```diff
@@ -210,6 +210,27 @@ processTestResources {
...
+def createSymlinks(archivePath, symlinkLocation) {
+    exec {
+        commandLine 'ln', '-fs', archivePath, symlinkLocation
+        ignoreExitValue = false
+    }
+}
+
+// Suffix is what will be added to the symlink
+def createAllSymlinks(destinationDir, archivePath, suffix) {
+    def finalSuffix = "-" + suffix
```
vruano: Is there a conditional assign in Groovy/Gradle? so that you can do these 4 lines in one?

```
def finalSuffix = suffix == "" ? "" : "-" + suffix;
```

build.gradle
```diff
@@ -210,6 +210,27 @@ processTestResources {
...
 
+def createSymlinks(archivePath, symlinkLocation) {
+    exec {
+        commandLine 'ln', '-fs', archivePath, symlinkLocation
+        ignoreExitValue = false
+    }
+}
+
+// Suffix is what will be added to the symlink
+def createAllSymlinks(destinationDir, archivePath, suffix) {
```
vruano: This method seems to be doing something very specific as it has some hardcoded file name bits in it (e.g. "/gatk???", "/hellbender-???" , ".jar" ...)

I think the name should be more specific then. Something like "createGatkJarSymlinks".

build.gradle
```diff
@@ -210,6 +210,27 @@ processTestResources {
...
+    def finalSuffix = "-" + suffix
+    if (suffix == "") {
+        finalSuffix = ""
+    }
+
+    def symlinkLocation = destinationDir.toString() + "/hellbender-protected" + finalSuffix + ".jar"
```
vruano: Can we extract this string literals into constants or perhaps can be deduced from some other existing properities/variables in the Gradle file?

build.gradle
```diff
@@ -347,6 +368,12 @@ shadowJar {
     configurations = [project.configurations.runtime]
     classifier = 'local'
```
vruano: Do we ever use the value of ```classifier```?

build.gradle
```diff
@@ -347,6 +368,12 @@ shadowJar {
...
     mergeServiceFiles('reference.conf')
+    doLast {
+        // Create a symlink to the newly created jar.  The name will be gatk.jar and
+        //  it will be at the same level as the newly created jar.  (overwriting symlink, if it exists)
+        // Please note that this will cause failures in Windows, which does not support symlinks.
+        createAllSymlinks(destinationDir.toString(), archivePath, "")
+    }
 }
 
 task localJar{ dependsOn shadowJar }
```
vruano: It seems to me that perhaps is more appropriate that ```shadowJar``` invokes ```localJar``` rather than the opposite in order to make is more coherent with ```sparkJar``` as ```local``` is better contrasting term (classifier?) to ```spark```.

build_docker.sh
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env bash
+
+## $ BRANCH=`git rev-parse --symbolic-full-name --abbrev-ref HEAD` &&  sudo bash build_docker.sh -e `git rev-parse ${BRANCH}` -s -r -d /home/BROAD.MIT.EDU/lichtens/gatk_docker_staging/
```
vruano: Either remove this comment or add another commend line that indicates why it is useful to keep it around.

build_docker.sh
```diff
@@ -0,0 +1,129 @@
...
+	esac
+done
+
+if [ -z "$GITHUB_TAG" ]; then
+	printf "Option -e requires an argument.\n \
+Usage: %s: -e <GITHUB_TAG> [-psl] \n \
+where <GITHUB_TAG> is the github tag (or hash when -s is used) to use in building the docker image\n \
+(e.g. bash build_docker.sh -e 1.0.0.0-alpha1.2.1)\n \
+Optional arguments:  \n \
+-p \t (DEV) push image to docker hub once complete.  This will use the GITHUB_TAG in dockerhub as well. \n \
```
vruano: What does ```(DEV)``` mean? ```(Development only)```, ```(Advance)```.  in such case it should go to the end of help.

vruano: Docker is kinda always for development so this would be redundant, right?

scripts/docker/gatkbase/build_docker_base.sh
```diff
@@ -0,0 +1,60 @@
...
+if [ -n "$IS_HELP" ]; then
+	printf "Build the GATK4 base image.\n \
+Usage: %s: [-p] [-h] \n \
+Optional arguments:  \n \
+-p \t push image to docker hub once complete.  This will use the GITHUB_TAG in dockerhub as well. \n \n" $0
+	exit 1
+fi
+
+# Make sure sudo or root was used.
+if [ "$(whoami)" != "root" ]; then
```
vruano: Why we need to be root, is it really necessary? ca we use sudo instead whenever privileges are needed?

src/test/java/org/broadinstitute/hellbender/utils/nio/GcsNioIntegrationTest.java
```diff
@@ -21,12 +21,12 @@
...
     final String largeFilePath = "large/human_g1k_v37.20.21.fasta";
 
-    @Test
+    @Test(groups={"bucket"})
     public void testGcsEnabled() {
         FileSystem fs = FileSystems.getFileSystem(URI.create("gs://domain-registry-alpha"));
     }
 
-    @Test
+    @Test(groups={"bucket"})
```
vruano: out of curiosity, what is this ```groups={"bucket"}``` declaration for?

build.gradle
LeeTL1220: Done

build.gradle
LeeTL1220: Done

build.gradle
LeeTL1220: Also eliminated ``-protected`` since this is gatk public.

LeeTL1220: Done

build.gradle
LeeTL1220: It shows up, so, I think the answer is yes.

build_docker.sh
LeeTL1220: Done

LeeTL1220: People make their own images all the time (for better or worse).

scripts/docker/gatkbase/build_docker_base.sh
LeeTL1220: This will check that sudo was used.

No action.

build.gradle
LeeTL1220: No action

build_docker.sh
LeeTL1220: Put in there by accident.  This is in the docs already.  I removed.

Done.

src/test/java/org/broadinstitute/hellbender/utils/nio/GcsNioIntegrationTest.java
LeeTL1220: "Requires GCS bucket access."

No action.

build.gradle
LeeTL1220: This is just to make the two equivalent.

I filed an issue for this, since this was not originally my decision.

No action.

LeeTL1220:
@vruano I have addressed comments.

PR #2716, opened 2017-05-15
Add command line to VCF outputs.

cmnbroad:
This is for VCFs only. BAM writers are mostly already done, though we'll also need to make a change to gatk-protected for HC/M2.


src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1107,4 +1110,13 @@ protected T computeNext() {
...
                 .collect(Collectors.toSet());
     }
+
+    /**
+     * Return the given {@code dateTime} formatted as string for display.
+     * @param dateTime the datte/time to be formatted
+     * @return String representing the {@code dateTime}.
+     */
+    public static String getDateTimeForDisplay(final ZonedDateTime dateTime) {
+        return dateTime.format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.LONG));
```
pshapiro4broad: Since `DateTimeFormatter` is immutable and thread safe, you can store the result of `DateTimeFormatter.ofLocalizedDateTime()` in a static final field and avoid the creation every time this method is called.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -666,6 +675,24 @@ protected SAMFileHeader getHeaderForSAMWriter(){
...
+    protected Set<VCFHeaderLine> getDefaultToolVCFHeaderLines() {
+        final Set<VCFHeaderLine> gatkToolHeaderLines = new HashSet<>();
+        if (addOutputVCFCommandLine) {
+            final Map<String, String> simpleHeaderLineMap = new HashMap<>(4);
+            simpleHeaderLineMap.put("ID", this.getClass().getSimpleName());
+            simpleHeaderLineMap.put("Version", getVersion());
+            simpleHeaderLineMap.put("Date", Utils.getDateTimeForDisplay((ZonedDateTime.now())));
+            simpleHeaderLineMap.put("CommandLine", getCommandLine());
+            gatkToolHeaderLines.add(new VCFHeaderLine("source", this.getClass().getSimpleName()));
+            gatkToolHeaderLines.add(new VCFSimpleHeaderLine("GATKCommandLine", simpleHeaderLineMap));
```
magicDGS: Could the ID "GATKCommandLine" be retrieve for an overridable method? It will be cool for other toolkits that wants to use the `GATKTool` interface for traversal but want to distinguish their command line into the header...

cmnbroad: Yes - if we add a "getToolKitName()", we can use it here and in getToolName(), which is in turn used in the getHeaderForSAMWriter code, which does something similar to this for SAM/BAM outputs. Also I really should move the rest of these to static constants...

magicDGS: That's a great idea and it will help for custom toolkits!

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
```diff
@@ -523,6 +527,28 @@ private File setupVCFWriter(
...
+        VCFHeader hdr = new VCFHeader(stdHeaderLines);
+
+        VCFHeaderLine sourceLine = hdr.getOtherHeaderLine("source");
+        Assert.assertEquals(sourceLine.getValue(), tool.getClass().getSimpleName());
+
+        VCFIDHeaderLine commandLine = (VCFIDHeaderLine) hdr.getOtherHeaderLine("GATKCommandLine");
+        Assert.assertEquals(commandLine.getID(), tool.getClass().getSimpleName());
+
+        String commandLineString = commandLine.toString();
+        Assert.assertTrue(commandLineString.contains("CommandLine="));
```
lbergelson: we have BaseTest.assertContains(String,String) which does this but with a nicer error message.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -88,6 +94,9 @@
     @Argument(fullName = StandardArgumentDefinitions.ADD_OUTPUT_SAM_PROGRAM_RECORD, shortName = StandardArgumentDefinitions.ADD_OUTPUT_SAM_PROGRAM_RECORD, doc = "If true, adds a PG tag to created SAM/BAM/CRAM files.", optional=true, common = true)
     public boolean addOutputSAMProgramRecord = true;
 
+    @Argument(fullName = StandardArgumentDefinitions.ADD_OUTPUT_VCF_COMMANDLINE, shortName = StandardArgumentDefinitions.ADD_OUTPUT_VCF_COMMANDLINE, doc = "If true, adds a command line header line to created VCF files.", optional=true, common = true)
+    public boolean addOutputVCFCommandLine = true;
```
lbergelson: I still think it's weird to have boolean flags that default to true.  I would invert this so the command line input is `suppressOutputVCFCommandline` or something like that, but I feel like we've had this conversation before and the conclusion was that this is fine for some flags.

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerGCSSupportIntegrationTest.java
```diff
@@ -60,7 +60,7 @@ public void testReadVCFOnGCS( final String vcf, final List<SimpleInterval> inter
         final IntegrationTestSpec testSpec = new IntegrationTestSpec(
                 " -V " + getGCPTestInputPath() + vcf +
                 intervalArg +
-                " -O %s",
+                " -O %s "+ " --addOutputVCFCommandLine false ",
```
lbergelson: this is unfortunate to have to add to every integration test

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
lbergelson: I"m not suer how valuable it is to put those in constants, but 👍 for the getToolKitName

lbergelson:
@cmnbroad Feel free to merge after addressing the issues brought up by the various reviewers.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
magicDGS: At least the `getToolKitName()`. The others are alright, because they are not toolkit-dependent...

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
cmnbroad: Done.

PR #2725, opened 2017-05-17
Added FPGA support to PairHMM via GKL

erniebrau:
This also includes:
* Added FPGA to ```Implementation``` enum of ```VectorLoglessPairHMM```
* Added FPGA to ```Implementation``` enum of ```PairHMM```
* Updated ```FASTEST_AVAILABLE``` value to try FPGA first (then AVX+OMP, then AVX, then regular)

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
@@ -65,6 +65,13 @@ public abstract class PairHMM implements Closeable{
             logger.info("Using the OpenMP multi-threaded AVX-accelerated native PairHMM implementation");
             return hmm;
         }),
+        /* FPGA implementation of LOGLESS_CACHING called through JNI. Throws if FPGA is not available */
+        EXPERIMENTAL_FPGA_LOGLESS_CACHING(args -> {
+            // Constructor will throw a UserException if FPGA is not available
+            final VectorLoglessPairHMM hmm = new VectorLoglessPairHMM(VectorLoglessPairHMM.Implementation.FPGA, args);
+            logger.info("Using the FPGA-accelerated native PairHMM implementation");
+            return hmm;
+        }),
         /* Uses the fastest available PairHMM implementation supported on the platform.
            Order of precedence:
             1. AVX_LOGLESS_CACHING_OMP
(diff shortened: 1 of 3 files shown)


erniebrau:
As requested, we've marked FPGA as experimental and removed FPGA from the list of alternatives when choosing ```FASTEST_AVAILABLE```.

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
```diff
@@ -65,13 +65,28 @@
             logger.info("Using the OpenMP multi-threaded AVX-accelerated native PairHMM implementation");
             return hmm;
         }),
+        /* FPGA implementation of LOGLESS_CACHING called through JNI. Throws if FPGA is not available */
+        EXPERIMENTAL_FPGA_LOGLESS_CACHING(args -> {
+            // Constructor will throw a UserException if OpenMP AVX is not available
```
droazen: Comments says throws if "OpenMP AVX" is not available, but comment above says "Throws if FPGA is not available" -- I assume the latter is correct?

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/PairHMM.java
```diff
@@ -65,13 +65,28 @@
...
+            return hmm;
+        }),
         /* Uses the fastest available PairHMM implementation supported on the platform.
            Order of precedence:
             1. AVX_LOGLESS_CACHING_OMP
             2. AVX_LOGLESS_CACHING
             3. LOGLESS_CACHING
          */
         FASTEST_AVAILABLE(args -> {
+//            try {
```
droazen: Add a comment explaining why this block is commented out for now.

src/main/java/org/broadinstitute/hellbender/utils/pairhmm/VectorLoglessPairHMM.java
```diff
@@ -72,7 +77,14 @@ public VectorLoglessPairHMM(Implementation implementation, PairHMMNativeArgument
                 if (!isSupported) {
                     throw new UserException.HardwareFeatureException("Machine does not support OpenMP AVX PairHMM.");
                 }
-                logger.info("Maximum number of OpenMP threads for AVX : " + args.maxNumberOfThreads);
```
droazen: Shouldn't we still output the number of threads here?

erniebrau: We removed this because the latest version of GKL (0.5.2) now outputs this information, along with the total number of available threads/cores. It also outputs a warning when more threads are requested than are available.

droazen:
@erniebrau Can you rebase this onto the latest master?

droazen:
@erniebrau Whoa, something went VERY wrong with your rebase! It looks like you accidentally rebased hundreds of commits from master, not just your own. Can you fix this?

What you need to do is `git reset` back to a sane state (`git reset --hard commit_id_of_branch_before_last_rebase`, which you can get from `git reflog`), then `git fetch && git rebase origin/master` to re-do the rebase.

erniebrau:
Sorry about that. Looking at my shell history, it looks like I pulled after I rebased for some reason, and that's what caused the mess.

I fixed it now (thanks for the instructions, by the way).

PR #2726, opened 2017-05-17
Issue 2697 docker testing

LeeTL1220:
- Attempts testing of docker

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,11 @@ jdk:
 - openjdk8
 env:
   matrix:
-  - TEST_TYPE=cloud 
+  - TEST_TYPE=cloud
   - TEST_TYPE=integration TEST_VERBOSITY=minimal
   - TEST_TYPE=unit TEST_VERBOSITY=minimal
-
+  - TEST_TYPE=integration TEST_DOCKER=true TEST_VERBOSITY=minimal
+  - TEST_TYPE=unit TEST_DOCKER=true TEST_VERBOSITY=minimal
   global:
   #for genomics db
   - LD_LIBRARY_PATH=$TRAVIS_BUILD_DIR/dependencies/libcsv/.libs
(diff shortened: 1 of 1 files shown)


LeeTL1220:
@vdauwera @droazen @lbergelson @cmnbroad Hey, this pull request will fail as long as the docker hub repo for broadinstitute/gatk has restricted read access.  Any objections to making reading of the gatk (not gatk-protected) dockerhub repo public?

lbergelson:
@LeeTL1220 No objections from me.

cmnbroad:
Improve testing and reduce costs. Sounds right to me.

LeeTL1220:
Got the go-ahead and the access has been changed.

LeeTL1220:
Ready to merge pending review.

.travis.yml
```diff
@@ -10,6 +10,8 @@ env:
   - TEST_TYPE=cloud
   - TEST_TYPE=integration TEST_VERBOSITY=minimal
   - TEST_TYPE=unit TEST_VERBOSITY=minimal
+  - TEST_TYPE=integration TEST_DOCKER=true TEST_VERBOSITY=minimal
+  - TEST_TYPE=unit TEST_DOCKER=true TEST_VERBOSITY=minimal
```
LeeTL1220: > Both this PR and the openjdk one appear to add to the overall time it takes to run the tests, so maybe we should just rely on the docker tests for our openjdk testing ?

The docker testing is probably adequate to cover openJDK vs oracleJDK, but remember that the docker testing does not test any cloud or bucket tests.  I would argue that we keep both unless we are running too many tests.

> I'm wondering if it makes sense to run both the integration and unit tests - do you know if there is additional cost to us in terms git-lfs bandwith ?

I think it makes sense.  
- Are you sure that no unit tests use files in git-lfs?
- There is a few minutes of additional time to download the git lfs files.  Building the docker image is still the bottleneck.
- I definitely want to see that both types of tests work in the docker image, so we have to do integration.  Also, I worry that the coverage is lower if we only do integration, so I also want unit.  The docker image should become standard in pipelines, so I want it as vetted as possible and I've already seen it fail on unit tests and not integration tests.

LeeTL1220: Tell me if I've convinced you, since I am taking no action.

LeeTL1220:
@cmnbroad Back to you....

.travis.yml
cmnbroad: @LeeTL1220 Sounds reasonable. We can alway tweak if this gets to be an issue.

LeeTL1220:
@cmnbroad Filed an issue for the repeating of tests of the docker image (openjdk vs. oraclejdk is irrelevant on the docker tests).  #2748

PR #2727, opened 2017-05-17
Eliminate allele count weighting in AlleleSubsetting likelihood sum to match GATK 3

davidbenjamin:
@lbergelson I believe this change is the only difference, but I would appreciate if you check whether this branch's likelihood sums match those of GATK 3.


davidbenjamin:
@droazen @meganshand Because you care, too.  Louis discovered that the likelihood sum, with allele count weighting, was a port of `IndependentAllelesDiploidExactAFCalculator`, while GATK 3 actually calls the likelihood sum of `DiploidExactAFCalculator`.

PR #2729, opened 2017-05-17
Add Hopscotch and Bloom filter classes for gigantic sets of longs

mwalker174:

There are 4 new classes here:

1. LongHopscotchSet - based on HopscotchCollection/Set but adapted to store primitive longs instead of objects. The most significant bit is used to tell if a bucket is null or not, so the longs being stored must be non-negative. This works for use with k-mers, which we are assume are odd-length up to 31 and thus consume up to 62 bits.
2. LargeLongHopscotchSet - for sets of longs greater than ~2 billion (the max Java array size) using a List of LongHopscotchSets.
3. LongBloomFilter - Bloom filter for long's
4. LargeLongBloomFilter - Bloom filter when the filter index size exceeds 2GB using a List of LongBloomFilters.

- LongIterator and QueryableLongSet interfaces for convenience.
- Minor change to HopscotchSet max legal size, which was higher than the actual allowed Java array size.

PS I just had a thought that the Bloom filters could use long instead of byte buckets to expand the max index size 8-fold. Could maybe be done for the Hopscotch sets as well, but with considerably more difficulty. Thoughts? On the other hand, the performance is already adequate so perhaps I'll save this idea for later.


src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
```diff
@@ -0,0 +1,224 @@
...
+        sets = new ArrayList<>(numSets);
+        for (int i = 0; i < numSets; i++) {
+            sets.add(kryo.readObject(stream, LongHopscotchSet.class));
+        }
+
+        kryo.setReferences(oldReferences);
+    }
+
+    private static int longHash(final long entryVal) {
+        return (int) SVUtils.fnvLong64(0x4c79ac6a, entryVal);
```
tedsharpe: Ha!  Where did this value come from?  Seems like a pretty good seed -- good balance of 1's and 0's with no obvious pattern -- but it would be better to compute a hash only once as mentioned immediately above.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
```diff
@@ -0,0 +1,224 @@
...
+        final int maxPartitionSize = LongHopscotchSet.getLegalSizeBelow(maxPartitionBytes / LongHopscotchSet.bytesPerBucket);
+        //Truncate numElements to largest int possible
+        final int idealPartitionSize = LongHopscotchSet.getPartitionSize(numElements);
+        //Number of elements per partition
+        final int elementsPerPartitionGuess = Math.min(idealPartitionSize, maxPartitionSize);
+        final int partitions = (int) Math.ceil(((double) numElements) / elementsPerPartitionGuess);
+        final int elementsPerPartition = (int)Math.ceil(numElements / (double)partitions);
+
+        for (int i = 0; i < partitions; i++) {
+            sets.add(new LongHopscotchSet(elementsPerPartition));
```
tedsharpe: I'm not sure, but I think you may not have taken into account that elementsPerPartition will be inflated by LOAD_FACTOR and getLegalSizeAbove.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
```diff
@@ -0,0 +1,224 @@
...
+        return setRef.add(entryValue);
+    }
+
+    public void addAll(final long[] entryValues) {
+        for (final long val : entryValues) {
+            add(val);
+        }
+    }
+
+    public long size() {
```
tedsharpe: Probably wouldn't be hard to keep your own total size.  But maybe this is unlikely to be called repetitively.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
```diff
@@ -0,0 +1,224 @@
...
+        long sum = 0;
+        for (final LongHopscotchSet s : sets) {
+            sum += s.size();
+        }
+        return sum;
+    }
+
+    public Collection<LongHopscotchSet> getSets() {
```
tedsharpe: I think this is an invitation to monkey business.  Maybe make it @VisibleForTesting rather than public?

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    private int setIndexOf(final long val) {
+        return Integer.remainderUnsigned(longHash(val), numSets);
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
tedsharpe: See comments in LongHopscotchSet about making this work with any QueryableLongSet.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        if (size != that.size) return false;
+        final LongIterator itr = this.iterator();
+        while (itr.hasNext()) {
+            if (!that.contains(itr.next())) return false;
+        }
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
```
tedsharpe: To meet the contract for hashCode and equals, this must be defined like it is in LargeLongHopscotchSet:  as a sum over the hashCodes of the elements.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
```diff
@@ -0,0 +1,224 @@
...
+
+        final LargeLongHopscotchSet that = (LargeLongHopscotchSet) o;
+
+        return numSets == that.numSets && sets.equals(that.sets);
+
+    }
+
+    @Override
+    public int hashCode() {
```
tedsharpe: might be a little faster to define this as:
```
sets.stream().mapToInt(LongHopscotchSet::hashCode).sum();
```

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
```diff
@@ -0,0 +1,224 @@
...
+            hopscotchSet.serialize(kryo, output);
+        }
+
+        @Override
+        public LargeLongHopscotchSet read(final Kryo kryo, final Input input, final Class<LargeLongHopscotchSet> klass) {
+            return new LargeLongHopscotchSet(kryo, input);
+        }
+    }
+
+    private final class LargeLongHopscotchSetIterator implements LongIterator {
```
tedsharpe: Just use an outer Iterator<LongHopscotchSet> and an inner LongIterator.
hasNext() checks the inner iterator for null or !hasNext and advances the outer iterator until it finds a new inner iterator that isn't empty, or until the outer iterator is exhausted.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
```diff
@@ -0,0 +1,224 @@
...
+ * not support dynamic resizing.
+ */
+@DefaultSerializer(LargeLongHopscotchSet.Serializer.class)
+public class LargeLongHopscotchSet implements Serializable, QueryableLongSet {
+
+    private static final long serialVersionUID = 1L;
+    private final List<LongHopscotchSet> sets;
+    private int numSets;
+
+    public LargeLongHopscotchSet(final long maxPartitionBytes, final long numElements) {
```
tedsharpe: actually, maybe getLegalSizeBelow(sqrt(numElements)) for numSets.  that way you'd know that numSets and inner-set capacity are relatively prime, which would be a good idea.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongBloomFilter.java
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Set of long's that is too large to fit in a single LongBloomFilter. Works by binning each entry into a
+ * corresponding LongBloomFilter partition using the entry hash value. Partitioning is handled internally given
+ * a maximum partition size, number of elements to be added, and false positive probability.
+ */
+@DefaultSerializer(LargeLongBloomFilter.Serializer.class)
+public class LargeLongBloomFilter implements QueryableLongSet {
+
+    private final transient Logger logger = LogManager.getLogger(this.getClass());
+    private List<LongBloomFilter> filters;
```
tedsharpe: I don't see any reason that this couldn't be an array rather than a list.  Size is fixed for all time.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongBloomFilter.java
```diff
@@ -0,0 +1,191 @@
...
+import org.broadinstitute.hellbender.utils.Utils;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Random;
+
+/**
+ * Set of long's that is too large to fit in a single LongBloomFilter. Works by binning each entry into a
+ * corresponding LongBloomFilter partition using the entry hash value. Partitioning is handled internally given
```
tedsharpe: I'm not at all sure about this, but I fear that this technique of choosing one of N filters to hold an entry might not be equivalent to having a single filter N times larger.  It feels like this way would have a greater false positive rate because of the stochasticity of the assignment:  some filters will end up with a few more entries and some with a few fewer.  But the error rate feels non-linear to me.  Certainly possible that I'm all wet.
A different possibility would be to use each filter as little more than a bit array -- just add setBit and testBit methods.  Entries get distributed across the entire set of filters:
```
public void add( final long entryVal ) {
   final long fnvHash = SVUtils.fnvLong64(entryVal);
   for ( i from 0 to numHashes ) {
      hash = applyHashFunc( i, fnvHash );
      filter[hash / numBits].setBit(hash % numBits);
   }
}
```
And then similarly with contains, using testBit instead of setBit.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+/**
+ * This class is based on the HopscotchCollection and HopscotchSet classes for storing Objects. Saves memory used by
+ * object headers by using primitive longs.
+ * <p>
+ * Non-negative longs only! We set the MSB to say that a bin is not empty.
+ */
+@DefaultSerializer(LongHopscotchSet.Serializer.class)
+public class LongHopscotchSet implements Serializable, QueryableLongSet {
+
+    protected static final int bytesPerBucket = 9;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+    // For power-of-2 table sizes add this line
+    //private static final int maxCapacity = Integer.MAX_VALUE/2 + 1;
+    // largest prime numbers less than each half power of 2 from 2^8 to 2^31
+    // Note the last size is the greatest prime that is an allowable Java array size (<=2^31-3)
+    @VisibleForTesting
+    static final int[] legalSizes = {
```
mwalker174: Added class SetSizeUtils for this

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+            251, 359, 509, 719, 1021, 1447, 2039, 2887, 4093, 5791, 8191, 11579, 16381, 23167, 32749, 46337, 65521,
+            92681, 131071, 185363, 262139, 370723, 524287, 741431, 1048573, 1482907, 2097143, 2965819, 4194301, 5931641,
+            8388593, 11863279, 16777213, 23726561, 33554393, 47453111, 67108859, 94906249, 134217689, 189812507,
+            268435399, 379625047, 536870909, 759250111, 1073741789, 1518500213, 2147483629
+    };
+    @VisibleForTesting
+    static final double LOAD_FACTOR = .85;
+    private static final long serialVersionUID = 1L;
+    private static final int NO_ELEMENT_INDEX = -1;
+    private static final int SPREADER = 241;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+    private int capacity;
+    // unused buckets contain null.  (this data structure does not support null entries.)
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+    // if the bucket is unused, the corresponding status byte is irrelevant, but is always set to 0.
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        this(12000);
+    }
+
+    /**
+     * make a LongHopscotchSet for a specified minimum capacity
+     */
+    public LongHopscotchSet(final int capacity) {
+        this.capacity = getLegalSizeAbove(capacity);
+        this.size = 0;
+        // Not unsafe, because the remainder of the API allows only elements known to be long's to be assigned to buckets.
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        this.buckets = new long[this.capacity];
+        this.status = new byte[this.capacity];
+        for (final long val : values) {
+            add(val);
+        }
+    }
+
+    protected LongHopscotchSet(final Kryo kryo, final Input stream) {
+        final boolean oldReferences = kryo.getReferences();
+        kryo.setReferences(false);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        capacity = stream.readInt();
+        size = 0;
+        buckets = new long[capacity];
+        status = new byte[capacity];
+        int nElements = stream.readInt();
+        while (nElements-- > 0) {
+            add(stream.readLong());
+        }
+
+        kryo.setReferences(oldReferences);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+    protected static int getLegalSizeAbove(final long minElements) {
+        final long augmentedSize = (long) (minElements / LOAD_FACTOR);
+        for (final int legalSize : legalSizes) {
+            if (legalSize >= augmentedSize) return legalSize;
+        }
+        throw new IllegalArgumentException("No legal sizes large enough for size " + minElements);
+    }
+
+    /**
+     * Computes next smallest legal table size for a given number of elements, accounting for load factor
```
mwalker174: Done, although I think it is getting the smallest table size greater than a specified capacity?

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        for (final int legalSize : legalSizes) {
+            if (legalSize >= augmentedSize) return legalSize;
+        }
+        return legalSizes[legalSizes.length - 1];
+    }
+
+    /**
+     * in a set, uniqueness is on the entry, so we just compare entries to see if we already have an identical one
+     */
+    private static LongBiPredicate entryCollides() {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        return (long t, long u) -> getValue(t) == getValue(u);
+    }
+
+    /**
+     * Returns the value for a given bucket entry (zeroing the bit used for declaring the bucket null)
+     */
+    private static long getValue(final long entry) {
+        return entry & ~Long.MIN_VALUE;
```
mwalker174: Even better

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+    public final long capacity() {
+        return capacity;
+    }
+
+    public final boolean contains(final long key) {
+        Utils.validateArg(isValidKey(key), "Tried to use negative key in LongHopScotchSet");
+        int bucketIndex = hashToIndex(key);
+        if (!isChainHead(bucketIndex)) return false;
+        long entryVal = getValue(buckets[bucketIndex]);
+        if (equivalentValue(entryVal, key)) return true;
```
mwalker174: Good catch

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        int offset;
+        while ((offset = getOffset(bucketIndex)) != 0) {
+            bucketIndex = getIndex(bucketIndex, offset);
+            entryVal = getValue(buckets[bucketIndex]);
+            if (equivalentValue(entryVal, key)) return true;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+    public final LongIterator iterator() {
+        return new LongCompleteIterator();
+    }
+
+    public final boolean remove(final long key) {
+        Utils.validateArg(isValidKey(key), "Tried to remove by negative key in LongHopScotchSet");
+        int bucketIndex = hashToIndex(key);
+        if (isNull(buckets[bucketIndex]) || !isChainHead(bucketIndex)) return false;
+        int predecessorIndex = NO_ELEMENT_INDEX;
+        while (!equivalentEntry(buckets[bucketIndex], key)) {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        removeAtIndex(bucketIndex, predecessorIndex);
+        return true;
+    }
+
+    public final int size() {
+        return size;
+    }
+
+    //Works with both raw entry with MSB set
+    private boolean equivalentEntry(final long entry, final long key) {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        return entry != 0 && equivalentValue(getValue(entry), key);
+    }
+
+    /**
+     * Returns whether a raw bucket entry (possibly with null bit set) and key (without null bit set) are equal
+     */
+    private boolean equivalentValue(final long entry, final long key) {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        int result = (SPREADER * longHash(entryVal)) % capacity;
+        if (result < 0) result += capacity;
+        return result;
+    }
+
+    private boolean isValidKey(final long key) {
+        return key >= 0;
+    }
+
+    private boolean insert(final long entryValue, final LongBiPredicate collision) {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+
+    private boolean isChainHead(final int bucketIndex) {
+        return (status[bucketIndex] & Byte.MIN_VALUE) != 0;
+    }
+
+    private int getOffset(final int bucketIndex) {
+        return status[bucketIndex] & Byte.MAX_VALUE;
+    }
+
+    private boolean isNull(final long val) {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+        }
+        if (nextOffset != 0) {
+            status[emptyBucketIndex] = (byte) (nextOffset - toEmptyDistance);
+        }
+        buckets[emptyBucketIndex] = buckets[bucketToMoveIndex];
+        buckets[bucketToMoveIndex] = 0;
+        status[bucketToMoveIndex] = 0;
+    }
+
+    @SuppressWarnings("unchecked")
```
mwalker174: Removed

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+    public void addAll(final long[] entryValues) {
+        for (final long val : entryValues) {
+            add(val);
+        }
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
mwalker174: I would do this except QueryableLongSet is also implemented by the Bloom filters (perhaps the interface name is not the best), which do not have a size and can't be iterated on.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+
+        final LongHopscotchSet that = (LongHopscotchSet) o;
+
+        if (size != that.size) return false;
+        final LongIterator itr = this.iterator();
```
mwalker174: See last comment

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
```diff
@@ -0,0 +1,681 @@
...
+
+    public boolean removeAll(final long[] set) {
+        boolean result = false;
+        for (final long val : set) {
+            if (remove(val)) result = true;
+        }
+        return result;
+    }
+
+    public interface LongBiPredicate {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/QueryableLongSet.java
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Interface for classes that provide ways to add long primitives and test for set membership
+ */
+public interface QueryableLongSet {
+
+    boolean add(final long val); //TODO needed for experimental kmer filtering approach
+    boolean contains(final long val);
+    boolean containsAll(final long[] vals);
+
```
mwalker174: See above

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchCollection.java
```diff
@@ -50,11 +50,12 @@
     private static final int NO_ELEMENT_INDEX = -1;
 
     // largest prime numbers less than each half power of 2 from 2^8 to 2^31
+    // Note the last size is the greatest prime that is an allowable Java array size (<=2^31-3)
     @VisibleForTesting static final int[] legalSizes = {
             251, 359, 509, 719, 1021, 1447, 2039, 2887, 4093, 5791, 8191, 11579, 16381, 23167, 32749, 46337, 65521,
             92681, 131071, 185363, 262139, 370723, 524287, 741431, 1048573, 1482907, 2097143, 2965819, 4194301, 5931641,
             8388593, 11863279, 16777213, 23726561, 33554393, 47453111, 67108859, 94906249, 134217689, 189812507,
-            268435399, 379625047, 536870909, 759250111, 1073741789, 1518500213, 2147483647
+            268435399, 379625047, 536870909, 759250111, 1073741789, 1518500213, 2147483629
```
mwalker174: You're welcome!

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+/**
+ * Set of longs that is larger than the max Java array size ( ~ 2^31 ~ 2 billion) and therefore cannot fit into a
+ * single LongHopscotchSet. Maintains partitions of LongHopscotchSets given a maximum partition size and the number
+ * of elements to be added. Bins each entry into a corresponding LongHopscotchSet using the entry hash value. Does
+ * not support dynamic resizing.
```
mwalker174: Definitely interested in how it works but for PathSeq we always know the number of elements at the start

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
mwalker174: maxPartitionBytes was originally there when I was broadcasting the kmer set and it was useful to try to tune. I like the sqrt(numElements) idea better so I'll use that. Very clever point to use getLegalSizeBelow.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongHopscotchSet.java
tedsharpe: Ah.  Roger.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
```diff
@@ -0,0 +1,224 @@
...
+        numSets = stream.readInt();
+        sets = new ArrayList<>(numSets);
+        for (int i = 0; i < numSets; i++) {
+            sets.add(kryo.readObject(stream, LongHopscotchSet.class));
+        }
+
+        kryo.setReferences(oldReferences);
+    }
+
+    private static int longHash(final long entryVal) {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
mwalker174: This line is removed now

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
mwalker174: Yes this is rarely called

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
mwalker174: Again I don't think it can be compared to a Bloom filter

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongHopscotchSet.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+
+import java.util.Arrays;
+
+/**
+ * Bloom filter for primitive longs. Useful for quickly querying whether a long is a part of a set when a
+ * a finite false positive probability can be tolerated. Optimal array size and number of hashes are determined
+ * for a given number of elements to be inserted and false prositive probability.
+ */
+@DefaultSerializer(LongBloomFilter.Serializer.class)
+public class LongBloomFilter implements QueryableLongSet {
```
mwalker174: Yes changed all to final

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+
+        final long optimal_num_bytes = getOptimalNumberOfBytes(num_elements, fpp);
+        this.numBuckets = getLegalSizeAbove(optimal_num_bytes);
+        this.numBits = (this.numBuckets * 8L);
+        final int optimal_num_hashes = (int) Math.round(-Math.log(fpp) / Math.log(2));
+        this.numHashes = optimal_num_hashes > 0 ? optimal_num_hashes : 1;
+        this.buckets = new byte[numBuckets];
+    }
+
+    public LongBloomFilter(final long minNumElements, final int numBytes) {
```
mwalker174: Sorry should have removed this beforehand. I don't know what it's doing either, but it isn't used anywhere. I see why it would be better to set numBits to a legal size even. Although this will decrease the max size by a factor of 8, it will not be an issue since I am going to implement your suggestion to maintain a list of bit arrays.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+        this.numBuckets = getLegalSizeAbove(minNumElements);
+        this.numBits = (this.numBuckets * 8L);
+        final int optimal_num_hashes = (int) Math.ceil(Math.log(2) * minNumElements / this.numBuckets);
+        this.numHashes = optimal_num_hashes > 0 ? optimal_num_hashes : 1;
+        this.buckets = new byte[numBuckets];
+    }
+
+    protected LongBloomFilter(final Kryo kryo, final Input input) {
+        final boolean oldReferences = kryo.getReferences();
+        kryo.setReferences(false);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+        numBuckets = input.readInt();
+        numHashes = input.readInt();
+        buckets = input.readBytes(numBuckets);
+
+        if (logger.isDebugEnabled()) {
+            final long X = countBits();
+            final long estimated_size = estimateSize(X);
+            logger.debug("Deserialized: numBits : " + numBits + ", numBuckets: " + numBuckets + ", numHashes: " + numHashes + ", bits set: " + X + ", approximate size: " + estimated_size);
+        }
+        kryo.setReferences(oldReferences);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+        return legalSizes[legalSizes.length - 1];
+    }
+
+    public static long getOptimalNumberOfBytes(final long num_elements, final double fpp) {
+        return (long) Math.ceil(-num_elements * Math.log(fpp) / (8 * Math.log(2) * Math.log(2)));
+    }
+
+    private long countBits() {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+            byte bb = b;
+            for (int i = 0; i < Byte.SIZE; i++) {
+                X += bb & 1;
+                bb >>= 1;
+            }
+        }
+        return X;
+    }
+
+    private long estimateSize(final long numBitsSet) {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+        return (long) (-(((double) numBits) / numHashes) * Math.log(1 - (((double) numBitsSet) / numBits)));
+    }
+
+    protected void serialize(final Kryo kryo, final Output output) {
+        final boolean oldReferences = kryo.getReferences();
+        kryo.setReferences(false);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+
+        output.writeLong(numBits);
+        output.writeInt(numBuckets);
+        output.writeInt(numHashes);
+        output.writeBytes(buckets);
+
+        kryo.setReferences(oldReferences);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    public boolean add(final long entryValue) {
+        final long[] bitIndex = entryIndices(entryValue);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+        for (final long val : entryValues) {
+            add(val);
+        }
+    }
+
+    /**
+     * Using different semantics than a Set because false positives are possible
+     */
+    public boolean contains(final long key) {
+        final long[] bitIndex = entryIndices(key);
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+            if (!contains(val))
+                return false;
+        }
+        return true;
+    }
+
+    /**
+     * Determines partition corresponding to the given the bit index
+     */
+    private int indexBucket(final long bitIndex) {
```
mwalker174: I've done away with this method but its replacements bitIndexToBucketArray() and bitIndexToBucketIndex() cannot be static

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+
+    /**
+     * Returns bucket bit mask with 1 in the position of given bit index
+     */
+    private byte bucketMask(final long bitIndex) {
+        return (byte) (1 << (bitIndex & 7));
+    }
+
+    public void clear() {
+        for (int i = 0; i < numBuckets; i++) {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+            if (b != 0) return false;
+        }
+        return true;
+    }
+
+    /**
+     * Computes hash values for a given entry. First, two hashes (h_, h_2) are computed using FNV1.
+     * Hash i is then computed by double hashing: h_i = h_1 + i * h_2. See reference:
+     * <p>
+     * Kirsch and Mitzenmacher. 2008. Less hashing, same performance: Building a better Bloom filter. Random
```
mwalker174: You're welcome! Moved to applyHashFunction()

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+     * Structures & Algorithms. 33:2, 187-218.
+     */
+    private long[] entryIndices(final long val) {
+        final long[] indices = new long[numHashes];
```
mwalker174: Yes I think this will be much faster and cut down on gc

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+            if (indices[i] < 0) indices[i] += numBits;
+        }
+        return indices;
+    }
+
+    @SuppressWarnings("SimplifiableIfStatement")
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+
+        final LongBloomFilter that = (LongBloomFilter) o;
+
+        if (numBits != that.numBits) return false;
+        if (numBuckets != that.numBuckets) return false;
```
mwalker174: Removing it from the equals check but I'd like to keep the number

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
```diff
@@ -0,0 +1,245 @@
...
+        if (numHashes != that.numHashes) return false;
+        return Arrays.equals(buckets, that.buckets);
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = (int) (numBits ^ (numBits >>> 32));
+        result = 31 * result + numBuckets;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongBloomFilter.java
mwalker174: I've done something like this now. It is cleaner than / faster than / obviates the need for LargeLongBloomFilter.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongBloomFilter.java
mwalker174: Noted but this class is now removed

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongBloomFilter.java
```diff
@@ -0,0 +1,191 @@
...
+                numPositive++;
+            }
+        }
+        return numPositive / (double) numTrials;
+    }
+
+    private void initialize(final long maxPartitionBytes, final long numElements, final double fpp) {
+        Utils.validateArg(maxPartitionBytes > 0, "Max partition size must be at greater than 0");
+        Utils.validateArg(numElements > 0, "Number of elements must be greater than 0");
+        Utils.validateArg(fpp > 0 && fpp < 1, "False positive probability must be between 0 and 1 (exclusive)");
```
mwalker174: Noted but this class is now removed

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LargeLongBloomFilter.java
```diff
@@ -0,0 +1,191 @@
...
+
+        logger.debug("Initializing Bloom filter with max partition size " + maxPartitionBytes + " bytes, " + numElements + " elements, and false positive probability " + fpp);
+
+        //Number of total bytes we'd like the set of filters to use
+        final long idealTotalBytes = LongBloomFilter.getOptimalNumberOfBytes(numElements, fpp);
+
+        //Largest legal size given max partition size
+        final int maxLegalPartitionBytes = LongBloomFilter.getLegalSizeBelow(maxPartitionBytes);
+
+        //Number of partitions, making sure to apply ceiling so that the actual false positive prob >= fpp
```
mwalker174: Noted

PR #2730, opened 2017-05-17
Add PathSeq BuildReferenceTaxonomy tool

mwalker174:
This tool creates a data file containing a map from reference accessions to NCBI taxonomic IDs, and the taxonomic tree, which includes parent/child relationships as well as other metadata like the reference length and scientific name of each node. 

The input files are available from the NCBI FTP server. One is a "catalog" file that gives the mapping from reference contig accession to taxonomic ID. There are catalog files available for RefSeq and for Genbank - the tool can take in either. 

There are two other files - a "names" and "nodes" file contained in a single tarball - that contain the scientific names of each node and parent/child relationships. For convenience, the tool takes in the path to the tarball and extracts the two files automatically.

The resulting database size is minimized using the given reference. Once the full NCBI taxonomy tree is built, any organism node that is neither in the reference nor an ancestor of a reference organism is removed.

The resulting datafile is read in by the ClassifyReads tool (coming in a future PR) to assign relative abundance scores to each taxonomic node.

Also made some changed to the way the PSTree and PSTreeNode are serialized (using Kryo read/writeObject instead of read/writeClassAndObject) so that loading old files won't break if these classes change packages.


src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBuildReferenceTaxonomy.java
```diff
@@ -0,0 +1,111 @@
...
+        logger.info("Building taxonomic database...");
+        final PSTree tree = PSBuildReferenceTaxonomyUtils.buildTaxonomicTree(taxIdToProperties);
+        final Map<String, String> accessionToTaxId = PSBuildReferenceTaxonomyUtils.buildAccessionToTaxIdMap(taxIdToProperties);
+
+        //Write output
+        PSBuildReferenceTaxonomyUtils.writeTaxonomyDatabase(outputPath, new PSTaxonomyDatabase(tree, accessionToTaxId));
+    }
+
+    @Override
+    public void traverse() {}
```
cwhelan: It seems a little silly to have this be a `GATKTool` with an empty `traverse` method. Would it be hard to make it into a `CommandLineProgram` instead? The only thing I can see that you're using from `GATKTool` is the reference initialization, which I think is only a line of code.

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBuildReferenceTaxonomyIntegrationTest.java
```diff
@@ -0,0 +1,104 @@
...
+        args.addFileArgument("refseqCatalogPath", catalogFile);
+        args.addFileArgument("taxdumpPath", taxdumpFile);
+        args.addOutput(outputFile);
+        this.runCommandLine(args.getArgsArray());
+
+        try {
+            final Kryo kryo = new Kryo();
+            kryo.setReferences(false);
+            Input input = new Input(new FileInputStream(expectedFile));
+            final PSTaxonomyDatabase db_expected = kryo.readObject(input, PSTaxonomyDatabase.class);
```
cwhelan: camelCase

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTaxonomyDatabase.java
```diff
@@ -1,16 +1,66 @@
...
-    public final Map<String, String> contigToTaxIDMap; //Reference contig name to taxonomic ID
+    public final Map<String, String> accessionToTaxId; //Reference contig name to taxonomic ID
 
     public PSTaxonomyDatabase(final PSTree tree, final Map<String, String> map) {
         this.tree = tree;
-        this.contigToTaxIDMap = map;
+        this.accessionToTaxId = map;
+    }
+
+    private PSTaxonomyDatabase(final Kryo kryo, final Input input) {
```
cwhelan: Since this is a pretty complicated serialization/deserialization, it'd be nice to write a unit test that exercises a round trip of this object with a small tree attached.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -0,0 +1,318 @@
...
+     * Input can be a RefSeq or Genbank catalog file. accNotFound is an initial list of
+     * accessions from the reference that have not been successfully looked up; if null,
+     * will be initialized to the accToRefInfo key set by default.
+     * <p>
+     * Returns a collection of reference accessions that could not be found, if any.
+     */
+    protected static Set<String> parseCatalog(final BufferedReader reader,
+                                              final Map<String, Tuple2<String, Long>> accessionToNameAndLength,
+                                              final Map<String, PSPathogenReferenceTaxonProperties> taxIdToProperties,
+                                              final boolean bGenBank, Set<String> accessionsNotFound) {
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -0,0 +1,318 @@
...
+        try {
+            String line;
+            final int taxIdColumnIndex, accessionColumnIndex;
+            if (bGenBank) {
+                taxIdColumnIndex = 6;
```
mwalker174: Unfortunately these files don't have any headers

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -0,0 +1,318 @@
...
+                lineNumber++;
+            }
+        } catch (final IOException e) {
+            throw new UserException.CouldNotReadInputFile("Error reading from catalog file", e);
+        }
+        return accessionsNotFound;
+    }
+
+    /**
+     * Parses scientific name of each taxon
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
mwalker174: Added an interface AccessionCatalogFormat with getTaxIdColumn() and getAccessionColumn() methods. Classes RefSeqCatalogFormat and GenBankCatalogFormat implement this interface for the two types of catalogs.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -0,0 +1,318 @@
...
+     */
+    protected static void parseNamesFile(final BufferedReader reader, final Map<String, PSPathogenReferenceTaxonProperties> taxIdToProperties) {
+        try {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                final String[] tokens = line.split("\t\\|\t");
```
mwalker174: The regex is looking for <tab>|<tab> as the delimiter. The double backslash is required for literals, otherwise Java thinks I'm trying to use an escape character called "\|" and throws an error.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /**
+     * Returns a PSTree representing a reduced taxonomic tree containing only taxa present in the reference
+     */
+    protected static PSTree buildTaxonomicTree(final Map<String, PSPathogenReferenceTaxonProperties> taxIdToProperties) {
+
+        //Build tree of all taxa
+        final PSTree tree = new PSTree(ROOT_ID);
+        final Collection<String> invalidIds = new ArrayList<>();
```
mwalker174: Changed to a hash set. I'm not sure exactly how many it is typically. I will just initialize it to the total number of ID's since this is not huge (~10,000's).

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -0,0 +1,318 @@
...
+        PSUtils.logItemizedWarning(logger, invalidIds, "The following taxonomic IDs did not have name/taxonomy information (this may happen when the catalog and taxdump files are inconsistent)");
+
+        final Set<String> unreachableNodes = tree.removeUnreachableNodes();
+        if (!unreachableNodes.isEmpty()) {
+            PSUtils.logItemizedWarning(logger, unreachableNodes, "Removed " + unreachableNodes.size() + " unreachable tree nodes");
+        }
+
+        tree.checkStructure();
+
+        //Trim tree down to nodes corresponding only to reference taxa (and their ancestors)
```
mwalker174: I don't think so.. for each node that has a reference, you add it and all its ancestors to relevantNodes. If you do that while constructing the tree, you're not guaranteed that all its ancestors have been added to the tree yet.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -0,0 +1,318 @@
...
+        } catch (final IOException e) {
+            throw new UserException.BadInput("Could not open compressed tarball file " + fileName + " in " + tarPath, e);
+        }
+    }
+
+    /**
+     * Wrapper for closing a reader
+     * @param reader The reader to close
+     */
+    public static void closeReader(final Reader reader) {
```
mwalker174: Done. Never used a try-with-resources block before!

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBuildReferenceTaxonomy.java
mwalker174: Done

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBuildReferenceTaxonomyIntegrationTest.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSTaxonomyDatabase.java
mwalker174: Added test class PSTaxonomyDatabaseTest with serialize/deserialize test

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
cwhelan: I see, makes sense.

PR #2731, opened 2017-05-18
Supporting empty loci emission for Spark tools

LeeTL1220:
Introducing the IntervalLocusIterator which will traverse every locus in intervals, regardless of coverage.

Minor changes

Removed imports.

AlignmentContextLocusIterator first cut.  Still needs unit tests.

Putting in the walker.  Still needs unit tests

Adding tests (and fixes) so that we can get AlignmentContexts.

Adding tests (and fixes) so that we can get AlignmentContexts.

Working tests.  Beginning migration to a LocusWalker change rather than a separate walker.

Merging the emit empty loci into locus walker.  Still need warnings and validation of parameters.

Next step is the LocusWalker testing.

Simple test of the new LocusWalker when it emit empty loci

Addressing PR requests and added ShardedIntervalIterator to save RAM on big intervals.

Addressing the rest of the PR comments

Rolling back to int from long.

Addressing second round of PR comments.

Wrapped LIBS in a factory so that we can encapsulate the retrieval of the best alignment context iterator.

Spark empty loci traversal being supported.

Rebasing based off of the other emit loci branch.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
@@ -9,18 +9,14 @@ import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.filters.WellformedReadFilter;
 import org.broadinstitute.hellbender.exceptions.UserException;
-import org.broadinstitute.hellbender.utils.IntervalUtils;
+import org.broadinstitute.hellbender.utils.locusiterator.AlignmentContextIteratorBuilder;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
-import org.broadinstitute.hellbender.utils.iterators.IntervalAlignmentContextIterator;
-import org.broadinstitute.hellbender.utils.iterators.IntervalLocusIterator;
-import org.broadinstitute.hellbender.utils.iterators.IntervalOverlappingIterator;
 import org.broadinstitute.hellbender.utils.locusiterator.LIBSDownsamplingInfo;
 import org.broadinstitute.hellbender.utils.locusiterator.LocusIteratorByState;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
 
 import java.util.*;
 import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
 /**
  * A LocusWalker is a tool that processes reads that overlap a single position in a reference at a time from
@@ -159,40 +155,26 @@ public abstract class LocusWalker extends GATKTool {
         final CountingReadFilter countedFilter = makeReadFilter();
         // get the filter and transformed iterator
         final Iterator<GATKRead> readIterator = getTransformedReadStream(countedFilter).iterator();
-        // get the LIBS
-        final LocusIteratorByState libs = new LocusIteratorByState(readIterator, getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
-        final Iterator<AlignmentContext> iterator = createAlignmentContextIterator(header, libs);
+
+        final AlignmentContextIteratorBuilder alignmentContextIteratorBuilder = new AlignmentContextIteratorBuilder();
+        alignmentContextIteratorBuilder.setDownsamplingInfo(getDownsamplingInfo());
+        alignmentContextIteratorBuilder.setEmitEmptyLoci(emitEmptyLoci());
+        alignmentContextIteratorBuilder.setIncludeDeletions(includeDeletions());
+        alignmentContextIteratorBuilder.setKeepUniqueReadListInLibs(keepUniqueReadListInLibs());
+        alignmentContextIteratorBuilder.setIncludeNs(includeNs());
+
+        final Iterator<AlignmentContext> iterator = alignmentContextIteratorBuilder.build(
+                readIterator, header, intervalsForTraversal, getBestAvailableSequenceDictionary(),
+                hasReference());
 
         // iterate over each alignment, and apply the function
-        final Spliterator<AlignmentContext> spliterator = Spliterators.spliteratorUnknownSize(iterator, 0);
-        StreamSupport.stream(spliterator, false)
-            .forEach(alignmentContext -> {
+        iterator.forEachRemaining(alignmentContext -> {
                         final SimpleInterval alignmentInterval = new SimpleInterval(alignmentContext);
                         apply(alignmentContext, new ReferenceContext(reference, alignmentInterval), new FeatureContext(features, alignmentInterval));
                         progressMeter.update(alignmentInterval);
                 }
             );
         logger.info(countedFilter.getSummaryLine());
-    }
-
-    private Iterator<AlignmentContext> createAlignmentContextIterator(SAMFileHeader header, LocusIteratorByState libs) {
-        Iterator<AlignmentContext> iterator;
-
-        validateEmitEmptyLociParameters();
-        if (emitEmptyLoci()) {
-
-            // If no intervals were specified, then use the entire reference (or best available sequence dictionary).
-            if (intervalsForTraversal == null) {
-                intervalsForTraversal = IntervalUtils.getAllIntervalsForReference(getBestAvailableSequenceDictionary());
-            }
-            final IntervalLocusIterator intervalLocusIterator = new IntervalLocusIterator(intervalsForTraversal.iterator());
-            iterator =  new IntervalAlignmentContextIterator(libs, intervalLocusIterator, header.getSequenceDictionary());
-
-        } else {
-            // prepare the iterator
-            iterator = (hasIntervals()) ? new IntervalOverlappingIterator<>(libs, intervalsForTraversal, header.getSequenceDictionary()) : libs;
-        }
-        return iterator;
     }
 
     /**
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/LocusWalkerSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/LocusWalkerSpark.java
@@ -1,8 +1,6 @@
 package org.broadinstitute.hellbender.engine.spark;
 
-import com.google.common.collect.ImmutableList;
 import htsjdk.samtools.SAMFileHeader;
-import htsjdk.samtools.SAMReadGroupRecord;
 import htsjdk.samtools.SAMSequenceDictionary;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.JavaSparkContext;
@@ -14,7 +12,7 @@ import org.broadinstitute.hellbender.engine.*;
 import org.broadinstitute.hellbender.engine.datasources.ReferenceMultiSource;
 import org.broadinstitute.hellbender.utils.IntervalUtils;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
-import org.broadinstitute.hellbender.utils.iterators.IntervalOverlappingIterator;
+import org.broadinstitute.hellbender.utils.locusiterator.AlignmentContextIteratorBuilder;
 import org.broadinstitute.hellbender.utils.locusiterator.LIBSDownsamplingInfo;
 import org.broadinstitute.hellbender.utils.locusiterator.LocusIteratorByState;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
@@ -66,6 +64,20 @@ public abstract class LocusWalkerSpark extends GATKSparkTool {
     }
 
     /**
+     * Does this tool emit information for uncovered loci? Tools that do should override to return {@code true}.
+     *
+     * NOTE:  Typically, this should only be used when intervals are specified.
+     * NOTE:  If MappedReadFilter is removed, then emitting empty loci will fail.
+     * NOTE:  If there is no available sequence dictionary and this is set to true, there should be a failure.  Please
+     *  consider requiring reads and/or references for all tools that wish to set this to {@code true}.
+     *
+     * @return {@code true} if this tool requires uncovered loci information to be emitted, {@code false} otherwise
+     */
+    public boolean emitEmptyLoci() {
+        return false;
+    }
+
+    /**
      * Loads alignments and the corresponding reference and features into a {@link JavaRDD} for the intervals specified.
      *
      * If no intervals were specified, returns all the alignments.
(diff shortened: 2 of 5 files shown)


LeeTL1220:
@droazen Feel free to assign to someone else.  If I have not heard from you in a little bit, I'll ask someone else.

src/main/java/org/broadinstitute/hellbender/utils/iterators/AlignmentContextIteratorFactory.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param downsamplingInfo how to downsample (for {@link LocusIteratorByState})
+     * @param reference the reference being used for this traversal, {@code null} if no reference being used.
+     * @param emitEmptyLoci whether loci with no coverage should be emitted.  In this case, the AlignmentContext will be empty (not null).
+     * @param isKeepUniqueReadListInLibs if true, we will keep the unique reads from the samIterator and make them
+     *                                       available via the transferReadsFromAllPreviousPileups interface (this parameter is specific to {@link LocusIteratorByState})
+     * @param isIncludeDeletions include reads with deletion on the loci in question
+     * @param isIncludeNs include reads with N on the loci in question
+     * @return Spliterator that produces AlignmentContexts ready for consumption (e.g. by a {@link org.broadinstitute.hellbender.engine.LocusWalker})
+     */
+    public static Iterator<AlignmentContext> createAlignmentContextIterator(final List<SimpleInterval> intervalsForTraversal,
```
magicDGS: If you are using a factory pattern, I would include the following fields and setters:

* downsamplingInfo/setDownsamplingInfo
* emitEmptyLoci/setEmitEmptyLoci
* keepUniqueReadList/setKeepUniqueReadList
* includeDeletetions/setIncludeDeletions
* includeNs/setIncludeNs
* dictionary/setSequenceDictionary - __Note: maybe this one is not a good idea, but we always expect to be traversing the same genome and thus the sequence dictionary should be the same...

Then, the method will be simplified by using:

```java
public Iterator<AlignmentContext> createAlignmentContextIterator(final Iterator<GATKRead> reads, final SAMFileHeader header, final List<SimpleInterval> intervalsForTraversal) {
   // the rest of the code will be the same, but using the fields
    ...
}
```

src/main/java/org/broadinstitute/hellbender/engine/spark/LocusWalkerSpark.java
```diff
@@ -105,13 +120,11 @@ protected final LIBSDownsamplingInfo getDownsamplingInfo() {
...
-            final Set<String> samples = header.getReadGroups().stream()
-                    .map(SAMReadGroupRecord::getSample)
-                    .collect(Collectors.toSet());
-            LocusIteratorByState libs = new LocusIteratorByState(readIterator, downsamplingInfo, false, samples, header, true, false);
-            IntervalOverlappingIterator<AlignmentContext> alignmentContexts = new IntervalOverlappingIterator<>(libs, ImmutableList.of(interval), sequenceDictionary);
-            final Spliterator<AlignmentContext> alignmentContextSpliterator = Spliterators.spliteratorUnknownSize(alignmentContexts, 0);
-            return StreamSupport.stream(alignmentContextSpliterator, false).map(alignmentContext -> {
+            final SAMSequenceDictionary referenceDictionary = reference == null? null : reference.getSequenceDictionary();
+
+            final Iterator<AlignmentContext> alignmentContextIterator = AlignmentContextIteratorFactory.createAlignmentContextIterator(Collections.<SimpleInterval>singletonList(interval), header, readIterator, sequenceDictionary,
```
magicDGS: If you make the factory pattern as I suggested, this could be much simpler and you can construct the factory just once keeping the state.

src/main/java/org/broadinstitute/hellbender/utils/iterators/AlignmentContextIteratorFactory.java
```diff
@@ -0,0 +1,113 @@
+package org.broadinstitute.hellbender.utils.iterators;
```
magicDGS: I will make this live in the `org.broadinstitute.hellbender.utils.locusiterator` package.

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
```diff
@@ -23,7 +24,15 @@
...
-    public IntervalAlignmentContextIterator(final Iterator<AlignmentContext> alignmentContextIterator, final IntervalLocusIterator intervalLocusIterator, final SAMSequenceDictionary dictionary) {
+
+    /**
+     *  Note:  Typically, if you are calling this from a walker tool, you want to use {@link AlignmentContextIteratorFactory}
+     *
+     * @param alignmentContextIterator iterator for alignment context
+     * @param intervalLocusIterator an iterator that will traverse all relevant loci
+     * @param dictionary reference or best available dictionary.
+     */
+    public IntervalAlignmentContextIterator(Iterator<AlignmentContext> alignmentContextIterator, IntervalLocusIterator intervalLocusIterator, SAMSequenceDictionary dictionary) {
```
magicDGS: I will make this live in the `org.broadinstitute.hellbender.utils.locusiterator` with the factory, and make package-protected. Reduce the visibility will avoid incorrect usages and you can remove the note.

It could also live as an static class inside the factory, but I always prefer not to implement nested classes...

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -162,14 +162,14 @@ public void traverse() {
         final CountingReadFilter countedFilter = makeReadFilter();
         // get the filter and transformed iterator
         final Iterator<GATKRead> readIterator = getTransformedReadStream(countedFilter).iterator();
-        // get the LIBS
-        final LocusIteratorByState libs = new LocusIteratorByState(readIterator, getDownsamplingInfo(), keepUniqueReadListInLibs(), samples, header, includeDeletions(), includeNs());
-        final Iterator<AlignmentContext> iterator = createAlignmentContextIterator(header, libs);
+
+        Iterator<AlignmentContext> iterator = AlignmentContextIteratorFactory.createAlignmentContextIterator(
```
magicDGS: This will be much more readable if you construct the factory and set the values and then creates the alignment context.

src/main/java/org/broadinstitute/hellbender/utils/iterators/AlignmentContextIteratorFactory.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/iterators/AlignmentContextIteratorFactory.java
LeeTL1220: This gives me default values for the creation, but I can get that with overloading.  If I am keeping a state that can be changed, it would be a ``Builder``, right?

I'm open to the idea, but need a little more convincing.

LeeTL1220: Actually, I just convinced myself.  The argument is that I can set a bunch of default parameters and don't have a combinatorial explosion of overloaded functions to set some and not others.

magicDGS: In fact, it is a builder pattern... sorry for the confusion. If you convince yourself, perfect! I also think that it will make the code clear while constructing the iterator and make possible to construct several iterators for different shards without the need of using again the full constructor, which makes things clear too.

Thanks for taking into account my comments!

LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/engine/spark/LocusWalkerSpark.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIterator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
LeeTL1220: Done

jonn-smith:
This looks good to me.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Mon May 01 15:13:13 PDT 2017
+#Wed May 10 13:25:04 EDT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.1-all.zip
```
jonn-smith: Is this intentional / will it break anything?

@lbergelson What do you think?

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/AlignmentContextIteratorBuilder.java
```diff
@@ -0,0 +1,178 @@
...
+            // If no intervals were specified, then use the entire reference (or best available sequence dictionary).
+            if (!hasIntervals(finalIntervals)) {
+                finalIntervals = IntervalUtils.getAllIntervalsForReference(dictionary);
+            }
+            final IntervalLocusIterator intervalLocusIterator = new IntervalLocusIterator(finalIntervals.iterator());
+            iterator =  new IntervalAlignmentContextIterator(libs, intervalLocusIterator, header.getSequenceDictionary());
+
+        } else {
+            // prepare the iterator
+            iterator = hasIntervals(finalIntervals) ? new IntervalOverlappingIterator<>(libs, finalIntervals, header.getSequenceDictionary()) : libs;
```
magicDGS: This will be clear if the hasIntervals condition is included in the if-else clause as following:

```java
if (emitEmptyLoci) {
    ...
} else if (hasIntervals(finalIntervals)) {
    ...
} else {
    iterator = libs;
}
```

Like that, it also can avoid the creation of `Iterator<AlignmentContext> iterator` by returning in each if-else branch the generated iterator...

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/AlignmentContextIteratorBuilder.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        return iterator;
+    }
+
+    private static boolean hasIntervals(final List<SimpleInterval> finalIntervals) {
+        return finalIntervals != null;
```
magicDGS: This means that if the intervals are empty, the iterator has intervals...is that what we want or should it account for an empty list too?

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/AlignmentContextIteratorBuilder.java
LeeTL1220: Done

src/test/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIteratorUnitTest.java
```diff
@@ -10,6 +10,7 @@
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.io.IOUtils;
+import org.broadinstitute.hellbender.utils.locusiterator.IntervalAlignmentContextIterator;
```
LeeTL1220: Done.  Forgot to move it the first time around.

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/AlignmentContextIteratorBuilder.java
```diff
@@ -0,0 +1,178 @@
...
+    private boolean isKeepUniqueReadListInLibs;
+    private boolean isIncludeDeletions;
+    private boolean isIncludeNs;
+    private LIBSDownsamplingInfo downsamplingInfo;
+
+    public boolean isEmitEmptyLoci() {
+        return isEmitEmptyLoci;
+    }
+
+    public void setEmitEmptyLoci(boolean emitEmptyLoci) {
```
LeeTL1220: Ugh... I expect setters to always return void by convention.  Maybe my time in J2EE-land is biasing me.
Alternatively, we could add ``withAttribute(...)`` and that would do the setting and return ``this``

No action.

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/AlignmentContextIteratorBuilder.java
LeeTL1220: I have renamed hasIntervals to areIntervalsSpecified.  Since I would not want that returning false for an empty list.

Done.

src/test/java/org/broadinstitute/hellbender/utils/iterators/IntervalAlignmentContextIteratorUnitTest.java
LeeTL1220: I made IntervalAlignmentContextIterator package private

gradle/wrapper/gradle-wrapper.properties
LeeTL1220: I reverted this.  Just to be safe.

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/AlignmentContextIteratorBuilder.java
magicDGS: I also expect setters to return void, but I thought that the builder pattern is usually returning `this` as a convention. Anyway, it was a suggestion to use the builder with chaining and not making a variable to call the setters in...

LeeTL1220: We can put ``withAttribute(...)`` in a future PR.

PR #2736, opened 2017-05-19
updating gathervcfs to work with paths

lbergelson:
This only works for the conventional gather.  adding useConventionalGather argument to force using conventional gather.

adding --ignoreSafetyChecks to skip pre-validating the headers

continuing the tradition of having no tests for this tool

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/GatherVcfs.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/GatherVcfs.java
@@ -12,12 +12,13 @@ import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.PeekableIterator;
 import htsjdk.samtools.util.RuntimeIOException;
 import htsjdk.tribble.AbstractFeatureReader;
+import htsjdk.tribble.FeatureReader;
 import htsjdk.variant.variantcontext.VariantContext;
 import htsjdk.variant.variantcontext.VariantContextComparator;
 import htsjdk.variant.variantcontext.writer.Options;
 import htsjdk.variant.variantcontext.writer.VariantContextWriter;
 import htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder;
-import htsjdk.variant.vcf.VCFFileReader;
+import htsjdk.variant.vcf.VCFCodec;
 import htsjdk.variant.vcf.VCFHeader;
 
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/GatherVcfs.java
```diff
@@ -56,11 +57,22 @@
...
             shortName = StandardArgumentDefinitions.INPUT_SHORT_NAME,  doc = "Input VCF file(s).")
-	public List<File> INPUT;
+	public List<String> inputs;
 
     @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
             shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "Output VCF file.")
-	public File OUTPUT;
+	public File output;
+
+    // default value of 40MB based on a test with CountReads (it's 5x faster than no prefetching)
```
droazen: Comment about default value of 40 MB does not match the tool default

src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/GatherVcfs.java
```diff
@@ -71,56 +83,82 @@ public GatherVcfs() {
...
+            if (pathString == null || pathString.endsWith(".bcf") || !AbstractFeatureReader.hasBlockCompressedExtension(pathString)){
                 return false;}
         }
         return true;
     }
 
+    private static FeatureReader<VariantContext> getReaderFromVCFUri(final Path variantPath, final int cloudPrefetchBuffer) {
+        final String variantURI = variantPath.toUri().toString();
+        final Function<SeekableByteChannel, SeekableByteChannel> cloudWrapper = (cloudPrefetchBuffer > 0 ? is -> SeekableByteChannelPrefetcher
+                .addPrefetcher(cloudPrefetchBuffer, is) : Function.identity());
```
droazen: No wrapper needed for the index?

lbergelson: we don't look at the indexes so we don't want to preload them i think.

src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/GatherVcfs.java
```diff
@@ -152,50 +194,58 @@ private static void assertSameSamplesAndValidOrdering(final List<File> inputFile
...
 
-                while (variantIterator.hasNext()) {
-                    lastContext = variantIterator.next();
-                    out.add(lastContext);
-                    progress.record(lastContext.getContig(), lastContext.getStart());
-                }
+                    if (lastContext != null && variantIterator.hasNext()) {
+                        final VariantContext vc = variantIterator.peek();
+                        if (comparator.compare(vc, lastContext) <= 0) {
+                            throw new IllegalStateException("First variant in file " + f.toUri().toString() + " is at " + vc.getSource() +
```
droazen: Interesting that you do this check even with `--ignoreSafetyChecks true` :)

lbergelson: @droazen Do you think we should have --ignoreSafetyChecks ignore this too?

droazen: I wasn't suggesting that, just pointing out the inconsistency of this arg.

lbergelson: I can rename to --ignoreExpensiveSafetyChecks....

src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/GatherVcfs.java
lbergelson: fixed

PR #2740, opened 2017-05-22
Cache the codec class used for a FeatureInput so we only have to discover it once.

cmnbroad:
Propose to reduce redundantly cracking open a path/stream to discover the correct feature codec. We do this twice for each feature input, which for multi-variant walkers with large # of inputs can be a lot. This caches the codec class in a FeatureInout the first time we find it. Ideally FeatureManager would remember it, but not all of the FeatureDataSources are created by Feature Manager (and fixing that is a bigger refactoring).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
@@ -288,11 +288,37 @@ public final class FeatureDataSource<T extends Feature> implements GATKDataSourc
                 throw new UserException("GenomicsDB inputs can only be used to provide VariantContexts.", e);
             }
         } else {
-            final Path featurePath = IOUtils.getPath(featureInput.getFeaturePath());
-            IOUtils.assertFileIsReadable(featurePath);
-            final FeatureCodec<T, ?> codec = (FeatureCodec<T, ?>) FeatureManager.getCodecForFile(featurePath, targetFeatureType);
+            FeatureCodec<T, ?> codec = getCodecForFeatureInput(featureInput, targetFeatureType);
             return getTribbleFeatureReader(featureInput, codec, cloudWrapper, cloudIndexWrapper);
         }
+    }
+
+    /**
+     * Get a new FeatureCodec instance to use for a FeatureInput. Avoid re-discovering which codec class to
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/engine/FeatureInput.java
```diff
@@ -60,6 +58,11 @@
     private final String featureFile;
 
     /**
+     * Cache the codec for this feature input the first time we discover it, so we only do it once
+     */
+    private transient Class<FeatureCodec<T, ?>> featureCodecClass;
```
droazen: Let me take a look to see if it would be possible to do this at the `FeatureDataSource` level rather than the `FeatureInput` level.

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
```diff
@@ -288,9 +288,21 @@ public static boolean isGenomicsDBPath(final String path) {
...
+                codec = (FeatureCodec<T, ?>) FeatureManager.getCodecForFile(featurePath, targetFeatureType);
+                featureInput.setFeatureCodecClass((Class<FeatureCodec<T, ?>>)codec.getClass());
+            } else {
+                try {
+                    codec = codecClass.newInstance();
+                }
+                catch ( InstantiationException | IllegalAccessException e ) {
+                    throw new GATKException("Unable to automatically instantiate codec " + codecClass.getName());
+                }
+            }
```
droazen: Can you extract a helper method here that takes a `FeatureInput` and returns a codec?

src/main/java/org/broadinstitute/hellbender/engine/FeatureInput.java
```diff
@@ -216,6 +219,21 @@ public FeatureInput(final String featureFile, final String name, final Map<Strin
...
+    }
+
+    /**
+     * @return The previously established FeatureCodec class to use for this input, if any. May return {@code null}.
+     */
+    public Class<FeatureCodec<T, ?>> getFeatureCodecClass() {
+        return this.featureCodecClass;
+    }
+
+    /**
```
droazen: Can you add a unit test for these new methods?

Issue #2685, opened 2017-05-09
NIO SSL Exception in GenomicsDBImport


Horneth:
I was running a WDL task with GenomicsDBImport followed by GenotypeGVCFs.
It failed with the following stderr:

```
[May 9, 2017 4:30:53 PM UTC] org.broadinstitute.hellbender.tools.genomicsdb.GenomicsDBImport  --genomicsDBWorkspace genomicsdb --batchSize 300 --sampleNameMap inputs_candidate.list --intervals chr1:820849-1000848  --genomicsDBSegmentSize 1048576 --genomicsDBVCFBufferSize 16384 --overwriteExistingGenomicsDBWorkspace false --consolidate false --interval_set_rule UNION --interval_padding 0 --interval_exclusion_padding 0 --readValidationStringency SILENT --secondsBetweenProgressUpdates 10.0 --disableSequenceDictionaryValidation false --createOutputBamIndex true --createOutputBamMD5 false --createOutputVariantIndex true --createOutputVariantMD5 false --lenient false --addOutputSAMProgramRecord true --cloudPrefetchBuffer 2 --cloudIndexPrefetchBuffer 2 --disableBamIndexCaching false --help false --version false --showHidden false --verbosity INFO --QUIET false --use_jdk_deflater false --use_jdk_inflater false --disableToolDefaultReadFilters false
[May 9, 2017 4:30:53 PM UTC] Executing as root@f045553cd221 on Linux 3.16.0-0.bpo.4-amd64 amd64; OpenJDK 64-Bit Server VM 1.8.0_121-8u121-b13-1~bpo8+1-b13; Version: Version:e1ffbff4-SNAPSHOT
Created workspace /cromwell_root/genomicsdb
[May 9, 2017 4:50:01 PM UTC] org.broadinstitute.hellbender.tools.genomicsdb.GenomicsDBImport done. Elapsed time: 19.15 minutes.
Runtime.totalMemory()=9498001408
java.lang.RuntimeException: java.util.concurrent.ExecutionException: com.google.cloud.storage.StorageException: Remote host closed connection during handshake
	at org.broadinstitute.hellbender.utils.nio.SeekableByteChannelPrefetcher.read(SeekableByteChannelPrefetcher.java:309)
	at htsjdk.samtools.seekablestream.SeekablePathStream.read(SeekablePathStream.java:86)
	at java.io.BufferedInputStream.fill(BufferedInputStream.java:246)
[...]

PR #2750, opened 2017-05-24
Increase retries and backoff

jean-philippe-martin:
A very simple and easy to review PR that backs off on retries a bit more, and retries for a bit longer. This may help when we're opening many cloud files in parallel.


droazen:
:+1: looks good to me -- merging

jean-philippe-martin:
Great, thank you! I believe we still need the changes described in https://github.com/broadinstitute/gatk/issues/2685#issuecomment-302798685 but this may help relieve some of the pressure.

Issue #2728, opened 2017-05-17
Add the ability log to file


ronlevine:
Replicate the behavior in [GATK3 ](https://software.broadinstitute.org/gatk/documentation/tooldocs/current/org_broadinstitute_gatk_engine_CommandLineGATK.php#--log_to_file).
May also require a script that runs a cron job that clean up "old" logs.

PR #2751, opened 2017-05-25
Add ability to log to file

ronlevine:
Implements #2728.
This is not a direct port from GATK3 since the log4j version in GATK4 (2.x) has a different architecture than the one in GATK3 (1.x).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -65,6 +65,9 @@ public abstract class CommandLineProgram implements CommandLinePluginProvider {
     @Argument(fullName = StandardArgumentDefinitions.VERBOSITY_NAME, shortName = StandardArgumentDefinitions.VERBOSITY_NAME, doc = "Control verbosity of logging.", common = true, optional = true)
     public Log.LogLevel VERBOSITY = Log.LogLevel.INFO;
 
+    @Argument(fullName = "log_to_file", shortName = "log", doc = "Set the logging location", common = true, optional = true)
+    protected String logToFile = null;
+
     @Argument(doc = "Whether to suppress job-summary info on System.err.", common=true)
     public Boolean QUIET = false;
 
@@ -130,6 +133,8 @@ public abstract class CommandLineProgram implements CommandLinePluginProvider {
         this.defaultHeaders.add(new StringHeader(commandLine));
         this.defaultHeaders.add(new StringHeader("Started on: " +
                                 startDateTime.format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.LONG))));
+
+        LoggingUtils.sendLogToFile(logToFile);
 
         LoggingUtils.setLoggingLevel(VERBOSITY);  // propagate the VERBOSITY level to logging frameworks
 
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -156,18 +156,21 @@ public abstract class CommandLineProgram implements CommandLinePluginProvider {
         }
 
         if (!QUIET) {
-            System.err.println("[" + ZonedDateTime.now().format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.LONG)) +
-                                "] " + commandLine);
+            final String zoneDateTimeMsg = "[" + ZonedDateTime.now().format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.LONG)) +
+                    "] " + commandLine;
+            System.err.println(zoneDateTimeMsg);
+            logger.trace(zoneDateTimeMsg);
 
             // Output a one liner about who/where and what software/os we're running on
             try {
-                System.err.println("[" + ZonedDateTime.now().format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.LONG)) +
+                final String softwareOsMsg = "[" + ZonedDateTime.now().format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.LONG)) +
                         "] Executing as " +
                         System.getProperty("user.name") + "@" + InetAddress.getLocalHost().getHostName() +
                         " on " + System.getProperty("os.name") + " " + System.getProperty("os.version") +
                         " " + System.getProperty("os.arch") + "; " + System.getProperty("java.vm.name") +
-                        " " + System.getProperty("java.runtime.version") +
-                        "; Version: " + getCommandLineParser().getVersion());
+                        " " + System.getProperty("java.runtime.version");
+                System.err.println(softwareOsMsg);
+                logger.trace(zoneDateTimeMsg);
 
                 // Print important settings to the logger:
                 printSettings();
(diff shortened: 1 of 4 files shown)


ronlevine:
@lbergelson Please review.

ronlevine:
@cmnbroad Please review.

ronlevine:
Ping @lbergelson.

src/main/java/org/broadinstitute/hellbender/utils/LoggingUtils.java
```diff
@@ -87,6 +93,22 @@ private static void setJavaUtilLoggingLevel(final Log.LogLevel verbosity) {
         return javaUtilLevelNamespaceMap.get(picardLevel);
     }
 
+    /**
+     * Send log output to file if file name is not null.
+     *
+     * @param fileName  File path for log output
+     */
+    public static void sendLogToFile(final String fileName) {
```
lbergelson: 1.  I would make this expect a non-null filename and have the null-check be at the call site I think.
2. If you call this repeatedly with multiple files can you direct the log to multiple files?  
3. Does this suppress log messages on the console?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -65,6 +65,9 @@
     @Argument(fullName = StandardArgumentDefinitions.VERBOSITY_NAME, shortName = StandardArgumentDefinitions.VERBOSITY_NAME, doc = "Control verbosity of logging.", common = true, optional = true)
     public Log.LogLevel VERBOSITY = Log.LogLevel.INFO;
 
+    @Argument(fullName = "log_to_file", shortName = "log", doc = "Set the logging location", common = true, optional = true)
```
lbergelson: I might just call both of these "logFile"

src/main/java/org/broadinstitute/hellbender/utils/LoggingUtils.java
```diff
@@ -87,6 +93,22 @@ private static void setJavaUtilLoggingLevel(final Log.LogLevel verbosity) {
...
+        if (fileName != null) {
+            final LoggerContext loggerContext = (LoggerContext) LogManager.getContext(false);
```
lbergelson: This cast is weird but I don't see how to avoid it.

lbergelson: The file should be created here if it doesn't exist because you get nasty error messages otherwise.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -131,6 +134,8 @@ public Object instanceMainPostParseArgs() {
         this.defaultHeaders.add(new StringHeader("Started on: " +
                                 startDateTime.format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.LONG))));
 
+        LoggingUtils.sendLogToFile(logToFile);
+
         LoggingUtils.setLoggingLevel(VERBOSITY);  // propagate the VERBOSITY level to logging frameworks
 
         for (final File f : TMP_DIR) {
```
lbergelson: There are a few messages that are output that are sent directly to stdout here which should also be included in a log file.  

ex:
```
[July 10, 2017 10:49:51 AM EDT] org.broadinstitute.hellbender.tools.PrintReads  --output output.bam --input src/test/resources/NA12878.chr17_69k_70k.dictFix.bam  --interval_set_rule UNION --interval_padd>
[July 10, 2017 10:49:51 AM EDT] Executing as louisb@WMD2A-31E on Mac OS X 10.11.6 x86_64; Java HotSpot(TM) 64-Bit Server VM 1.8.0_112-b16; Version: Version:removeNativeBiuld-380-g4267ce3-SNAPSHOT
[July 10, 2017 10:49:52 AM EDT] org.broadinstitute.hellbender.tools.PrintReads done. Elapsed time: 0.01 minutes.
Runtime.totalMemory()=192937984
```

src/main/java/org/broadinstitute/hellbender/utils/LoggingUtils.java
```diff
@@ -87,6 +93,22 @@ private static void setJavaUtilLoggingLevel(final Log.LogLevel verbosity) {
...
+            final Configuration loggerContextConfig = loggerContext.getConfiguration();
+            final Layout<?> layout = PatternLayout.createLayout(PATTERN_STRING, null, loggerContextConfig,
+                    null, null, false, false, null, null);
+            final FileAppender appender = FileAppender.createAppender(fileName, "", "", fileName, "",
```
lbergelson: Do we want append behavior or overwrite?  I would expect overwrite based on the behavior of other gatk arguments, but I'm not sure how the log file is usually used.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
ronlevine: Yep. Was going to stderr.

src/main/java/org/broadinstitute/hellbender/utils/LoggingUtils.java
ronlevine: Agreed. I wrestled with that too but the code "pinned" me (pun intended).

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
ronlevine: I was using the GATK3 names. The other argument names in the file use the same convention.

lbergelson: Ah, that makes sense.  Keep them then I guess.  We can revisit them if we ever do the global naming pass.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
lbergelson: So I think we want to keep that message going to stderr, because we want it output always even with restrictive logging modes.  We want to make sure it also is duplicated in the log file though.

src/main/java/org/broadinstitute/hellbender/utils/LoggingUtils.java
ronlevine: `FileAppender.createAppender`,  which is a couple of lines down, creates the file. So, that should not be an issue.

src/main/java/org/broadinstitute/hellbender/utils/LoggingUtils.java
ronlevine: The second argument, append, should be changed to false, the default is true.

src/main/java/org/broadinstitute/hellbender/utils/LoggingUtils.java
ronlevine: 1. I think it's safer to do the non-null check in sendLogToFile. If the application programmer forgets to do the check, there will  be an error
2. Yes. Added a test.
3. No.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
ronlevine: The stderr messages will also got to the log file.

ronlevine:
I added the ability to send log4J and Java utility logs to file. Unfortunately, [Esotericsoft's MinLog](https://github.com/EsotericSoftware/minlog),  that is used by kryo, [only has the ability to write to stdout](https://github.com/EsotericSoftware/minlog/blob/master/src/com/esotericsoftware/minlog/Log.java#L218).

magicDGS:
@ronlevine, I think that the following should work for output the Esotericsoft's MinLog to write to a file:
* Implement in `LoggerUtils` a class extending `com.esotericsoftware.minlog.Log.Logger` that overrides the print method to append to the provided file.
* Use`Log.setLogger()` with an instance of that class.

ronlevine:
@magicDGS Esotericsoftware's minLog is used by Esotericsoftware's kryo. So, they would have to extend the class. :sad:

lbergelson:
@magicDGS Would this nightmare go away on it's own if we used slfj like you suggested a while ago?

magicDGS:
@lbergelson: I don't think so, for that they should use https://github.com/jdanbrown/minlog-slf4j and I do not think that they are using it.

@ronlevine: I think that using `com.esotericsoftware.minlog.Log.setLogger()` with a GATK's implementation of `com.esotericsoftware.minlog.Log.Logger` will solve the problem. Actuallty, maybe it's worthy to set up this directly on startup to a custom implementation using the log4j logger to print the log, although that requires to override the `Logger.log()` method to log in the correct level.

ronlevine:
@magicDGS I think you're right about using the `com.esotericsoftware.minlog.Log.setLogger()` with the file logger that overrides `com.esotericsoftware.minlog.Log.print`.

cmnbroad:
@ronlevine @lbergelson Have you guys thought about the Spark case - I don't think this will capture logging output from Spark workers without propagating unless it gets propagated through the config (we also don't propagate the verbosity IIRC). It might be a little misleading to allow specification of a file for Spark tools if it doesn't capture all of the output.

ronlevine:
Need to modify `htsjdk.samtools.util.Log` so it can log to file, not just stderr.

droazen:
Closing in favor of https://github.com/broadinstitute/gatk/issues/3587

PR #2752, opened 2017-05-25
Update license copyright date

vdauwera:
straightforward change -- date was out of date

also there was a missing period


LICENSE.TXT
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2009-2015, GATK Authors. All rights reserved.
+Copyright (c) 2009-2017, GATK Authors. All rights reserved.
```
droazen: Same question here as in https://github.com/broadinstitute/gatk-protected/pull/1099 -- should the copyright holder be "Broad Institute, Inc.", or is "GATK Authors" correct?

Issue #2753, opened 2017-05-25
Contig ordering in VCF header is scrambled by either GenomicsDB, GenomicsDBImport, or GenotypeGVCFs


droazen:
Broad production is finding that after running HG38 GVCFs through `GenomicsDBImport`, `GenomicsDB`, and then `GenotypeGVCFs`, the sort order of the contigs in the header is scrambled. Specifically, the contig names appear to be sorted by ASCII value, so, eg., "chr10" comes before "chr9". 

This is a problem, since there are tools that use the contig ordering in the header to generate contig indices used for comparison of relative contig ordering. `GatherVcfs` in Picard is one example.

We need to identify which of the three possible suspects above (`GenomicsDBImport`, `GenomicsDB`, or `GenotypeGVCFs`) is clobbering the contig ordering, and patch it quickly.

droazen:
@kgururaj @kdatta Can you guys comment on this? Do you think it's possible that `GenomicsDB` is failing to preserve the ordering of the contig header lines? We need a fix for this soon, as it's causing a bit of havoc in our pipeline.

droazen:
@lbergelson please also comment on whether it's possible that `GenotypeGVCFs` itself could be responsible.

droazen:
Another possible culprit here is the `VCFUtils.smartMergeHeaders()` call in `GenomicsDBImport` when running without a sample name map file (which is how production was running, in this case) -- @cmnbroad can you comment on this possibility?

ronlevine:
@droazen I fixed a similar issue in https://github.com/samtools/htsjdk/pull/848. The header from the input VCF or dictionary is put into a `TreeMap`, which stores the data in natural order. Since the key is a String, it's lexicographical. It should be stored in a `LinkedHashMap`, which maintains the order.

ronlevine:
@droazen `VCFUtils.smartMergeHeaders()` was fixed in https://github.com/samtools/htsjdk/pull/848.

cmnbroad:
@droazen Yeah, what @ronlevine mentioned could possibly do it, depending on where the data flows to next, i.e., if the header lines are then put back into a header, the contig line list in the new header should be correct (the big metadata list would still be off) though.

cmnbroad:
Hm. It looks like GDB importer merges, then creates a header from the resulting lines, then retrieves a sequence dictionary, which doesn't correct the order. So that may be it.

cmnbroad:
[...]

PR #2755, opened 2017-05-25
Add a --disableContigOrderingCheck argument to GatherVcfs to disable the comparison of relative contig ordering during a conventional gather

droazen:
With this argument, we still check start positions to make sure they're increasing within a given contig
during a conventional gather, but we don't enforce relative ordering of contigs using the ordering of
the sequence dictionary in the VCF header.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/GatherVcfs.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/GatherVcfs.java
@@ -73,6 +73,9 @@ public final class GatherVcfs extends PicardCommandLineProgram {
             "This is necessary when using NIO paths")
     public boolean useConventionalGather = false;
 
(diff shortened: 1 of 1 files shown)


droazen:
@lbergelson please review. This is a quick-and-dirty temporary fix to work around https://github.com/broadinstitute/gatk/issues/2753, and if we can fix the underlying problem quickly this may not be worth even merging.

lbergelson:
I disapprove of this gross duplication, but since it's running in production I think it's best to merge this thing anyway for now.

ronlevine:
IIRC, this use of this feature was discouraged in GATK3 . Right @vdauwera? Making a new release of  HTSJDK and incorporating it should fix the underlying problem.

droazen:
I want this argument as a fallback for now in case production runs into more trouble with the upcoming callset. Merging in so @lbergelson can incorporate this change in his upcoming patch to `GatherVcfs`

PR #2758, opened 2017-05-25
ChangedSelectVariantsDoc

chandrans:
Explanation of GATK4 SelectVariants

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
@@ -49,13 +49,12 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 /**
- * Select a subset of variants from a larger callset
+ * Select a subset of variants from a VCF file
  *
- * <p>
- * Often, a VCF containing many samples and/or variants will need to be subset in order to facilitate certain analyses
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
```diff
@@ -84,6 +84,16 @@
  * A new VCF file containing the selected subset of variants.
  * </p>
  *
+ * * <h3>Usage example</h3>
+ * <pre>
+ *     ./gatk-launch \
+ *     SelectVariants \
+ *     -R reference.fasta \
```
cmnbroad: GATK4 doesn't generally require that a reference be provided except in the case where the tool specifically requires it, or the input or output is a CRAM), so I'm wondering if we should remove these from the examples.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariants.java
```diff
@@ -84,6 +84,16 @@
...
+ *     -V input.vcf \
+ *     -selectType SNP \
+ *     -o output.vcf
```
sooheelee: `-o` to `-O`?

vdauwera:
can be merged once tests pass

PR #2764, opened 2017-05-25
ChangedLeftAlignIndelsDoc

chandrans:
For GATK4 docs


sooheelee:
It may be good to mention that HaplotypeCaller already left-aligns and (I am assuming) there would be no need to run this tool on HaplotypeCaller outputs.

PR #2766, opened 2017-05-26
partition crossing

tedsharpe:


tedsharpe:
Master stats:
Discovered 31120 intervals.
Killed 388 intervals that were near reference gaps.
Killed 174 intervals that had >1000x coverage.
Discovered 9480784 mapped template names.
Ignoring 19200460 genomically common kmers.
Discovered 39739968 kmers.
Discovered 34170333 unique template names for assembly.
Wrote SAM file of aligned contigs.
Discovered 6255 variants.
INV: 239
DEL: 3644
DUP: 1123
INS: 1249
Elapsed time: 47.34 minutes.

This PR:
Discovered 31125 intervals.
Killed 390 intervals that were near reference gaps.
Killed 174 intervals that had >1000x coverage.
Discovered 9480874 mapped template names.
Ignoring 19200460 genomically common kmers.
Discovered 39730495 kmers.
Discovered 34154214 unique template names for assembly.
Wrote SAM file of aligned contigs.
Discovered 6234 variants.
INV: 233
DEL: 3635
DUP: 1125
INS: 1241
Elapsed time: 46.77 minutes.

We did find a few extra intervals by gluing evidence across partition boundaries.  The number of variants detected has decreased by just a little.  I think this is likely due to calculating read metadata at the library level, rather than the read group level.

cwhelan:
Just curious, if you run it multiple times do the results still change slightly? I was wondering if fixing this would get rid of some or all of our run-to-run fluctuations.

tedsharpe:
I haven't tried it, but I will.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
```diff
@@ -39,39 +39,60 @@ public ReadMetadata( final Set<Integer> crossContigIgnoreSet,
...
                      .collect();
-        nPartitions = perPartitionStatistics.size();
         nReads = perPartitionStatistics.stream().mapToLong(PartitionStatistics::getNReads).sum();
         maxReadsInPartition = perPartitionStatistics.stream().mapToLong(PartitionStatistics::getNReads).max().orElse(0L);
         final long nReadBases = perPartitionStatistics.stream().mapToLong(PartitionStatistics::getNBases).sum();
         final long nRefBases = header.getSequenceDictionary().getSequences()
                 .stream().mapToLong(SAMSequenceRecord::getSequenceLength).sum();
-        coverage = (int) ((nReadBases + nRefBases - 1) / nRefBases);
-        readGroupToFragmentStatistics = new HashMap<>(SVUtils.hashMapCapacity(header.getReadGroups().size()));
+        coverage = (int) ((nReadBases + nRefBases/2) / nRefBases);
```
cwhelan: Why this change to the coverage calculation? Can you put in a comment explaining the calculation?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
```diff
@@ -254,49 +299,72 @@ public ReadMetadata read( final Kryo kryo, final Input input, final Class<ReadMe
...
+                lastContig = null;
+                lastLocation = 0;
+            } else {
+                long reads = 0L;
+                long bases = 0L;
+                GATKRead read = readItr.next();
+                firstContig = read.getContig();
+                firstLocation = read.getStart();
+                libraryToFragmentSizeCountMap = new HashMap<>(SVUtils.hashMapCapacity(readGroupToLibraryMap.size()));
+                while ( true ) {
```
cwhelan: why not just `while ( readItr.hasNext() )`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -184,15 +254,7 @@ protected void serialize( final Kryo kryo, final Output output ) {
...
-            final StringBuffer out = new StringBuffer();
-            out.append(super.toString());
-            out.append("\t");
-            out.append("Split");
-            out.append("\t");
-            out.append(cigar);
-            out.append("\t");
-            out.append((tagSA == null ? " SA: None" : (" SA: " + tagSA)));
-            return out.toString();
+            return super.toString()+"\tSplit\t"+cigar+"\t"+(tagSA == null ? " SA: None" : (" SA: " + tagSA));
```
tedsharpe: Mindless consistency.  All the other classes implement it this way.
As far as efficiency goes, they are completely equivalent, I believe.  So it's a matter of clarity.  To my eye the concatenation is a little clearer, but YMMV.  Happy to change it back if you think the StringBuffer way is clearer.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidenceClusterer.java
```diff
@@ -0,0 +1,55 @@
...
+        final SVInterval interval = evidence.getLocation();
+        final int weight = evidence.getWeight();
+        if ( curInterval == null ) {
+            curInterval = interval;
+            curWeight = weight;
+        } else if ( curInterval.gapLen(interval) < gapSize ) {
+            curInterval = curInterval.join(interval);
+            curWeight += weight;
+        } else {
+            result = new SingletonIterator<>(new BreakpointEvidence(curInterval,curWeight,true));
```
tedsharpe: I feel pretty confident about the coherent evidence check.  As long as the regions near partition boundaries are large enough we ought to get the same results evaluating supporting evidence (coherent or not) across the boundaries as we do within a partition.
Can't think of any edge conditions, but that doesn't mean that there aren't any.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/FlatMapGluer.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.function.Function;
+
+public class FlatMapGluer<I,O> implements Iterator<O> {
+    private final Function<I,Iterator<O>> flatMapFunc;
+    private final Iterator<? extends I> inputIterator;
+    private I sentinel;
```
tedsharpe: Actually, any subtype of I can be passed as an I (because it _IsA_ I).  So your example of a special subclass will work fine.
It's just the type of the collection that needs to be covariant (because an Iterator<J> is not in an _IsA_ relationship with Iterator<I>, even if J extends I).
Java generics are big, big fun!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
tedsharpe: I was just rounding instead of truncating.  Comment has been added.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
tedsharpe: Because I've already grabbed the first read from the iterator (to get the firstContig and firstLocation data), but I still have to account for its reads, bases, and tLen.  So I have to enter the loop once, even when there isn't an additional read to process.  I could express it as a do/while, but I'm not sure it would be any clearer:
```
    GATKRead read = mappedReadItr.next();
    firstContig = read.getContig();
    firstLocation = read.getStart();
    do {
        a bunch of stuff
    } while ( (read = (readItr.hasNext() ? readItr.next() : null)) != null );
```
Do you prefer the do/while form?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointDensityFilter.java
```diff
@@ -0,0 +1,131 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import org.broadinstitute.hellbender.utils.Utils;
+
+import java.util.*;
+
+/**
+ * A class that acts as a filter for breakpoint evidence.
+ * It passes only that evidence that is part of a putative cluster.
+ */
+public final class BreakpointDensityFilter implements Iterator<BreakpointEvidence> {
```
tedsharpe: You have a point.  Filtering is what it does, not what it is.  On the other hand, BreakpointDensityMapPartitioner states what it is, but obscures what it does.  I think I'm gonna punt.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVUtils.java
```diff
@@ -122,13 +121,24 @@ public static void writeIntervalsFile( final String intervalsFile,
...
-                writer.write(seqName + "\t" + (interval.getStart()+1) + "\t" + interval.getEnd() + "\n");
+                writer.write(seqName + "\t" + interval.getStart() + "\t" + interval.getEnd() + "\n");
             }
         } catch (final IOException ioe) {
             throw new GATKException("Can't write intervals file " + intervalsFile, ioe);
         }
     }
 
+    public static int matchLen( final Cigar cigar ) {
+        int len = 0;
```
tedsharpe: Yes.  Nice.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/PartitionEdges.java
```diff
@@ -0,0 +1,58 @@
+package org.broadinstitute.hellbender.tools.spark.sv;
+
+public class PartitionEdges {
```
tedsharpe: It's now a PartitionCrossingChecker.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
```diff
@@ -28,9 +28,9 @@
     private final Map<String, Integer> contigNameToID;
     private final long nReads;
     private final long maxReadsInPartition;
-    private final int nPartitions;
     private final int coverage;
-    private final Map<String, ReadGroupFragmentStatistics> readGroupToFragmentStatistics;
+    private final PartitionBounds[] partitionBounds;
+    private final Map<String, LibraryFragmentStatistics> readGroupToFragmentStatistics;
```
tedsharpe: The fragment length stats are still looked up by read group.  It's just that they're accumulated by library, and then the same set of statistics is shared over all the groups derived from that library.
I'll say that in the doc.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/BreakpointEvidence.java
```diff
@@ -119,39 +57,171 @@ public boolean hasDistalTargets() {
...
+            private final String value;
+        }
+
+        /**
+         * evidence offset and width is set to "the rest of the fragment" not covered by this read
+         */
+        protected ReadEvidence( final GATKRead read, final ReadMetadata metadata ) {
+            super(restOfFragmentInterval(read,metadata), 1, false);
+            this.templateName = read.getName();
+            if ( templateName == null ) throw new GATKException("Read has no name.");
```
tedsharpe: The call to ```super``` has to be the first line.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/PartitionEdges.java
```diff
@@ -0,0 +1,58 @@
...
+    public boolean onEdge( final SVInterval interval ) {
+        return onLeadingEdge(interval) || onTrailingEdge(interval);
+    }
+
+    public boolean onLeadingEdge( final SVInterval interval ) {
+        return interval.getContig() == beginningContigID && interval.getStart() < beginningPosition;
+    }
+
+    public boolean onTrailingEdge( final SVInterval interval ) {
+        return interval.getContig() == endingContigID && interval.getStart() >= endingPosition;
```
tedsharpe: I've drawn some pictures and rethought this, and I think that you're right.  It should be interval.getEnd().

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/FlatMapGluer.java
cwhelan: Yes, of course you're right, sorry. Got myself turned around.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadMetadata.java
cwhelan: No, it's ok the way it is. I understand now.

tedsharpe:
Comments addressed.  Thanks for your input.  Back to @cwhelan for any further thoughts.

PR #2767, opened 2017-05-26
GenomicsDBImport: preserve ordering of contig lines in GVCF headers from the inputs

### LINKED ISSUE #2753: Contig ordering in VCF header is scrambled by either GenomicsDB, GenomicsDBImport, or GenotypeGVCFs

droazen:
This is fixed by switching to the latest HTSJDK snapshot, which contains a
fix for VCFUtils.smartMergeHeaders() to preserve ordering.

Add a test that fails without the new HTSJDK snapshot and passes with it.

Resolves #2753


droazen:
@lbergelson please review

Issue #963, opened 2015-10-07
Make GCS authentication work without using dataflow options classes


droazen:

droazen:
If possible! This would potentially allow us to remove the entire `com.google.cloud.genomics.dataflow` dependency.

droazen:
Affects https://github.com/broadinstitute/hellbender/issues/954

davidadamsphd:
This has partially been handled by JP already :) with his `AuthHolder` class. The only missing piece is removing our dependance on `GCSOptions.Methods.createGCSAuth(options)` by instead calling the create auth stuff directly.

droazen:
This one is definitely for beta

droazen:
This one is almost done -- assigning to @jean-philippe-martin to do the last remaining piece.

droazen:
@jean-philippe-martin Can we close this one?

jean-philippe-martin:
I think the question is about `com.google.cloud.dataflow.sdk.options.PipelineOptions`. In that case we're not done, there is still some of that about.

jean-philippe-martin:
I've removed some in #2786. The rest are all related to the `ReferenceAPISource`.

PR #2786, opened 2017-05-31
Remove some uses of dataflow.sdk.options.PipelineOptions

jean-philippe-martin:
Those options are used to carry auth for the ReferenceAPISource,
so we can't remove all instances without changing those.

This work is part of #963.


jean-philippe-martin:
The Travis failure looks like a network transient: it's complaining it can't download something from Maven. Trying again.

jean-philippe-martin:
Awesome, thank you!

PR #2788, opened 2017-05-31
fix bug in ElementIterator

tedsharpe:


tedsharpe:
@mwalker174 I've chosen you as a victim because:
1)  you know more about the hopscotch algorithm than anyone else
2)  i just spent a mess of time on one of your PRs and turnabout is fair play
sorry for the imposition.  it's a very small change, and i hope it doesn't take long to review.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchCollection.java
```diff
@@ -562,6 +557,25 @@ public T next() {
...
+
+        @Override
+        public void remove() {
+            super.remove();
+            if ( currentIndex != NO_ELEMENT_INDEX ) {
+                checkEquivalence();
+            }
+        }
+
+        private void checkEquivalence() {
```
cwhelan: I'm always a little wary about methods with names like `checkXXX` that actually have side effects, like in this case advancing the `currentIndex` pointer. Maybe something like `ensureEquivalence` would be better? Alternatively, could you just have `advanceUntilEquivalent` contain the initial check for equivalence and just use that?

src/test/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchMultiMapTest.java
```diff
@@ -80,4 +80,20 @@ void removeEachTest() {
         }
         Assert.assertEquals(hopscotchMultiMap.size(), 0);
     }
+
+    // found a bug in removeEach when multiple keys hash to the same bucket
+    // (non-equivalent keys were getting deleted)
+    // this test demonstrates the fix
+    @Test
+    void removeEachBugTest() {
```
cwhelan: Nice simple little test to catch a tricky bug.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/HopscotchCollection.java
tedsharpe: Good idea.

SHuang-Broad: I agree with Chris on this one.
What I've observed (noticing that this is my first time looking at this code):
  * `checkEquivalence()` essentially adds a `while` check that also appears in `advanceUntilEquivalent()`, which is its delegate, so `checkEquivalence()` could be removed with a re-structured `advanceUntilEquivalent()`
  * `advanceUntilEquivalent()` could be&mdash;but not necessary&mdash;renamed `advanceUntilEquivalentOrExhausted()`

tedsharpe: Can't eliminate either method without doing a redundant equivalence test.
ensureEquivalence is effectively: ```while ( !equivalent ) advance;```
advanceUntilEquivalent is: ```do advance; while ( !equivalent );```
If I called ensureEquivalence in the next() method rather than advanceUntilEquivalent I'd be doing an extra equivalence test on the current key, which is already known to be equivalent.
Maybe an extra key compare is trivial, but maybe it's not -- you'd need to know something about the key.
I did the rename to ensureEquivalence, which does seem like a better name.

PR #2789, opened 2017-05-31
ChangedPileupDoc

chandrans:
Minor changes to doc.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/Pileup.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/Pileup.java
@@ -5,6 +5,7 @@ import htsjdk.tribble.Feature;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.broadinstitute.barclay.argparser.Hidden;
+import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 import org.broadinstitute.hellbender.cmdline.programgroups.QCProgramGroup;
 import org.broadinstitute.hellbender.engine.*;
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/Pileup.java
```diff
@@ -26,13 +26,13 @@
  *
  * <p>This tool emulates the 'samtools pileup' command. It prints the alignment in a format that is very similar to the
  * Samtools pileup format (see the <a href="http://samtools.sourceforge.net/pileup.shtml">Pileup format
- * documentation</a> for more details about the original format). There is one line per genomic position, listing the
+ * documentation</a> for more details about the original format). There is one column per genomic position, listing the
```
magicDGS: I think that each line is a genomic position, and then each column shows the information for the list in the following line. I do not really know how to phrase it better, but the current one looks like each column is a genomic position and the first row is the chromosome name, the second the coordinate and so on (transposed format).

sooheelee: It would be great to include an example snippet of the output.

sooheelee:
I find it funny we show the samtools command that our tool emulates....

src/main/java/org/broadinstitute/hellbender/tools/walkers/qc/Pileup.java
vdauwera: there is an example snippet lower down in the file

vdauwera: And yes it is one line per position, not one column. Fixed.

vdauwera:
@droazen Not sure why we're getting a build failure here, can you or @lbergelson please take a look?

vdauwera:
added missing documented feature annotation

vdauwera:
@sooheelee Your comment about the missing doc feature annotation in another PR clued me in to the nature of the problem in this branch, thank you

PR #2800, opened 2017-06-01
ApplyBQSRDoc

chandrans:
Added documentation to ApplyBQSR doc.


sooheelee:
For the next iteration perhaps mention that you can omit inclusion of the OQ tags with XYZ parameter to save space as base qualities do not compress well.

vdauwera:
This doc currently starts with several side notes before the main tool description, and the first line (which becomes the summary description) is actually two sentences. All this breaks the pattern we've established for the other tools.

vdauwera:
I fixed the main description, separated out the notes and improved the argument docs while I was at it

PR #2803, opened 2017-06-02
Check for bad sequence dictionaries in ReferenceUtils.loadFastaDictionary

### LINKED ISSUE #2609: ReferenceUtils.loadFastaDictionary should throw a UserException if given an invalid .dict file.

jonn-smith:
Now throws a UserException.MalformedFile if the resulting header has no
dictionary.

Added testLoadFastaDictionaryWithFastaFile to test this case.

resolves #2609

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtils.java
@@ -73,10 +73,10 @@ public final class ReferenceUtils {
         final SAMFileHeader header = codec.decode(reader, fastaDictionaryStream.toString());
 
         // Make sure we have a valid sequence dictionary before continuing:
-        if (header.getSequenceDictionary().isEmpty()) {
+        if (header.getSequenceDictionary() == null || header.getSequenceDictionary().isEmpty()) {
             throw new UserException.MalformedFile (
                     "Could not read sequence dictionary from given fasta stream " +
-                            fastaDictionaryStream.toString()
+                            fastaDictionaryStream
             );
         }
 
--- a/src/main/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtils.java
@@ -49,6 +49,12 @@ public final class ReferenceUtils {
         catch ( IOException e ) {
             throw new UserException.CouldNotReadInputFile("Error loading fasta dictionary file " + fastaDictionaryFile, e);
         }
+        catch ( UserException.MalformedFile e ) {
+            throw new UserException.MalformedFile(
+                    "Could not read sequence dictionary from given fasta file " +
+                            fastaDictionaryFile
+            );
+        }
     }
 
     /**
@@ -65,6 +71,15 @@ public final class ReferenceUtils {
 
         final SAMTextHeaderCodec codec = new SAMTextHeaderCodec();
         final SAMFileHeader header = codec.decode(reader, fastaDictionaryStream.toString());
+
+        // Make sure we have a valid sequence dictionary before continuing:
+        if (header.getSequenceDictionary().isEmpty()) {
+            throw new UserException.MalformedFile (
+                    "Could not read sequence dictionary from given fasta stream " +
+                            fastaDictionaryStream.toString()
+            );
+        }
+
         return header.getSequenceDictionary();
     }
 }
--- a/src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
@@ -7,6 +7,10 @@ import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
 import org.broadinstitute.hellbender.utils.test.BaseTest;
 import org.testng.Assert;
 import org.testng.annotations.Test;
+import org.testng.annotations.DataProvider;
+
+import java.lang.reflect.Method;
+import java.util.*;
 
 import java.io.*;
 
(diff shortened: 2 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtils.java
```diff
@@ -65,6 +71,15 @@ public static SAMSequenceDictionary loadFastaDictionary( final InputStream fasta
 
         final SAMTextHeaderCodec codec = new SAMTextHeaderCodec();
         final SAMFileHeader header = codec.decode(reader, fastaDictionaryStream.toString());
+
+        // Make sure we have a valid sequence dictionary before continuing:
+        if (header.getSequenceDictionary().isEmpty()) {
```
droazen: Protect against null as well: `if (header.getSequenceDictionary() == null || header.getSequenceDictionary().isEmpty())`

src/main/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtils.java
```diff
@@ -49,6 +49,12 @@ public static SAMSequenceDictionary loadFastaDictionary( final File fastaDiction
         catch ( IOException e ) {
             throw new UserException.CouldNotReadInputFile("Error loading fasta dictionary file " + fastaDictionaryFile, e);
         }
+        catch ( UserException.MalformedFile e ) {
+            throw new UserException.MalformedFile(
```
droazen: A little weird, but I guess worth doing to increase the specificity of the error message.

src/main/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtils.java
```diff
@@ -65,6 +71,15 @@ public static SAMSequenceDictionary loadFastaDictionary( final InputStream fasta
 
         final SAMTextHeaderCodec codec = new SAMTextHeaderCodec();
         final SAMFileHeader header = codec.decode(reader, fastaDictionaryStream.toString());
+
+        // Make sure we have a valid sequence dictionary before continuing:
+        if (header.getSequenceDictionary().isEmpty()) {
+            throw new UserException.MalformedFile (
+                    "Could not read sequence dictionary from given fasta stream " +
+                            fastaDictionaryStream.toString()
```
droazen: Explicit `toString()` call unnecessary (Java invokes it for you)

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
```diff
@@ -52,6 +53,25 @@ public void testLoadFastaDictionaryFromStream() throws IOException {
         }
     }
 
+    @Test
+    public void testLoadFastaDictionaryWithFastaFile() throws IOException {
```
droazen: The name of this test case will be confusing to anyone not familiar with the original issue -- recommend renaming to something like `testLoadWrongFormatFastaDictionary()`

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
```diff
@@ -52,6 +53,25 @@ public void testLoadFastaDictionaryFromStream() throws IOException {
...
+        Assert.assertThrows(UserException.MalformedFile.class, () -> {
+            ReferenceUtils.loadFastaDictionary(testFastaFile);
+        });
+
+        // Test loadFastaDictionary with Stream argument:
+        try ( final ClosingAwareFileInputStream referenceDictionaryStream = new ClosingAwareFileInputStream(testFastaFile) ) {
+
+            Assert.assertThrows(UserException.MalformedFile.class, () -> {
+                ReferenceUtils.loadFastaDictionary(referenceDictionaryStream);
+            });
```
droazen: You should be using TestNG's `expectedExceptions` feature here. Exceptions you expect to be thrown should go in the `@Test` annotation itself: `(expectedExceptions = UserException.MalformedFile.class)`. That way you don't have to call `assertThrows()` yourself. 

You'll need to break this up into two test cases to do this -- you can use a `DataProvider` with two test cases that passes in a `File` plus a boolean flag indicating whether to call the stream version.

src/main/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtils.java
jonn-smith: Done!

src/main/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtils.java
jonn-smith: Removed!

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
jonn-smith: Renamed as suggested.

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
jonn-smith: Updated to use expectedExceptions and a DataProvider.

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
```diff
@@ -52,6 +57,34 @@ public void testLoadFastaDictionaryFromStream() throws IOException {
         }
     }
 
+    @DataProvider(name = "testData_testLoadWrongFormatFastaDictionary")
+    public Object[][] testData_testLoadWrongFormatFastaDictionary(final Method testMethod) {
```
droazen: Your `DataProvider` method should take no parameters.

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
```diff
@@ -52,6 +57,34 @@ public void testLoadFastaDictionaryFromStream() throws IOException {
...
+        Object[][] tests = {
+                {hg19MiniReference, true},
+                {hg19MiniReference, false}
+        };
+
+        return tests;
+    }
+
+    @Test(expectedExceptions = UserException.MalformedFile.class, dataProvider = "testData_testLoadWrongFormatFastaDictionary")
+    public void testLoadWrongFormatFastaDictionary(final String fastaFileName, boolean doLoadAsStream ) throws IOException {
```
droazen: `loadAsStream` is a bit less awkward than `doLoadAsStream`

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
droazen: Also, if you don't provide an explicit name for your `DataProvider`, it will default to using the method name as the `DataProvider` name.

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
```diff
@@ -52,6 +57,34 @@ public void testLoadFastaDictionaryFromStream() throws IOException {
         }
     }
 
+    @DataProvider(name = "testData_testLoadWrongFormatFastaDictionary")
+    public Object[][] testData_testLoadWrongFormatFastaDictionary(final Method testMethod) {
+
+        Object[][] tests = {
```
droazen: Add a comment explaining that we're trying to load a fasta file as a `dict` file to trigger a malformed file error.

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
```diff
@@ -52,6 +57,34 @@ public void testLoadFastaDictionaryFromStream() throws IOException {
...
+    public void testLoadWrongFormatFastaDictionary(final String fastaFileName, boolean doLoadAsStream ) throws IOException {
+
+        File testFastaFile = new File(fastaFileName);
+
+        if ( doLoadAsStream ) {
+            // Test loadFastaDictionary with Stream argument:
+            try ( final ClosingAwareFileInputStream referenceDictionaryStream = new ClosingAwareFileInputStream(testFastaFile) ) {
+                ReferenceUtils.loadFastaDictionary(referenceDictionaryStream);
+            }
+        }
```
lbergelson: We typically do braces like 
```
} else {
```
 rather than
``` 
}
 else  {
```

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
lbergelson: I would probably just inline this variable

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
lbergelson: David claims he writes with this style. So I guess do whatever you want.

droazen: Not me :)

src/test/java/org/broadinstitute/hellbender/utils/reference/ReferenceUtilsUnitTest.java
droazen: ie., `return new Object[][] {`

PR #2804, opened 2017-06-02
removing two redundant travis matrix entries to save vms

lbergelson:
removing the non-docker unit and integration test matrix entries because
they were redundant with the docker ones

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ jdk:
 env:
   matrix:
   - TEST_TYPE=cloud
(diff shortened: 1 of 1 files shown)


droazen:
@lbergelson Not sure we should merge this until we solve the intermittent timeout issue in the docker tests.

droazen:
Blocked by https://github.com/broadinstitute/gatk/issues/2817

lbergelson:
this was made redundant by #3353

PR #2807, opened 2017-06-02
(Ported from gatk-protected) Isolate a problematic change to ReferenceConfidenceModel for further discussion

droazen:
I've extracted this change to ReferenceConfidenceModel from the now-closed PR
https://github.com/broadinstitute/gatk-protected/pull/1022, since it causes tests
to fail spectacularly (including concordance tests against GATK3), so that we can
unblock the badly-needed merge of https://github.com/broadinstitute/gatk-protected/pull/1027

Let's review and test this change in isolation to be sure we understand it fully
before accepting it into master.


droazen:
Closing in favor of https://github.com/broadinstitute/gatk/pull/3519, which includes this fix.

PR #2811, opened 2017-06-03
CNV performallelefractionsegmentation: give example command

sooheelee:


sooheelee:
Could one of the Lee's please review?

LeeTL1220:
@sooheelee As per the other PR for Sam and me...  If @samuelklee approves, I approve too.

src/main/java/org/broadinstitute/hellbender/tools/exome/segmentation/PerformAlleleFractionSegmentation.java
```diff
@@ -18,7 +19,7 @@
  * Groups contiguous targets with the same minor allele fraction for a single sample.
  *
  *  <p>
- *     The --tumorHets file is from {@link GetBayesianHetCoverage}, preferably, but can also come from {@link org.broadinstitute.hellbender.tools.exome.GetHetCoverage}.
+ *     The --tumorHets file is from {@link org.broadinstitute.hellbender.tools.exome.GetBayesianHetCoverage}, preferably, but can also come from {@link org.broadinstitute.hellbender.tools.exome.GetHetCoverage}.
```
samuelklee: You can import `org.broadinstitute.hellbender.tools.exome` to clean up both javadoc links here.  I think you can set IntelliJ to do this automatically; e.g., https://stackoverflow.com/questions/24266917/how-to-make-intellij-autocompletion-not-insert-the-full-path-in-javadoc.

sooheelee: Thanks for this! I looked and it appears I am already following these preferences:

screenshot 2017-06-05 23 43 32

For some links, I don't get this appendage. For other links, I get the red font and then so I use the appendage. Not clear to me why there is a distinction between tools.

sooheelee: We can clean up the code by removing the appendages if these links appear funny via the javadoc/gatkDoc generation.

samuelklee: I think you might have to change the dropdown from "If not already imported" to "Never, use short name and add import" if you want the IntelliJ autocorrect to use the short name.

PR #2816, opened 2017-06-04
Document Mutect2: revamp documentation and example commands to reflect algorithmic changes

sooheelee:
I've completely rewritten the documentation portion to be more helpful to users (and to reflect the new M2 rather than gatk3's M2), and I have updated the example commands.

---
### Questions for @davidbenjamin 
- `--af_of_alleles_not_in_resource`: is this allele frequency used only in certain contexts, e.g. with matched normal analyses, or towards tumor sample variant alleles, etc.? I need to add to the doc details how this argument factors into calculations.
- I need a sentence or two describing the new algorithmic improvement on the new Mutect2 integration over uncertainty. 
- The WDLs do not include use of a contamination.table and so I did not include it in the commands. Is this something we want to nudge users to use, i.e. should I put in a sentence in the documentation section about the new tool CalculateContamination?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
@@ -20,78 +20,155 @@ import java.util.ArrayList;
 import java.util.List;
 
 /**
- * Call somatic SNPs and indels via local re-assembly of haplotypes
+ * Call somatic short variants, both SNVs and indels, via local assembly of haplotypes
  *
- * <p>MuTect2 is a somatic SNP and indel caller that combines the DREAM challenge-winning somatic genotyping engine of the original MuTect (<a href='http://www.nature.com/nbt/journal/v31/n3/full/nbt.2514.html'>Cibulskis et al., 2013</a>) with the assembly-based machinery of <a href="https://www.broadinstitute.org/gatk/documentation/tooldocs/org_broadinstitute_gatk_tools_walkers_haplotypecaller_HaplotypeCaller.php">HaplotypeCaller</a>.</p>
+ * <p>
+ *     Mutect2 calls somatic single nucleotide (SNV) and insertion and deletion (indel) variants.
+ *     The caller combines the DREAM challenge-winning somatic genotyping engine of the original MuTect
+ *     (<a href='http://www.nature.com/nbt/journal/v31/n3/full/nbt.2514.html'>Cibulskis et al., 2013</a>) with the
+ *     assembly-based machinery of <a href="https://www.broadinstitute.org/gatk/documentation/tooldocs/org_broadinstitute_gatk_tools_walkers_haplotypecaller_HaplotypeCaller.php">HaplotypeCaller</a>.
+ *     Although we present the tool for somatic analyses, it may also apply to other contexts.
+ * </p>
  *
- * <h3>How MuTect2 works</h3>
- * <p>The basic operation of MuTect2 proceeds similarly to that of the HaplotypeCaller, with a few key differences.</p>
- * <p>While the HaplotypeCaller relies on a ploidy assumption (diploid by default) to inform its genotype likelihood and variant quality calculations, MuTect2 allows for a varying allelic fraction for each variant, as is often seen in tumors with purity less than 100%, multiple subclones, and/or copy number variation (either local or aneuploidy). MuTect2 also differs from the HaplotypeCaller in that it does apply some hard filters to variants before producing output. Finally, some of the parameters used for ActiveRegion determination and graph assembly are set to different default values in MuTect2 compared to HaplotypeCaller.</p>
- * <p>Note that MuTect2 is designed to produce somatic variant calls only, and includes some logic to skip variant sites that are very clearly germline based on the evidence present in the Normal sample compared to the Tumor sample. This is done at an early stage to avoid spending computational resources on germline events. As a result the tool is NOT capable of emitting records for variant calls that are clearly germline unless it is run in artifact-detection mode, which is used for Panel-Of-Normals creation.</p>
- * <p>Note also that the GVCF generation capabilities of HaplotypeCaller are NOT available in MuTect2, even though some of the relevant arguments are listed below. There are currently no plans to make GVCF calling available in MuTect2.</p>
+ * <h3>How GATK4 Mutect2 differs from GATK3 MuTect2</h3>
  *
- * <h3>Usage examples</h3>
- * <p>These are example commands that show how to run Mutect2 for typical use cases. Square brackets ("[ ]")
- * indicate optional arguments. Note that parameter values shown here may not be the latest recommended; see the
- * Best Practices documentation for detailed recommendations. </p>
+ * <dl>
+ *     <dd>(i) The filtering functionality is now a separate tool called {@link FilterMutectCalls}.
+ *     To filter further based on sequence context artifacts, additionally use {@link FilterByOrientationBias}.</dd>
+ *     <dd>(ii) If using a known germline variants resource, then it must contain population allele frequencies, e.g.
+ *     from gnomAD or the 1000 Genomes Project. See below or the GATK Resource bundle for an example.</dd>
+ *     <dd>(iii) To create the panel of normals (PoN), call on normal samples as if they are tumor samples with Mutect2 and then use GATK4's {@link CreateSomaticPanelOfNormals}.
+ *     This contrasts with the GATK3 workflow, which uses an artifact mode in MuTect2 and CombineVariants for PoN creation.</dd>
+ * </dl>
  *
- * <br />
- * <h4>Tumor/Normal variant calling</h4>
+ * <p>
+ *     What remains unchanged is that neither tool versions call on seeming loss of heterozygosity (LoH) events.
+ *     To detect LoH, see the Copy Number Variant (CNV) and AllelicCNV workflows.
+ * </p>
+ *
+ * <p>Here is an example of a known variants resource with populaton allele frequencies:</p>
+ *
  * <pre>
- *   java
- *     -jar GenomeAnalysisTK.jar \
- *     -T Mutect2 \
- *     -R reference.fasta \
- *     -I tumor.bam \
- *     -I normal.bam \
- *     -tumor tumorSampleName \ // as in the BAM header
- *     -normal normalSampleName \ // as in the BAM header
- *     [--dbsnp dbSNP.vcf] \
- *     [--cosmic COSMIC.vcf] \
- *     [-L targets.interval_list] \
- *     -o output.vcf
+ *     #CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO
+ *      1       10067   .       T       TAACCCTAACCCTAACCCTAACCCTAACCCTAACCCTAACCC      30.35   PASS    AC=3;AF=7.384E-5
+ *      1       10108   .       CAACCCT C       46514.32        PASS    AC=6;AF=1.525E-4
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
```diff
@@ -20,78 +20,155 @@
...
+ * <h3>How GATK4 Mutect2 differs from GATK3 MuTect2</h3>
  *
- * <h3>Usage examples</h3>
- * <p>These are example commands that show how to run Mutect2 for typical use cases. Square brackets ("[ ]")
- * indicate optional arguments. Note that parameter values shown here may not be the latest recommended; see the
- * Best Practices documentation for detailed recommendations. </p>
+ * <dl>
+ *     <dd>(i) The filtering functionality is now a separate tool called {@link FilterMutectCalls}.
+ *     To filter further based on sequence context artifacts, additionally use {@link FilterByOrientationBias}.</dd>
+ *     <dd>(ii) If using a known germline variants resource, then it must contain population allele frequencies, e.g.
```
davidbenjamin: Maybe specify that population allele frequencies will be the `AF` INFO field in a vcf.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
```diff
@@ -20,78 +20,155 @@
...
+ *     from gnomAD or the 1000 Genomes Project. See below or the GATK Resource bundle for an example.</dd>
+ *     <dd>(iii) To create the panel of normals (PoN), call on normal samples as if they are tumor samples with Mutect2 and then use GATK4's {@link CreateSomaticPanelOfNormals}.
```
davidbenjamin: Perhaps specify that the normals for the PoN should be called in tumor-only mode.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
```diff
@@ -20,78 +20,155 @@
...
+ *     This contrasts with the GATK3 workflow, which uses an artifact mode in MuTect2 and CombineVariants for PoN creation.</dd>
```
davidbenjamin: Perhaps note that in GATK4, the artifact mode is achieved by skipping `FilterMutectCalls`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
```diff
@@ -20,78 +20,155 @@
...
+ * </dl>
  *
- * <br />
- * <h4>Tumor/Normal variant calling</h4>
+ * <p>
+ *     What remains unchanged is that neither tool versions call on seeming loss of heterozygosity (LoH) events.
+ *     To detect LoH, see the Copy Number Variant (CNV) and AllelicCNV workflows.
+ * </p>
+ *
+ * <p>Here is an example of a known variants resource with populaton allele frequencies:</p>
```
davidbenjamin: typo: populat[i]on

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
```diff
@@ -20,78 +20,155 @@
...
+ * <p>Overall, Mutect2 works similarly to HaplotypeCaller, but with a few key differences. </p>
+ * <p>
+ *     (i) GVCF calling is not a feature of Mutect2.
+ *     (ii) While HaplotypeCaller relies on a fixed ploidy assumption to inform its genotype likelihoods that are the basis for genotype probabilities (PL),
+ *     Mutect2 allows for varying ploidy in the form of allele fractions for each variant.
+ *     Varying allele fractions is often seen within a tumor sample due to fractional purity, multiple subclones and/or copy number variation.
+ *     (iii) Mutect2 also differs from the HaplotypeCaller in that it can apply various prefilters to sites and variants depending on the use of
+ *     a matched normal (--normalSampleName), a panel of normals (PoN; --normal_panel) and/or a common population variant resource containing allele-specific frequencies (--germline_resource).
+ *     If provided, Mutect2 uses the PoN to filter sites and the germline resource and matched normal to filter alleles.
+ *     (iv) Mutect2 variant site annotations differ from that of HaplotypeCaller. See the --annotation parameter description for a list.
```
davidbenjamin: differ from *those* of HaplotpyeCaller

davidbenjamin: Rather, Mutect's *default* annotations are different, but in principle any `Annotation` can be used with both.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
```diff
@@ -20,78 +20,155 @@
...
+ *     (v) Finally, Mutect2 has additional parameters not available to HaplotypeCaller that factor in the decision to reassemble a genomic region,
+ *     factor in likelihood calculations that then determine whether to emit a variant, or factor towards filtering.
+ *     These parameters include the following and are each described further in the arguments section.
+ * </p>
+ *
+ * <dl>
+ *     <dd>--min_variants_in_pileup ==> active region determination</dd>
+ *     <dd>--minNormalVariantFraction ==> active region determination</dd>
+ *     <dd>--tumorStandardDeviationsThreshold ==> active region determination</dd>
```
davidbenjamin: Just FYI, we realized recently that M2's active region determination, while better than GATK3, still has a very long way to go.  We're going to blow up many of these parameters soon.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
```diff
@@ -20,78 +20,155 @@
...
+ * </p>
+ *
+ * <h3>Examples</h3>
+ *
+ * <p>Example commands show how to run Mutect2 for typical scenerios.</p>
+ *
+ * <h4>Tumor with matched normal</h4>
+ * <p>
+ *     Given a matched normal, Mutect2 is designed to call somatic variants only. The tool includes logic to skip
+ *     emitting variants that are clearly germline based on the evidence present in the matched normal. This is done at
```
davidbenjamin: We on the GATK team are insiders who use "germline" as an adjective, but for the standard user I might use more explicit language like "present in the germline".

davidbenjamin:
>--af_of_alleles_not_in_resource: is this allele frequency used only in certain contexts, e.g. with matched normal analyses, or towards tumor sample variant alleles, etc.? I need to add to the doc details how this argument factors into calculations.

This is the population AF assigned to a variant not found in the germline resource and is inversely proportional to the number of samples used to create that resource.  For example, gnomAD contains about 16,000 wgs samples, or 32,000 homologs per locus.  Thus the AF of an allele not found in gnomAD is probably 1/32,000 or less.  This is useful because it lets us use absence from the germline resource as evidence that an allele is *not* a germline variant.

>I need a sentence or two describing the new algorithmic improvement on the new Mutect2 integration over uncertainty.

Since allele depths (ADs) are subject to statistical error -- i.e. the exact number of reads for an allele is a random variable -- alleles' allele fractions are not known.  Previously, M2 estimated allele fractions ffrom the ADs and proceeded with the likelihoods calculation.  Now M2 uses nifty math to account for the uncertainty in the allele fraction when calculating likelihoods.  In statistical language (which *will* be familiar to a decent-sized minority of users) we marginalize over allele fractions instead of using a maximum likelihood estimate.

>The WDLs do not include use of a contamination.table and so I did not include it in the commands. Is this something we want to nudge users to use, i.e. should I put in a sentence in the documentation section about the new tool CalculateContamination?

`CalculateContamination` is invoked inside the `Filter` task in mutect2.wdl.  We want users to use the new tool.  You might mention that in the interest of speed you can pass it `-L 1` to use only variants on chromosome 1, which gives a very good estimate in a couple of minutes.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
sooheelee: Added "The VCF INFO field contains the allele frequency (AF) tag.".

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
sooheelee: Clarified "To create the panel of normals (PoN), call on each normal sample using Mutect2's tumor-only mode and then use...."

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
sooheelee: Added "In GATK4, omitting filtering with FilterMutectCalls achieves the same artifact mode."

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
sooheelee: Thanks!

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
sooheelee: Fixed

sooheelee: Amended to "Mutect2's default variant site annotations differ..."

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
sooheelee: Good to know. FYI in turn, I think HaplotypeCaller has an argument where you can pass in a VCF for sites that you want considered active. I think it's called `--useAllelesTrigger`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
sooheelee: Amended to "...that are clearly present in the germline based on the evidence present..."

sooheelee:
Added in

[1] more details to differences with gatk3:
```
 *     <dd>(iv) Instead of using a maximum likelihood estimate, GATK4 Mutect2 marginalizes over allele fractions. 
 *     GATK3 MuTect2 directly uses allele depths (AD) to estimate allele fractions and calculate likelihoods. In contrast, GATK4 Mutect2
 *     factors for the statistical error inherent in allele depths by marginalizing over allele fractions when calculating likelihoods.</dd>
 *     <dd>(v) GATK4 Mutect2 recommends including contamination estimates with the -contaminationFile option from {@link CalculateContamination}, 
 *     which in turn relies on the results of {@link GetPileupSummaries}.</dd>
```
[2] another section:
```
 * <h3>Further points of interest</h3>
```

[3] and a new point to the new section:
```
* <p>
 *     If a variant is absent from a given germline resource, then the value for --af_of_alleles_not_in_resource applies. 
 *     For example, gnomAD's 16,000 samples (~32,000 homologs per locus) becomes a probability of one in 32,000 or less.
 *     Thus, an allele's absence from the germline resource becomes evidence that it is not a germline variant.
 * </p>
```

sooheelee:
I've also incorporated @davidbenjamin 's feedback. Please let me know if there are additional fixes for the documentation section.

sooheelee:
Thanks DavidB! @droazen, may I please have permission to merge PRs? Or, if you feel I'm not ready yet as a git newb, please, can you merge this PR for me? Thank you.

PR #2879, opened 2017-06-05
Re-enable openPrivateFileWithExplicitCredentials test

jean-philippe-martin:
Also updated code and documentation to indicate how to properly run this
test. Sadly it has to be done manually because I don't know of a
reasonable way to disable default credentials.

Nevertheless it's good that the test is there even for automated runs,
so we can check that loading the explicit credentials does not break
anything.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
@@ -150,11 +150,13 @@ public abstract class BaseTest {
     }
 
     /**
-     * A local path where the service account credentials are stored
-     * @return GOOGLE_APPLICATION_CREDENTIALS env. var if defined, throws otherwise.
+     * A local path where the non-default service account credentials are stored.
+     * GOOGLE_APPLICATION_CREDENTIALS contains the default credentials, but this is
+     * to test the code path for explicit credentials.
+     * @return HELLBENDER_JSON_SERVICE_ACCOUNT_KEY env. var if defined, throws otherwise.
      */
     public static String getGoogleServiceAccountKeyPath() {
-      return getNonNullEnvironmentVariable("GOOGLE_APPLICATION_CREDENTIALS");
+      return getNonNullEnvironmentVariable("HELLBENDER_JSON_SERVICE_ACCOUNT_KEY");
     }
 
     protected static String getNonNullEnvironmentVariable(String envVarName) {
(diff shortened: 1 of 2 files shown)


jean-philippe-martin:
It looks like the errors are all of the flavor:

```
Unable to load Maven meta-data from https://artifactory.broadinstitute.org/artifactory/libs-snapshot/com/github/samtools/htsjdk/2.9.1-34-gd7bae17-SNAPSHOT/maven-metadata.xml.
            > Could not GET 'https://artifactory.broadinstitute.org/artifactory/libs-snapshot/com/github/samtools/htsjdk/2.9.1-34-gd7bae17-SNAPSHOT/maven-metadata.xml'. 
Received status code 401 from server: Unauthorized
```
Perhaps Maven's temporarily in a bad mood, we'll have to try again later.

lbergelson:
@jean-philippe-martin They moved our artifactory so there was probably a downtime where it was inaccessible.  I've restarted.  There's supposed to be a redirect in place from the old link so it should work.

jean-philippe-martin:
Ah that explains it, thank you. I see it failed again in the same way. Perhaps it needs more time?

lbergelson:
@jean-philippe-martin Seems to have resolved now.

jean-philippe-martin:
Great! Now that we know it works, it can wait to be reviewed. It's a small change but this feature had stayed untested for too long.

In fact an even better way of doing this, if we have the energy & desire, would be to set up a separate bucket with a separate project. I *think* if we do it right we may then be able to have a fully automated test of the explicit credentials, as the default credentials would have access to the "normal" test data but we'd make sure the default user is not authorized to access the separate bucket (so we have to use the explicit credentials).

jean-philippe-martin:
@droazen? This should be straightforward to review... if we don't wait so long that the codebase drifts away.

jean-philippe-martin:
@droazen it's been a month. Perhaps we're ready to review this code now? It's a small change.

lbergelson:
It seems like travis isn't really testing this properly, it's hard to prove that it's not using the default credentials.  Would it make sense to do the following?
1.  create a new gcloud project
2.  create a private file in a bucket only accessible by that account 
3.  add a service account json for that account to travis, and use it in this test.

jean-philippe-martin:
@lbergelson yes that sounds perfect.

jean-philippe-martin:
You should create the project, though, so you can control it.

jean-philippe-martin:
@lbergelson still waiting on the new test project...

droazen:
Closing this ancient PR -- this is hard to test/would take a lot of work to get in, and it's a somewhat exotic use case. We do already have tests covering access to private files with default credentials, which seems sufficient.

PR #3026, opened 2017-06-05
shl CalculateContamination tag doc, flesh out example command, mentio…

sooheelee:
…n replaces ContEst

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
@@ -5,6 +5,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
+import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hellbender.cmdline.CommandLineProgram;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 import org.broadinstitute.hellbender.cmdline.programgroups.VariantProgramGroup;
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
```diff
@@ -18,20 +19,35 @@
 import java.util.stream.IntStream;
 
 /**
- * Given pileup data from {@link GetPileupSummaries} calculate the fraction of reads coming from exogenous human DNA.
+ * Given pileup data from {@link GetPileupSummaries}, calculates the fraction of reads coming from cross-sample contamination.
  *
- * This tool is run as follows:
+ * <p>
+ *     The resulting contamination table is used with either {@link org.broadinstitute.hellbender.tools.walkers.mutect.Mutect2}
```
davidbenjamin: Just `FilterMutectCalls`, I believe.  You can check their associated `ArgumentCollection` classes to verify.

davidbenjamin:
@sooheelee back to you.

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
sooheelee: If I `gatk-launch Mutect2`, I see that Mutect2 also has 
```
--contamination_fraction_per_sample_file,-contaminationFile:File
                              Tab-separated File containing fraction of contamination in sequencing data (per sample) to
                              aggressively remove. Format should be "<SampleID><TAB><Contamination>" (Contamination is
                              double) per line; No header.  Default value: null. 
```

But I can remove this if you think that it is better to only filter contamination using CalculateContamination.

davidbenjamin: @sooheelee Excellent point, however these are relics of shared HaplotypeCaller code.  Mutect2 doesn't do anything with these arguments.  This probably means I should move them out of shared code and into HaplotypeCaller-only code.

sooheelee: That is good to know. I've removed mention of filtering with M2.

sooheelee:
@davidbenjamin Let me know if the changes look good to you. Thank you.

davidbenjamin:
@sooheelee I'm satisfied.

Issue #3030, opened 2017-06-06
CollectSequencingArtifactMetrics (GATK) has errors and requires workaround in downstream tools


sooheelee:
Currently, FilterByOrientation requires replacing a required input's contents with a sed command: 

```
sed -r "s/picard\.analysis\.artifacts\.SequencingArtifactMetrics\\\$PreAdapterDetailMetrics/org\.broadinstitute\.
 hellbender\.tools\.picard\.analysis\.artifacts\.SequencingArtifactMetrics\$PreAdapterDetailMetrics/g" \
 "metrics.pre_adapter_detail_metrics" > "gatk.pre_adapter_detail_metrics"
```

The required input is the detailed pre-adapter summary metrics from Picard's CollectSequencingArtifactMetrics. The sed command replaces
```
## METRICS CLASS	picard.analysis.artifacts.SequencingArtifactMetrics$PreAdapterDetailMetrics
```
with 
```
## METRICS CLASS	org.broadinstitute.hellbender.tools.picard.analysis.artifacts.SequencingArtifactMetrics$PreAdapterDetailMetrics
```

If this class is not replaced, then the tool errors with: 
```
htsjdk.samtools.SAMException: Could not locate class with name gatk.analysis.artifacts.SequencingArtifactMetrics$PreAdapterDetailMetrics
	at htsjdk.samtools.metrics.MetricsFile.read(MetricsFile.java:356)
	at org.broadinstitute.hellbender.tools.exome.FilterByOrientationBias.onTraversalStart(FilterByOrientationBias.java:102)
	at org.broadinstitute.hellbender.engine.GATKTool.doWork(GATKTool.java:777)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:115)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:170)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:189)
	at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:122)
	at org.broadinstitute.hellbender.Main.mainEntry(Main.java:143)
	at org.broadinstitute.hellbender.Main.main(Main.java:221)
Caused by: java.lang.ClassNotFoundException: gatk.analysis.artifacts.SequencingArtifactMetrics$PreAdapterDetailMetrics
[...]

PR #3031, opened 2017-06-06
shl FilterByOrientationBias tag doc, flesh out example command, introduce tool

sooheelee:
Please see <https://github.com/broadinstitute/gatk/issues/3030> for the errors I encountered when running this tool. I have not been able to run the tool successfully and so ask what are we missing in the documentation that will help users get this tool running.

Otherwise, I recommend labeling this tool experimental as well as in BETA.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/FilterByOrientationBias.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/FilterByOrientationBias.java
@@ -32,7 +32,7 @@ import java.util.stream.Collectors;
  * Additionally filter Mutect2 somatic variant calls for sequence-context dependent artifacts.
  *
  * <p>
- *     This tool is complementary to {@link org.broadinstitute.hellbender.tools.walkers.mutect.FilterMutectCalls}.
+ *     This tool is complementary to {@link FilterMutectCalls}.
  *     The tool requires the pre-adapter detailed metrics calculated by Picard CollectSequencingArtifactMetrics.
  *     Specify the base substitution to consider for orientation bias. For a given base substitution specified with
  *     the --artifactModes argument, the tool considers both the forward and reverse complement for filtering.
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/FilterByOrientationBias.java
```diff
@@ -25,9 +26,44 @@
 import java.util.*;
 import java.util.stream.Collectors;
 
+/**
+ * Additionally filter Mutect2 somatic variant calls for sequence-context dependent artifacts.
+ *
+ * <p>
+ *     This tool is complementary to {@link org.broadinstitute.hellbender.tools.walkers.mutect.FilterMutectCalls}.
+ *     The tool requires the pre-adapter detailed metrics calculated by Picard CollectSequencingArtifactMetrics.
+ *     Specify the sequence context(s) to consider for filtering with the --artifactModes argument.
```
davidbenjamin: >Specify the sequence contexts(). . .

This isn't quite right.  Rather it specifies what types of substitutions eg G --> T to consider.  For every substitution, the Picard detailed metrics measure orientation bias at all 3-base contexts, but the user can't control that.  Note also that if you specify a substitution the tool filters both it and its reverse complement eg G --> T *and* C --> A.  Only one substitution in a reverse complement pair should be specified.

davidbenjamin:
@sooheelee I think the experimental label is fine for now.  Lee is still evaluating it.

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterByOrientationBias.java
sooheelee: Incorporated

sooheelee:
Lee is looking for someone else to evaluate it @davidbenjamin. In the meanwhile, because we have the BETA/Experimental label on this tool, it should be ok to release.

sooheelee:
I will let someone else be the decider on this one. Perhaps @LeeTL1220 or @droazen?

sooheelee:
@davidbenjamin I've incorporated an experimental/BETA label. Am I missing something from your review?

davidbenjamin:
Thumbs up!

PR #3032, opened 2017-06-06
shl SplitIntervals tag doc, give example command, clarify some details

sooheelee:
### Question for @davidbenjamin 
- Not clear if SplitIntervals actually divides by the count of intervals or by genomic territory covered. Specifically, what is the difference between --subdivision_mode INTERVAL_SUBDIVISION and BALANCING_WITHOUT_INTERVAL_SUBDIVISION? The argument description is unclear.
- What is the point of having the default scatter count 1? For future reference, can we change this default.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/SplitIntervals.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/SplitIntervals.java
@@ -45,7 +45,7 @@ import java.util.stream.IntStream;
  *    The default --scatter_count is 1 and so this value should be changed to utilize the tool's functionality.
  * </p>
  *
- * @author David Benjamin &lt;davidben@brgoadinstitute.org&gt;
+ * @author David Benjamin &lt;davidben@broadinstitute.org&gt;
  */
 @CommandLineProgramProperties(
         summary = "Split intervals into sub-interval files.",
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/SplitIntervals.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/SplitIntervals.java
@@ -6,6 +6,7 @@ import htsjdk.samtools.util.IntervalList;
 import htsjdk.samtools.util.RuntimeIOException;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
+import org.broadinstitute.barclay.help.DocumentedFeature;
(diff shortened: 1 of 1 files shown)


davidbenjamin:
@sooheelee `INTERVAL_SUBDIVISION` splits the original intervals into interval lists with spanning equal genomic territory, and is willing to break up some of the original intervals to do so.  `BALANCING_WITHOUT_INTERVAL_SUBDIVISION` has the same goal of equal numbers of covered bases but is not willing to split an original interval.

For example, suppose scatter count is 3 and your intervals are `-L 1:1-1000 -L 1:2000-3000 -L 2`.  Then the output interval lists under `INTERVAL_SUBDIVISION` will be
1) 1:1-1000, 1:2000-3000 and about 1/3 of chr2
2) the next 1/3 of chr2
3) the last 1/3 of chr2

davidbenjamin:
@sooheelee One reason to make the default 1 is that doing nothing, while silly, is also the least arbitrary choice.  Another reason is that omitting the scatter count in a wdl is handled naturally i.e. by keeping the original intervals list intact.

src/main/java/org/broadinstitute/hellbender/tools/walkers/SplitIntervals.java
```diff
@@ -21,17 +22,37 @@
...
+ *   -O interval-files
+ * </pre>
+ *
+ * <p>
+ *    The -O argument specifies a directory name for the scatter intervals files. Each file will be named, e.g 0000-scattered.intervals,
+ *    0001-scattered.intervals, 0002-scattered.intervals and so on.
+ *    The default --scatter_count is 1 and so this value should be changed to utilize the tool's functionality.
+ * </p>
+ *
+ * @author David Benjamin &lt;davidben@brgoadinstitute.org&gt;
```
davidbenjamin: typo: brgoad

davidbenjamin:
@sooheelee Just the one typo; otherwise all is good.

davidbenjamin:
@sooheelee looks good.

sooheelee:
Thanks for all the clarifications DavidB. Thanks for the merge DavidR.

sooheelee:
Just noticing that 

```
Default value: INTERVAL_SUBDIVISION. Possible values:
                              {INTERVAL_SUBDIVISION, BALANCING_WITHOUT_INTERVAL_SUBDIVISION}.
```

So we need to warn users to be careful when dividing lengthy genomic intervals. Perhaps it would be wise to specify the workflow in which this tool would be used. Something for the second pass.

PR #3036, opened 2017-06-06
Made changes to joint segmentation and added temporary AllelicCNVHMM CLI.

samuelklee:
Migrated from https://github.com/broadinstitute/gatk-protected/pull/962.

Changes by @davidbenjamin
-changes to docs
-got rid of weights and concentration
-got rid of constant states to simplify before CRP pre-training
-smarter transition matrix
-switched to binomial AF likelihoods for segmentation
-got rid of attempt big change in memory length
-fixed outlier likelihood
Changes by @samuelklee
-ACNV with joint segmentation
-tweaked convergence criteria and removed extraneous MCMC fit
-sorted acc in AF segmentation
-NaN fixes in binomial likelihood
-fixed some tests and added EXPERIMENTAL tags
-disabled JointAFCRSegmenterUnitTest

This introduces a new command line (AllelicCNVHMM---@sooheelee, this command line is experimental and should not be used) that performs joint segmentation and then fits model parameters using MCMC. It performs relatively well on some samples (and was used to generate results for the AACR poster), but others result in oversegmentation and convergence issues. It's possible that this could be due to the naive copy-ratio model used. @davidbenjamin may want to do some additional tweaking, but I think we will also explore other iHMM variants concurrently.


sooheelee:
Ok thanks for letting me know about this new tool and thanks for adding the appropriate (Experimental) label. We also need to add the BETA class label until a time `experimental` gets its own official barclay label (DR opened a ticket requesting this today). The BETA tag goes as such underneath the DocumentedFeature tag (which your tool may not have yet):

```
 @DocumentedFeature
 @BetaFeature
```

sooheelee:
I added the tags since this PR has been sitting here for a while.

PR #3045, opened 2017-06-06
ChangedHaplotypeCallerDoc

chandrans:
Added some websites with tool information and updated the command


sooheelee:
Should we mention that when calls are being made (non-GVCF workflow) there is an option for a new QUAL model?

PR #3051, opened 2017-06-07
Fix javadoc HTML in ReadLengthReadFilter and WellformedReadFilter

magicDGS:
Using character entities in javadoc to correct HTML formatting.


magicDGS:
Thanks!

PR #3057, opened 2017-06-07
Temporarily disable IndexFeatureFile and tests to avoid creation of bad indices.

cmnbroad:
Not sure if there is a better way to do this. UserException seems wrong since its displays USER Error.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
@@ -9,6 +9,7 @@ import htsjdk.tribble.util.TabixUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.barclay.argparser.BetaFeature;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hellbender.cmdline.CommandLineProgram;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
```diff
@@ -33,7 +35,7 @@
         oneLineSummary = "Creates indices for Feature-containing files (eg VCF and BED files)",
```
droazen: Add in the text `(Disabled) ` at the start of both the `summary` and `oneLineSummary` here.

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
```diff
@@ -33,7 +35,7 @@
...
         programGroup = VariantProgramGroup.class
 )
-@DocumentedFeature
+@BetaFeature
```
droazen: Add comment to restore `@DocumentedFeature` when the tool is fixed.

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
```diff
@@ -51,6 +53,9 @@
 
     @Override
     protected Object doWork() {
+        if (true) {
+            throw new GATKException("The IndexFeatureFile tool is temporarily disabled.");
```
droazen: `UserException` -- it's their fault for trying to run a disabled tool! :)

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
cmnbroad: Done.

magicDGS:
Is there any time-plan to revert this tool?

cmnbroad:
@magicDGS I expect we'll try to get this done as quickly as we can, though it will require a fix in htsjdk.

magicDGS:
Thanks @cmnbroad!

PR #3062, opened 2017-06-07
ChangedGenotypeGVCfsDoc

chandrans:
Added some minor changes for easier reading. Should insert some docs for CombineGVCFs when we know where the doc lives.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
@@ -26,35 +26,48 @@ import java.io.File;
 import java.util.*;
 
 /**
- * Perform joint genotyping on gVCF files produced by HaplotypeCaller
+ * Perform joint genotyping on one or more samples pre-called with HaplotypeCaller
  *
  * <p>
- * GenotypeGVCFs merges gVCF records that were produced as part of the Best Practices workflow for variant discovery
- * (see Best Practices documentation for more details) using the '-ERC GVCF' or '-ERC BP_RESOLUTION' mode of the
- * HaplotypeCaller, or result from combining such gVCF files using CombineGVCFs. This tool will produce correct genotype
- * likelihoods, re-genotype the newly merged record, and then re-annotate it.</p>
+ * This tool is designed to perform joint genotyping on multiple samples pre-called with HaplotypeCaller to produce a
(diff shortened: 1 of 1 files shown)


sooheelee:
CombineGVCFs hasn't been ported to GATK4 @chandrans. If people need to use it, they will have to use the GATK3 version. To replace it, there are GenomicsDB tools instead. 

In lieu of mentioning a new tool we haven't yet documented, perhaps we can mention that CombineGVCFs is only in GATK3?

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -26,35 +26,48 @@
...
+ *   -O output.vcf
+ * </pre>
+ *
+ * <h4>Perform joint genotyping on a set of GVCFs listed in a text file, one per line</h4>
  * <pre>
- * gatk-launch GenotypeGVCFs \
+ * ./gatk-launch GenotypeGVCFs \
  *   -R reference.fasta \
- *   -V sample1.g.vcf \
+ *   -V input_gvcfs.list \
```
vdauwera: @droazen we can still do this in GATK4, right?

vdauwera:
one question for @droazen or @lbergelson re: list input capabilities

vdauwera:
@sooheelee we'll document that separately as part of the pipeline docs; that doesn't belong in individual tool docs imho

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
vdauwera: (I'm going to assume this is the case)

vdauwera:
will merge when tests pass

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
ronlevine: Should this be camelCase, for consistency?

vdauwera: I prefer snake case for file names as it's easier to parse.

lbergelson:
@vdauwera Any gatk4 argument that takes a collection can accept a list file.

PR #3064, opened 2017-06-07
Porting forceJVMLocaleToUSEnglish from GATK3.

### LINKED ISSUE #2557: Implement equivalent of forceJVMLocaleToUSEnglish() from GATK3
I noticed this in the gatk3.  It seems like maybe we should port it to 4?
```
        /**
         * The very first thing that any GATK application does is forces the JVM locale into US English, so that we don't have
         * to think about number formatting issues.
         */
        forceJVMLocaleToUSEnglish();
```

jonn-smith:
Placed a call to forceJVMLocaleToUSEnglish in Main.runCommandLineProgram so that integration tests can take advantage of this.

Fixes #2557

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -115,6 +115,15 @@ public class Main {
      * @return the result of running  {program} with the given args, possibly null
      */
     protected static Object runCommandLineProgram(CommandLineProgram program, String[] rawArgs) {
+
+        /**
+         * The very first thing that any GATK application does is forces the JVM locale into US English, so that we don't have
+         * to think about number formatting issues.
+         *
+         * Placing the call here so that it is enabled for integration tests as well.
+         */
+        Utils.forceJVMLocaleToUSEnglish();
+
         if (null == program) return null; // no program found!  This will happen if help was specified with no other arguments
         // we can lop off the first two arguments but it requires an array copy or alternatively we could update CLP to remove them
         // in the constructor do the former in this implementation.
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -34,6 +34,14 @@ import java.util.*;
  */
 public class Main {
 
+    static {
+        /**
+         * The very first thing that any GATK application does is forces the JVM locale into US English, so that we don't have
+         * to think about number formatting issues.
+         */
+        Utils.forceJVMLocaleToUSEnglish();
+    }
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -115,6 +115,15 @@ public Object instanceMain(final String[] args, final List<String> packageList,
...
      */
     protected static Object runCommandLineProgram(CommandLineProgram program, String[] rawArgs) {
+
+        /**
+         * The very first thing that any GATK application does is forces the JVM locale into US English, so that we don't have
+         * to think about number formatting issues.
+         *
+         * Placing the call here so that it is enabled for integration tests as well.
+         */
+        Utils.forceJVMLocaleToUSEnglish();
```
cmnbroad: It might be better to do this in a static block so the locale is established right from the start, and would be in place for unit tests that don't execute this code path.

cmnbroad: Also, we should probably see if we can propagate this setting to Spark executors.

droazen: I agree with @cmnbroad that this call should happen in a static initializer block in `Main.java`, which was the case in GATK3.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1121,4 +1121,11 @@ protected T computeNext() {
     public static String getDateTimeForDisplay(final ZonedDateTime dateTime) {
         return dateTime.format(longDateTimeFormatter);
     }
+
+    /**
+     * A hack to ensure that numbers are always formatted in the US style.
```
droazen: This is not really a hack :)  Method description should just be "Set the Locale to US English so that numbers will always be formatted in the US style."

src/main/java/org/broadinstitute/hellbender/Main.java
droazen: I'd be satisfied if we create a separate ticket to address the Spark issue, unless you have a solution at hand @cmnbroad

cmnbroad: I'm not sure how to do it so probably a separate ticket.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Fixed.  (I had just copied over the method from GATK 3 and had preserved the comment.)

PR #3072, opened 2017-06-09
Increase NIO retries from 3 to 20, and make the timeout settings even more generous

droazen:


droazen:
@jean-philippe-martin We've found that with the current NIO retry settings, we're still exhausting all retries and failing ~1-2% of the time on large runs. This PR increases the number of retries from 3 to 20, which should trigger waits of several minutes rather than several seconds in the later retries. It also increases the timeout settings in `BucketUtils.setGenerousTimeouts()` by quite a bit, also with the goal of allowing waits in minutes rather than seconds when necessary.

We'll try running with this and see what happens, but in the meantime please review.

jean-philippe-martin:
The change is good :thumbsup:

You may also want to investigate having fewer parallel reads, so we get throttled less.

droazen:
Thanks JP! We do have an argument to control the number of simultaneous GCS readers in the tool in question. However, even setting it to a low value is triggering the throttling, since we're running about 10,000 ways parallel, and each job is accessing the same set of files. We can currently only control the number of simultaneous readers within each of the 10,000 individual jobs, not across all jobs.

jean-philippe-martin:
That's a lot of jobs. How many jobs per machine? (You can reply over email if you prefer)

droazen:
No failures so far on the latest run -- think this may have done the trick :)

droazen:
Not sure about the number of jobs per physical machine @jean-philippe-martin , I'd have to check with @Horneth who is running the tests.

droazen:
Turns out that the failure rate with this patch was greatly reduced, but there were still a few failures. We're trying a run now where we tell the tool to import one file at a time (as opposed to a batch of 25 or 100), and we'll see how that goes.

PR #3076, opened 2017-06-09
modified init script to put ref idx img in local ssd

SHuang-Broad:
rather than the default boot HDD.
Thanks to @mwalker174 for discovering this.

In terms of runtime, this didn't change runtime for the SV pipeline much.

### DIFF
--- a/scripts/sv/default_init.sh
+++ b/scripts/sv/default_init.sh
@@ -15,5 +15,5 @@ if [[ ! "${ROLE}" == 'Master' ]]; then
     if  [[ ! $REFLOC =~ .+/$ ]]; then
       REFLOC+="/"
     fi
-    mkdir -p /reference && gsutil -m cp "$REFLOC"*.img /reference/
+    mkdir -p /mnt/1/reference && gsutil -m cp "$REFLOC"*.img /mnt/1/reference/
 fi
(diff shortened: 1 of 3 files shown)


scripts/sv/default_init.sh
```diff
@@ -15,5 +15,5 @@ if [[ ! "${ROLE}" == 'Master' ]]; then
     if  [[ ! $REFLOC =~ .+/$ ]]; then
       REFLOC+="/"
     fi
-    mkdir -p /reference && gsutil -m cp "$REFLOC"*.img /reference/
+    mkdir -p /mnt/1/reference && gsutil -m cp "$REFLOC"*.img /mnt/1/reference/
```
mwalker174: Maybe add a comment explaining /mnt/1/ like you have in runWholePipeline here, just in case this fails if the default path changes in the future.

SHuang-Broad:
@mwalker174 . Thanks for taking a look! I've added the requested comments and also added a check, in case the cluster was NOT created with a local SSD.

SHuang-Broad:
@mwalker174 Hi Mark, are you OK with the new changes? Thanks!

mwalker174:
Yes looks great! Sorry for the late reply - didn't realize I was the only reviewer.

Issue #3104, opened 2017-06-13
Port IndelRealignment pipeline


magicDGS:
After discussion in #3084, I offer myself to port the indel realignment pipeline. After exploring the GATK3 implementation, I will split the port in the following independent tasks:

- [ ] Port `RealignerTargetCreator` (require test data after run with GATK3)
- [ ] Port `ConstrainedMateFixingManager`
- [ ] Port `NWaySAMFileWriter` (requires some change in the engine to get the ID for the inputs)

The previous port will be integrated in the `IndelRealigner` tool implementation.

magicDGS:
Can you provide me some test data for include in the tools integration test, @vdauwera and/or @sooheelee? If not, I will try to use some BAM files already in the repository...

lbergelson:
It's not clear to me that we want these tools in Gatk4.  We deliberately didn't port them because we felt they were unnecessary going forward.  

I understand that there are some legitimate use cases that require them: ex low coverage naive variant calling from high ploidy pools which haplotype caller would do poorly on.   (Also, do we know that haplotype caller doesn't do well on those sorts of things?  Maybe we should consider modifications there if it doesn't?)   I'm not sure that supporting that use case is worth the added complexity of maintaining and supporting these tools.  Especially since we don't provide a pileup based variant caller as part of gatk4...  

@vdauwera Can you comment?  

@sooheelee I'm not sure I agree with you that supporting this for mutect 1 is useful. 
 A) We don't want to support the use of mutect 1 anymore and would like to encourage people to switch to mutect 2 which I think we now believe is a better variant caller for both snps and indels.  
B) Mutect 1 users are already using gatk3, so they have access to these tools already.  Mutect 1 also requires co-cleaning which I believe is a different but related tool to indel realignment.   

For the variant review issue, we have  thoughts on implementing a much better solution for variant review by creating an assembly plugin for igv.

magicDGS:
I thought after the last comment of @vdauwera in the [blog post about the removal](http://gatkforums.broadinstitute.org/gatk/discussion/7131/is-indel-realignment-removed-from-gatk4) that it will be possible to port it here as a contribution (without too much effort from your dev team). If the final solution is that this is not going to be maintained in GATK4, I would port this code to my own software if you give me the permission; but it is definitely something that the community is interested.

[...]

PR #3107, opened 2017-06-13
Port ConstrainedMateFixingManager from GATK3

magicDGS:
As part of #3104


magicDGS:
Could you nominate someone for this one, @droazen?

In addition, I would like to use the `GATKRead` interface instead of `SAMRecord`, but I do not look into the logic enough to do the change yet...

droazen:
For @cmnbroad to review when he has the time

magicDGS:
@cmnbroad - any chance to get to this one?

magicDGS:
@cmnbroad - this is also part of the port of indel realignment, which should be easier to review because it is not a tool but a helper class (thus, documentation and arguments formatting shouldn't be a problem here).

magicDGS:
Closing (will be part of a bigger PR with the port for both tools from the pipeline)

PR #3108, opened 2017-06-13
UpdatedConcordanceDoc

chandrans:
Simple fixes to capitalize VCF and changed command to reflect ./gatk-launch

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
@@ -27,27 +27,38 @@ import java.util.EnumMap;
 import java.util.Set;
 
 /**
+ * Evaluate concordance of an input VCF against a validated truth VCF
  *
- * This tool evaluates a vcf against a validated truth vcf. We assume that the truth vcf only contains PASS variants.
- * The summary statistics (# true positives, # false positives, # false negatives, sensitivity, precision)
- * are reported as a summary tsv (--summary).
+ * <p>This tool evaluates an input VCF against a VCF that has been validated and is considered to represent ground truth.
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
```diff
@@ -27,27 +27,38 @@
...
  *
- * Optionally, the tool also produces vcfs of
- * 1) true positives and false negatives (i.e. all variants in the truth vcf)
- *      This mode is useful for calculating sensitivity
- * 2) true positives and false positives (i.e. all variants in the eval vcf)
- *      This mode is useful for obtaining a training data set for machine learning classifiers of artifacts.
+ * <p>Optionally, the tool also produces VCFs of the following:</p>
+ * <ul>
+ *     <li>True positives and false negatives (i.e. all variants in the truth VCF): useful for calculating sensitivity</li>
+ *     <li>True positives and false positives (i.e. all variants in the eval VCF): useful for obtaining a training data
```
vdauwera: Not sure I understand the point of these two -- what's different compared to the actual eval and truth vcfs we started with?

vdauwera:
One question for either @takutosato or @davidbenjamin

src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
davidbenjamin: @vdauwera These difference is that these have an additional `INFO` field tag for true positive, false negative, false positive, filtered true negative, and filtered false negative.  These tags turn the input vcfs into training data for classifiers.

vdauwera:
Will merge once tests pass

PR #3110, opened 2017-06-13
Disable NIO prefetching in GenomicsDBImport

droazen:
Empirical testing has shown that this tool performs best at scale with cloud buffering
disabled. With cloud buffering on and thousands of concurrent GenomicsDBImport tasks,
we do too many simultaneous GCS accesses (since the prefetcher spawns a new thread for each
reader upon a query) and start seeing intermittent failures, even with aggressive retries.


droazen:
@jean-philippe-martin Could you review?

PR #3112, opened 2017-06-14
Port RealignerTargetCreator

magicDGS:
As part of #3104


droazen:
@cmnbroad please review when you have the time

serge2016:
Hello! What is the progress here?

magicDGS:
Still waiting for review, @serge2016...

magicDGS:
@cmnbroad - any chance to review this one?

cmnbroad:
@magicDGS My apologies for the long delay on this. It looks like there are still quite a few standard porting issues that need to be addressed in this PR (brackets, finals, multiple top-level public classes, outdated GATK3 usage example, remove references to RODs, kebabify, etc., etc.). There is also the bigger issue of testing and validation - ideally at a minimum we'd reproduce the existing GATK3 tests, but since these are dependent on large, private files, those tests will have to be replaced with new tests, and validated/compared against GATK3. These same issues will come up with IndelRealigner.

@vdauwera @sooheelee Even with @magicDGS graciously volunteering to do the work of porting the code, retaining the indel realignment tools will require internal review, support. Before we commit to that, I guess I want to make sure that this is indeed a high priority, and that you think porting this to GATK4 is a better option than relying on GATK3, or letting @magicDGS port it to ReadTools ?

chandrans:
My two cents. I think we have had about 3-4 posts on the forum about this. People do want this in GATK4.

magicDGS:
My first idea was to port it to ReadTools (that's why I asked to change it to the public code when GATK3 was divided into the non-free version and the plan was not to port indel-realignment). Nevertheless, someone showed interest into this and that's why I did this PRs; in addition, it is much better than the original authors (if still at Broad) can contribute to see if the port is correct. Anyway, I don't have any strong feeling about this, and I can always port the pipeline to ReadTools and maintain it there (actually, I had some ideas for other realignment methods and it will be interesting to compare with this).

@cmnbroad - this was a long time ago, when even some of the standards (documentation and kebab-case arguments) were not even implemented. Regarding the tests, the last commit includes one using a BAM file in the repository, and the output was generated with GATK3 (if I remember correctly, although I can't remember the version).

I think that the best way to go is to re-open a PR to have a cleaner history and proper standards. The idea will be:

1. Copy of the GATK3 files (from the latest release)
2. Reformat (license and coding-style)
3. Port to the new framework.
4. Convert documentation and arguments to the GATK4 format
5. Add tests by running the latest GATK3 release on data in the repository and checking concordance.

If a different dataset should be used for the tests, it will be nice if you can provide some. Otherwise, I will use data in the repository (b37_reference_20_21 and NA12878_20_21_WGS_bam)

cmnbroad:
@magicDGS We face this same issue with many of these tool ports - porting the code is often the easy part by far. I think we'd want to have a set of tests with reasonably good code coverage to start reviewing these PRs, as well as getting concordance with GATK3. And we'll have the same issue with IndelRealigner. I don't have any dataset at the ready other than the one used in GATK3. I can see if any of that data, which is currently private, can be shared with you if thats helpful.

magicDGS:
@cmnbroad - I can add more tests if there is available data for it, so it will be nice if you can figure out if the data can be made public. Now, the coverage for the tool is 74-75%, because there is no test for known indels; but the test outputs were generated with GATK3 (I can't remember, as I told you before, with which version).

Should I close this PR and open a new one with the documentation and arguments with the current style of GATK4, and the tests that are now? I can also add the tests from the tutorial that @sooheelee has gave me, but they are kind of big (probably they should be added with lfs). Just let me know the approach that we should take here...

cmnbroad:
@magicDGS I will check on whether any of the data can be made available to you. However, the existing GATK3 tests look to be written against the b36 reference, which we can't include in GATK4 even using git-lfs. So either way, I think new tests have to be developed. A new PR would be fine (please include a link to this one), but my suggestion would be that we make sure we have a plan for how to get good test coverage for IndelRealigner before we go any further with RealignerTargetCreator.

magicDGS:
I think that a proper example would be the one in the tutorial from @sooheelee (https://software.broadinstitute.org/gatk/blog?id=7847, see also https://github.com/broadinstitute/gatk/issues/3104#issuecomment-314886000).

I divided the PRs for the `IndelRealignment` into 4 different sections for better review (2 components of indel-realignment, `RealignerTargetCreator`and `IndelRealignment`). This strategy is because I dissected the pipeline into the easy `RealignerTargetCreator` to found regions worth to look at (this could be marked as experimental/beta before the indel-realignment is in) and the more complicated and component-based `IndelRealigner` (the same as with other tools, this can be marked as experimental/beta until a really good coverage is achieved - in the meantime, I have some test with the current data in the repository and the GATK3 counterpart). There are two parts that are usable outside `IndelRealigner` that are worthy to separate into two commits, and might be useful for other tools/downstream projects: `ConstrainedMateFixingManager` and `NWaySAMFileWriter`. That's the reason of making the port in split PRs.

One option can be to have the PRs open, and reviewed independently without acceptance until every component is ready. Otherwise, I think that an experimental tag would be good until we find a good set of tests for edge cases. Does this approach make sense for you, @cmnbroad?

cmnbroad:
The tutorial data would be an option, except that it covers reference territory not currently in the repo. It looks like that would add nearly 1 gig just for the reference data, which I think we'd really want to avoid.

Ideally we'd have just one PR for each of the (two) tools. If you want to keep `ConstrainedMateFixingManager` and `NWaySAMFileWriter` as separate PRs, thats fine, but I don't think we'd take them until we know there is, or will very soon be, code that depends on them.

Certainly using `@Experimental` for the tools could make sense, once we're certain that we have a way forward for test coverage. One other note, it's very helpful to include the original GATK3 file as the first commit, as you did in this PR.

magicDGS:
Sorry, I completely missed this one in the last weeks. I was waiting for your answer (and @sooheelee's):

1. I need the test data for the public repository if we would like to reproduce the complete test suite that GATK3 has (see my comment above, https://github.com/broadinstitute/gatk/pull/3112#issuecomment-364866520)
1. The tutorial data might be a possibility for the `IndelRealignment` test suite, but it requires the full reference genome (which is huge and not possible to include in the repository, see https://github.com/broadinstitute/gatk/issues/3104#issuecomment-325612165 for more information). @sooheelee might have an idea about the data and if it can be limited to the reference already on the repository (if I am correct, 20 and 21).
1. This just came out now when I was looking at my port branches, both this and some backup for others part of the code. If the tool is going to be experimental, I would like to know if it is ok to do not have fully featured tools and fill in some issues in the meantime. For example, in this PR the part of the code uncovered by tests is the `known` indels: the main idea is to comment out the code and throw an exception if the argument is specified, saying that it is not supported yet. In the case of `IndelRealignment`, I would like to leave out in the first iteration for the tool the `NWaySAMFileWriter`, to speed up development and add the support for split output later.

Once I know this I will close the PR and sent another one with the `RealignmentTargetCreator`. On the other hand, the `IndelRealigner` tool will take a bit longer, because I haven't look at it for the longest time.

Thanks for looking into this @cmnbroad!

cmnbroad:
@magicDGS I think we covered all of these points earlier in this thread, but both the GATK3 test data and the tutorial data are written against reference(s) we don't have in GATK4, and are much larger than we want to include in the repo. We would need tests that work against one of the references we already have included.

sooheelee:
@magicDGS I'm just seeing this ticket. Sorry, I've been focused elsewhere, so many tickets come through my inbox, and my main work is mostly not through this repo.  I can make test data.

cmnbroad:
Thanks @sooheelee. There are two large-ish reference files in the repo, both of which contain contigs for both 20 and 21: Homo_sapiens_assembly38.20.21.fasta and human_g1k_v37.20.21.fasta. So the tests would need to work against one of those.

sooheelee:
@cmnbroad could you point me to the relevant tests so I can figure out the various resource files that I assume the tests use in their commands?

sooheelee:
Also, it may be prudent for me to run the data through the commands the tests use, as the data I will make comes from an external source and may not validate in its current state, depending on the tool.

magicDGS:
Sorry for the late answer @sooheelee and thanks for looking into this. I guess that the tests for the two tools that are in the GATK3 code are the ones that we should reproduce (maybe the large scale will be too large):

* https://github.com/broadgsa/gatk/blob/0b73e380436aaa5a41fb3aab97ab651207669f47/public/gatk-tools-public/src/test/java/org/broadinstitute/gatk/tools/walkers/indels/RealignerTargetCreatorLargeScaleTest.java
* https://github.com/broadgsa/gatk/blob/0b73e380436aaa5a41fb3aab97ab651207669f47/public/gatk-tools-public/src/test/java/org/broadinstitute/gatk/tools/walkers/indels/RealignerTargetCreatorIntegrationTest.java
* https://github.com/broadgsa/gatk/blob/0b73e380436aaa5a41fb3aab97ab651207669f47/public/gatk-tools-public/src/test/java/org/broadinstitute/gatk/tools/walkers/indels/IndelRealignerLargeScaleTest.java
* https://github.com/broadgsa/gatk/blob/7f75fa449605851bc84d1da4e77a7cfed489ea33/public/gatk-tools-public/src/test/java/org/broadinstitute/gatk/tools/walkers/indels/IndelRealignerIntegrationTest.java

cmnbroad:
@magidDGS We'd definitely prefer to not take partially developed versions of either of these tools (or @Experimental or @Beta versions). We should aim to have either a single PR with everything, or two PRs, one for each tool, with complete functionality, including tests and test data, rather than an ongoing stream of partial PRs.

magicDGS:
@sooheelee - the main idea is to run both tools with GATK3 on the test data, and then check if the port provides the same answer. Thanks a lot @cmnbroad and @sooheelee to look into this. I will try to make today a new PR for `RealignerTargetCreator` and start the work from scratch for `IndelRealignment` (it will take a bit longer, although without some support it might be actually a bit lighter too).

sooheelee:
@magicDGS I glanced at the tests and it appears there may be a scientific validation component, which I am unfamiliar with (I'm in an intro to Java course currently). Is this the case? If scientific validation is needed, then it is best to involve someone familiar with validation, e.g. Laura or Yossi. If all you need is data that can be run through these commands, I can put this together. Let me know.

I'm late to these efforts, but I'd like to check one thing. Because of the way GRCh38 contigs are parsed, e.g. the HLAs that contain colons in their names, I believe we now prefer Picard-style intervals lists that tab-separate values instead of the `10:96000399-96000421` format that RealignerTargetCreator produces. I'm not certain of the status of GATK-style intervals lists, but I do know that the CNV developers have swiched to Picard-style. Is this what is produced by the new RealignerTargetCreator?

magicDGS:
@sooheelee - I think that to have a proper test suite similar to GATK's 3, we need also some test that exercise some code paths that requires some specific scientifically meaningful data (for example, known indels that are also included in some sample). I am not that familiar with the test data on this or GATK3 repository, so I am not sure if that was already covered.

@sooheelee - this is going to be an exact port of the indel-realignment pipeline, as it is in the GATK3 code, so that means that I won't modify the interval list format or anything (although I will use the HTSJDK/Picard classes as used on GATK3). Because this will be an experimental/beta feature, I think that I can have a look to the new format after acceptance of the original port.

@cmnbroad - I understand that a fully functional tool is a requirement for acceptance, but what I mean is that some specific features might require more work than others. I am only concerned about the `NWaySAMFileWriter`, which is just an specific way of output the data but does not add anything to the real realignment process (actually, I think that I've never heard about anyone around me using it). That is a nice feature, but I don't think that it is a high-priority - I care more about having the algorithm implemented to test if the actual processing of the data works, and add support for some way of output the data in a different PR. In addition, if the people still using indel-realignment does not require the n-way output, then it is pointless to spend time on it. I was also thinking about the mate-fixing algorithm in the tool, because it can be performed afterwards with Picard, which is not constraining by any distance between reads or records in RAM - nevertheless, this is really a drop of functionality that will change results, and that's why I didn't propose that. About the target-creator, known indels are really easy to port because the code is within the tool and is simpler - the only problem might be code coverage if there is no data for known indels.

I will propose very soon two PRs with fully functional tools (without the n-way out feature for indel-realignment), and trying to add simple integration tests with the data already available on the repository and running with GATK3.8-1. If that is OK for you, I will proceed with this approach.

sooheelee:
Thanks for the clarification @magicDGS. What I have are 1000 Genomes Project BAMs, which were preprocessed by the 1000 Genomes Project. The BAMs are aligned to GRCh38 and have undergone indel realignment (GATK3), so they represent a state of preprocessing against which we can validate the new indel realignment. If some representation of indels in such a BAM is sufficient as a truthset, then I can make this VCF. These samples, being from the 1000 Genomes Project, are well-analyzed and their indels will be represented in dbSNP and gnomAD resources. Of course, the test data will have been re-mapped with BWA and be without any trace of indel realignment.

We should probably get someone familiar with the test suite to weigh in on the scope of the validation.

cmnbroad:
@sooheelee I can't speak for CNV, but there isn't any general reason to prefer Picard interval lists in GATK. There was previously an issue with parsing interval queries that used contig names that contained ":", but thats fixed now. The only time we prefer a Picard list is the theoretical case were you use a query interval against a sequence dictionary that contains contigs that make that query ambiguous (hg38 is not one of those). GATK will detect and reject such a query and suggest using a Picard interval file to disambiguate it.

@magicDGS I'm not sure how/if writing tests against existing files in the repository will be useful. I want to restate that we don't want to take ports of these tools if they're marked `@Experimental `or `@Beta` because they haven't been validated, or don't have good test coverage. We need to find a way need to have valid tests so they'll be production ready.

magicDGS:
Sorry @cmnbroad, I misunderstood your comment before - what I've got is that it is ok to have beta/experimental for this, but it's obviosly not what you said, so my fault for not reading carefully. In that case, I would like to have a proposal for how to proceed here:

* I will implement the port for the tools in two independent PRs - just direct translation into the new framework, documentation and kebab-case argument style.
* The port for `RealignerTargetCreator` will be similar to this one
* The port for `IndelRealigner` would not have support for n-way output, although it will be fully functional in other ways. The n-way option can be ported in the future as an extra feature if necessary (maybe the communications team can weight in, and tell if it is really a needed feature)
* For the first test, which will be removed before merging, I will use the data from the tutorial. This will be the validation for the port before test data valid for the repository is provided
* For the final tests, I will draft the class with the tests from GAKT3 without the data and disabled, waiting for @sooheelee for meaningful tests (or other people from your team)

The main idea is to have two PRs with the port validated with the tutorial data, and add other tests similar to the GATK3's ones for extra validation and/or covering missing codepaths using @sooheelee or ported GATK3 data. Does it makes sense, @cmnbroad and @sooheelee?

@sooheelee - I think that the 1000G data can be a good validation if chromosomes 20/21 have realigned reads without a pair mapping on other chromosomes. In that case, some of that reads can be extracted and reset to the state previous to realignment to validate the new tool.

sooheelee:
@magicDGS `--nWayOut` is absolutely necessary for co-cleaning tumor-normal pairs of samples. This allows upfront filtering of tumor mutations against the matched-normal, the panel of normals and the germline sites to be much more efficient (in somatic mutation calling). 

I will then go ahead and start making a base test dataset. The repo tests look like they tested all sorts of things, e.g. bad CIGARs. Will you be altering the base data I provide towards these other tests or are we focused just on the primary functionality? If more than just a normal dataset is needed, then do specify these requirements. Remember that I do not as of yet understand Java code. It would be helpful to also see the previous test data if its features are what we want to recapitulate. If we are instead making up new tests, then that is fine too.

Since you bring up nwayout, it seems the test should accommodate two samples. I was thinking to make alt-ware mapped GRCh38 snippets that included ALT/HLA contigs, just to cover our bases. In fact, I can make test data on post-alt processed reads, to further accommodate recent advances in genomics. Let me know your thoughts.

sooheelee:
I have to say, too bad we don't have a mechanism in place that allows for the full reference, e.g. NIO only the contigs or portions thereof that are needed for a particular analysis. That would make making test data so much easier. I would imagine this is simple to implement, given the reference is indexed and such a feature would be useful for cloud analyses. I have to jump through ridiculous hoops to make small test data.

cmnbroad:
I totally sympathize - it is super painful. But unfortunately we can't put reference-sized files into a Github repository without running into both disk size and time limits with both Github and Travis.

sooheelee:
The test data I have for you requires its own mini-reference, which I have prepared this afternoon. I am in the process of working on the alignment data.

sooheelee:
@cmnbroad, is 400MB (fasta, dictionary & index) too large for you?
It is possible for me to ~halve this but I have to construct artificial contigs.

sooheelee:
@cmnbroad says a 400MB reference is too large and the largest file they have is 250MB. So I need to pare down the reference and data further. This will involve making artificial contigs that are basically just snippets of the original GRCh38 contigs.

sooheelee:
Actually, this was fairly easy to do and the new reference set is ~85MB.

magicDGS:
Thanks a lot for looking into this @sooheelee - I understand the pain of making the small data for the tests, so I really appreciate your work there. I have almost prepare a PR for the new port of `RealignerTargetCreator` that will have TODOs for the test data prepared by you, and with some tests using files already in the repository (for the target-creator, I guess that the validation for getting the regions to realign would be enough, because thanks to that tests I realized of a small bug due to not including all the loci in the `LocusWalker`).

On the other hand, I will still fight for the `--nWayOut` not blocking the inclusion of `IndelRealigner` in the first place. My reasons are the following:
 
1. Looking a bit into the code of GATK3, there is a lot of complication to get the reader ID for each read. It will require to modify the `GATKRead` interface, the data source for reads, or find an *ad hoc* solution on `IndelRealignment` to set the procedence of the read. This requires going into the engine-level code, which in my experience is difficult to port from GATK3 and also slow on the reviewing/acceptance process.
1. My idea for developing a new writer of general use as the n-way output (which can be used in other tools as well) is to factor out some code from `SplitReads` to have a custom `GATKReadWriter` for arbitrary splitting. i'm already using a similar solution on `ReadTools`, so backporting the code to GATK might be a solution. Nevertheless, this still requires that the `GATKRead` has somehow the identity store at the object level, which requires to address point 1.
1. The use case of the tumor-normal pair can be resolved by an extra processing step (split by read group). I understand that it is quite convenient to add this argument, but I would suggest that until it can be develop.
1. Last, bu quite important for me as a developer, I don't have time to spend looking at that engine-level features required to include that argument. I would definitely use some of my time on such a feature if it wasn't possible to workaround the use case of tumor-normal data, but my previous suggestion is enough until someone (even myself) can spend some time on developing the feature.

All this said, I am really interested in getting the indel realignment pipeline out in GATK4, and that's why I am implementing it. If the only way is adding support for every extra-feature of the tools, I have no other chance than doing it, but I can't promise that my schedule allows me to do it soon. I already started porting `IndelRealigner` without some features, and thus I think that if the n-way out can wait a bit, I can have the tool sooner than if I need to spend some time on working around the problems at the engine level.

sooheelee:
@magicDGS, after much discussion with @droazen, we will make this test set even smaller. Also, the current test set will be temporary, to unimpede you, until updates to the main test set can be made (discussion to start next week). @cmnbroad says we need to keep an eye out for coverage in the tests, given the need to make this dataset extremely small.

So I removed chr17 from the mini-reference, such that only four small snippets of reference from various chromosomes remained. However, it appears there are only two usable indel realignments going on with GATK3. Note thate these are targeted exome samples with comparatively low coverage.

[for_magicDGS.zip](https://github.com/broadinstitute/gatk/files/1820785/for_magicDGS.zip)

Note there are two samples, so as to enable testing nWayOut, and a artificial reference `hg38_Shl01`. The two sites to hone in on are  chr11:177568 and chr11:207134. Note also that the BAMs are in an invalid state according to ValidateSamFile. However, GATK3 RealignerTargetCreator and IndelRealigner did not seem to mind. I think these tools should allow processing of BAMs in any validation state. Good luck @magicDGS.

sooheelee:
Let me know if you need me to rework the test data @magicDGS.

sooheelee:
I'm leaving the Broad @magicDGS. So, I won't be able to follow up on this. Best wishes to you.

serge2016:
Dear colleagues, any news? I am very interested in this feature...

cmnbroad:
I'm not aware of any recent activity on this.

droazen:
Closing this PR as abandoned -- we can resurrect it in the future if someone is willing to take over the branch and do the work needed to push it across the finish line.

PR #3113, opened 2017-06-14
Implement PathSeqBwaSpark tool and tests

mwalker174:
Implements a Bwa Spark tool for the PathSeq pipeline. 

Tool input:
1) BAM of paired reads
2) BAM of unpaired reads
3) Bwa index image file

Output:
1) BAM of paired alignments
2) BAM of unpaired alignments

Notes:
- The tool does not generate secondary/supplementary alignments. 
- Alternate alignments are written to the SA tag. 
- Only the sequences for which there was at least 1 alignment are written to the BAM headers (instead of writing all sequences in the reference, which is 10,000's for the pathogen database and substantially increases the run time of the subsequent scoring tool).


tedsharpe:
This seems fine.  I'll make a few suggestions, but nothing important.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
```diff
@@ -0,0 +1,100 @@
...
+ * If preserveAlignments is set, existing alignment data in each read is not overwritten, and any new alignments
+ * are added to the SA tag.
+ */
+public final class PSBwaAligner {
+
+    private final BwaMemIndex bwaIndex;
+    private final int minSeedLength, numThreads, maxAltHits, scoreThreshold;
+    private final boolean pairedAlignment;
+
+    public PSBwaAligner(final PSBwaArgumentCollection bwaArgs, final boolean pairedAlignment) {
```
tedsharpe: Might as well just save the bwaArgs collection, rather than copying each value into this class.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
```diff
@@ -0,0 +1,100 @@
...
+        while (itr.hasNext()) {
+            reads.add(itr.next());
+        }
+        final int numReads = reads.size();
+        if (pairedAlignment && numReads % 2 != 0) {
+            throw new UserException.BadInput("Expected paired reads but there are an odd number");
+        }
+
+        //Align read sequences
+        final List<Tuple2<GATKRead, List<BwaMemAlignment>>> results;
```
tedsharpe: Rather than preparing this list, why don't you prepare your final output list directly by revising your input list.  Just refactor the code that starts on line 81 into a separate function that maps a GATKRead and list of alignments into a (possibly new) GATKRead.  On line 71 just modify the input list with the result of that call.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
```diff
@@ -0,0 +1,100 @@
...
+        if (numReads == 0) {
+            results = Collections.emptyList();
+        } else {
+            final List<byte[]> seqs = reads.stream().map(GATKRead::getBases).collect(Collectors.toList());
+            final List<List<BwaMemAlignment>> alignments = aligner.alignSeqs(seqs);
```
tedsharpe: This also works, and would save you from having to build the ```seqs``` list.
```
alignments = aligner.alignSeqs(reads, GATKRead::getBases);
```
But it would call GATKRead::getBases twice on each read.  I'm not sure which way would be faster, but my suggestion would definitely save memory.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
```diff
@@ -0,0 +1,100 @@
...
+            for (int i = 0; i < numReads; i++) {
+                results.add(new Tuple2<>(reads.get(i), alignments.get(i)));
+            }
+        }
+
+        //Map alignments back to the reads
+        final List<String> refNames = bwaIndex.getReferenceContigNames();
+        return Utils.stream(results).map(item -> {
+            final GATKRead read = item._1;
+            SAMRecord rec = read.convertToSAMRecord(header);
```
tedsharpe: Why convert to SAMRecord?  Just create a new GATKRead down below (line 87 and 93) to replace the input read.  Unaligned reads just pass through.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
```diff
@@ -0,0 +1,100 @@
...
+            for (final BwaMemAlignment alignment : alignmentList) {
+                //Only get primary alignments
+                if (SAMFlag.NOT_PRIMARY_ALIGNMENT.isUnset(alignment.getSamFlag())
+                        && SAMFlag.SUPPLEMENTARY_ALIGNMENT.isUnset(alignment.getSamFlag())) {
+                    if (rec.getReadUnmappedFlag()) {
+                        //Record is currently unmapped, so apply first alignment
+                        rec = BwaMemAlignmentUtils.applyAlignment(rec.getReadName(), rec.getReadBases(), rec.getBaseQualities(),
+                                rec.getReadGroup().getId(), alignment, refNames, header, false, false);
+                    } else if (alignment.getRefId() != -1) {
+                        //Record is currently mapped, but if the alignment is not empty then add as an alternate alignment
```
tedsharpe: I don't see how it's possible to enter this branch.  If the read is mapped its refId can't be -1, can it?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
```diff
@@ -0,0 +1,100 @@
...
+                        final String alignmentTag = BwaMemAlignmentUtils.asTag(alignment, refNames);
+                        final String currentTag = rec.getStringAttribute("SA") != null ? rec.getStringAttribute("SA") : "";
+                        rec.setAttribute("SA", currentTag + alignmentTag);
+                    }
+                }
+            }
+            return SAMRecordToGATKReadAdapter.headerlessReadAdapter(rec).copy();
```
tedsharpe: Not sure why you need a copy.  Whenever you're altering a GATKRead, you'll make a new SAMRecord which will be independent of the input, and wrap it with the headerlessReadAdapter method.  The altered GATKRead is already independent.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAlignerSpark.java
```diff
@@ -0,0 +1,41 @@
...
+import org.broadinstitute.hellbender.utils.bwa.BwaMemIndexSingleton;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+/**
+ * Wrapper class for using the PathSeq Bwa aligner class in Spark. Encapsulates closing the index when done.
+ */
+public final class PSBwaAlignerSpark implements AutoCloseable {
+
+    private final JavaSparkContext ctx;
+    private final Broadcast<PSBwaArgumentCollection> bwaArgsBroadcast;
```
tedsharpe: This is too small to broadcast.  It's already Serializable, so you can pass it via lambda capture instead.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAlignerSpark.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public PSBwaAlignerSpark(final JavaSparkContext ctx, final PSBwaArgumentCollection bwaArgs) {
+        this.ctx = ctx;
+        this.bwaArgsBroadcast = ctx.broadcast(bwaArgs);
+    }
+
+    private static JavaRDD<GATKRead> doBwaAlignmentHelper(final JavaRDD<GATKRead> reads,
```
tedsharpe: Don't see how this 1-liner is helpful.  Was it for testing?  Why not just put this code directly into doBwaAlignment.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
```diff
@@ -0,0 +1,66 @@
...
+import org.broadinstitute.hellbender.utils.SerializableFunction;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+import java.util.*;
+
+/**
+ * Utility functions for PathSeq Bwa tool
+ */
+public final class PSBwaUtils {
```
tedsharpe: I assume that extracting only referenced ref sequences is essential because you have so many of them.  Nonetheless, aren't you going to end up with a bunch of BAMs with inconsistent headers?
Sounds really complicated unless it's essential.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Returns list of sequence names that were aligned to at least once in the reads
+     */
+    static List<String> getAlignedSequenceNames(final JavaRDD<GATKRead> reads) {
+        return reads.flatMap(read -> {
+            if (read.isUnmapped() || read.getAssignedContig().equals("*")) Collections.emptySet().iterator();
+            final Set<String> sequenceNames;
+            if (read.hasAttribute("SA")) {
+                final String[] saTokens = read.getAttributeAsString("SA").split(";");
```
tedsharpe: Maybe SA tag parsing could be put into a utility class somewhere?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
```diff
@@ -0,0 +1,100 @@
...
+        if (pairedAlignment && numReads % 2 != 0) {
+            throw new UserException.BadInput("Expected paired reads but there are an odd number");
+        }
+
+        //Align read sequences
+        final List<Tuple2<GATKRead, List<BwaMemAlignment>>> results;
+        if (numReads == 0) {
+            results = Collections.emptyList();
+        } else {
+            final List<byte[]> seqs = reads.stream().map(GATKRead::getBases).collect(Collectors.toList());
```
tedsharpe: Rather than ```Collectors.toList()```, you should probably use ```SVUtils.arrayListCollector(reads.size())```.  More efficient when you know the size of your output list.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
```diff
@@ -0,0 +1,66 @@
...
+                sequenceNames = new HashSet<>(1 + saTokens.length);
+                sequenceNames.add(read.getAssignedContig());
+                for (final String token : saTokens) {
+                    final String[] alignmentTokens = token.split(",", 1);
+                    sequenceNames.add(alignmentTokens[0]);
+                }
+            } else {
+                sequenceNames = Collections.singleton(read.getAssignedContig());
```
tedsharpe: I like putting the short branch first (especially where one branch is a single line of code, like here).  Seems more readable that way to me.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+            return sequenceNames.iterator();
+        }).filter(Objects::nonNull).distinct().collect();
```
tedsharpe: Aren't they already non-null and distinct?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBwaSpark.java
```diff
@@ -0,0 +1,142 @@
...
+     */
+    private void writeBam(final JavaRDD<GATKRead> reads, final String inputBamPath, final boolean isPaired,
+                          final JavaSparkContext ctx, final SAMFileHeader header) {
+
+        //Only retain header sequences that were aligned to
+        reads.cache();
+        final List<String> usedSequences = PSBwaUtils.getAlignedSequenceNames(reads);
+        final List<SAMSequenceRecord> usedSequenceRecords = usedSequences.stream()
+                .map(seqName -> header.getSequence(seqName))
+                .filter(Objects::nonNull)
```
tedsharpe: Shouldn't they be non-null already?  Are we concealing a bug somewhere else?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Returns list of sequence names that were aligned to at least once in the reads
+     */
+    static List<String> getAlignedSequenceNames(final JavaRDD<GATKRead> reads) {
+        return reads.flatMap(read -> {
+            if (read.isUnmapped() || read.getAssignedContig().equals("*")) Collections.emptySet().iterator();
+            final Set<String> sequenceNames;
+            if (read.hasAttribute("SA")) {
+                final String[] saTokens = read.getAttributeAsString("SA").split(";");
+                sequenceNames = new HashSet<>(1 + saTokens.length);
```
tedsharpe: Oops.  I'd intended to make a comment here, and forgot to initially.
You've made your HashSet too small.  You need to inflate the size of HashMaps and HashSets to account for the max load factor.  There's a function SVUtils.hashMapCapacity that does this for you.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
mwalker174: Done. Looks much cleaner this way too.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
mwalker174: Great - done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
mwalker174: If the original read is already mapped (ie in the input bam), it's possible the aligner doesn't come up with any alignments in which case it returns an alignment with refId -1. 

This isn't likely in the current tool implementation because it checks that the input BAM header doesn't have any sequences, but could be in future versions if we want to be able to align to multiple reference files in the pipeline. 

I added a comment to explain.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAlignerSpark.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAlignerSpark.java
mwalker174: My intention was basically to encapsulate the lambda in a static function. In order to do that, I needed a non-static method as well that would grab the bwaArgs.

For some reason I have the impression that calling lambdas in Spark will force Spark to capture `this`. Is that right? I think this came up in my first code review and I encountered a reason for this somewhere but have since forgotten.

In any case, I've refactored doBwaAlignment to be static and take in bwaArgs.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
mwalker174: It's essential. For some reason GATK is very slow at reading header sequences. This optimization cuts the final Score tool run time from ~5 min to ~30 seconds.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
mwalker174: Factored getting the sequence names for readability.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
mwalker174: Yes. Removed filter() and distinct()

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAligner.java
tedsharpe: Duh, yes.  I get it.  I looked and looked at that line of code and read it as checking the contig ID in the read rather than the alignment every time.  My mistake.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAlignerSpark.java
tedsharpe: Capturing ```this``` (and thus forcing it to be serialized) happens when the lambda references a field from ```this```.  Avoid it by copying the field to a temporary final variable.
Example of the bad thing:
```
class Foo {
   private final GiantKitchenSinkFullOfGarbage = new GiantKitchenSinkFullOfGarbage();
   private final int minMapQ = 47;
   JavaRDD<GATKRead> filter( final JavaRDD<GATKRead> reads ) {
        return reads.filter(read -> read.getMappingQuality() >= minMapQ);
    }
}
```
Fixed:
```
class Foo {
   private final GiantKitchenSinkFullOfGarbage = new GiantKitchenSinkFullOfGarbage();
   private final int minMapQ = 47;
   JavaRDD<GATKRead> filter( final JavaRDD<GATKRead> reads ) {
        final int tmpMinMapQ = minMapQ;
        return reads.filter(read -> read.getMappingQuality() >= tmpMinMapQ);
    }
}
```

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBwaSpark.java
mwalker174: I hope not. The nonNull is there in case header.getSequence() returns null, which should only happen if a read aligned to a sequence that is not in the input BAM header or reference dictionary. In theory, this could happen if someone deleted header sequences from the input BAM after aligning it. Again, the tool currently forbids this but we may want to allow pre-aligned inputs in the future.

I've added a check to print a warning message if one comes up null so at least we know something is up.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
mwalker174: I always assumed HashSets did that for you... thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaAlignerSpark.java
mwalker174: Thanks for explaining. I'll use that pattern for bwaArgs then.

PR #3115, opened 2017-06-14
Update PathSeqFilterSpark and PathSeqBuildKmers tools

mwalker174:
Adds updated tools for creating the host reference kmer set (PathSeqBuildKmers) and filtering reads that are low-quality, low-complexity, or come from the host (PathSeqFilterSpark). Sorry for the especially large size on this PR.

**PathSeqBuildKmers tool**

Note this has been renamed from PathSeqKmerSpark.

Input:
1) Host reference FASTA
2) False positive probability (0 create a hash set, >0 to create a Bloom filter)
3) Kmer length (1-31)
4) Kmer base indices to mask (optional)

Output:
1) Serialized kmer Hopscotch set (.hss) or Bloom filter (.bfi) file

For each reference record, the tool generates a list of long's containing the canonicalized/masked kmers. The result is a Collection<long[]> variable, which is then converted to either a PSKmerSet (Hopscotch set) or PSKmerBloomFilter, depending on the desired false positive probability. 

The PSKmerSet/BloomFilter classes are basically wrappers for LargeLongHopscotchSet and LongBloomFilter, respectively. They both inherit PSKmerCollection, which provides a contains() function for querying new kmers for set membership and makes loading the kmers for filtering more convenient. These classes also store the kmer size, mask, and false positive probability. They also handle canonicalization/masking on queried kmers.

**PathSeqFilterSpark tool**

Input:
1) Input BAM
2) Host kmer set file (optional)
3) Host reference bwa image (optional)

Output:
1) BAM containing paired reads that still have mates
2) BAM containing unpaired reads / reads whose mates were filtered out
3) Metrics file containing read counts and elapsed wall time at each step (optional)

Filtering steps performed on each read:
- If the user sets the --isHostAligned, the read will first be filtered if it is aligned sufficiently well 
- Alignment info is stripped
- A series of quality filters (same as in the previous version of this tool)
- Kmerized and filtered out if at least a threshold number of kmers are in the host set (default 1)
- Aligned to the host reference and filtered if it maps sufficiently well
- Sequence duplicates are removed

Other:
-Fixed bugginess in very large LongBloomFilters by changing a size variable from int to long. 
- Also realized we can't get away with using just 1 hash function in the Bloom filter. Before, I was using a single 64-bit hash and splitting it into 2 32-bit hashes, then using the hash1 + i*hash2 trick to generate each hash value. I don't think we can do this now because we allow for tables of size >2 billion bits in a single filter, so we need 2 64-bit hashes to use the trick.
-A couple of utility functions have been moved around


src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaFilter.java
```diff
@@ -0,0 +1,83 @@
...
+        while (itr.hasNext()) {
+            reads.add(itr.next());
+        }
+        final int numReads = reads.size();
+        if (bPaired && (numReads & 1) != 0) {
+            throw new GATKException("Cannot do paired alignment with an odd number of reads");
+        }
+
+        //Do alignment
+        final List<byte[]> seqs = reads.stream().map(GATKRead::getBases).collect(Collectors.toList());
```
tedsharpe: Use ```SVUtils.arrayListCollector(reads.size())``` rather than ```Collectors.toList()```.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaFilter.java
```diff
@@ -0,0 +1,83 @@
...
+        final List<List<BwaMemAlignment>> alignments = aligner.alignSeqs(seqs);
+        final List<Tuple2<GATKRead, List<BwaMemAlignment>>> results = new ArrayList<>(numReads);
```
tedsharpe: I'd suggest a similar reorganization as in the last PR:  Just create an output list, move lines 73-80 into a separate method that tests whether a read is cool or not, and either add the read to the output or not on line 67.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -0,0 +1,334 @@
...
+
+    private final JavaSparkContext ctx;
+    private final PSFilterArgumentCollection filterArgs;
+    private final MetricsState metricsState;
+    private JavaRDD<GATKRead> reads;
+    private final SAMFileHeader header;
+
+    public PSFilter(final JavaSparkContext ctx, final PSFilterArgumentCollection filterArgs, final JavaRDD<GATKRead> inputReads,
+                    final SAMFileHeader header) {
+        Utils.nonNull(ctx);
```
tedsharpe: IMHO this 1-arg version of Utils.nonNull is nearly useless.  I'd use the 2-arg version or skip it:
```
Utils.nonNull(ctx, "The JavaSparkContext object cannot be null.");
```
Now at least we can grep the code for that message if we don't get a stack trace.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -0,0 +1,334 @@
...
+        return reads.map(read -> clearReadAlignment(read, header));
+    }
+
+    /**
+     * Returns input read with alignment-related info cleared
+     */
+    private static GATKRead clearReadAlignment(final GATKRead read, final SAMFileHeader header) {
+        final GATKRead newRead = new SAMRecordToGATKReadAdapter(new SAMRecord(header));
+        newRead.setName(read.getName());
+        newRead.setBases(read.getBases());
```
tedsharpe: Potential issue here -- don't know whether it's important or not:  An aligned read that mapped to the negative strand will have had its bases RC'd and its quals reversed.  But after you strip away the alignment info you won't have the SAM flag that indicates that, and so you won't ever be able to recover the original orientation of the read.  Do you care?  I don't know.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -0,0 +1,334 @@
...
+
+    /**
+     * Sets proper pairedness flags
+     */
+    private static Iterator<GATKRead> setPartitionUnpairedFlags(final Iterator<GATKRead> iter, final int readsPerPartitionGuess) {
+        final Tuple2<List<GATKRead>, List<GATKRead>> lists = getPairedAndUnpairedLists(iter, readsPerPartitionGuess);
+        final List<GATKRead> pairedReadsList = lists._1;
+        final List<GATKRead> unpairedReadsList = lists._2;
+        for (final GATKRead unpairedRead : unpairedReadsList) {
+            unpairedRead.setIsPaired(false);
```
tedsharpe: This won't clear any mate info (mate contig, position, and template length) that happens to be hanging around.  Also leaves the firstInPair/secondInPair flags untouched.  Don't know if that matters.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -0,0 +1,334 @@
...
+        }
+        final List<GATKRead> newPartitionList = new ArrayList<>(pairedReadsList.size() + unpairedReadsList.size());
+        newPartitionList.addAll(pairedReadsList);
+        newPartitionList.addAll(unpairedReadsList);
+        return newPartitionList.iterator();
+    }
+
+    /**
+     * Maps partition to a Tuple of two Lists, the first containing the paired reads, the second containing unpaired
```
tedsharpe: Comment doesn't describe action.  Looks like a cut and paste boo-boo.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSKmerUtils.java
```diff
@@ -0,0 +1,176 @@
...
+    public static void writeKryoObject(final Object obj, String uri) {
+        final Output output = new Output(BucketUtils.createFile(uri));
+        final Kryo kryo = new Kryo();
+        kryo.writeObject(output, obj);
+        output.close();
+    }
+
+    public static void writeKmerSet(final String uri, final PSKmerSet set) {
+        String filePath = uri;
+        if (HOPSCOTCH_SET_EXTENSION != null && !uri.toLowerCase().endsWith(HOPSCOTCH_SET_EXTENSION.toLowerCase())) {
```
tedsharpe: Seems odd to be testing a static final String for null and lower casing it.  It's non-null and lower case for all time, right?  Could you put in a static initializer that checks to make sure that it's defined properly?  Or was this here just for testing or something?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSKmerUtils.java
```diff
@@ -0,0 +1,176 @@
...
+            filePath = filePath + HOPSCOTCH_SET_EXTENSION;
+        }
+        writeKryoObject(set, filePath);
+    }
+
+    public static void writeKmerBloomFilter(final String uri, final PSKmerBloomFilter bloomFilter) {
+        String filePath = uri;
+        if (BLOOM_FILTER_EXTENSION != null && !uri.toLowerCase().endsWith(BLOOM_FILTER_EXTENSION.toLowerCase())) {
```
tedsharpe: same comment as above

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
```diff
@@ -57,24 +51,6 @@ public static void logItemizedWarning(final Logger logger, final Collection<Stri
     }
 
     /**
-     * Writes two objects using Kryo to specified local file path.
```
tedsharpe: Seven lines above this point (old code, so I can't comment there), you're doing a string concatenation in a loop:
```str += acc + ", "```
You should use a StringBuilder for situations like this.

File: src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java [hunk 10]
```diff
@@ -107,6 +110,10 @@ public static long getOptimalNumberOfBits(final long numElements, final double f
         return (long) Math.ceil(-numElements * Math.log(fpp) / (Math.log(2) * Math.log(2)));
     }
 
+    public double getTheoreticalFPP(final long numElements) {
+        return Math.pow(1.0 - Math.pow(1.0 - (1.0/totalBits), numHashes * numElements), numHashes);
+    }
+
```
tedsharpe: A few lines below here, on line 125 you have:
```sum += bitCountMap[0 & b];```
That's clearly wrong -- 0 & b is always 0.

File: src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilterTest.java [hunk 11]
```diff
@@ -0,0 +1,243 @@
...
+            final GATKRead read = ArtificialReadUtils.createArtificialRead(bases, qual, "*");
+            read.setName(name);
+            return read;
+        }).collect(Collectors.toList());
+    }
+
+    final static List<GATKRead> getPairedReadsFromFastq(final File fastqFile1, final File fastqFile2) {
+        final List<GATKRead> readList1 = getReadsFromFastq(fastqFile1);
+        final List<GATKRead> readList2 = getReadsFromFastq(fastqFile2);
+        readList1.stream().forEach(read -> {
```
tedsharpe: you can leave out ```stream()```:  collection has a forEach method.  (same comment 4 lines below.)

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSKmerUtilsTest.java
```diff
@@ -0,0 +1,172 @@
...
+    public void testLongArrayCollectionToSet() {
+        final int numLists = 1000;
+        final int listSize = 1000;
+        final Collection<long[]> longArrays = new ArrayList<>(numLists);
+        final Set<Long> truthSet = new HashSet<>(numLists * listSize);
+        final Random rand = new Random(SEED);
+        for (int i = 0; i < listSize; i++) {
+            final List<Long> longList = rand.longs(listSize).map(val -> val >>> 2).boxed().collect(Collectors.toList());
+            longArrays.add(longList.stream().mapToLong(Long::longValue).toArray());
+            truthSet.addAll(longList.stream().collect(Collectors.toList()));
```
tedsharpe: can't you just:
```truthSet.addAll(longList);```

File: src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSKmerUtilsTest.java [hunk 13]
```diff
@@ -0,0 +1,172 @@
...
+    public void testLongArrayCollectionToBloomFilter() {
+        final int numLists = 1000;
+        final int listSize = 1000;
+        final Collection<long[]> longArrays = new ArrayList<>(numLists);
+        final Set<Long> truthSet = new HashSet<>(numLists * listSize);
+        final Random rand = new Random(SEED);
+        for (int i = 0; i < listSize; i++) {
+            final List<Long> longList = rand.longs(listSize).map(val -> val >>> 2).boxed().collect(Collectors.toList());
+            longArrays.add(longList.stream().mapToLong(Long::longValue).toArray());
+            truthSet.addAll(longList.stream().collect(Collectors.toList()));
```
tedsharpe: ditto

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBuildKmersSparkIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+        final String libraryPath = "src/test/resources/" + PathSeqBuildKmers.class.getPackage().getName().replace(".", "/") + "/hg19mini.hss";
+        final File expectedFile = new File(libraryPath);
+        final File ref = new File(hg19MiniReference);
+        final File output = createTempFile("test", ".bfi");
+        if (!output.delete()) {
+            Assert.fail();
+        }
+        final ArgumentsBuilder args = new ArgumentsBuilder();
+        args.addFileArgument("referencePath", ref);
+        args.addArgument("bloomFalsePositiveProbability", new Double(BLOOM_FPP).toString());
```
tedsharpe: Double.toString(BLOOM_FPP)

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/ContainsKmerReadFilter.java
```diff
@@ -1,33 +1,52 @@
...
-    private int kSize;
+    private static PSKmerCollection kmerLib = null;
+    private final int kSize, kmerCountThreshold;
 
-    public ContainsKmerReadFilter(final Set<SVKmer> kmer_lib, int kmer_size) {
-        kmerLib = kmer_lib;
-        kSize = kmer_size;
+    public ContainsKmerReadFilter(final String kmerLibPath, final int kmerCountThreshold) {
+        this.kmerCountThreshold = kmerCountThreshold;
+        if (kmerLib == null) {
```
mwalker174: Done. I was actually getting infrequent null pointer exceptions on giant BAMs in Spark mode. I'll bet this was the culprit.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/ContainsKmerReadFilterSpark.java
```diff
@@ -0,0 +1,25 @@
+package org.broadinstitute.hellbender.tools.spark.pathseq;
+
+import org.apache.spark.api.java.function.Function;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+/**
+ * Spark version of ContainsKmerReadFilter
+ */
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/ContainsKmerReadFilterSpark.java
```diff
@@ -0,0 +1,25 @@
...
+    private transient ContainsKmerReadFilter filter;
+
+    public ContainsKmerReadFilterSpark(final String kmerSetPath, final int kmerCountThreshold) {
+        this.kmerSetPath = kmerSetPath;
+        this.kmerCountThreshold = kmerCountThreshold;
+    }
+
+    @Override
+    public Boolean call(final GATKRead read) {
+        if (filter == null) filter = new ContainsKmerReadFilter(kmerSetPath, kmerCountThreshold);
```
mwalker174: Very slick. I think I figured it out.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaFilter.java
mwalker174: Code removed by using the other alignSeqs(), but noted.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaFilter.java
mwalker174: Yes looks much better. Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
mwalker174: This could change things with downstream bwa alignment. Thanks for pointing this out - I've added a check now that sets the read back to its original orientation.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
mwalker174: What I decided on was to run all the reads back through clearReadAlignment() to clean that stuff up. It's a little hacky but at this point we're down to a relatively small number of reads where I'm not concerned about performance. (I thought it was written this way but I must have done this on another branch somewhere and forgotten, I've added it now).

Also realized that I should only set firstInPair/secondInPair in clearReadAlignment() if the read is paired so I've added a check for that too.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSKmerUtils.java
```diff
@@ -0,0 +1,176 @@
...
+        logger.info("Generating kmers from " + totalBases + " bases in " + records.size() + " records...");
+        final ProgressCounter counter = new ProgressCounter(totalBases, 1e6, "Mbp", logger);
+
+        //Get kmers from each reference record
+        for (final String recName : records.keySet()) {
+
+            //Kmerize the record
+            final byte[] bases = records.get(recName).getBases();
+            final long[] list = SVKmerizer.stream(bases, kSize, kSpace, new SVKmerShort(kSize))
+                    .mapToLong(kmer -> new Long(PSKmerCollection.canonicalizeAndMask((SVKmerShort) kmer, kSize, mask)))
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSKmerUtils.java
mwalker174: I suppose I was being overly-cautious here. I'll compromise and just assume it's non-null at least. What I really need is endsWithIgnoreCase().

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSKmerUtils.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
mwalker174: This got moved to PSUtils.java. Using String.join() instead since it's fewer lines.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/LongBloomFilter.java
mwalker174: Got rid of `0 &` and added a small unit test for this.

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilterTest.java
mwalker174: Done

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSKmerUtilsTest.java
mwalker174: Yep, no idea what happened there

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSKmerUtilsTest.java
mwalker174: Done

src/test/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBuildKmersSparkIntegrationTest.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -1,268 +1,115 @@
...
+        header.setSequenceDictionary(new SAMSequenceDictionary());
+
+        try (final PSFilter filter = new PSFilter(ctx, filterArgs, getReads(), header)) {
+
+            final Tuple2<JavaRDD<GATKRead>, JavaRDD<GATKRead>> result = filter.doFilter();
+            final JavaRDD<GATKRead> pairedReads = result._1;
+            final JavaRDD<GATKRead> unpairedReads = result._2;
+
+            if (!pairedReads.isEmpty()) {
+                final SAMFileHeader.SortOrder oldSortOrder = header.getSortOrder();
```
tedsharpe: I thought you had already grouped the mate pairs together in the pairedReads list.  In which case, don't you want the writer to leave the order unperturbed, rather than doing what I fear is a global sort on query name?  And restoring the oldSortOrder at the end -- doesn't that leave you with a lie?  Could you just set the sort order as "unsorted" or "unknown" and leave it that way?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -1,268 +1,115 @@
...
+                header.setSortOrder(SAMFileHeader.SortOrder.queryname);
+                writeReads(ctx, outputPath + ".paired.bam", pairedReads, header);
+                header.setSortOrder(oldSortOrder);
+            } else {
+                logger.info("No paired reads to write - BAM will not be written.");
+            }
+            if (!unpairedReads.isEmpty()) {
+                writeReads(ctx, outputPath + ".unpaired.bam", unpairedReads, header);
```
tedsharpe: Don't you need to dink around with the sort order specifier in the header, just like for the paired case?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -0,0 +1,334 @@
...
+        final long hashForward = fnvByteArray64(1099511628211L, bases);
+        SequenceUtil.reverseComplement(bases);
+        final long hashReverse = fnvByteArray64(1099511628211L, bases);
+        return new Tuple2<>(Math.min(hashForward, hashReverse), read);
+    }
+
+    /**
+     * 64-bit FNV-1a hash for byte arrays
+     */
+    private static long fnvByteArray64(long start, final byte[] toHash) {
```
tedsharpe: Maybe stick this into SVUtils with the other FNV methods? and provide a one-arg overload that encapsulates the 1099511628211L?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -0,0 +1,334 @@
...
+                    }
+                    return iter.iterator().next();
+                });
+    }
+
+    /**
+     * Pairs reads with canonical Long ID using the lesser 64-bit hash of the sequence and its reverse complement
+     */
+    @VisibleForTesting
+    static Tuple2<Long, GATKRead> canonicalizeRead(final GATKRead read) {
```
tedsharpe: Would you get better de-duplication if you only required agreement on the first, say, 31 bases of the read?  There are often quite a few low-quality sequencing errors toward the end of the read.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
mwalker174: Good point that we don't have to sort by query name here, but the subsequent tool, PathSeqBwaSpark, requires that input read pairs be on the same partition, and the GATK bam reader will only do this if it is query name sorted.

A pipeline tool in a future PR will avoid all this.

I'm not sure what oldSortOrder was for but I'm deleting it.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
mwalker174: Yes. Unpaired reads don't have the issue above, so I'm setting it to unsorted.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
mwalker174: This is a good idea. Once I have the tool fully validated I'll keep this in mind as an improvement.

tedsharpe: When I was writing the assembler, we made a joint kmer out of the first few (31, I think) bases from each of the mates.  (Keeping in mind that reads mapped to the negative strand are reversed in the BAM.)  We calculated an RC-invariant hash of the joint kmers and used that as our duplication criterion.  Sounds a little seat-of-the-pants, but it's fast and seemed to be pretty effective in practice.

mwalker174:
@tedsharpe Thank you for your, as always, thoughtful review!

@cwhelan I'm assuming you delegated this one to Ted or were you planning to review as well?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -0,0 +1,330 @@
...
+                pairedReadsList.add(unpairedReads.remove(readName));
+            } else {
+                unpairedReads.put(readName, read);
+            }
+        }
+        //Get the unpaired reads out of the hashmap
+        final List<GATKRead> unpairedReadsList = new ArrayList<>(unpairedReads.values());
+
+        //Minimize unpairedReads memory footprint (don't rely on readsPerPartitionGuess)
+        final List<GATKRead> pairedReadsListResized = new ArrayList<>(pairedReadsList.size());
```
cwhelan: You could use the same `trimToSize` method on `ArrayList` we talked about in your previous PR.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilterArgumentCollection.java
```diff
@@ -0,0 +1,100 @@
...
+            fullName = "minIdentity",
+            minValue = 1,
+            optional = true)
+    public int minIdentity = 30;
+    @Argument(doc = "Host kmer count threshold.",
+            fullName = "hostKmerThreshold",
+            minValue = 1,
+            optional = true)
+    public int hostKmerThresh = 1;
+    @Argument(doc = "The bwa mem index image file of the host reference, distributed on each executor.",
```
cwhelan: This seems to be a point of confusion with users trying to run the other tools that require BWA indices, so I'd be very explicit here that this needs to be a path on the local filesystem of each worker node in the cluster.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
mwalker174: Forgot I had done that here - done.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilterArgumentCollection.java
mwalker174: Done

PR #3122, opened 2017-06-16
Restore ability to process bams with repeated reads

davidbenjamin:
External bams sometimes use faulty adapter trimming algorithms that leave a few identical, repeated reads in a bam (i.e. not PCR duplicates but exactly the same read name etc).  While these bams are faulty there is no reason not to be able to handle them, as we could as of GATK 3.6.  This patch prevents an error without changing how we process good bams.


davidbenjamin:
@droazen this is what I was talking about.

Issue #3123, opened 2017-06-16
Update StandardSomaticAnnotation to reflect M2's default annotations, currently under M2ArgumentCollection as annotation


sooheelee:
So that the StandardSomaticAnnotation can be used by Mutect2's `--group` argument instead of each individually under `--annotation`. Opening this ticket as a result of separate discussion with @davidbenjamin and @droazen.

Here's what is in StandardSomaticAnnotation currently:
![screenshot 2017-06-16 16 50 50](https://user-images.githubusercontent.com/11543866/27244317-007cb358-52b4-11e7-89f0-3d7f119d677e.png)

As a note. HaplotypeCaller uses the `--group` option for two standard annotation sets. I was wondering what would happen if I provided M2 with `--group StandardHCAnnotation` or `--group StandardAnnotation`.

PR #3124, opened 2017-06-16
Shl Correct and expand the M2ArgumentCollection documentation

sooheelee:
I also removed the obsolete errorProbability variable line of code in the SomaticGenotypingEngine.java and noted this argument is deprecated in the M2ArgumentCollection.

Somewhat relatedly, see request in #3123.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
@@ -34,13 +34,13 @@ public class M2ArgumentCollection extends AssemblyBasedCallerArgumentCollection 
     /**
      * A panel of normals can be a useful (optional) input to help filter out commonly seen sequencing noise that may appear as low allele-fraction somatic variants.
      */
-    @Argument(fullName="normal_panel", shortName = "PON", doc="VCF file of sites observed in normal", optional = true)
+    @Argument(fullName="normal_panel", shortName = "PON", doc="VCF file of sites observed in normal.", optional = true)
     public FeatureInput<VariantContext> pon;
 
     /**
      * A resource, such as gnomAD, containing population allele frequencies of common and rare variants.
      */
-    @Argument(fullName="germline_resource", doc="Population vcf of germline sequencing containing allele fractions", optional = true)
+    @Argument(fullName="germline_resource", doc="Population vcf of germline sequencing containing allele fractions.", optional = true)
     public FeatureInput<VariantContext> germlineResource;
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -62,54 +65,72 @@
...
      */
-    @Argument(fullName = "minNormalVariantFraction", optional = true, doc = "Minimum number of reads in pileup to be considered active region.")
+    @Argument(fullName = "minNormalVariantFraction", optional = true, doc = "Minimum fraction of variant reads in normal pileup to be considered germline.")
     public double minNormalVariantFraction = 0.1;
 
     /**
-     * Only variants with tumor LODs exceeding this threshold can pass filtering.
+     * Only variants with tumor LODs exceeding this threshold will be written to the VCF, regardless of filter status.
+     * Set to less than or equal to tumor_lod. Increase argument value to reduce false positives in the callset.
+     * Default setting of 3 is permissive and will emit some amount of false-positives that
```
davidbenjamin: I worry people will read too quickly and misinterpret this to say that by default M2 has a false positive problem.  Could you rewrite it without the words "false positives."

Also "filtering should filter" reads a little roughly to my ears.  Maybe "{@link FilterMutectCalls} should filter."

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -62,54 +65,72 @@
...
+     * filtering should filter and that allows for downstream training outside of the current workflow.
      */
-    @Argument(fullName = "tumor_lod_to_emit", optional = true, doc = "LOD threshold for emit tumor variant")
+    @Argument(fullName = "tumor_lod_to_emit", optional = true, doc = "LOD threshold to emit tumor variant to results.")
```
davidbenjamin: IMO, "to results" --> "to output vcf"

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -62,54 +65,72 @@
...
+    @Argument(fullName = "normal_lod", optional = true, doc = "LOD threshold for calling normal variant non-germline.")
     public double NORMAL_LOD_THRESHOLD = 2.2;
 
     /**
-     * This argument is used for the M1-style strand bias filter
+     * @deprecated This argument is obsolete as of June 2017. It was previously used for the M1-style strand bias filter.
      */
-    @Argument(fullName="power_constant_qscore", doc="Phred scale quality score constant to use in power calculations", optional = true)
+    @Deprecated
+    @Argument(fullName="power_constant_qscore", doc="Phred scale quality score constant to use in power calculations.", optional = true)
```
davidbenjamin: Instead of deprecating, you may delete this argument along with the line `final double errorProbability = QualityUtils.qualToErrorProb(MTAC.POWER_CONSTANT_QSCORE);` in `SomaticGenotypingEngine`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -62,54 +65,72 @@
...
     public int POWER_CONSTANT_QSCORE = 30;
 
     /**
-     * Which annotations to add to the output VCF file. See the VariantAnnotator -list argument to view available annotations.
+     * Which annotations to add to the output VCF file. By default the tool adds all annotations.
```
davidbenjamin: By default the tool adds all *of the following* annotations.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
davidbenjamin: . . . which, I see, you already did.

davidbenjamin:
Back to @sooheelee.  Thanks for doing this!

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -34,26 +34,29 @@
...
-    @Argument(fullName="af_of_alleles_not_in_resurce", shortName = "default_af",
+    @Argument(fullName="af_of_alleles_not_in_resource", shortName = "default_af",
             doc="Population allele fraction assigned to alleles not found in germline resource.  A reasonable value is" +
                     "1/(2* number of samples in resource) if a germline resource is available; otherwise an average " +
                     "heterozygosity rate such as 0.001 is reasonable.", optional = true)
     public double afOfAllelesNotInGermlineResource = 0.001;
 
     /**
-     * Prior probability that any given site has a somatic allele
+     * Prior probability that any given site has a somatic allele. Impacts germline probability calculation.
```
sooheelee: Specified "Prior log-10 probability..."

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -34,26 +34,29 @@
...
+     * For example, -6 translates to one in a million or ~3000 somatic mutations per human genome.
+     * Depending on tumor type, mutation rate ranges vary (Lawrence et al. Nature 2013), and so adjust parameter accordingly.
+     * For higher expected rate of mutation, adjust number up, e.g. -5. For lower expected rate of mutation, adjust number down, e.g. -7.
```
sooheelee: Added two sentences "The workflow uses this parameter only towards the germline event filter. It does NOT relate to the LOD threshold."

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -62,54 +65,72 @@
...
+     * setting minimum variants in pileup to zero forces the tool to consider all provided regions as active.
      */
-    @Argument(fullName = "tumorStandardDeviationsThreshold", optional = true, doc = "How many standard deviations above the expected number of variant reads due to error we require for a tumor pielup to be considered active.")
+    @Argument(fullName = "tumorStandardDeviationsThreshold", optional = true, doc = "How many standard deviations above the expected number of variant reads due to error we require for a tumor pileup to be considered active.")
     public int tumorStandardDeviationsThreshold = 2;
 
     /**
-     * Minimum allele fraction of variant reads in normal for a pileup to be considered inactive
+     * Minimum fraction of variant reads in normal for a pileup to be considered inactive. Applies to normal data in a tumor with matched normal analysis.
+     * For value of 0.1, at least one tenth of pileup must be variant for tool to consider it a germline variant site and therefore not a region of interest
```
sooheelee: Changed to "therefore not a locus of interest". I use _region of interest_ (ROI) out of habit from my bench days.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -62,54 +65,72 @@
...
+     * in the somatic analysis.
      */
-    @Argument(fullName = "minNormalVariantFraction", optional = true, doc = "Minimum number of reads in pileup to be considered active region.")
+    @Argument(fullName = "minNormalVariantFraction", optional = true, doc = "Minimum fraction of variant reads in normal pileup to be considered germline.")
```
sooheelee: Changed to "Minimum fraction of variant reads in normal pileup to be considered a germline variant site and thus not of further interest."

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
sooheelee: Sure. Smoothed to "Default setting of 3 is permissive and will emit some amount of negative training data that {@link FilterMutectCalls} should then filter."

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
sooheelee: I try to avoid words inputs and outputs. How about we simplify this back to "LOD threshold to emit tumor variant to VCF."?

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
sooheelee: Yes, I leave the deprecated parameter left in so folks don't wonder if they had imagined the argument for prior versions. Then after a while, perhaps in the next iteration of tool changes, we can remove this.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
sooheelee: amended

sooheelee:
I've incorporated your feedback @davidbenjamin. Many thanks for the review. Waiting for checks to pass.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
davidbenjamin: Sure, that's fine with me.

davidbenjamin:
@sooheelee Looks good to me. You can merge once tests pass.

Issue #3131, opened 2017-06-20
SVFastqRead with reversed strand mapped reads.


vruano:
Fastq files writen using this ulitity contains reads with bases and qualities as they appear in the original aligned bam file. 

Reads that map to reverse strand have their based changed to the reverse complement with respect with the original sequenced bases. This results in a lost of information where we do not know whether the read in the .fastq file generated contains either the original base calls or their reverse complement. 

We need to fix that.

SHuang-Broad:
Should this be closed now, after #3133 ?

vruano:
yes.

PR #3133, opened 2017-06-20
FastqRead created from GATKRead records reverse their base calls and …

vruano:
…qualities if the original read is mapped against the reverse strand.

Addresses issue #3131.


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
```diff
@@ -5,7 +5,12 @@
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
 import htsjdk.samtools.SAMUtils;
+import org.apache.commons.lang3.ArrayUtils;
+import org.bdgenomics.adam.models.DNAAlphabet;
```
tedsharpe: is this import necessary?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
```diff
@@ -5,7 +5,12 @@
...
 import org.broadinstitute.hellbender.exceptions.GATKException;
+import org.broadinstitute.hellbender.utils.BaseUtils;
+import org.broadinstitute.hellbender.utils.Nucleotide;
```
tedsharpe: is this import necessary?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
```diff
@@ -5,7 +5,12 @@
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
 import htsjdk.samtools.SAMUtils;
+import org.apache.commons.lang3.ArrayUtils;
```
tedsharpe: maybe use htsjdk.samtools.util.SequenceUtil.reverseQualities instead?  seems more mainstream.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
```diff
@@ -26,10 +31,26 @@
         private final byte[] bases;
         private final byte[] quals;
 
+        public FastqRead( final GATKRead read ) {
+            this(Utils.nonNull(read).getName(),
+                 read.isReverseStrand() ? BaseUtils.simpleReverseComplement(read.getBases()) : read.getBases().clone(),
```
tedsharpe: clone is unnecessary: getBases makes a copy, as does getBaseQualities.  so you could simplify this constructor considerably, i think.  maybe something like:
```
public FastqRead( final GATKRead read ) {
    name = Utils.nonNull(read.getName());
    if ( !read.isReverseStrand() ) {
        bases = read.getBases();
        quals = read.getBaseQualities();
    } else {
        bases = BaseUtils.simpleReverseComplement(read.getBases());
        quals = read.getBaseQualities();
        SequenceUtils.reverseQualities(quals);
    }
}
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
```diff
@@ -112,10 +133,16 @@ public FastqRead read( final Kryo kryo, final Input input, final Class<FastqRead
...
         final String nameSuffix = read.isPaired() ? (read.isFirstOfPair() ? "/1" : "/2") : "";
-        String mapLoc = "";
+        final String mapLoc;
         if ( includeMappingLocation ) {
-            if ( read.isUnmapped() ) mapLoc = " mapping=unmapped";
-            else mapLoc = " mapping=" + read.getContig() + ":" + read.getStart() + ";" + read.getCigar().toString();
+            if ( read.isUnmapped() ) {
+                mapLoc = " mapping=unmapped";
+            } else {
+                mapLoc = String.format(" mapping=%s:%d;%s;strand=%s", read.getContig(), read.getStart(), read.getCigar(),
```
tedsharpe: String.format is considerably slower than just gluing together the pieces with +.  But if you find it cleaner, I'm fine with it.  I don't think our code parses this information currently, but I'm guessing you probably will be parsing out the strand info.  If so, please add a method to parse the line to this class so that the formatting and the parsing live together.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
vruano: Ok, I was looking for that actually.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
vruano: Probably not.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
vruano: Probably not.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
vruano: I found it rather cleaner this way. considering that most of the time this is associated with slower IO operations I think is better to use format even if it is less efficient. I wonder whether there would be a "pre-compiled" version such as Pattern.compile's but for string composition rather than parsing.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
vruano: Ok, I see, I guess that I had the old SAMRecord in mind when I wrote it.

vruano:
@tedsharpe, please take another look. I have changed a few things apart from your comments... Now I'm addressing another separated yet related issue (getName returns now the read name and nothing else), and get the tests to pass (hopefully).

vruano:
@tedsharpe, I have changed the implementation to use a lazy non-cached recreation of the read mapping information. 

I have not added the additional mapping information in your branch like the MQ value. I would do that in a separate commit. 

Please take a look and perhaps run your benchmark.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
```diff
@@ -4,49 +4,245 @@
...
+    private static final String MAPPING_FIELD_EQUAL_TO = MAPPING_FIELD_NAME + HEADER_FIELD_EQUAL_CHR;
+    private static final String UNMAPPED_DESCRIPTION_STR = MAPPING_FIELD_EQUAL_TO + UNMAPPED_STR;
+    private static final String HEADER_FIELD_ARRAY_SEPARATOR_STR = "" + HEADER_FIELD_ARRAY_SEPARATOR_CHR;
+    private static final Pattern MAPPING_DESCRIPTION_PATTERN = Pattern.compile("^" + MAPPING_FIELD_EQUAL_TO +
+            "(.+):(\\d+);(" + Strand.PATTERN.pattern() + ");(.+)$");
+    private static final Pattern FASTQ_READ_HEADER_PATTERN = Pattern.compile("^" + HEADER_PREFIX_CHR +
+            "([^" + HEADER_FIELD_SEPARATOR_REGEXP +  "]+)(" + HEADER_FIELD_SEPARATOR_REGEXP + "(.*))?$");
+
+    // TODO htsjdk has it own Strand annotation enum. These classes could be merged if that Strand would me updated
+    // TODO so that one can get the enum constant char encoding; currently one can only do the transformation the other way.
```
tedsharpe: There's an enum called org.broadinstitute.hellbender.tools.spark.sv.BreakpointEvidence.ReadEvidence.TemplateEnd
that might be helpful.  (You could move it to a top-level class if you'd like.)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -373,7 +373,7 @@ public AlignedAssemblyOrExcuse apply( final Tuple2<Integer, List<SVFastqUtils.Fa
             if ( fastqDir != null ) {
                 final String fastqName = String.format("%s/assembly%06d.fastq",fastqDir,intervalAndReads._1());
                 final ArrayList<SVFastqUtils.FastqRead> sortedReads = new ArrayList<>(intervalAndReads._2());
-                sortedReads.sort(Comparator.comparing(SVFastqUtils.FastqRead::getName));
+                sortedReads.sort(Comparator.comparing(SVFastqUtils.FastqRead::getId));
```
tedsharpe: Compare on ::getHeader instead for efficiency.  We don't want to be tokenizing and substringing with each key comparison.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadsForQNamesFinder.java
```diff
@@ -44,7 +44,7 @@ public ReadsForQNamesFinder( final HopscotchUniqueMultiMap<String, Integer, QNam
                 if ( FastqRead == null ) {
                     final String readName =
                             dumpFASTQs ? SVFastqUtils.readToFastqSeqId(read, includeMappingLocation) : null;
```
tedsharpe: Kill this line and the function it calls.  It's no longer necessary.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
```diff
@@ -4,49 +4,245 @@
...
         private final byte[] bases;
         private final byte[] quals;
 
-        public FastqRead( final String name, final byte[] bases, final byte[] quals ) {
-            this.name = name;
+        public FastqRead( final GATKRead read ) {
+            this(Utils.nonNull(read).getName(),
+                 read.isPaired() ? OptionalInt.of(read.isFirstOfPair() ? 1 : 2) : OptionalInt.empty(),
+                 read.isReverseStrand() ? BaseUtils.simpleReverseComplement(read.getBases()) : read.getBases(),
+                 reverseIf(read.isReverseStrand(), read.getBaseQualities()), new Mapping(read));
```
tedsharpe: You need to handle the unmapped read case:
```read.isUnMapped() ? null : new Mapping(read)```

File: src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java [hunk 10]
```diff
@@ -4,49 +4,245 @@
...
     public static final class FastqRead implements FermiLiteAssembler.BasesAndQuals {
-        private final String name;
+        private final String header;
         private final byte[] bases;
         private final byte[] quals;
 
-        public FastqRead( final String name, final byte[] bases, final byte[] quals ) {
-            this.name = name;
+        public FastqRead( final GATKRead read ) {
+            this(Utils.nonNull(read).getName(),
```
tedsharpe: I'd rather see simple inline code here to initialize the three fields correctly, rather than calling another constructor with a bunch of ternary operators.  You can then also eliminate reverseIf, composeHeaderLine and the next two constructors that are otherwise unused.  I think it would be much clearer that way.  But this is a matter of personal taste, I suppose, so if you like it this way better I won't object.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
vruano: In the SAM format spec these are named 'Fragments' or 'Segments' of the template so perhaps we should use that instead of TemplateEnds?

vruano: Also I guess you mention ```TemplateEnd``` here in relation to /1 /2... outputting rather than the strand + or -, which is what is addressed by the ```enum Strand``` this **TODO** is referring to.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
vruano: Ok.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
vruano: I guess this is bit up in the air... if it is unmapped is-reverse-strand is supposed to be false.... but I guess is better worry than sorry.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
vruano: Ok.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadsForQNamesFinder.java
vruano: Done, also I did a bit of refactoring in respect to the fastqRead initialization conditional within the inner loop.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
tedsharpe: I'm sorry.  Yes, I temporarily left leave of my senses.  Strand and template end are completely unrelated.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/SVFastqUtils.java
tedsharpe: My comment pertained to the ```new Mapping(read)``` part of the line.  Since you allowed for a null Mapping object in composeHeaderLine, I thought you'd want to pass null for an unmapped read.  The way you have it, you'll produce a Mapping with contig=null, start = -1, strand = +, and cigar = null.  This will cause Mapping::toString to blow up, because you don't account for null cigars.

tedsharpe: Maybe it should be an exception to try to create a Mapping from an unmapped read?

vruano:
@tedsharpe what about this one.

tedsharpe:
Yes, this is fine.

vruano:
Last changes in and squashed.

PR #3135, opened 2017-06-20
Use gatk-launch convention and show how to specify java option Xmx in…

sooheelee:
… example commands of previously protected tools. Convention is now `gatk-launch --javaOptions "-Xmx4g"`.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/PadTargets.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/PadTargets.java
@@ -22,7 +22,7 @@ import java.util.List;
  * <p>For whole exome sequencing (WES) targets: </p>
  *
  * <pre>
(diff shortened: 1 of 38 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/PadTargets.java
```diff
@@ -22,7 +22,7 @@
  * <p>For whole exome sequencing (WES) targets: </p>
  *
  * <pre>
- * java -Xmx1g -jar $gatk_jar PadTargets \
+ * gatk-launch --javaOptions "-Xmx4g" PadTargets \
```
droazen: Preserve the original `-Xmx1g` here?

sooheelee: Done. Good catch.

PR #3136, opened 2017-06-20
Rename zipBundle target to bundle, and make it the default build target

droazen:


droazen:
@lbergelson please review

PR #3138, opened 2017-06-20
add small indel evidence

tedsharpe:


tedsharpe:
Boosts discovery of all classes of variants, but nothing dramatic.  Runtime down to 41 minutes.
Contains quite a bit of cleanup (sorry).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindSmallIndelRegions.java
```diff
@@ -0,0 +1,135 @@
...
+
+        for ( final BreakpointEvidence evidence : evidenceList ) {
+            System.out.println(evidence);
+        }
+    }
+
+    public static final class Finder {
+        private static final int BLOCK_SIZE = 250;
+        private static final int EVIDENCE_WEIGHT = 5;
+        private static final float KS_SIGNIFICANCE = 1.E-5f;
```
cwhelan: What was your justification for setting this value? Do you have a plot of what they looked like empirically? We are doing a lot of tests across the genome.

It would likely be tricky or inefficient to implement, but if we could gather the p-values for many of these tests we could use a q-value FDR approach.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/IntHistogram.java
```diff
@@ -0,0 +1,233 @@
...
+        public IntHistogram createEmptyHistogram() {
+            return new IntHistogram(cdfFractions.length - 2);
+        }
+
+        /** Is the specified histogram significantly different (at the p < .01 level) from the CDF. */
+        public boolean isDifferent( final IntHistogram intHistogram ) { return isDifferent(intHistogram, .01f); }
+
+        /** Using the Kolmogorov-Smirnov statistic for two samples:
+         *  Is the specified histogram significantly different from the CDF. */
+        public boolean isDifferent( final IntHistogram sampleHistogram, final float significance ) {
```
cwhelan: Do you have a sense of what size indels are causing the test to return true under your chosen parameters? You mentioned before that you thought it might be ringing the bell even for very small indels.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindSmallIndelRegions.java
tedsharpe: The reasoning was:
We perform ~12million tests (3Gb genome / new window every 250b).  So the expected number of false positives is 120 (12E6 * 1E-5).  We're ringing the bell 13,680 times, so one might expect < 1% of our calls to be FP.  Seemed reasonable to me.
I think we could pretty easily do something far more sophisticated, however:  we could pretty easily bring all 12million window values back to the driver and calculate q-values, or do something Bayesian to try to discover the right parameters.
BTW, BLOCK_SIZE and EVIDENCE_WEIGHT are equally seat-of-the-pants.  I don't think BLOCK_SIZE is too sensitive a parameter, but I don't really know that.  EVIDENCE_WEIGHT ought to be calculated from truth data.

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/IntHistogram.java
tedsharpe: Not yet.  I was beginning to scrutinize that a few days ago, but got diverted by PR reviews.

tedsharpe:
Back to you @cwhelan.

cwhelan:
Looks good to me! I like the way it looks with the new filtering approach, thanks.

PR #3139, opened 2017-06-20
gatherVcfs add block copy for NIO

lbergelson:


src/main/java/org/broadinstitute/hellbender/tools/GatherVcfs.java
```diff
@@ -324,6 +368,10 @@ private static void gatherWithBlockCopying(final List<File> vcfs, final File out
                                 lastByteNewline = thisByteNewline;
                             }
 
+                            if( firstNonHeaderByteIndex == -1 ){
+                                log.warn("Scanned the entire file " + f.toUri().toString() + " and found no variants");
```
droazen: It seems to me that at this point you have not necessarily scanned the entire file

droazen: I suspect that you intended to put this check after the while loop, not within it.

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsIntegrationTest.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+    
+    @Test(dataProvider = "getVcfsToShard")
+    public void testBlockGather(final File vcf, final int numShards) throws IOException {
+        try (final FeatureDataSource<VariantContext> input = new FeatureDataSource<>(vcf)) {
+            final ArrayList<VariantContext> expected = Lists.newArrayList(input);
+            final List<File> shards = scatterVariants(expected, (VCFHeader) input.getHeader(), numShards,
+                                                      createTempDir("vcfshards"));
+
+            final File output = new File("failAgain.vcf.gz");//createTempFile("gathered", ".vcf.gz");
```
droazen: The output here should be a temp file, not a hardcoded filename

droazen:
Review complete -- found 2 minor issues that I will fix myself by pushing into this branch, since @lbergelson is away.

droazen:
Addressed code review comments:

* Pulled `if( firstNonHeaderByteIndex == -1 ){` check out of while loop in `gatherWithBlockCopying()`
* `testBlockGather()` writes output to a temp file, not a hardcoded path
* Corrected docs for `--gatherType` argument.

Will merge once tests pass in travis

droazen:
I've had to temporarily disable `testBlockGather()` to get tests to pass here -- the test passes locally, but appears to run out of memory on travis (gets killed with signal 9). Will open a ticket for this.

PR #3140, opened 2017-06-20
Fix for randomly failing XHMM-related tests

mbabadi:
Still not sure why the tests failed randomly! all XHMM-related tests use their own RNG with fixed seeds and there are no RNG calls in any parallel streams. Therefore, the randomly generated test data must be identical and fully deterministic across all runs. However, it did not appear to be the case! some test runs triggered a bug in HMMPostProcessor (see below) and some runs didn't. I removed a few unnecessary RNGs and the issue is not reproducible anymore. In particular, both XHMMModel and XHMMEmissionProbabilityCalculator had their own RNG but then again, if the tests are run in a deterministic order, it shouldn't matter. The good news is the bug in HMMPostProcessor is fixed; the bad news is, I still don't know why the tests were not deterministic. I bet the failing issue is (magically!) fixed as a result of pulling out the RNG from XHMMModel and XHMMEmissionProbabilityCalculator. If it occurs again, I'll investigate more.

- fixed a bug in HMMPostProcessor that required all samples to be queried in the given list of genotyping segments every time (origin of the failing tests: sometimes the randomly generated genotyping segments contained fewer samples than all samples available for genotyping)
- got rid of the unnecessary RNG in XHMMModel to make it stateless (sampling requires an external RNG)
- also made XHMMEmissionProbabilityCalculator stateless (sampling requires an external RNG)
- truncated the target list used in XHMM integration tests (cuts down the test time by a factor of 10)

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/germlinehmm/xhmm/XHMMEmissionProbabilityCalculator.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/germlinehmm/xhmm/XHMMEmissionProbabilityCalculator.java
@@ -10,6 +10,7 @@ import org.broadinstitute.hellbender.tools.exome.germlinehmm.CopyNumberTriState;
 import org.broadinstitute.hellbender.utils.param.ParamUtils;
 
 import javax.annotation.Nonnull;
+import java.util.Random;
 
 /**
  * Implements the {@link TargetLikelihoodCalculator} interface for the original XHMM-based germline model.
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/germlinehmm/xhmm/XHMMEmissionProbabilityCalculator.java
```diff
@@ -10,30 +10,28 @@
...
         this.deletionMean = ParamUtils.isNegativeOrZero(deletionMean, "Deletion coverage shift must be negative.");
         this.duplicationMean = ParamUtils.isPositiveOrZero(duplicationMean, "Duplication coverage shift must be positive");
         emissionStandardDeviation = ParamUtils.isPositive(emissionStdDev, "Emission standard deviation must be positive");
-        this.rng = rng;
     }
 
     @Override
     public double logLikelihood(@Nonnull final XHMMEmissionData emissionData, final double copyRatio, final Target target) {
-        return new NormalDistribution(rng, getEmissionMean(copyRatio), emissionStandardDeviation)
+        return new NormalDistribution(getEmissionMean(copyRatio), emissionStandardDeviation)
```
samuelklee: According to the documentation, passing null as the rng to NormalDistribution will avoid some overhead.

samuelklee:
Looks good to me except for one minor thing.  Hopefully this does the trick!  Thanks also for cutting down on the test runtime.

droazen:
@mbabadi Can you rebase this onto the latest master (which has a fix for the travis failures) and merge ASAP? Thanks!

src/main/java/org/broadinstitute/hellbender/tools/exome/germlinehmm/xhmm/XHMMEmissionProbabilityCalculator.java
mbabadi: done.

droazen:
M2 WDL tests had a transient failure on this branch -- re-running them.

droazen:
Merging this despite the unrelated "No space left on device" error in the M2 WDL tests, which is definitely unrelated to this change.

PR #3146, opened 2017-06-22
Clarify two uses of "Raw" in CNV workflow

sooheelee:
I'd mistaken the two meanings of RAW in the CNV workflow and have clarified with @samuelklee and on behalf of our users.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
@@ -57,9 +57,8 @@ import java.util.stream.IntStream;
  * </pre>
  *
  * <p>
- *     To collect raw coverage, omit the --transform PCOV option.
- *     "Raw" in the context of the workflow can refer to either (i) raw coverage counts from this step or 
- *     (ii) non-normalized raw proportional coverage counts.
+ *     To collect raw coverage in terms of integer count, i.e. not normalized by the total number of targets,
+ *     use the --transform RAW option.
  * </p>
  * <p>
  *     The interval targets are exome target intervals padded, e.g. with 250 bases on either side.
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -57,8 +57,9 @@
  * </pre>
  *
  * <p>
- *     Panel of Normal (PoN) creation uses the --transform PCOV parameter as shown.
- *     To collect raw coverage for tumor sample copy number analysis, omit this option.
+ *     To collect raw coverage, omit the --transform PCOV option.
```
samuelklee: Perhaps something like: `To collect coverage in terms of integer counts (i.e., not normalized by the total number of targets), omit the --transform PCOV option.

Note: The use of "raw" in subsequent documentation may refer to either (i) coverage in terms of integer counts from this step or (ii) non-denoised coverage in general, which may be in terms of either integer counts or proportional coverage.`

And do we really use (ii) in the javadoc, or does that just happen in the tutorial? If the latter, then I would omit the note entirely. But in any case, I think you want to use the word "integer" somewhere.

samuelklee:
Just some suggestions, thanks!

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
sooheelee: Just the tutorial, which I updated to clarify. So I'll remove this note.

sooheelee:
I've incorporated your feedback @samuelklee. Thanks for the review.

samuelklee:
👍

samuelklee:
@droazen @sooheelee Let's get this doc change into the beta.

sooheelee:
Restarted the travis tests.

Issue #3156, opened 2017-06-23
Update CNV tool docs to clarify log2 status of inputs and outputs


sooheelee:
As requested in the forum by SL.

----
@esalinas The segment means in the .seg file should be non log_2 (this is the output of the GATK CNV tool, as indicated in the javadoc), but the tangent normalized coverages in the tn.tsv should be log_2.

@shlee Perhaps it's worth making this more explicit in the javadoc.

Future versions of the CNV pipeline will only deal with raw (integer) read-count coverage and only output non log_2 copy-ratio estimates, which will hopefully prevent such sources of confusion.

This Issue was generated from your [forums] 
[forums]: http://gatkforums.broadinstitute.org/gatk/discussion/comment/39744#Comment_39744

sooheelee:
The PR is undergoing Travis tests.

PR #3157, opened 2017-06-23
Shl doc explicate log2 for PerformSegmentation and NormalizeSomaticReadCounts

sooheelee:
@samuelklee's request described in #3156.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
@@ -48,9 +48,9 @@ import java.io.File;
  *   --output tumor.seg
  * </pre>
  *
- * While the tangent normalized (tn) data are log2 transformed, as indicated by the --log2Input true option, 
- * the resulting segment means in the .seg file are not. Currently, NormalizeSomaticReadCounts can only produce 
- * log2-transformed counts.
+ * {@link NormalizeSomaticReadCounts}, which calculates pretangent and tangent normalized coverage, can only produce log2-transformed counts.
+ * These example commands show the tangent normalized (tn) coverage data are log2 transformed with the --log2Input true option.
+ * PerformSegmentation accepts either non-log2 (specify --log2Input false) or log2 data and for either case gives segment means as non-log2 values.
  *
  */
 @CommandLineProgramProperties(
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
@@ -48,6 +48,10 @@ import java.io.File;
  *   --output tumor.seg
  * </pre>
  *
+ * While the tangent normalized (tn) data are log2 transformed, as indicated by the --log2Input true option, 
+ * the resulting segment means in the .seg file are not. Currently, NormalizeSomaticReadCounts can only produce 
+ * log2-transformed counts.
+ *
  */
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
```diff
@@ -48,6 +48,10 @@
  *   --output tumor.seg
  * </pre>
  *
+ * While the tangent normalized (tn) data are log2 transformed, as indicated by the --log2Input true option,
```
samuelklee: Perhaps: "While the tangent normalized (tn) input coverage may be log2 transformed, if indicated by the --log2Input true option..."

That is, PerformSegmentation will still work if you give non-log2 input (produced by some other tool besides NormalizeSomaticReadCounts) and specify --log2Input false, and the output will again be non-log2.

src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
```diff
@@ -48,6 +48,10 @@
  *   --output tumor.seg
  * </pre>
  *
+ * While the tangent normalized (tn) data are log2 transformed, as indicated by the --log2Input true option, 
+ * the resulting segment means in the .seg file are not. Currently, NormalizeSomaticReadCounts can only produce
```
samuelklee: Could add Javadoc link to NormalizeSomaticReadCounts.

samuelklee:
@droazen @sooheelee Let's get this doc change into the beta.

src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
sooheelee: I avoid use of "input" "output" in documentation. Also, this description refers specifically to what is happening in the two example commands immediately above. So, how about
```
 * In these example commands, while the tangent normalized (tn) coverage data are log2 transformed, as indicated by the --log2Input true option, 
 * the resulting segment means in the .seg file are not. PerformSegmentation accepts either non-log2 (specify --log2Input false) and log2 data.
 * However, {@link NormalizeSomaticReadCounts}, which calculates pretangent and tangent normalized coverage, can only produce log2-transformed counts.
```

src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
sooheelee: Yup.

src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
sooheelee: No wait, hold on.

sooheelee: How about:
```
* {@link NormalizeSomaticReadCounts}, which calculates pretangent and tangent normalized coverage, can only produce log2-transformed counts.
 * These example commands show the tangent normalized (tn) coverage data are log2 transformed with the --log2Input true option.
 * PerformSegmentation accepts either non-log2 (specify --log2Input false) or log2 data and for either case gives segment means as non-log2 values.
```

sooheelee:
@samuelklee Expanded on your feedback. Let me know if the changes are okay.

sooheelee:
Based on DR's email, I rebased these commits to get the travis tests going.

samuelklee:
👍 Thanks!

Issue #3125, opened 2017-06-19
Add the gcs-connector as a GATK dependency, and write a test showing that GCS access with the Spark local runner works


droazen:
As a stopgap solution to allow `gs://` access on Spark with the local runner, let's add the `gcs-connector` as a project dependency, and craft a test case the runs a simple Spark tool like `PrintReadsSpark` using the local runner with GCS inputs and outputs.

I've already started this in the branch https://github.com/broadinstitute/gatk/compare/dr_fix_gcs_spark_writing, but it's not working yet since the gcs-connector requires some extra authentication-related setup.

droazen:
@jean-philippe-martin Mind taking this one on?

jean-philippe-martin:
Sure I'll take a look!

PR #3159, opened 2017-06-23
Add gcs-connector for Spark unit tests running locally

jean-philippe-martin:
For the tests to work we must define:
HELLBENDER_TEST_PROJECT : Google Project to use
HELLBENDER_JSON_SERVICE_ACCOUNT_KEY : path to a JSON file with service
account credentials

(I've updated the README accordingly)

(We've used both before so they should already be configured)

This fixes issue #3125

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -126,6 +126,10 @@ dependencies {
     compile 'com.google.cloud:google-cloud-nio:0.19.0-alpha:shaded'
     compile 'com.google.cloud.genomics:google-genomics-dataflow:v1beta2-0.15'
     compile 'com.google.cloud.genomics:gatk-tools-java:1.1'
+    // this comes built-in when running on Google Dataproc, but the library
+    // allows us to reads from GCS also when testing locally (or on non-Dataproc clusters,
+    // should we want to)
+    compile group: 'com.google.cloud.bigdataoss', name: 'gcs-connector', version: '1.6.0-hadoop2'
     compile 'org.apache.logging.log4j:log4j-api:2.3'
     compile 'org.apache.logging.log4j:log4j-core:2.3'
     compile 'org.apache.commons:commons-lang3:3.4'
--- a/build.gradle
+++ b/build.gradle
@@ -129,7 +129,7 @@ dependencies {
     // this comes built-in when running on Google Dataproc, but the library
     // allows us to reads from GCS also when testing locally (or on non-Dataproc clusters,
     // should we want to)
-    compile group: 'com.google.cloud.bigdataoss', name: 'gcs-connector', version: '1.6.0-hadoop2'
+    compile group: 'com.google.cloud.bigdataoss', name: 'gcs-connector', version: '1.6.1-hadoop2'
     compile 'org.apache.logging.log4j:log4j-api:2.3'
     compile 'org.apache.logging.log4j:log4j-core:2.3'
     compile 'org.apache.commons:commons-lang3:3.4'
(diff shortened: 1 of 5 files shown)


jean-philippe-martin:
Travis failed with "fatal: destination path 'broadinstitute/gatk' already exists and is not an empty directory." I'm not sure why; we should probably just give it some time and then retry.

jean-philippe-martin:
Same issue:

```
error: external filter 'git-lfs filter-process' failed
fatal: src/test/resources/large/cnv/create-pon-control-target-coord-only.pcov: smudge filter lfs failed
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'
The command "eval git clone --depth=9999999 https://github.com/broadinstitute/gatk.git broadinstitute/gatk " failed. Retrying, 2 of 3.
fatal: destination path 'broadinstitute/gatk' already exists and is not an empty directory.
```

droazen:
Don't worry @jean-philippe-martin, it's not just your branch. We are seeing `git lfs` errors across-the-board due to quota issues. We've opened a support ticket with github.

jean-philippe-martin:
OK @droazen, thanks, I'll let it be for now.

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
```diff
@@ -51,6 +53,13 @@
             .put("spark.ui.enabled", Boolean.toString(SPARK_DEBUG_ENABLED))
             .put("spark.kryoserializer.buffer.max", "256m")
             .put("spark.hadoop.fs.file.impl.disable.cache", "true") // so NonChecksumLocalFileSystem is not cached between tests
+            // configure the gcs-connector
+            .put("spark.hadoop.fs.gs.impl", "com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem")
```
lbergelson: We're going to have to add these to the spark defaults in gatk-launch as well I think in order to get this to work outside of tests.  Lets do that as a separate PR though since it will require some changes to gatk-launch. #3237

lbergelson:
@jean-philippe-martin This needs a rebase unfortunately.  There's some horrible travis problem as well, but I think it should be fixed with a rebase.

build.gradle
```diff
@@ -126,6 +126,10 @@ dependencies {
     compile 'com.google.cloud:google-cloud-nio:0.19.0-alpha:shaded'
     compile 'com.google.cloud.genomics:google-genomics-dataflow:v1beta2-0.15'
     compile 'com.google.cloud.genomics:gatk-tools-java:1.1'
+    // this comes built-in when running on Google Dataproc, but the library
+    // allows us to reads from GCS also when testing locally (or on non-Dataproc clusters,
+    // should we want to)
+    compile group: 'com.google.cloud.bigdataoss', name: 'gcs-connector', version: '1.6.0-hadoop2'
```
lbergelson: Actually, should we update to 1.6.1 since it's out now?

src/main/java/org/broadinstitute/hellbender/engine/spark/SparkContextFactory.java
jean-philippe-martin: I think we can also try putting them in DEFAULT_PROPERTIES instead of TEST_PROPERTIES, if Gatk always calls the SparkContextFactory to get its context. But agreed that this is best done as part of a separate CL.

build.gradle
jean-philippe-martin: Yes indeed!

jean-philippe-martin:
Rebased.

jean-philippe-martin:
Tests with `TEST_DOCKER = true` failed, I'm not entirely clear why. Here's a bit of the log:

> Building 85% > :test > Resolving dependencies ':jacocoAgent'aven.org/maven2/org/jacoco/org.jacoco.agent/0.7.7.201606060606/org.jacoco.agent-0.7.7.201606060606.jar
> Building 85% > :test > 207 KB/233 KB downloaded> Building 85% > :test > 0 tests completed> Resolving dependencies ':testRuntime':test FAILED

lbergelson:
@jean-philippe-martin Yeah, I'm seeing that.  I suspect it's because we aren't passing the environment variables correctly to the docker tests so it's exploding with an unhelpful error during test initialization.

lbergelson:
Since JP is away on vacation for a while I'm resolving the issues myself in #3398, so I'm closing this pr.

Issue #3163, opened 2017-06-25
roundToInteger in HDF5PCACoveragePoNCreationUtils is set to true causing CreatePanelOfNormals to drop all count columns 


sooheelee:
For discussion with @LeeTL1220 @samuelklee @mbabadi @vruano.

---
### CreatePanelOfNormals gives unexpected error
Running CNV's CreatePanelOfNormals on forty 1000 Genomes Project WES samples using conservatively derived and padded target intervals (189,493 targets). Tool gives the following error.

```
Using GATK jar /usr/shlee/gatk-4.alpha.2-1134-ga9d9d91-SNAPSHOT/gatk-package-4.alpha.2-1134-ga9d9d91-SNAPSHOT-local.jar
Running:
    java -Dsamjdk.use_async_io_read_samtools=false -Dsamjdk.use_async_io_write_samtools=true -Dsamjdk.use_async_io_write_tribble=false -Dsamjdk.compression_level=1 -Dsnappy.disable=true -Xmx16g -jar /usr/shlee/gatk-4.alpha.2-1134-ga9d9d91-SNAPSHOT/gatk-package-4.alpha.2-1134-ga9d9d91-SNAPSHOT-local.jar CreatePanelOfNormals -I shlee-dev/CNV/1kgmix_gccorrect_pon/CreateCNVPon/95bf88a5-c5aa-45eb-9178-efc7d5c75946/call-CombineReadCounts/String_combinedcoverage.tsv -O String.pon --disableSpark
Picked up _JAVA_OPTIONS: -Djava.io.tmpdir=/cromwell_root/tmp
[June 23, 2017 6:54:00 PM UTC] CreatePanelOfNormals  --input shlee-dev/CNV/1kgmix_gccorrect_pon/CreateCNVPon/95bf88a5-c5aa-45eb-9178-efc7d5c75946/call-CombineReadCounts/String_combinedcoverage.tsv --output String.pon --disableSpark true  --minimumTargetFactorPercentileThreshold 25.0 --maximumColumnZerosPercentage 2.0 --maximumTargetZerosPercentage 5.0 --extremeColumnMedianCountPercentileThreshold 2.5 --truncatePercentileThreshold 0.1 --numberOfEigensamples auto --noQC false --dryRun false --sparkMaster local[*] --help false --version false --showHidden false --verbosity INFO --QUIET false --use_jdk_deflater false --use_jdk_inflater false
[...]

PR #3164, opened 2017-06-25
Do not round to integer during CreatePanelOfNormals filtering (addresses #3163).

sooheelee:
Green light from devs to change this. However, perhaps more discussion is warranted. Please see #3163 for details leading to this PR.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollectionUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollectionUtils.java
@@ -474,8 +474,14 @@ public final class ReadCountCollectionUtils {
      * <p>The imputation is done in-place, thus the input matrix is modified as a result of this call.</p>
      *
      * @param readCounts the input and output read-count matrix.
+     * @param percentile must be in [0, 100].
      */
     public static void truncateExtremeCounts(final ReadCountCollection readCounts, final double percentile, final Logger logger) {
+        ParamUtils.inRange(percentile, 0., 100., "Percentile must be in [0, 100].");
+
+        if (percentile == 0.) {
+            return;
+        }
 
         final RealMatrix counts = readCounts.counts();
         final int targetCount = counts.getRowDimension();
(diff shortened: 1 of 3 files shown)


droazen:
This needs a unit test.

sooheelee:
In the meanwhile, I've created a jar with this branch to use in my Docker @vdauwera. Thought you should know.

src/main/java/org/broadinstitute/hellbender/tools/pon/coverage/pca/HDF5PCACoveragePoNCreationUtils.java
```diff
@@ -209,8 +209,8 @@ private static ReadCountCollection cleanNormalizedCounts(final ReadCountCollecti
         // Remove column and targets with too many zeros and targets with extreme median coverage:
         final int maximumColumnZerosCount = calculateMaximumZerosCount(readCounts.targets().size(), maximumPercentageZeroColumns);
         final int maximumTargetZerosCount = calculateMaximumZerosCount(readCounts.columnNames().size(), maximumPercentageZeroTargets);
-        ReadCountCollection cleanedCounts = ReadCountCollectionUtils.removeColumnsWithTooManyZeros(readCounts, maximumColumnZerosCount, true, logger);
-        cleanedCounts = ReadCountCollectionUtils.removeTargetsWithTooManyZeros(cleanedCounts, maximumTargetZerosCount, true, logger);
+        ReadCountCollection cleanedCounts = ReadCountCollectionUtils.removeColumnsWithTooManyZeros(readCounts, maximumColumnZerosCount, false, logger);
+        cleanedCounts = ReadCountCollectionUtils.removeTargetsWithTooManyZeros(cleanedCounts, maximumTargetZerosCount, false, logger);
```
droazen: I'd be satisfied (for now) with a direct unit test for `cleanNormalizedCounts()` that checks that rounding is not performed on the return value. If "not rounding" is a post-condition for this method and part of the method's new contract, then there should be a direct unit test for that post-condition, even if this will later be supplemented by an end-to-end test for the affected tool(s). This could be done by making the method `@VisibleForTesting`.

The method documentation should also be updated to state that rounding is not performed.

samuelklee:
@droazen Please review.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollectionUtils.java
```diff
@@ -474,8 +474,14 @@ public double visit(final int row, final int column, final double value) {
      * <p>The imputation is done in-place, thus the input matrix is modified as a result of this call.</p>
      *
      * @param readCounts the input and output read-count matrix.
+     * @param percentile must be in [0, 100].
```
droazen: Under normal circumstances I'd request unit tests for these new method preconditions -- but these are not normal circumstances!

Issue #2769, opened 2017-05-26
add documentation stylesheet to gatk repo


lbergelson:
Currently the html doc we build looks pretty ugly in the browser.  We don't have the appropriate style sheets available locally.  We should check those into the repo so that it can build a nice looking copy of the documentation.

droazen:
@vdauwera Could you make this happen?

vdauwera:
Sure. But we'll probably need to use some simpler stylesheets than what the website uses.

vdauwera:
addressed in #3165

PR #3165, opened 2017-06-25
Updated templates to fit current website logic

vdauwera:
This PR updates the Freemarker templates so that the resulting pages will work with the current state of the website code. Most of the changes have to do with functionality I put in to enable hosting multiple doc versions and easy switching between them via a dropdown menu. 

I had already done some retrofitting on older tooldocs so the versioned tool docs go back to 3.5, and we can add beta versions of 4 without changing the "latest supported version". 

The only remaining problem is that I couldn't figure out how to output php instead of html. To test the web integration, I just renamed all *.html to *.php with `for f in *.html; do mv -- "$f" "${f%.html}.php"; done` but that doesn't take care of internal links, which are of course broken as a result. @cmnbroad please let me know if I missed something obvious on this front ^^. 

That being said this PR is fully functional as far as I'm concerned.

### DIFF
--- a/src/main/resources/org/broadinstitute/hellbender/utils/helpTemplates/generic.index.template.html
+++ b/src/main/resources/org/broadinstitute/hellbender/utils/helpTemplates/generic.index.template.html
@@ -1,66 +1,96 @@
 <?php
 
-    include '../../../common/include/common.php';
-    include_once '../../config.php';
+    include '../../../../common/include/common.php';
+    include_once '../../../config.php';
     $module = modules::GATK;
-    printHeader($module, "GATK | Tool Documentation Index", "Guide");
+    $name = docSN::toolDocs;
+
+    printHeader($module, $name, "Documentation");
+
+    $dirs = array_diff(scandir("../", SCANDIR_SORT_DESCENDING), array('..', '.'));
+    $curr_version = file_get_contents(engConst::versions_dir.versionFiles::gatk);
 ?>
+
+<link type='text/css' rel='stylesheet' href='gatkDoc.css'>
 
 <div class='row-fluid'>
 
-<?php printGATKDocsNav($module); ?>
+    <aside class="span3">
 
-<div class='span9'>
+        <?php echo produceGuideNav($module, $name) . makeTwitterFeed(); ?>
+
+    </aside>
+
+    <div class='span9'>
 
 <#include "common.html"/>
 
 <#macro emitGroup group>
-    <div class="accordion-group">
-        <div class="accordion-heading">
-            <a class="accordion-toggle" data-toggle="collapse" data-parent="#index" href="#${group.id}">
-                <h4>${group.name}</h4>
-            </a>
-        </div>
-        <div class="accordion-body collapse" id="${group.id}">
-            <div class="accordion-inner">
-                <p class="lead">${group.summary}</p>
-                <table class="table table-striped table-bordered table-condensed">
-                    <tr>
-                        <th>Name</th>
-                        <th>Summary</th>
-                    </tr>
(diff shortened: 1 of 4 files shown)


src/main/resources/org/broadinstitute/hellbender/utils/helpTemplates/generic.index.template.html
```diff
@@ -1,66 +1,89 @@
...
+                    </a>
+                    <ul class="dropdown-menu">
+                        <?php foreach($dirs as $dir) {
+                             print "<li><a tabindex='-1' href='../".$dir."'>".$dir."</a></li>";
+                        } ?>
+                    </ul>
+                </div>
+            </div>
+        </div>
+        <h1><small>Showing docs for version ${version} | The latest version is <?php print $curr_version; ?></small></h1>
```
cmnbroad: Seems like there is something funny going on right around here (no version after "version is.." and the "".$dir.""; } ?>" stuff. Not sure if its php-related:

screen shot 2017-06-25 at 10 32 13 am

vdauwera: Ah yes this is entirely due to the php. It doesn't get evaluated in html files; it only does in php files processed through a server-side interpreter and the browser tries to just ignore it, but the sequence of brackets apparently throws off the html interpreter in the browser. I may be able to switch to a different syntax that will just hide all the bits that we can't show correctly without php. 

Good to know about the extension switching, thanks -- it would be ideal to be able to set a gradle switch in the command line, so we can set the release build to emit the html version for packaging and the php version for publishing to the website.

vdauwera:
Added styling to make the standalone html look nicer + hide elements that are website-specific and don't really make sense to show in the standalone docs.

Addresses https://github.com/broadinstitute/gatk/issues/2769 although here I stuck the styling into common.html instead of making a separate stylesheet -- wdyt?

vdauwera:
Ah shoot, just realized that now the styling is in the php version and overrides the website's... will need to make it a separate css after all; the website should be able to ignore a missing stylesheet

vdauwera:
I separated out the styling into a css file, but that means we need to copy the css into the gatkDoc output dir at build time. @cmnbroad I assume you can add this easily?

cmnbroad:
@vdauwera Yes, its an easy change to copy the css file. PR coming.

The "Tool Docs Index" link at the bottom of these pages is broken though (missing ".html" ?). Having said that, I just built this and the local doc looks great with it!

cmnbroad:
I made a [PR](https://github.com/broadinstitute/gatk/pull/3169) for copying the CSS file that separate from the command line arg one since its dependent on this PR.

vdauwera:
Ah that's because I left out the extension on purpose; our php server is configured to grab whatever file is present with that basename, with a rule of precedence in case there are several with different extensions (which is useful because of reasons). But locally the browser doesn't know to do that. I'm putting in some logic to handle this, thanks for pointing it out.

vdauwera:
Alright I think I've addressed all the issues, this should be good to go

vdauwera:
No waivers! ;) 

I updated the wiki with a note about the PHP functionality (anticipating the merge of your PR slightly) and describing the outputs a tad more. 

@droazen Are we ok to merge this or holding off on everything until tests run again?

vdauwera:
(Oh maybe you meant what I was asking David... I thought you were asking whether the changes required documenting, which would be amusingly circular in a way)

droazen:
Needs to be rebased onto latest master so that tests can pass in travis.

vdauwera:
Doing it now

PR #3167, opened 2017-06-26
Add a doc build command line argument to emit php doc files.

cmnbroad:


vdauwera:
Ah this is great, thanks @cmnbroad. Do I understand correctly that the new usage will be `./gradlew gatkDoc -PphpDoc` ?

cmnbroad:
Right. And just "./gradlew gatkDoc" (with no arg) for html.

vdauwera:
Perfect, thank you 👍

droazen:
Would it be better/simpler to just have both `./gradlew gatkDoc` and `./gradlew phpDoc`?

droazen:
Needs to be rebased onto latest master so that tests can pass in travis.

vdauwera:
Doing it now

vdauwera:
@droazen re: your earlier question, I think it's preferable to use the same base command and add a qualifier -- we may add other output format shortcuts in future, and I find it more intuitive to specify the format as an extra arg rather than a different target altogether.

vdauwera:
@cmnbroad This branch is failing, I don't know why (don't understand the failures). I just rebased it on latest master, which passes everything afaict.

droazen:
@vdauwera Well, the full truth is that the M2 WDL tests are experiencing intermittent failures at the moment due to out-of-disk-space issues on travis. You can safely merge this if only the M2 WDL tests failed.

cmnbroad:
Yeah, this failure looks the same as the others, and unrelated to this change.

vdauwera:
splendid -- then i will merge it.

PR #3170, opened 2017-06-26
kmer PoN training data

davidbenjamin:
@thebkaufman1995 I will assign this to you formally once you are on the dsde methods github team.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
@@ -333,7 +333,9 @@ public final class Mutect2Engine implements AssemblyRegionEvaluator {
     }
 
     private boolean isNonRef(final byte refBase, final PileupElement p) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/engine/PileupWalker.java
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * Internally, the reads are loaded in chunks called read shards, which are then subdivided into active/inactive regions
+ * for processing by the tool implementation. Read shards should typically be much larger than the maximum assembly
+ * region size to achieve good performance, and should have sufficient padding on either end to avoid boundary artifacts
+ * for events near the shard boundaries.
+ *
+ * Read shards exist mainly as a proof-of-concept that we can shard the reads without introducing calling artifacts,
+ * which will be important for the Spark equivalent of this traversal.
+ */
+public abstract class PileupWalker extends GATKTool {
```
magicDGS: What's the difference between this traversal mode an a `LocusWalker`? It looks like it is adding a bit more complexity by using shards, but the only gain there is that it will process some positions twice due to the padding.

davidbenjamin: @magicDGS You are correct.  I was mistaken about how `LocusWalker` works and will rewrite this code.  Thank you!

davidbenjamin:
@thebkaufman1995 I may reopen this PR once this achieves its final form.

PR #3171, opened 2017-06-26
Polishing hapmap M2 validation

davidbenjamin:
@takutosato These are the changes I made for the most recent GP validation.


scripts/mutect2_wdl/unsupported/hapmap_sensitivity.wdl
```diff
@@ -59,23 +67,101 @@ task AnalyzeSensitivity {
...
+
+        cat header body > "${prefix}_combined.txt"
+    }
+
+    runtime {
+            memory: "5 GB"
+        }
+
+    output {
+        File table = "${prefix}_combined.txt"
```
davidbenjamin: I guess so -- it works when I run it.

PR #3176, opened 2017-06-27
Create TestResources for GATK-specific file constants and GenomeLocParser

### LINKED ISSUE #2125: Remove GenomeLocParser from BaseTest
BaseTest instantiates a GenomeLocParser that uses one of our test references.  This causes BaseTest to be unusable by external projects that don't have the file included.  We should either package the reference file in the jar or we should remove the GenomeLocParser from BaseTest and instantiate it in the classes that actually use it.

### LINKED ISSUE #3029: For projects built on GATK framework, automated testing is broken
This applies to projects that import the GATK jar as part of the build process, but are not part of the GATK itself.

All unit and integration tests are (by default) broken, since the BaseTest class requires the mini fasta, even when it should not be required.  This causes breakage, since a project built on the GATK should not be expected to have that file at the exact correct place in the filesystem.

The tests do not even start.

magicDGS:
* Remove repo-specific files from `BaseTest` in favor of `TestResources` constants
* Remove `GenomeLocParser`initialization in `BaseTest` in favor of on-demand initialization in `TestResources` and getters for related objects.

Closes #3029 
Closes #2125

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/test/TestResources.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/test/TestResources.java
@@ -1,0 +1,131 @@
+package org.broadinstitute.hellbender.utils.test;
+
+import htsjdk.samtools.SAMFileHeader;
+import org.broadinstitute.hellbender.utils.GenomeLoc;
+import org.broadinstitute.hellbender.utils.GenomeLocParser;
+import org.broadinstitute.hellbender.utils.fasta.CachingIndexedFastaSequenceFile;
+import org.testng.Assert;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Class containing helper methods and resolved directories for test resources.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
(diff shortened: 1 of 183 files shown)


lbergelson:
@LeeTL1220 This pr would fix the issue you're having.

src/main/java/org/broadinstitute/hellbender/utils/test/TestResources.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Class containing helper methods and resolved directories for test resources.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public final class TestResources {
```
droazen: Instead of extracting a `TestResources` class like this, it would be much less invasive/disruptive to instead create a `GATKBaseTest` subclass of `BaseTest` with the references to GATK-specific test data. That way, you won't have to change every reference to every file -- you'll just have to change `extends BaseTest` to `extends GATKBaseTest`

magicDGS: True, I will close and open a new PR for this because it is already quite old.

magicDGS:
Closing in favor of #3475

PR #3179, opened 2017-06-27
Authenticate with github from Travis

droazen:
This should resolve our git-lfs quota issues, since the quotas
for unauthenticated requests are much stingier than those for
authenticated requests.


.travis.yml
```diff
@@ -3,6 +3,7 @@ sudo: required
 dist: trusty
 git:
   depth: 9999999
+  lfs_skip_smudge: true
```
cmnbroad: Doesn't this turn off git-lfs download ? Or is this a temporary change ?

droazen: This was the part that took me a while to figure out :) We have to disable git-lfs download for the initial git clone, since there is no way in travis to run commands before the initial clone. Then, we authenticate with github in the `before_install` phase, then run `scripts/download_lfs_files.sh` to manually download the lfs files.

cmnbroad: I see.

droazen:
@cmnbroad It looks like this resolves the git-lfs quota issues, but apparently the M2 WDL tests are failing.

cmnbroad:
@droazen are you sure thats not transient ? The first (connection timeout) failures that happened don't appear in previous builds that succeeded.

droazen:
@cmnbroad Yeah, the M2 failures went away in the most recent build, only to be replaced with the `XHMMSegmentCallerBaseIntegrationTest` transient failure :)  

Since @mbabadi has a PR that fixes `XHMMSegmentCallerBaseIntegrationTest`, we can merge this into master, rebase that PR on top of it and merge, and then we should have passing tests again!

PR #3189, opened 2017-06-28
Split over-large SV intervals

TedBrookings:
Split SVIntervals larger than maxIntervalLength (1500), maintaining a
minimum amount of overlap (splitIntervalMinOverlap = 500)

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
@@ -102,6 +102,7 @@ public final class FindBreakpointEvidenceSpark extends GATKSparkTool {
                 new Params(argumentCollection.kSize, argumentCollection.maxDUSTScore,
                         argumentCollection.minEvidenceMapQ,
                         argumentCollection.minEvidenceMatchLength, argumentCollection.maxIntervalCoverage,
+                        argumentCollection.maxIntervalLength, argumentCollection.splitIntervalMinOverlap,
                         argumentCollection.minEvidenceWeight, argumentCollection.minCoherentEvidenceWeight,
                         argumentCollection.minKmersPerInterval,
                         argumentCollection.cleanerMaxIntervals, argumentCollection.cleanerMinKmerCount,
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -681,14 +682,47 @@ public AlignedAssemblyOrExcuse apply( final Tuple2<Integer, List<SVFastqUtils.Fa
...
+            // of overlap. In this case, shorten the split intervals a bit
+            final int maxOverlap = Math.max( params.maxIntervalLength / 3, params.splitIntervalMinOverlap );
+            final int overlap = (numIntervals * params.maxIntervalLength - originalLen) / (numIntervals - 1);
+            final int splitIntervalLen = (numIntervals == 2 && overlap > maxOverlap) ?
+                    (originalLen + maxOverlap) / 2:
+                    params.maxIntervalLength;
+            // Have numIntervals and splitIntervalLen, so make the intervals
+            final double startShift = (originalLen - splitIntervalLen) / (double)(numIntervals - 1);
+            splitIntervals =
+                    java.util.stream.IntStream.range(0, numIntervals)
```
tedsharpe: Import java.util.stream.IntStream.  Always import when you can to make the code more compact.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -681,14 +682,47 @@ public AlignedAssemblyOrExcuse apply( final Tuple2<Integer, List<SVFastqUtils.Fa
...
+                            .map( (idx) -> interval.getStart() + (int)Math.round( startShift * (double)idx ) )
```
tedsharpe: You don't need the parens around idx -- most people would leave them out:
```
.map( idx -> ... )
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -681,14 +682,47 @@ public AlignedAssemblyOrExcuse apply( final Tuple2<Integer, List<SVFastqUtils.Fa
...
+                            .mapToObj( (startIdx) -> new SVInterval( interval.getContig(), startIdx, startIdx + splitIntervalLen ) )
+                            .collect( java.util.stream.Collectors.toList() );
```
tedsharpe: It totally doesn't matter in this case (total number of intervals to check is small and numIntervals is generally quite modest), but since you're just learning idiomatic Java I'll point this out:
Collectors.toList() doesn't know how many elements you're going to end up with, so it can't be very efficient.  There's no guarantee on these details, but what our JVM does is to allocate an ArrayList of default size, which is 10, and then grows it by 50% every time it runs out of room.  That's a lot of reallocation and copying if the list is large.
But you know the size of the resulting list ahead of time.  So do this instead:
```
.collect( SVUtils.arrayListCollector(numIntervals) );
```
That will allocate an ArrayList with an adequate initial capacity, and Bob's your uncle.
Even when you don't know the exact size of the final list, it's good to take a guess so that you don't end up reallocating a tiny list a gazillion times.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSparkUnitTest.java
```diff
@@ -53,6 +68,16 @@ public void getIntervalsTest() {
     }
 
     @Test(groups = "spark")
+    public void splitIntervalTest() {
+        final List<SVInterval> splitIntervals = Arrays.asList( testBigIntervals )
+                .stream()
```
tedsharpe: Arrays has a stream method, so you don't need to make a list first.  Just do  ```Arrays.stream(testBigIntervals)```

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSparkUnitTest.java
```diff
@@ -53,6 +68,16 @@ public void getIntervalsTest() {
...
+                .flatMap( interval -> FindBreakpointEvidenceSpark.splitInterval( interval, params ).stream() )
+                .collect( java.util.stream.Collectors.toList() );
```
tedsharpe: Collectors.toList() is fine here:  you know you have a tiny list.  (But import java.util.stream.Collectors so that the code is more compact.)

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSparkUnitTest.java
```diff
@@ -22,6 +22,21 @@
 public final class FindBreakpointEvidenceSparkUnitTest extends BaseTest {
     private static final SVInterval[] testIntervals =
             { new SVInterval(1, 33140934, 33141485), new SVInterval(1, 33143116, 33143539) };
+    private static final SVInterval[] testBigIntervals = {
```
cwhelan: Rather than put these into two separate arrays, you probably want to use an `@DataProvider` to feed individual inputs to the method along with their expected results. You can find documentation for that in the testng package. It'll make your test a little more compact, and you won't have to do the stream in the test method to collect the results.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindBreakpointEvidenceSpark.java
```diff
@@ -681,14 +682,47 @@ public AlignedAssemblyOrExcuse apply( final Tuple2<Integer, List<SVFastqUtils.Fa
...
+                    / (double)(params.maxIntervalLength - params.splitIntervalMinOverlap ) );
+            // If the interval is *slightly* larger than params.maxIntervalLength, you may end up with an absurd amount
+            // of overlap. In this case, shorten the split intervals a bit
+            final int maxOverlap = Math.max( params.maxIntervalLength / 3, params.splitIntervalMinOverlap );
+            final int overlap = (numIntervals * params.maxIntervalLength - originalLen) / (numIntervals - 1);
+            final int splitIntervalLen = (numIntervals == 2 && overlap > maxOverlap) ?
+                    (originalLen + maxOverlap) / 2:
+                    params.maxIntervalLength;
+            // Have numIntervals and splitIntervalLen, so make the intervals
+            final double startShift = (originalLen - splitIntervalLen) / (double)(numIntervals - 1);
```
tedsharpe: I disagree with Chris on this point.  If startShift was integral you'd accumulate rounding errors when numIntervals was largish.  You'd end up not covering the entire input interval (missing up to numIntervals-1 bases at the end).  I think Ted B. has got this right.
(But you don't really need the cast to double on idx on line 719 -- since startShift is a double the JVM will use floating point math.)

cwhelan: That's fair. I guess I wasn't anticipating numIntervals being that large but maybe it could get there in some cases.

cwhelan:
Before we merge, it might be good to get you to run the full pipeline with this change on a real data set. That'll require a little bit of an intro to GCP and dataproc. Come by and one of us can help step you through that.

TedBrookings:
Closing and deleting. After more testing, change didn't help.

Issue #3188, opened 2017-06-28
Deprecate --allowMissingData and add --errorIfMissingData in VariantsToTable


ronlevine:
Port https://github.com/broadinstitute/gsa-unstable/pull/1606.
Also fixes a NullPointerException when trying to run VariantsToTable when requesting the genotype filter field ("-GF FT") for an unfiltered genotype.

ronlevine:
Closing since #3190 is merged.

PR #3190, opened 2017-06-28
Deprecate --allowMissingData, add --errorIfMissingData and print NA for null object in VariantsToTable

ronlevine:
Implements https://github.com/broadinstitute/gatk/issues/3188


ronlevine:
@vdauwera @lbergelson Please review.

ronlevine:
Ping @vdauwera @lbergelson

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/VariantsToTable.java
```diff
@@ -74,7 +75,7 @@
  *
  * <h3>Caveats</h3>
  * <ul>
- *     <li>Some annotations cannot be applied to all variant sites, so VCFs typically contain records where some annotation values are missing. By default this tool throws an error if you request export of an annotation for which not all records have values. You can override this behavior by setting `--allowMissingData` in the command line. As a result, the tool will emit the special value NA for the missing annotations in those records.</li>
+ *     <li>Some annotations cannot be applied to all variant sites, so VCFs typically contain records where some annotation values are missing. By default this tool the tool will emit the special value NA for the missing annotations if you request export of an annotation for which not all records have values. You can override this behavior by setting --errorIfMissingData in the command line. As a result, the tool will throw an error if a record is missing a value.</li>
```
vdauwera: "this tool the tool" -> fix

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/VariantsToTable.java
```diff
@@ -143,9 +144,19 @@
...
-    @Advanced
-    @Argument(fullName="allowMissingData", shortName="AMD", doc="If provided, we will not require every record to contain every field", optional=true)
+    @Hidden
+    @Deprecated
+    @Argument(fullName="allowMissingData", shortName="AMD", doc="This argument is no longer used, refer to the documentation for --errorIfMissingData", optional=true)
     private boolean allowMissingData = false;
+
+    /**
+     * By default, this tool will write out NA values indicating missing data when it encounters a field without a value in a record.
+     * If this flag is added to the command, the tool will instead exit with an error if missing data is encountered..
```
vdauwera: extra "."

vdauwera:
Doc/microcopy looks good, just a couple of typos to fix. 

I'll leave the functional/code review to the esteemed @lbergelson

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/VariantsToTable.java
ronlevine: Wow, I didn't realize I stutter when I type. 😉 
Changed "this tool the tool" -> "this tool".

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/VariantsToTable.java
ronlevine: Removed extra period.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/VariantsToTable.java
```diff
@@ -143,9 +144,19 @@
      * will cause VariantsToTable to write out NA values for missing fields instead of throwing an error.
      * Note that this flag only applies to standard columns (CHROM, ID, QUAL) and the INFO field and it does not apply to the genotype field.
      */
-    @Advanced
-    @Argument(fullName="allowMissingData", shortName="AMD", doc="If provided, we will not require every record to contain every field", optional=true)
+    @Hidden
+    @Deprecated
+    @Argument(fullName="allowMissingData", shortName="AMD", doc="This argument is no longer used, refer to the documentation for --errorIfMissingData", optional=true)
```
lbergelson: It seems like we should either delete this argument so it errors if someone uses it, or have it check and throw a reasonable exception if someone uses it.  Hiding and deprecating it just means that it will silently do the wrong thing if someone tries to use it.

ronlevine: The reasoning is that if a someone was using it, they still have the functionality. But, new users would not see this option.
@vdauwera Please weigh in.

vdauwera: It's not to keep the functionality, it's that if you use the argument in a command you get a nice message saying "don't use this, use that one instead" rather than just an error about the argument not being found

vdauwera: It's not to keep the functionality, it's that if you use the argument in a command you get a nice message saying "don't use this, use that one instead" rather than just an error about the argument not being found

vdauwera: Although now that I think about it, in GATK4 we could go for a clean slate and just get rid of all the old GATK3 arguments since the upcoming refactoring is going to break all arguments anyway (we're still doing that, right?)

ronlevine: @vdauwera Thanks for correcting me. I forgot that I removed all of the wiring to make this argument functional ...

lbergelson: @vdauwera I don't think you'll get a nice message the way it is right now.  I think it will just silently continue.  You'd have to decide to look at the documentation, while enabling `--showHidden` in order to see the message.  ( I don't think barclay understands automatic warning on deprecated arguments...  )  If we want to show a message for a while as a transitionary step we'd have to add one manually.  I'd vote for just killing it though.

ronlevine: I removed the deprecated `allowMisingData` argument. @lbergelson please confirm you are you are still OK with that.

lbergelson:
@ronlevine 👍

lbergelson:
When you merge, be sure to update the commit comment and message to reflect the fact that it's been replaced instead of deprecated and that the default behavior is the opposite of what it used to be.

ronlevine:
Change commit message to `Replace --allowMissingData with --errorIfMissingData (gives opposite default behavior as previously) and print NA for null object in VariantsToTable`.

PR #3195, opened 2017-06-29
PrimaryAlignmentReadFilter change

magicDGS:
Solves #2165 without breaking previous behavior

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
@@ -180,21 +180,22 @@ public final class ReadFilterLibrary {
             return read.isProperlyPaired();}}
 
     /**
-     * Keep only reads representing primary alignments.
+     * Keep only reads representing primary alignments (those that satisfy both the NotSecondaryAlignment and
+     * NotSupplementaryAlignment filters, or in terms of SAM flag values, must have neither of the 0x100 or
+     * 0x800 flags set).
      *
-     * <p>Warning: currently PrimaryAlignmentReadFilter filters out only secondary alignments (0x100).
+     * Note that this filter represents a stronger criteria for "primary alignment" than the
+     * SAM flag 0x100 (representing ""not primary alignment" in some contexts).
      *
-     * <p>The NotSecondaryAlignmentReadFilter also filters based solely on secondary alignment status.
-     * The PrimaryAlignmentReadFilter will therefore undergo changes in the future to also filter supplementary alignments.
+     * For example, a read that has only the supplementary flag (0x800) set, but not the secondary (0x100)
+     * flag will be filtered out from processing by the PrimaryLineReadFilter, but would NOT be filtered out by
+     * other software that uses the looser notion of "not primary" that only depends on the "secondary" flag being set.
      */
-    //TODO: should this reject supplementary alignments ? If not it should be
-    //removed since its redundant with NotSecondaryAlignmentReadFilter
-    //https://github.com/broadinstitute/gatk/issues/2165
-    @DocumentedFeature(groupName=HelpConstants.DOC_CAT_READFILTERS, groupSummary=HelpConstants.DOC_CAT_READFILTERS_SUMMARY, summary = "Keep only reads representing primary alignments", extraDocs = NotSecondaryAlignmentReadFilter.class)
-    public static class PrimaryAlignmentReadFilter extends ReadFilter {
+    @DocumentedFeature(groupName=HelpConstants.DOC_CAT_READFILTERS, groupSummary=HelpConstants.DOC_CAT_READFILTERS_SUMMARY)
+    public static class PrimaryLineReadFilter extends ReadFilter {
         private static final long serialVersionUID = 1L;
         @Override public boolean test(final GATKRead read) {
-            return ! read.isSecondaryAlignment();}}
+            return ! read.isSecondaryAlignment() && ! read.isSupplementaryAlignment();}}
 
     /**
      * Filter out reads where the read and CIGAR do not match in length.
@@ -275,7 +276,7 @@ public final class ReadFilterLibrary {
     public static final PairedReadFilter PAIRED = new PairedReadFilter();
     public static final ProperlyPairedReadFilter PROPERLY_PAIRED = new ProperlyPairedReadFilter();
     public static final PassesVendorQualityCheckReadFilter PASSES_VENDOR_QUALITY_CHECK = new PassesVendorQualityCheckReadFilter();
-    public static final PrimaryAlignmentReadFilter PRIMARY_ALIGNMENT = new PrimaryAlignmentReadFilter();
+    public static final PrimaryLineReadFilter PRIMARY_ALIGNMENT = new PrimaryLineReadFilter();
     public static final ReadLengthEqualsCigarLengthReadFilter READLENGTH_EQUALS_CIGARLENGTH = new ReadLengthEqualsCigarLengthReadFilter();
     public static final SecondOfPairReadFilter SECOND_OF_PAIR = new SecondOfPairReadFilter();
     public static final SeqIsStoredReadFilter SEQ_IS_STORED = new SeqIsStoredReadFilter();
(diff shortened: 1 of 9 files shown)


magicDGS:
Because you opened the issue, could you have a look to this @cmnbroad?

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -145,14 +145,12 @@ private ReadFilterLibrary(){ /*no instance*/ }
...
-    //TODO: should this reject supplementary alignments ? If not it should be
-    //removed since its redundant with NotSecondaryAlignmentReadFilter
-    //https://github.com/broadinstitute/gatk/issues/2165
+    /** Keep reads representing primary alignments (not 0x100 and not 0x800). */
     @DocumentedFeature(groupName=HelpConstants.DOC_CAT_READFILTERS, groupSummary=HelpConstants.DOC_CAT_READFILTERS_SUMMARY)
     public static class PrimaryAlignmentReadFilter extends ReadFilter {
         private static final long serialVersionUID = 1L;
         @Override public boolean test(final GATKRead read) {
-            return ! read.isSecondaryAlignment();}}
+            return ! read.isSecondaryAlignment() && ! read.isSupplementaryAlignment();}}
```
droazen: Add a unit test for the updated read filter that tests with primary, secondary, and supplementary alignments.

magicDGS: Done.

magicDGS:
Back to you @droazen

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
```diff
@@ -145,14 +145,12 @@ private ReadFilterLibrary(){ /*no instance*/ }
         @Override public boolean test(final GATKRead read) {
             return read.isProperlyPaired();}}
 
-    //TODO: should this reject supplementary alignments ? If not it should be
-    //removed since its redundant with NotSecondaryAlignmentReadFilter
-    //https://github.com/broadinstitute/gatk/issues/2165
+    /** Keep reads representing primary alignments (not 0x100 and not 0x800). */
```
cmnbroad: I think this PR looks ready to go now, but I propose that we add a more detailed javadoc comment here highlighting how this filter's notion of "primary" differs from both GATK3 and htsjdk meanings (but matches my reading of the SAM spec):
    /**
     * Keep only reads representing primary alignments (those that satisfy both the NotSecondaryAlignment and
     * NotSupplementaryAlignment filters, or in terms of htsjdk flag values, must have neither of the 0x100 or
     * 0x800 flags set).
     *
     * Note that this filter represents a stronger criteria for "primary alignment" than either the GATK3
     * NotPrimaryAlignmentFilter, or the htsjdk NOT_PRIMARY_ALIGNMENT flag (0x100).
     *
     * For example, a read that has only the supplementary flag (0x800) set, but not the secondary (0x100)
     * flag - note that htsjdk uses the symbolic name "NOT_PRIMARY_ALIGNMENT" for this value - will be filtered
     * out from processing by the GATK4 PrimaryAlignmentFilter, but would NOT be filtered out by the GATK3
     * NotPrimaryAlignmentFilter, which uses the looser notion of "not primary" that only depends on the 
     * "secondary" flag being set.
     */
I think I got this right; @lbergelson @droazen seeking opinions on whether this is helpful, or adds to the confusion.

lbergelson: This seems like a good addition.   It both explains it and serves as a big flag that says "HEY THIS IS MORE SUBTLE THAN YOU EXPECT"

tedsharpe:
Not that anyone asked me, but I think it might be helpful to rename PrimaryAlignmentReadFilter to PrimaryLineReadFilter.  This would match the language in the SAM spec, and might forewarn users familiar with the old filter that something has changed.

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
magicDGS: I agree partially with the text, but I have some concerns:

* The javadoc is populated into the online help, and mentions to htsjdk may confuse to a GATK user (or downstream toolkit), which does not know the implementation details. I recommend to pass all that information to a javadoc `@implNote`.
* Is it required to add the mention to GATK3 change in behaviour? I think that it is better to add a warning about the notion of "primary alignment" represented by this filter independent of htsjdk and GATK3, except for the API point of view.

As @tedsharpe point out, a change of the name can be useful to point out that something changed. Otherwise, I propose the following javadoc (and other information can be included in the forum/help pages for the transition from GATK3 to GATK4):

```java
/**
* Keep only reads representing primary alignments (those that satisfy both the NotSecondaryAlignment and
* NotSupplementaryAlignment filters, or in terms of SAM flag values, must have neither of the 0x100 or
* 0x800 flags set).
*
* Note that this filter represents a stronger criteria for "primary alignment" than the
* SAM flag 0x100 (representing ""not primary alignment" in some contexts).
*
* For example, a read that has only the supplementary flag (0x800) set, but not the secondary (0x100)
* flag will be filtered out from processing by the GATK4 PrimaryAlignmentFilter, but would NOT be
* filtered out by other software that uses the looser notion of "not primary" that only depends on the
* "secondary" flag being set.
*
* @implNote Note that HTSJDK uses the symbolic name {@link SAMFlag#NOT_PRIMARY_ALIGNMENT}
* for the "secodary" flag (0x100).
* @implNote Note that previous GATK versions implementation differs from the current
* by filtering out reads with 0x800 and not 0x100
*/
```

This also brings to me a question about GATK-specific mentions in the javadoc (which may be populated to the help pages) for downstream projects. It is a different issue, but it may be worthy to add a javadoc-tag to be used in the docgen code to include GATK-specific documentation, to allow downstream projects to skip them if necessary. But that is part of a different issue.

cmnbroad: @magicDGS The reason for mentioning GATK3 was to bring attention to users and developers (especially those who are porting tools) that this filter isn't a direct substitute for the similarly named GATK3 filter. Using @implNote, these notes wouldn't (a far as I can tell)  show up in the gatk doc or the javadoc.

I think @tedsharpe 's idea of calling it "PrimaryLineReadFilter" is reasonable though.

The other idea of scrubbing "GATK" references from the javadoc may be quixotic, though. I'm not quite sure how the tag you suggest would be used, but there are already quite a few appearances of "GATK" in the code now, not to mention all of the classes that have "GATK" as part of the class name, like "GATKRead".

magicDGS: @cmnbroad:

* The implNote should show up in the javadoc, and not in the gatkdoc. That's actually what is worrying me, to show it in the documentation.
* For the rest of documentation, I am fine with havind GATK mentioned in class names and javadoc, but I refer to documented features that may be populated in downstream documentation pages. That can get confusing for users reading other toolkit documentation, which is referring to GATK when it is not the software that they are using.

cmnbroad: @magicDGS I don't think having no references to "GATK" in the documented features is going to be maintainable in general, but in the interest of moving this PR forward, I suggest we:

- Change the name to PrimaryLineReadFilter (including in the javadoc)
- Move the two impl notes into a non-javadoc comment
- Remove the broken javadoc link {@link SAMFlag#NOT_PRIMARY_ALIGNMENT}, which will not be navigable since its a link to separate javadoc
- Fix up the rest of the text (spelling, i.e. "secodary"-> secondary" etc.)

and then we can merge.

magicDGS:
Next commit includes:

* Change the name to PrimaryLineReadFilter
* Remove  impl notes completely, because if they aren't tags, they will be populated to the user documentation. With the name change, I believe that it isn't necessary anymore: with the current text is clear that the concept of primary alignment is more stringent for this filter, the name change clarify that it is a different filter than the previous GATK versions, and the name of HTSJDK flag is also different.

Back to you @cmnbroad

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java
magicDGS: Regarding the documentation issue, I can understand the references to the "GATK" in documented tools, because they are part of the toolkit; but framework utilities, such as read filters, are supposed to be re-used by client-code (including documentation). Or at least, have a custom tag that is processed by the GATK-doclet and that can be pinpoint by custom doclets to remove or reference in a different way...

cmnbroad:
@magicDSG Can you rebase this since it now has conflicts with the doc update PR, and then we can merge when test pass (might as well squash as well.) Thx.

magicDGS:
Rebased/Squashed @cmnbroad - back to you!

cmnbroad:
Thanks @magicDGS.

magicDGS:
Thanks you for accepting it... and all the reviewers for the feedback!

PR #3201, opened 2017-06-30
Add default CharSet for retrieve bytes from String in GATKRead

magicDGS:
Currently, the GATKRead.getAttributeAsByteArray for a `String` attribute use the default CharSet, but it will be nice if the method sticks to a common representation by using a default charset. This PR adds a UTF-8 default encoding for this case in the GATKRead implementations to retrieve the bytes.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
@@ -12,6 +12,7 @@ import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.utils.Utils;
 
 import java.io.Serializable;
+import java.nio.charset.Charset;
 import java.util.*;
 
 /**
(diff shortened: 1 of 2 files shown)


lbergelson:
This seems like a good idea to specify.  👍

lbergelson:
@magicDGS Thanks, I suspect there are other places we haven't been careful about charsets as well.

magicDGS:
Thanks @lbergelson! I will open an issue to track the issue about charsets, and I will try to find usages of `String.getBytes()` without charset to force UTF-8.

magicDGS:
Opps, I've just realized that the SAM specs will use the US-ASCII encoding (https://github.com/samtools/hts-specs/pull/205) unless it is specified UTF-8. Maybe a method to retrieve the `String` with a provided charset should be added to the interface...

Thoughts on that, @lbergelson? I think that it will be a good idea to include US-ASCII or whatever is the default in `ReadConstants` and specify that in the documentation...

lbergelson:
@magicDGS Isn't it the case that an ascii String converted to bytes using UTF-8 will be the same as one converted using ascii?  I think UTF-8 is a superset of US ASCII so it should work fine I think.  Going the other way is more of an issue I think since then you have to decide what to do with characters that aren't representable in ASCII.

lbergelson:
There are definitely going to be a number of exceptions to the ASCII only rule in the bam spec as well, because traditionally you've been allowed to represent non-ascii characters in many attributes.  Those will continue to be encoded with UTF8 I believe.

magicDGS:
Thanks for the answer, @lbergelson... I will keep it as it is!

PR #3206, opened 2017-06-30
Make PathSeq test BAMs pass ValidateSamFile

mwalker174:
@LeeTL1220 Fixed PathSeq test BAMs so that all reads have read groups with an SM tag. This will make them pass the WellFormedReadFilter.

To be thorough, I made sure they also check out with ValidateSameFile. The BAMs are now properly sorted, and also I made a small fix so that unmapped mate flags are set properly in the filter tool.


src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -89,6 +91,7 @@ private static GATKRead clearReadAlignment(final GATKRead read, final SAMFileHea
             newRead.setAttribute(SAMTag.RG.name(), readGroup);
         }
         return newRead;
+
```
LeeTL1220: Extra whitespace.

PR #3213, opened 2017-07-05
Add SortOrder to ReadsDataSource.SamFileHeaderMerger based on inputs

magicDGS:
Currently, if `ReadsDataSource` comes from several files, the `getHeader()` method is always returning a coordinate sorted header. Thus, it is difficult to assume that the input is pre-sorted or not for other sorting orders.

This PR implements a method to obtain the consensus between the files (ignoring the unknown sort order), and if there is no consensus falls back in the previous behavior (coordinate). In addition, it logs a warning with the current assumptions.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
@@ -443,12 +443,32 @@ public final class ReadsDataSource implements GATKDataSource<GATKRead>, AutoClos
      */
     private SamFileHeaderMerger createHeaderMerger() {
         List<SAMFileHeader> headers = new ArrayList<>(readers.size());
+        final Set<SAMFileHeader.SortOrder> sortOrders = new HashSet<>(SAMFileHeader.SortOrder.values().length);
         for ( Map.Entry<SamReader, CloseableIterator<SAMRecord>> readerEntry : readers.entrySet() ) {
-            headers.add(readerEntry.getKey().getFileHeader());
+            final SAMFileHeader h = readerEntry.getKey().getFileHeader();
+            headers.add(h);
+            sortOrders.add(h.getSortOrder());
         }
 
-        // TODO: don't require coordinate ordering
-        SamFileHeaderMerger headerMerger = new SamFileHeaderMerger(SAMFileHeader.SortOrder.coordinate, headers, true);
+        // unknown order is removed and assumes that it is the same as the others
+        final boolean containUnknown = sortOrders.remove(SAMFileHeader.SortOrder.unknown);
+        final boolean detectedOrder;
+        final SAMFileHeader.SortOrder order;
+        if (sortOrders.size() == 1) {
+            order = sortOrders.iterator().next();
+            detectedOrder = true;
+        } else {
+            // TODO: default value should be unknown/unsorted but kept as coordinate to maintain previous behaviour
+            order = SAMFileHeader.SortOrder.coordinate;
+            detectedOrder = false;
+        }
+        if (detectedOrder && containUnknown) {
+            logger.warn("Some inputs have unknown sort orders. Assuming {} sorted reads for them according to other inputs", order);
+        } else if (!detectedOrder) {
+            logger.warn("Inputs have different sort orders. Assuming {} sorted reads for all of them.", order);
+        }
+
+        SamFileHeaderMerger headerMerger = new SamFileHeaderMerger(order, headers, true);
         return headerMerger;
     }
 
--- a/src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.hellbender.engine;
 
+import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.*;
 import htsjdk.samtools.util.CloseableIterator;
 import htsjdk.samtools.util.IOUtil;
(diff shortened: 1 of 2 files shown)


magicDGS:
Do you mind to have a look to this, @droazen?

droazen:
@cmnbroad could you review?

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -443,12 +443,32 @@ public SAMFileHeader getHeader() {
...
-            headers.add(readerEntry.getKey().getFileHeader());
+            final SAMFileHeader h = readerEntry.getKey().getFileHeader();
+            headers.add(h);
+            sortOrders.add(h.getSortOrder());
         }
 
-        // TODO: don't require coordinate ordering
-        SamFileHeaderMerger headerMerger = new SamFileHeaderMerger(SAMFileHeader.SortOrder.coordinate, headers, true);
+        // unknown order is removed and assumes that it is the same as the others
+        final boolean containUnknown = sortOrders.remove(SAMFileHeader.SortOrder.unknown);
```
lbergelson: To clarify since it's very unclear and not really documented `unknown` is the order we get when the sort order doesn't match any of the known spec compliant orders, `unsorted` is what you get when the sort order is unspecified or is specifically marked as unsorted.  

I don't think it's a good idea to ignore unknowns in this case, they should be left in for this check.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -443,12 +443,32 @@ public SAMFileHeader getHeader() {
...
+        final boolean detectedOrder;
+        final SAMFileHeader.SortOrder order;
+        if (sortOrders.size() == 1) {
+            order = sortOrders.iterator().next();
+            detectedOrder = true;
+        } else {
+            // TODO: default value should be unknown/unsorted but kept as coordinate to maintain previous behaviour
+            order = SAMFileHeader.SortOrder.coordinate;
```
lbergelson: I think if we have multiple sort orders we should default to "unsorted" instead of coordinate.  We will end up producing invalid output if we continue defaulting to coordinate.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -443,12 +443,32 @@ public SAMFileHeader getHeader() {
...
+            detectedOrder = false;
+        }
+        if (detectedOrder && containUnknown) {
+            logger.warn("Some inputs have unknown sort orders. Assuming {} sorted reads for them according to other inputs", order);
```
lbergelson: Since I think we should drop this special handling of unknown we don't need this warning.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -443,12 +443,32 @@ public SAMFileHeader getHeader() {
...
+        } else if (!detectedOrder) {
+            logger.warn("Inputs have different sort orders. Assuming {} sorted reads for all of them.", order);
```
lbergelson: We should keep a warning like this if we have multiple input orders even if we change to produce `unsorted` in the case that we have multiples.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
lbergelson: Can you refactor this code so the check for the sort order is in a new method called 
`SortOrder identifySortOrder(List<SamFileHeader> headers){...}` instead of being inline with the rest of the code here.  It will make it easier to read.

lbergelson: Would you mind adding a unit test when you do the refactoring as well?

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
```diff
@@ -443,12 +443,32 @@ public SAMFileHeader getHeader() {
      */
     private SamFileHeaderMerger createHeaderMerger() {
         List<SAMFileHeader> headers = new ArrayList<>(readers.size());
+        final Set<SAMFileHeader.SortOrder> sortOrders = new HashSet<>(SAMFileHeader.SortOrder.values().length);
```
lbergelson: Instead of keeping a separate list of sort orders here, lets add a new method that encapsulates the new logic your adding.  Then just pass the existing list of headers into the new method.
```
SortOrder identifySortOrder(List<SamFileHeader> headers){...} 
```
I think it will make things more readable, and make it easy to test the new functionality.  (If you don't mind could you add a unit test for the new method as well? )

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
magicDGS: I think that a refactoring for sort orders in htsjdk will be required eventually, if https://github.com/samtools/hts-specs/issues/204 is added to the specs. Anyway, I will treat for this fix the unknown as any other sort order.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
magicDGS: I've just kept the same behavior just in case there is some dependent code in GATK. I will modify this in the next commit, because I agree that `unsorted` is the proper default.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
magicDGS: Removed in the next commit.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
magicDGS: Sure, will do!

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
lbergelson: I appreciate that, but in this case it's definitely a buggy case we didn't think about and if we're relying on it we should fix our code.

src/main/java/org/broadinstitute/hellbender/engine/ReadsDataSource.java
magicDGS: I'll do.

magicDGS:
I added the tests for the new method and default to unsorted. Back to you @lbergelson and thanks for reviewing!

lbergelson:
@magicDGS Thank you for finding and fixing this bug.

PR #3220, opened 2017-07-06
fix crash when GatherVcfs gathers files with no variants

### LINKED ISSUE #3218: --gatherType BLOCK in GatherVcfs fails with IllegalStateException on empty VCFs
If one of the block compressed VCFs in the list is empty (i.e. it does have proper header lines but there are no variant records, which is perfectly valid) then the tool fails with an IllegalStateException:

java.lang.IllegalStateException: Could not read available bytes from BlockCompressedInputStream.
at org.broadinstitute.hellbender.utils.Utils.validate(Utils.java:697)
at org.broadinstitute.hellbender.tools.GatherVcfs.gatherWithBlockCopying(GatherVcfs.java:354)

lbergelson:
GatherVcf was crashing in BLOCK mode if any file after the first had a valid vcf header but no variants
fixes #3218


src/main/java/org/broadinstitute/hellbender/tools/GatherVcfs.java
```diff
@@ -345,7 +345,7 @@ private static void gatherWithBlockCopying(final List<Path> vcfs, final File out
                         boolean lastByteNewline = true;
 
                         int firstNonHeaderByteIndex = -1;
-                        while (!in.eof()) {
+                        while (!in.eof() && blockIn.available() > 0) {
```
droazen: This history of this line is a bit unsettling. Prior to the blocked NIO PR we had `while (in.available() > 0) {`. Then with that PR it changed to `while (!in.eof()) {`, and now we have `while (!in.eof() && blockIn.available() > 0) {`. 

Are we **sure** this latest version is 100% correct? Could `in` ever reach `eof` while `blockIn` still has available bytes?

PR #3223, opened 2017-07-07
Fix GATKSAMRecordToGATKReadAdapter.getAttributeAsString for byte[]

magicDGS:
For `byte[]` attributes, the `GATKSAMRecordToGATKReadAdapter.getAttributesAsString()` uses the default `Object.toString()` (className@hashCode). This is something unexpected for my point of view, because the following code may fail:

```java
public void testGATKReadGetAttributeAsString(final GATKRead read) {
    read.setAttribute("BC", new byte[]{'A', 'C', 'T', 'G'});
    // this will fail with the current implementation
    Assert.assertEquals(read.getAttributesAsString("BC").getBytes(Charset.forName("UTF-8")),
                        read.getAttributeAsByteArray("BC"));
}
```

This PR fixes the issue by identifying `byte[]` attributes and converting them to Strings by using the default GATKRead charset (UTF-8).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/read/SAMRecordToGATKReadAdapter.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/read/SAMRecordToGATKReadAdapter.java
@@ -507,7 +507,13 @@ public class SAMRecordToGATKReadAdapter implements GATKRead, Serializable {
     public String getAttributeAsString( final String attributeName ) {
         ReadUtils.assertAttributeNameIsLegal(attributeName);
         final Object attributeValue = samRecord.getAttribute(attributeName);
-
+        if ( attributeValue instanceof byte[]) {
+            // in case that the attribute is a byte[] array, the toString method will format it as name@hashCode
+            // for a good representation of the byte[] as String, it encodes the bytes with the default charset (UTF-8)
+            final byte[] val = (byte[]) attributeValue;
+            return (val.length == 0) ? "" : new String(val, DEFAULT_CHARSET);
+        }
+        // otherwise, just use the toString() method unless it is null
         return attributeValue != null ? attributeValue.toString() : null;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/utils/read/SAMRecordToGATKReadAdapter.java
```diff
@@ -507,7 +507,13 @@ else if ( attributeValue instanceof Integer ) {
...
         ReadUtils.assertAttributeNameIsLegal(attributeName);
         final Object attributeValue = samRecord.getAttribute(attributeName);
-
+        if ( attributeValue instanceof byte[]) {
+            // in case that the attribute is a byte[] array, the toString method will format it as name@hashCode
+            // for a good representation of the byte[] as String, it encodes the bytes with the default charset (UTF-8)
+            final byte[] val = (byte[]) attributeValue;
+            return (val.length == 0) ? "" : new String(val, DEFAULT_CHARSET);
+        }
+        // otherwise, just use the toString() method unless it is null
```
droazen: Can you add a test case for this @magicDGS ?

lbergelson:
@magicDGS Good catch, that's a bad bug.  Would you mind adding a unit test for the new code?

src/main/java/org/broadinstitute/hellbender/utils/read/SAMRecordToGATKReadAdapter.java
magicDGS: Added a previously failing test.

magicDGS:
Back to you @droazen and/or @lbergelson

lbergelson:
@magicDGS Thank you.  Looks good.  I will merge this once tests pass.

magicDGS:
Thank you for reviewing!

PR #3226, opened 2017-07-07
using travis' git lfs instead of installing it

lbergelson:
Travis-CI now has git-lfs baked into the images so we can delete our installation script and just use their instance.  
deleted download_lfs_files.sh


ronlevine:
:clap: 🎆

PR #3228, opened 2017-07-07
Fix CountFalsePositives

takutosato:
With a large interval file e.g. WGS intervals it would return garbage as target territory and 0 as false positive rates. Now it's fixed

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/tools/walkers/validation/CountFalsePositivesIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/walkers/validation/CountFalsePositivesIntegrationTest.java
@@ -15,9 +15,16 @@ import static org.broadinstitute.hellbender.tools.walkers.validation.FalsePositi
  * Created by tsato on 12/28/16.
  */
 public class CountFalsePositivesIntegrationTest extends CommandLineProgramTest {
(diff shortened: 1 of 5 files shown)


takutosato:
@davidbenjamin will you review this?

src/test/java/org/broadinstitute/hellbender/tools/walkers/validation/CountFalsePositivesIntegrationTest.java
```diff
@@ -37,6 +37,29 @@ public void testSimple() throws Exception {
...
+                "-L", wgsIntervalFile,
+                "-O", output.toString()
+        };
+        runCommandLine(args);
+
+        FalsePositiveRecordReader reader = new FalsePositiveRecordReader(output);
+        FalsePositiveRecord record = reader.readRecord();
+
+        Assert.assertEquals(record.getTargetTerritory(), 2832598739L);
+        Assert.assertEquals(record.getSnpFalsePositiveRate(), 173.0/2832598739L * 1e6);
```
davidbenjamin: Extract constants: `final int expectedNumFalsePositiveSNPs = 173`, `final int wgsIntervalsTotalSize = 2832598739` etc.

davidbenjamin:
@takutosato Are tests failing because you still need to commit the wgs intervals file?

PR #3217, opened 2017-07-05
travis PR builds now use the correct commit in docker

lbergelson:
Previously we had an issue where our travis builds would use the wrong
commit for the docker builds in the travis pull-request builds but not for the push
builds.

This caused the tests from master to run and usually pass.  However,
since we are mounting the test data from the correct commit into the
docker, this would result in confusing mismatches where old tests would
try to run on new test data.

Fixing the problem by using the $TRAVIS_COMMIT environment variable
instead of the TRAVIS_BRANCH.

fixes #3216

PR #3229, opened 2017-07-07
always use upstream read when looking at template lengths

tedsharpe:
fixes a flaw where i was looking at the 1st read, and the event could've been upstream or downstream.  now i always choose the upstream read so that the event will be downstream.


SHuang-Broad:
No objection.

SHuang-Broad:
@tedsharpe , for the failing PR, I tried restarting it and it failed again, so I would assume it would be fixed after #3217 is in.

PR #3238, opened 2017-07-10
Add features to PositionalDownsampler to handle hotspots

davidbenjamin:
Here's what this PR does:

* Adds an argument to skip a locus rather than downsample if depth is too high.  This handles horrible regions so rife with mapping errors that we wouldn't believe any calls from them.  Basically, it blacklists on the fly intervals that should have been blacklisted ahead of time.
* Adds a stride argument for positional downsampling so that we downsample within a range of alignment start positions.  This regularizes statistical fluctuations in coverage and, more importantly, makes positional downsampling much more useful for data where many reads share the same start position while most start positions have no reads.  I believe @fleharty deals with this kind of stuff.
* ~~Deletes a few methods in the `ReadsDownsampler` API that were only used in their own unit tests.~~
* Adds an option to bias downsampling in favor of reads with high mapping quality.
* Beefs up the unit tests for `PositionalDownsampler`.

The new functionality is a superset of the old, and the old stuff is unchanged.  HaplotypeCaller is not affected, unless of course someone wants to run it with the new options.  The new options allow Mutect2 to run much faster.


davidbenjamin:
@droazen Let me know if you would like me to add the new features more cautiously eg by making a separate `PositionalDownsampler` class with the new features or by verifying exact matching of HaplotypeCaller output on bams other than those in the repo.

davidbenjamin:
@takutosato Here it is.

droazen:
@davidbenjamin We should chat about this in person -- have some questions for you.

davidbenjamin:
@droazen yes, let's discuss.  There's no need to remove those methods as far as Mutect is concerned.  As long as we implement the new functionality in *some* downsampler class, we have what we need.

droazen:
@davidbenjamin This branch seems to have a lot of unrelated refactoring bundled in with the substantive fix for `Mutect2`, and some of this refactoring seems problematic to me (eg., deletion of some downsampler API methods that will be needed when `ReadWalker` downsampling is ported).  

Since mixing refactoring into the same PR as behavioral changes is asking for trouble as a general rule, could I ask you to strip out everything except the substantive change to `PositionalDownsampler` from this PR? Happy to do a quick review once this is done.

davidbenjamin:
@droazen @takutosato Now it only has the changes needed for M2.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -63,6 +63,15 @@
     @Argument(fullName = "maxReadsPerAlignmentStart", shortName = "maxReadsPerAlignmentStart", doc = "Maximum number of reads to retain per alignment start position. Reads above this threshold will be downsampled. Set to 0 to disable.", optional = true)
     protected int maxReadsPerAlignmentStart = defaultMaxReadsPerAlignmentStart();
 
+    @Argument(fullName = "alignmentStartStride", shortName = "alignmentStartStride", doc = "Length in bases to include in a single unit of positional downsampling.", optional = true)
```
magicDGS: The number of arguments for `AssemblyRegionWalker` is increasing exponentially, and this is a pity for other usages outside `HaplotypeCaller` and `Mutect2` if they aren't need it. Maybe it's time to add an abstract `AssemblyRegionArgumentCollection` ( #argumentCollection to allow hiding some arguments (related with #2371 too, because a `ShardingArgumentCollection` will be useful too).

Like that, instead of a bunch of abstract default getters, only an `AssemblyRegionArgumentCollection` getter will be needed with a default implementation returned by `HaplotypeCaller` and `Mutect2`. I can open a PR with this idea and/or implement it in #2371 (either before or after this PR). What do you think, @droazen?

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReservoirDownsampler.java
```diff
@@ -78,6 +84,22 @@ public ReservoirDownsampler(final int targetSampleSize, final boolean expectFewO
     /**
      * Construct a ReservoirDownsampler
      *
+     * @param targetSampleSize Size of the reservoir used by this downsampler.
```
davidbenjamin: I'm being cautious since it's not my class.  I'll do whatever David R prefers.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/ReservoirDownsampler.java
```diff
@@ -118,8 +151,9 @@ public boolean hasFinalizedItems() {
         if (hasFinalizedItems()) {
             // pass reservoir by reference rather than make a copy, for speed
             final List<GATKRead> downsampledItems = reservoir;
+            final int depth = totalReadsSeen;
             clearItems();
-            return downsampledItems;
+            return depth < depthToIgnoreLocus ? downsampledItems : Collections.emptyList();
```
davidbenjamin: You're right that we could speed the downsampling with this kind of shortcut, but this is an edge case and almost all the benefit comes from skipping the reads downstream of the downsampler.  The cost of downsampling itself is really small.  Now, if somehow we could downsample in such a way that we could actually make reading the bam file cheaper we would want to short-circuit as much as possible to avoid unnecessary I/O.  Since the reads are already parsed into memory before downsampling there's not a lot to be gained.

Of course, there's not a lot to be lost either, but for me the tie-breaker is keeping changes to code shared with HaplotypeCaller as minimal as possible.

src/test/java/org/broadinstitute/hellbender/utils/downsampling/ReservoirDownsamplerUnitTest.java
```diff
@@ -124,6 +127,41 @@ public void testSignalNoMoreReadsBefore() throws Exception {
...
+                .mapToObj(n -> {
+                    final GATKRead read = ArtificialReadUtils.createArtificialRead("100M");
+                    read.setMappingQuality(n);
+                    return read;
+                }).collect(Collectors.toList());
+        Collections.shuffle(reads, Utils.getRandomGenerator());
+        rd.submit(reads);
+        rd.signalEndOfInput();
+        final List<GATKRead> downsampledReads = rd.consumeFinalizedItems();
+        Assert.assertEquals(downsampledReads.size(), 10);
```
davidbenjamin: done

src/test/java/org/broadinstitute/hellbender/utils/downsampling/ReservoirDownsamplerUnitTest.java
```diff
@@ -124,6 +127,41 @@ public void testSignalNoMoreReadsBefore() throws Exception {
...
+        final double averageMappingQuality = downsampledReads.stream()
+                .mapToInt(GATKRead::getMappingQuality).average().getAsDouble();
+        Assert.assertTrue(averageMappingQuality > 70);
+    }
+
+    @Test
+    public void testDepthToIgnoreLocus() {
+        Utils.resetRandomGenerator();
+        final ReservoirDownsampler rd100 = new ReservoirDownsampler(10, false, false, 100);
```
davidbenjamin: done and did something in the same spirit.  Instead I extracted `numReads = 100` and set `depthToIgnore` to `numReads` and `numReads + 1`.

src/main/java/org/broadinstitute/hellbender/utils/downsampling/PositionalDownsampler.java
```diff
@@ -22,59 +22,64 @@
...
 
-    private List<GATKRead> finalizedReads;
+    private final int alignmentStartStride;
 
     /**
      * Construct a PositionalDownsampler
      *
      * @param targetCoverage Maximum number of reads that may share any given alignment start position. Must be > 0
-     * @param header SAMFileHeader to use to determine contig ordering. Non-null.
+     * @param downsampleByMappingQuality    If true, bias downsampling toward reads with higher mapping quality.
```
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
```diff
@@ -179,6 +181,15 @@
     protected int defaultMaxReadsPerAlignmentStart() { return DEFAULT_MAX_READS_PER_ALIGNMENT; }
 
     @Override
+    protected int defaultAlignmentStartStride() { return DEFAULT_ALIGNMENT_START_STRIDE; }
+
+    @Override
+    protected boolean defaultDownsampleByMappingQuality() { return false; }
```
davidbenjamin: Probably so; done.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
davidbenjamin: I like that suggestion.

davidbenjamin:
@droazen Do you still want to give me a memory patch to test?

droazen:
@davidbenjamin Yes, I am testing that patch out now -- should be ready for you to try out tomorrow!

droazen:
@davidbenjamin As discussed in person, it's my hope that the `AssemblyRegionWalker` changes in https://github.com/broadinstitute/gatk/commit/1ef09b3ca265209e0777c77a8519da74480908ce (which have now been merged into master!) will address `Mutect2` memory usage, and make these somewhat confusing new downsampling arguments unnecessary. That patch reduces the number of reads stored in memory at once by the engine by roughly an order of magnitude without doing any extra downsampling at all.

I suggest that we do an evaluation to test whether this really resolves the issues you encountered. `Mutect2` is already hooked up to the new, lower-memory traversal code in the latest gatk/master, so all you have to do is re-run your benchmarking test. I'd suggest that you:

1. Run with default settings in the latest master, and see if that alone does the trick!

2. If not, try turning up the existing downsampling a bit. Eg., run with `--maxReadsPerAlignmentStart 10` instead of the default of 50.

3. If that still doesn't resolve the problem, we can revisit this PR and consider a simplified version of the downsampling args here for merge.

davidbenjamin:
@lbergelson As David R asked I compared the run-time of this branch to the master branch, with his most recent changes to memory use in shards.  His changes helped, but with equivalent amounts of downsampling this branch remains 30% faster in total CPU time and about 40% faster in wall clock time.  He instructed me to pass the review on to you if a difference like this remained.

droazen:
@davidbenjamin Let me do a bit of profiling/optimization on the new code path to see if I can narrow the gap. The patch was intended to address memory use rather than runtime, and could use a profiling pass.

droazen:
@davidbenjamin Can you tell me whether you were running with or without an interval list in your tests?

davidbenjamin:
I was using the wgs whitelisted intervals used for HaplotypeCaller (/seq/references/Homo_sapiens_assembly19/v1/variant_calling/wgs_calling_regions.v1.interval_list) for wgs and ICE exome intervals (/seq/references/HybSelOligos/whole_exome_illumina_coding_v1/whole_exome_illumina_coding_v1.Homo_sapiens_assembly19.targets.interval_list) for wes.

droazen:
Aha, good -- that's consistent with @skwalker's results showing a ~30%-40% performance regression vs. GATK3 when using a large interval list + the latest GATK4 HC. I think this is something we can resolve through profiling -- I'll update you in a few days with my progress.

In the mean time, it would be valuable to me to know whether you see the same performance difference in Mutect2 when running *without* an interval list (latest master vs. this downsampling branch). Perhaps you could create a large-but-not-too-large bam snippet to test that out?

droazen:
Actually, if you have the time, an even more valuable test would be to repeat your comparison with latest master vs. a rebased copy of this branch onto latest master. That would tell us whether the performance difference you saw is due to the downsampling, or due to the differences in the traversal code.

davidbenjamin:
@droazen Everything was rebased to master as of 9/12, but are you saying there are changes since then that make a rebase worthwhile?

davidbenjamin:
@droazen I rebased against master on 11/1, and ran on several tumor-normal pairs with and without interval lists.  The total CPU time remains about 30-40% better in the rebased downsampling branch than in the latest master branch.  Also, wall clock time turns out to be a stronger argument in favor of these additions: scattering wgs bams 50 ways, the wall clock time tends to be 3-5x greater in master.

droazen:
@davidbenjamin I think one essential test you haven't done is to run master with more aggressive downsampling settings using the existing `--maxReadsPerAlignmentStart` argument, and compare that against your branch. This would help us determine whether or not the existing downsampling functionality really is inadequate to control peak memory use and runtime.

The current default in `Mutect2` for `--maxReadsPerAlignmentStart` is 50, which is almost certainly much higher than it needs to be. That's 50 * 300 = 15,000 reads per assembly region, with on the order of ~2 regions in memory at any time = 30,000 reads.

I propose that we try a run with the current vanilla master and `--maxReadsPerAlignmentStart 3`, and compare that to an identical run on your branch with the default `--maxReadsPerAlignmentStart 50` + your new downsampling options.

davidbenjamin:
@droazen I agree that running master with default settings is not a fair comparison.  Therefore, I set `-maxReadsPerAlignmentStart` to be equal and much lower than the default in both branches.  I have tried a few values, but none lower than 10.  I expect that using master downsampling to 3 reads per alignment start will handle most of the runtime issue at some small cost to sensitivity and precision, but let me run it and see what happens.

davidbenjamin:
@droazen Setting master to 3 reads per alignment start and this branch to 30 reads per alignment start with a stride of 10 (i.e. equal downsampling on average) and scattering four wgs bams 50 ways each, CPU time is almost identical but wall clock time is anywhere from 20% to 100% slower in master.  Also, sensitivity in master goes down by 0.5% but is unchanged in this branch.

I think this makes a good case for the alignment start stride and biasing downsampling to higher mapping quality, but those are not urgent.  As we discussed in person, I *don't* like my maxReadsToIgnore region argument, which provides most of the benefit, and I want to explore a read start filter as an alternative.  I think that will take less than a week to prototype.

davidbenjamin:
Doing this in a different way without affecting code shared with HaplotypeCaller.

PR #3241, opened 2017-07-10
Set global NIO options.

jean-philippe-martin:
This uses a snapshot version of gcloud, to get the setDefaultCloudStorageConfiguration feature.

Do not submit, because of the reliance of a self-compiled gcloud.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -148,7 +148,6 @@ public class Main {
      * Note: this is the only method that is allowed to call System.exit (because gatk tools may be run from test harness etc)
      */
     protected final void mainEntry(final String[] args) {
-        BucketUtils.setGlobalNIODefaultOptions();
         final CommandLineProgram program = extractCommandLineProgram(args, getPackageList(), getClassList(), getCommandLineName());
         try {
             final Object result = runCommandLineProgram(program, args);
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -15,6 +15,7 @@ import org.broadinstitute.hellbender.utils.Utils;
 import java.io.PrintStream;
 import java.io.Serializable;
 import java.util.*;
+import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
 
 /**
  * This is the main class of Hellbender and is the way of executing individual command line programs.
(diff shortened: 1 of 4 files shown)


jean-philippe-martin:
@droazen @lbergelson here's the code you were asking for. I'm not certain as to where's the best place to set global defaults; for now I put them in Main.

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -147,6 +148,7 @@ public Object instanceMain(final String[] args) {
      * Note: this is the only method that is allowed to call System.exit (because gatk tools may be run from test harness etc)
      */
     protected final void mainEntry(final String[] args) {
+        BucketUtils.setGlobalNIODefaultOptions();
```
droazen: Can you move to `CommandLineProgram.instanceMainPostParseArgs()`, just below the call to `BlockGunzipper.setDefaultInflaterFactory()`? Should also add logger output to `CommandLineProgram.printSettings()` to output the current default retries/reopens for GCS.

droazen:
@jean-philippe-martin I get the following error when trying to use the shaded google-cloud-java snapshot jar:

```
/Users/droazen/.m2/repository/com/google/cloud/google-cloud-nio/0.20.2-alpha-SNAPSHOT/google-cloud-nio-0.20.2-alpha-SNAPSHOT-shaded.jar(com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.class): warning: Cannot find annotation method 'value()' in type 'AutoService': class file for com.google.auto.service.AutoService not found
```

Checking the jar, `AutoService` is indeed not present, but is an annotation on `CloudStorageFileSystemProvider`. Any thoughts on what could be causing this?

jean-philippe-martin:
I'm looking into this right now.

droazen:
Thanks @jean-philippe-martin. If there's some change I can make in my clone of `google-cloud-java` to get the shaded snapshot working, let me know!

Interestingly, the `AutoService` annotation is marked as `@Retention(SOURCE)`...

jean-philippe-martin:
Marking `AuthService` as `optional` in gcloud-nio's pom fixes it:

```
<dependency>
    <groupId>com.google.auto.service</groupId>
    <artifactId>auto-service</artifactId>
    <version>1.0-rc3</version>
    <optional>true</optional>
    <scope>provided</scope> <!-- to leave out of the all-deps jar -->
</dependency>
```

$ ./gradlew sparkJar
BUILD SUCCESSFUL

$ ./gatk-launch ExampleNioCountReads (...)
(starts without error)

droazen:
Thanks @jean-philippe-martin -- this worked, though it took me a while to realize that the update from `1.0-rc2` to `1.0-rc3` was also essential (still failed for me with `1.0-rc2`)

jean-philippe-martin:
ah sorry @droazen , I should have mentioned it. I didn't test with 1.0-rc2.

droazen:
Closing -- this was done in https://github.com/broadinstitute/gatk/pull/3295

PR #3243, opened 2017-07-11
No more prepareReadLikelihoodsForAnnotation in Mutect

davidbenjamin:
This simplifies the code and didn't affect specificity in our validations.  @takutosato can you review this?


takutosato:
yup absolutely

PR #3250, opened 2017-07-11
Adding the ability to generate intervals on filtered variants as well as unfiltered variants.

LeeTL1220:
Default is previous behavior.

Added testing.

Note that this change may need to be made in picard as well.


LeeTL1220:
@TedBrookings Thanks!

src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/VcfToIntervalList.java
```diff
@@ -23,6 +23,8 @@
         programGroup = VariantProgramGroup.class
 )
 public final class VcfToIntervalList extends PicardCommandLineProgram {
+    public static final String INCLUDE_FILTERED_LONG_NAME = "includeFiltered";
```
TedBrookings: <Hair-Splitting>Might be clearer to move declaration of INCLUDE_FILTERED names to just before they are first used.</Hair-Splitting>

TedBrookings: Splitting hairs: might be clearer to move declaration of INCLUDE_FILTERED names to just before they are first used.

LeeTL1220: This is the convention used elsewhere in GATK.

No action.

cmnbroad:
@LeeTL1220 Does GATK have a dependency on this change ? If so this will definitely need to be made in Picard as well since these GATK copies of the Picard tool are soon to be obsoleted.

LeeTL1220:
@cmnbroad I put in a test in GATK that should fail if the dependency is moved to Picard.  So, you are 100% correct that these changes will need to go into Picard.

PR #3252, opened 2017-07-11
Return NA if VariantsToTable is printing a null object, replace --allowMissingData with --errorIfMissingData

ronlevine:
Related to https://github.com/broadgsa/gatk/pull/22.
Port of https://github.com/broadinstitute/gsa-unstable/pull/1606

Also replaced --allowMissingData with --errorIfMissingData since we'd rather not fail  by default for missing data.
The root cause of the problem is due the implementation of `Genotype.hasAnyAttribute`. It always returns true for FT, even if it's missing.


bbimber:
@ronlevine just pointed out this is redundant w/ another issue (see above).  for that issue you guys requested I write a unit test, which I did this morning.  Feel free to apply that or not.  It's attached to that thread as a patch (sorry, dont have a good local enlistment right now).  It's nothing special, but tests are rarely a bad thing.

ronlevine:
Closing. This is a duplicate of https://github.com/broadinstitute/gatk/pull/3190.

PR #3262, opened 2017-07-12
[SV] Fix cross-contig exclusion issues

cwhelan:
This PR fixes two problems I noticed relative to how we are treating cross-contig evidence that comes from the alt contigs:

- The cross-contig exclusion rule in `ReadClassifier` was incorrect. Due to the structure of the `if`-`else if` logic in `checkDiscordantPair()`, `SameStrandPair` or `OutiesPair` evidence was still being created for cross-contig pairs based on the strand configuration of the two reads (even though they were aligned to different contigs).
- The `runWholePipeline` script we've been using was not actually setting the cross-contig kill list parameter, meaning no cross-contig evidence was being filtered out.

The change makes the number of intervals for CMHMIX WGS1 drop from 31962 to 27645.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadClassifier.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadClassifier.java
@@ -125,27 +125,29 @@ public class ReadClassifier implements Function<GATKRead, Iterator<BreakpointEvi
     private void checkDiscordantPair( final GATKRead read, final List<BreakpointEvidence> evidenceList ) {
         if ( read.mateIsUnmapped() ) {
             evidenceList.add(new BreakpointEvidence.MateUnmapped(read, readMetadata));
-        } else if ( isInterContig(read.getContig(),read.getMateContig()) ) {
-            evidenceList.add(new BreakpointEvidence.InterContigPair(read, readMetadata));
-        } else if ( read.isReverseStrand() == read.mateIsReverseStrand() ) {
-            evidenceList.add(new BreakpointEvidence.SameStrandPair(read, readMetadata));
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/ReadClassifier.java
```diff
@@ -125,25 +125,33 @@ private void checkBigIndel( final List<CigarElement> cigarElements,
...
         } else {
-            final float zIshScore = readMetadata.getZishScore(read.getReadGroup(), Math.abs(read.getFragmentLength()));
-            if ( zIshScore > MAX_ZISH_SCORE && zIshScore < MIN_CRAZY_ZISH_SCORE ) {
-                evidenceList.add(new BreakpointEvidence.WeirdTemplateSize(read, readMetadata));
+
+            final int contigID1 = readMetadata.getContigID(read.getContig());
+            final int contigID2 = readMetadata.getContigID(read.getMateContig());
+
+            if (contigID1 != contigID2) {
+                if (isValidIntercontigEvidence(contigID1, contigID2)) {
```
tedsharpe: I'd suggest killing the tiny function, isValidIntercontigEvidence, which now contains a superfluous condition that's already been tested (equality of contigs) and inline this code:
```
    if ( !readMetadata.ignoreCrossContigID(contigID1) && !readMetadata.ignoreCrossContigID(contigID2) )
```
I think this would read more clearly, but YMMV.  I don't insist.  And by the way, thanks for finding and fixing my faulty logic.

cwhelan: done

PR #3268, opened 2017-07-13
oops

tedsharpe:
lines 125-127 fix a bug:  need to clear out 1st half window when we switch contigs
changes on lines 62-67 are cosmetic

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindSmallIndelRegions.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindSmallIndelRegions.java
@@ -1,6 +1,8 @@
 package org.broadinstitute.hellbender.tools.spark.sv;
 
+import com.netflix.servo.util.VisibleForTesting;
 import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.SAMSequenceRecord;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.apache.spark.broadcast.Broadcast;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindSmallIndelRegions.java
```diff
@@ -58,8 +59,12 @@ protected void runTool( final JavaSparkContext ctx ) {
                 })
                 .collect();
 
+        System.out.println();
+        final List<SAMSequenceRecord> seqs = header.getSequenceDictionary().getSequences();
         for ( final BreakpointEvidence evidence : evidenceList ) {
-            System.out.println(evidence);
+            final SVInterval interval = evidence.getLocation();
+            final String contigName = seqs.get(interval.getContig()).getSequenceName();
+            System.out.println(contigName+"\t"+(interval.getStart()+1)+"\t"+interval.getEnd());
```
pshapiro4broad: Need spaces around operators here, e.g. 
```
System.out.println(contigName + "\t" + (interval.getStart() + 1) + "\t" + interval.getEnd())
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindSmallIndelRegions.java
```diff
@@ -117,6 +122,9 @@ public void test( final GATKRead read, final List<BreakpointEvidence> evidenceLi
                 } else {
                     curContig = read.getContig();
                     curEnd = read.getStart() + BLOCK_SIZE;
+                    // clear 1st-half window when we switch contigs
+                    final int oldIdx = fillIdx ^ 1;
+                    libraryToHistoPairMap.forEach((lib,histPair) -> histPair[oldIdx].clear());
```
pshapiro4broad: IMO it would be clearer to select only the values in the map so it's more apparent that the keys aren't used, e.g.
```
libraryToHistoPairMap.values().forEach(histPair -> histPair[oldIdx].clear());
```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/FindSmallIndelRegions.java
```diff
@@ -117,6 +122,9 @@ public void test( final GATKRead read, final List<BreakpointEvidence> evidenceLi
                 } else {
                     curContig = read.getContig();
                     curEnd = read.getStart() + BLOCK_SIZE;
+                    // clear 1st-half window when we switch contigs
+                    final int oldIdx = fillIdx ^ 1;
```
pshapiro4broad: It would be good to have a test for this fix, if possible.

tedsharpe:
Ready for re-review @pshapiro4broad

PR #3271, opened 2017-07-14
Polish up PathSeq and add pipeline tool

mwalker174:
Adds PathSeqPipelineSpark master tool, which required some minor changes:
-Refactored PSScoreUtils class to a PSScorer, which now includes the main Score tool code
-Moved code for paring down the pathogen header into a new function removeUnmappedHeaderSequences()

Spark-related optimizations:
-Removed cache() calls when possible, and replaced with persist(), spilling to disk with serialization, if necessary
-Removed try-with-resources in Filter and Bwa tools, which seemed to be causing the BWA/kmer references to be unloaded prematurely

Other changes:
-Changed ambiguous base filter from using a fraction of bases to number of bases
-Added function for closing all kmer filter instances
-Optimized PSBwaAligner SA tag construction
-Renamed repartitionPairedReads() to repartitionReadsByName()
-Resolved some conflicting tool argument names
-Created PathSeq tool program group
-Filled out tool summary strings


src/main/java/org/broadinstitute/hellbender/engine/filters/AmbiguousBaseReadFilter.java
```diff
@@ -15,30 +15,29 @@
...
-    public AmbiguousBaseReadFilter(final double n_frac) {
-        this.nFrac = n_frac;
+    public AmbiguousBaseReadFilter(final int maxAmbiguousBases) {
+        this.maxAmbiguousBases = maxAmbiguousBases;
     }
 
     //Filters out reads with more than a threshold number of N's
     @Override
     public boolean test(final GATKRead read) {
-        final int N_max = (int) (read.getLength() * nFrac);
```
magicDGS: I have some code dependent on this filter using the percentage of Ns. If this is strictly necessary, I suggest to add `maxAmbiguousBases` as a nullable `Integer` argument that is mutually exclusive with the fraction, that shouldn't be removed. In that case, this line will be the following and the rest of the method unchanged:

```java
final int N_max = (maxAmiguousBases  != null) ? maxAmiguousBases : (int) (read.getLength() * nFrac);
```

Does this makes sense? Another option is to allow both parameter and use `Math.min` to always filter based on the minimum. Actually this could be more desirable to being able to stablish a hard bound and once it passes use the fraction provided...

mwalker174: Just added your first suggestion to use another variable for maxAmbiguousBases that is null by default but can be set using a different constructor.

magicDGS: Thanks for this, @mwalker174!

src/main/java/org/broadinstitute/hellbender/engine/filters/AmbiguousBaseReadFilter.java
```diff
@@ -7,38 +7,42 @@
...
 /**
- * Filters out reads with more than a threshold number of N's
+ * Filters out reads with more than a threshold number of N's. By default will determine the threshold by the given
+ * maximum fraction of bases in the read. This can be overridden by setting the maximum number of ambiguous bases instead.
  */
 @DocumentedFeature(groupName= HelpConstants.DOC_CAT_READFILTERS, groupSummary=HelpConstants.DOC_CAT_READFILTERS_SUMMARY)
 public final class AmbiguousBaseReadFilter extends ReadFilter {
 
     private static final long serialVersionUID = 1L;
+    private final Integer maxAmbiguousBases;
```
magicDGS: I think that this should be exposed to the user in the command line, because if not it could rise unexpected behavior. Imagine that a default filter is applied using the constructor with maxAmbigBases, and the user provides an `--ambigFilterFrac` argument to the CLI. The fraction specified by the user is ignored without noticing. To expose the argument as mutually-exclusive with the fraction, something like this should be done:

```java
@Argument(doc = "Maximum number of ambiguous bases. If null, use the fraction of ambiguous bases; otherwise, ignores the fraction", fullName = "maxAmbigBases", optional = true, mutex = {"ambigFilterFrac})
public final Integer maxAmbiguousBases = null;
```

It also needs to update the `@Argument` for the maxAmbiguousBaseFraction by adding `mutex = {"maxAmbigBases"}`.

mwalker174: @magicDGS Thanks for your suggestion - I do see how this could be confusing for a user.

In PathSeq, I have a `--filterMaxAmbiguousBases` argument, but it does something fundamentally different from the default read filter because it applies the filter after masking the reads with more N's, whereas default read filters are applied directly to the input. 

For my tool, I will rename my argument `--maxMaskedBases` to be clearer that this filter occurs after applying masking to the reads. Also, I'll print a warning message if the user tried to use `--ambigFilterFrac` or `--ambigFilterBases`, in case they mistook these for the tool's argument.

I also took your suggestion to add the number of bases as an argument, which I named `--ambigFilterBases` for consistency.

magicDGS: Agreed that `--ambigFilterBases` is a better name for keep consitency. Thanks for taking into account my comments and suggestions!

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaArgumentCollection.java
```diff
@@ -7,35 +7,35 @@
 public final class PSBwaArgumentCollection implements Serializable {
     private static final long serialVersionUID = 1L;
 
-    @Argument(doc = "URI of the BWA index image file",
-            fullName = "bwamemIndexImage")
+    @Argument(doc = "URI of the BWA image file for pathogen alignment. This must be distributed to local disk on each node.",
```
cwhelan: Is this for the host (human) reference? Might be helpful to make it explicit.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
```diff
@@ -43,6 +45,25 @@ static void addReferenceSequencesToHeader(final SAMFileHeader header,
...
     /**
+     * Returns header with sequences that were aligned to at least once in reads
+     */
+    static SAMFileHeader removeUnmappedHeaderSequences(final SAMFileHeader header,
+                                                       final JavaRDD<GATKRead> reads,
+                                                       final Logger logger) {
+        final List<String> usedSequences = PSBwaUtils.getAlignedSequenceNames(reads);
+        usedSequences.forEach(seqName -> {
+            if (header.getSequence(seqName) == null)
+                logger.warn("One or more reads are aligned to sequence " + seqName + " but it is not in the header");
```
cwhelan: you could probably fold this logger into the stream process below if you wanted, perhaps by making the `filter` call take a lambda that logs a warning if it returns false.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBwaSpark.java
```diff
@@ -137,13 +125,14 @@ protected void runTool(final JavaSparkContext ctx) {
...
-                throw new UserException.BadInput("No reads were loaded. Ensure --pairedInput and/or --unpairedInput are set and valid.");
-            }
+
+        final PSBwaAlignerSpark aligner = new PSBwaAlignerSpark(ctx, bwaArgs);
+        boolean bPairedSuccess = alignBam(inputPaired, aligner, true, ctx, options, readsSource);
+        boolean bUnpairedSuccess = alignBam(inputUnpaired, aligner, false, ctx, options, readsSource);
+        if (!bPairedSuccess && !bUnpairedSuccess) {
+            throw new UserException.BadInput("No reads were loaded. Ensure --pairedInput and/or --unpairedInput are set and valid.");
         }
+        aligner.close();
```
cwhelan: Do you not care about having this get closed in case of exception any more? Just wondering why you got rid of the try block.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaArgumentCollection.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBwaUtils.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBwaSpark.java
mwalker174: There weren't any exceptions I wanted to catch, I just had it there to automatically call close() when finished. I noticed that in the Filter tool, using a try-with-resources caused the reference to sometimes get closed prematurely, so I removed it there and here.

Issue #2799, opened 2017-06-01
ReferenceBases annotation fails with NPE if reference is null


lbergelson:
The `ReferenceBases` annotation fails with an NPE if there is no reference available.  It should fail with a helpful UserException instead.

```
java.lang.NullPointerException
	at org.broadinstitute.hellbender.tools.walkers.annotator.ReferenceBases.annotate(ReferenceBases.java:33)
	at org.broadinstitute.hellbender.tools.walkers.annotator.VariantAnnotatorEngine.annotateContext(VariantAnnotatorEngine.java:161)
	at org.broadinstitute.hellbender.tools.walkers.annotator.VariantAnnotatorEngineUnitTest.testAllAnnotations(VariantAnnotatorEngineUnitTest.java:225)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:108)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:661)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:869)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1193)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:126)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:109)
	at org.testng.TestRunner.privateRun(TestRunner.java:744)
	at org.testng.TestRunner.run(TestRunner.java:602)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:380)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:375)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:340)
	at org.testng.SuiteRunner.run(SuiteRunner.java:289)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1301)
[...]

PR #3299, opened 2017-07-18
Changed referenceBases to silently fail and emit a single warning if there is no reference

jamesemery:


src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReferenceBases.java
```diff
@@ -22,14 +23,19 @@
...
+
     @Override
     public List<String> getKeyNames() { return Collections.singletonList(REFERENCE_BASES_KEY); }
 
     @Override
     public Map<String, Object> annotate(final ReferenceContext ref,
                                         final VariantContext vc,
                                         final ReadLikelihoods<Allele> likelihoods) {
-
+        if (ref==null)  {
```
magicDGS: I though that https://github.com/broadinstitute/gatk/issues/2799 suggest to throw an `UserException`, which actually makes sense if the user provides an annotation that couldn't be apply because the requirements are not met...

src/main/java/org/broadinstitute/hellbender/utils/logging/OneShotLogger.java
```diff
@@ -0,0 +1,26 @@
+package org.broadinstitute.hellbender.utils.logging;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * A logger wrapper class which only outputs the first warning provided to it
+ */
+public class OneShotLogger {
+    Logger logger;
```
droazen: Mark as `@VisibleForTesting`

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReferenceBases.java
droazen: The warning seems good for now, and consistent with existing behavior in the annotation classes -- we can open a separate ticket to discuss whether there should be a mode in which GATK throws in these cases.

Issue #3221, opened 2017-07-06
Write a tool for filtering locally-assembled contig's alignments


SHuang-Broad:
Right now the alignments are filtered in various places in SV discovery stage. 
Having a single logic unit for doing this makes
  * debugging and modular development, 
  * complex SV discovery/interpretation, and 
  * future improvements (e.g. not filtering but downgrading certain alignments and use an optimization approach)

easier

SHuang-Broad:
Implemented in a branch and results to be investigated.

SHuang-Broad:
Closed by #3432

PR #3240, opened 2017-07-10
Finally re-packaging SV classes; replaces #3232

SHuang-Broad:
Finally working after weeks....
Replaces #3232 and a few previous PR's.
@cwhelan @tedsharpe @mwalker174 please review. Thanks!

PR #3302, opened 2017-07-19
Proposed heuristic alignment filtering of long reads (DONOT MERGE YET)

SHuang-Broad:
Tool used for making the new call set evaluated [here](https://github.com/broadinstitute/dsde-methods-sv/pull/6).

Solves #3221.


SHuang-Broad:
Note that this is NOT  PR into master branch, but a branch after the repackaging PR #3240

SHuang-Broad:
replaced by #3432

PR #3304, opened 2017-07-19
Huge improvement in Mutect active region determination

davidbenjamin:
@takutosato @LeeTL1220 As mentioned, this change scraps all the p values and replaces it with a simple and cheap probabilistic model.  All our validations either improve or stay the same and speed is much better.

* Spurious active regions are reduced by almost 50%.
* DREAM 4 goes from 40 hours total CPU time to 20 hours.  All DREAM genomes now take less than a day.
* Hapmap sensitivity is the same.
* DREAM sensitivities for SNVs and indels all go up a bit.
* Upon manual review we no longer make any obviously bad inactive calls, except for very long deletions, which remain an issue.

@takutosato This is a higher priority review than either of the documentation PRs.


takutosato:
@davidbenjamin Got it, going to review now

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -275,55 +279,79 @@ public void shutdown() {
...
 
-        if (tumorNonRef < expectedTumorNonRefDueToError + MTAC.tumorStandardDeviationsThreshold * tumorNonRefStdev) {
+        if (tumorLog10Odds < MTAC.initialTumorLodThreshold) {
             return new ActivityProfileState(refInterval, 0.0);
+        } else if (hasNormal()) {
+            final ReadPileup normalPileup = pileup.getPileupForSample(MTAC.normalSampleName, header);
+            final Pair<Integer, Double> normalAltCountAndQualSum = altCountAndQualSum(normalPileup, refBase);
+            final int normalAltCount = normalAltCountAndQualSum.getFirst();
+            final double normalQualSum = normalAltCountAndQualSum.getSecond();
+            if (normalAltCount > normalPileup.size() * 0.3 && normalQualSum > 100) {
```
takutosato: parameterize 0.3 and 100

takutosato: How did you pick 0.3 and 100? 0.3 used to be 0.1 (= `minNormalVariantFraction`), is this an attempt to boost sensitivity?

takutosato: I thought you were against qual sum? Maybe use average base quality per read? `(1/d) * \sum_i bq_i`

davidbenjamin:
@takutosato The speedup is due to the fact that far fewer pileups are called active, so we trigger assembly and realignment much less.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
davidbenjamin: In this case the qual sum turns out to be a term in a fairly principled approximation to a likelihood-based calculation.  I'm still against it standing on its own as a heuristic.  I'll run the calculation by you in front of a whiteboard.

davidbenjamin:
Back to @takutosato

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -275,55 +279,79 @@ public void shutdown() {
...
+        final int tumorAltCount = tumorAltCountAndQualSum.getFirst();
+        final int tumorRefCount = tumorPileup.size() - tumorAltCount;
 
-        // since errors are rare, the number of errors (if reads are independent) is approximately a Poisson random variable,
-        // with mean equal to its variance
-        final double expectedTumorNonRefDueToError = StreamSupport.stream(tumorContext.getBasePileup().spliterator(), false)
-                .mapToDouble(pe -> QualityUtils.qualToErrorProb(pe.getQual()))
-                .sum();
-        final double tumorNonRefStdev = Math.sqrt(expectedTumorNonRefDueToError);
+        final double tumorLog10Odds = -QualityUtils.qualToErrorProbLog10(tumorAltCountAndQualSum.getSecond()) +
```
takutosato: Got it, thanks for writing up the doc

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
takutosato: Sounds good!

takutosato:
Please fix the typo and merge!

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -275,55 +281,79 @@ public void shutdown() {
...
+            }
+        } else {
+            final List<VariantContext> germline = featureContext.getValues(MTAC.germlineResource, refInterval);
+            if (!germline.isEmpty() && germline.get(0).getAttributeAsDoubleList(VCFConstants.ALLELE_FREQUENCY_KEY, 0.0).get(0) > MTAC.maxPopulationAlleleFrequency) {
+                return new ActivityProfileState(refInterval, 0.0);
+            }
         }
 
-        if (hasNormal() && normalContext != null && countNonRef(refBase, normalContext) > normalContext.getBasePileup().size() * MTAC.minNormalVariantFraction) {
+        if (!MTAC.genotypePonSites && !featureContext.getValues(MTAC.pon, new SimpleInterval(context.getContig(), (int) context.getPosition(), (int) context.getPosition())).isEmpty()) {
```
byoo: Hi, I understood this excludes variants at PON sites entirely and save computing resources. Consequently, variants falling into this will not be part of Mutect output vcf (correct?).
If so, this seems like an inconsistent behavior because there are other variants at PON sites that are part of Mutect output and explicitly filtered as panel_of_normals as opposed to be excluded silently. 
I wonder if the default value of --genotypePonSites would better be true to be consistent and explicit.

davidbenjamin: I deliberately made `--genotypePonSites` false by default because running local assembly and realignment of PoN sites is very expensive, especially so because PoN sites are frequently in regions that yield very messy assembly graphs, hence many haplotypes.  It's true that explicit results can be useful, and we frequently want them in the course of development, but a tenet of the GATK is to make the tools work as well as possible with default settings.

byoo:
Thank you for the kind explanation, *@davidbenjamin*. I understand your
rationale. BTW, in my test with a PoN of ~100 sample, this set of changes
makes significant performance improvement regardless of the value of
--genotypePonSites. Could you advise how many samples you used to create
the PoN you test? Thanks!

PR #3305, opened 2017-07-19
Fixing hash collision issue when adding filter value to the variant context

### LINKED ISSUE #3291: FilterByOrientationBias gives random failures
Out of 11 runs on exactly the same data, FilterByOrientationBias fails 6 times and succeeds 5 times. Assigning @LeeTL1220, given prior interaction with user.

- User reports this error in: https://gatkforums.broadinstitute.org/gatk/discussion/comment/40412#Comment_40412
- My recapitulation is in: https://github.com/broadinstitute/dsde-docs/issues/2294

Data is at `/humgen/gsa-scr1/pub/incoming/byoo_FilterByOrientationBias.zip`.

Command is:
```
gatk-launch FilterByOrientationBias \
    -A 'G/T' -A 'C/T' \
    -V test2.vcf \
    -P test2.pre_adapter_detail_metrics \
    --output ob_filtered2.vcf
```

Error message changes between:
```
java.lang.IllegalStateException: Allele in genotype C* not in the variant context [G*, T]
	at htsjdk.variant.variantcontext.VariantContext.validateGenotypes(VariantContext.java:1360)
	at htsjdk.variant.variantcontext.VariantContext.validate(VariantContext.java:1298)
	at htsjdk.variant.variantcontext.VariantContext.<init>(VariantContext.java:401)
	at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:494)
	at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:488)
	at org.broadinstitute.hellbender.tools.exome.orientationbiasvariantfilter.OrientationBiasFilterer.annotateVariantContextsWithFilterResults(OrientationBiasFilterer.java:216)
	at org.broadinstitute.hellbender.tools.exome.FilterByOrientationBias.onTraversalSuccess(FilterByOrientationBias.java:168)
	at org.broadinstitute.hellbender.engine.GATKTool.doWork(GATKTool.java:781)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:115)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:170)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:189)
	at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:122)
	at org.broadinstitute.hellbender.Main.mainEntry(Main.java:143)
	at org.broadinstitute.hellbender.Main.main(Main.java:221)
```
and
```
java.lang.IllegalStateException: Allele in genotype G* not in the variant context [C*, T]
	at htsjdk.variant.variantcontext.VariantContext.validateGenotypes(VariantContext.java:1360)
	at htsjdk.variant.variantcontext.VariantContext.validate(VariantContext.java:1298)
	at htsjdk.variant.variantcontext.VariantContext.<init>(VariantContext.java:401)
	at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:494)
	at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:488)
	at org.broadinstitute.hellbender.tools.exome.orientationbiasvariantfilter.OrientationBiasFilterer.annotateVariantContextsWithFilterResults(OrientationBiasFilterer.java:216)
	at org.broadinstitute.hellbender.tools.exome.FilterByOrientationBias.onTraversalSuccess(FilterByOrientationBias.java:211)
	at org.broadinstitute.hellbender.engine.GATKTool.doWork(GATKTool.java:840)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:115)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:170)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:189)
	at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:131)
	at org.broadinstitute.hellbender.Main.mainEntry(Main.java:152)
	at org.broadinstitute.hellbender.Main.main(Main.java:230)
```

I tested using `gatk-package-4.beta.1-local.jar` and pre-release alpha version `gatk-4.alpha.2-1134-ga9d9d91-SNAPSHOT/gatk-package-4.alpha.2-1134-ga9d9d91-SNAPSHOT-local.jar`.

LeeTL1220:
Closes #3291 

- Detects the issue earlier and fails with a better error message that is local (in code) to the problem.
- Removes the spew of warnings that are not useful except in debug.
- Fixes an issue where filter values were being replaced instead of appended.
- Implemented a wrapper to always guarantee sorting never collides.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
@@ -16,6 +16,7 @@ import org.broadinstitute.hellbender.utils.param.ParamUtils;
 import org.broadinstitute.hellbender.utils.variant.GATKVCFConstants;
 import org.broadinstitute.hellbender.utils.variant.GATKVariantContextUtils;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
@@ -58,7 +59,8 @@ public class OrientationBiasFilterer {
 
             final List<Allele> alleles = genotype.getAlleles();
             if (genotype.getPloidy() != 2) {
-                logger.debug("No action required:  This tool will skip non-diploid sites.  Saw GT: " + genotype.getGenotypeString() + " at " + vc.toStringWithoutGenotypes());
+                logger.debug("No action required:  This tool will skip non-diploid sites.  Saw GT: {} at {}",
+                        genotype::getGenotypeString, vc::toStringWithoutGenotypes);
             }
 
             // Get the reference allele as a String and make sure that there is only one ref allele and that it is length
@@ -131,7 +133,7 @@ public class OrientationBiasFilterer {
         }
 
         final List<String> sampleNames = preAdapterQAnnotatedVariants.get(0).getSampleNamesOrderedByName();
-        final Map<String, SortedMap<Genotype, VariantContext>> sampleNameToVariants = createSampleToGenotypeVariantContextSortedMap(sampleNames, preAdapterQAnnotatedVariants);
+        final Map<String, SortedMap<UniqueIDWrapper<Genotype>, VariantContext>> sampleNameToVariants = createSampleToGenotypeVariantContextSortedMap(sampleNames, preAdapterQAnnotatedVariants);
 
         // This map will hold all updated genotypes (across samples)
         final Map<VariantContext, List<Genotype>> newGenotypes = new HashMap<>();
@@ -143,7 +145,7 @@ public class OrientationBiasFilterer {
             //  I.e. only remove filtered or ref/ref genotypes and count the rest.
             final long unfilteredGenotypeCount = OrientationBiasUtils.calculateUnfilteredNonRefGenotypeCount(preAdapterQAnnotatedVariants, sampleName);
 
-            final SortedMap<Genotype, VariantContext> genotypesToConsiderForFiltering = sampleNameToVariants.get(sampleName);
+            final SortedMap<UniqueIDWrapper<Genotype>, VariantContext> genotypesToConsiderForFiltering = sampleNameToVariants.get(sampleName);
 
             // Save some time, especially for the normal sample
             if (genotypesToConsiderForFiltering.keySet().size() == 0) {
@@ -171,7 +173,8 @@ public class OrientationBiasFilterer {
 
             final Map<Transition, Long> transitionCutSoFar = new HashMap<>();
             relevantTransitions.stream().forEach(transition -> transitionCutSoFar.put(transition, 0L));
-            for (final Genotype genotype : genotypesToConsiderForFiltering.keySet()) {
+            for (final UniqueIDWrapper<Genotype> genotypeWrapped : genotypesToConsiderForFiltering.keySet()) {
+                final Genotype genotype = genotypeWrapped.getWrapped();
                 final GenotypeBuilder genotypeBuilder = new GenotypeBuilder(genotype);
                 // Since we only have the ALT_F2R1 and ALT_F1R2 counts for the first alt allele, we will not consider a site where transition is not artifact mode in the first alt allele.
                 final Transition transition = Transition.transitionOf(genotype.getAllele(0).getBaseString().charAt(0), genotype.getAllele(1).getBaseString().charAt(0));
@@ -195,7 +198,7 @@ public class OrientationBiasFilterer {
                     }
                 }
 
-                newGenotypes.computeIfAbsent(genotypesToConsiderForFiltering.get(genotype), v -> new ArrayList<>()).add(genotypeBuilder.make());
+                newGenotypes.computeIfAbsent(genotypesToConsiderForFiltering.get(genotypeWrapped), v -> new ArrayList<>()).add(genotypeBuilder.make());
             }
         }
 
@@ -214,7 +217,7 @@ public class OrientationBiasFilterer {
                 // Add the orientation bias filter to the variant context.
                 if (newGenotypesForThisVariantContext.stream().anyMatch(g -> (g != null) && (g.getFilters() != null) && (g.getFilters().contains(OrientationBiasFilterConstants.IS_ORIENTATION_BIAS_CUT)))) {
                     final List<String> result = GATKVariantContextUtils.createFilterListWithAppend(vc, OrientationBiasFilterConstants.IS_ORIENTATION_BIAS_CUT);
-                    variantContextBuilder.filters(result.toArray(new String[0]));
+                    variantContextBuilder.filters(result.toArray(new String[result.size()]));
                 }
                 final VariantContext updatedVariantContext = variantContextBuilder.make();
                 finalVariants.add(updatedVariantContext);
@@ -228,7 +231,7 @@ public class OrientationBiasFilterer {
     }
 
 
-    private static Map<Transition, Long> createTransitionToNumCutPrePreAdapterQ(double fdrThresh, String sampleName, long unfilteredGenotypeCount, final SortedMap<Genotype, VariantContext> genotypesToConsiderForFiltering, final Map<Transition, Long> transitionCount) {
+    private static Map<Transition, Long> createTransitionToNumCutPrePreAdapterQ(double fdrThresh, String sampleName, long unfilteredGenotypeCount, final SortedMap<UniqueIDWrapper<Genotype>, VariantContext> genotypesToConsiderForFiltering, final Map<Transition, Long> transitionCount) {
         final long allTransitionCount = transitionCount.values().stream().mapToLong(Long::longValue).sum();
         final int totalNumToCut = calculateTotalNumToCut(fdrThresh, unfilteredGenotypeCount, genotypesToConsiderForFiltering);
 
@@ -244,9 +247,9 @@ public class OrientationBiasFilterer {
         return transitionNumToCut;
     }
 
-    private static int calculateTotalNumToCut(final double fdrThresh, final long unfilteredGenotypeCount, final SortedMap<Genotype, VariantContext> genotypesToConsiderForFiltering) {
+    private static int calculateTotalNumToCut(final double fdrThresh, final long unfilteredGenotypeCount, final SortedMap<UniqueIDWrapper<Genotype>, VariantContext> genotypesToConsiderForFiltering) {
         final List<Double> pArtifactScores = genotypesToConsiderForFiltering.keySet().stream()
-                .map(g -> OrientationBiasUtils.getGenotypeDouble(g, OrientationBiasFilterConstants.P_ARTIFACT_FIELD_NAME, 0.0))
+                .map(g -> OrientationBiasUtils.getGenotypeDouble(g.getWrapped(), OrientationBiasFilterConstants.P_ARTIFACT_FIELD_NAME, 0.0))
                 .collect(Collectors.toList());
 
         // When doing the Benjamini-Hochberg procedure, we need to include the non-artifact mode SNVs to guarantee the
@@ -277,12 +280,12 @@ public class OrientationBiasFilterer {
                 .findFirst().orElse(pArtifactScoresIncludingNonArtifact.size() - 1);
     }
 
-    private static Map<Transition, Long> createTransitionCountMap(SortedSet<Transition> relevantTransitions, SortedMap<Genotype, VariantContext> genotypesToConsiderForFiltering) {
+    private static Map<Transition, Long> createTransitionCountMap(SortedSet<Transition> relevantTransitions, SortedMap<UniqueIDWrapper<Genotype>, VariantContext> genotypesToConsiderForFiltering) {
         final Map<Transition, Long> transitionCount = new HashMap<>();
         relevantTransitions.stream().forEach(transition -> transitionCount.put(transition, 0L));
-        for (final Genotype g : genotypesToConsiderForFiltering.keySet()) {
+        for (final UniqueIDWrapper<Genotype> g : genotypesToConsiderForFiltering.keySet()) {
             relevantTransitions.stream()
-                    .filter(transition -> OrientationBiasUtils.isGenotypeInTransition(g, transition))
+                    .filter(transition -> OrientationBiasUtils.isGenotypeInTransition(g.getWrapped(), transition))
                     .forEach(transition -> transitionCount.put(transition, transitionCount.get(transition) + 1));
         }
         return transitionCount;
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
```diff
@@ -58,7 +58,7 @@ public static VariantContext annotateVariantContextWithPreprocessingValues(final
 
             final List<Allele> alleles = genotype.getAlleles();
             if (genotype.getPloidy() != 2) {
-                logger.warn("No action required:  This tool will skip non-diploid sites.  Saw GT: " + genotype.getGenotypeString() + " at " + vc.toStringWithoutGenotypes());
+                logger.debug("No action required:  This tool will skip non-diploid sites.  Saw GT: " + genotype.getGenotypeString() + " at " + vc.toStringWithoutGenotypes());
```
lbergelson: Do we output this at every non-diploid site?  Should we change that?

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
```diff
@@ -210,8 +210,11 @@ private static double calculateFob(final Genotype genotype, final boolean isRele
                 final List<Genotype> newGenotypesForThisVariantContext = newGenotypes.get(vc);
                 newGenotypesForThisVariantContext.forEach(gcc::replace);
                 final VariantContextBuilder variantContextBuilder = new VariantContextBuilder(vc).genotypes(gcc);
+
+                // Add the orientation bias filter to the variant context.
                 if (newGenotypesForThisVariantContext.stream().anyMatch(g -> (g != null) && (g.getFilters() != null) && (g.getFilters().contains(OrientationBiasFilterConstants.IS_ORIENTATION_BIAS_CUT)))) {
-                    variantContextBuilder.filter(OrientationBiasFilterConstants.IS_ORIENTATION_BIAS_CUT);
+                    final List<String> result = GATKVariantContextUtils.createFilterListWithAppend(vc, OrientationBiasFilterConstants.IS_ORIENTATION_BIAS_CUT);
+                    variantContextBuilder.filters(result.toArray(new String[0]));
```
lbergelson: Might as well use the size of the list here.

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
LeeTL1220: I set it to debug.  Most of those are M2 making bad calls anyway.

No action.

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
LeeTL1220: Why does it matter?

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
LeeTL1220: (This tool is not run on germline data)

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
lbergelson: save an array allocation?  it doesn't really matter much

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
```diff
@@ -70,7 +70,7 @@ private VariantContext makeVC(String source, List<Allele> alleles, Genotype... g
     }
 
     private VariantContext makeVC(String source, List<Allele> alleles, String filter) {
-        return makeVC(source, alleles, filter.equals(".") ? null : new LinkedHashSet<>(Collections.singletonList(filter)));
+        return makeVC(source, alleles, filter.equals(".") ? null : new LinkedHashSet<String>(Collections.singletonList(filter)));
```
lbergelson: why did you add a type here? Wasn't it inferred?

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
```diff
@@ -1469,4 +1469,21 @@ public static int calculateGQFromPLs(final int[] plValues) {
...
         return second - first;
     }
+
+    /**
+     *  Create a string with existing filters plus the one to append
+     * @param vc VariantContext to base initial filter values.  Not {@code null}
+     * @param filterToAppend the filter value to append to the strings Not {@code null}
+     * @return String of filter values.  Sorted alphabetically.
+     */
+    public static List<String> createFilterListWithAppend(VariantContext vc, String filterToAppend) {
```
lbergelson: This seems clunky and pretty specific to this one use case.  Why not have a method `addToFilterList(List<String>, String)` which adds and sorts the values instead?

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
```diff
@@ -294,11 +297,12 @@ static int calculateTotalNumToCut(final double fdrThreshold, final long unfilter
...
 
         // Sorts in reverse order (highest p_artifact goes first and will not allow anything to be equal
-        //  unless they share the same reference)
+        //  unless they share the same reference).  Unfortunately, we cannot check the actual reference, so we rely on
+        //  hashCode and the genotype rendered as a string to be universally unique.
         // Note the negative sign is to sort in reverse error.
         final Comparator<Genotype> genotypePArtifactComparator = Comparator
                 .comparingDouble((Genotype g) -> -OrientationBiasUtils.getGenotypeDouble(g, OrientationBiasFilterConstants.P_ARTIFACT_FIELD_NAME, 0.0))
-                .thenComparingInt(g -> g.hashCode());
+                .thenComparingInt(g -> g.hashCode()).thenComparing(Genotype::toString);
```
lbergelson: This still seems like a gross hack to me.  It seems like it would be better to solve this problem properly rather than just kicking it out to less and less probable territory.

Why not create a private subclass of Genotype that includes a static atomic counter that's incremented each time one is created, and then use that in the comparator?

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
magicDGS: Maybe use the `{}` syntax of the logger to avoid the construction of the genotype/variant strings:

```java
logger.debug("No action required:  This tool will skip non-diploid sites.  Saw GT: {} at {}",
    genotype::getGenotypeString, vc::toStringWithoutGenotypes);
```

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
lbergelson: Alternatively you could use a stateful comparator that tracks collision resolutions to convert the partial ordering by score into a TreeMap compatible total ordering.  http://gafter.blogspot.com/2007/03/compact-object-comparator.html

droazen: :+1: to this idea

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
LeeTL1220: Done

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
LeeTL1220: Not according to IntelliJ.

No action.

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
LeeTL1220: Doing UniqueIdWrapper.  Makes the code a bit messy, but looks like it works.  Also, slows down the tool a bit, but not drastically.

Done

LeeTL1220:
@lbergelson Please review again...  Thanks

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
LeeTL1220: Then it can just be appending to a list and sorting.  And this is in GATKVariantContextUtils, after all.

I liked the convenience of just being able to get an updated filter set in one call.

No action, unless you strongly disagree.  

No action.

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
```diff
@@ -322,6 +329,21 @@ static int calculateTotalNumToCut(final double fdrThreshold, final long unfilter
...
+     *
+     * @param genotypesToConsiderForFiltering SortedMap, changed in place
+     * @param genotype key to insert into the sorted map
+     * @param vc value to insert into the sorted map
+     */
+    private static void putDisallowingKeyOverwrite(SortedMap<UniqueIDWrapper<Genotype>, VariantContext> genotypesToConsiderForFiltering, Genotype genotype, VariantContext vc) {
+
+        final VariantContext vcTemp = genotypesToConsiderForFiltering.putIfAbsent(new UniqueIDWrapper<>(genotype), vc);
+        if (vcTemp != null) {
+            throw new GATKException.ShouldNeverReachHereException("This error may disappear on a subsequent run of this tool, so please try running this tool again.  Otherwise, this can only be fixed by a GATK developer.  Attempting to overwrite a key in the genotypes for filtering map.  This means that genotypePArtifactComparator is (still) not generating unique keys.  See https://github.com/broadinstitute/gatk/issues/3291");
```
lbergelson: This comment is out of date now, but it's not a bad idea to keep the check in place.

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
```diff
@@ -362,4 +384,23 @@ public static VCFHeader createVCFHeader(final VCFHeader inputVCFHeader, final St
         final Set<String> sampleNameSet = samples.asSetOfSamples();
         return new VCFHeader(headerLines, sampleNameSet);
     }
+
+    static class UniqueIDWrapper<A>{
```
lbergelson: If it would make the code cleaner you could drop the type parameter and just hardcode it as a genotype.  Probably want to add a comment explaining why we have this class.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
lbergelson: If you're going to leave it as is do you want to bake the conversion to array into it as well?

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
lbergelson: Interesting, your intellij and my intellij don't agree for some reason.

lbergelson:
@LeeTL1220 A few minor remaining comments.  Do what you will.  How much of a performance impact does the change have?  You said it slows it down, is it significant?  It might be faster if you make it a long instead of an atomic long which should be safe it it's single threaded and you don't use parallel streams anywhere.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtils.java
LeeTL1220: Leave as is

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
LeeTL1220: Thanks

No action

LeeTL1220:
@lbergelson Performance hit is not enough to worry about.

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
LeeTL1220: Moving to a separate class.

src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
pshapiro4broad: @LeeTL1220 maybe your intellij isn't set up to use JDK7+ features? You should be able to fix this on the Project Settings page.

lbergelson:
@LeeTL1220 👍 Good to merge.

sooheelee:
I have notified the user who reported this error of the fix.

Issue #2443, opened 2017-03-08
UserExceptions should explain how to see the stacktrace


lbergelson:

cmnbroad:
@lbergelson I think most use cases for this would be for post-arg-parsing problems, so it would make sense to add an Advanced, common command line argument for it. Much easier to use, and as you say we could include instructions in the exception message itself for how to enable it.

magicDGS:
Do you think that this is really necessary for a normal user? I mean, usually `UserException` should have a well-defined message String for point to the user what happened. If the stack traces are necessary, are for debugging and I think that the final user won't benefit for having an extra argument.

In addition, it is in the repository README, which made it discoverable for developers and it is what it is really meant for (I guess). Other option may be to print the stack trace for `UserException`only if the verbosity is set to DEBUG, and that will get rid of the environmental variable....

lbergelson:
@cmnbroad 👍 to adding an advanced command line option for it.  

@magicDGS Our goal is to make it unnecessary for normal users to ever need to see a stacktrace.  We're definitely not at the point yet where every UserException produces either a) the complete information necessary to debug, or even b) the correct information. We're trying to fix all those cases, but there's a lot of possible failure modes between cloud access, spark, filesystem plugins, etc, so it's going to be an issue for a while.  

I don't think it will hurt the user experience to have an extra commandline argument for it.  Printing the stacktrace when debug is on isn't a bad idea, but I think it's better to have finer grained control over it.  

The other issue is that it's easier to explain to an unsophisticated user how to set an extra commmand line argument rather than trying to get them to set the environment variable which well be helpful for our support team when they're trying to debug someone's problem.  (especially since setting the environment variables may be different on a spark cluster than on a local run).

PR #3307, opened 2017-07-19
Fix stack trace error message.

cmnbroad:
When an error occurs, we currently print this:

> Use -DGATK_STACKTRACE_ON_USER_EXCEPTIONto print the stack trace.

In addition to missing a space, this doesn't work with gatk-launch. Replace this with:

> Use the system property GATK_STACKTRACE_ON_USER_EXCEPTION (--javaOptions '-DGATK_STACKTRACE_ON_USER_EXCEPTION=true') to print the stack trace.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -195,7 +195,9 @@ public class Main {
         if(printStackTraceOnUserExceptions()) {
             e.printStackTrace();
         } else {
-            System.err.println("Use -D" + STACK_TRACE_ON_USER_EXCEPTION_PROPERTY + "to print the stack trace.");
+            System.err.println("Use the system property "
+                            + STACK_TRACE_ON_USER_EXCEPTION_PROPERTY
+                            + " (--javaOptions '-DGATK_STACKTRACE_ON_USER_EXCEPTION=true') to print the stack trace.");
         }
     }
 
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -195,9 +195,10 @@ public class Main {
         if(printStackTraceOnUserExceptions()) {
             e.printStackTrace();
         } else {
-            System.err.println("Use the system property "
-                            + STACK_TRACE_ON_USER_EXCEPTION_PROPERTY
-                            + " (--javaOptions '-DGATK_STACKTRACE_ON_USER_EXCEPTION=true') to print the stack trace.");
+            System.err.println(String.format(
+                    "Set the system property %s (--javaOptions '-D%s=true') to print the stack trace.",
+                    STACK_TRACE_ON_USER_EXCEPTION_PROPERTY,
+                    STACK_TRACE_ON_USER_EXCEPTION_PROPERTY));
         }
     }
 


src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -195,7 +195,9 @@ protected void handleUserException(Exception e) {
         if(printStackTraceOnUserExceptions()) {
             e.printStackTrace();
         } else {
-            System.err.println("Use -D" + STACK_TRACE_ON_USER_EXCEPTION_PROPERTY + "to print the stack trace.");
+            System.err.println("Use the system property "
+                            + STACK_TRACE_ON_USER_EXCEPTION_PROPERTY
+                            + " (--javaOptions '-DGATK_STACKTRACE_ON_USER_EXCEPTION=true') to print the stack trace.");
```
magicDGS: The `--javaOptions` is not an option in the GATK-framework but specific gatk-launch. This can cause problems in downstream projects using `java -jar ToolKit.jar` syntax...

magicDGS:
Please, @lbergelson, take into account that this will be confusing for downstream projects before accepting this PR. Thanks in advance!

src/main/java/org/broadinstitute/hellbender/Main.java
lbergelson: @magicDGS this is a good point, thanks for pointing that out.  

Would expanding the comment to say something like "Use the system property STACKTRACE to print a stack trace when user exceptions occur.   If you're running with gatk-launch you can set it with "--javaOptions '-DSTACKTRACE=true".

Would that work for you?  Alternatively, you could override the method in your Main instance.

lbergelson: @cmnbroad I obviously didn't pay enough attention here because I missed the fact that the property name is hardcoded now in part of the message, we should have it rely on the constant instead.

magicDGS: If the gatk-launch is mentioned, it works for avoiding confusion while runing with the `java -jar` syntax. Although I think that I will override anyway as you suggested, I would like to ask/point out the following:

* Is it really necessary to point out that the system properties for gatk-launch should be set  by `--javaOptions`? I think that if a user is advance enough to print the stack-trace, they will be familiar on how to provide to gatk-launch the property. But most of the times I am wrong, so maybe it is really needed.
* One of my worries about this line hardcoded here is not only for downstream projects but for the GATK toolkit itself: what will happen if the gatk-launch option or scrip name changes? Although it is unlikely for mature problems, this can be easily overlooked and can get out of sync.

droazen: It is important to mention `--javaOptions` explicitly, as that is a question we get asked a lot by our users.

cmnbroad: @magicDGS Agreed that the embedded strings are fragile, but the current message is almost always incorrect, and its doubly frustrating since it leads the user, who is already on an error path, to a different error. I think we did previously open up this method for override at your request, so that should help somewhat for your case.

cmnbroad: Also, I updated the code to use the constant.

magicDGS: Maybe in the near future will be easier to set this with a special command line argument to make it work for downstream projects too (as discussed in https://github.com/broadinstitute/gatk/issues/2443). Anyway, I will override the method as a temporary solution...

Thanks for the discussion!

Issue #2885, opened 2017-06-05
CalculateTargetCoverage command line printed in header of output is incorrect.


droazen:
@samuelklee commented on [Thu Sep 01 2016](https://github.com/broadinstitute/gatk-protected/issues/663)

--groupBy parameter is output in lowercase, but is required to be input as uppercase.  Thus, if one tries to rerun the printed command line, an exception will be thrown.

cmnbroad:
I randomly came across this ticket - this is happening because the GroupBy enum in CalculateTargetCoverage [overrides toString](https://github.com/broadinstitute/gatk/blob/b58baa5ff8d69e69ae4ac28c863a4126529cf094/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java#L675) for some reason, and returns a lower cased string. Sending to @davidbenjamin for routing since he looks like the main contributor for that file.

cmnbroad:
Loos good to me.

davidbenjamin:
Closed by #3314.

PR #3314, opened 2017-07-19
undid override of CalculateTargetCoverage.GroupBy::toString

davidbenjamin:
This seems to address issue #2885.  @cmnbroad @samuelklee


samuelklee:
:+1:

PR #3323, opened 2017-07-21
Changed the build_docker.sh to run git LFS dependent tests

### LINKED ISSUE #3191: build_docker.sh needs to download and mount the large files when running unit tests
Currently the script fails when configured to run tests, since it doesn't have access to the large files in the docker image. These need to be downloaded and mounted for the tests to pass, as we do in the docker tests on travis.

jamesemery:
Also deletes the staging directory when finished.

fixes #3191

### DIFF
--- a/build_docker.sh
+++ b/build_docker.sh
@@ -75,6 +75,8 @@ if [ -n "${IS_PUSH}" ]; then
 	docker login
 fi
 
+ORIGINAL_WORKING_DIRECTORY=$(pwd)
+
 if [ -n "$STAGING_DIR" ]; then
     GITHUB_DIR="tags/"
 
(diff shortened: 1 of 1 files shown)


build_docker.sh
```diff
@@ -89,7 +91,9 @@ if [ -n "$STAGING_DIR" ]; then
     set -e
     GIT_LFS_SKIP_SMUDGE=1 git clone https://github.com/${REPO}/${PROJECT}.git ${STAGING_DIR}/${STAGING_CLONE_DIR}
     cd ${STAGING_DIR}/${STAGING_CLONE_DIR}
-    echo "Now in ${PWD}"
+    STAGING_ABSOLUTE_PATH=$(pwd)
+
+    echo "Now in $(pwd)"
         if [ ${PULL_REQUEST_NUMBER} ]; then
```
lbergelson: the indentation here is funny although it looks like it was like that before you got here...

PR #3328, opened 2017-07-21
Added 'nio_max_retries' option as a command line acessable option for command line tools

jamesemery:


src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -365,16 +369,16 @@ public static void setGlobalNIODefaultOptions() {
         final String[] split = gcsUrl.split("/", -1);
         final String BUCKET = split[2];
         final String pathWithoutBucket = String.join("/", Arrays.copyOfRange(split, 3, split.length));
-        CloudStorageConfiguration cloudConfig = getCloudStorageConfiguration();
+        CloudStorageConfiguration cloudConfig = getCloudStorageConfiguration(DEFAULT_NIO_MAX_REOPENS);
```
jonn-smith: This call will only use the config with DEFAULT_NIO_MAX_REOPENS rather than what is set on the command-line.  You may need to pass in the new nioMaxReopens to this method.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -74,6 +74,9 @@
     @Argument(fullName = "use_jdk_inflater", shortName = "jdk_inflater", doc = "Whether to use the JdkInflater (as opposed to IntelInflater)", common=true)
     public boolean useJdkInflater = false;
 
+    @Argument(fullName = "nio_max_retries", shortName = "nio_retries", doc = "If the NIO channel errors out, how many times it will attempt to retry", optional = true)
+    private int NIO_MAX_REOPENS = 20;
```
jamesemery: Responded by relying on the globally set defaults instead.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
jamesemery: Responded by relying on the globally set defaults instead.

jamesemery:
@jonn-smith I responded to your comments.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -74,6 +74,9 @@
     @Argument(fullName = "use_jdk_inflater", shortName = "jdk_inflater", doc = "Whether to use the JdkInflater (as opposed to IntelInflater)", common=true)
     public boolean useJdkInflater = false;
 
+    @Argument(fullName = "nio_max_retries", shortName = "nio_retries", doc = "If the NIO channel errors out, how many times it will attempt to retry", optional = true)
+    public int NIO_MAX_REOPENS = BucketUtils.DEFAULT_NIO_MAX_REOPENS;
```
droazen: Rename this to `gcs_max_retries` and explain in the doc that it applies to Google Cloud Storage requests -- the term "NIO" is meaningless to our users.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -39,7 +40,7 @@
     public static final String FILE_PREFIX = "file:";
 
     // if the channel errors out, re-open up to this many times
-    public static final int NIO_MAX_REOPENS = 20;
+    public static final int DEFAULT_NIO_MAX_REOPENS = 20;
```
droazen: Rename to `GCS_MAX_REOPENS`

src/main/java/org/broadinstitute/hellbender/utils/nio/ReadsIterable.java
```diff
@@ -48,7 +48,7 @@ public ReadsIterator() throws IOException {
             // set high-level retries to deal with servers that might be temporarily overloaded
             // while we're reading a very long file from them.
             SeekableByteChannelPrefetcher chan = new SeekableByteChannelPrefetcher(
-                Files.newByteChannel(fpath, CloudStorageOptions.withChannelReopen(BucketUtils.NIO_MAX_REOPENS)), BUFSIZE);
+                Files.newByteChannel(fpath), BUFSIZE);
```
droazen: Verify that the default reopens as set in `setGlobalNIODefaultOptions()` are applied here in the absence of passing it in explicitly.

jamesemery: It appears that this codepath is only touched by the unconnected class ExampleNioCountReads.java. I have run an integration test of this class and found that the GCS max retries is properly as it implements CommandLineTool

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
jamesemery: Responded

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
jamesemery: Responded

jonn-smith:
Looks good.  Merging.

Issue #3320, opened 2017-07-21
Improve error message when Genotype PL and PP do not have 3 items when calculatePosteriorGLs


ronlevine:
Port https://github.com/broadinstitute/gsa-unstable/pull/1614.

PR #3329, opened 2017-07-21
Make error informative for non-diploid family likelihoods

ronlevine:
Implements https://github.com/broadinstitute/gatk/issues/3320.
Was throwing a ArrayIndexOutOfBoundsException because the number of PLs is set to 3. This is only true for diploid biallelics.


ronlevine:
@vdauwera Please review.

ronlevine:
Ping @vdauwera

PR #3331, opened 2017-07-21
Porting over gvcf validation option to ValidateVariants

jsotobroad:
Wanted to add the functionality added in this commit to gatk3 - https://github.com/broadinstitute/gsa-unstable/commit/56ac3ad78c34ad5f7b7b067770ff12fcef4f1d99


src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -140,28 +142,39 @@
...
     @Argument(fullName = "warnOnErrors", shortName = "warnOnErrors", doc = "just emit warnings on errors instead of terminating the run at the first instance", optional = true)
     Boolean WARN_ON_ERROR = false;
 
     /**
+     *  Validate this file as a gvcf. In particular, every variant must include a <NON_REF> allele, and that
+     *  every base in the territory under consideration is covered by a variant (or a reference block).
+     *  If you specifed intervals (using -L or -XL) to restrict analysis to a subset of genomic regions,
+     *  those intervals will need to be covered in a valid gvcf.
+     */
+    @Argument(fullName = "validateGVCF", shortName = "gvcf", doc = "Validate this file as a GVCF", optional = true, mutex = "doNotValidateFilteredRecords")
```
lbergelson: Could you pull this argument name out as a public constant in the class and then use the constant since it's referenced in two places now?  Likewise with the doNotValidateFilteredRecords argument name.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -140,28 +142,39 @@
...
+    /**
      * Contains final set of validation to apply.
      */
     private Collection<ValidationType> validationTypes;
 
-    @Override
+    private GenomeLocSortedSet genomeLocSortedSet;
+
     public void onTraversalStart() {
+        if (VALIDATE_GVCF)
```
lbergelson: You lost the override here for some reason.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -140,28 +142,39 @@
...
+            genomeLocSortedSet = new GenomeLocSortedSet(new GenomeLocParser(getBestAvailableSequenceDictionary()));
```
lbergelson: I think you want to check that getBestAvailableSequenceDictionary() didn't return null, and if it did throw a helpful UserException.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -225,6 +269,13 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
         }
     }
 
+    private void ValidateGVCFVariant(final VariantContext vc) {
```
lbergelson: This should start with a lowercase letter, in java only types should start with a capital, ex classes and interfaces.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -225,6 +269,13 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
...
+        if (!vc.hasAllele(GATKVCFConstants.NON_REF_SYMBOLIC_ALLELE)) {
+            throw new TribbleException.InternalCodecException(String.format("In a GVCF all records must contain a %s allele. Offending record: %s",
```
lbergelson: This doesn't seem like the right exception to throw.  It seems like the others throw a UserException if validation fails, shouldn't this match?  Should it respect the warning only mode that other validations seem to handle?

src/test/resources/org/broadinstitute/hellbender/tools/walkers/ValidateVariants/gvcf.basepairResolution.vcf
```diff
@@ -0,0 +1,10078 @@
+##fileformat=VCFv4.1
+##ALT=<ID=NON_REF,Description="Represents any possible alternative allele at this location">
```
lbergelson: This is a pretty big file.  Could you shrink it to fewer lines and still have a useful test or is the length important?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -203,6 +241,12 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
      * @return the final set of type to validate. May be empty.
      */
     private Collection<ValidationType> calculateValidationTypesToApply(final List<ValidationType> excludeTypes) {
+        if (VALIDATE_GVCF && !excludeTypes.contains(ValidationType.ALLELES)) {
```
lbergelson: Should this be an error instead of a warning?  Why is GVCF validation implemented with it's own argument instead of being part of validation enum? (just curious, don't have to change it necessarily...)

src/test/resources/org/broadinstitute/hellbender/tools/walkers/ValidateVariants/gvcf.basepairResolution.vcf
jsotobroad: I don't think it should matter, I cut it down a bit.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
jsotobroad: I think if we added GVCF to the validation enum then wouldn't everyone need to exclude it in their command lines when validating vcfs?

jsotobroad:
lots of copy pasta problems you noticed, thanks for that!  Should be good for a second round when you get a chance @lbergelson

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
lbergelson: That's a good reason.

src/test/resources/org/broadinstitute/hellbender/tools/walkers/ValidateVariants/gvcf.basepairResolution.vcf
lbergelson: 👍

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -85,6 +88,9 @@
 public final class ValidateVariants extends VariantWalker {
     static final Logger logger = LogManager.getLogger(ValidateVariants.class);
 
+    public static final String validateGVCF = "validateGVCF";
```
lbergelson: @jsotobroad Our convention is that static constants are UPPERCASE_WITH_UNDERSCORES.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -250,4 +309,12 @@ private void applyValidationType(VariantContext vc, Allele reportedRefAllele, Al
                 break;
         }
     }
-}
\ No newline at end of file
+
+    private void throwOrWarn(UserException e) {
```
lbergelson: Good idea!

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -170,11 +194,16 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
...
+            genomeLocSortedSet.add(genomeLocSortedSet.getGenomeLocParser().createGenomeLoc(ref.getInterval().getContig(), ref.getInterval().getStart(), vc.getEnd()), true);
+            validateGVCFVariant(vc);
+        }
+
         for (final ValidationType t : validationTypes) {
             try{
                 applyValidationType(vc, reportedRefAllele, observedRefAllele, rsIDs, t);
             } catch (TribbleException e) {
-                if ( WARN_ON_ERROR ) {
+                if (WARN_ON_ERROR) {
```
lbergelson: You probably want to use throwOrWarn here too.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -164,9 +168,16 @@
...
 
+    @Override
     public void onTraversalStart() {
-        if (VALIDATE_GVCF)
-            genomeLocSortedSet = new GenomeLocSortedSet(new GenomeLocParser(getBestAvailableSequenceDictionary()));
+        if (VALIDATE_GVCF) {
+            SAMSequenceDictionary dict = getBestAvailableSequenceDictionary();
+
+            if (dict == null)
+                throw new UserException("Could not find valid sequence dictionary for this command.");
```
lbergelson: This would be more useful if it was a bit more specific I think.  ex:  something like: "Validating a GVCF requires a sequence dictionary but no dictionary was found in your input."

Issue #3173, opened 2017-06-26
Small documentation issues


vdauwera:
This ticket aims to centralize small documentation errors such as typos and syntax errors that can be addressed in bulk.

- [x] HaplotypeCaller doc has some syntax errors in links causing entire paragraphs to be included in the link. 
- [x] CollectAllelicCounts has a syntax error that causes a code format block to extend to most of the page (probably a missing closing tag). 
- [x] CalculateContamination has a missing `</pre>` tag that also causes a code format block to be extended to the rest of the page.

PR #3332, opened 2017-07-21
Minor doc fixes - addresses issue 3173

vdauwera:
Addresses issue #3173 

I'd welcome advice from @cmnbroad on how to deal with the issue in CalculateContamination where the `@` starting the header lines in the example of Picard interval list breaks the docgen.  My hacky workaround is to add a dot before the @ but it's icky.


vdauwera:
Ack I meant CollectAllelicCounts

cmnbroad:
@vdauwera You should be able to just replace the "@" with "{@literal @}". You can also use the html entities for these kinds of things, but I can never remember which ones are which.

vdauwera:
ah great thanks @cmnbroad !

vdauwera:
Thanks @lbergelson

PR #3335, opened 2017-07-23
Fix stack overflow at high depths in strand artifact annotation

davidbenjamin:
due to recursive implementation of Legendre abscissas in Apache Commons.

@vdauwera @takutosato this is very simple; it just caps the number of subdivisions of the integral to avoid the recursive stack overflow.  I tested it on absurdly high coverage (100,000) and reproduced the error with the old code.  Whichever one of you gets to this first should review.

While this isn't the most beautiful thing in the world, it will work reasonably while new integration code is pending.


takutosato:
I'll review

PR #3347, opened 2017-07-24
Fixes for FOB being too generous, broken FOB summary table, oncotator MAF override values, and updating tests.

### LINKED ISSUE #3309: FilterByOrientationBias summary table always lists that there were no variants filtered, even though there were in the VCF
For example:
```
lichtens@OncobuntuMk6:~$ more /home/lichtens/work/ob_filtered2.tmp.vcf.summary 
Sample	Artifact_Mode	Artifact_Mode_Complement	OBQ	Num_Artifact_Mode	Num_Artifact_Mode_Filtered	Num_Not_Artifact_Mode	Num_NonRef_Passing_Variants
SAMPLE1Tumor	C>T	G>A	42.68076887123618	82339	0	205577	287916
SAMPLE1Tumor	G>T	C>A	32.01295410609514	36517	0	251399	287916
lichtens@OncobuntuMk6:~$ egrep orientation_bias /home/lichtens/work/ob_filtered2.tmp.vcf | wc -l
26279

```

Not high priority.

Just stems from the VariantContext now being filtered as well as the genotype.  Fix can be placed in ``OrientationBiasUtils.getGenotypeStream``

### LINKED ISSUE #3343: M2 WDL does not override oncotator for the case and control samples.
When oncotator tries to infer this, it can be incorrect.  Since the WDL knows the correct answer, it should pass this into Oncotator.

### LINKED ISSUE #3344: FilterByOrientationBias is not aggressive enough when two artifact modes are specified and one has a suppression factor of 0.0
In this case, the FDR threshold is not honored.  The explanation of this is complex, but essentially has to do with the Benjamini-Hochberg procedure not playing well with suppression factor when extended to more than one artifact mode.

The definition of ``FilterByOrientationBias`` will have to be changes from "guaranteeing less than 1% FDR over all mutations" to "guaranteeing less than 1% FDR in each specified artifact mode".  This could make the filter more aggressive, so we may have to adjust the FDR threshold.

- [x] code fix
- [x] doc updates

LeeTL1220:
Closes #3344 
Closes #3343 
Closes #3309

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
@@ -16,6 +16,7 @@ import org.broadinstitute.hellbender.utils.genotyper.SampleList;
 import org.broadinstitute.hellbender.utils.param.ParamUtils;
 import org.broadinstitute.hellbender.utils.variant.GATKVCFConstants;
 import org.broadinstitute.hellbender.utils.variant.GATKVariantContextUtils;
+
 import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
@@ -204,6 +205,17 @@ public class OrientationBiasFilterer {
 
         // Create the final variants with the modified genotypes
         logger.info("Updating genotypes and creating final list of variants...");
+        return createFilteredVariantContextsIfAnyGenotypesAreFiltered(preAdapterQAnnotatedVariants, newGenotypes);
+    }
+
+    /**
+     *  Simply put, if any genotype within a variant context is filtered, then mark the variant context as filtered.
+     *
+     * @param preAdapterQAnnotatedVariants
+     * @param newGenotypes
+     * @return Updated list of variant contexts
+     */
+    private static List<VariantContext> createFilteredVariantContextsIfAnyGenotypesAreFiltered(List<VariantContext> preAdapterQAnnotatedVariants, Map<VariantContext, List<Genotype>> newGenotypes) {
         final List<VariantContext> finalVariants = new ArrayList<>();
         final ProgressMeter resultProgressMeter = new ProgressMeter();
         resultProgressMeter.start();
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
```diff
@@ -204,6 +205,17 @@ private static double calculateFob(final Genotype genotype, final boolean isRele
 
         // Create the final variants with the modified genotypes
         logger.info("Updating genotypes and creating final list of variants...");
+        return createFilteredVariantContextsIfAnyGenotypesAreFiltered(preAdapterQAnnotatedVariants, newGenotypes);
+    }
+
+    /**
+     *  Simply put, if any genotype within a variant context are filtered, then mark the variant context as filtered.
```
davidbenjamin: is filtered

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
```diff
@@ -231,36 +243,31 @@ private static double calculateFob(final Genotype genotype, final boolean isRele
...
-        // When doing the Benjamini-Hochberg procedure, we need to include the non-artifact mode SNVs to guarantee the
-        //  FDR is kept to the specified threshold.
-        //   Pad the list of pArtifact scores so that the the non-artifacts (w/ a p-value of zero) are included.
-        final int numToPadZeroes = (int) unfilteredGenotypeCount - pArtifactScores.size();
-        List<Double> finalPArtifactScores = new ArrayList<>();
-        finalPArtifactScores.addAll(pArtifactScores);
-        finalPArtifactScores.addAll(Collections.nCopies(numToPadZeroes, 0.0));
-
-        return calculateTotalNumToCut(fdrThresh, unfilteredGenotypeCount, finalPArtifactScores);
+        return calculateTotalNumToCut(fdrThresh, pArtifactScores.size(), pArtifactScores);
```
davidbenjamin: Can this method now take only two arguments since `pArtifactScores.size()` can be obtained from `pArtifactScores` itself?

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasUtils.java
```diff
@@ -248,17 +249,27 @@ static long calculateNumNotTransition(final String sampleName, final List<Varian
                 .filter(g -> OrientationBiasUtils.isGenotypeInTransition(g, complement) || OrientationBiasUtils.isGenotypeInTransition(g, transition));
     }
 
+    private static Stream<Genotype> getNumArtifactGenotypeStreamKeepingOBFilteredVCs(String sampleName, List<VariantContext> variantContexts, Transition transition, Transition complement) {
```
davidbenjamin: Does `complement` really need to be an argument i.e. how much do you save by not just calculating on the fly?

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasUtils.java
```diff
@@ -248,17 +249,27 @@ static long calculateNumNotTransition(final String sampleName, final List<Varian
...
+        return getGenotypeStreamKeepingOBFilteredVCs(sampleName, variantContexts)
+                .filter(g -> OrientationBiasUtils.isGenotypeInTransition(g, complement) || OrientationBiasUtils.isGenotypeInTransition(g, transition));
+    }
     private static Stream<Genotype> getGenotypeStream(String sampleName, List<VariantContext> variantContexts) {
         return variantContexts.stream()
                 .filter(vc -> !vc.isFiltered())
                 .map(vc -> vc.getGenotype(sampleName));
     }
 
+    private static Stream<Genotype> getGenotypeStreamKeepingOBFilteredVCs(String sampleName, List<VariantContext> variantContexts) {
```
davidbenjamin: Can you absorb this into `getNumArtifactGenotypeStreamKeepingOBFilteredVCs`?

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasFilterer.java
LeeTL1220: Technically, yes, but I left it separate in case we want to (effectively) zero pad.  I just added another method that will grab the .size() for this case.

LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasUtils.java
LeeTL1220: I had already made a method for that -- I just forgot to use it.

Done

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasUtils.java
LeeTL1220: I thought it might be useful later, but I've rolled it in.

Done

PR #3351, opened 2017-07-24
Simplify Mutect annotations

davidbenjamin:
@LeeTL1220 Could you take a look at this?  In particular, I want to be sure that this will not change the functionality of the orientation bias filter, even though it expands the annotation to all alleles.


LeeTL1220:
@davidbenjamin You have conflicts.  Can you rebase before I review?

davidbenjamin:
@LeeTL1220 rebased

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVCFConstants.java
```diff
@@ -94,11 +94,8 @@
     public static final String TRANSMISSION_DISEQUILIBRIUM_KEY =    "TDT";
     public static final String VARIANT_TYPE_KEY =                   "VariantType";
     public static final String VQS_LOD_KEY =                        "VQSLOD";
-    public static final String OXOG_ALT_F1R2_KEY =                  "ALT_F1R2";
-    public static final String OXOG_ALT_F2R1_KEY =                  "ALT_F2R1";
-    public static final String OXOG_REF_F1R2_KEY =                  "REF_F1R2";
-    public static final String OXOG_REF_F2R1_KEY =                  "REF_F2R1";
-    public static final String OXOG_FRACTION_KEY =                  "FOXOG";
+    public static final String F1R2_KEY =                           "F1R2";
```
LeeTL1220: Please keep the old name to distinguish whether it is ALT or REF.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/OxoGReadCounts.java
```diff
@@ -34,26 +38,16 @@
...
 public final class OxoGReadCounts extends GenotypeAnnotation {
 
-    public static final Allele REF_C = Allele.create(C.base, true);
-    public static final Allele REF_A = Allele.create(A.base, true);
-
     @Override
     public List<String> getKeyNames() {
-        return Arrays.asList(GATKVCFConstants.OXOG_ALT_F1R2_KEY,
-                             GATKVCFConstants.OXOG_ALT_F2R1_KEY,
-                             GATKVCFConstants.OXOG_REF_F1R2_KEY,
```
LeeTL1220: Before you delete REF_F2R1, etc, did you check with any users to see if they use it?  I don't, but you might want to double check with someone in CGA (for example).

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/OxoGReadCounts.java
```diff
@@ -65,55 +59,37 @@ public void annotate(final ReferenceContext refContext,
...
-                } else if (allele.equals(alt, true)) {
-                    if (read.isReverseStrand() == read.isFirstOfPair()) {
-                        alt_F2R1++;
-                    } else {
-                        alt_F1R2++;
-                    }
-                }
-            }
-        }
+        final Map<Allele, MutableInt> f1r2Counts = likelihoods.alleles().stream()
```
LeeTL1220: If you do ALT_F1R2 and ALTF2R1 counts per allele (I'm assuming it will be on a genotype), we need a test that this works for FilterByOrientationBias.  The current behavior is to ditch all but the first allele and that was done because the counts were per genotype.  I don't want to leave that hack in place if we can do the right thing.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVCFConstants.java
LeeTL1220: Both in the variable name and the string.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/OxoGReadCounts.java
davidbenjamin: For this and the above comment, how about I just make this `VCFHeaderLineCount.R` instead of `VCFHeaderLineCount.A`?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/OxoGReadCounts.java
davidbenjamin: I changed the annotation to do all alleles (still a genotype annotation) but left the filtering to only look at the first alt allele -- the pipeline's output should be identical.  I don't think I understand what you're getting at about FilterByOrientationBias.  Do you mean including multiallelic variants in our tests?  

Also, I agree that looking only at the first allele is a hack but I think addressing that should be in the context of a big refactoring of our filtering engine, which does similar things elsewhere.

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVCFConstants.java
takutosato: How about `ALT_F1R2`?

LeeTL1220: Yeah ALT_F1R2 and ALT_F2R1, same as before.

davidbenjamin: What if I make it `VCFHeaderLineCount.R` instead of `VCFHeaderLineCount.A` and unify the ref's and alts' annotations under the single name `F1R2`?  This would be consistent with other GATK annotations.

takutosato: Oh yea that sounds good!

davidbenjamin: @LeeTL1220 what do you think?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/OxoGReadCounts.java
```diff
@@ -65,55 +59,37 @@ public void annotate(final ReferenceContext refContext,
...
-        gb.attribute(GATKVCFConstants.OXOG_ALT_F1R2_KEY, alt_F1R2);
-        gb.attribute(GATKVCFConstants.OXOG_ALT_F2R1_KEY, alt_F2R1);
-        gb.attribute(GATKVCFConstants.OXOG_REF_F1R2_KEY, ref_F1R2);
-        gb.attribute(GATKVCFConstants.OXOG_REF_F2R1_KEY, ref_F2R1);
-        gb.attribute(GATKVCFConstants.OXOG_FRACTION_KEY, fraction);
+        final Map<Allele, MutableInt> f2r1Counts = likelihoods.alleles().stream()
+                .filter(Allele::isNonReference)
+                .collect(Collectors.toMap(a -> a, a -> new MutableInt(0)));
+
+        Utils.stream(likelihoods.bestAlleles(g.getSampleName()))
```
takutosato: The code skips ref allele. We decided to include them, right?

src/main/java/org/broadinstitute/hellbender/utils/variant/GATKVCFHeaderLines.java
```diff
@@ -87,11 +86,8 @@ private static void addFilterLine(final VCFFilterHeaderLine line) {
...
         addFormatLine(new VCFFormatHeaderLine(ALLELE_FRACTION_KEY, VCFHeaderLineCount.A, VCFHeaderLineType.Float, "Allele fractions of alternate alleles in the tumor"));
-        addFormatLine(new VCFFormatHeaderLine(OXOG_ALT_F1R2_KEY, 1, VCFHeaderLineType.Integer, "Count of reads in F1R2 pair orientation supporting the alternate allele"));
-        addFormatLine(new VCFFormatHeaderLine(OXOG_ALT_F2R1_KEY, 1, VCFHeaderLineType.Integer, "Count of reads in F2R1 pair orientation supporting the alternate allele"));
-        addFormatLine(new VCFFormatHeaderLine(OXOG_REF_F1R2_KEY, 1, VCFHeaderLineType.Integer, "Count of reads in F1R2 pair orientation supporting the reference allele"));
-        addFormatLine(new VCFFormatHeaderLine(OXOG_REF_F2R1_KEY, 1, VCFHeaderLineType.Integer, "Count of reads in F2R1 pair orientation supporting the reference allele"));
-        addFormatLine(new VCFFormatHeaderLine(OXOG_FRACTION_KEY, 1, VCFHeaderLineType.Float, "Fraction of alt reads indicating OxoG error"));
+        addFormatLine(new VCFFormatHeaderLine(F1R2_KEY, 1, VCFHeaderLineType.Integer, "Count of reads in F1R2 pair orientation supporting the alternate allele"));
```
takutosato: These two lines need to be updated to reflect that the annotation is now an array of ref and alt values.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/OxoGReadCounts.java
davidbenjamin: @takutosato I haven't put the ref allele back in because we're still waiting to hear back from CGA.

takutosato: @davidbenjamin got it

PR #3352, opened 2017-07-24
future proofing our R dependency references

lbergelson:
hopefully this will prevent problems like the one Lee is seeing


lbergelson:
Closing this because it turns out it requires adding an additional dependency on devtools which adds more complication.  It lets us specify versions of our existing dependencies but we end up with an unversioned dependency on devtools instead.  Doesn't really seem worth it since devtools is a fairly heavy thing with lots of it's own dependencies.

PR #3353, opened 2017-07-24
Added functionality for travis tests to upload reports from a docker image 

jamesemery:
Also reduced the overall number of tests to execute by eliminating non-docker unit and integration tests from the travis.yml.

I would suggest checking that the docker tests in this travis execution actually uploaded to where it says they do on gcloud.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,7 @@ jdk:
 - openjdk8
 env:
   matrix:
-  - TEST_TYPE=cloud
-  - TEST_TYPE=integration TEST_VERBOSITY=minimal UPLOAD=true
-  - TEST_TYPE=unit TEST_VERBOSITY=minimal
+  - TEST_TYPE=cloud UPLOAD=true
   - TEST_TYPE=integration TEST_DOCKER=true TEST_VERBOSITY=minimal
   - TEST_TYPE=unit TEST_DOCKER=true TEST_VERBOSITY=minimal
   - RUN_CNV_SOMATIC_WDL=true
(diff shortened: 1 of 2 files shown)


build_docker.sh
```diff
@@ -133,3 +141,6 @@ else
 	echo "Not pushing to dockerhub"
 fi
 
+cd ${ORIGINAL_WORKING_DIRECTORY}
+rm -Rf ${STAGING_DIR}/${STAGING_CLONE_DIR}
```
lbergelson: This line scares me... 

if $STAGING_DIR and $STAGING_CLONE_DIR are unset, you're just doing a `sudo rm -rf /`  which is bad...

Could you add a guard that only runs this if `STAGING_DIR` is set?

.travis.yml
```diff
@@ -9,8 +9,8 @@ jdk:
 env:
   matrix:
   - TEST_TYPE=cloud
-  - TEST_TYPE=integration TEST_VERBOSITY=minimal UPLOAD=true
-  - TEST_TYPE=unit TEST_VERBOSITY=minimal
+#  - TEST_TYPE=integration TEST_VERBOSITY=minimal UPLOAD=true
```
lbergelson: Just delete these lines.

build_docker.sh
jamesemery: Can do, Also it looks like my branches merged together when I did not intend them to. Either way I shall add this.

lbergelson:
@jamesemery  could you kill those two commented out lines and then 👍  to merge

PR #3354, opened 2017-07-24
PathSeq Illumina adapter trimming and simple repeat masking

mwalker174:
Adds additional filtering steps to the PathSeq filter to 1) trim adapter sequences and 2) mimic a simple filter used in RepeatMasker that masks windows with excessive A/T or G/C content.


src/main/java/org/broadinstitute/hellbender/transformers/SimpleRepeatMaskTransformer.java
```diff
@@ -0,0 +1,55 @@
...
+ * Masks read bases with a supra-threshold number of A/T's or G/C's within a given window size.
+ */
+public class SimpleRepeatMaskTransformer implements ReadTransformer {
+    public static final long serialVersionUID = 1L;
+    private static final byte MASKED_BASE_QUAL = 2;
+
+    final int windowSize, threshAT, threshGC;
+
+    public SimpleRepeatMaskTransformer(final int threshAT, final int threshGC, final int windowSize) {
+        this.threshAT = threshAT;
```
TedBrookings: Maybe good to put in an assert that threshAT and threshGC are less than windowSize (if not, it indicates a typo or logic error).

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -14,12 +15,11 @@
 import org.broadinstitute.hellbender.tools.spark.sv.utils.SVUtils;
 import org.broadinstitute.hellbender.tools.spark.utils.ReadFilterSparkifier;
 import org.broadinstitute.hellbender.tools.spark.utils.ReadTransformerSparkifier;
-import org.broadinstitute.hellbender.transformers.BaseQualityClipReadTransformer;
-import org.broadinstitute.hellbender.transformers.BaseQualityReadTransformer;
-import org.broadinstitute.hellbender.transformers.DUSTReadTransformer;
+import org.broadinstitute.hellbender.transformers.*;
```
TedBrookings: I don't feel strongly about this, but maybe better to import the classes individually so that dependencies can more easily be managed.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -213,6 +223,14 @@ private static GATKRead clearReadAlignment(final GATKRead read, final SAMFileHea
 
         if (!filterArgs.skipFilters) {
 
+            //Adapter trimming
+            reads = reads.map(new ReadTransformerSparkifier(new IlluminaAdapterTrimTransformer(filterArgs.maxAdapterMismatches, filterArgs.minAdapterLength, ADAPTER_SEQUENCES)));
+
+            //Apply simple repeat masking
+            //See "Low-complexity DNA and simple repeats" at http://www.repeatmasker.org/webrepeatmaskerhelp.html
+            reads = reads.map(new ReadTransformerSparkifier(new SimpleRepeatMaskTransformer(29, 29, 30)));
```
TedBrookings: If these numbers may need to be tweaked, it would be good to add them to filterArgs. If they're final numbers, it may still be good to name them as constants (e.g. shortWindowLen, maxShortWindowGCCounts or similar, so that it's easier to figure out what they do).

src/main/java/org/broadinstitute/hellbender/transformers/SimpleRepeatMaskTransformer.java
```diff
@@ -0,0 +1,55 @@
...
+        this.threshGC = threshGC;
+        this.windowSize = windowSize;
+    }
+
+    @Override
+    public GATKRead apply(final GATKRead read) {
```
SHuang-Broad: A suggestion for how to implement it. 

I'll start with what my understanding is:
This function, rolls along the provided sequence with a fixed-size window in a base--by-base fashion, calculates the AT/GC (with N) content within the window and mask out the bases and changes the qual accordingly if GC content is high or low.

Now my suggestion:
Since there are many reads to be checked against, I would imagine this could be a hot spot. So making it fast is probably a worthy thing to do:
* instead of summing up base counts for each window, why not store a `previousATNcount` and `previousGCNcount`, considering that each time we step into a new window, these two variables are going to change only +/-1. This would get rid of the `basesCount` array;
* as you have already did in the current implementation, the changes in bases count would be +/-1, why not, instead of setting bases and quals in each window, simply mark if the current window defined by `start` and `end` should be masked, and finally maximally extend the `start` and `end`, then go change those bases and quals
* in `maskWindow`, instead of closed interval, why not make it semi-open [start, end)? this would also get rid of some of the "-1"'s in the `apply()` function.

src/main/java/org/broadinstitute/hellbender/transformers/IlluminaAdapterTrimTransformer.java
```diff
@@ -0,0 +1,56 @@
...
+                for (int j = 0; j < alignmentLength; j++) {
+                    if (!SequenceUtil.isNoCall((byte)adapterSequence.charAt(j)) && bases[alignmentStart + j] != adapterSequence.charAt(j)) {
+                        if (++numMismatches > maxMismatches) break;
+                    }
+                }
+                if (numMismatches <= maxMismatches) {
+                    //We have a match. Clip from the beginning of the adapter to the end of the read.
+                    final String basesString = read.getBasesString();
+                    final String qualString = new String(read.getBaseQualities());
+                    read.setBases(basesString.substring(0, alignmentStart).getBytes());
```
SHuang-Broad: This part confuses me: by "clipping", do you mean set the bases to 'N' and qual to 2? In addition, a read should have a CIGAR, and clipping operations should update the CIGAR (either soft or hard), otherwise the object is in a incoherent state.

src/main/java/org/broadinstitute/hellbender/transformers/IlluminaAdapterTrimTransformer.java
```diff
@@ -0,0 +1,56 @@
...
+            int alignmentLength = minClipLength;
+            for (int alignmentStart = bases.length - minClipLength; alignmentStart >= 0; alignmentStart--) {
+                int numMismatches = 0;
+                //Check each base for mismatches
+                for (int j = 0; j < alignmentLength; j++) {
+                    if (!SequenceUtil.isNoCall((byte)adapterSequence.charAt(j)) && bases[alignmentStart + j] != adapterSequence.charAt(j)) {
+                        if (++numMismatches > maxMismatches) break;
+                    }
+                }
+                if (numMismatches <= maxMismatches) {
```
mwalker174: I think what you are saying is that if some adapter ends with the sequence of another adapter, or if a particular adapter ends with itself e.g.:

1) `AGCAT`
2) `GAGCAT`

Both end in `AGCAT` but if we had the read ending in `GAGCAT` and (2) happens to be tested before (1), it would only trim 4 bases instead of 5. Also if you had an adapter `AAAAA` and minimum clip length of 4, it would only clip 4 bases.

This is a very good point I didn't consider. I don't believe this is the case for any standard adapter sequences (Picard's MarkIlluminaAdapters also doesn't seem to account for it), but since this class exposes the use of custom adapters we should check for it.

src/main/java/org/broadinstitute/hellbender/transformers/SimpleRepeatMaskTransformer.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
mwalker174: This is what IntelliJ does

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
mwalker174: Added these as constants to the PSFilter class.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqFilterSpark.java
```diff
@@ -47,8 +49,9 @@
  *
  */
 @CommandLineProgramProperties(summary = "First, low-quality and repetitive sequences reads are filtered: \n\t(1) Remove secondary and " +
-        "supplementary reads; \n\t(2) mask repetitive sequences using the sDUST algorithm; \n\t(3) hard clip according to masked " +
-        "and low-quality bases; \n\t(4) remove reads below minimum length; \n\t(5) mask low-quality bases; \n\t(6) filter reads with" +
+        "supplementary reads; \n\t(2) trim adapter sequences; \n\t(3) mask sequences with excessive A/T or G/C content; " +
```
mwalker174: Fair point. Done

mwalker174: I think I'm going to leave it as is for now until we get official documentation up, otherwise it's impossible to tell what the parameters do.

src/main/java/org/broadinstitute/hellbender/transformers/IlluminaAdapterTrimTransformer.java
mwalker174: In light of this, I'm also renaming the class to `AdapterTrimTransformer`

src/main/java/org/broadinstitute/hellbender/transformers/IlluminaAdapterTrimTransformer.java
```diff
@@ -0,0 +1,56 @@
...
+import htsjdk.samtools.util.SequenceUtil;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+import java.util.List;
+
+/**
+ * Trims adapter sequences from read ends. If the adapter sequence ends before the end of the read, trims the remaining
+ * bases as well.
+ */
+public class IlluminaAdapterTrimTransformer implements ReadTransformer {
```
mwalker174: Yes, will also make SimpleRepeatMaskTransformer final.

src/main/java/org/broadinstitute/hellbender/transformers/IlluminaAdapterTrimTransformer.java
```diff
@@ -0,0 +1,56 @@
...
+    public static final long serialVersionUID = 1L;
```
mwalker174: Whoops - done.

src/main/java/org/broadinstitute/hellbender/transformers/IlluminaAdapterTrimTransformer.java
mwalker174: Hard clip. Good point about cigars. I found a handy ReadClipper class to use that will take care of this.

src/test/java/org/broadinstitute/hellbender/transformers/SimpleRepeatMaskTransformerTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class SimpleRepeatMaskTransformerTest {
+
+    @DataProvider(name = "testData")
+    public Object[][] getTestData() {
+        return new Object[][]{
+                {"GCGCGCGCGCGCGCGCGCGCGCGCGCGCGCATGCATGCATGCATGC",
```
mwalker174: Not all of them are, but point taken.

src/main/java/org/broadinstitute/hellbender/transformers/SimpleRepeatMaskTransformer.java
```diff
@@ -0,0 +1,55 @@
+package org.broadinstitute.hellbender.transformers;
+
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+/**
+ * Masks read bases with a supra-threshold number of A/T's or G/C's within a given window size.
+ */
+public class SimpleRepeatMaskTransformer implements ReadTransformer {
+    public static final long serialVersionUID = 1L;
```
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/transformers/SimpleRepeatMaskTransformer.java
```diff
@@ -0,0 +1,55 @@
...
+    private static final byte MASKED_BASE_QUAL = 2;
+
+    final int windowSize, threshAT, threshGC;
```
mwalker174: Made private and put each variable on a separate line

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSFilter.java
```diff
@@ -42,6 +42,16 @@
     private JavaRDD<GATKRead> reads;
     private final SAMFileHeader header;
 
+    private static final List<String> ADAPTER_SEQUENCES = CollectionUtil.makeList(
```
mwalker174: I didn't intend for these sequences to always be used, I just happened to use the same sequences in the test as in PSFilter().

SHuang-Broad:
👍

PR #3355, opened 2017-07-24
Option to filter short pathogen reference contigs

mwalker174:
In PathSeqBuildReferenceTaxonomy, adds option to exclude non-virus contigs below a specified length. This helps filter out a lot of false hits to unplaced contigs that were assembled from contaminant (usually human) sequences.


src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -212,20 +213,34 @@ private static void updateAccessionTaxonProperties(final String accession, final
...
+    protected static Map<String, String> buildAccessionToTaxIdMap(final Map<String, PSPathogenReferenceTaxonProperties> taxIdToProperties,
+                                                                  final PSTree tree,
+                                                                  final int minNonVirusContigLength) {
         final Map<String, String> accessionToTaxId = new HashMap<>();
         for (final String taxId : taxIdToProperties.keySet()) {
+            final boolean isVirus = tree.getPathOf(taxId).contains(VIRUS_ID);
             final PSPathogenReferenceTaxonProperties taxonProperties = taxIdToProperties.get(taxId);
-            for (String name : taxonProperties.accessions) {
-                accessionToTaxId.put(name, taxId);
+            for (String name : taxonProperties.getAccessions()) {
```
SHuang-Broad: `final String`

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -212,20 +213,34 @@ private static void updateAccessionTaxonProperties(final String accession, final
         } else {
             taxonProperties = new PSPathogenReferenceTaxonProperties();
```
SHuang-Broad: Several other places in this file could also be handled like this.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSPathogenReferenceTaxonProperties.java
```diff
@@ -11,10 +14,34 @@
     public String name = null;
     public String rank = null;
     public String parentTaxId = null;
-    public long length = 0;
-    public final Set<String> accessions;
+    private long length = 0;
```
SHuang-Broad: Not a comment on the `length` and `accessions` fields, but on the name and rank fields: is it possible to make them immutable  (final). I looked at where they are written, and it seems like they are written only when the default ctor of this class is used and the `name` and `rank` are immediately put after that. What about having a ctor that takes this two arguments?

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -212,20 +213,34 @@ private static void updateAccessionTaxonProperties(final String accession, final
...
     }
 
     /**
+     * Removes nodes not in the tree from the tax_id-to-properties map
+     */
+    static void removeUnusedTaxIds(final Map<String, PSPathogenReferenceTaxonProperties> taxIdToProperties,
+                                   final PSTree tree) {
+        final Set<String> unusedNodes = new HashSet<>(taxIdToProperties.keySet());
+        unusedNodes.removeAll(tree.getNodeIDs());
+        taxIdToProperties.keySet().removeAll(unusedNodes);
```
pshapiro4broad: Isn't this the same as `taxIdToProperties.keySet().retainAll(tree.getNodeIDs());`?

src/test/resources/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBuildReferenceTaxonomy/test.dict
```diff
@@ -13,4 +13,4 @@
 @SQ	SN:ref|ACC_3c	LN:99	M5:b091f1dc6e3c6ef4391fabc5d8149f05	UR:file:/Users/markw/IdeaProjects/gatk/src/test/resources/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqBuildReferenceTaxonomy/test.fasta
```
mwalker174: This is a fair point but it's inconsequential (maybe the UR's are used in production?), plus it seems like all dict files in the repo have the same issue.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
mwalker174: Great. Did this here and a couple of other places where it was easy.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
mwalker174: Yep, thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
```diff
@@ -212,20 +213,34 @@ private static void updateAccessionTaxonProperties(final String accession, final
...
-        taxonProperties.accessions.add(name);
-        taxonProperties.length += length;
+        taxonProperties.addAccession(name, length);
         taxIdToProperties.put(accession, taxonProperties);
     }
 
     /**
+     * Removes nodes not in the tree from the tax_id-to-properties map
+     */
+    static void removeUnusedTaxIds(final Map<String, PSPathogenReferenceTaxonProperties> taxIdToProperties,
```
mwalker174: Test added and using @pshapiro4broad's suggestion below.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSBuildReferenceTaxonomyUtils.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSPathogenReferenceTaxonProperties.java
mwalker174: It appears that way but I sometimes set them directly, i.e. taxonProperties.name = "blah"

mwalker174: Isn't quite possible we usually read in the taxonomic ID long before we get its name, in parseNamesFile(). I added get/set methods for PSPathogenReferenceTaxonProperties.

PR #3358, opened 2017-07-25
Changed the landing directory for the docker image to be /gatk

### LINKED ISSUE #3194: /root in docker image should ideally contain (links to) gatk-launch plus the two jars (local and spark)
Currently the user has to cd into `/gatk` before running `gatk-launch`

jamesemery:
fixes #3194

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@ WORKDIR /root
 # Make sure we can see a help message
 RUN ln -sFv /gatk/build/libs/gatk.jar
 RUN java -jar gatk.jar -h
+RUN ln -s /gatk /root
 
 #Setup test data
 WORKDIR /gatk
@@ -22,5 +23,7 @@ RUN ln -s /testdata src/test/resources
 # Create a simple unit test runner
 ENV CI true
 RUN echo "cd /gatk/ && ./gradlew test" >/root/run_unit_tests.sh
+RUN cp -r . /gatk
+RUN echo "./gradlew test" >/gatk/run_unit_tests.sh
 
-WORKDIR /root
+WORKDIR /gatk
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,11 +19,13 @@ WORKDIR /gatk
 RUN rm -rf src/test/resources
 # Create link to where test data is expeced
 RUN ln -s /testdata src/test/resources
-RUN cp -r . /gatk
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -11,6 +11,7 @@ WORKDIR /root
 # Make sure we can see a help message
 RUN ln -sFv /gatk/build/libs/gatk.jar
 RUN java -jar gatk.jar -h
+RUN ln -s /gatk /root
```
droazen: This should be removed

Dockerfile
```diff
@@ -22,5 +23,9 @@ RUN ln -s /testdata src/test/resources
 # Create a simple unit test runner
 ENV CI true
 RUN echo "cd /gatk/ && ./gradlew test" >/root/run_unit_tests.sh
+RUN echo "./gradlew test" >/gatk/run_unit_tests.sh
```
droazen: This should also be removed -- this will just get overwritten by the cp command later

Dockerfile
```diff
@@ -22,5 +23,9 @@ RUN ln -s /testdata src/test/resources
...
 
 WORKDIR /root
+RUN cp -r . /gatk
```
droazen: Copy selected items from `/root` into `/gatk` rather than everything, for safety reasons.

README.md
```diff
@@ -77,6 +77,8 @@ You can download and run pre-built versions of GATK4 from the following places:
 
 * Starting with the beta release, you can download a GATK4 docker image from [our dockerhub repository](https://hub.docker.com/r/broadinstitute/gatk/)
 
+* A docker image for the current release can be found in the broadinstitute/gatk repository on dockerhub. Within the image simply run ./gatk-launch from the default directory.
```
droazen: Comment is redundant with the preceding bullet point. Add a sub-bullet-point below that one instead that just says "Within the docker image, run gatk-launch commands as usual from the default startup directory (/gatk)"

vdauwera:
Will this have an effect on how we call gatk when using the docker via WDL?

jamesemery:
@vdauwera It shouldn't. It looks like production uses a different docker build script. Only the mutect2 and cnv wdls should have been affected and the tests for each of those tools still worked.

vdauwera:
great, thanks @jamesemery

PR #3359, opened 2017-07-25
Extends BwaMemImageSingleton into a cache, BwaMemImageCache, that can…

vruano:
… manage more than one index image


src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexCache.java
```diff
@@ -0,0 +1,51 @@
...
+    public static synchronized BwaMemIndex getInstance( final String indexImageFile ) {
+        Utils.nonNull(indexImageFile, "the index image file name provided cannot be null");
+        if (!instances.containsKey(indexImageFile)) {
+            instances.put(indexImageFile, new BwaMemIndex(indexImageFile));
+        }
+        return instances.get(indexImageFile);
+    }
+
+    /**
+     * Closes all instances in the VM.
```
tedsharpe: I don't know what your plans are for using multiple indices, but it's certainly a reasonable use case that you might want to close some one particular instance, rather than being forced to close them all.  Do we need another method that closes a particular instance?

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexCache.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public static synchronized void closeInstances() {
+        instances.values().forEach(BwaMemIndex::close);
+        instances.clear();
```
tedsharpe: It strikes me as a little dangerous to close each one, and only then to clear the Map.  I think a more recoverable way to do this would be to iterate over the map, doing the close and then an iterator.remove().  That way, all successfully closed indexes will have been removed from the map.

vruano: Yeah, you are right, most likely that would result in an unrecoverable exception but as a matter of principle is better to follow your approach.

tedsharpe: There is a conceivably recoverable scenario:  we keep a positive reference count on indexes that are currently in use for alignment, and throw an IllegalStateException if we attempt to close an index that's in use.  So later in the program when the index is truly unused you could try again and it would close just fine.  It's also true that no one should be attempting to close an index that may or may not be in use, but I suppose it could happen.

vruano: I guess, if we take the extra mile one can use a References (e.g. SoftReference) and just rely on the GC to dispose unused indexes when there is lack of mem. Would that be better?

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexCache.java
vruano: I've already encountered this when I have to realign read pairs against several haplotypes, each is a separate "reference".
Ok will add individual index close operations.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemIndexCache.java
tedsharpe: No, no, no!  Don't do it!
A typical reference index claims a gigantic amount of memory -- and that memory is largely outside of the JVM heap.  You need a definite way of saying, "I'm done with this little teeny Java object now.  Please recover the gigantic amount of non-Java memory I control right now."

vruano: Didn't plan to do it in this PL anyway. No SoftReferences, nor counters for now.

vruano:
@tedsharpe please review changes to address your comments.

tedsharpe:
Looks fine.

PR #3363, opened 2017-07-27
Documentation for local assembly in the GATK

davidbenjamin:
@vdauwera @ldgauthier @vruano This describes the assembly engine pretty much start to finish. 
 It is part of the M2 team's goal of complete, self-contained documentation for all of M2, but it applies equally well to HaplotypeCaller and seems like it could be of general interest to the whole group.  (I have been told that the old HC manuscript was short on technical details and I *hope* I'm not duplicating earlier efforts.  If I am, it will at least have been a useful exercise for my own education).

Anyone care to review or suggest a reviewer?


docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+\date{\today}
+
+\begin{abstract}
+The GATK tool \HC~ and \Mutect~ assemble reads aligned within a window of several hundred base pairs into a de Bruijn graph of local variation.  Local haplotypes correspond to paths in this de Bruijn graph, and the downstream likelihood calculations in both tools involve aligning reads to these haplotypes.  Here we describe how this de Bruijn graph is generated.
+\end{abstract}
+
+\maketitle
+
+\section{Correcting Reads} \label{correcting-reads}
+Read error correction is turned off by default in \HC~ and \Mutect~ and we are not familiar with it.  The idea is as follows: large kmers are good because they contain more phasing information and therefore yield a simpler de Bruijn graph.  For example, two SNVs 20 bases apart with sequenced with error-free reads yield a de Bruijn graph with $2 \times 2 = 4$ paths if $k = 10$ but only $2$ paths if $k = 40$, because the latter spans the phased SNVs.  The drawback is that larger kmers are more liable to be lost due to sequencing error simply because they contain more bases.  The idea of read error correction is to rescue kmers from the occasional error.
```
ronlevine: Why 2*2 paths? Why 2 paths? Please explicitly say where the values come from.

davidbenjamin: @ronlevine Thanks so much for looking at this! (Not that I'm conscripting you to review, but I appreciate any participation).  Long kmers yield two paths because there is a bubble one side of which is the reference and one side of which contains the two phased variants (or one side contains one variant if they are anti-phased).  With short kmers you get a bubble for the first variant, followed by a stretch of ref-only, followed by a second bubble.  I can clarify this in the docs.  Or maybe even make a figure.

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+First, the \code{ReadErrorCorrector} kmerizes every read and counts the occurrences of each kmer.  Then it builds a map of kmer corrections where kmers that appear often\footnote{By default, 20 times or more.  This threshold is set by the \code{minObservationsForKmerToBeSolid} command line argument.} map to themselves and kmers that appear only once\footnote{This is a hardcoded threshold.} map to their nearest neighbor in Hamming distance\footnote{For simplicity we consider only substitution errors, eg the distance between ACGGT and AGGTG is 3} within a maximum of two mismatches\footnote{This is a hardcoded threshold.}.  Kmers that appear between 1 and 20 times are not part of the correction map.
```
ronlevine: Should `map of kmer corrections` be `map of kmer to kmer corrections`?

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+Then, for every base in every read, the \code{ReadErrorCorrector} queries the kmer correction map for each overlapping kmer.  For example, to correct the fourth base of read ACGTATTC if k = 3, it looks at the corrections for CGT, GTA, and TAT at their thirds, second, and first positions, respectively.  If and only if the corrections are unanimous, the base is corrected.  Note that corrected reads are used only for assembly and the original reads are used downstream.
+
+\section{Building the graph} \label{graph-assembly}
+Next, the \code{ReadThreadingAssembler} assembles the (corrected) reads over several different kmer sizes specified by the \code{kmerSize} argument\footnote{By default, 10 and 25.  This size is a compromise and no single value is the best choice for all regions.  Large kmers are more likely to be unique and to yield a graph with no cycles, which is especially important in low-complexity regions, but they are more sensitive to errors and low coverage.}.  If the given kmer sizes fail to produce a graph without cycles, the \code{ReadThreadingAssembler} repeatedly increases $k$ by 10 bases until assembly succeeds.
```
ronlevine: The graph doesn't always succeed without cycles because `MAX_KMER_ITERATIONS_TO_ATTEMPT = 6` in `ReadThreadingAssembler` So,  `repeatedly increases k by 10 bases until assembly succeeds until assembly` should be something like `repeatedly increases k by 10 bases until assembly succeeds until assembly succeeds or k has been increased by 60 bases`.  The attempts at increasing the kmer size can be overridden by `-dontIncreaseKmerSizesForCycles`.

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+First, the \code{ReadErrorCorrector} kmerizes every read and counts the occurrences of each kmer.  Then it builds a map of kmer corrections where kmers that appear often\footnote{By default, 20 times or more.  This threshold is set by the \code{minObservationsForKmerToBeSolid} command line argument.} map to themselves and kmers that appear only once\footnote{This is a hardcoded threshold.} map to their nearest neighbor in Hamming distance\footnote{For simplicity we consider only substitution errors, eg the distance between ACGGT and AGGTG is 3} within a maximum of two mismatches\footnote{This is a hardcoded threshold.}.  Kmers that appear between 1 and 20 times are not part of the correction map.
+
+Then, for every base in every read, the \code{ReadErrorCorrector} queries the kmer correction map for each overlapping kmer.  For example, to correct the fourth base of read ACGTATTC if k = 3, it looks at the corrections for CGT, GTA, and TAT at their thirds, second, and first positions, respectively.  If and only if the corrections are unanimous, the base is corrected.  Note that corrected reads are used only for assembly and the original reads are used downstream.
```
ronlevine: typo: thirds -> third

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+\section{Building the graph} \label{graph-assembly}
+Next, the \code{ReadThreadingAssembler} assembles the (corrected) reads over several different kmer sizes specified by the \code{kmerSize} argument\footnote{By default, 10 and 25.  This size is a compromise and no single value is the best choice for all regions.  Large kmers are more likely to be unique and to yield a graph with no cycles, which is especially important in low-complexity regions, but they are more sensitive to errors and low coverage.}.  If the given kmer sizes fail to produce a graph without cycles, the \code{ReadThreadingAssembler} repeatedly increases $k$ by 10 bases until assembly succeeds.
+
+The first step is to create a set of sequences to be kmerized and put into the de Bruijn graph.  These sequences are: the reference haplotype, any requested alleles in \HC's GGA mode, and maximal subsequences of reads with base quality at least 10 and at least one kmer long.  That is, a read with 100 bases and a base of quality 7 at position 70 yields sequences of length 69 and 30 if $k = 10$ and a single sequence of length 69 if $k = 40$.
```
ronlevine: `base quality at least 10` can be changed to `base quality at least 10 by default` because the default value of `--min_base_quality_score` is `10`.

ronlevine: What is `GGA mode`? I assume it's referring to genotyping. Is this referring to requesting alleles with the `--alleles` argument?

ronlevine: I think changing `69 and 30` to `69, 30` would be clearer and read better.

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+\item dangling tails: The assembler only outputs haplotypes that start and end with a reference kmer, so it attempts to rescue paths in the graph that do not.  To rescue a ``dangling tails" -- a path that ends in a non-reference kmer vertex -- the assembler first traverses the graph backwards from this vertex to a reference vertex.  If during traversal it encounters a vertex with more than one incoming edge it gives up\footnote{as opposed to doing eg depth-first search of all possible paths back to the reference.}  It also gives up if it encounters a vertex with more than one outgoing edge, that is, if the path branches again after diverging from the reference\footnote{It seems like this could be changed to increase sensitivity.}.  Then it generates the Smith-Waterman alignment of the branching path versus the reference path after the vertex at which they diverge.  If the alignment's CIGAR contains three or fewer elements, that is, if the alignment has at most one indel,
+
```
ronlevine: `,` should be `.`.

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+To merge the dangling tail back into the reference path, the assembler finds the beginning of the maximal common suffix of the dangling path and the reference path, that is, the point at which the sequences coverges\footnote{this is \textit{not} where the \textit{paths in the graph} converge (they don't) because kmers in the suffix disagree with the ref at upstream bases.} and adds an edge between the dangling path's vertex and the reference path's vertex at this position.  Note that this means that the graph is no longer a valid de Bruijn graph because the dangling vertex kmer and its succeeding reference vertex kmer do not overlap by $k - 1$ bases.  Nonetheless, this graph yields valid haplotypes when we later ``zip'' the graph by accumulating the last base of each kmer.
```
ronlevine: I think `Note that` in not necessary.

ronlevine: What do you mean by `zip`?

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+\item dangling heads: This is the mirror image of dangling tails.
```
ronlevine: Are they handled the same way as dangling heads?

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+\item non-reference paths: after attempting to merge dangling heads and tails into the reference path, the assembler deletes edges and vertices belonging to dead ends i.e. non-branching subgraphs that start at non-reference source vertices or end at non-reference sink vertices.  This is achieved by performing a breadth-first search of vertices moving forward from the reference source and a breadth-first search of vertices moving backwards from the reference sink and keeping only vertices found in both searches.
```
ronlevine: `after` -> `After`

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+\item zipping chains: a de Bruijn graph of kmers is convenient for performing assembly but is inefficient for summarizing the results of assembly.  In this step the assembler converts the de Bruijn graph\footnote{or rather, if dangling paths have been merged, an \textit{almost} de Bruijn graph.} into a sequence graph by combining all vertices in each maximal non-branching subgraph (i.e. each linear chain) into a single vertex containing the sequence implied by those vertices' kmers\footnote{That is, the concatenation of the last bases of all kmers, except for the first vertex which contributes its entire kmer if and only if it is a source.}.  Thus, for example, a de Bruijn graph containing one reference path and one bubble for a single variant yields a sequence graph with four vertices: the reference subgraphs before and after the bubble and the two sides of the bubble.
+
```
ronlevine: Capitalize: `a de Bruijn graph` -> `A de Bruijn graph`

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+\item the clean-up steps of zipping chains, merging diamonds, tails, and common suffixes, and splitting common prefixes and suffixes are repeated until no more transformations occur.
+
+\end{itemize}
+
+\section{Finding Haplotypes} \label{finding-haplotypes}
+At this point there is one cleaned sequence graph for each successful kmer size.  For each graph we find the best haplotypes\footnote{This is 128 by default and set by the \code{maxNumHaplotypesInPopulation} argument.} according the following score: the score of a path (haplotype) in a sequence graph is the sum over all branching vertices in the path of the log of the multiplicity of the outgoing edge in the path minus the log of the total multiplicity of all outgoing edges.  This description completely defines the best haplotypes; the actual implementation is a somewhat complicated recursive algorithm with lots of polymorphism.
```
ronlevine: I think the recursive algorithm should be explained here at a high level.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/readthreading/ReadThreadingGraph.java
```diff
@@ -788,6 +788,17 @@ final DanglingChainMergeHelper generateCigarAgainstUpwardsReferencePath(final Mu
         return findPath(vertex, pruneFactor, v -> isReferenceNode(v) || outDegreeOf(v) != 1, v -> isReferenceNode(v), v -> outgoingEdgeOf(v), e -> getEdgeTarget(e));
     }
 
+    /**
+     * Finds a path starting from a given vertex and satisfying various predicates
+     *
+     * @param vertex   the original vertex
+     * @param pruneFactor  the prune factor to use in ignoring chain pieces
+     * @param done predicate on vertices testing whether the end of the path has been reached
```
ronlevine: Could this be something like: test for whether a vertex is at the end of the path?

ldgauthier:
@davidbenjamin I will take a look, probably tomorrow.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/readthreading/ReadThreadingGraph.java
```diff
@@ -1128,8 +1139,8 @@ private MultiDeBruijnVertex extendChainByOne(final MultiDeBruijnVertex prevVerte
     }
 
     /**
-     * Add the given read to the sequence graph.  Ultimately the read will get sent through addSequence(), but first
-     * this method ensures we only use high quality bases and accounts for reduced reads, etc.
+     * Add a read to the sequence graph.  Finds maximal consecutive runs of bases with sufficient quality
+     * and applies addSequence() to these subreads if they are longer than the kmer size.
```
ronlevine: Can use `@see` for `addSequence()`.

sooheelee:
It would be nice to have illustrative examples or diagrams.

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+\date{\today}
+
+\begin{abstract}
+The GATK tool \HC~ and \Mutect~ assemble reads aligned within a window of several hundred base pairs into a de Bruijn graph of local variation.  Local haplotypes correspond to paths in this de Bruijn graph, and the downstream likelihood calculations in both tools involve aligning reads to these haplotypes.  Here we describe how this de Bruijn graph is generated.
+\end{abstract}
+
+\maketitle
+
+\section{Correcting Reads} \label{correcting-reads}
```
sooheelee: Error correction is news to me. So you are saying we don't actually use it?

sooheelee:
I've just finished reading through this white paper this morning. One thing the users ask about, that this paper omits, is how the reads are filtered by the engine and then by HaplotypeCaller. What reads are actually used in building the graph etc.

docs/local_assembly.tex
davidbenjamin: Clarified and added figure.

docs/local_assembly.tex
davidbenjamin: done

docs/local_assembly.tex
davidbenjamin: done

docs/local_assembly.tex
davidbenjamin: fixed

docs/local_assembly.tex
davidbenjamin: > base quality at least 10 can be changed to base quality at least 10 by default because the default value of --min_base_quality_score is 10.

Done.

> 69, 30 would be clearer and read better.

Done

> Is this referring to requesting alleles with the --alleles argument?

Yes, and clarified.

docs/local_assembly.tex
davidbenjamin: fixed

docs/local_assembly.tex
davidbenjamin: done and added figure

docs/local_assembly.tex
davidbenjamin: yes, clarified

docs/local_assembly.tex
davidbenjamin: done x 5

docs/local_assembly.tex
davidbenjamin: done

docs/local_assembly.tex
davidbenjamin: My off-hand comment was intended as a cop-out to explain my omission, not to suggest that the complication causes problems.  But it's a moot point because it seems like I really should get the job done and explain it here.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/readthreading/ReadThreadingGraph.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/readthreading/ReadThreadingGraph.java
davidbenjamin: done

docs/local_assembly.tex
davidbenjamin: > I don't think we ever use error correction.... does it even work?

Unknown.

> Has Palantir evaluated it?

No.

> Error correction is news to me. So you are saying we don't actually use it?

We do not.  I document it because it might enable larger kmers.

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+\title{Local Assembly in HaplotypeCaller and Mutect}
+\author{David Benjamin\footnote{The author took no part in development of the methods described below -- credit belongs to several others on the GATK team. }}
+\email{davidben@broadinstitute.org}
+\affiliation{Broad Institute, 75 Ames Street, Cambridge, MA 02142}
+
+\date{\today}
+
+\begin{abstract}
+The GATK tool \HC~ and \Mutect~ assemble reads aligned within a window of several hundred base pairs into a de Bruijn graph of local variation.  Local haplotypes correspond to paths in this de Bruijn graph, and the downstream likelihood calculations in both tools involve aligning reads to these haplotypes.  Here we describe how this de Bruijn graph is generated.
```
davidbenjamin: @vruano I get that once we merge dangling tails and zip chains it's not a de Bruijn graph, but before then why isn't it a de Bruijn graph?

docs/local_assembly.tex
davidbenjamin: Done

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+Before building the graph, the \code{ReadThreadingAssembler} finds the set of all kmers that appear twice in the same sequence i.e. twice in the reference or twice in any read.  Then each sequence is kmerized starting from the first kmer that is not in this set.  When a new kmer appears, a vertex is added to the graph; otherwise an edge is added (or the multiplicity of an existing edge incremented) from the preceding kmer to the current one.  Up to this point, we have produced a de Bruijn graph in textbook fashion, with the exception that we have excluded kmers containing low-quality bases and non-unique kmers at the beginning of reads.  Note that this is a directed graph of kmers from aligned reads (i.e. aligned to the forward strand of the reference), not a bidirected graph of unaligned reads that could come from either strand, with the complications of associating a kmer with its reverse complement.
```
davidbenjamin: done

docs/local_assembly.tex
```diff
@@ -0,0 +1,77 @@
...
+
+
+\section{Cleaning the Graph} \label{graph-cleaning}
+Before deciding on candidate haplotypes, the assembler simplifies the graph with the following heuristics to remove spurious paths and to merge variant paths that diverge from the reference.
+\begin{itemize}
+\item pruning: the assembler finds all maximal non-branching subgraphs and removes those that 1) do not share an edge with the reference path and 2) contain no edges with sufficient multiplicity\footnote{By default 2.  This is controlled by the \code{minPruning} argument.}  While the default multiplicity threshold of 2 is quite permissive, it \textit{does} cause \Mutect~ to lose sensitivity for deletions occurring in a single read\footnote{While a SNV occurring on a single read would never be called due to its low likelihood, a long deletion in a non-STR context could easily be supported by a single read.}.
```
davidbenjamin: clarified it

davidbenjamin:
@sooheelee I added a few figures and will add more.  I agree that read filtering needs to be documented but not in the assembly docs.  We'll get to it eventually.

docs/local_assembly.tex
davidbenjamin: Done, elaborated a bit on the algorithm.

davidbenjamin:
@ronlevine @vruano Thanks for the suggestions.  I think I addressed everything, except I need to understand why it's not a de Bruijn graph even before cleaning.

docs/local_assembly.tex
ldgauthier: If you believe wikipedia(https://en.wikipedia.org/wiki/De_Bruijn_graph), a de Bruijn graph is required to represent all possible sequences: ```In graph theory, an n-dimensional De Bruijn graph of m symbols is a directed graph representing overlaps between sequences of symbols. It has m^n vertices, consisting of all possible length-n sequences of the given symbols```

davidbenjamin: But who in comp bio uses *that* definition?

docs/local_assembly.tex
```diff
@@ -0,0 +1,120 @@
...
+
+\title{Local Assembly in HaplotypeCaller and Mutect}
+\author{David Benjamin\footnote{The author took no part in development of the methods described below -- credit belongs to several others on the GATK team. }}
+\email{davidben@broadinstitute.org}
+\affiliation{Broad Institute, 75 Ames Street, Cambridge, MA 02142}
+
+\date{\today}
+
+\begin{abstract}
+The GATK tool \HC~ and \Mutect~ assemble reads aligned within a window of several hundred base pairs into a de Bruijn graph of local variation.  Local haplotypes correspond to paths in this de Bruijn graph, and the downstream likelihood calculations in both tools involve aligning reads to these haplotypes.  Here we describe how this de Bruijn graph is generated.
```
ldgauthier: "GATK tool HC and" -> "GATK tools HaplotypeCaller and"

docs/local_assembly.tex
```diff
@@ -0,0 +1,120 @@
...
+
+The first step is to create a set of sequences to be kmerized and put into the de Bruijn graph.  These sequences are: the reference haplotype, any requested alleles given by the \code{alleles} argument in \HC's ``genotype given alleles" (GGA) mode, and maximal subsequences of reads with base quality at least 10 by default\footnote{This is controlled by the \code{minBaseQualityScore} argument.} and at least one kmer long.  That is, a read with 100 bases and a base of quality 7 at position 70 yields sequences of length 69, 30 if $k = 10$ and a single sequence of length 69 if $k = 40$.
```
ldgauthier: I don't know if it's worth mentioning GGA mode because we don't support it in conjunction with reference-confidence mode, which is our recommended best practices.

docs/local_assembly.tex
```diff
@@ -0,0 +1,120 @@
...
+
+Before building the graph, the \code{ReadThreadingAssembler} finds the set of all kmers that appear twice in the same sequence i.e. twice in the reference or twice in any read.  Then each sequence is kmerized starting from the first kmer that is not in this set.  When a new kmer appears, a vertex is added to the graph; otherwise an edge is added (or the multiplicity of an existing edge incremented) from the preceding kmer to the current one.  Up to this point, we have produced a de Bruijn graph in textbook fashion, with the exception that we have excluded kmers containing low-quality bases and non-unique kmers at the beginning of reads.  Note that this is a directed graph of kmers from aligned reads (i.e. aligned to the forward strand of the reference), not a bidirected graph of unaligned reads that could come from either strand, with the complications of associating a kmer with its reverse complement.
```
ldgauthier: It might be worth explicitly mentioning that we don't use mate information anywhere.  That's an obvious direction for improvement.

docs/local_assembly.tex
```diff
@@ -0,0 +1,120 @@
...
+
+To merge the dangling tail back into the reference path, the assembler finds the beginning of the maximal common suffix of the dangling path and the reference path, that is, the point at which the sequences coverges\footnote{this is \textit{not} where the \textit{paths in the graph} converge (they don't) because kmers in the suffix disagree with the ref at upstream bases.} and adds an edge between the dangling path's vertex and the reference path's vertex at this position.  This means that the graph is no longer a valid de Bruijn graph because the dangling vertex kmer and its succeeding reference vertex kmer do not overlap by $k - 1$ bases.  Nonetheless, this graph yields valid haplotypes when we later ``zip'' the graph's chains (see below) by accumulating the last base of each kmer.
+
+Figure \ref{fig:dangling} shows the result of the tail from alt path ACCTGGA(T$\rightarrow$C)CC being merged back into the reference path.
```
ldgauthier: It would be good to indicate which path is the reference in a different color.

docs/local_assembly.tex
```diff
@@ -0,0 +1,120 @@
...
+
+\item zipping chains: A de Bruijn graph of kmers is convenient for performing assembly but is inefficient for summarizing the results of assembly.  In this step the assembler converts the de Bruijn graph\footnote{or rather, if dangling paths have been merged, an \textit{almost} de Bruijn graph.} into a sequence graph by combining all vertices in each maximal non-branching subgraph (i.e. each linear chain) into a single vertex containing the sequence implied by those vertices' kmers\footnote{That is, the concatenation of the last bases of all kmers, except for the first vertex which contributes its entire kmer if and only if it is a source.}.  Thus, for example, a de Bruijn graph containing one reference path and one bubble for a single variant, such as the graph of Figure \ref{fig:phased} yields a sequence graph with four vertices: the reference subgraphs before and after the bubble and the two sides of the bubble, as in Figure \ref{fig:zipped}.
```
ldgauthier: You allude to the fact that this has implications for phasing in the title of the figure, but it would be good to be explicit about the kmer size tradeoffs here.

docs/local_assembly.tex
```diff
@@ -0,0 +1,120 @@
...
+\begin{figure}
+\center
+\includegraphics[scale=0.5]{diamond.png}
+\caption{A merged diamond.}
+\label{fig:diamond}
+\end{figure}
+
+\item splitting common prefixes and suffixes: If there are multiple nodes $S_i$ with a common predecessor and successor of the form $S_i = A + x_i + B$, where $A$ is a common prefix, $B$ is a common suffix, and $x_i$ is unique to $S_i$, the assembler splits each $S_i$ such that $A$ and $B$ become independent nodes, with the $x_i$ in between.
+
+\item The clean-up steps of zipping chains, merging diamonds, tails, and common suffixes, and splitting common prefixes and suffixes are repeated until no more transformations occur.
```
ldgauthier: Do you want to mention that you can output the various stages of the graphs with some args?

docs/local_assembly.tex
davidbenjamin: I'm comfortable calling it a de Bruijn graph even if it doesn't follow the wikipedia definition as long as it follows the usage in Pevzner and basically every paper on assembly since then.  Does our graph violate *their* definition?

docs/local_assembly.tex
davidbenjamin: fixed

docs/local_assembly.tex
davidbenjamin: I put in a big ol' caveat footnote.

docs/local_assembly.tex
davidbenjamin: Done.

docs/local_assembly.tex
davidbenjamin: done

docs/local_assembly.tex
davidbenjamin: I put some more discussion about the drawbacks of large k above.

docs/local_assembly.tex
davidbenjamin: done

davidbenjamin:
@ldgauthier @vruano back to you.

docs/local_assembly.tex
ldgauthier: What is their definition?  Or is the issue that they don't give one?

davidbenjamin: From Pevzner's *Nature* review:

>Modern short-read assembly algorithms construct a de Bruijn graph by representing all k-mer prefixes and suffixes as nodes and then drawing edges that represent k-mers having a particular prefix and suffix. For example, the k-mer edge ATG has prefix AT and suffix TG. 

This is the definition that I'm trying to use here.

davidbenjamin:
@ldgauthier @vruano @ronlevine If you accept the Pevzner definition of a de Bruijn graph that I cited above, I think I addressed everything.  Anything else?

docs/local_assembly.tex
```diff
@@ -0,0 +1,121 @@
...
+\end{abstract}
+
+\maketitle
+
+\section{Correcting Reads} \label{correcting-reads}
+Read error correction is turned off by default in \HC~ and \Mutect~ and we are not familiar with it, nor have we validated it.  Nonetheless, the code exists and can be turned on.  The idea is as follows: large kmers are good because they contain more phasing information and therefore yield a simpler de Bruijn graph.  For example, two SNVs 20 bases apart with sequenced with error-free reads yield a de Bruijn graph with $2 \times 2 = 4$ paths if $k = 10$ but only $2$ paths if $k = 40$, because the latter spans the phased SNVs.  For example, consider a reference sequence TGAAACGTATTTGGG and an alt sequence with two phased SNVs, TGAAA(C$\rightarrow$T)GTA(T$\rightarrow$C)TTGGG.  If $k=3$ no kmer spans both SNVs and so the assembly graph containing two bubbles, one for each SNV, as in Figure \ref{fig:unphased}.  If $k = 5$, a kmer spans both SNVs and thus only two paths exist in the assembly graph, as in Figure \ref{fig:phased}.
```
ronlevine: Should the read error correction discussion be a footnote and the sentence `Nonetheless, the code exists and can be turned on.` be removed? It seems out of place.

davidbenjamin: One of my goals here is to create a guide for people (me, for example) who may want to hack local assembly in the GATK.  I realize that it reads poorly to start with a feature that nobody uses, but as long as I'm very clear that it's not generally used I would rather not bury it in a footnote.  It's potentially a useful feature.

ronlevine: Understood. Perhaps reword it so it flows better into the the kmer discussion. If I understand this correctly, perhaps say that the reads used for generating the kmers are not error corrected by default, but can be turned on with the --pcr_indel_model option.

ldgauthier:
LGTM :+1:

davidbenjamin:
@vruano Waiting for your approval and/or condemnation.

docs/local_assembly.tex
```diff
@@ -0,0 +1,121 @@
...
+
+Before building the graph, the \code{ReadThreadingAssembler} finds the set of all kmers that appear twice in the same sequence i.e. twice in the reference or twice in any read.  Then each sequence is kmerized starting from the first kmer that is not in this set.  When a new kmer appears, a vertex is added to the graph; otherwise an edge is added (or the multiplicity of an existing edge incremented) from the preceding kmer to the current one.  Up to this point, we have produced a de Bruijn graph in textbook fashion, with the exception that we have excluded kmers containing low-quality bases and non-unique kmers at the beginning of reads.  Note that this is a directed graph of kmers from aligned reads (i.e. aligned to the forward strand of the reference), not a bidirected graph of unaligned reads that could come from either strand, with the complications of associating a kmer with its reverse complement.
```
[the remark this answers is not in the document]
vruano: There are a few innacuracies in this paragraph for what I remember, I'm going to addreses those here without confirming looking at the code so there is a risk that I'm wrong and you are right...

> "Before building the graph, the ReadThreadingAssembler funds the set of all kners that appear twice in the same sequence i.e. ref or read". 

I think that you mean to say twice or more, so not once. Also you don't say what those kmers are (the **non-unique kmer set**)... just that you won't use them as the start "threading the sequence"... I think it is important to explicitly state the concepts of unique and non-unique kmers as they have a important influence as how the rest of the algorithm progresses. 

The very first sequence that we thread is the reference sequence. In that special case we add all the kmers to the graph (I believe) unique or non-unique.
I'm not sure what is done with the given alleles in GGA mode... I think that if GGA is on we don't even bother to thread the reads at all but just to add the small bubbles to account for the alleles as they appear on the input VCF. Perhaps it actually composes the alternative haplotypes and does the threading as is done with the ref or the reads (describe bellow) but that would be inefficient if you have many variants in the active region ... why would you thread all the possible haplotypes when you can deal with each variant position individually?

Now come the reads... these are threaded starting a the first kmer that in unique (so is important to define what that actually means) and that is present already in the graph. So you may have a early kmer that is unique but because is not part of the reference sequence is not use as start.  I don't know the reason for this but I guess that authors (M. DePristo, R Poplin) wanted to have a connected graph for the very start. Anyway this process may result in **dangling tail** sequence that leave the reference path and never come back either because a suffix of the read that is not present in the reference sequence or it contains diverging non unique-kmers.

I think that after this is done the prefix of the read before such  a start kmer is thread upstream potentially resulting into a **dangling head**. 

Don't know what would happen with those sequence that don't have  any unique kmer that is not present in the reference.... or no unique kmer at all... are they simply discarded? possibly considering that this is typically unlikely expect for a read-long insertion and HC is not considered a good tool for those.

Also when we thread a read that has diverged from the reference and the next kmer is non-unique it will create a new node for that kmer even it is already present in the graph unless the it is already present as an extension of the previous kmer node.

docs/local_assembly.tex
```diff
@@ -0,0 +1,121 @@
...
+
+
+\section{Cleaning the Graph} \label{graph-cleaning}
+Before deciding on candidate haplotypes, the assembler simplifies the graph with the following heuristics to remove spurious paths and to merge variant paths that diverge from the reference.
+\begin{itemize}
+\item pruning: The assembler finds all maximal non-branching subgraphs and removes those that 1) do not share an edge with the reference path and 2) contain no edges with sufficient multiplicity\footnote{By default 2.  This is controlled by the \code{minPruning} argument.}  While the default multiplicity threshold of 2 is quite permissive, it \textit{does} cause \Mutect~ to lose sensitivity for deletions occurring in a single read\footnote{While a SNV occurring on a single read would not yield a confident somatic variant call, a long deletion in a non-STR context could easily be supported by a single read be due to the tiny probability of its arising from sequencing error.}.
```
vruano: "no longer a valid de Bruijn"... it never was a valid de-Bruijn graph I think.

docs/local_assembly.tex
```diff
@@ -0,0 +1,121 @@
...
+
+\end{itemize}
+
+When the \code{debugGraphTransformations} argument is used, the assembly engine emits graphviz .dot files for each of the steps above.
+
+\section{Finding Haplotypes} \label{finding-haplotypes}
+At this point there is one cleaned sequence graph for each successful kmer size.  For each graph we find the best haplotypes\footnote{This is 128 by default and set by the \code{maxNumHaplotypesInPopulation} argument.} according the following score: the score of a path (haplotype) in a sequence graph is the sum over all branching vertices in the path of the log of the multiplicity of the outgoing edge in the path minus the log of the total multiplicity of all outgoing edges.  This formula is implemented as a recursive depth-first search: Starting from the reference source vertex, the haplotype finder adds branching scores and instantiates new haplotype finders for each edge whenever a path diverges and otherwise simply moves to the succeeding vertex without changing the accumulated score.  The procedure ends when every haplotype finder reaches a sink vertex, that is, when the exhaustive search of all paths is finished.
```
vruano: It uses dynamic programing where the reused subproblems are the haplotype suffixes.... I'm not sure that depth-first is an accurate descriptor... it seems that you are describing the first algorithm that was use (leaving their also different scoring system aside). 

"The procedure ends when every haplotype finder reaches a sink vertex, that is, when the exhaustive search of all paths is finished" That seem measleading... It never explicitly enumerates all possible haplotypes... in some extraordinary cases these could be more than Intger.MAX_VALUE haplotipes. 
The algoritm figures out the best K haplotypes buy reusing subproblem computation. 

The current algorithm has a few short comings... First the code is more complex that it needs to be ... then there is an alternative algorithm base of dijstra that can produce the best K-paths just as efficiently. 

Also is has been shown that taking the K best haplotypes won't necessarely recover all plaussible variation; so this could also be improved.

docs/local_assembly.tex
```diff
@@ -0,0 +1,121 @@
...
+
+\title{Local Assembly in HaplotypeCaller and Mutect}
+\author{David Benjamin\footnote{The author took no part in development of the methods described below -- credit belongs to several others on the GATK team. }}
+\email{davidben@broadinstitute.org}
+\affiliation{Broad Institute, 75 Ames Street, Cambridge, MA 02142}
+
+\date{\today}
+
+\begin{abstract}
+The GATK tools \HC~and \Mutect~assemble reads aligned within a window of several hundred base pairs into a de Bruijn graph of local variation.  Local haplotypes correspond to paths in this de Bruijn graph, and the downstream likelihood calculations in both tools involve aligning reads to these haplotypes.  Here we describe how this de Bruijn graph is generated.
```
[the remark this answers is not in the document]
davidbenjamin: You might be overestimating me when you say "I know you know that."  I'm trying to figure out exactly what makes our read threading graph (before pruning and clean-up, of course) not a de Bruijn graph.  The accepted definition in the literature seems to me to be:

* Vertices are kmers
* Two vertices get a directed edge whenever they appear consecutively in the same read.

Our graph differs in two ways that I can think of:

* We keep track of the multiplicity of edges.
* We thread the reference haplotype and GGA sequences as if they were reads.

I can certainly accept being more precise and calling our graph a weighted de Bruijn graph with a reference pseudo-read, but I fear that you have another, more fundamental difference in mind.

docs/local_assembly.tex
```diff
@@ -0,0 +1,121 @@
...
+
+\section{Building the graph} \label{graph-assembly}
+Next, the \code{ReadThreadingAssembler} assembles the (corrected) reads over several different kmer sizes specified by the \code{kmerSize} argument\footnote{By default, 10 and 25.  This size is a compromise and no single value is the best choice for all regions.  Large kmers are more likely to be unique and to yield a graph with no cycles, which is especially important in low-complexity regions, but they are more sensitive to errors and low coverage.}.  If the given kmer sizes fail to produce a graph without cycles, or if more than $1/5$ of the kmers in the graph are not unique in the sequences from which they come\footnote{This uniqueness condition is waived on the last attempt when $k$ has been increased by 60 bases.}, the \code{ReadThreadingAssembler} repeatedly increases $k$ by 10 bases until assembly succeeds or until $k$ has been increased 6 times, that is, by 60 bases\footnote{These constants, 10 and 6, are hard-coded, but the attempts to increase $k$ can be disabled completely with the \code{dontIncreaseKmerSizesForCycles} argument.}.
```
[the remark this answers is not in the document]
davidbenjamin: Agreed and done.

docs/local_assembly.tex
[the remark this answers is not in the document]
davidbenjamin: > "Up to this point, we have produced a de Bruijn graph in textbook fashion" is a very inaccurate statement as argued above.

See above - I'm trying to understand.

> I think that you mean to say twice or more

That's right.  I changed it to be clearer.

> I think it is important to explicitly state the concepts of unique and non-unique kmers

Agreed and done.

> The very first sequence that we thread is the reference sequence. In that special case we add all the kmers to the graph (I believe) unique or non-unique.

You're right.  Even after you told me it was easy to miss.  Thank you!

> I'm not sure what is done with the given alleles in GGA mode

I just checked and we use the GGA `VariantContexts` to create GGA haplotypes via substitution on ref allele --> alt allele in the ref haplotype.  Then we thread these as non-ref sequences after the ref haplotype and before the reads.  That is, we build GGA haplotypes and treat them as long reads.  I put this in the docs.

> So you may have a early kmer that is unique but because is not part of the reference sequence is not use as start

I *very hesitantly* think I got this right originally.  Let me try to make the case.  Looking at the code in `ReadThreadingGraph::threadSequence` it calls `findStart`, which goes through every kmer in the read and calls `isThreadingStart`, which is a one-liner:

`return startThreadingOnlyAtExistingVertex ? uniqueKmers.containsKey(kmer) : !nonUniqueKmers.contains(kmer)`

Since `startThreadingOnlyAtExistingVertex` is hard-coded to be `false`, a kmer is considered a start if it is *not non-unique*, even if hasn't been put on the graph yet.  After returning from `findStart`, we invoke `getOrCreateKmerVertex`, which creates a vertex for the kmer if one doesn't exist already.

> I think that after this is done the prefix of the read before such a start kmer is threaded upstream. . . Don't know what would happen with those sequence that don't have any unique kmer that is not present in the reference

If I'm right in the discussion above, these concerns would be moot.

docs/local_assembly.tex
davidbenjamin: See above and let me know what you think.  Maybe we're just debating how much you can change and still call it a de Bruijn graph, but maybe I'm overlooking something.

docs/local_assembly.tex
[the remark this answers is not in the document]
vruano: Kmer <-> vertex is not a bijective relationship, a kmer may have more than one vertex (non-unique kmers at the very least). Also partly as consequence of that we don't have many cycles that otherwise are present in the deBruijn graph.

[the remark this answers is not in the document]
vruano: Also due to the way we construct the graph... this may well be different depending on what order we process the sequences.... scary huh!? The union of the individual sequences deBruijn's would be the same graph regardless of the order the sequences are processed... I think.

docs/local_assembly.tex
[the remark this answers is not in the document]
vruano: All what you saying seems reasonable.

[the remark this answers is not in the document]
davidbenjamin: > Also when we thread a read that has diverged from the reference and the next kmer is non-unique it will create a new node for that kmer even it is already present in the graph unless the it is already present as an extension of the previous kmer node

Let's check this out in the code of `ReadThreadingGraph::extendChainByOne`.  First it looks at all outgoing edges from the previous kmer.  If one of these edges goes to the new kmer we just increment its multiplicity, as you said.  Then we call `getUniqueKmerVertex`, which checks whether this a unique kmer (i.e. a non-non-unique kmer that has already been added to the graph).  If it's unique, we add an edge to the already existing vertex.  Otherwise, that is, if it is non-unique *or* hasn't been added to the graph yet, we create a new vertex.  At the end of this long analysis, you are correct.  I suppose this is another place where we diverge from a vanilla de Bruijn graph.

docs/local_assembly.tex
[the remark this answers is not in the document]
davidbenjamin: Addendum: another difference based on verifying your points is that we create a new vertex for non-unique kmers, even those that already appear on the graph.  Is there anything else?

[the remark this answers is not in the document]
davidbenjamin: > Kmer <-> vertex is not a bijective relationship, a kmer may have more than one vertex (non-unique kmers at the very least)

Okay, I get how this is fundamental.  I think I will change the discussion to explicitly highlight the differences.  That way people who are familiar with de Bruijn graphs can find comfort in the fact that most of their knowledge carries over to these docs, but will not be led to believe that everything is the same.

docs/local_assembly.tex
davidbenjamin: Got it.

docs/local_assembly.tex
davidbenjamin: > I'm not sure that depth-first is an accurate descriptor

It seems to me that *if not for caching* the search would be depth-first, that is, the recursion essentially has the form
```
find(vertex) {
   for (edge in outgoing edges) {
      find(target(edge));
   }
}
```
What I mean is that I don't want to just say that it's dynamic programming because that explains the caching but not the traversal.  This matters because if you implemented it, say, as breadth-first search you would get an explosion on sub-problems with no benefit from the caching.

I certainly agree that it doesn't explicitly enumerate the haplotypes.  How about I say that it obtains equivalent results to an exhaustive depth-first search but greatly reduces the cost by caching the best scores starting from every previously considered vertex?

docs/local_assembly.tex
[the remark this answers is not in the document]
davidbenjamin: I changed the docs.

davidbenjamin:
@vruano Thanks for a great review.  You made me earn it!  I think this latest commit covers everything, with the exception that my description of the haplotype finder might still be inadequate.

davidbenjamin:
@vruano Final thoughts?

docs/local_assembly.tex
```diff
@@ -58,12 +58,19 @@ \section{Correcting Reads} \label{correcting-reads}
...
+\item A non-unique kmer (see below) may be associated with multiple vertices.
+\end{itemize}
+
+Next, the \code{ReadThreadingAssembler} assembles the (corrected) reads over several different kmer sizes specified by the \code{kmerSize} argument\footnote{By default, 10 and 25.  This size is a compromise and no single value is the best choice for all regions.  Large kmers are more likely to be unique and to yield a graph with no cycles, which is especially important in low-complexity regions, but they are more sensitive to errors and low coverage.}.  If the given kmer sizes fail to produce a graph without cycles, or if more than $1/5$ of the kmers in the graph are not unique in the sequences from which they come\footnote{This uniqueness condition is waived on the last attempt when $k$ has been increased by 60 bases.}, the \code{ReadThreadingAssembler} repeatedly increases $k$ by 10 bases, starting from the largest initial $k$, until assembly succeeds or until $k$ has been increased 6 times, that is, by 60 bases\footnote{These constants, 10 and 6, are hard-coded, but the attempts to increase $k$ can be disabled completely with the \code{dontIncreaseKmerSizesForCycles} argument.}.
```
vruano: > Large kmers are more likely to be unique and to yield...
> but they are more sensitive to errors and low coverage.

mmm... why you're saying that? has it been shown to be the case.

I think larger could would be less sensitive sometimes
for example they may fail to close back into the reference after a even close to the end of the active region ... should be recoverable as a dangling tail but not always. The viceversa could be true with dangling heads.

I don't think that anyone has ever do a proper analysis of this so is all conjecture... but if you are confident that what you are saying is accurate then go ahead.

docs/local_assembly.tex
```diff
@@ -58,12 +58,19 @@ \section{Correcting Reads} \label{correcting-reads}
...
...a set of sequences to be kmerized and put into the de Bruijn graph.  These sequences are: the reference haplotype, any requested alleles given by the \code{alleles} argument in \HC's ``genotype given alleles" (GGA) mode\footnote{This mode is disabled in \Mutect~ and incompatible with the reference confidence mode of \HC, which is the recommended best practice.  We document it here for completeness.  Before assembly, the GATK engine creates GGA haplotypes by substituting the reference bases with the alt bases from the GGA vcf in the reference haplotype.  These GGA haplotypes are treated as non-reference sequences, that is, as long reads.}, and maximal subsequences of reads with base quality at least 10 by default\footnote{This is controlled by the \code{minBaseQualityScore} argument.} and at least one kmer long.  That is, a read with 100 bases and a base of quality 7 at position 70 yields sequences of length 69, 30 if $k = 10$ and a single sequence of length 69 if $k = 40$.  We do not use mate information; that is, a read and its mate yield completely independent sequences, nor do we later post-process the graph using mate information.  This is a potential direction for improvement.
```
vruano: If we have more than one variant in an active region do we induce all the possible allele combinations or we only consider one variant/allele at a time. In that case a small kmer size is important to capture all plausible haplotypes.

docs/local_assembly.tex
```diff
@@ -58,12 +58,19 @@ \section{Correcting Reads} \label{correcting-reads}
...
...e same sequence i.e. more than once in the reference or in any read.  These are called ``non-unique'' kmers and a kmer that is not non-unique is called ``unique.''  Then each sequence is kmerized starting from the first unique kmer\footnote{Note that a kmer that occurs only once in this particular sequence but more than once in some other sequence in non-unique.}, with the exception that the reference sequence begins at its first kmer, unique or not.  When a new kmer \textit{or a non-unique kmer (even if it is already on the graph)} appears, a new vertex is added to the graph; otherwise an edge is added (or the multiplicity of an existing edge incremented) from the preceding kmer to the current one.  The advantage of creating a new vertex for a non-unique kmer rather than re-using an old vertex is that it avoids spurious cycles.  Up to this point, we have strayed from a canonical de Bruijn graph only as noted above.  Note that this is a directed graph of kmers from aligned reads (i.e. aligned to the forward strand of the reference), not a bidirected graph of unaligned reads that could come from either strand, with the complications of associating a kmer with its reverse complement.
```
vruano: > When a new kmer appears, a new vertex is added to the graph; otherwise an edge is added...

The otherwise is only true for unique kmers as I guess otherwise means an old kmer appears; old non-unique kmers may add a vertex if an edge from the previous kmer vertex with that extending base does not exists (adds a vertex and an edge to that vertex).

vruano:
A few more comments for you @davidbenjamin

docs/local_assembly.tex
davidbenjamin: I think I was saying what you're saying but with a confusing overload of the word "sensitive."  I meant that large kmers would be more *affected* by errors, which could decrease sensitivity.  I'll make that more clear.

docs/local_assembly.tex
davidbenjamin: We do not induce all possible GGA haplotypes.  I added a sentence to the GGA footnote.

docs/local_assembly.tex
davidbenjamin: It's now: `When a new kmer \textit{or a non-unique kmer (even if it is already on the graph)} appears, a new vertex is added to the graph; otherwise an edge is added (or the multiplicity of an existing edge incremented) from the preceding kmer to the current one.`

davidbenjamin:
I addressed your comments @vruano.

docs/local_assembly.tex
vruano: Is a bit more complex than that I'm afraid.... and old non-unique kmer may or may not result in a new vertex.
When a new vertex appear this is added to the graph, when and old vertex appear a few things may happen. If it is unique then no new vertex is added (and a edge is draw from the current vertex to this (old) next), if it is non-unique then if there is already a edge between the current vertex and the old next then we only increase the weight on that edge otherwise a new vertex and edge are added to the graph.

davidbenjamin:
@vruano Done with the last comment.  Thanks for your help!

PR #3370, opened 2017-07-27
Add scripts to manage SV spark jobs and copy result

TedBrookings:
Add new scripts to gatk/scripts/sv/ folder, and alter action (but not
passed parameters) of older scripts to make running sv spark jobs
more convenient.
Added:
  -copy_sv_results.sh: copy files to time and git-stamped folder on
   google cloud storage
     -> results folder on cluster
     -> command line arguments to SV discover pipeline
     -> console log file (if present)

  -manage_sv_pipeline.sh: create cluster, run job, copy results, and
   delete cluster. Manage cluster naming, time and git-stamping,
   and log file production.

Altered:
  -create_cluster.sh: control GCS zone and numbers of workers via
   environmental variables. Defaults to previous hard-coded values.

  -runWholePipeline.sh: accept command-line arguments for sv
   discovery pipeline, work with clusters having NUM_WORKERS != 10

### DIFF
--- a/scripts/sv/copy_sv_results.sh
+++ b/scripts/sv/copy_sv_results.sh
@@ -9,7 +9,7 @@ set -eu
 set -o pipefail
 
 if [[ "$#" -lt 3 ]]; then
-    echo -e
+    echo -e \
 "Please provide:
   [1] GCS project name (required)
   [2] GCS cluster name (required)
@@ -17,7 +17,11 @@ if [[ "$#" -lt 3 ]]; then
   [4] GCS user name (defaults to local user name)
   [5] path to local log file (default to empty, i.e. no log)
   [*] additional arguments that were passed to
-      StructuralVariationDiscoveryPipelineSpark"
+      StructuralVariationDiscoveryPipelineSpark
+To leave a value as default but specify a later value, use an empty
+  string. e.g. to use default user name but override log file:
+\$ copy_sv_results.sh my_project my_cluster_name /output/dir \\
+     \"\" /my/log/file.log"
   exit 1
 fi
 
--- a/scripts/sv/copy_sv_results.sh
+++ b/scripts/sv/copy_sv_results.sh
@@ -15,7 +15,9 @@ if [[ "$#" -lt 3 ]]; then
   [2] GCS cluster name (required)
   [3] cluster output directory (required)
   [4] GCS user name (defaults to local user name)
-  [5] path to local log file (default to empty, i.e. no log)
+  [5] GCS save bucket/path (defaults to \$PROJECT_NAME/\$GCS_USER if
+      omitted or empty)
+  [6] path to local log file (default to empty, i.e. no log)
   [*] additional arguments that were passed to
       StructuralVariationDiscoveryPipelineSpark
 To leave a value as default but specify a later value, use an empty
@@ -29,9 +31,10 @@ PROJECT_NAME=$1
 CLUSTER_NAME=$2
 OUTPUT_DIR=$3
 GCS_USER=${4:-${USER}}
-LOCAL_LOG_FILE=${5:-"/dev/null"}
+GCS_SAVE_PATH=${5:-"${PROJECT_NAME}/${GCS_USER}"}
+LOCAL_LOG_FILE=${6:-"/dev/null"}
 
-shift 5
+shift $(($# < 6 ? $# : 6))
 SV_ARGS=${*:-${SV_ARGS:-""}}
 
 # get appropriate ZONE for cluster
@@ -55,11 +58,11 @@ if [ -z "${RESULTS_DIR}" ]; then
     # directory is empty (presumably the job crashed). Use OUTPUT_DIR (without leading slash)
     RESULTS_DIR=$(echo "${OUTPUT_DIR}" | sed -e 's/^\///')
     echo "RESULTS_DIR=${RESULTS_DIR}" 2>&1 | tee -a ${LOCAL_LOG_FILE}
-    GCS_RESULTS_DIR="gs://${PROJECT_NAME}/${GCS_USER}/${RESULTS_DIR}"
+    GCS_RESULTS_DIR="gs://${GCS_SAVE_PATH}/${RESULTS_DIR}"
 else
     # copy the latest results to google cloud
     echo "RESULTS_DIR=${RESULTS_DIR}" 2>&1 | tee -a ${LOCAL_LOG_FILE}
-    GCS_RESULTS_DIR="gs://${PROJECT_NAME}/${GCS_USER}/${RESULTS_DIR}"
+    GCS_RESULTS_DIR="gs://${GCS_SAVE_PATH}/${RESULTS_DIR}"
     # chose semi-optimal parallel args for distcp
     # 1) count number of files to copy
     COUNT_FILES_CMD="hadoop fs -count /${RESULTS_DIR}/ | tr -s ' ' | cut -d ' ' -f 3"
--- a/scripts/sv/copy_sv_results.sh
+++ b/scripts/sv/copy_sv_results.sh
@@ -1,0 +1,98 @@
+#!/bin/bash
+
+# This script copies SV analysis results from a Google Dataproc cluster
+# to an appropriate bucket/directory on GCS. It also uploads contents of
+# local output logs
+
+# terminate script on error or if a command fails before piping to another command
+set -eu
+set -o pipefail
+
+if [[ "$#" -lt 3 ]]; then
+    echo -e
+"Please provide:
+  [1] GCS project name (required)
+  [2] GCS cluster name (required)
+  [3] cluster output directory (required)
+  [4] GCS user name (defaults to local user name)
+  [5] path to local log file (default to empty, i.e. no log)
+  [*] additional arguments that were passed to
+      StructuralVariationDiscoveryPipelineSpark"
+  exit 1
+fi
+
+PROJECT_NAME=$1
+CLUSTER_NAME=$2
+OUTPUT_DIR=$3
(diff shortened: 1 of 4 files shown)


scripts/sv/runWholePipeline.sh
```diff
@@ -34,6 +35,16 @@ KMER_KILL_LIST=$(echo "${REF_FASTA}" | sed 's/.fasta$/.kill.kmers/')
...
 REF_TWOBIT=$(echo "${REF_FASTA}" | sed 's/.fasta$/.2bit/')
 
+# extract any extra arguments to StructuralVariationDiscoveryPipelineSpark
+shift 6
+SV_ARGS=${*:-${SV_ARGS:-""}}
+# expand any local variables passed as strings (e.g. PROJECT_OUTPUT_DIR)
+eval "SV_ARGS=\"${SV_ARGS}\""
+
+# Choose NUM_EXECUTORS = 2 * NUM_WORKERS
+NUM_WORKERS=$(gcloud compute instances list --filter="name ~ ${CLUSTER_NAME}-[sw].*" | grep RUNNING | wc -l)
```
SHuang-Broad: I got warnings saying "This argument is deprecated". 
How about
```bash
gcloud dataproc clusters list | grep -F "${CLUSTER_NAME}" | awk '{print $2}'
```

scripts/sv/manage_sv_pipeline.sh
```diff
@@ -0,0 +1,174 @@
...
+if [[ "$#" -lt 4 ]]; then
+    echo -e \
+"Please provide:
+  [1] local directory of GATK build (required)
+  [2] project name (required)
+  [3] GCS path to indexed BAM (required)
+  [4] GCS path to reference fasta (required)
+      OPTIONAL arguments:
+  [5] path to initialization script (local or GCS, defaults to
+      ${GATK_DIR}/scripts/sv/default_init.sh if omitted or empty)
```
SHuang-Broad: I have a concern here: if the 5th and 6th argument are omitted, is the `shift 6` statement below going to skip two arguments that are intended to be sent to java be "shifted" away?

scripts/sv/manage_sv_pipeline.sh
```diff
@@ -0,0 +1,174 @@
...
+INIT_SCRIPT=${5:-${INIT_SCRIPT:-"${GATK_DIR}/scripts/sv/default_init.sh"}}
+GCS_USER=${6:-${GCS_USER:-${USER}}}
+shift 6
+SV_ARGS=${*:-${SV_ARGS:-""}} && SV_ARGS=${SV_ARGS:+" ${SV_ARGS}"}
+
+# add GATK SV scripts to PATH
+PATH="${GATK_DIR}/scripts/sv:${PATH}"
+
+# configure caching .jar files
+export GATK_GCS_STAGING="gs://${PROJECT_NAME}/${GCS_USER}/staging/"
```
SHuang-Broad: I would prefer a check if an environment variable has been set or not, because a user may already have this set in my shell profile ( I personally do), and preferably would like the jars being cached there.

scripts/sv/manage_sv_pipeline.sh
```diff
@@ -0,0 +1,174 @@
...
+
+# set cluster name based on user and target bam file
+# (NOTE: can override by defining SV_CLUSTER_NAME)
```
SHuang-Broad: This doesn't seem to be exposed to the user (i.e. it is hidden knowledge).

scripts/sv/manage_sv_pipeline.sh
```diff
@@ -0,0 +1,174 @@
...
+SANITIZED_BAM=$(basename "${GCS_BAM}" | awk '{print tolower($0)}' | sed 's/[^a-z0-9]/-/g')
+CLUSTER_NAME=${SV_CLUSTER_NAME:-"${GCS_USER}-${SANITIZED_BAM}"}
+
+# update gcloud
+gcloud components update
```
SHuang-Broad: requires interaction from user, may be a surprise to someone.

scripts/sv/manage_sv_pipeline.sh
```diff
@@ -0,0 +1,174 @@
...
+                ;;
+        *)      echo "Please answer yes, no, or cancel."
+                ;;
+    esac
+done
+
+# copy results into gcloud
+while true; do
+    echo "#############################################################" 2>&1 | tee -a ${LOCAL_LOG_FILE}
+    read -p "Copy results? (yes/no/cancel)" yn
```
SHuang-Broad: sorry for being edge-case picky: if a user answered no to run whole pipeline, this would end up copying nothing, correct?

scripts/sv/copy_sv_results.sh
```diff
@@ -0,0 +1,98 @@
...
+  exit 1
+fi
+
+PROJECT_NAME=$1
+CLUSTER_NAME=$2
+OUTPUT_DIR=$3
+GCS_USER=${4:-${USER}}
+LOCAL_LOG_FILE=${5:-"/dev/null"}
+
+shift 5
```
SHuang-Broad: Same concern: if we only provide 3 required arguments and the rest is java arguments, are 4 and 5 "shifted" away.

SHuang-Broad:
Turns out a typo prevents running the "manage_sv_pipeline" script, saying GATK_DIR is an unbound variable. Please fix.

scripts/sv/manage_sv_pipeline.sh
TedBrookings: It depends. They may have run previously (perhaps not using this script, or perhaps there was an error). If there's data, the copy script should copy it off, and regardless, it should copy any log files or command line arguments.

scripts/sv/manage_sv_pipeline.sh
TedBrookings: Agreed. I'll fix that.

scripts/sv/runWholePipeline.sh
TedBrookings: That would eliminate the warning.  I think the warning itself is actually a bug (it corresponds to passing the NAME argument, not using a filter). I've avoided listing all instances and using grep because that corresponds to potentially downloading a large number of instances, whereas this in principle might filter before download, saving data. If the warning is disconcerting, I can make the fix you suggest and then revert once google gets their act together on this.

scripts/sv/manage_sv_pipeline.sh
TedBrookings: The user would have to pass empty strings (signaling default values) for argument 4 and 5 to then pass arguments 6 and above. I added a description of this to the documentation and pushed the fix.

scripts/sv/copy_sv_results.sh
TedBrookings: Here as well, I just added a description of how to specify later values and leave early ones as default.

scripts/sv/manage_sv_pipeline.sh
TedBrookings: Sorry, I somehow missed this comment. I just added a line to inform the user what the CLUSTER_NAME was set to. I'll push again.

TedBrookings:
Unbound variable bug fixed.

scripts/sv/runWholePipeline.sh
TedBrookings: I found a filter command for the gcloud dataproc clusters list command, so I changed and pushed that. It has the downside of not seeing preemptible instances, but preemptible instances actually crash the pipeline right now, so it's not a super-important problem.

SHuang-Broad:
I cannot seem to be able to run the pipeline with the following way of running it:

```bash
./manage_sv_pipeline.sh \
    /Users/shuang/GATK \
    broad-dsde-methods \
    gs://broad-dsde-methods/sv/samples/G94797_CHM_MIX/WGS1/G94794.CHMI_CHMI3_WGS1.cram.bam \
    gs://broad-dsde-methods/sv/reference/GRCh38/Homo_sapiens_assembly38.fasta \
    gs://broad-dsde-methods/shuang/tmp/gatk-jars/default_init.sh
```

And if I run with the following way (only adding a user name)

```bash
./manage_sv_pipeline.sh \
    /Users/shuang/GATK \
    broad-dsde-methods \
    gs://broad-dsde-methods/sv/samples/G94797_CHM_MIX/WGS1/G94794.CHMI_CHMI3_WGS1.cram.bam \
    gs://broad-dsde-methods/sv/reference/GRCh38/Homo_sapiens_assembly38.fasta \
    gs://broad-dsde-methods/shuang/tmp/gatk-jars/default_init.sh \
    shuang
```
The script runs as expected.

I believe line 47 is the culprit:
```bash
SV_ARGS=${*:-${SV_ARGS:-""}} && SV_ARGS=${SV_ARGS:+" ${SV_ARGS}"}
```

SHuang-Broad:
My test run has successfully finished. Awesome, @TedBrookings !

Two suggestions that you could decide when to address them:
1. This script requires a lot of user interaction (gcloud update, whether to create a cluster, whether to copy results, whether to delete cluster), I could imagine this being inconvenient for some, so we could have some upfront arguments specifying answers to these questions when the script is launched.
2. The bucket to which the results are copied is not overridable by caller of script. A user might want to copy the results to a specific location.

TedBrookings:
I addressed the crash (caused by calling shift with too large a number and set-ing strict error checks in bash).

I also added the requested options. Due to the larger number of options I decided that using only positional arguments would be impractical (requiring counting the number of empty "" default arguments) so I added a unix-like option parsing scheme to manage_sv_pipeline.sh
--quiet or -q will cause the script to run without any user prompting
--save [bucket/path] or -s [bucket/path] will save results and initialization scripts within the specified bucket/path

PR #3371, opened 2017-07-27
Added support for uploading the zipped gatk file on a nightly basis to google buckets

### LINKED ISSUE #3168: Publish nightly builds for GATK4
The zip produced by `./gradlew clean bundle` should be published nightly on the GATK website, as we did for GATK3.

jamesemery:
There is a second and perhaps more proper method for accomplishing the same thing commented out in the travis.yml which I could not get to work immediately. I invite suggestion as to whether this should be our approach, It would still require separate code in order to remove older zips from the repository.

fixes #3168

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,6 +85,12 @@ before_install:
 - if [[ $RUN_M2_WDL == true ]]; then
     wget -O ~/picard.jar https://github.com/broadinstitute/picard/releases/download/2.9.0/picard.jar;
   fi
+# Installing Travis for test report uploading
+- if [[ $TRAVIS_EVENT_TYPE == cron ]]; then
+    curl -sSL https://get.rvm.io | bash -s stable --ruby;
+    source /home/travis/.rvm/scripts/rvm;
+    gem install travis;
+  fi
 # Download git lfs files
 - git lfs version
 - git lfs install
@@ -106,7 +112,9 @@ install:
 script:
 # run a basic sanity check to be sure that gatk-launch doesn't explode
 # run tests
-- if [[ $TRAVIS_SECURE_ENV_VARS == false && $TEST_TYPE == cloud ]]; then
+- if [[ $TRAVIS_EVENT_TYPE == cron ]]; then
+    echo "Not running any tests for nightly builds";
+  elif [[ $TRAVIS_SECURE_ENV_VARS == false && $TEST_TYPE == cloud ]]; then
     echo "Can't run cloud tests without keys so don't run tests";
   elif [[ $RUN_CNV_SOMATIC_WDL == true ]]; then
     echo "Running CNV somatic workflows";
@@ -147,12 +155,24 @@ after_success:
     git tag master;
     ./gradlew uploadArchives; 
   fi
+
+# This creates and uploads the gatk zip file to the nightly build bucket, only keeping the 10 newest entries
+- if [[ $TRAVIS_BRANCH == master && $TRAVIS_EVENT_TYPE == cron && $UPLOAD == true && "$(travis show master | head -2 | grep 'State:' | awk '{print $NF}')" == passed ]]; then
+    $GCLOUD/gcloud components -q update gsutil;
+    gsutil ls -l gs://gatk-nightly-builds | grep gatk | sort -r -k 2 | grep -o '\S\+$' | tail -n +11 | xargs -I {} gsutil rm {};
+    ./gradlew bundle;
+    ZIP_FILE="$(ls build/ | grep .zip)";
+    echo "Uploading zip to gs://gatk-nightly-builds/";
+    $GCLOUD/gsutil -m cp build/$ZIP_FILE gs://gatk-nightly-builds/"$(date +%Y-%m-%d)"-$ZIP_FILE;
+  fi
 after_failure:
 - dmesg | tail -100
 after_script:
-- if [ $TRAVIS_SECURE_ENV_VARS == true ]; then
+- if [[ $TRAVIS_SECURE_ENV_VARS == true && $TRAVIS_EVENT_TYPE != cron ]]; then
     $GCLOUD/gcloud components -q update gsutil;
     REPORT_PATH=$TRAVIS_BRANCH_$TRAVIS_JOB_NUMBER;
     $GCLOUD/gsutil -m cp -z html -z js -z xml -z css -r build/reports/tests gs://hellbender/test/build_reports/$REPORT_PATH/;
     echo "See the test report at https://storage.googleapis.com/hellbender/test/build_reports/$REPORT_PATH/tests/test/index.html";
   fi
+
+


.travis.yml
```diff
@@ -150,9 +158,26 @@ after_success:
...
-- if [ $TRAVIS_SECURE_ENV_VARS == true ]; then
+- if [ $TRAVIS_SECURE_ENV_VARS == true && $TRAVIS_EVENT_TYPE != cron ]; then
     $GCLOUD/gcloud components -q update gsutil;
     REPORT_PATH=$TRAVIS_BRANCH_$TRAVIS_JOB_NUMBER;
     $GCLOUD/gsutil -m cp -z html -z js -z xml -z css -r build/reports/tests gs://hellbender/test/build_reports/$REPORT_PATH/;
     echo "See the test report at https://storage.googleapis.com/hellbender/test/build_reports/$REPORT_PATH/tests/test/index.html";
   fi
+# This creates and uploads the gatk zip file to the nightly build bucket, only keeping the 10 newest entries
+- if [[ $TRAVIS_BRANCH == master && $UPLOAD == true && "$(travis show master | head -2 | grep 'State:' | awk '{print $NF}')" == passed ]]; then
+    gsutil ls -l gs://gatk-nightly-builds | awk '{$1=""}1' | sort -r | awk '{$1=""}1' | awk 'NR>9' | xargs -I {} gsutil rm {};
```
lbergelson: bleeeeech awk

.travis.yml
```diff
@@ -150,9 +158,26 @@ after_success:
...
+    ./gradlew bundle;
+    $GCLOUD/gsutil -m cp build/"$(ls build/ | grep .zip)" gs://gatk-nightly-builds/"$(date +%Y-%m-%d)"-"$(ls build/ | grep .zip)";
```
lbergelson: we probably want to include the hash in here as well?

.travis.yml
```diff
@@ -150,9 +158,26 @@ after_success:
...
 after_script:
-- if [ $TRAVIS_SECURE_ENV_VARS == true ]; then
+- if [ $TRAVIS_SECURE_ENV_VARS == true && $TRAVIS_EVENT_TYPE != cron ]; then
     $GCLOUD/gcloud components -q update gsutil;
     REPORT_PATH=$TRAVIS_BRANCH_$TRAVIS_JOB_NUMBER;
     $GCLOUD/gsutil -m cp -z html -z js -z xml -z css -r build/reports/tests gs://hellbender/test/build_reports/$REPORT_PATH/;
     echo "See the test report at https://storage.googleapis.com/hellbender/test/build_reports/$REPORT_PATH/tests/test/index.html";
   fi
+# This creates and uploads the gatk zip file to the nightly build bucket, only keeping the 10 newest entries
+- if [[ $TRAVIS_BRANCH == master && $UPLOAD == true && "$(travis show master | head -2 | grep 'State:' | awk '{print $NF}')" == passed ]]; then
```
lbergelson: Do cron jobs run on branches other than master?  
Shouldn't this check TRAVIS_EVEN_TYPE == cron?
Does this need to check that it's not a pull request build?

jamesemery: Currently they don't, one can set the cron job to run off of any branch one wants. We aren't using it though.

jamesemery:
@lbergelson I just responded to your comments in the most to-date version of the branch

.travis.yml
```diff
@@ -85,6 +85,12 @@ before_install:
 - if [[ $RUN_M2_WDL == true ]]; then
     wget -O ~/picard.jar https://github.com/broadinstitute/picard/releases/download/2.9.0/picard.jar;
   fi
+# Installing Travis for test report uploading
+- if [[ $TRAVIS_EVENT_TYPE == cron ]]; then
+    curl -sSL https://get.rvm.io | bash -s stable --ruby;
+    source ~/.rvm/scripts/rvm;
```
lbergelson: Bye matt, we'll miss you!

.travis.yml
```diff
@@ -150,9 +158,26 @@ after_success:
...
-- if [ $TRAVIS_SECURE_ENV_VARS == true ]; then
+- if [ $TRAVIS_SECURE_ENV_VARS == true && $TRAVIS_EVENT_TYPE != cron ]; then
     $GCLOUD/gcloud components -q update gsutil;
     REPORT_PATH=$TRAVIS_BRANCH_$TRAVIS_JOB_NUMBER;
     $GCLOUD/gsutil -m cp -z html -z js -z xml -z css -r build/reports/tests gs://hellbender/test/build_reports/$REPORT_PATH/;
     echo "See the test report at https://storage.googleapis.com/hellbender/test/build_reports/$REPORT_PATH/tests/test/index.html";
   fi
+# This creates and uploads the gatk zip file to the nightly build bucket, only keeping the 10 newest entries
+- if [[ $TRAVIS_BRANCH == master && $TRAVIS_EVENT_TYPE == cron && $UPLOAD == true && "$(travis show master | head -2 | grep 'State:' | awk '{print $NF}')" == passed ]]; then
+    gsutil ls -l gs://gatk-nightly-builds | grep gatk | sort -r -k 2 | grep -o '\S\+$' | tail -n +11 | xargs -I {} gsutil rm {};
```
lbergelson: Is my understanding of this chain of command correct?

1. List all the things
2.  grab the ones that look like gatk
3 sort by oldest first? 
4. select a contiguous block of whitespace until the end of the line
5.  grab everything after line 10
6 delete them

.travis.yml
```diff
@@ -150,9 +158,26 @@ after_success:
...
 - dmesg | tail -100
 after_script:
-- if [ $TRAVIS_SECURE_ENV_VARS == true ]; then
+- if [ $TRAVIS_SECURE_ENV_VARS == true && $TRAVIS_EVENT_TYPE != cron ]; then
     $GCLOUD/gcloud components -q update gsutil;
     REPORT_PATH=$TRAVIS_BRANCH_$TRAVIS_JOB_NUMBER;
     $GCLOUD/gsutil -m cp -z html -z js -z xml -z css -r build/reports/tests gs://hellbender/test/build_reports/$REPORT_PATH/;
     echo "See the test report at https://storage.googleapis.com/hellbender/test/build_reports/$REPORT_PATH/tests/test/index.html";
   fi
+# This creates and uploads the gatk zip file to the nightly build bucket, only keeping the 10 newest entries
```
lbergelson: @jamesemery I just noticed, shouldn't this really go into the after success block?  If the ruby install fails there's no point in doing the upload.

.travis.yml
```diff
@@ -150,9 +158,26 @@ after_success:
...
+- if [[ $TRAVIS_BRANCH == master && $TRAVIS_EVENT_TYPE == cron && $UPLOAD == true && "$(travis show master | head -2 | grep 'State:' | awk '{print $NF}')" == passed ]]; then
+    gsutil ls -l gs://gatk-nightly-builds | grep gatk | sort -r -k 2 | grep -o '\S\+$' | tail -n +11 | xargs -I {} gsutil rm {};
+    ./gradlew bundle;
+    $GCLOUD/gsutil -m cp build/"$(ls build/ | grep .zip)" gs://gatk-nightly-builds/"$(date +%Y-%m-%d)"-"$(ls build/ | grep .zip)";
+  fi
+#deploy:
```
lbergelson: We could probably use the deploy blocks instead of our own gsutil installation for this upload and the test reports, but since we have to install gsutil for gatk tests anyway it never seemed worthwhile figuring out how to configure it.  Lets delete these commented out lines, but if you want to revive them later as a different mechanism for upload it might be worth looking at.

.travis.yml
```diff
@@ -150,9 +158,26 @@ after_success:
...
+- if [ $TRAVIS_SECURE_ENV_VARS == true && $TRAVIS_EVENT_TYPE != cron ]; then
     $GCLOUD/gcloud components -q update gsutil;
     REPORT_PATH=$TRAVIS_BRANCH_$TRAVIS_JOB_NUMBER;
     $GCLOUD/gsutil -m cp -z html -z js -z xml -z css -r build/reports/tests gs://hellbender/test/build_reports/$REPORT_PATH/;
     echo "See the test report at https://storage.googleapis.com/hellbender/test/build_reports/$REPORT_PATH/tests/test/index.html";
   fi
+# This creates and uploads the gatk zip file to the nightly build bucket, only keeping the 10 newest entries
+- if [[ $TRAVIS_BRANCH == master && $TRAVIS_EVENT_TYPE == cron && $UPLOAD == true && "$(travis show master | head -2 | grep 'State:' | awk '{print $NF}')" == passed ]]; then
+    gsutil ls -l gs://gatk-nightly-builds | grep gatk | sort -r -k 2 | grep -o '\S\+$' | tail -n +11 | xargs -I {} gsutil rm {};
+    ./gradlew bundle;
```
lbergelson: Might want to just capture this`ls build/ | grep .zip` block into an environment variable and reuse that.  It might look cleaner that way.  It might also trigger a round of typo fixing in travis which I'm sure you're happy to avoid, so feel free to disregard the suggestion.

.travis.yml
jamesemery: 1. yes
2. yes (filtering out the summary line at the end)
3a. Actually youngest first, because the date is year-month-day the format 2017-07-26 appears after 2017-07-27 in alphanumeric order so we reverse it.
3b. Its actually non whitespace '\S' is the opposite of '\s', making this grabbing the last block of non whitespace at the end of the line
4. yes
6 yes

.travis.yml
jamesemery: This is a good point, I originally had it there but I was running into issues getting the code to install gsutil . Its notable however that even if ruby fails to install in the install code but it passes all the subsequent tests (none in this case so its a low bar) it will consider the execution a success and still run the after-success script. I will move it back there because I fixed the issue that was giving me a headache and it might catch some other issue that would pop up.

.travis.yml
lbergelson: ack, yes, meant non-whitespace.  I wonder what happened to 5....

lbergelson: also, I like the formatting match :)

.travis.yml
lbergelson: Oh interesting, I would have thought that ruby failure would cause build failure, but that makes sense given some of the runs that continued right through a git lfs failure.

jamesemery:
@lbergelson Responded to all of your comments

.travis.yml
```diff
@@ -147,12 +155,24 @@ after_success:
...
+    ZIP_FILE="$(ls build/ | grep .zip)";
+    echo "Uploading zip to gs://gatk-nightly-builds/";
+    $GCLOUD/gsutil -m cp build/$ZIP_FILE gs://gatk-nightly-builds/"$(date +%Y-%m-%d)"-$ZIP_FILE;
+  fi
 after_failure:
 - dmesg | tail -100
 after_script:
-- if [ $TRAVIS_SECURE_ENV_VARS == true ]; then
```
lbergelson: good catch here

PR #3375, opened 2017-07-27
M2 wdl outputs the contamination

davidbenjamin:
@LeeTL1220 as discussed today.


davidbenjamin:
@LeeTL1220 Did it.  I didn't make the output optional because the only natural way to do that would be to extract contamination as an independent task.  Which maybe we'll do some time, but not in this PR.

LeeTL1220:
@davidbenjamin Ready to merge whenever you want.

PR #3376, opened 2017-07-27
improved filtering for smallish indels

tedsharpe:
I think this is ready for prime time.  Seems to improve things a bit.


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVIntervalTree.java
```diff
@@ -26,10 +32,38 @@
  * all this weirdness.  The sentinel value is null by default, so put and remove will behave like you might expect them
  * to if you're not worrying about this stuff:  they'll return null for novel insertions and failed deletions.
  */
-public final class SVIntervalTree<V> implements Iterable<SVIntervalTree.Entry<V>> {
+    @DefaultSerializer(SVIntervalTree.Serializer.class)
+    public final class SVIntervalTree<V> implements Iterable<SVIntervalTree.Entry<V>> {
```
cwhelan: Is the indentation here funny?

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/IntHistogram.java
```diff
@@ -198,6 +202,46 @@ public boolean isDifferentByKSStatistic( final IntHistogram sampleHistogram, fin
...
+            if ( mCounts == 0L ) return null;
+
+            int locationOfMaxDiff = -1;
+            float maxDiff = 0.f;
+            float maxArea = 0.f;
+            long sum = 0L; // sum from 0 to idx of sample histogram counts
+            int idx = 0; // current index into sample histogram
+            float curVal = 0.f; // sum/mCounts as a float (between 0 and 1)
+            float curArea = 0.f;
+            while ( sum < mCounts ) {
```
cwhelan: In your previous implementation of this you had a check for "an extremum can occur at the end of a run of zero counts" but I don't see that in this version. Was it not longer needed?

src/main/java/org/broadinstitute/hellbender/tools/spark/utils/IntHistogram.java
```diff
@@ -198,6 +202,46 @@ public boolean isDifferentByKSStatistic( final IntHistogram sampleHistogram, fin
             return false;
         }
 
+        public KSData evalKS( final IntHistogram sampleHistogram ) {
```
cwhelan: It looks like this method is also only used in the `FindBreakpointEvidenceSpark` path that emits a lot more info about the data. It might be nice to comment on that and say when you might want to use this method as opposed to `isDifferentByKSStatistic`.

SHuang-Broad: If this function is to be moved to `FindSmallIndelRegions`, then the associated class `KSData` should be too.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindSmallIndelRegions.java
```diff
@@ -99,9 +229,18 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+                       final String cdfDir ) {
             this.readMetadata = readMetadata;
             this.filter = filter;
-            final Map<String, FragmentLengthStatistics> libraryStatisticsMap = readMetadata.getAllLibraryStatistics();
+            this.trueIntervals = trueIntervals;
+            this.cdfDir = cdfDir;
+            final Map<String, LibraryStatistics> libraryStatisticsMap = readMetadata.getAllLibraryStatistics();
             libraryToHistoPairMap = new HashMap<>(SVUtils.hashMapCapacity(libraryStatisticsMap.size()));
             libraryStatisticsMap.forEach( (libName, stats) -> {
                 final IntHistogram[] pair = new IntHistogram[2];
```
SHuang-Broad: Since the array is always size 2 (for the two halves), using a `Tuple2` is probably a better choice; but your call.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindSmallIndelRegions.java
```diff
@@ -114,49 +253,140 @@ public Finder( final ReadMetadata readMetadata, final SVReadFilter filter ) {
...
+            if ( !(sameContig = read.getContig().equals(curContig)) || read.getStart() >= curEnd ) {
+                getStatus(statusList);
+                advanceBlock(sameContig, read);
+            }
+            final IntHistogram[] histoPair =
+                    libraryToHistoPairMap.get(readMetadata.getLibraryName(read.getReadGroup()));
+            histoPair[fillIdx].addObservation(Math.abs(read.getFragmentLength()));
+        }
+
+        public void getStatus( final List<String> statusList ) {
```
SHuang-Broad: Not sure what "status" means here, so adding a documentation (if worth it) would be helpful.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindSmallIndelRegions.java
```diff
@@ -114,49 +253,140 @@ public Finder( final ReadMetadata readMetadata, final SVReadFilter filter ) {
...
+                    }
                 }
                 oldHisto.clear();
             }
-            fillIdx = oldIdx;
+        }
+
+        private void advanceBlock( final boolean sameContig, final GATKRead read ) {
+            final int oldIdx = fillIdx;
+            fillIdx ^= 1; // bit magic to switch halves
```
SHuang-Broad: This bit magic that affects `fillIdx` is executed in the master version only conditionally: i.e. when `sameContig` is true. Not sure if this change is intended.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindSmallIndelRegions.java
SHuang-Broad: The name of this function is not suggesting what it is doing: its name is `get`-* but actually are putting new elements into the provided argument. Why not return a `List` and let the caller add the returned list to the `statusList`?

tedsharpe:
Have another look, if you care to @SHuang-Broad and @cwhelan.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
```diff
@@ -271,6 +282,14 @@ public int getMaxMedianFragmentSize() {
         return contigNameToID;
     }
 
+    public static String[] buildContigIDToNameArray( final Map<String, Integer> nameToIDMap ) {
+        String[] result = new String[nameToIDMap.size()];
```
SHuang-Broad: You are right. Not sure what I was thinking when making the suggestion...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
```diff
@@ -317,11 +337,71 @@ public ReadMetadata read( final Kryo kryo, final Input input, final Class<ReadMe
...
+        }
+
+        public long getNReads() { return nReads; }
+        public long getNBases() { return nBases; }
+
+        public LibraryStatistics createLibraryStatistics( final long nRefBases ) {
+            return new LibraryStatistics(fragmentSizes.getCDF(), nBases, nReads, nRefBases);
+        }
+
+        public static LibraryRawStatistics reduce( final LibraryRawStatistics stats1,
```
SHuang-Broad: If so, I would suggest adding a documentation for that assumption.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/experimental/FindSmallIndelRegions.java
```diff
@@ -0,0 +1,347 @@
...
+import java.io.*;
+import java.util.*;
+
+import static org.broadinstitute.hellbender.tools.spark.sv.StructuralVariationDiscoveryArgumentCollection.FindBreakpointEvidenceSparkArgumentCollection;
+
+@CommandLineProgramProperties(summary="Experimental code to dump statistics about regions containing small indels due to fragment length anomalies.",
+        oneLineSummary="Experimental do not use.",
+        usageExample="gatk-launch FindSmallIndelRegions -O hdfs://cluster-name:8020/path/to/statusDir -I hdfs://cluster-name:8020/path/to/bam --alignerIndexImage notUsed --kmersToIgnore notUsed",
+        programGroup=StructuralVariationSparkProgramGroup.class)
+public final class FindSmallIndelRegions extends GATKSparkTool {
```
SHuang-Broad: Please add `@BetaFeature`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/experimental/FindSmallIndelRegions.java
```diff
@@ -0,0 +1,347 @@
...
+                        final String fileName =
+                                cdfDir + "/" + libName + "." + curContig + "." + start + "." + truthStatus + ".cdf";
+                        dumpCDF(fileName, oldHisto.getCDF(), status, trueIntervals.overlappers(curInterval));
+                    }
+                }
+                oldHisto.clear();
+            }
+        }
+
+        private KSData evalKS( final IntHistogram.CDF librarySample, final IntHistogram.CDF windowSample ) {
```
SHuang-Broad: Could be `static`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindSmallIndelRegions.java
```diff
@@ -40,44 +62,152 @@ public boolean requiresReads()
...
+            }
+
+            @Override
+            public SVTypeLen read( final Kryo kryo, final Input input, final Class<SVTypeLen> klass ) {
+                return new SVTypeLen(kryo, input);
+            }
         }
     }
 
+    private static SVIntervalTree<SVTypeLen> readTruthFile( final String truthFile,
```
SHuang-Broad: Currently in master, the VCF output is guaranteed to have a SVLEN field. But when it comes the BND and IMPRECISE records, no such guarantee can be made (BND records shouldn't even have a `SVLEN` IMO).

SHuang-Broad:
@tedsharpe Looked at the new changes. Very minor suggestions. Deal with them as you like.

PR #3377, opened 2017-07-28
Add option to run Spark tests only

tomwhite:


lbergelson:
@tomwhite Could you update the part of the readme that lists the test type options to include the new one?  It's near the top of the test section.  Aren't you on vacation?

PR #3383, opened 2017-07-28
Turn off tranches plots if no output Rscript is specified (for annota…

ldgauthier:
…tion plots)

Should prevent VariantRecalibrator from failing in a docker without R.  I tested by building a new docker from the image with the NIO fix, adding a jar from this branch, then running the SNPSVariantRecalibratorCreateModel task from the joint calling pipeline.


ldgauthier:
@droazen Here's the fix, but I'm only about 99.6% sure the tests will pass, so don't merge yet.

PR #3385, opened 2017-07-31
fixed contamination calculation, added error bars to output

davidbenjamin:
@takutosato This corrects the math error you pointed out in your code review of the docs.  While we're at it, it also outputs the error bars according to the formula given in the docs.  I have tested it on an in silico contamination series and it improves results slightly.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.hellbender.tools.walkers.contamination;
 
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.commons.math3.stat.descriptive.rank.Median;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -12,6 +13,7 @@ import org.broadinstitute.hellbender.cmdline.programgroups.VariantProgramGroup;
 import org.broadinstitute.hellbender.tools.exome.HashedListTargetCollection;
 import org.broadinstitute.hellbender.tools.exome.TargetCollection;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.tools.walkers.mutect.FilterMutectCalls;
 
 import java.io.File;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
```diff
@@ -67,31 +69,32 @@
...
         final long totalReadCount = homAltSites.stream().mapToLong(PileupSummary::getTotalCount).sum();
         final long totalRefCount = homAltSites.stream().mapToLong(PileupSummary::getRefCount).sum();
 
         // if eg ref is A, alt is C, then # of ref reads due to error is roughly (# of G read + # of T reads)/2
         final long errorRefCount = homAltSites.stream().mapToLong(PileupSummary::getOtherAltCount).sum() / 2;
         final long contaminationRefCount = Math.max(totalRefCount - errorRefCount, 0);
-        final double proportionRefReads = (double) contaminationRefCount / totalReadCount;
-        final double proportionOfContaminatingReadsThatAreRef = homAltSites.stream().mapToDouble(PileupSummary::getRefFrequency).average().getAsDouble();
-        final double contamination = proportionRefReads / proportionOfContaminatingReadsThatAreRef;
+        final double totalDepthWeightedByRefFraction = homAltSites.stream()
```
takutosato: Eh fraction is probably just as good. You use fraction in the docs. Your call

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
```diff
@@ -67,31 +69,32 @@
...
+        final double standardError = Math.sqrt(contamination / totalDepthWeightedByRefFraction);
 
         logger.info(String.format("In %d homozygous variant sites we find %d reference reads due to contamination and %d" +
                         " due to to sequencing error out of a total %d reads.", homAltSites.size(), contaminationRefCount, errorRefCount, totalReadCount));
-        logger.info(String.format("We estimate the fraction of contaminating ref reads in these homAltSites to be %.3f." +
-                "Based on population data, we expect reference reads to comprise a fraction %.3f of all contaminating reads." +
-                "Therefore, we estimate a contamination of %.3f", proportionRefReads, proportionOfContaminatingReadsThatAreRef, contamination));
-
-        return contamination;
+        logger.info(String.format("Based on population data, we would expect %d reference reads in a contaminant with the same depths at these sites.", (long) totalDepthWeightedByRefFraction));
```
takutosato: as* these sites? Alternatively, maybe we can say that `totalDepthWeightedByRefFraction` is the expected number of ref reads if we had 100% contamination

PR #3386, opened 2017-07-31
Wrote wdl for making Mutect pon

davidbenjamin:
@LeeTL1220 Tested.  Works.


LeeTL1220:
There might still be an issue here for Firecloud, but I would still merge this.

davidbenjamin:
@LeeTL1220 I will merge now and open a separate ticket to update the multi sample wdl and all wdls that depend on it.

PR #3388, opened 2017-07-31
Use ReadClipper in BaseQualityClipReadTransformer

mwalker174:
The earlier version of BaseQualityClipReadTransformer clipped bases by just removing them from the bases and base qualities arrays, but did not adjust the cigar, coordinate, etc. (thanks to @SHuang-Broad for catching this in #3354).

This commit fixes this behavior by invoking the ReadClipper class and adds cigar checking to the unit test. It also makes some minor style changes to the transformer and test code.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
+++ b/src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
@@ -8,8 +8,8 @@ import org.broadinstitute.hellbender.utils.read.GATKRead;
 /**
  * Clips reads on both ends using base quality scores
  */
-public class BaseQualityClipReadTransformer implements ReadTransformer {
-    public static final long serialVersionUID = 1L;
+public final class BaseQualityClipReadTransformer implements ReadTransformer {
+    private static final long serialVersionUID = 1L;
     private int qTrimmingThreshold = 15;
 
     public BaseQualityClipReadTransformer(final int trim_thresh) {
@@ -35,6 +35,9 @@ public class BaseQualityClipReadTransformer implements ReadTransformer {
         return clipReadLeftEnd(readClippedRightEnd);
     }
 
+    /**
+     * Returns right clip point or -1 if no clip
+     */
     public int getRightClipPoint( final byte[] quals ) {
         int clipSum = 0, lastMax = -1, clipPoint = -1; // -1 means no clip
         final int readLength = quals.length;
@@ -48,6 +51,9 @@ public class BaseQualityClipReadTransformer implements ReadTransformer {
         return clipPoint;
     }
 
+    /**
+     * Returns left clip point or -1 if no clip
+     */
     public int getLeftClipPoint( final byte[] quals ) {
         int clipSum = 0, lastMax = -1, clipPoint = -1; // -1 means no clip
         final int readLength = quals.length;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -29,8 +30,8 @@ public BaseQualityClipReadTransformer(int trim_thresh) {
      *
      */
     @Override
-    public GATKRead apply(GATKRead read) {
-        GATKRead readClippedRightEnd = clipReadRightEnd(read);
+    public GATKRead apply(final GATKRead read) {
+        final GATKRead readClippedRightEnd = clipReadRightEnd(read);
         return clipReadLeftEnd(readClippedRightEnd);
     }
```
SHuang-Broad: I know this exists before this PR, but while you are at it, could you add a documentation to `getRightClipPoint(byte[])` and `getLeftClipPoint(byte[])` explicitly saying that a negative return value means no clipping necessary?

SHuang-Broad: And, while looking at these two functions, they are very similar. 
Considering that even though it is public but seemingly only this class itself is using it, I suggest the following change, but it is totally up to you:

```java
    @override
    public GATKRead apply(final GATKRead read) {
        return clipRead(clipRead(read, false), true);
    }

    private int getClipPoint( final byte[] quals, final boolean fromLeft ) {
        final int readLength = quals.length;
        final int start, end, step;
        if (fromLeft) {
            start = 0;              end = readLength; step = 1;
        } else {
            start = readLength - 1; end = -1;         step = -1;
        }
        int clipSum = 0, lastMax = -1, clipPoint = -1; // -1 means no clip
        for (int i = start; i != end; i+= step) {
            clipSum += (qTrimmingThreshold - quals[i]);
            if (clipSum >= 0 && (clipSum >= lastMax)) {
                lastMax = clipSum;
                clipPoint = i;
            }
        }
        return clipPoint;
    }
    
    private GATKRead clipRead(final GATKRead read, final boolean fromLeft) {
        final byte[] quals = read.getBaseQualities();
        final int clipPoint = getClipPoint(quals, fromLeft);
        if (clipPoint != -1) {
            final ReadClipper readClipper = new ReadClipper(read);
            readClipper.addOp(fromLeft ? new ClippingOp(0, clipPoint) : new ClippingOp(clipPoint, read.getLength()));
            return readClipper.clipRead(ClippingRepresentation.HARDCLIP_BASES);
        } else {
            return read;
        }
    }
    
```

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
```diff
@@ -1,17 +1,18 @@
...
+import org.broadinstitute.hellbender.utils.clipping.ClippingRepresentation;
+import org.broadinstitute.hellbender.utils.clipping.ReadClipper;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
 
-import java.util.Arrays;
-
 /**
  * Clips reads on both ends using base quality scores
  */
 public class BaseQualityClipReadTransformer implements ReadTransformer {
```
SHuang-Broad: I forgot to mention that you probably intended this class to be `final` too, and its `serialVersionUID` to be `private`.

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
mwalker174: Done

mwalker174: That was my original approach, but I think it's easier to have two simple functions instead of one complicated function in this case.

src/main/java/org/broadinstitute/hellbender/transformers/BaseQualityClipReadTransformer.java
mwalker174: Yes - done.

Issue #3166, opened 2017-06-26
System.exit in Main.java lead to a failure in yarn cluster mode


popboy126:
Line 154(System.exit(0)) in Main.java will lead a unormal exitcode(143) for a container in yarn cluster mode. Remove this will get the correct output.
` final Object result = runCommandLineProgram(program, args);
            handleResult(result);
            //System.exit(0);`

lbergelson:
This is a duplicate of https://github.com/broadinstitute/gatk/issues/2666.

Issue #2666, opened 2017-05-05
Shutdown hook called before final status was reported.


david-wb:
Hi, 
I am experimenting with submitting a PrintReadsSpark job to a yarn spark cluster in AWS. I run the job with the following command.

```
spark-submit --deploy-mode cluster --class org.broadinstitute.hellbender.Main --deploy-mode cluster --master yarn gatk-package-4.alpha.2-248-gcd449bf-SNAPSHOT-spark.jar PrintReadsSpark -I hdfs://chr1.bam -O hdfs://output.bam
```

I can see from the output files that the job finished successfully, however the cluster tells me that it failed. It shows the following error message:

```
17/05/05 06:03:53 INFO ApplicationMaster: Final app status: FAILED, exitCode: 16, (reason: Shutdown hook called before final status was reported.)
```

I believe this may be due to the `System.exit(0)` statement at line 144 in hellbender.Main, though I am not sure. 

Here is a more complete snippet from the stderr log. 

```
17/05/05 06:03:50 INFO FileOutputCommitter: File Output Committer Algorithm version is 1
17/05/05 06:03:52 WARN DFSClient: Caught exception 
java.lang.InterruptedException
	at java.lang.Object.wait(Native Method)
	at java.lang.Thread.join(Thread.java:1249)
	at java.lang.Thread.join(Thread.java:1323)
	at org.apache.hadoop.hdfs.DFSOutputStream$DataStreamer.closeResponder(DFSOutputStream.java:609)
	at org.apache.hadoop.hdfs.DFSOutputStream$DataStreamer.endBlock(DFSOutputStream.java:370)
	at org.apache.hadoop.hdfs.DFSOutputStream$DataStreamer.run(DFSOutputStream.java:546)
17/05/05 06:03:52 INFO FileOutputCommitter: Saved output of task 'attempt_20170505060336_0011_r_000001_0' to hdfs://ip-172-30-0-86.ec2.internal:8020/output.bam.parts/_temporary/0/task_20170505060336_0011_r_000001
17/05/05 06:03:52 INFO Executor: Finished task 1.0 in stage 2.0 (TID 3). 1921 bytes result sent to driver
17/05/05 06:03:52 INFO TaskSetManager: Finished task 1.0 in stage 2.0 (TID 3) in 10260 ms on localhost (executor driver) (1/4)
[...]

PR #3400, opened 2017-08-01
removing the explicit System.exit(0) from Main

lbergelson:
this fixes the yarn early termination bug reported in #2666 and #3166

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -151,7 +151,6 @@ public class Main {
         try {
             final Object result = runCommandLineProgram(program, args);
             handleResult(result);
-            System.exit(0);
         } catch (final CommandLineException e){
             System.err.println(program.getUsage());
             handleUserException(e);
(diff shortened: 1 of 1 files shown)


lbergelson:
@droazen @magicDGS Can any of you think of a reason this would cause problems?  It seems like it should just continue and exit with 0 unless there's a non-daemon thread holding it open which is either something deliberate that should stay open until it's done or a bug.

magicDGS:
For me this is a 👍 (I will complain as always if it does not work for me, no worries), because I think that this should work.

Just in case, it will be useful to add a documentation note in `handleResult()` suggesting that if it needs to be explicitly call it should be added there.

droazen:
@lbergelson Is there a way we can test this fix?

lbergelson:
@david-wb @popboy126 I'm having trouble reproducing the shutdown issue on our own cluster, I sometimes get the message `Shutdown hook called before final status was reported.` but the job status is SUCCEEDED.  This happens if I run with the System.exit(0) or not.  Could one of you test with this branch and let me know if it solves your issue?  I think my cluster configuration must be different then yours in some way.

lbergelson:
I haven't been able to reproduce the issue myself, and I haven't heard back from either of the people who reported this issue, but it seems likely that this will solve it so I'm going to merge and they can reopen if they see it again.

PR #3401, opened 2017-08-02
Update to Intel GKL version 0.5.5

erniebrau:
GKL 0.5.5 fixes a critical bug in the FPGA PairHMM implementation. The bug produces a segmentation fault when FPGA PairHMM is invoked in 'private' mode. GKL 0.5.5 is otherwise identical to 0.5.3.


lbergelson:
@erniebrau thank you

PR #3403, opened 2017-08-02
Upgrade to htsjdk 2.11.0. Make TargetCodec indexable.

cmnbroad:
TargetCodec is dependent on the old htsjdk indexing scheme whereby the indexer called readActualHeader on the codec first (for the side effect of initializing the codec header state), and then manually processed the feature file contents by:

- re-creating the input SOURCE/stream a second time
- NOT calling readActualHeader
- extracting and passing the features one at a time to the codec's decode method, using the stream position to find the feature file offsets

Although this scheme worked with TargetCodec, it had several other failure modes (see https://github.com/samtools/htsjdk/pull/906). With https://github.com/samtools/htsjdk/pull/906, the SOURCE/stream is only opened once for indexing. However, TargetCodec uses an underlying CSVReader that automatically buffers input, which confounds the indexer. This PR works around that issue for indexing.

Note: this won't compile until there is an htsjdk snapshot available with https://github.com/samtools/htsjdk/pull/906.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/codecs/TargetCodec.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/codecs/TargetCodec.java
@@ -1,16 +1,20 @@
 package org.broadinstitute.hellbender.utils.codecs;
 
+import htsjdk.samtools.util.BlockCompressedInputStream;
+import htsjdk.samtools.util.LocationAware;
 import htsjdk.tribble.AbstractFeatureReader;
 import htsjdk.tribble.AsciiFeatureCodec;
-import htsjdk.tribble.readers.LineIterator;
+import htsjdk.tribble.readers.*;
 import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.tools.exome.Target;
 import org.broadinstitute.hellbender.tools.exome.TargetTableReader;
-
+import org.broadinstitute.hellbender.utils.tsv.TableUtils;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.URI;
-import java.net.URISyntaxException;
+import java.util.Collections;
+import java.util.Iterator;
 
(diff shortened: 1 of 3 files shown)


cmnbroad:
@droazen @lbergelson This is what we need to do to be able to take and htsjdk version that includes https://github.com/samtools/htsjdk/pull/906. Please take a look when you get a chance and see what you think - I'm waiting to merge https://github.com/samtools/htsjdk/pull/906 until we make sure we're ok with this, since its quite possible there are other codecs in the wild that have similar issues. I don't see any other good options.

src/main/java/org/broadinstitute/hellbender/utils/codecs/TargetCodec.java
```diff
@@ -34,9 +38,63 @@ public Target decode(final String line) {
...
+     * The rest of the indexing process doesn't delegate the remaining input stream contents to the codec (the
+     * indexer parses the stream manually, and passes the resulting features to the codec's decode method), so
+     * the features are then consumed and indexed with the correct stream offsets.
+     *
+     * @param inputStream stream to be indexed
+     * @return An AsciiLineReaderIterator implementation.
+     */
+    @Override
+    public LocationAware makeIndexableSourceFromStream(final InputStream inputStream) {
+        if (inputStream instanceof BlockCompressedInputStream) {
```
lbergelson: The fact that we have to duplicate this logic here makes me lean towards Magic's suggestion of creating a factory method in AsciiLineReader that encapsulates this in one place.

cmnbroad: Fair enough that. Done.

lbergelson:
@cmnbroad Htsjdk 2.11.0 is on it's way to maven.

cmnbroad:
@lbergelson Not sure if you're finished reviewing this one or not. This one has the htsjdk upgrade needed to re-enable IndexFeatureFile, so this one needs to go in first.

PR #3405, opened 2017-08-03
Add option to skip pre-Bwa repartitioning in PSFilter

mwalker174:
The PathSeq filter repartitioning the reads directly before the BWA host filtering step to even out the load on each partition. This is helpful for typical samples with low pathogen abundance, when ~90% of the reads are filtered before this step. 

However, in some sample types such as stool, saliva, or environmental samples, one can have a large number of reads (i.e. ~10M) at this stage so the cost of the repartition shuffle outweighs the benefit of load balancing for the slow BWA step. 

This PR exposes a tool argument to skip the repartitioning.


mwalker174:
Added an in-depth explanation in a comment block. 

Will work on a new PR to do this for other arguments.

PR #3406, opened 2017-08-03
Implement PathSeq taxon hit scoring in Spark

mwalker174:
Upgrades PathSeqScoreSpark to perform abundance score calculations on the executors rather than the driver. This was crashing on inputs with a lot of pathogen reads. 

This also required some minor changes to the `PSPathogenTaxonScore` class to be able to keep track of abundance score contributions that come directly from hits to that taxon and those that are from the taxon's descendents.

As a result, some of the test output changed when using bitwise, exact checks on the output. So the tests now check for output equivalence, meaning parsing the scores table, checking that all the taxa are the same, and that the scores are equal to within some defined epsilon.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSScorer.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSScorer.java
@@ -5,10 +5,12 @@ import com.esotericsoftware.kryo.io.Input;
 import htsjdk.samtools.*;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.apache.spark.broadcast.Broadcast;
 import org.broadinstitute.hellbender.exceptions.UserException;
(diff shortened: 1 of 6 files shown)


tedsharpe:
On vacation--won't have a chance to look at this until next week. If Chris or others approve just go for it.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSScorer.java
```diff
@@ -305,43 +306,69 @@ public static boolean isValidAlignment(final Cigar cigar, final int numMismatche
...
             //Scores normalized by genome length and degree of ambiguity (number of hits)
             final Set<String> hitPathNodes = new HashSet<>(); //Set of all unique hits and ancestors
             for (final String taxId : hitTaxIds) {
+                if (!tree.hasNode(taxId) || tree.getLengthOf(taxId) == 0) {
+                    invalidIds.add(taxId);
+                    continue;
+                }
                 final Double score = SCORE_GENOME_LENGTH_UNITS * hit.numMates / (numHits * tree.getLengthOf(taxId));
-                sum += score;
+                //Git list containing this node and its ancestors
```
cwhelan: Git -> Get?

mwalker174: Done

PR #3416, opened 2017-08-07
Upgrade to gkl-0.5.6, and added log4j-1.2-api bridge to dependencies,…

erniebrau:
… as well as excluding log4j 1.x.

GKL 0.5.6 now uses the log4j 1.x API for logging, and we use the log4j-1.2-api bridge JAR to redirect to log4j2 implementation. See [here](https://logging.apache.org/log4j/2.0/faq.html#which_jars) for details. This change was made because GATK 3.x uses log4j 1.x, and users were reporting errors in the output. This release fixes those errors.

GATK 4 uses log4j2 and, in order to make the API compatible with the GKL, we need to add a dependency on the log4j-1.2-api bridge. Unfortunately, the log4j 1.X JAR is also brought in due to some transitive dependency from another package, which causes conflicts with the log4j-1.2-api bridge package. To solve that, we need to exclude log4j 1.X from the dependencies, and let log4j-1.2-api take care of any calls to the log4j 1.X API, redirecting them to the log4j2 implementation. See [here](https://logging.apache.org/log4j/2.0/faq.html#exclusions) for details.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,7 +78,6 @@ configurations.all {
     }
     all*.exclude group: 'org.slf4j', module: 'slf4j-jdk14' //exclude this to prevent slf4j complaining about to many slf4j bindings
     all*.exclude group: 'com.google.guava', module: 'guava-jdk5'
-    all*.exclude group: 'log4j' //exclude log4j (1.x) and let log4j-1.2-api from log4j2 handle it
 }
 
 
@@ -133,7 +132,6 @@ dependencies {
     compile 'com.google.cloud.bigdataoss:gcs-connector:1.6.1-hadoop2'
     compile 'org.apache.logging.log4j:log4j-api:2.3'
     compile 'org.apache.logging.log4j:log4j-core:2.3'
-    compile 'org.apache.logging.log4j:log4j-1.2-api:2.3'
     compile 'org.apache.commons:commons-lang3:3.4'
     compile 'org.apache.commons:commons-math3:3.5'
     compile 'org.apache.commons:commons-collections4:4.1'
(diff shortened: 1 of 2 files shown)


lbergelson:
@erniebrau I don't think we can exclude log4j 1.x.  A lot of our dependencies use it.  This branch is failing with various ClassNotFound errors.

ex: https://storage.googleapis.com/hellbender/test/build_reports/11489.3/tests/test/classes/org.broadinstitute.hellbender.engine.spark.datasources.ReadsSparkSourceUnitTest.html#testReadFromFileAndHDFS

erniebrau:
@lbergelson You are correct. Let us do a bit more research to see if we can get GKL to log properly to both GATK 3 and 4. If not, we might have to go the route of having a GKL specific to GATK 3. Can we leave this PR open for now?

erniebrau:
OK. I found a potential solution. For this solution, we do not need to add or remove any dependencies. The only change is to the `log4j.properties` file (which configures log4j 1.x) to match the config specified in `log4j2.xml` (which configures log4j2). Now, GKL will use log4j 1.x to log, but the format will match the rest of  GATK, which uses log4j2.

This means that we have only one GKL for both GATK 3 and 4, at the expense of having to keep to config files, `log4j.properties` and `log4j2.xml`, in sync (which they probably should have been anyway, thought they weren't).

lbergelson:
So we typically override the config files on the command line.  We'll have to make sure we wire the log4j 1.x logger to respect our command line overrides if it doesn't already.  You can check that by testing if you can control the log output with the --verbosity command.  If not we'll have to update `LoggingUtils.setLoggingLevel()`

lbergelson:
I have to say, I loathe the logger ecosystem...  So much complexity for such boring results.

erniebrau:
@lbergelson As you foreshadowed, the `--versbosity` option does not affect log4j 1.x code, which means that, with the changes in 0.5.6, GKL will not listen to this option.

I think it is time to close this PR, since any fix to this issue, if one exists, will require a new release of GKL.

Thoughts?

PR #3418, opened 2017-08-07
Added a .dockerignore file to the gatk

### LINKED ISSUE #3414: Remove test resources from the docker image
Right now the docker image is too large. It appears that the if the build_docker.sh script its set to run the tests, then it is uploading the test resources into the docker and removing them, resulting in multiple layers which uses unneeded space. We need to remove the resources files from the docker.

jamesemery:
This reduces the size of the docker image from ~9GB when LFS tests were being run to 2.74GB

fixes #3414

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+# These are files that for whatever reason we don't want to include in our distribution docker images
+src/test/resources
+src/test/resources/*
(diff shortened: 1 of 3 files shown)


droazen:
@LeeTL1220 Can you review?

Issue #569, opened 2015-06-16
Validate all existing test BAMs


lbergelson:
We've discovered a number of bam files being used in tests which are not valid BAM files.  We should go through all the checked in BAMs, validate them, and replace broken ones.  (Except ones that are intentionally broken for testing.)

nenewell:
I'll work on this.

akiezun:
@nenewell go for it. it's yours. I don't seem to be able to 'assign' it to you.

lbergelson:
@nenewell I took a brief look yesterday and it looks like there are many bams that fail to validate due to errors with mate pairs.  I'm not sure how concerned we are about those errors.  It's the weirder ones like 

```
htsjdk.samtools.SAMException: SAMFormatException on record 01
	at htsjdk.samtools.SamFileValidator.validateSamRecordsAndQualityFormat(SamFileValidator.java:308)
	at htsjdk.samtools.SamFileValidator.validateSamFile(SamFileValidator.java:199)
	at htsjdk.samtools.SamFileValidator.validateSamFileVerbose(SamFileValidator.java:159)
	at org.broadinstitute.hellbender.tools.picard.sam.ValidateSamFile.doWork(ValidateSamFile.java:129)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:97)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:150)
	at org.broadinstitute.hellbender.cmdline.PicardCommandLineProgram.instanceMain(PicardCommandLineProgram.java:51)
	at org.broadinstitute.hellbender.Main.instanceMain(Main.java:71)
	at org.broadinstitute.hellbender.Main.main(Main.java:86)
Caused by: htsjdk.samtools.SAMFormatException: Error parsing text SAM file. Not enough fields; File src/test/resources/org/broadinstitute/hellbender/tools/picard/analysis/CompareMetrics/colqualyldmtcs.bam; Line 1
```

That seem particularly concerning.

lbergelson:
I did this yesterday but didn't look into all the output.

```
find src/test/resources -name "*.bam" > bams

for x in $(cat bams):
do build/install/hellbender/bin/hellbender ValidateSamFile -I $x  >> validateBamOutput 2>&1
done
```

output is here:
https://gist.github.com/lbergelson/002289ce94b0eace183c

lbergelson:
[...]

PR #3421, opened 2017-08-08
updating to htsjdk snapshot 2.10.1-26-g9c6c364-20170731.190658-1

### LINKED ISSUE #3417: update to a current htsjdk snapshot
update htsjdk to a current snapshot and fix the test failures

lbergelson:
updating tests
closes #3417

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ repositories {
     mavenLocal()
 }
 
-final htsjdkVersion = System.getProperty('htsjdk.version','2.10.1')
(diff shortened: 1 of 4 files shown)


lbergelson:
@cmnbroad please review (assuming there aren't any test failures that I missed in my local testing)

lbergelson:
@cmnbroad Someone once actually fixed all the bam files so they were valid and we rejected it because people were afraid that the fixes were not going to maintain the test conditions properly.  So my guess is no, we will never do #569.

PR #3422, opened 2017-08-08
Fix MuTect2 INFO field types in the VCF header

DonFreed:
The MuTect2 headers ECNT, TLOD and NLOD are currently strings. It is more accurate to describe these fields as integer, float and float.


DonFreed:
This should fix chapmanb/bcbio-nextgen#2033 with GATK4

lbergelson:
@davidbenjamin This looks like one for you.

davidbenjamin:
@DonFreed I appreciate it!

Issue #3306, opened 2017-07-19
Failure using samtools-generated .idx/.dict when reference has ambiguity code "."


cmnbroad:
This issue is for tracking https://github.com/samtools/samtools/issues/704.

The samtools-generated .idx and .dict entries for a reference sequence that contains a "." have different lengths (the .dict appears to not count the "."), and the MD5s are different. This causes GATK to throw:

> A USER ERROR has occurred: Couldn't read file /Users/cnorman/projects/htsjdk/src/test/resources/htsjdk/samtools/cram/amb.fa. Error was: Index length does not match dictionary length for contig: iupac with exception: Index length does not match dictionary length for contig: iupac

and causes failures when reading a CRAM file using that reference. It works fine if the .dict is regenerated using  GATK.

Issue #3154, opened 2017-06-22
CRAM reading gives MD5 mismatch with correct reference that samtools can decram fine


sooheelee:
This **CRAM file** and **its reference** are identical to those I have in the cloud because I uploaded these exact files to the cloud after downloading them from the 1000 Genomes Project FTP site. Back then, in the cloud, I was able to use samtools to decram and index this CRAM file alongside 39 others.

On our local server, I cannot get readwalkers PrintReads nor CalculateTargetCoverage to correctly decipher the CRAM. Both tools give the same error.

Here is the PrintReads command:
```
/humgen/gsa-hpprojects/GATK/gatk4/gatk-4.alpha.2-1134-ga9d9d91-SNAPSHOT/gatk-launch \
  PrintReads \
  -R /humgen/gsa-hpprojects/dev/shlee/ref/GRCh38_1kg/GRCh38_full_analysis_set_plus_decoy_hla.fa \
  -I /humgen/gsa-hpprojects/dev/shlee/1kg_GRCh38_exome/cram/HG02759.alt_bwamem_GRCh38DH.20150826.GWD.exome.cram \
  -O HG02759.alt_bwamem_GRCh38DH.20150826.GWD.exome.bam
```
And here is the error:
```
17:47:15.362 INFO  ProgressMeter -       chr1:198467627              2.6               8432000        3202552.3
17:47:25.402 INFO  ProgressMeter -       chr1:236860077              2.8              10019000        3577916.1
ERROR   2017-06-22 17:47:27     Slice   Reference MD5 mismatch for slice 0:248681942-248858764, ATAGCGGTCA...AGTGGCGGTG
17:47:27.292 INFO  CalculateTargetCoverage - Shutting down engine
[June 22, 2017 5:47:27 PM EDT] org.broadinstitute.hellbender.tools.exome.CalculateTargetCoverage done. Elapsed time: 2.87 minutes.
Runtime.totalMemory()=10377756672
htsjdk.samtools.cram.CRAMException: Reference sequence MD5 mismatch for slice: sequence id 0, start 248681942, span 176823, expected MD5 4b8526e90896b01860301e5a1ef4988b
        at htsjdk.samtools.CRAMIterator.nextContainer(CRAMIterator.java:187)
        at htsjdk.samtools.CRAMIterator.hasNext(CRAMIterator.java:261)
        at htsjdk.samtools.SamReader$AssertingIterator.hasNext(SamReader.java:601)
[...]

PR #3430, opened 2017-08-10
Test to validate that CRAM MD5 slice calculation matches samtools.

cmnbroad:
Added a test to verify that https://github.com/broadinstitute/gatk/issues/3154 is fixed now that we've upgraded htsjdk (though we should keep that ticket open until @sooheelee can verify her particular incarnation of this issue).

Also note that while the CRAM MD5 slice calculation is fixed, GATK users can still have problem reading CRAMs made from references containing ambiguity codes if the .dict accompanying the reference was generated with samtools. This is tracked by https://github.com/broadinstitute/gatk/issues/3306, but is really a samtools issue. The simple workaround is to recreate the .dict using CreateSequenceDictionary, which is what I've done to create the test in this PR.


sooheelee:
I've tested my scenario and made notes in <https://github.com/broadinstitute/gatk/issues/3154>. I used `gatk-4.beta.5` to test. The issue is solved. Thank you for the fix.

PR #3433, opened 2017-08-11
Pair HMM docs

davidbenjamin:
@ronlevine here are the pair HMM docs I promised.  Do you want to review officially?


ronlevine:
@davidbenjamin I'll be a reviewer. I think you should also include @vruano, @sooheelee, and @gspowley (for the Intel implementation).

docs/pair_hmm.tex
```diff
@@ -0,0 +1,101 @@
...
+
+The read-alignment likelihood has two components.  First is the probability of the sequence of match, insertion, and deletion states, which is
+\begin{equation}
+P(\mc{A}) = \prod_k T_{s_k, s_{k+1}},
+\end{equation}
+where $T$ is a constant matrix of state transition probabilities\footnote{The gap continuation probability corresponds to a phred-scaled quality that is set by the \code{gcpHMM} argument, which is 10 by default, implying that $T_{DD} = T_{II} = 0.1$ and $T_{DM} = T{IM} = 0.9$.  The other transitions are hard-coded based on a phred-scaled indel start quality of 45.  That is, $T_{MD} = T{MI} = 10^{-4.5}$ and $T_{MM} = 1 - 2 \times 10^{-4.5}$.  This fact is somewhat obscured in the code, which gives a position index to the transition matrices.  Nonetheless, the array representing transitions is constant over this index.}.  Next is the emission probability of the read bases given the haplotype bases they align to and the base qualities:
```
ronlevine: T{IM} should be T_{IM} and T{MI} should be T_{MI}. How is T_{II} = 0.1? is it 1/10? Would it be clearer to have $T_{DM} = T{IM} = 1 - T_{II}?

docs/pair_hmm.tex
```diff
@@ -0,0 +1,101 @@
...
+\begin{itemize}
+\item $(i, j, M/D/I) \rightarrow (i+1, j+1, M)$: match of read position $i+1$ with haplotype position $j + 1$
+\item $(i, j, M/D) \rightarrow (i, j+1, D)$: deletion after read position $i$ -- haplotype position $j+1$ is deleted
+\item $(i, j, M/I) \rightarrow (i + 1, j, I)$: insertion at read position $i + 1$
+\end{itemize}
+Note that the state label $s$ seems redundant because it can be reconstructed from the sequence of $i$ an $j$.  While this is true, our model treats indel starts differently from indel continuations, and by including the state label we can distinguish these conveniently\footnote{That is, we can tell which type of transition it is by looking back one unit instead of two.  Thus we have a first-order Markov model instead of a second-order Markov model.}.
+
+The read-alignment likelihood has two components.  First is the probability of the sequence of match, insertion, and deletion states, which is
+\begin{equation}
+P(\mc{A}) = \prod_k T_{s_k, s_{k+1}},
```
ronlevine: What is k for the product? What are the max/min values?

docs/pair_hmm.tex
```diff
@@ -0,0 +1,101 @@
...
+{\rm alignment} = \{ (i_1, j_1, s_1), (i_2, j_2, s_2) \ldots (i_N, j_N, s_N) \},
+\end{equation}
+where $i$ and $j$ represent positions within the read and haplotype, respectively, and $s_n \in \{ M, I, D \}$ represents the states of match, insertion, and deletion of the read relative to the haplotype.  For example, an alignment $\{ (1,10, M), (2,11, M), (3,12, M), (4,13, M), (4,14, D), (4,15, D), (5,16, M), (6,17, M), (7,17, I), (8,18, M) \}$ means that positions 1 - 4 of the read match positions 10 - 13 of the haplotype, followed by a two-base deletion (advancing from 13 to 15 in the haplotype without advancing in the read), followed by a match of read positions 5 - 6 with haplotype positions 16 - 17, followed by an insertion at read position 7, followed by a match.  The allowable transitions $(i_n, j_n) \rightarrow (i_{n+1}, j_{n+1})$ are
+\begin{itemize}
+\item $(i, j, M/D/I) \rightarrow (i+1, j+1, M)$: match of read position $i+1$ with haplotype position $j + 1$
+\item $(i, j, M/D) \rightarrow (i, j+1, D)$: deletion after read position $i$ -- haplotype position $j+1$ is deleted
+\item $(i, j, M/I) \rightarrow (i + 1, j, I)$: insertion at read position $i + 1$
```
ronlevine: It would be nice to have an image like the "Example Model" in  https://web.stanford.edu/class/cs262/archives/notes/lecture8.pdf.

docs/pair_hmm.tex
ronlevine: If I am not mistaken, `The other transitions are hard-coded based on a phred-scaled indel start quality of 45` is not entirely correct. It is true if the read does not have BQSR Base Insertion (BI) or Deletion (BD) Quality values. See `ReadUtils.getBaseInsertionQualities` and `ReadUtils.getBaseDeletionQualities`.

docs/pair_hmm.tex
vruano: I sent this detailed hand-made diagram to @dekling a while back, perhaps of some use here.

docs/pair_hmm.tex
vruano: @ronlevine is correct.

ldgauthier: Agreed.  We don't do insertion or deletion qualities in production because they make the bams too big.

docs/pair_hmm.tex
```diff
@@ -0,0 +1,101 @@
...
+		\State Calculate $D_{ij}$ via Equation \ref{del}.
+	\EndFor
+\EndFor
+\State Total likelihood $P(\mc{R} | \mc{H})$ is $\sum_j \left(M_{\mc{R},j} + I_{\mc{R},j} \right)$.
+\end{algorithmic}
+\caption{Pair HMM algorithm}
+\label{pairHMM}
+\end{algorithm}
+
+That the $i = 0$ rows of $M$ and $I$ are initialized to zero corresponds to starting at an imaginary position one base before the read start in a deletion state.  The factor of $1/|\mc{H}|$ corresponds to a flat prior on which $j$ an alignment starts at.  The curious factor of $2^{1020}$ is a huge number to prevent underflow -- all multiplications are by numbers less than 1, so we needn't worry about overflow -- which is a much more efficient approach than performing the computation in log space.  The omission of $D$ from the returned value recognizes the fact that a terminal deletion is meaningless.
```
ldgauthier: I would phrase this differently by pointing out that those boundary conditions are set that way because this is a local alignment and we don't want a penalty associated with where the read starts with respect to the beginning of the haplotype.

docs/pair_hmm.tex
```diff
@@ -0,0 +1,101 @@
...
+
+Finally, we note a shortcut that the GATK exploits: when two consecutive haplotypes agree up to the $k$th position, the first $k$ columns of $M$, $D$ and $I$ are recycled and the inner loop is over $k < j \le \mc{H}$.
```
davidbenjamin: > Only between consecutive haplotypes with the same length. . .

Done.

> We need to note that there is a asymmetry. . .

Done.

docs/pair_hmm.tex
```diff
@@ -0,0 +1,101 @@
...
+\date{\today}
+
+\begin{abstract}
+After generating candidate haplotypes, the GATK tools \HC~and \Mutect~realign reads against these haplotypes to obtain a matrix of likelihoods for each read to be derived from each haplotype.  Here we describe the probabilistic model specifying this likelihood as well as its computational implementation.  We do not describe the translation of this implementation into native code optimized for vectorized architectures.
+\end{abstract}
+
+\maketitle
+
+\section{The Pair HMM model}
+We want to calculate the probability $P( \mc{R} | \mc{H})$ of read $\mc{R}$ to be sequenced from haplotype $\mc{H}$, where the haplotypes are sufficiently long that reads are contained within them.  This likelihood is the sum of likelihoods of all possible alignments $\mc{A}$ of $\mc{R}$ to $\mc{H}$:
```
davidbenjamin: This seems okay to me because the no-mismatch read will (I hope) have a good likelihood to align to the ref haplotype as well, and thus any model with a prior against the alt should not consider this read as evidence for the deletion.

docs/pair_hmm.tex
davidbenjamin: done

docs/pair_hmm.tex
davidbenjamin: done on all counts

docs/pair_hmm.tex
davidbenjamin: It's 1 to read length + number of deletions, since the latter is the total number of CIGAR elements.  This is now explicit.

docs/pair_hmm.tex
davidbenjamin: done

davidbenjamin:
@ronlevine @ldgauthier @vruano Thanks for the comments.  I addressed all suggestions.

docs/pair_hmm.tex
ronlevine: It looks like this change didn't make it in.

docs/pair_hmm.tex
```diff
@@ -0,0 +1,101 @@
...
+\begin{equation}
+P(\mc{R} | \mc{A}, \mc{H}) = \prod_k P(r_{i_k} | h_{j_k}, q_{i_k})^{{\rm I}[s_k = M]},
+\end{equation}
+where $r_m$ and $h_n$ are the $m$th read bases and $n$th haplotype base and $q_m$ is the quality of the read's $m$th base.  Note that this only includes alignments in the match state.  The per-base emission is given directly from the definition of base quality:
+\begin{equation}
+P(b_2 | b_1, q) = \left\{ \begin{array}{cc} \epsilon(q)/3 & (b_1 \ne b_2) \\ 1 - \epsilon(q) & (b_1 = b_2)  \end{array} \right.
```
ronlevine: What is epsilon? I don't think it's the same epsilon in the HMM diagram.

docs/pair_hmm.tex
davidbenjamin: > It looks like this change didn't make it in.

Which one?  I wrote "done" before pushing the changes; maybe that's it?

ronlevine: Yes. That's it.

docs/pair_hmm.tex
davidbenjamin: It's the error rate associated with the base quality, as I added just now.

docs/pair_hmm.tex
ronlevine: I'm not seeing T_{II} = 1/10 replacing of T_{II} = 0.1.

davidbenjamin: Whoops, I interpreted the 0.1 = 1/10 comment as being part of the`T{IM} = 1 - T_{II}` comment.  Now I have the explicit phred-scale calculation 10^{-10/10} = 1/10.

ronlevine: Thanks, that's much clearer.

docs/pair_hmm.tex
```diff
@@ -0,0 +1,109 @@
...
+\label{fig:fsm}
+\end{figure}
+
+The read-alignment likelihood has two components.  First is the probability of the sequence of match, insertion, and deletion states, which is
+\begin{equation}
+P(\mc{A}) = \prod_k T_{s_k, s_{k+1}},
+\end{equation}
+where $T$ is a matrix of state transition probabilities\footnote{The gap continuation probability corresponds to a phred-scaled quality that is set by the \code{gcpHMM} argument, which is 10 by default, implying that $T_{DD} = T_{II} = 10^{-10/10} = 1/10$ and $T_{DM} = T_{IM} = 1 - T_{II}$.  The indel start transitions are derived from the read's BQSR base insertion and base deletion qualities, if they exist, or a phred-scaled indel start quality of 45 if they do not.  In production at the Broad Institute bams do not have BQSR indel qualities, hence the constant default is used.  That is, $T_{MD} = T_{MI} = 10^{-4.5}$ and $T_{MM} = 1 - 2 \times 10^{-4.5}$.}.  The index $k$ runs from 1 to the number of states in the alignment, that is, the read length plus the number of deleted reference bases.  Next is the emission probability of the read bases given the haplotype bases they align to and the base qualities:
+\begin{equation}
```
ronlevine: The following legend can accompany the state transition diagram:
image
taken from https://docs.google.com/presentation/d/1hCxMjf1VF8eJRiKt5wQQAehhMkjUS-h1n5Ubr68nRGE/edit#slide=id.p30

davidbenjamin: I now include this information in the figure caption.

docs/pair_hmm.tex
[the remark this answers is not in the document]
ldgauthier: It's certainly not substantial evidence, no matter the prior.  But did you have any thoughts you wanted to add on why we sum?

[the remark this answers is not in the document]
davidbenjamin: Summing over alignments is marginalizing over all possible values of a latent variable.  If we believe the Pair HMM model then it's the mathematically correct thing to do and our hand is forced.  In the IGV screenshot above, if we didn't marginalize but instead including only the likeliest alignment, the bottom gray read would still be in an unsettling state, a tie between the ref and the deletion.  It's possible that marginalizing breaks the tie in favor of the deletion, but this is a perturbation.

From the point of view of likelihoods, however, this seems correct to me, because likelihoods (as opposed to posterior probabilities) answer the question: "*if* this allele is in fact present, does this read make sense?"  Given that the read is a perfect match to both haplotypes, the likelihoods should be basically equal.

docs/pair_hmm.tex
ronlevine: Why is`epsilon(q)/3`? Is dividing by 3 empirical?

docs/pair_hmm.tex
```diff
@@ -0,0 +1,109 @@
...
+\begin{algorithmic}[1]
+\State Initialize $M_{0,j} = I_{0,j} = 0$ and $D_{0,j} = 2^{1020}/|\mc{H}|$ for $1 \le j \le |\mc{H}|$.
+\For{$1 \le i \le | \mc{R} |$}
+	\For{$1 \le j \le | \mc{H} |$}
+		\State Calculate $M_{ij}$ via Equation \ref{match}.
+		\State Calculate $I_{ij}$ via Equation \ref{ins}.
+		\State Calculate $D_{ij}$ via Equation \ref{del}.
+	\EndFor
+\EndFor
+\State Total likelihood $P(\mc{R} | \mc{H})$ is $\sum_j \left(M_{\mc{R},j} + I_{\mc{R},j} \right)$.
```
ronlevine: Is the likelihood calculation missing the deletions (D) in the summation?

docs/pair_hmm.tex
```diff
@@ -0,0 +1,109 @@
...
+
+\begin{equation}
+P(\mc{R} | \mc{H}) = \sum_{\mc{A}} P(\mc{R}, \mc{A} | \mc{H})
+\end{equation}
+
+We represent alignments as sequences
+\begin{equation}
+{\rm alignment} = \{ (i_1, j_1, s_1), (i_2, j_2, s_2) \ldots (i_N, j_N, s_N) \},
+\end{equation}
+where $i$ and $j$ represent positions within the read and haplotype, respectively, and $s_n \in \{ M, I, D \}$ represents the states of match, insertion, and deletion of the read relative to the haplotype.  For example, an alignment $\{ (1,10, M), (2,11, M), (3,12, M), (4,13, M), (4,14, D), (4,15, D), (5,16, M), (6,17, M), (7,17, I), (8,18, M) \}$ means that positions 1 - 4 of the read match positions 10 - 13 of the haplotype, followed by a two-base deletion (advancing from 13 to 15 in the haplotype without advancing in the read), followed by a match of read positions 5 - 6 with haplotype positions 16 - 17, followed by an insertion at read position 7, followed by a match.  The allowable transitions $(i_n, j_n) \rightarrow (i_{n+1}, j_{n+1})$ are
```
ronlevine: I think it would be useful to say how the M, I and D are calculated. For example, something like the `Parameters` bullet in slide 40, `Deriving HMMs from multiple
alignments` in http://homes.soic.indiana.edu/classes/spring2017/info/i529-yye/lectures/ProfileHMM.pdf

docs/pair_hmm.tex
davidbenjamin: It's not missing because terminal deletions are not considered a valid alignment.  It's in the doc somewhere or other.

docs/pair_hmm.tex
davidbenjamin: @ronlevine Do we mean how the T matrix elements are derived?  The thing is, while you can learn them from Baum-Welch, as those slides described, we don't actually do that.  The parameters are just empirical constants for us.

ronlevine: It might be good to mention that they're empirical constants.

davidbenjamin: Done -- see the new footnote.

ronlevine:
@davidbenjamin 👍  LGTM

docs/pair_hmm.tex
davidbenjamin: @ronlevine the error probability is the probability for any of 3 possible substitution errors, so we divide by 3 to get the likelihood of any particular one.

PR #3445, opened 2017-08-15
ValidateVariants gvcf validation

jsotobroad:
Merge adjacent blocks when validating GVCFs so we use less memory and dont fail when not using an interval argument


src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -195,7 +201,15 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
         final Set<String> rsIDs = getRSIDs(featureContext);
 
         if (VALIDATE_GVCF) {
-            genomeLocSortedSet.add(genomeLocSortedSet.getGenomeLocParser().createGenomeLoc(ref.getInterval().getContig(), ref.getInterval().getStart(), vc.getEnd()), true);
+            SimpleInterval refInterval = ref.getInterval();
```
yfarjoun: finals

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -168,10 +168,16 @@
 
     private GenomeLocSortedSet genomeLocSortedSet;
 
+    // information to keep track of when validating a GVCF
+    private int currentEnd;
```
yfarjoun: instead of keeping the end and the contig, keep a currentInterval and use the available math functions to check if the next interval abuts the current.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -195,7 +201,15 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
         final Set<String> rsIDs = getRSIDs(featureContext);
 
         if (VALIDATE_GVCF) {
-            genomeLocSortedSet.add(genomeLocSortedSet.getGenomeLocParser().createGenomeLoc(ref.getInterval().getContig(), ref.getInterval().getStart(), vc.getEnd()), true);
+            SimpleInterval refInterval = ref.getInterval();
+            // Take advantage of adjacent blocks and just merge them so we dont have to keep so many objects in the set.
+            if (refInterval.getContig().equals(currentContig) && currentEnd == (refInterval.getStart() - 1)) {
```
yfarjoun: there's a Interval::abuts() function. please use it.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -211,7 +225,13 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
     @Override
     public Object onTraversalSuccess() {
         if (VALIDATE_GVCF) {
-            GenomeLocSortedSet intervalArgumentGenomeLocSortedSet = GenomeLocSortedSet.createSetFromList(genomeLocSortedSet.getGenomeLocParser(), IntervalUtils.genomeLocsFromLocatables(genomeLocSortedSet.getGenomeLocParser(), intervalArgumentCollection.getIntervals(getBestAvailableSequenceDictionary())));
+            GenomeLocSortedSet intervalArgumentGenomeLocSortedSet;
```
yfarjoun: final

jsotobroad:
back to you @yfarjoun

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -195,7 +200,14 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
         final Set<String> rsIDs = getRSIDs(featureContext);
 
         if (VALIDATE_GVCF) {
-            genomeLocSortedSet.add(genomeLocSortedSet.getGenomeLocParser().createGenomeLoc(ref.getInterval().getContig(), ref.getInterval().getStart(), vc.getEnd()), true);
+            final SimpleInterval refInterval = ref.getInterval();
+            // Take advantage of adjacent blocks and just merge them so we dont have to keep so many objects in the set.
+            if (previousInterval != null && previousInterval.overlapsWithMargin(refInterval, 1)) {
+                genomeLocSortedSet.add(genomeLocSortedSet.getGenomeLocParser().createGenomeLoc(refInterval.getContig(), previousInterval.getEnd(), vc.getEnd()), true);
```
yfarjoun: I'm confused ad to why you are adding an interval in both cases of the if statment. I thought that the logic should be

```
if( contiguous)
  merge currentInterval into previous one
else
  add previous to set 
  replace previous with current
end
```

Specifically, only one branch add the interval. 

ofcourse, you still need to remember to add at the end..of the traversal. Are you doing that?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -211,7 +223,13 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
     @Override
     public Object onTraversalSuccess() {
         if (VALIDATE_GVCF) {
-            GenomeLocSortedSet intervalArgumentGenomeLocSortedSet = GenomeLocSortedSet.createSetFromList(genomeLocSortedSet.getGenomeLocParser(), IntervalUtils.genomeLocsFromLocatables(genomeLocSortedSet.getGenomeLocParser(), intervalArgumentCollection.getIntervals(getBestAvailableSequenceDictionary())));
+            final GenomeLocSortedSet intervalArgumentGenomeLocSortedSet;
+
```
yfarjoun: need to remember to add the last interval here.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -195,7 +200,14 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
         final Set<String> rsIDs = getRSIDs(featureContext);
 
         if (VALIDATE_GVCF) {
-            genomeLocSortedSet.add(genomeLocSortedSet.getGenomeLocParser().createGenomeLoc(ref.getInterval().getContig(), ref.getInterval().getStart(), vc.getEnd()), true);
+            final SimpleInterval refInterval = ref.getInterval();
+            // Take advantage of adjacent blocks and just merge them so we dont have to keep so many objects in the set.
+            if (previousInterval != null && previousInterval.overlapsWithMargin(refInterval, 1)) {
```
yfarjoun: I think there's some pre-optimizing going on here. Please create `currentInterval` as a local final variable

Use it for checking overlap and for merging or replacing as needed.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
jsotobroad: the logic I was going for was always add the interval to the set, but use the `mergeIfIntervalOverlaps` in our favor the previous and current interval are adjacent

jsotobroad: specifically because i wanted all the set creation logic in one place and not have a straggler in `onTraversalSuccess`

yfarjoun: I like it. Could you add comments to that effect? (I wonder who didn't comment about this idea previously?!!)

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
yfarjoun: now that I understand what you were trying to do, I think that the comment is not clarifying. you need to state that we would like to merge abutting intervals, but that the genomeLocSortedSet only knows how to merge **overlapping** intervals.

yfarjoun: given that the difference between these two lines is so small, perhaps you should define the start with a ternary:

```java
final long start = (previousInterval != null && previousInterval.overlapsWithMargin(refInterval, 1)) ? 
previousInterval.getEnd() : refInterval.getStart()
```

jsotobroad:
Oh I really like that

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -168,10 +168,15 @@
...
+    // information to keep track of when validating a GVCF
+    private SimpleInterval previousInterval;
+
+    SAMSequenceDictionary dict;
+
     @Override
     public void onTraversalStart() {
         if (VALIDATE_GVCF) {
-            SAMSequenceDictionary dict = getBestAvailableSequenceDictionary();
+            dict = getBestAvailableSequenceDictionary();
```
lbergelson: I wouldn't store this, I would just call getBestAvailableSequenceDictionary() again when you need it again.   It's not used in any sort of tight loop, so the expense is negligible.  Adding fields should always be a last resort.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -195,7 +200,14 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
         final Set<String> rsIDs = getRSIDs(featureContext);
 
         if (VALIDATE_GVCF) {
-            genomeLocSortedSet.add(genomeLocSortedSet.getGenomeLocParser().createGenomeLoc(ref.getInterval().getContig(), ref.getInterval().getStart(), vc.getEnd()), true);
+            final SimpleInterval refInterval = ref.getInterval();
+
+            // Take advantage of adjacent blocks and just merge them so we dont have to keep so many objects in the set.
+            final int start = (previousInterval != null && previousInterval.overlapsWithMargin(refInterval, 1)) ?
+                    previousInterval.getEnd() : refInterval.getStart();
```
lbergelson: I'm not sure I understand this change, doesn't this still add the same number of elements to the set?  Does the set do internal merging somehow?  If so, why is doing the merging externally necessary?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -211,7 +223,13 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
     @Override
     public Object onTraversalSuccess() {
         if (VALIDATE_GVCF) {
-            GenomeLocSortedSet intervalArgumentGenomeLocSortedSet = GenomeLocSortedSet.createSetFromList(genomeLocSortedSet.getGenomeLocParser(), IntervalUtils.genomeLocsFromLocatables(genomeLocSortedSet.getGenomeLocParser(), intervalArgumentCollection.getIntervals(getBestAvailableSequenceDictionary())));
+            final GenomeLocSortedSet intervalArgumentGenomeLocSortedSet;
+
+            if (intervalArgumentCollection.intervalsSpecified()){
+                intervalArgumentGenomeLocSortedSet = GenomeLocSortedSet.createSetFromList(genomeLocSortedSet.getGenomeLocParser(), IntervalUtils.genomeLocsFromLocatables(genomeLocSortedSet.getGenomeLocParser(), intervalArgumentCollection.getIntervals(dict)));
+            } else {
+                intervalArgumentGenomeLocSortedSet = GenomeLocSortedSet.createSetFromSequenceDictionary(dict);
```
lbergelson: You should add an integration test that exercises this new path.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
jsotobroad: it does the same number of `.add()` calls but the number of elements in the set is smaller.  The set internally merges things that overlap with the `true` in the `.add()` call.  In a gvcf most records are adjacent to each other so nothing was being merged and we had a huge set.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
jsotobroad: 👍

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
jsotobroad: so if we notice the current interval is adjacent to the previous one, we "overlap" them before we add to the set

lbergelson: Ah, thank you for the explanation, that makes sense.  Could you add a more detailed comment explaining that, because i suspect the inner workings of the `GenomeLocSortedSet` are not well known to most people.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
jsotobroad: positive and negative tests added

jsotobroad:
@yfarjoun this is a memory optimization but I have confirmed it uses less memory and runs at the same speed

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -195,7 +198,17 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
         final Set<String> rsIDs = getRSIDs(featureContext);
 
         if (VALIDATE_GVCF) {
-            genomeLocSortedSet.add(genomeLocSortedSet.getGenomeLocParser().createGenomeLoc(ref.getInterval().getContig(), ref.getInterval().getStart(), vc.getEnd()), true);
+            final SimpleInterval refInterval = ref.getInterval();
+
+            // GenomeLocSortedSet will automatically merge intervals that are overlapping when setting `mergeIfIntervalOverlaps`
+            // to true.  In a GVCF most blocks are adjacent to each other so they wouldn't normally get merged.  We check
+            // if the current record is adjacent to the previous record and "overlap" them if they are so our set is as
+            // small as possible.
```
yfarjoun: "...while still containing the same bases."

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -168,15 +168,18 @@
...
     private GenomeLocSortedSet genomeLocSortedSet;
 
+    // information to keep track of when validating a GVCF
+    private SimpleInterval previousInterval;
+
     @Override
     public void onTraversalStart() {
         if (VALIDATE_GVCF) {
-            SAMSequenceDictionary dict = getBestAvailableSequenceDictionary();
+            SAMSequenceDictionary seqDictionary = getBestAvailableSequenceDictionary();
```
yfarjoun: final

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -211,7 +224,14 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
     @Override
     public Object onTraversalSuccess() {
         if (VALIDATE_GVCF) {
-            GenomeLocSortedSet intervalArgumentGenomeLocSortedSet = GenomeLocSortedSet.createSetFromList(genomeLocSortedSet.getGenomeLocParser(), IntervalUtils.genomeLocsFromLocatables(genomeLocSortedSet.getGenomeLocParser(), intervalArgumentCollection.getIntervals(getBestAvailableSequenceDictionary())));
+            final GenomeLocSortedSet intervalArgumentGenomeLocSortedSet;
+            SAMSequenceDictionary seqDictionary = getBestAvailableSequenceDictionary();
```
yfarjoun: final

yfarjoun:
I'm happy with this except for the missing `final`s. @lbergelson ?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -195,7 +198,17 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
         final Set<String> rsIDs = getRSIDs(featureContext);
 
         if (VALIDATE_GVCF) {
-            genomeLocSortedSet.add(genomeLocSortedSet.getGenomeLocParser().createGenomeLoc(ref.getInterval().getContig(), ref.getInterval().getStart(), vc.getEnd()), true);
+            final SimpleInterval refInterval = ref.getInterval();
+
+            // GenomeLocSortedSet will automatically merge intervals that are overlapping when setting `mergeIfIntervalOverlaps`
+            // to true.  In a GVCF most blocks are adjacent to each other so they wouldn't normally get merged.  We check
```
lbergelson: thanks, this is helpful!

PR #3106, opened 2017-06-13
Run MD+BQSR+HC pipeline on full genome

tomwhite:
Here are the changes needed to get the full pipeline running on WGS BAMs.

This contains commits from other PRs.

PR #3452, opened 2017-08-16
Extend ReadsPipelineSpark to run HaplotypeCallerSpark

tomwhite:
Based on https://github.com/broadinstitute/gatk/pull/3106, so not ready to be merged until after that PR is in.


scripts/spark_eval/exome_pipeline_single_gcs.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+# Run the pipeline (ReadsPipelineSpark) on exome data on a GCS Dataproc cluster. Data is in GCS.
+
+# TODO: change output to a GCS bucket when writing works again
```
lbergelson: I think writing is supposed to work now.

tomwhite: I tried it out, but unfortunately it doesn't :( I've created a fix here: #3485.

tomwhite:
@lbergelson can this go in now? I've opened a separate fix for writing variants to GCS, since it needs a Hadoop-BAM change too.

scripts/spark_eval/exome_pipeline_single_gcs.sh
lbergelson: Ah, sorry about that, wasn't thinking about variants when we did the bam fix.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -234,16 +266,15 @@ protected void runTool(final JavaSparkContext ctx) {
      *
      * This will be replaced by a parallel writer similar to what's done with {@link org.broadinstitute.hellbender.engine.spark.datasources.ReadsSparkSink}
      */
-    private void writeVariants(JavaRDD<VariantContext> variants, HaplotypeCallerEngine hcEngine) {
+    private static void writeVariants(String outputFile, JavaRDD<VariantContext> variants, HaplotypeCallerEngine hcEngine, SAMSequenceDictionary referenceDictionary, SAMSequenceDictionary readsDictionary) {
```
lbergelson: I think I had a comment about this in another PR but I can't find it again, so I'm sorry if this is redundant.  It seems like we probably don't need two different sequence dictionaries here.  Shouldn't we be able to use getBestAvailableSequenceDictionary() for both?  The fact that they're potentially different, seems problematic to me, probably a bug in the existing code.e

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
```diff
@@ -129,12 +134,17 @@ protected void runTool(final JavaSparkContext ctx) {
...
 
-        JavaPairRDD<GATKRead, ReadContextData> rddReadContext = AddContextDataToReadSpark.add(ctx, markedFilteredReadsForBQSR, getReference(), bqsrKnownVariants, baseRecalibrationKnownVariants, joinStrategy, getHeaderForReads().getSequenceDictionary(), readShardSize, readShardPadding);
+        JavaPairRDD<GATKRead, ReadContextData> rddReadContext = AddContextDataToReadSpark.add(ctx, markedFilteredReadsForBQSR, getReference(), bqsrKnownVariants, baseRecalibrationKnownVariants, joinStrategy, getHeaderForReads().getSequenceDictionary(), shardingArgs.readShardSize, shardingArgs.readShardPadding);
         final RecalibrationReport bqsrReport = BaseRecalibratorSparkFn.apply(rddReadContext, getHeaderForReads(), getReferenceSequenceDictionary(), bqsrArgs);
 
         final Broadcast<RecalibrationReport> reportBroadcast = ctx.broadcast(bqsrReport);
         final JavaRDD<GATKRead> finalReads = ApplyBQSRSparkFn.apply(markedReads, reportBroadcast, getHeaderForReads(), applyBqsrArgs.toApplyBQSRArgumentCollection(bqsrArgs.PRESERVE_QSCORES_LESS_THAN));
 
-        writeReads(ctx, output, finalReads);
+        // Run Haplotype Caller
```
lbergelson: @tomwhite I think we should keep an option to write the final bam file even if you're going to continue on and call variants.  I suspect it's something a lot of people will want to do.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -234,16 +266,15 @@ protected void runTool(final JavaSparkContext ctx) {
...
         final List<VariantContext> sortedVariants = collectedVariants.stream()
             .sorted((o1, o2) -> IntervalUtils.compareLocatables(o1, o2, referenceDictionary))
             .collect(Collectors.toList());
 
-        try(final VariantContextWriter writer = hcEngine.makeVCFWriter(output, getBestAvailableSequenceDictionary())) {
-            hcEngine.writeHeader(writer, getHeaderForReads().getSequenceDictionary(), new HashSet<>());
+        try(final VariantContextWriter writer = hcEngine.makeVCFWriter(outputFile, referenceDictionary)) {
+            hcEngine.writeHeader(writer, readsDictionary, new HashSet<>());
             sortedVariants.forEach(writer::add);
         }
```
lbergelson: Could you remove the now unused method `getShardBoundaryOverlapDetector` on line 286?  It became dead code in a previous PR but I didn't notice until now.

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSparkIntegrationTest.java
```diff
@@ -57,31 +56,23 @@ private String getResourceDir(){
...
         final String more20Sites = getResourceDir() + "dbsnp_138.b37.20.10m-10m100.vcf"; //for testing 2 input files
 
-        final String expectedSingleKnownSites = "expected.CEUTrio.HiSeq.WGS.b37.ch20.1m-1m1k.NA12878.noMD.noBQSR.md.bqsr.bam";
-        final String expectedMultipleKnownSites = "expected.MultiSite.reads.pipeline.bam";
-        final String expectedMultipleKnownSitesCram = "expected.MultiSite.reads.pipeline.cram";
+        final String expectedMultipleKnownSites = "expected.MultiSite.reads.pipeline.vcf";
 
         return new Object[][]{
                 // input local, computation local.
-                //Note: these output files were created by running Picard 1.130 and GATK3.46
```
lbergelson: I think it would be best if we could keep these tests, and just generate expected variant files from them.   Ideally we could run them in a mode the produces a bam and variants so we can detect subtler changes that may not be reflected strongly in the variants.  Why delete them all?

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
tomwhite: I need to investigate this further since I was hitting problems with the sequence dictionary in a whole genome BAM being different to the reference.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
tomwhite: That's a good idea. I'll add it.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
tomwhite: Done

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSparkIntegrationTest.java
tomwhite: I initially removed them due to expediency, so I've gone back and added them back. The only ones I think we don't need are the shuffle join partitioner ones - they fail due to a serialization error, but we plan on removing the shuffle join partitioner since broadcast or overlaps partitioner are superior.

Also, CRAM isn't working since it doesn't work with 2bit, and HaplotypeCaller requires 2bit. I'll add a note about this.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
tomwhite: I've changed this to just take a single reference.

tomwhite:
@lbergelson back to you.

tomwhite:
I've addressed all the feedback and all tests are passing so I'm going to squash and merge this now.

lbergelson:
@tomwhite Sorry, I was away for a bit, thanks for addressing issues and merging 👍

Issue #2703, opened 2017-05-10
Make discoveries in complex events


SHuang-Broad:
As there are many possibilities, will tackle different types differently first.
First deal with #2702

SHuang-Broad:
__The series of PRs waiting line is following the chart below__

![cpxsvflowchart](https://user-images.githubusercontent.com/16310888/29530831-968c631e-8673-11e7-98c2-765827941f91.png)

SHuang-Broad:
closed by the last PR-8. Flowchart to be updated for documentation purposes.

PR #3456, opened 2017-08-17
PR series for complex SV, part 2

SHuang-Broad:
New tool aiming to call all types of precise variants detectable by long read alignments (not fully functioning yet in the sense that not all types of variants are detected yet&mdash;to be handled by later PRs in this series).
This new tool splits the input long reads by scanning their alignment characteristics (number of alignments, if strand switch is involved, if mapped to the same chromosome, if have equally good alignment configurations based on the scoring tool, etc), and send them down different code path/logic units for variant type inference and VCF output.
This PR would only deal with simple INSDEL, for long reads having exactly 2 alignments (no other equally good alignment configuration) mapped to the same chromosome without strand switch or order switch (translocation or large tandem duplications), because we already have this type of variant covered in master.

__UPDATE__
See updated roadmap in #2703.

NEEDS TO WAIT UNTIL PART 1 IS IN.


SHuang-Broad:
Step 3 towards #2703

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ForSimpleInsDel.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+
+final class ForSimpleInsDel implements VariantDetectorFromLongReadAlignments {
+
+    public void inferSvAndWriteVCF(final JavaRDD<AlignedContig> longReads, final String vcfOutputFileName,
+                                   final Broadcast<ReferenceMultiSource> broadcastReference, final String fastaReference,
+                                   final GCSOptions options, final Logger toolLogger){
```
cwhelan: It's kind of annoying to have to keep passing around `GCSOptions` and I think the engine team has said these are being phased out eventually. It looks like it's only ever used if accessing a reference from the Google Genomics API, which we never do. It might be worth asking the engine team if we should still be using it, and/or just pass in null when we create the reference object.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ForSimpleInsDel.java
```diff
@@ -0,0 +1,76 @@
...
+                                noveltyTypeAndEvidence._2._1, noveltyTypeAndEvidence._2._2, broadcastReference));
+
+        SVVCFWriter.writeVCF(options, vcfOutputFileName, fastaReference, annotatedVariants, toolLogger);
+    }
+
+    /**
+     * Very similar to {@link ChimericAlignment#parseOneContig(AlignedContig, int)}, except the head/tail filtering.
+     */
+    private static Tuple2<byte[], List<ChimericAlignment>> convertAlignmentIntervalToChimericAlignment
+    (final AlignedContig contig) {
```
cwhelan: I'm not sure what the proper indentation is for parameters, but I think this should be indented some..

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ForSimpleInsDel.java
```diff
@@ -0,0 +1,76 @@
...
+        final List<AlignmentInterval> alignmentIntervals = contig.alignmentIntervals;
+        final Iterator<AlignmentInterval> iterator = alignmentIntervals.iterator();
+        AlignmentInterval current = iterator.next();
+        final List<ChimericAlignment> results = new ArrayList<>(alignmentIntervals.size() - 1);
+        final List<String> insertionMappings = new ArrayList<>();
+        while ( iterator.hasNext() ) {
+            final AlignmentInterval next = iterator.next();
+            if (ChimericAlignment.nextAlignmentMayBeNovelInsertion(current, next,
+                    StructuralVariationDiscoveryArgumentCollection.DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection
+                            .DEFAULT_MIN_ALIGNMENT_LENGTH)) {
```
cwhelan: Would it be too much of a pain to actually access the parameter here (ie have the tool pass it in)?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ForSimpleInsDel.java
```diff
@@ -0,0 +1,76 @@
...
+                        .groupByKey()
+                        .mapToPair(noveltyAndEvidence -> DiscoverVariantsFromContigAlignmentsSAMSpark.inferType(noveltyAndEvidence._1, noveltyAndEvidence._2))
+                        .map(noveltyTypeAndEvidence -> DiscoverVariantsFromContigAlignmentsSAMSpark.annotateVariant(noveltyTypeAndEvidence._1,
+                                noveltyTypeAndEvidence._2._1, noveltyTypeAndEvidence._2._2, broadcastReference));
+
+        SVVCFWriter.writeVCF(options, vcfOutputFileName, fastaReference, annotatedVariants, toolLogger);
+    }
+
+    /**
+     * Very similar to {@link ChimericAlignment#parseOneContig(AlignedContig, int)}, except the head/tail filtering.
```
cwhelan: Not sure what you mean by head/tail filtering - you mean starting the parsing at the first high-MQ mapping in the other method? Maybe make the comment a little more explicit?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
```diff
@@ -0,0 +1,228 @@
...
+
+import java.util.EnumMap;
+import java.util.HashSet;
+import java.util.Set;
+
+
+/**
+ * This tool takes a SAM file containing alignments of single-ended long read
+ * (be it long read sequencing, or contigs assembled from standard Illumina short reads),
+ * searches for reads with complicated split alignments indicating the presence of complex structural variations (cxSV),
```
cwhelan: Since we're also calling simple SVs (as done in this PR), this description seems incomplete.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
```diff
@@ -0,0 +1,228 @@
...
+public final class InternalCpxSvDiscoverFromLongReadsSpark extends GATKSparkTool {
+    private static final long serialVersionUID = 1L;
+    private final Logger localLogger = LogManager.getLogger(InternalCpxSvDiscoverFromLongReadsSpark.class);
+
+    @ArgumentCollection
+    private StructuralVariationDiscoveryArgumentCollection.DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection
+            discoverStageArgs
+            = new StructuralVariationDiscoveryArgumentCollection.DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection();
+
+    @Argument(doc = "file containing non-canonical contig names (e.g chrUn_KI270588v1) in the reference, human reference assumed when omitted",
```
cwhelan: Since in the default it's assuming that chromosome names start with "chr", the default is more accurately "human hg19 or hg38".

cwhelan: It also occurs to me that the alt contig names are in the BWA reference files, so maybe we could eventually just read them out of there automatically.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
```diff
@@ -0,0 +1,228 @@
...
+            final Broadcast<SAMFileHeader> headerBroadcast = ctx.broadcast(header);
+            splitUpLongReads.forEach((k, v) -> writeSAM(v, k.name(), reads, headerBroadcast, outputDir, localLogger));
+        }
+
+        new ForSimpleInsDel()
+                .inferSvAndWriteVCF(splitUpLongReads.get(RawTypes.InsDel), outputDir+"/"+RawTypes.InsDel.name()+".vcf",
+                        ctx.broadcast(getReference()), discoverStageArgs.fastaReference, getAuthenticatedGCSOptions(),
+                        localLogger);
+
+//        new ForSimpleStrandSwitch()
```
cwhelan: For now remove commented code.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
```diff
@@ -0,0 +1,228 @@
...
+//                .inferSvAndWriteVCF(splitUpLongReads.get(RawTypes.Inv), outputDir+"/"+RawTypes.Inv.name()+".vcf",
+//                        ctx.broadcast(getReference()), discoverStageArgs.fastaReference, getAuthenticatedGCSOptions(),
+//                        localLogger);
+    }
+
+    private enum RawTypes {
+        MultiConf, Inv, InsDel, SegDupOrMEI, Cpx;
```
cwhelan: `MultiConf` doesn't really make it obvious what these things are. Might `Ambiguous` be a better name?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    private static boolean hasEquallyGoodConfigurations(final AlignedContig longread) {
```
cwhelan: This method seems unnecessary, as done the method below.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
```diff
@@ -0,0 +1,228 @@
...
+        return longread.hasEquallyGoodAlnConfigurations;
+    }
+
+    private static boolean hasOnly2Alignments(final AlignedContig longreadWithOnlyOneConfig) {
+        return longreadWithOnlyOneConfig.alignmentIntervals.size() == 2;
+    }
+
+    private static boolean isSameChromosomeMapping(final AlignedContig longreadWithOnlyOneConfigAnd2Aln) {
```
cwhelan: You might want to add an argument validation assertion to this method.

cwhelan: And to the rest of this family of methods.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
```diff
@@ -0,0 +1,228 @@
...
+
+        final AlignmentInterval intervalOne, intervalTwo;
+        if (longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(0).forwardStrand) {
+            intervalOne = longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(0);
+            intervalTwo = longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(1);
+        } else {
+            intervalOne = longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(1);
+            intervalTwo = longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(0);
+        }
+        return ( intervalOne.referenceSpan.getEnd() <= intervalTwo.referenceSpan.getStart() )
```
cwhelan: If the first case of this clause is true, won't the second one always be true too? Seems like you could just check the second one..

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
```diff
@@ -0,0 +1,228 @@
...
+
+        final AlignmentInterval intervalOne, intervalTwo;
+        if (longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(0).forwardStrand) {
+            intervalOne = longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(0);
+            intervalTwo = longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(1);
+        } else {
+            intervalOne = longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(1);
+            intervalTwo = longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(0);
+        }
+        return intervalOne.referenceSpan.getEnd() > intervalTwo.referenceSpan.getStart() &&
```
cwhelan: It seems like this clause is also redundant to me. If the second clause is true, the first clause will always be true, right?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
```diff
@@ -0,0 +1,228 @@
...
+            intervalOne = longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(1);
+            intervalTwo = longreadWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(0);
+        }
+        return ( intervalOne.referenceSpan.getEnd() <= intervalTwo.referenceSpan.getStart() )
+                ||
+                ( intervalOne.referenceSpan.getStart() < intervalTwo.referenceSpan.getStart() &&
+                        intervalOne.referenceSpan.getEnd() < intervalTwo.referenceSpan.getEnd());
+    }
+
+    private static boolean isLikelySegDupOrMEI(final AlignedContig longreadWithOnlyOneConfigAnd2AlnToSameChr) {
```
cwhelan: SegDup (I'd call it a dispersed duplication actually) or MEI is probably the most likely explanation for a normal germline sample, but there could in rare cases be actual translocations of sequence in the same chromosome. Also, this only catches one side of a purported MEI or dispersed duplication -- a mapping that happened to be on the other side would look more like a deletion, right? A more agnostic name for these alignment patterns might be better.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/VariantDetectorFromLongReadAlignments.java
```diff
@@ -0,0 +1,19 @@
...
+import org.broadinstitute.hellbender.engine.datasources.ReferenceMultiSource;
+import org.broadinstitute.hellbender.tools.spark.sv.discovery.AlignedContig;
+
+/**
+ * A base class for workflow of variant breakpoint detection from split alignments, variant type interpretation,
+ * and resolving complication {@link org.broadinstitute.hellbender.tools.spark.sv.discovery.BreakpointComplications}.
+ */
+interface VariantDetectorFromLongReadAlignments {
+
+    abstract void inferSvAndWriteVCF(final JavaRDD<AlignedContig> longReads, final String vcfOutputFileName,
```
cwhelan: Don't need abstract keyword

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/FileUtils.java
```diff
@@ -20,4 +23,36 @@ public static void writeLinesToSingleFile(final Iterator<String> linesToWrite, f
...
+                                    final boolean preOrdered) {
+        try (final SAMFileWriter writer = ReadUtils.createCommonSAMWriter(new File(outputName), null, header,
+                preOrdered, true, false) ) {
+            alignments.forEachRemaining(writer::addAlignment);
+        } catch ( final UncheckedIOException ie) {
+            throw new GATKException("Can't write SAM file to the specified location: " + outputName, ie);
+        }
+    }
+
+    public static boolean createDirInBucketToWriteTo(final String pathString) {
```
cwhelan: This method isn't really bucket-specific so it probably doesn't need `bucket` in the name.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ForSimpleInsDel.java
```diff
@@ -0,0 +1,76 @@
...
+import org.broadinstitute.hellbender.tools.spark.sv.discovery.DiscoverVariantsFromContigAlignmentsSAMSpark;
+import org.broadinstitute.hellbender.tools.spark.sv.utils.SVVCFWriter;
+import scala.Tuple2;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+
+final class ForSimpleInsDel implements VariantDetectorFromLongReadAlignments {
```
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ForSimpleInsDel.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public void inferSvAndWriteVCF(final JavaRDD<AlignedContig> longReads, final String vcfOutputFileName,
```
SHuang-Broad: Yes, that's the intention, but will change back for now.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ForSimpleInsDel.java
SHuang-Broad: just talked with the Engine team and tool developers shouldn't need to worry about them and essentially always use `null` if have to pass in an argument for that

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ForSimpleInsDel.java
SHuang-Broad: I was referring to lines 137 - 140
```
AlignmentInterval current = iterator.next();
    while (mapQualTooLow(current) && iterator.hasNext()) {
        current = iterator.next();
}
```

SHuang-Broad: But will add clarification to the doc

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ForSimpleInsDel.java
SHuang-Broad: not sure why I did that, sorry....

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ForSimpleInsDel.java
SHuang-Broad: nope, will change as suggested.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
SHuang-Broad: will remove

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/FileUtils.java
SHuang-Broad: changed, and added documentation to signal that it can work with buckets too.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/VariantDetectorFromLongReadAlignments.java
SHuang-Broad: removed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
SHuang-Broad: updated.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
SHuang-Broad: removed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
SHuang-Broad: not sure what you mean, should I open a ticket for the 2nd comment?

SHuang-Broad: But we could also make the point that a custom set of chromosomes is to be excluded, e.g. "chrM" or EBV, this is possible with the current implementation, am I misunderstanding something?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
SHuang-Broad: not really. it was intended to check for two cases:
1. first alignment (alignment with a lower corresponding start location on the input locally assembled contig) reference span is earlier than that of the second, i.e. the two doesn't overlap, 
2. the two overlaps but doesn't do "order switch", meaning the second alignment has a lower reference span
And indeed this method has a bug: just a check that 
```
intervalOne.referenceSpan.getStart() <= intervalTwo.referenceSpan.getStart()
```
is enough to guarantee no order switch is occurring.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalCpxSvDiscoverFromLongReadsSpark.java
SHuang-Broad: correct. Though i just found that this has a bug in it, meaning I should be checking for an reference block order switch, and strandness should be involved
It should be 
```
intervalOne.referenceSpan.getStart() > intervalTwo.referenceSpan.getStart() == intervalOne.forwardStrand
```

SHuang-Broad:
@cwhelan, would you please take another look, since I discovered a bug and fixed it in the latest commit? Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalSvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
```diff
@@ -0,0 +1,230 @@
...
+                .equals(contigWithOnlyOneConfigAnd2Aln.alignmentIntervals.get(1).referenceSpan.getContig());
+    }
+
+    private static boolean isLikelyInvBreakpointOrInsInv(final AlignedContig contigWithOnlyOneConfigAnd2AlnToSameChr) {
+        Utils.validateArg(hasOnly2Alignments(contigWithOnlyOneConfigAnd2AlnToSameChr),
+                "assumption that input contig has only 2 alignments is violated. \n" +
+                        onErrorStringRepForAlignedContig(contigWithOnlyOneConfigAnd2AlnToSameChr));
+        final AlignmentInterval intervalOne = contigWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(0),
+                                intervalTwo = contigWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(1);
+        Utils.validateArg(intervalOne.referenceSpan.getContig().equals(intervalTwo.referenceSpan.getContig()),
```
cwhelan: You could use the `isSameChromosomeMapping` method above

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalSvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
```diff
@@ -0,0 +1,230 @@
...
+        Utils.validateArg(hasOnly2Alignments(contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch),
+                "assumption that input contig has only 2 alignments is violated. \n" +
+                        onErrorStringRepForAlignedContig(contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch));
+
+        final AlignmentInterval intervalOne = contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch.alignmentIntervals.get(0),
+                                intervalTwo = contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch.alignmentIntervals.get(1);
+        Utils.validateArg(intervalOne.forwardStrand == intervalTwo.forwardStrand,
+                "assumption that input contig's 2 alignment map to the same strand is violated. \n" +
+                        onErrorStringRepForAlignedContig(contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch));
+        Utils.validateArg(intervalOne.referenceSpan.getContig().equals(intervalTwo.referenceSpan.getContig()),
```
cwhelan: Again you could use `isSameChromosomeMapping`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalSvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
```diff
@@ -0,0 +1,230 @@
...
+    private enum RawTypes {
+        Ambiguous, Inv, InsDel, DispersedDupOrMEI, Cpx;
+    }
+
+    private static boolean hasOnly2Alignments(final AlignedContig contigWithOnlyOneConfig) {
+        return contigWithOnlyOneConfig.alignmentIntervals.size() == 2;
+    }
+
+    private static boolean isSameChromosomeMapping(final AlignedContig contigWithOnlyOneConfigAnd2Aln) {
+        Utils.validateArg(hasOnly2Alignments(contigWithOnlyOneConfigAnd2Aln),
```
cwhelan: it seems like you could shorten these up if you made methods out of them, ie. `validateTwoAlignments()`, so you wouldn't have to repeat the logic and error message everywhere

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/FileUtils.java
```diff
@@ -34,7 +34,10 @@ public static void writeSAMFile(final Iterator<SAMRecord> alignments, final SAMF
         }
     }
 
-    public static boolean createDirInBucketToWriteTo(final String pathString) {
+    /**
+     * Creates a directory, in local FS, HDFS, or Google buckets to write to.
+     */
+    public static boolean createDirToWriteTo(final String pathString) {
```
cwhelan: In this method, I think that you could use IOUtils.getPath() you could combine this clause with the third clause.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalSvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
SHuang-Broad: I think the following change suggestions will cover this, so no action on this.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalSvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/InternalSvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/FileUtils.java
SHuang-Broad: Nice!

PR #3457, opened 2017-08-17
PR series for complex SV, part 3

SHuang-Broad:
This PR deals with long reads with exactly two alignments (no other equally good alignment configuration), mapped to the same chromosome with strand switch, but NOT significantly overlapping each other.

We used to call inversions from such alignments, but it is more appropriate to emit BND records because a lot of times such signal is actually generated from inverted segmental duplications, or simply inverted mobile element insertions. To confidently interpret and distinguish between such events, we need other types of evidence, and is better to be dealt with downstream logic units.

Inverted duplications are NOT dealt with in this PR and is going to be in the next.

NEEDS TO WAIT UNTIL PART 1 & 2 ARE IN.


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AlignedContig.java
```diff
@@ -14,22 +14,25 @@
 
 /**
  * Locally assembled contig:
- * its name
- * its sequence as produced by the assembler (no reverse complement like in the SAM record if it maps to '-' strand), and
- * its stripped-down alignment information.
+ *   its name
```
pshapiro4broad: When formatted as javadoc this will all appear as a single paragraph of text. To write a list in javadoc you need to use some basic HTML:
```
/**
  * Locally assembled contig:
  * <ul><li>its name
  * <li>its sequence ...
  * <li>its stripped-down ...
  * </ul>
  */
```
Note that you don't need need `</li>` as it's automatically closed by `<li>`.

SHuang-Broad:
Step 4 towards #2703

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/FileUtils.java
```diff
@@ -0,0 +1,56 @@
...
+import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
+import org.broadinstitute.hellbender.utils.read.ReadUtils;
+
+import java.io.*;
+import java.util.Iterator;
+
+public final class FileUtils {
+
+    public static void writeLinesToSingleFile(final Iterator<String> linesToWrite, final String fileName) {
```
pshapiro4broad: I believe that this can be replaced with [`Files.write()`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#write-java.nio.file.Path-java.lang.Iterable-java.nio.file.OpenOption...-)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/FileUtils.java
```diff
@@ -0,0 +1,56 @@
...
+            alignments.forEachRemaining(writer::addAlignment);
+        } catch ( final UncheckedIOException ie) {
+            throw new GATKException("Can't write SAM file to the specified location: " + outputName, ie);
+        }
+    }
+
+    public static boolean createDirInBucketToWriteTo(final String pathString) {
+        try {
+            Utils.nonNull(pathString);
+            if ( java.nio.file.Files.exists(java.nio.file.Paths.get(pathString)) )
```
pshapiro4broad: If possible, you should modify your imports so a fully qualified name isn't required for `java.io.file.` symbols.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AlignmentInterval.java
```diff
@@ -36,9 +37,11 @@
     public final int mismatches;
     public final int alnScore;
 
-    // if this is true, fields "mapQual", "mismatches", "alnScore" should be viewed with care as they were simply copied from the
-    // original alignment (not for "mismatches"), which after the split are wrong (we didn't recompute them because that would require expensive SW re-alignment)
+    // if any of the following boolean fields are true, fields "mapQual", "mismatches", "alnScore" should be viewed
```
pshapiro4broad: It looks like at least part of this should be put in a javadoc for this field.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AnnotatedVariantProducer.java
```diff
@@ -22,9 +22,36 @@
...
+        final List<VariantContext> result = new ArrayList<>();
+        result.add(record);
+        // hack for now because up to this point inferredType would have max of 2 only
+        while (it.hasNext()) {
+            final VariantContext mateRecord =
+                    produceAnnotatedVcFromInferredTypeAndRefLocations(novelAdjacencyReferenceLocations.leftJustifiedRightRefLoc, -1,
+                            novelAdjacencyReferenceLocations.complication, it.next(), contigAlignments, broadcastReference);
+            result.add(mateRecord);
+        }
+        return result.iterator();
```
pshapiro4broad: Why not return `List<>` here? A caller can easily convert it to an iterator if needed. Returning `List` provides more flexibility to the caller.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/FileUtils.java
SHuang-Broad: Thanks for taking a look, @pshapiro4broad. This would also us to write to local FS, hadoop, and google buckets, i.e. more general than `Files.write()`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/FileUtils.java
SHuang-Broad: This is to deal with name clashes between nio Files and hadoop Files.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AnnotatedVariantProducer.java
SHuang-Broad: That's constraint by API in Spark.

SHuang-Broad:
@cwhelan this is now ready. Please review. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
```diff
@@ -0,0 +1,301 @@
...
+            final CigarElement ce = cigarElements.get(idx);
+            if ( ce.getOperator().equals(CigarOperator.HARD_CLIP))
+                continue;
+            if ( ce.getOperator().consumesReadBases() ) {
+                if (readBasesConsumed + ce.getLength() < clipLength)
+                    readBasesConsumed += ce.getLength();
+                else { // enough read bases would be clipped
+
+                    if ( !ce.getOperator().isAlignment() && !ce.getOperator().equals(CigarOperator.I))
+                        throw new GATKException.ShouldNeverReachHereException("Logic error, should not reach here");
```
cwhelan: Could you give a more descriptive error message explaining what is wrong?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
```diff
@@ -0,0 +1,301 @@
...
+
+                    // dead with cigar first
```
cwhelan: Not sure what dead means in the context of this comment

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
```diff
@@ -0,0 +1,301 @@
...
+                    final List<CigarElement> resultCEs = threeSections._1();
+                    final int a = readBasesConsumed + ce.getLength() - clipLength;
+                    final CigarOperator op = ce.getOperator().isAlignment() ? CigarOperator.M : CigarOperator.S;
+                    if (clipFrom3PrimeEnd) {
```
cwhelan: This code might be a little more concise if you put these things in the sublist in the same order all the time, and then reverse the sublist if `clipFrom3PrimeEnd` with `Collections.reverse()`, and then append the sublist to `resultCEs`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
```diff
@@ -0,0 +1,301 @@
...
+                        }
+                        resultCEs.add(new CigarElement(clipLength, CigarOperator.S));
+                    } else {
+                        resultCEs.add(new CigarElement(clipLength, CigarOperator.S));
+                        if (a!=0) {
+                            resultCEs.add( new CigarElement(a, op) );
+                        }
+                        resultCEs.addAll(cigarElements.subList(idx+1, cigarElements.size()));
+                    }
+                    if (!threeSections._3().isEmpty())
```
cwhelan: You don't really need this if check, you could still add it in if it's empty.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        if (middle.isEmpty())
+            throw new GATKException("Logic error: cigar elements corresponding to alignment block is empty. " + input.toPackedString());
+
+        return new Tuple3<>(left, middle, right);
+    }
+
+    /**
+     * Taking advantage of the fact that for input read, we know it has only two alignments that map to the same reference
```
cwhelan: I'd add an assertion to this method too if that's the case.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
```diff
@@ -0,0 +1,301 @@
...
+
+    // =================================================================================================================
+
+    private JavaRDD<VariantContext> dealWithSimpleStrandSwitchBkpts(final JavaRDD<AlignedContig> longReads,
+                                                                    final Broadcast<ReferenceMultiSource> broadcastReference,
+                                                                    final Logger toolLogger) {
+
+        final JavaPairRDD<ChimericAlignment, byte[]> simpleStrandSwitchBkpts =
+                longReads
+                        .mapToPair(SimpleStrandSwitchVariantDetector::convertAlignmentIntervalToChimericAlignment)
```
cwhelan: To me a more straightforward flow would be to `filter` them on `splitPairStrongEnoughEvidenceForCA` and then convert the remaining ones to `ChimericAlignment`s, rather than filtering on `nonNull`.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/VariantDetectorFromLongReadAlignmentsForSimpleStrandSwitchUnitTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+import scala.Tuple2;
+
+import java.io.IOException;
+
+public class VariantDetectorFromLongReadAlignmentsForSimpleStrandSwitchUnitTest extends BaseTest {
```
cwhelan: It might be worth adding a test for `extractCigarElements` as well. Perhaps some of the other methods in the class, too.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
```diff
@@ -0,0 +1,301 @@
...
+
+        final JavaRDD<VariantContext> simpleStrandSwitchBkpts =
+                dealWithSimpleStrandSwitchBkpts(split._2, broadcastReference, toolLogger);
+        SVVCFWriter.writeVCF(null, vcfOutputFileName.replace(".vcf", "_simpleSS.vcf"),
+                fastaReference, simpleStrandSwitchBkpts, toolLogger);
+
+        // TODO: 8/23/17 add inv dup code in the next pr
+    }
+
+    public static final class IsLikelyInvertedDuplication implements SerializablePredicate<AlignedContig> {
```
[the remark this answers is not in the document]
tedsharpe: Your SerializablePredicate is semi-redundant with org.apache.spark.api.java.function.Function<T, Boolean>, which is what JavaRDD::filter expects.  You could eliminate it, and just use the spark Function interface in RDDUtils:
```
import org.apache.spark.api.java.function.Function;
...
    public static <T> Tuple2<JavaRDD<T>, JavaRDD<T>> split(
            final JavaRDD<T> input,
            final Function<T, Boolean> predicate,
            final boolean unpersistInputRDD) {

        final JavaRDD<T> first = input.filter(predicate).cache();
        final JavaRDD<T> second = input.filter(t -> !predicate.call(t)).cache();

        if (unpersistInputRDD) input.unpersist();
        return new Tuple2<>(first, second);
    }
```

And then in this simple circumstance you can eliminate your explicit static class in favor of a lambda.  The call at line 53 above becomes:
```
        final Tuple2<JavaRDD<AlignedContig>, JavaRDD<AlignedContig>> split
                = RDDUtils.split(contigs.map(SimpleStrandSwitchVariantDetector::removeOverlap),
                                  contig ->
                                          BreakpointComplications.isLikelyInvertedDuplication(
                                                  contig.alignmentIntervals.get(0),
                                                    contig.alignmentIntervals.get(1)),
                                  true);
```

This doesn't prevent you from implementing Function in an explicit static class when it's complicated enough to warrant it.

Does that help?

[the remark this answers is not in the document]
tedsharpe: By the way, I'm a little skeptical of putting the caching behavior into RDDUtils.split.
I fear it might not always be appropriate (if the RDD is very large, for example), but a client might forget that the caching is happening, and might easily forget to unpersist the split RDDs.
Better, IMO, to have the cache and unpersist be called in the same code block, so that it's absolutely obvious how executor memory is being managed.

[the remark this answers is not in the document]
SHuang-Broad: Thanks for the suggestion, @tedsharpe. For some unknown reason, the lambda lead to a task not serializable exception when I first attempted as in your suggestion, and that is why I had this Functor wrapper around a function. I'll try again and see if it works with a simple lambda.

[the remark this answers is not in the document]
SHuang-Broad: For caching the split-up two RDDs, I agree and the caching will be removed in the next commit.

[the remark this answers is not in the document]
tedsharpe: The trick is that the spark functional interfaces are all marked Serializable.  So you have to use one of those, rather than, for example, Java's Predicate interface, which isn't Serializable out of the box.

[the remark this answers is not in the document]
SHuang-Broad: Right, that was why I had this trivial SerializablePredicate interface, without knowing that there is this `Function` class in Spark as well.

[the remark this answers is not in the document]
tedsharpe: It would have been nice if Spark had extended Predicate with serializability, rather than reusing Function with a Boolean return type for its filter method.  That way the negative test could have used predicate.negate().  Oh, well.

SHuang-Broad:
ping @cwhelan .

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
```diff
@@ -67,7 +72,8 @@ public boolean test(final AlignedContig contig) {
 
     /**
      * Removes overlap from a designated alignment interval, so that the inverted duplicated reference span is minimal.
-     * If the two alignment intervals are NOT overlapping, return the original read.
+     * If the two alignment intervals are NOT overlapping, return the original aligned contig.
+     * For algorithm {@see <a href="https://github.com/broadinstitute/dsde-methods-sv/pull/8"}
```
cwhelan: This is our private repo so I'm not sure we should reference it in comments in the public code -- it would just be frustrating for someone external.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
```diff
@@ -77,6 +83,8 @@ private static AlignedContig removeOverlap(final AlignedContig contig) {
         } else {
             final AlignmentInterval one = contig.alignmentIntervals.get(0),
                                     two = contig.alignmentIntervals.get(1);
+            // js is for "the shoot-off reference location of a jump that linked to alignment intervals", and
```
cwhelan: Is this supposed to be "linked two alignment intervals"?

I'm not sure I understand what "shoot-off" and "landing" are -- or, I guess I can imagine, but maybe something like "jumpStart" and "jumpEnd" might be more intuitive?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
```diff
@@ -93,73 +101,75 @@ private static AlignedContig removeOverlap(final AlignedContig contig) {
...
-                    refBasesConsumed += ce.getOperator().isAlignment() ? (clipLength - readBasesConsumed)
+                } else { // enough read bases would be clipped, note that here we can have only either 'M' or 'I'
+
+                    if (!ce.getOperator().isAlignment() && !ce.getOperator().equals(CigarOperator.I))
+                        throw new GATKException.ShouldNeverReachHereException("Logic error, should not reach here: operation consumes read bases but is neither alignment nor insertion.\n " +
+                                "Original cigar(" + TextCigarCodec.encode(input.cigarAlong5to3DirectionOfContig) + ")\tclipLengthOnRead(" + clipLengthOnRead + ")\t" + (clipFrom3PrimeEnd ? "clipFrom3PrimeEnd" : "clipFrom5PrimeEnd"));
+
+                    // deal with cigar first
+                    newMiddleSection.add( new CigarElement(clipLengthOnRead, CigarOperator.S) );
+                    final int a = readBasesConsumed + ce.getLength() - clipLengthOnRead;
```
cwhelan: Would a more descriptive name for `a` be `basesRemainingOnCigarElementAfterClipping`? Just trying to make sure I understand the algorithm.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AnnotatedVariantProducer.java
```diff
@@ -90,7 +90,7 @@ static VariantContext produceAnnotatedVcFromInferredTypeAndRefLocations(final Si
                 .id(inferredType.getInternalVariantId())
                 .attribute(GATKSVVCFConstants.SVTYPE, inferredType.toString());
 
-        if (inferredType instanceof SimpleSVType)
```
cwhelan: I don't think there's anything wrong with using `instanceof`.

SHuang-Broad: Strangely it failed... Plus it might be better if an alt route exists that doesn't use reflection and has equal line of code works

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
SHuang-Broad: Agreed and removed. 
But we are more and more in need of a central place for documenting the algorithms used in the various stages of the pipeline.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
SHuang-Broad: !OoO!, yes, it should be "two" instead of "to".

I have in mind a cartoonish depiction of a tiny anime figure sliding along the reference guided by the alignments of the read, sometimes jumping back and forth. Sorry for this too-pictorial naming scheme.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SimpleStrandSwitchVariantDetector.java
SHuang-Broad: yes, that's right and updated with documentation

Issue #3459, opened 2017-08-17
BwaMemAlignmentUtils.applyAlignments produces invalid SAMRecords.


vruano:
It seems that it fails to hard-clip the bases array if the cigar has hard-clips.

```
java.lang.IllegalStateException: CIGAR covers 4 bases but the sequence is 8 read bases 

	at org.broadinstitute.hellbender.utils.bwa.BwaMemAlignmentUtils.applyAlignment(BwaMemAlignmentUtils.java:92)
	at org.broadinstitute.hellbender.tools.spark.sv.discovery.AlignmentIntervalUnitTest.testConstructionFromGATKRead(AlignmentIntervalUnitTest.java:133)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:108)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:661)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:869)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1193)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:126)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:109)
	at org.testng.TestRunner.privateRun(TestRunner.java:744)
	at org.testng.TestRunner.run(TestRunner.java:602)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:380)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:375)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:340)
	at org.testng.SuiteRunner.run(SuiteRunner.java:289)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1301)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1226)
[...]

PR #3461, opened 2017-08-18
Fixes a bug in applyAlignment where some times hard-clipped based (as per the cigar) re not removed for the bases array.

vruano:
Fixes issue #3459

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
@@ -4,6 +4,7 @@ import htsjdk.samtools.*;
 import htsjdk.samtools.util.SequenceUtil;
 import org.broadinstitute.hellbender.tools.spark.sv.utils.SVUtils;
 import org.broadinstitute.hellbender.utils.BaseUtils;
+import org.broadinstitute.hellbender.utils.Utils;
 
 import java.util.*;
 
@@ -24,12 +25,21 @@ public class BwaMemAlignmentUtils {
                                            final String readGroup, final BwaMemAlignment alignment,
                                            final List<String> refNames, final SAMFileHeader header,
                                            final boolean softclipAlts, final boolean alwaysGenerateASandXS ) {
+        Utils.nonNull(alignment, "the input alignment cannot be null");
+        Utils.nonNull(basesArg, "the input bases argument cannot be null");
+        Utils.nonNull(refNames, "the input reference name list cannot be null");
+        final int refId = alignment.getRefId();
+        final int refMateId = alignment.getMateRefId();
+        if (refId >= refNames.size()) throw new IllegalArgumentException("reference names array to short for refId");
+        if (refMateId >= refNames.size()) throw new IllegalArgumentException("reference names array to short for refMateId");
+        if (refId >= 0 && refNames.get(refId) == null) throw new IllegalArgumentException("reference name cannot be null");
+        if (refMateId >= 0 && refNames.get(refMateId) == null) throw new IllegalArgumentException("mate reference name cannot be null");
         final SAMRecord samRecord = new SAMRecord(header);
         samRecord.setReadName(readName);
         final int samFlag = alignment.getSamFlag();
         samRecord.setFlags(samFlag);
-        if ( alignment.getRefId() >= 0 ) samRecord.setReferenceName(refNames.get(alignment.getRefId()));
-        else if ( alignment.getMateRefId() >= 0 ) samRecord.setReferenceName(refNames.get(alignment.getMateRefId()));
+        if ( refId >= 0 ) samRecord.setReferenceName(refNames.get(refId));
+        else if ( refMateId >= 0 ) samRecord.setReferenceName(refNames.get(refMateId));
         if ( alignment.getRefStart() >= 0 ) samRecord.setAlignmentStart(alignment.getRefStart()+1);
         else if ( alignment.getMateRefStart() >= 0 ) samRecord.setAlignmentStart(alignment.getMateRefStart()+1);
         if ( alignment.getMapQual() >= 0 ) samRecord.setMappingQuality(alignment.getMapQual());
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
```diff
@@ -57,9 +67,15 @@ public static SAMRecord applyAlignment( final String readName, final byte[] base
                         }
                     }
                 }
-                bases = Arrays.copyOfRange(bases, alignment.getSeqStart(), alignment.getSeqEnd());
+            }
+            if (tmpCigar.getFirstCigarElement().getOperator() == CigarOperator.H ||
+                    tmpCigar.getLastCigarElement().getOperator() == CigarOperator.H) {
+                final int leftHardClip = tmpCigar.getFirstCigarElement().getOperator() == CigarOperator.H ? tmpCigar.getFirstCigarElement().getLength() : 0;
```
tedsharpe: You don't need to calculate the length of the hard clips -- the alignment tells you the proper bounds.  The old code was fine, and simpler.
```
bases = Arrays.copyOfRange(bases, alignment.getSeqStart(), alignment.getSeqEnd());
if ( quals.length != 0 ) quals = Arrays.copyOfRange(quals, alignment.getSeqStart(), alignment.getSeqEnd());
```

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AlignmentIntervalUnitTest.java
```diff
@@ -59,7 +59,7 @@ public void testAlignmentIntervalOverlap() throws Exception {
      *                                      intended to be used for testing concordance between the two constructors)
      */
     @DataProvider(name = "AlignmentIntervalCtorTestForSimpleInversion")
-    Object[][] createInputsAndExpectedResults_BwaMemAlignmentConstruction() {
+    public static Object[][] createInputsAndExpectedResults_BwaMemAlignmentConstruction() {
```
tedsharpe: The problem you're trying to solve is actually caused by bogus data generation by this unit test.
BWA would never create a primary alignment with hard clips, but this class does.  The problem would go away if you changed line 76 to these 4 lines:
```
            int samFlag = 0;
            if ( !strandedness[i] ) samFlag = SAMFlag.READ_REVERSE_STRAND.intValue();
            if ( cigarStrings[i].indexOf('H') != -1 ) samFlag |= SAMFlag.SUPPLEMENTARY_ALIGNMENT.intValue();
            final BwaMemAlignment bwaMemAlignment = new BwaMemAlignment(samFlag,
```
You could argue that BwaMemAlignmentUtils ought to do the right thing, anyway.  So, if you want to make the change to BwaMemAlignmentUtils, go right ahead.

vruano:
The other option here is to make applyAlignment to check the assumption it makes on the input data. I don't care if it fails as long as it does it early. What do you think @tedsharpe?

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
vruano: Do ```getSeqStart/End``` return the base index range in the read that is actually aligned to the ref (i.e. involving first M and last M cigar elements) or that plus the soft-clipped range? I suspected is the former thus may change. Otherwise, happy to revert to that code.

tedsharpe:
The change is fine:  It's more tolerant of bogus input, and the cost is nearly 0.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
tedsharpe: The start and end of the alignment always exclude clipping, as far as I know.  Kind of by definition, right?

tedsharpe: Otherwise these fields would be useless -- they'd always be 0 and read length - 1.

vruano: There still a diff between soft and hard-clipping, where the the soft determine the positions of the bases[] array relative to the full read sequence. The accessor names do not rule that possibility. I just wanted to make sure. 

The reason the old more succinct code works is because it only get executed with that ```if``` statement where is certain that all clipping is to be excised, however that is not in general true any longer if we are to support those "invalid" inputs. That is the reason for the more complex code, unless I have not realized of something.

tedsharpe: I think I haven't explained well enough what I'm proposing.  I propose that you replace lines 73 through 78 with this code:
```
            bases = Arrays.copyOfRange(bases, alignment.getSeqStart(), alignment.getSeqEnd());
            if ( quals.length != 0 ) {
                quals = Arrays.copyOfRange(quals, alignment.getSeqStart(), alignment.getSeqEnd());
            }
```

src/test/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtilsTest.java
```diff
@@ -1,13 +1,23 @@
...
 import org.testng.annotations.Test;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
-public class BwaMemAlignmentUtilsTest {
+public class BwaMemAlignmentUtilsTest extends BaseTest {
```
magicDGS: Should extend `GATKBaseTest` after #3475

droazen:
Closing this ancient PR, since as @tedsharpe mentioned it appears to fix a problem that can't occur in practice. Feel free to re-open if someone feels strongly that this patch should go in.

PR #3469, opened 2017-08-21
Imprecise variant calling -- gather clustered evidence links

cwhelan:
This PR is the initial stage of implementing the calling of IMPRECISE variants in the SV pipeline. It introduces the concept of an evidence-target link, which joins an evidence interval to its distal target. This is an extension of the 'coherent' evidence concept previously used in determining evidence thresholds for assembly. The code in this PR contains the following changes:

- Evidence intervals and distal targets now are treated as stranded, and evidence-target link clustering depends on overlaps between both intervals and strands.
- Evidence target interval and distal target interval calculations have been modified to make sure that evidence supporting the same event clusters together (has overlapping intervals). This includes several changes such as extending the 'rest-of-fragment-size' calculation to try to capture almost all non-outlier fragment sizes in the library; increasing the split read location uncertainty a little; and being more precise about the boundaries of distal target intervals by taking advantage of information in the MD and MC tags if available.
- Evidence target links are gathered for every piece of evidence supporting a high-quality distal target. 
- Evidence target links are clustered together and store the amount of split-read and read-pair evidence that went into each cluster.
- All evidence target link clusters that are composed of at least 1 split read or at least 2 read pairs are collected in the driver and emitted in a BEDPE formatted file specified in the command line parameters.
- A `PairedStrandedIntervalTree` data structure is introduced to allow `SVIntervalTree`-style lookups for paired intervals.

To finish this work, future PRs will 1) use the collected evidence target links to annotate our assembly called-variants with the number of split reads and read pairs observed in the original mappings and 2) create IMPRECISE VCF records for events that have enough evidence-target-link support, first for deletions and then possibly for other variant types.

Initial testing shows that these changes slightly increase the number of variants called by the current pipeline, on both the CHM mix and NA12878 data sets, without greatly affecting run time.


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLink.java
```diff
@@ -0,0 +1,121 @@
...
+        output.writeBoolean(sourceForwardStrand);
+        intervalSerializer.write(kryo, output, target);
+        output.writeBoolean(targetForwardStrand);
+
+        output.writeInt(splitReads);
+        output.writeInt(readPairs);
+    }
+
+    public String toBedpeString(ReadMetadata readMetadata) {
+        final Map<Integer, String> contigIdToContigNameMap =
```
tedsharpe: You don't need this.  ReadMetadata stores the inverse map.  Use readMetadata.getContigName( int id ) to turn a contigID into a contigName.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
```diff
@@ -0,0 +1,122 @@
...
+        return forwardStrand1 != null && forwardStrand2 != null && forwardStrand1.equals(forwardStrand2);
+    }
+
+    /**
+     * Combines links that agree on interval-pair and orientation but have source and target switched. For
+     * example, if link1 has intervals s1 and t1, and link2 has s2 and t2, and s1 overlaps t1 and s2 overlaps t2,
+     * the links will be combined as long as strands agree. Returned links have the intersection of two paired intervals as
+     * source and target, with the lower-coordinate interval appearing as source.
+     */
+    public static List<EvidenceTargetLink> deduplicateTargetLinks(final List<EvidenceTargetLink> evidenceTargetLinks) {
```
tedsharpe: Wouldn't it be easier to always record the links in canonical order (source < target)?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
```diff
@@ -0,0 +1,122 @@
...
+                                strandsMatch(nextEvidence.getDistalTargetStrands(readMetadata).get(0), oldLink.targetForwardStrand))) {
+                            // if it does, intersect the source and target intervals to refine the link
+                            it.remove();
+                            final SVInterval newSource = sourceIntervalEntry.getInterval().intersect(nextEvidence.getLocation());
+                            final SVInterval newTarget = oldLink.target.intersect(nextEvidence.getDistalTargets(readMetadata).get(0));
+                            updatedLink = new EvidenceTargetLink(newSource,
+                                    oldLink.sourceForwardStrand,
+                                    newTarget,
+                                    oldLink.targetForwardStrand,
+                                    nextEvidence instanceof BreakpointEvidence.DiscordantReadPairEvidence
```
tedsharpe: Should be BreakpointEvidence.SplitRead, right?

cwhelan:
Re: strandedness, while it would be much easier not to have to track strand (and would probably save me from writing a lot of bugs), I think that we do need to do so if we ever want these data elements to be used for calling more complicated events like inversions or translocations. For example, for inversions we need to record whether we saw evidence linking +/+ strands or -/- strands, since the breakpoints might appear different at the two junctions due to insertions or deletions. For translocations, you definitely need to know which strands were involved so that you can extrapolate the rest of the event from the breakpoint. Agreed that it's a pain to keep track of it, though, and I find myself introducing strand errors all the time.

tedsharpe:
I'm not suggesting that you ignore strand.
I'm suggesting that these two alignments are consistent, even though their strandedness doesn't agree:
readX +flag chr1 10001 60 75M76S ...    SA:chr1,85275,-,75S76M,60,1;
readY -flag chr1 10001 60 75M76S ...    SA:chr1,85275,+,75S76M,60,1;

Whereas, this alignment is inconsistent:
readZ -flag chr1 10001 60 75M76S ...    SA:chr1,85275,-,75S76M,60,1;

tedsharpe:
Should have said,
Whereas, these alignments are inconsistent with the pair above, but consistent with each other:
readZ -flag chr1 10001 60 75M76S ... SA:chr1,85275,-,75S76M,60,1;
readZ +flag chr1 10001 60 75M76S ... SA:chr1,85275,+,75S76M,60,1;

cwhelan:
That's right, but for split reads the strand I'm using is not the actual strand of the alignment but the side of each aligned region on which the clipping occurred (clip on the left == negative strand, clip on the right == positive strand). Strandedness is more important for read pair data but I wanted to give it an analogue for split reads so that the evidence from the two types of reads could play together. Let's discuss more in person.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -695,7 +724,56 @@ public AlignedAssemblyOrExcuse apply( final Tuple2<Integer, List<SVFastqUtils.Fa
             intervals.add(evidenceIterator2.next().getLocation());
         }
 
-        return intervals;
+        return new Tuple2<>(intervals, evidenceTargetLinks);
+    }
+
+    private static void writeTargetLinks(final FindBreakpointEvidenceSparkArgumentCollection params, final Broadcast<ReadMetadata> broadcastMetadata, final List<EvidenceTargetLink> targetLinks) {
```
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLink.java
cwhelan: Oops, I missed that, thanks. done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
cwhelan: I don't think so.. These ternaries are hard to read but in this clause and the next I'm saying:

- if it's a read pair use the old split read count, otherwise increment the split read count
- if it's a read pair increment the read pair count, otherwise use the old read pair count

Maybe if I pull these out into named variables it will be clearer.

tedsharpe:
Light begins to dawn on Marblehead.
You're modeling a breakpoint, not a pair of alignments.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
tedsharpe: It's fine. I just misread it.
The code does assume that if it's not a discordant pair, it must be a split read, however. Might be a little fragile.

cwhelan: I changed it to check SplitRead for the split read count..

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
cwhelan: This is an interesting idea that I'm still pondering..

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
tedsharpe: Might also want to test that it's one or the other.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
```diff
@@ -0,0 +1,122 @@
...
+                            nextEvidence instanceof BreakpointEvidence.DiscordantReadPairEvidence
+                                    ? 0 : 1,
+                            nextEvidence instanceof BreakpointEvidence.DiscordantReadPairEvidence
+                                    ? 1 : 0);
+                }
+                currentIntervalsWithTargets.put(updatedLink.source, updatedLink);
+            }
+        }
+
+        for (Iterator<SVIntervalTree.Entry<EvidenceTargetLink>> it = currentIntervalsWithTargets.iterator(); it.hasNext(); ) {
```
tedsharpe: SVIntervalTree implements Iterable, so this could be expressed a little more tersely if you wanted to:
```
        currentIntervalsWithTargets.forEach(entry -> links.add(entry.getValue()));
``

File: src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java [hunk 6]
```diff
@@ -0,0 +1,122 @@
...
+        this.readMetadata = readMetadata;
+    }
+
+    public Iterator<EvidenceTargetLink> cluster(final Iterator<BreakpointEvidence> breakpointEvidenceIterator) throws Exception {
+        final List<EvidenceTargetLink> links = new ArrayList<>();
+        final SVIntervalTree<EvidenceTargetLink> currentIntervalsWithTargets = new SVIntervalTree<>();
+        while (breakpointEvidenceIterator.hasNext()) {
+            final BreakpointEvidence nextEvidence = breakpointEvidenceIterator.next();
+            if (nextEvidence.hasDistalTargets(readMetadata)) {
+                final SVInterval target = nextEvidence.getDistalTargets(readMetadata).get(0);
```
tedsharpe: Unused variable.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
```diff
@@ -0,0 +1,122 @@
...
+                    pairedStrandedIntervalTree.overlappers(reversedPair);
+
+            EvidenceTargetLink newLink = null;
+            while (psiOverlappers.hasNext()) {
+                EvidenceTargetLink existingLink = psiOverlappers.next()._2();
+                newLink = new EvidenceTargetLink(link.target.intersect(existingLink.source), link.targetForwardStrand,
+                        link.source.intersect(existingLink.target), link.sourceForwardStrand,
+                        link.splitReads + existingLink.splitReads, Math.max(link.readPairs, existingLink.readPairs));
+                psiOverlappers.remove();
+                break;
```
tedsharpe: Is this right?  You intended to break here?
If so, the while on line 102 should just be an if, but I'm guessing you intended to process all the overlappers rather than just the first.

cwhelan: You're right, this logic is wrong / incomplete.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryArgumentCollection.java
```diff
@@ -119,6 +119,10 @@
         @Argument(doc = "output dir for assemblies", fullName = "gfaDir", optional = true)
         public String gfaDir;
 
+        @Argument(doc = "file for non-assembled breakpoints in bedpe format",
```
tedsharpe: seems obvious to us, but you might want to say "output file"

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -695,7 +724,56 @@ public AlignedAssemblyOrExcuse apply( final Tuple2<Integer, List<SVFastqUtils.Fa
...
+                    }
+                });
+            } catch ( final IOException ioe ) {
+                throw new GATKException("Can't write target links to "+params.targetLinkFile, ioe);
+            }
+        }
+    }
+
+
+    private static List<BreakpointEvidence> collectAllEvidence( final JavaRDD<BreakpointEvidence> evidenceRDD,
```
tedsharpe: It appears that this is not called.  Should it have been?

cwhelan:
Re: Adding the `SVReadFilter` to the `ReadMetadata` object. It would have been a little bit of a hassle to pass this mapq threshold around -- It'd have to be passed into `BreakpointDensityFilter` and `EvidenceTargetLinkClusterer`. Not the end of the world, though. 

My thought was that the metadata is really tied to the filter's parameters -- if we were to save out the metadata for a bam, for example, it would be helpful to know if the metadata was generated with an override of the default filter values if we wanted to parse and reinterpret it. But maybe that's not something we need to be overly concerned about right now. If you object to adding the filter to the metadata I'll remove it and pass the values around instead.

tedsharpe:
Sorry.  I get your thinking, but I think it would be cleaner to pass the param to each of those two classes.  It cleans up the dependency tree -- those two classes don't depend on any of the ReadMetadata state but for that one param -- and avoids having to have the SVReadFilter be both Java and Kryo serializable.  I'd prefer it, but won't insist upon it.

cwhelan:
No worries, I'll take it out. I didn't like the dual serialization that was necessary either. No big deal.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
cwhelan: Added a validation check in my iteration loop.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
cwhelan: sorry, removed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
cwhelan: I realized that way I'm setting up the data that I'm calling this on, there can never be more than one overlapper. I'll either fix this to make it more clear or re-do it if I decide to take your suggestion to canonicalize the links.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryArgumentCollection.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLink.java
```diff
@@ -0,0 +1,121 @@
...
+    private static final SVInterval.Serializer intervalSerializer = new SVInterval.Serializer();
+
+    final SVInterval source;
+    final boolean sourceForwardStrand;
+    final SVInterval target;
+    final boolean targetForwardStrand;
+    final int splitReads;
+    final int readPairs;
+
+    public EvidenceTargetLink(final SVInterval source, final boolean sourceForwardStrand, final SVInterval target, final boolean targetForwardStrand, final int splitReads, final int readPairs) {
```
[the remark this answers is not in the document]
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
cwhelan: No, sorry. Dev code that I mistakenly left in. Removed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/PairedStrandedIntervals.java
```diff
@@ -0,0 +1,69 @@
+package org.broadinstitute.hellbender.tools.spark.sv.utils;
+
+public final class PairedStrandedIntervals {
```
tedsharpe: I think it might be helpful to have this class more clearly model an uncertain breakpoint, rather than just a pair of intervals with strand.  As it stands, it's not clear what the relationship between the two sequences is.  I think I've figured out that in the existing code the two sequences are extensions of the alignments of a read pair -- and therefore expected to be on opposite strands.
So the data held by this class is fine, but the interpretation of the relationship between left and right could be made simpler and more explicit, I think.
I'd like to see a head sequence and a tail sequence (rather than a left and right) with the strandedness manipulated so that the actual sample sequence over the breakpoint -- if we were able to figure it out exactly -- would be some prefix of the head sequence followed by some suffix of the tail sequence.
This is probably totally unclear, and we can chat.

tedsharpe:
I've tried to study the code that turns a split read into a stranded interval pair so that it can be compared directly to the disparate read pair info, and I can't convince myself that it's correct.

There seem to be a bunch of cases that aren't covered.  For example, it seems to me that the first step would have to be reversing the cigars on negative strand SA's, so that you can see which pairs of SA's are genuinely supplemental to -- and potentially partially overlapping -- each other, and the actual order of the aligned sequences.  (I'm sure this is unclear, too.  We can talk.)

Is there a unit test that mocks the reads we'd expect to get from an inversion breakpoint, both discordant and split.  And that demonstrates that the stranded interval pairs generated from the reads are consistent with each other and with that breakpoint?

cwhelan:
@tedsharpe I've addressed some of your comments here -- all the simpler stuff plus:

- I now only make distal targets for split reads with one supplementary alignment. We can make a ticket to handle more complex cases at some point.
- I renamed the concept of strand in the `EvidenceTargetLink` and related classes -- I'm now calling it `evidenceUpstreamOfBreakpoint`.
- I canonicalize `EvidenceTargetLinks` and only create them when the source is upstream of the target. This allowed me to get rid of the de-duplication code, so thanks for the suggestion.

It seemed tricky to me to try to cluster these links during the initial pass over the reads while at the same time keeping track of coherent evidence. In my testing it doesn't seem like it is slow to run over the `EvidenceRDD` again to do this, but we could think about trying to change this sometime if we're looking for optimizations. 

Want to take another look?

tedsharpe:
It would be nice to have a StrandedInterval class, and then the distal targets could be stored as one of these in the various BreakpointEvidence subclasses that have distal targets, and the PairedStrandedIntervals class could just have two of them.  This would also let you eliminate the methods hasDistalTargets and getDistalTargetsUpstreamOfBreakpoints -- getDistalTargets would do the job of all three.
I assume that PairedStrandedIntervals shouldn't ever have null intervals for source or target, so you could just validate in the constructor and then you wouldn't have to test nullness in equals and hashCode.

cwhelan:
@tedsharpe Thanks for the suggestions, I think it makes the code a bit cleaner. Do these changes match what you were thinking?

tedsharpe:
Definitely cleaner.  Thanks.  Merge away.

PR #3472, opened 2017-08-21
Correct gVCF header lines

DonFreed:
Corrects a couple of allele-specific annotations.


lbergelson:
@DonFreed Thanks for the contribution!

@jamesemery Please review this one.
it looks like org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCallerIntegrationTest.testGVCFModeIsConsistentWithPastResults_AlleleSpecificAnnotations is failing because it checks for an exact header match.

droazen:
Please don't merge this until after the palantir `HaplotypeCaller` evaluation is complete @jamesemery, as this is a change with respect to GATK3.

lbergelson:
@droazen Don't you think palantir could probably work around this change in their evaluation?  It's not like our headers match gatk3 exactly anyway.

jamesemery:
@DonFreed Both of these changes look more correct than what we were doing before. But as louis pointed out two of the integration tests failed on parsing the header. You should fix these tests.

droazen:
I don't want the palantir evaluation to be any more complicated than it needs to be, and I'm also in the process of ripping apart the `HaplotypeCaller` tests in another branch -- let's put this PR on hold until after that's done.

DonFreed:
Sorry I missed the test case. I updated the header of the test case so the pull request passes the integration tests. Thanks!

jamesemery:
@DonFreed Thanks for the fix, though unfortunately as @droazen says the branch is going to have to be on hold for a little while until the Haplotype Caller has been updated.

droazen:
I think we've held this one up long enough -- merging!

DonFreed:
Thanks!

PR #3474, opened 2017-08-22
Updates bwamem-jni depedency to 1.0.2 and adds the possibility of ali…

vruano:
…gning singletons to  BwaEngine classes.

Updates bwamem-jni depedency to 1.0.2 and adds the possibility of aligning singletons to  BwaEngine classes.


vruano:
This is part of a bigger pull-request... just trying to get it into easier to digest self-standing PRs.

src/main/java/org/broadinstitute/hellbender/tools/BwaMemIndexImageCreator.java
```diff
@@ -19,25 +23,51 @@
...
+            if (checkIndexFilesArePresent(referenceFastaLoc)) {
+                BwaMemIndex.createIndexImageFromIndexFiles(referenceFastaLoc, referenceIndexImageOutputLoc);
+                return null;
+            } else {
+                logger.warn(String.format("User requested to use existing index files, " +
+                        "however no complete set of index file (%s.*) was found; " +
+                        "trying to create the image from the reference file directly", referenceFastaLoc));
+            }
+        } else {
+            logger.warn("User requested to use existing index files however no complete set was found");
```
mwalker174: Isn't this line only reached when `useExistingIndexFilesIfAvailable` is false?

src/main/java/org/broadinstitute/hellbender/tools/BwaMemIndexImageCreator.java
```diff
@@ -19,25 +23,51 @@
...
 public final class BwaMemIndexImageCreator extends CommandLineProgram {
 
+    public static final String USE_EXISTING_INDEX_FULL_NAME = "useExistingIndexFilesIfAvailable";
+    public static final String USE_EXISTING_INDEX_SHORT_NAME = "useIndexFiles";
+
     @Argument(fullName = StandardArgumentDefinitions.INPUT_LONG_NAME,
-              shortName = StandardArgumentDefinitions.INPUT_SHORT_NAME,
-              doc = "Input reference fasta file. The five bwa index files are assumed living in the same directory with the same prefix.")
+            shortName = StandardArgumentDefinitions.INPUT_SHORT_NAME,
+            doc = "Input reference fasta file. The five bwa index files are assumed living in the same directory with the same prefix.")
```
mwalker174: I think the second sentence needs to be updated to something like: "The index image will be built directly from the reference FASTA, unless " + USE_EXISTING_INDEX_FULL_NAME + " is true."

src/main/java/org/broadinstitute/hellbender/tools/BwaMemIndexImageCreator.java
```diff
@@ -19,25 +23,51 @@
...
-              doc = "Output reference index image file (ending in \".img\").",
-              optional = true)
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+            doc = "Output reference index image file (ending in \".img\").",
+            optional = true)
     private String referenceIndexImageOutputLoc = null;
 
+    @Argument(fullName = USE_EXISTING_INDEX_FULL_NAME,
+            shortName = USE_EXISTING_INDEX_SHORT_NAME,
+            doc = "If there are some index files present, build the image out of those rather than the reference.",
```
mwalker174: When this is used, the original FASTA is not needed, right? Could we clarify this with "If the index files are present, build the image out of those rather than the reference. The reference FASTA need not be present in this case."

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java
```diff
@@ -48,10 +48,35 @@ public BwaSparkEngine(final JavaSparkContext ctx,
...
+        return align(unalignedReads, true);
+    }
+
+    /**
+     * Performs single-end alignment on a RDD.
+     *
+     * @param unalignedReads the reads to align.
+     * @return never {@code null}.
+     */
+    public JavaRDD<GATKRead> alignSingletons(final JavaRDD<GATKRead> unalignedReads) {
```
tedsharpe: Singleton doesn't seem to me to be quite the right word.  alignReads?  alignUnpairedReads?

src/test/java/org/broadinstitute/hellbender/tools/BwaMemIndexImageCreatorIntegrationTest.java
```diff
@@ -35,4 +37,20 @@ public void testImageFileGeneration() throws Exception {
...
+                "--input", testReferenceFasta.getAbsolutePath(),
+                "--output", imgFile.getAbsolutePath(),
+                "--" + BwaMemIndexImageCreator.USE_EXISTING_INDEX_FULL_NAME));
+        runCommandLine(args);
+        Assert.assertTrue(imgFile.exists());
+        try( final BwaMemIndex index = new BwaMemIndex(imgFile.getAbsolutePath()) ){
+            BwaMemTestUtils.assertCorrectSingleReadAlignment(index);
+            BwaMemTestUtils.assertCorrectChimericContigAlignment(index);
+        }
+        Assert.assertTrue(imgFile.delete());
```
tedsharpe: Maybe wrap this in a finally clause?

vruano:
Heng said that the only thing missing from Apache2's branch is the fast algorithm to create the index for a reference over 2Gb such as the human one, and so that is why it is worth while to keep it for now I guess (we can run bwa index with either algorithm to see how bad it is).

It is easy to remove the feature, if that is what people wanna do. We can talk it thru tomorrow during the stand-up. @broadinstitute/gatk-sv-methods

src/main/java/org/broadinstitute/hellbender/tools/BwaMemIndexImageCreator.java
magicDGS: It is needed for the base name of the index files, I guess...

src/main/java/org/broadinstitute/hellbender/tools/BwaMemIndexImageCreator.java
magicDGS: I think that it is much better to move the "The five bwa index files are assumed living in the same directory with the same prefix." to the `USE_EXISTING_INDEX_FULL_NAME` arguments, and include here what @mwalker174 suggested.

src/main/java/org/broadinstitute/hellbender/tools/BwaMemIndexImageCreator.java
magicDGS: I guess that this else clause should be removed now.

src/main/java/org/broadinstitute/hellbender/tools/BwaMemIndexImageCreator.java
```diff
@@ -19,25 +23,51 @@
...
     @Argument(fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
-              shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
-              doc = "Output reference index image file (ending in \".img\").",
-              optional = true)
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+            doc = "Output reference index image file (ending in \".img\").",
+            optional = true)
     private String referenceIndexImageOutputLoc = null;
 
+    @Argument(fullName = USE_EXISTING_INDEX_FULL_NAME,
```
magicDGS: As pointed by @tedsharpe, this option can lead to bugs if the indexes were created with previous or later versions, that are not coupled with the JNI. If you want to maintain this argument for convenience, I will add the barclay annotation `@Advance` to mark as a __"only use if you know what you are doing"__.

src/main/java/org/broadinstitute/hellbender/tools/BwaMemIndexImageCreator.java
```diff
@@ -19,25 +23,51 @@
...
 
-        BwaMemIndex.createIndexImage(referenceFastaLoc, referenceIndexImageOutputLoc);
+        if (useExistingIndexFilesIfAvailable) {
+            if (checkIndexFilesArePresent(referenceFastaLoc)) {
+                BwaMemIndex.createIndexImageFromIndexFiles(referenceFastaLoc, referenceIndexImageOutputLoc);
+                return null;
+            } else {
+                logger.warn(String.format("User requested to use existing index files, " +
+                        "however no complete set of index file (%s.*) was found; " +
+                        "trying to create the image from the reference file directly", referenceFastaLoc));
```
magicDGS: If the useExistingIndexFilesIfAvailable argument is advanced, I think that it is better to throw an `UserException` to make clear that it is not using previous indexes. I believe that if someone is asking for this option, he should be sure that the indexes are present (and generated with the same version or a compatible one).

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -23,14 +23,23 @@
...
+    public static final String SINGLE_END_ALIGNMENT_FULL_NAME = "singleEndAlignment";
+    public static final String SINGLE_END_ALIGNMENT_SHORT_NAME = "SE";
+
     @Argument(doc = "the output bam", shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
             fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME)
     private String output;
 
     @Argument(doc = "the bwa mem index image file name that you've distributed to each executor",
-            fullName = "bwamemIndexImage")
+            fullName = "bwaMemIndexImage")
```
magicDGS: This should be extracted as a static final `String` for visibility (and maybe re-use).

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -23,14 +23,23 @@
...
     private String indexImageFile;
```
magicDGS: Now that the indexing code is available in the JNI, I think that it will be nice to add a mutex argument to this one to provide the reference sequence instead: if this argument is provided, the pre-generated image will be used; otherwise, it will create the image file from the reference sequence.

Like that, the BwaSpark tool can be self-contained and be used directly with input reads and a reference sequence, without any additional parameter.

src/main/java/org/broadinstitute/hellbender/tools/BwaMemIndexImageCreator.java
```diff
@@ -20,24 +24,22 @@
 public final class BwaMemIndexImageCreator extends CommandLineProgram {
 
     @Argument(fullName = StandardArgumentDefinitions.INPUT_LONG_NAME,
-              shortName = StandardArgumentDefinitions.INPUT_SHORT_NAME,
-              doc = "Input reference fasta file. The five bwa index files are assumed living in the same directory with the same prefix.")
+            shortName = StandardArgumentDefinitions.INPUT_SHORT_NAME,
+            doc = "Input reference fasta file. The five bwa index files are assumed living in the same directory with the same prefix.")
```
tedsharpe: This doc is out of date, I think.  Index files aren't assumed to be living anywhere, or even to exist.

vruano:
Addressed comments. Tests pass.

Notice that now  src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java takes longer to run (3+ minutes) due to the change from creating the image from the index file to using the fasta file. Just giving you the heads-up. @tedsharpe

vruano:
I forgot to update the docs about the no need for index files... will do that before merging, anything else that needs to be changed?

vruano:
Addresses the last comment, once test pass this can be merged.

PR #3475, opened 2017-08-22
Factor out a GATKBaseTest for separate test resources from test utilities in BaseTest

### LINKED ISSUE #2125: Remove GenomeLocParser from BaseTest

### LINKED ISSUE #3029: For projects built on GATK framework, automated testing is broken

magicDGS:
Changes to the testing framework to remove references to the test resources, keeping them into the src/test package. This changes include:

* Factor out a `GATKBaseTest` for separate test resources from test utilities in `BaseTest`
* Remove duplicated `CleanSamIntegrationTest`
* Repackage `CommandLineProgramTest` to be in the test sources, and use it's interface in testers
* Move some testers to the src/test package because they are tool-specific (added TODO to other ones that aren't that clear)
* Refactor `TargetsToolsTestUtils` to use a provided reference

Closes #3029
Closes #2125


magicDGS:
Here is a new PR for the issues related with the test code. Back to you @droazen.

magicDGS:
@droazen, I updated the PR message to describe the changes, because I tried to get all the code referring to test resources packaged in src/test. Now it should pass the tests too.

jamesemery:
@magicDGS Could you rebase this and update newer references to baseTest that have been added since you created this branch?

magicDGS:
@jamesemery - I think that the rebase is done. I'd like to have this in as soon as it can be, to avoid the extra-work of rebasing due to new tests or refactoring of them.... Thank you in advance!

src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
```diff
@@ -168,25 +90,6 @@ protected static String getNonNullEnvironmentVariable(String envVarName) {
...
-        return intervalStringsToGenomeLocs(Arrays.asList(intervals));
-    }
-
-    protected List<GenomeLoc> intervalStringsToGenomeLocs( List<String> intervals ) {
-        List<GenomeLoc> locs = new ArrayList<>();
-        for (String interval: intervals)
-            locs.add(hg19GenomeLocParser.parseGenomeLoc(interval));
-        return Collections.unmodifiableList(locs);
-    }
-
```
jamesemery: For purity sake I think getToolTestDataDir() should be left abstract in BaseTest and instantiated in GATKBaseTest.

src/main/java/org/broadinstitute/hellbender/utils/test/TargetsToolsTestUtils.java
```diff
@@ -15,25 +14,21 @@
...
-    private TargetsToolsTestUtils(){} //don't instantiate this
-
-    /**
-     * Returns a {@link File} pointing to the directory that contains the test data.
-     * @return never {@code null}.
-     */
-    public static File getTestDataDir(){
-        return new File(CommandLineProgramTest.getTestDataDir(),"exome");
+    /** Initialize the reference file and dictionary to use for creating intervals. */
+    public TargetsToolsTestUtils(final File referenceFile){
```
jamesemery: It should not be the case that you are ever instantiating a utils class.

jamesemery: I think if you are trying to avoid calling = new File(getTestDataDir(),"test_reference.fasta"), create a getter which farms out to GATKBaseTest getTestDataDir() to get these files and leave it to external tool authors to avoid using these files.

src/main/java/org/broadinstitute/hellbender/utils/test/TargetsToolsTestUtils.java
```diff
@@ -77,7 +72,8 @@ public static SimpleInterval createOverEntireContig(final String contig) {
      * @return never {@code null}.
      * @throws UserException if there was some problem when creating the location.
      */
-    public static SimpleInterval createInterval(final String contig, final int start) {
+    public SimpleInterval createInterval(final String contig, final int start) {
+        // TODO: should this really be createInterval(contig, start, start) instead of using the constructor supplied here?
```
jamesemery: I agree with you, this method is kindof pointless, it seems to only be called in one place and furthermore it seems to be used in one place, and given the the other overload it seems confusing to just use SimpleInterval

src/main/java/org/broadinstitute/hellbender/utils/test/testers/CleanSamIntegrationTest.java
```diff
@@ -1,58 +0,0 @@
...
-import picard.sam.CleanSam;
-import org.testng.Assert;
-
-import java.io.PrintWriter;
-import java.util.*;
-
-/**
- * This class is the extension of the SamFileTester to test CleanSam with SAM files generated on the fly.
- */
-public final class CleanSamIntegrationTest extends SamFileTester {
```
jamesemery: Nice catch on the duplicated code

src/main/java/org/broadinstitute/hellbender/utils/test/testers/AbstractMarkDuplicatesTester.java
```diff
@@ -50,7 +53,7 @@ public File getMetricsFile() {
     }
 
     @Override
-    public String getTestedClassName() { return getProgram().getClass().getSimpleName(); }
+    public final String getTestedToolName() { return getProgram().getClass().getSimpleName(); }
```
jamesemery: Why did you make this change in a few places?

src/main/java/org/broadinstitute/hellbender/utils/test/testers/AbstractMarkDuplicatesTester.java
```diff
@@ -19,6 +19,9 @@
  * This class is an extension of SamFileTester used to test AbstractMarkDuplicatesCommandLineProgram's with SAM files generated on the fly.
  * This performs the underlying tests defined by classes such as AbstractMarkDuplicatesCommandLineProgramTest.
  */
+// TODO: is this really necessary for the packaged testing framework?
+// TODO: it looks like that this is a SamFileTester exclusive for MarkDuplicates with the parameter in GATK/Picard
+// TODO: and thus, it should live in the test resources
 public abstract class AbstractMarkDuplicatesTester extends SamFileTester {
```
jamesemery: I agree, though I would leave it for a separate pull request to make that change for the testers more broadly.

src/main/java/org/broadinstitute/hellbender/utils/test/TargetsToolsTestUtils.java
jamesemery: More specifically, refactor the other methods in this class to take the reference dictionary as an argument and have the REFERENCE_FILE just be a string pointer to the file. Utils classes should not have state.

src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
magicDGS: This can be overridden, but I think that it is a good practice to have a standard folder structure for test data directories. In addition, this is not project specific, because the java source structure (`src/test/resources`) is more or less an standard.

No action for this comment.

src/main/java/org/broadinstitute/hellbender/utils/test/TargetsToolsTestUtils.java
magicDGS: This was before an utility class to create `SimpleIntervals` with the reference in the exome test source. It is true that it is not a utils class, but a factory/builder class.

I think that I prefer in this case to hold the dictionary and the refererence to be sure that it correspond to the same one. I renamed to `SimpleIntervalTestFactory` to be clear that it keeps state and it is a factory.

src/main/java/org/broadinstitute/hellbender/utils/test/TargetsToolsTestUtils.java
magicDGS: I removed the TODO and call the `createInterval(contig, start, start)` instead, to be sure that it is using the sequence dictionary.

src/main/java/org/broadinstitute/hellbender/utils/test/testers/AbstractMarkDuplicatesTester.java
magicDGS: Changed the TODO to a clear procedure and open a new ticket for move it to the test sources (#3762)

src/main/java/org/broadinstitute/hellbender/utils/test/testers/AbstractMarkDuplicatesTester.java
magicDGS: This classes were implementing `CommandLineProgramTest`, which is the `CommandLineProgramTester` implementation for GATK (extending also `GATKBaseTest`).

Because this classes are in the testing framework, and they are only tester classes, they do not inherit from `BaseTest` anymore. The method for returning the tool name is called `getTestedToolName` in `CommandLineProgramTester` and thus here it should be the one overridden.

Thus, this is one of the results of finally separating the gatk-testing framework (for re-use with downstream projects) and the test from GATK.

src/main/java/org/broadinstitute/hellbender/utils/test/testers/CleanSamIntegrationTest.java
magicDGS: Thank you!

magicDGS:
@jamesemery - we should get this merge as soon as possible to avoid conflicts that pop up in every round of comments. Once this is in, I can go to the open PRs to point out the conflicts and the new structure (e.g., change the new tests to extend `GATKBaseTest`).

I added a new commit addressing the issues and I will rebase to resolve conflicts again.

src/main/java/org/broadinstitute/hellbender/utils/test/TargetsToolsTestUtils.java
jamesemery: This seems like a good solution, you should update the documentation on the file to reflect this change and make it clear to future authors what this class is doing.

jamesemery:
@magicDGS You need to resolve the conflicts yet again and respond to the comments I made about the `SimpleIntervalTestFactory` then this could probably be merged

src/main/java/org/broadinstitute/hellbender/utils/test/TargetsToolsTestUtils.java
magicDGS: Can we left this to a different PR? The main problem with this issue is the `BaseTest` and `GATKBaseTest`, and removing this class by adding new methods to `IntervalUtils` will delay even more the reviewing process...

I open an issue to remove it (https://github.com/broadinstitute/gatk/issues/3771) and I promise to prepare a PR for it once this is in...

magicDGS:
@jamesemery - I rebased to solve the conflict and I open an issue regarding the `SimpleIntervalTestFactory` to do not block this work any longer, because this is already a big change in the testing framework. 

Because the changes are big and we are working in different timezones, conflicts pop up everyday when another PR is accepted before this if they modify any of the test files (which is often the case). If we continue delaying this, it would never be possible to merge...

jamesemery:
@magicDGS Once these conflicts are resolved (which appear to be quick fixes) I will merge this branch.

magicDGS:
Thanks @jamesemery - that's the complication of this big PR. I hope that the tests pass after my last commit and that we can get this in before another PR gets in.

Thanks a lot for reviewing!

lbergelson:
@magicDGS merged!  Thank you for your this.

magicDGS:
Thanks to all of you!

magicDGS:
I tried to warn in every affected open PR as promised @jamesemery

PR #3476, opened 2017-08-22
Structural Variant Context

vruano:
Wrapper around VC object to access SVContext specific annotations.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
@@ -1,0 +1,280 @@
+package org.broadinstitute.hellbender.tools.spark.sv;
+
+import htsjdk.samtools.*;
+import htsjdk.variant.variantcontext.StructuralVariantType;
+import htsjdk.variant.variantcontext.VariantContext;
+import htsjdk.variant.vcf.VCFConstants;
+import org.broadinstitute.hellbender.engine.datasources.ReferenceMultiSource;
+import org.broadinstitute.hellbender.exceptions.GATKException;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.tools.spark.sv.utils.GATKSVVCFConstants;
+import org.broadinstitute.hellbender.tools.spark.sv.utils.GATKSVVCFHeaderLines;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.haplotype.Haplotype;
+import org.broadinstitute.hellbender.utils.param.ParamUtils;
+import org.broadinstitute.hellbender.utils.reference.ReferenceBases;
+import org.broadinstitute.hellbender.utils.variant.GATKVCFConstants;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Supplier;
+
+/**
+ * Variant context with additional method to mind the structural variant specific information from structural
+ * variant records.
+ */
+public class StructuralVariantContext extends VariantContext {
+
+    private static final long serialVersionUID = 1L;
+
+    private int end = -1;
+    private int length = -1;
+
+    /**
+     * Returns a instance given a plain {@link VariantContext} instance.
+     * <p>
+     *     This method will fail with a {@link IllegalArgumentException} if the input variant context is not
+     *     structural. Currently we only check that it has been properly annotated with the {@value VCFConstants#SVTYPE}
+     *     annotation.
+     * </p>
+     * @param vc the input variant context.
+     * @throws IllegalArgumentException if {@code vc} is {@code null} or does not seem to be
+     * a structural variant context based on its annotations.
+     * @return never {@code null}, potentially the same as the input if it happens to be an instance of this class.
+     */
+    public static StructuralVariantContext create(final VariantContext vc) {
+        if (vc instanceof StructuralVariantContext) {
+            return (StructuralVariantContext) vc;
+        } else {
+            checkIsStructuralVariantContext(vc);
+            return new StructuralVariantContext(vc);
+        }
+    }
+
+    private static void checkIsStructuralVariantContext(final VariantContext vc) {
+        Utils.nonNull(vc, "the input variant context must not be null");
+        Utils.nonNull(vc.getAttributeAsString(VCFConstants.SVTYPE, null), "the SVTYPE annotation is missing");
+    }
+
+    private StructuralVariantContext(final VariantContext other) {
+        super(other);
+    }
+
+    /**
+     * The end position of this variant context.
+     * This end position is determined with this priorityq:
+     * <ul>
+     *     <li>the value of the annotation {@value VCFConstants#END_KEY} if present, otherwise</li>
+     *     <li>the value that results from applying the length annotated under {@value  GATKSVVCFConstants#SVLEN} based on the structural type if such annoation is present, otherwise</li>
+     *     <li>the variant context start position + the length of the reference allele - 1.</li>
+     * </ul>
+     * @return whatever is returned by {@link #getStart()} or greater.
+     */
+    @Override
+    public int getEnd() {
+        if (end <= 0) {
+            if (hasAttribute(VCFConstants.END_KEY)) {
+                end = getAttributeAsInt(VCFConstants.END_KEY, -1);
+            } else if (getStructuralVariantType() == StructuralVariantType.DEL) {
+                end = getStart() + getStructuralVariantLength();
+            } else {
+                end = super.getEnd();
+            }
+        }
+        return end;
+    }
+
+    /**
+     * Returns the assembly ids for this context's structural variant.
+     * <p>
+     *     The list returned is an immutable list.
+     * </p>
+     * @throws IllegalStateException if the {@link GATKSVVCFConstants#CONTIG_NAMES} annotation contains
+     * undefined contig names (.)
+     *
+     * @return never {@code null}, an empty list if no structural variant is specified.
+     */
+    public List<String> getContigNames() {
+        if (!hasAttribute(GATKSVVCFConstants.CONTIG_NAMES)) {
+            return Collections.emptyList();
+        } else {
+            final List<String> contigNames = getAttributeAsStringList(GATKSVVCFConstants.CONTIG_NAMES, null);
+            if (contigNames.contains(null)) {
+                throw new IllegalStateException("the contig names annotation contains undefined values");
+            }
+            return contigNames;
+        }
+    }
+
+    /**
+     * Composes the haplotype that corresponds to an allele based on the reference sequence.
+     * <p>
+     *     The resulting haplotype will be annotated with the appropriate Cigar and genome location.
+     * </p>
+     * @param index the index of the target allele.
+     * @param paddingSize extra bases from the reference sequence to be added on either side.
+     * @param reference the reference to use as source.
+     * @return never {@code null}.
+     */
+    public Haplotype composeHaplotypeBasedOnReference(final int index, final int paddingSize, final ReferenceMultiSource reference)  {
+        Utils.nonNull(reference, "the input reference cannot be null");
+        ParamUtils.isPositiveOrZero(paddingSize, "the input padding must be 0 or greater");
+        ParamUtils.inRange(index, 0, 1, "the input allele index must be 0 or 1");
+
+        final SAMSequenceDictionary dictionary = reference.getReferenceSequenceDictionary(null);
+        if (dictionary == null) {
+            throw new IllegalArgumentException("the input reference does not have a dictionary");
+        }
+        final SAMSequenceRecord contigRecord = dictionary.getSequence(getContig());
+        if (contigRecord == null) {
+            throw new IllegalArgumentException("the input reference does not have a contig named: " + getContig());
+        }
+        final int contigLength = contigRecord.getSequenceLength();
+        if (contigLength < getEnd()) {
+            throw new IllegalArgumentException(String.format("this variant goes beyond the end of "
+                    + "the containing contig based on the input reference: "
+                    + "contig %s length is %d but variant end is %d", getContig(), contigLength, getEnd()));
+        }
+        final SimpleInterval referenceInterval = new SimpleInterval(getContig(),
+                Math.max(1, getStart() - paddingSize + 1),
+                Math.min(contigLength, getEnd() + paddingSize));
+
+        final ReferenceBases bases;
+        try {
+            bases = reference.getReferenceBases(null, referenceInterval);
+        } catch (final IOException ex) {
+            throw new GATKException("could not read reference file");
+        }
+        if (index == 0) {
+            final Haplotype result = new Haplotype(bases.getBases(), true);
+            result.setCigar(new Cigar(Collections.singletonList(new CigarElement(referenceInterval.size(), CigarOperator.M))));
+            result.setGenomeLocation(referenceInterval);
+            return result;
+        } else { //index == 1
+            switch (getStructuralVariantType()) {
+                case INS:
+                    return composeInsertionHaplotype(bases);
+                case DEL:
+                    return composeDeletionHaplotype(bases);
+                default: // not jet supported.
+                    throw new UnsupportedOperationException("not supported yet");
+            }
+        }
+    }
+
+    private Haplotype composeDeletionHaplotype(final ReferenceBases referenceBases) {
+        final int deletionSize = getStructuralVariantLength();
+        final byte[] resultBases = new byte[referenceBases.getInterval().size() - deletionSize];
+        final int leftPaddingSize = getStart() - referenceBases.getInterval().getStart() + 1;
+        final int rightPaddingSize = referenceBases.getInterval().getEnd() - getStart() - deletionSize;
+        final byte[] referenceBaseBytes = referenceBases.getBases();
+        System.arraycopy(referenceBaseBytes, 0, resultBases, 0, leftPaddingSize);
+        System.arraycopy(referenceBaseBytes, leftPaddingSize + deletionSize, resultBases, leftPaddingSize, rightPaddingSize);
+        final Cigar cigar = new Cigar(Arrays.asList(new CigarElement(leftPaddingSize, CigarOperator.M),
+                new CigarElement(deletionSize, CigarOperator.D),
+                new CigarElement(rightPaddingSize, CigarOperator.M)));
+        final Haplotype result = new Haplotype(resultBases, false);
+        result.setCigar(cigar);
+        result.setGenomeLocation(referenceBases.getInterval());
+        return result;
+    }
+
+    private Haplotype composeInsertionHaplotype(final ReferenceBases referenceBases) {
+        final byte[] insertedSequence = getInsertedSequence();
+        final byte[] referenceBaseBytes = referenceBases.getBases();
+        final byte[] resultBases = new byte[referenceBases.getInterval().size() + insertedSequence.length];
+        final int leftPaddingSize = getStart() - referenceBases.getInterval().getStart() + 1;
+        final int rightPaddingSize = referenceBases.getInterval().getEnd() - getStart();
+        System.arraycopy(referenceBaseBytes, 0, resultBases, 0, leftPaddingSize);
+        System.arraycopy(insertedSequence, 0, resultBases, leftPaddingSize, insertedSequence.length);
+        System.arraycopy(referenceBaseBytes, leftPaddingSize    , resultBases, leftPaddingSize + insertedSequence.length, rightPaddingSize);
+        final Cigar cigar = new Cigar(Arrays.asList(new CigarElement(leftPaddingSize, CigarOperator.M),
+                new CigarElement(insertedSequence.length, CigarOperator.I),
+                new CigarElement(rightPaddingSize, CigarOperator.M)));
+        final Haplotype result = new Haplotype(resultBases, false);
+        result.setCigar(cigar);
+        result.setGenomeLocation(referenceBases.getInterval());
+        return result;
+    }
+
+    /**
+     * Returns the inserted sequence.
+     * <p>
+     *     Currently this method relies on the annotation {@value GATKSVVCFConstants#INSERTED_SEQUENCE}.
+     * </p>
+     * @return {@code null} if there is no inserted sequence.
+     */
+    public byte[] getInsertedSequence() {
+        if (hasAttribute(GATKSVVCFConstants.INSERTED_SEQUENCE)) {
+            final String asString = getAttributeAsString(GATKSVVCFConstants.INSERTED_SEQUENCE, null);
+            if (asString == null) {
+                return null;
+            } else {
+                return asString.getBytes();
+            }
+        } else {
+            return null;
+        }
+    }
+
+    /**
+     * Returns the absolute structural variant length as recorded in the SVLEN annotation.
+     * <p>
+     *     This method always return a 0 or positive value, so for example for a deletion of 10bp
+     *     it will return 10 despite that the SVLEN annotation should be -10.
+     * </p>
+     * @return -1 if there is no SVLEN annotation, 0 or greater otherwise.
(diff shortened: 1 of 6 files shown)


vruano:
This is part of a large piece of work, I'm trying to submit different part in separated self-standing pull-requests. @SHuang-Broad please review.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+                end = getStart() + getStructuralVariantLength();
+            } else {
+                end = super.getEnd();
+            }
+        }
+        return end;
+    }
+
+    /**
+     * Returns the assembly ids for this context's structural variant.
```
SHuang-Broad: In addition, the name of this function is easily confused with `getContig()`. 
I suggest `getSupportingContigs()`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+    public static StructuralVariantContext create(final VariantContext vc) {
+        if (vc instanceof StructuralVariantContext) {
+            return (StructuralVariantContext) vc;
+        } else {
+            checkIsStructuralVariantContext(vc);
+            return new StructuralVariantContext(vc);
+        }
+    }
+
+    private static void checkIsStructuralVariantContext(final VariantContext vc) {
```
SHuang-Broad: Please use existing method `getStructuralVariantType` in `VariantContext`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+     *     In case the padding is set two zero, since the 0-length interval is not valid, this method would
+     *     return an interval included the base just before the break-point.
+     * </p>
+     *
+     * @param padding the padding around the exact location of the break point to be included in the padding interval.
+     * @param dictionary reference meta-data.
+     * @return never {@code null}, potentially 0-length but typically at least one element.
+     */
+    public List<SimpleInterval> getBreakPointIntervals(final int padding, final SAMSequenceDictionary dictionary) {
+        Utils.nonNull(dictionary, "the input dictionary cannot be null");
```
SHuang-Broad: You might also want to make sure that `padding` is not negative.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+        final String contigName = getContig();
+        final int contigLength = dictionary.getSequence(contigName).getSequenceLength();
+        final StructuralVariantType type = getStructuralVariantType();
+        final int start = getStart();
+        if (type == StructuralVariantType.INS) {
+            return Collections.singletonList(new SimpleInterval(contigName,
+                    Math.max(1, padding > 0 ? (start - padding + 1) : start),
+                    Math.min(contigLength, start + padding)));
+        } else if (type == StructuralVariantType.DEL) { // must be <DEL>
```
SHuang-Broad: comment near end of line is redundant

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+            final int length = getStructuralVariantLength();
+            return Arrays.asList(
+                    new SimpleInterval(contigName,
+                            Math.max(1, padding > 0 ? (start - padding + 1) : start) ,
+                            Math.min(contigLength, start + padding)), // left-end.
```
SHuang-Broad: Isn't `start + length` the same as `end` for `DEL`?

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContextUnitTest.java
```diff
@@ -0,0 +1,219 @@
...
+        try (final VCFFileReader reader = new VCFFileReader(VALID_VARIANTS_FILE, false)) {
+            return Utils.stream(reader)
+                    .map(vc -> new Object[]{vc})
+                    .toArray(Object[][]::new);
+        } catch (final Throwable ex) {
+            throw new TestException("could not load the valid context file: " + VALID_VARIANTS_FILE.getAbsolutePath(), ex);
+        }
+    }
+
+    public static ReferenceMultiSource referenceMultiSource() {
```
SHuang-Broad: `private` instead of `public`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+     * Composes the haplotype that corresponds to an allele based on the reference sequence.
+     * <p>
+     *     The resulting haplotype will be annotated with the appropriate Cigar and genome location.
+     * </p>
+     * @param index the index of the target allele.
+     * @param paddingSize extra bases from the reference sequence to be added on either side.
+     * @param reference the reference to use as source.
+     * @return never {@code null}.
+     */
+    public Haplotype composeHaplotypeBasedOnReference(final int index, final int paddingSize, final ReferenceMultiSource reference)  {
```
SHuang-Broad: should have extra parameter pipelineoptions instead of a implicit null below when accessing reference bases

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+        if (contigLength < getEnd()) {
+            throw new IllegalArgumentException(String.format("this variant goes beyond the end of "
+                    + "the containing contig based on the input reference: "
+                    + "contig %s length is %d but variant end is %d", getContig(), contigLength, getEnd()));
+        }
+        final SimpleInterval referenceInterval = new SimpleInterval(getContig(),
+                Math.max(1, getStart() - paddingSize + 1),
+                Math.min(contigLength, getEnd() + paddingSize));
+
+        final ReferenceBases bases;
```
SHuang-Broad: code above this line could be refactored into a "argument checking and get reference interval" utility method (perferably either private or package private to be tested)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+        try {
+            bases = reference.getReferenceBases(null, referenceInterval);
+        } catch (final IOException ex) {
+            throw new GATKException("could not read reference file");
+        }
+        if (index == 0) {
```
SHuang-Broad: This has the implicit assumption that ref allele is at index 0. I remember from the days of fixing HaplotypeCaller this kind of implicit assumptions can give surprises. Please document clearly in the Javadoc for the parameters of this function that `index==0 <=> reference allele`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+            final Haplotype result = new Haplotype(bases.getBases(), true);
+            result.setCigar(new Cigar(Collections.singletonList(new CigarElement(referenceInterval.size(), CigarOperator.M))));
+            result.setGenomeLocation(referenceInterval);
+            return result;
+        } else { //index == 1
```
SHuang-Broad: This demands documentation for the biallelic assumption.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
+package org.broadinstitute.hellbender.tools.spark.sv;
```
vruano: Ok utils it is.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+import org.broadinstitute.hellbender.utils.param.ParamUtils;
+import org.broadinstitute.hellbender.utils.reference.ReferenceBases;
+import org.broadinstitute.hellbender.utils.variant.GATKVCFConstants;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Supplier;
```
vruano: Ain't there anymore.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+
+/**
+ * Variant context with additional method to mind the structural variant specific information from structural
```
vruano: The latter, thanks.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+ * variant records.
+ */
+public class StructuralVariantContext extends VariantContext {
```
vruano: I'm gland that VC is not final otherwise I would need to use delegation... however as a matter of principle all classes should be final unless proven to be useful to be otherwise, so will change it to final.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    private int end = -1;
```
vruano: Ok.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+    private int length = -1;
+
+    /**
+     * Returns a instance given a plain {@link VariantContext} instance.
```
vruano: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
vruano: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+    private static void checkIsStructuralVariantContext(final VariantContext vc) {
+        Utils.nonNull(vc, "the input variant context must not be null");
+        Utils.nonNull(vc.getAttributeAsString(VCFConstants.SVTYPE, null), "the SVTYPE annotation is missing");
+    }
+
+    private StructuralVariantContext(final VariantContext other) {
+        super(other);
+    }
+
+    /**
```
vruano: For individual breakends I guess the length is in fact 0 a breakpoint does not have dimensions. We can always add some ```if``` to handle this if needed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+     * The end position of this variant context.
+     * This end position is determined with this priorityq:
```
vruano: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+     * <ul>
+     *     <li>the value of the annotation {@value VCFConstants#END_KEY} if present, otherwise</li>
+     *     <li>the value that results from applying the length annotated under {@value  GATKSVVCFConstants#SVLEN} based on the structural type if such annoation is present, otherwise</li>
```
vruano: done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+     *     <li>the variant context start position + the length of the reference allele - 1.</li>
```
vruano: Ok.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
vruano: ```getSupportingContigIds()``` if you don't mind.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
vruano: To be honest I don't have a grasp on the PipelineOptions stuff yet, we follow your advice.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+     * @param reference the reference to use as source.
+     * @return never {@code null}.
+     */
+    public Haplotype composeHaplotypeBasedOnReference(final int index, final int paddingSize, final ReferenceMultiSource reference)  {
+        Utils.nonNull(reference, "the input reference cannot be null");
+        ParamUtils.isPositiveOrZero(paddingSize, "the input padding must be 0 or greater");
+        ParamUtils.inRange(index, 0, 1, "the input allele index must be 0 or 1");
+
+        final SAMSequenceDictionary dictionary = reference.getReferenceSequenceDictionary(null);
+        if (dictionary == null) {
```
vruano: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+            throw new IllegalArgumentException("the input reference does not have a dictionary");
+        }
+        final SAMSequenceRecord contigRecord = dictionary.getSequence(getContig());
+        if (contigRecord == null) {
```
vruano: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
SHuang-Broad: My bad. I just talked with the Engine team yesterday and they suggest, for tool developers, when `PipelineOptions` are needed, always use `null` and they'll have a smart way of dealing with the case when `null` is not enough. So please keep it the old way. Sorry!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
vruano: What is the concern? that the method code is too long? @SHuang-Broad

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
vruano: In think that given that this is a VC it is guaranteed that ref is at index 0. I imagine those problems where due to assume that the index 0 in order Allele containing objects also meant reference for example Allele array that was populated in a arbitrary manner... that said there is no much harm in calling getAllele(i) to verify, is just a added instructions.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
vruano: For now StructuralVariantContext is intrinsically balletic. What I can do is to make it "explode" if that is not the case.

vruano: create() is going to take care of making sure that is the case before returning a value.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+     * Returns the inserted sequence.
+     * <p>
+     *     Currently this method relies on the annotation {@value GATKSVVCFConstants#INSERTED_SEQUENCE}.
+     * </p>
+     * @return {@code null} if there is no inserted sequence.
+     */
+    public byte[] getInsertedSequence() {
+        if (hasAttribute(GATKSVVCFConstants.INSERTED_SEQUENCE)) {
+            final String asString = getAttributeAsString(GATKSVVCFConstants.INSERTED_SEQUENCE, null);
+            if (asString == null) {
```
vruano: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+        }
+    }
+
+    /**
+     * Returns the absolute structural variant length as recorded in the SVLEN annotation.
+     * <p>
+     *     This method always return a 0 or positive value, so for example for a deletion of 10bp
+     *     it will return 10 despite that the SVLEN annotation should be -10.
+     * </p>
+     * @return -1 if there is no SVLEN annotation, 0 or greater otherwise.
```
vruano: Wouldn't BND have length 0?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
```diff
@@ -0,0 +1,280 @@
...
+     */
+    public int getStructuralVariantLength() {
+        if (length < 0) {
```
vruano: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
vruano: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
vruano: Done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
vruano: It is, also I think I found a bug here... I reckon the right break end should be:  padding + end-end + padding

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
SHuang-Broad: not too long, but seems a utility method would serve the purpose. not a must though.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariantContext.java
SHuang-Broad: Not by my understanding. The VCF spec says that the length is the difference between reference allele and alt allele, so technically for a simple inversion, the length would also be 0, although there are different opinions about this definition. So personally I prefer to have no `END` and `SVLEN` at all for BND records (they have `MATE`'s). We could, for the moment, have a static const specifying a value to indicate that `END` and `SVLEN` don't apply for certain records.

vruano:
@SHuang-Broad take a look and tell me if anything else needs to be changed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/StructuralVariantContext.java
```diff
@@ -0,0 +1,284 @@
...
+            return asString == null ? null : asString.getBytes();
+        } else {
+            return null;
+        }
+    }
+
+    /**
+     * Returns the absolute structural variant length as recorded in the SVLEN annotation.
+     * <p>
+     *     When SVLEN is absent this method returns -1.
```
SHuang-Broad: I suggest changing the "-1"'s mentioned in this doc to 

```
public static final int MISSING_LENGTH = -1; // note currently it is private
```

vruano:
@SHuang-Broad I made a few "last" changes... MISSING_XX constants are only used internally to identify when the values have not been calculated. For length I have NO_LENGTH to indicate that in fact there is no length for that variant. Please take a look; hope to have this merged in today.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/StructuralVariantContext.java
```diff
@@ -127,7 +140,8 @@ public int getEnd() {
      * @param pipelineOptions pipeline options.
      * @return never {@code null}.
      */
-    public Haplotype composeHaplotypeBasedOnReference(final int index, final int paddingSize, final ReferenceMultiSource reference, final PipelineOptions pipelineOptions)  {
+    public Haplotype composeHaplotypeBasedOnReference(final int index, final int paddingSize, final ReferenceMultiSource reference,
+                                                      final PipelineOptions pipelineOptions)  {
```
SHuang-Broad: My last comment about `PipelineOptions` is probably overwelmed by other changes, but here's the summary:
My suggestion about adding PipelineOptions are wrong, and please remove them. We need only null for them practically.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/StructuralVariantContext.java
```diff
@@ -82,13 +94,14 @@ private StructuralVariantContext(final VariantContext other) {
...
     @Override
     public int getEnd() {
-        if (end <= 0) {
-            if (hasAttribute(VCFConstants.END_KEY)) {
-                end = getAttributeAsInt(VCFConstants.END_KEY, -1);
-            } else if (getStructuralVariantType() == StructuralVariantType.DEL) {
-                end = getStart() + getStructuralVariantLength();
-            } else {
-                end = super.getEnd();
+        if (end == MISSING_END) {
```
SHuang-Broad: Not suggesting that you change them now (though it will be good if you do so), I still believe that BND records should have no `END` concept associated, so having a `hasEnd()` method is good, and preferably users should call `hasEnd()` first, then `getEnd()`.
But as said, you don't need to add this now.

vruano:
No problem adding hasEnd()... but can you break down what SV types have it which don't .... 

*BUT* I think is important to consider here what "end" means in reality.. I think that "end" here should be the last position continuously overlapped by the variant from its start position. So for insertions, translocations and bnds, typically it would be set equal to start. 

Think about "start" itself.... it does not make reference to the first overlapped based but the ones before it. If a BND would not have an "end" why should it have an "start"?

I think start-end is just defined to what is practical for the sake of working with VCFs.

What do you think? @SHuang-Broad

vruano:
@SHuang-Broad  what is PipelineOptions needed for ... do you need it to access the reference if it stored in something that is not a ordinary file? (e.g. GS bucket?)

SHuang-Broad:
* For `PipelineOptions`, my understanding is that it is used for Google genomics API, and we seem to use it very infrequently in GATK (and never in SV), so it is safe to use null whenever engine level or other utility functions API needs it
* For the `END` and `START` annotation, there is NO`START` in VCF spec, but `POS`, so I don't know where the `start` comes from. And yes, I agree that BND records don't have a `start` either, it is merely a novel adjacency between two genomic locations, none of which is a start or end.

PR #3479, opened 2017-08-22
use correct logger class in MathUtils

lbergelson:
MathUtils was using the log4j1 logger, fixed to use the log4j 2 logger


lbergelson:
@jonn-smith Could you take a look at this super simple PR when you get a chance?

jonn-smith:
Looks fine to me.  I'll merge when checks are done.

PR #3480, opened 2017-08-22
allowing variant walkers to configure their caching behavior

### LINKED ISSUE #3471: add mechanism for VariantWalkers to override FEATURE_CACHE_LOOKAHEAD
Currently all VariantWalkers use `100_000` as the feature cache look-ahead settings.   This works well for most, but is likely causing memory issues for GenotypeGVCFs.

lbergelson:
Adding a new method `getVariantCacheLookAheadBases` to `VariantWalkerBase` which allows subclasses to set how far to look ahead when caching variants.  This may help reduce memory use in GenotypeGVCFs.

This also changes the side inputs to use FeatureDataSource.DEFAULT_QUERY_LOOKAHEAD_BASES which is `1000`, this is the value used by the other tools.  I'm not sure if that's the right thing to do, but it makes variant walkers more consistent with other tools.  Alternatively we could add a separate configuration method that lets tools change the side input value.  We could also expose an optional parameter in the feature input that lets you set that on a per input basis if we need it.  

This doesn't seem to have any negative effect on performance for genotypegvcfs, but it's hard to tell from short runs.  It's also hard to tell if it's improving memory usage.  It doesn't seem to make an appreciable difference at random places in the genome, but I'm hoping it will make a difference in very bad locations that have a lot of variation.  Ideally our caches would be based on size rather than number of variants, but that's a more complicated change.

fixes #3471

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
@@ -29,10 +29,10 @@ import java.util.stream.StreamSupport;
 public abstract class VariantWalkerBase extends GATKTool {
 
     /**
-     * Default value to control the size of the cache for our primary FeatureInputs
-     * (specifically, the number of additional bases worth of overlapping records to cache when querying feature sources).
+     * Default value to control the size of the cache for our driving variants input(s)
+     * (specifically, the number of additional bases worth of overlapping records to cache for driving variants).
      */
-    public static final int DEFAULT_FEATURE_CACHE_LOOKAHEAD = 100_000;
+    public static final int DEFAULT_DRIVING_VARIANTS_LOOKAHEAD_BASES = 100_000;
 
     @Override
     public boolean requiresFeatures() { return true; }
@@ -43,9 +43,15 @@ public abstract class VariantWalkerBase extends GATKTool {
     @Override
     void initializeFeatures() {
 
-        //Note: we override this method because we don't want to set feature manager to null if there are no FeatureInputs.
-        //This is because we have at least 1 source of features (namely the driving dataset).
-        features = new FeatureManager(this, FeatureDataSource.DEFAULT_QUERY_LOOKAHEAD_BASES, cloudPrefetchBuffer, cloudIndexPrefetchBuffer,
+        // We override this method to prevent our FeatureManager from being set to null when no side feature inputs
+        // are specified, since VariantWalkers always have at least one (driving) variants input. Note that the query
+        // lookahead size used here for side inputs is not necessarily the same as for driving variants, which is
+        // determined by {@link #getDrivingVariantCacheLookAheadBases}.
+        //
+        // TODO: I think reducing the lookahead for side inputs from DEFAULT_DRIVING_VARIANTS_LOOKAHEAD_BASES to
+        // FeatureDataSource.DEFAULT_QUERY_LOOKAHEAD_BASES will likely hurt performance for tools like VQSR, but lets
+        // test it
+        features = new FeatureManager(this, DEFAULT_DRIVING_VARIANTS_LOOKAHEAD_BASES, cloudPrefetchBuffer, cloudIndexPrefetchBuffer,
                                       referenceArguments.getReferencePath());
         initializeDrivingVariants();
     }
@@ -98,12 +104,13 @@ public abstract class VariantWalkerBase extends GATKTool {
     protected abstract Spliterator<VariantContext> getSpliteratorForDrivingVariants();
 
     /**
-     * When performing a query on the primary variant input how many overlapping records should be cached when querying.
+     * When performing a query on the driving variants input(s), the number of lookahead bases for which overlapping
+     * variants should be pre-fetched and cached.
      * Subclasses can set this value by overriding this method.
      * @return the number of bases ahead of a query to prefetch
      */
-    protected int getVariantCacheLookAheadBases(){
-        return DEFAULT_FEATURE_CACHE_LOOKAHEAD;
+    protected int getDrivingVariantCacheLookAheadBases(){
+        return DEFAULT_DRIVING_VARIANTS_LOOKAHEAD_BASES;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -95,6 +94,15 @@ public final SAMSequenceDictionary getBestAvailableSequenceDictionary() {
     protected abstract Spliterator<VariantContext> getSpliteratorForDrivingVariants();
 
     /**
+     * When performing a query on the primary variant input how many overlapping records should be cached when querying.
+     * Subclasses can set this value by overriding this method.
+     * @return the number of bases ahead of a query to prefetch
+     */
+    protected int getVariantCacheLookAheadBases(){
+        return DEFAULT_FEATURE_CACHE_LOOKAHEAD;
```
magicDGS: Maybe it is a good idea to add a `TODO`saying that this could be included in the config file...

lbergelson: Good idea, I added a new issue #3489

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -95,6 +94,15 @@ public final SAMSequenceDictionary getBestAvailableSequenceDictionary() {
     protected abstract Spliterator<VariantContext> getSpliteratorForDrivingVariants();
 
     /**
+     * When performing a query on the primary variant input how many overlapping records should be cached when querying.
```
droazen: It's not how many records -- it's how many additional **bases** should be queried and cached beyond the current variant upon a cache miss.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -26,10 +25,10 @@
 public abstract class VariantWalkerBase extends GATKTool {
 
     /**
-     * This number controls the size of the cache for our primary and auxiliary FeatureInputs
+     * Default value to control the size of the cache for our primary FeatureInputs
      * (specifically, the number of additional bases worth of overlapping records to cache when querying feature sources).
      */
-    public static final int FEATURE_CACHE_LOOKAHEAD = 100_000;
+    public static final int DEFAULT_FEATURE_CACHE_LOOKAHEAD = 100_000;
```
droazen: Rename this constant to make it clear it only applies to the driving input

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -42,7 +41,7 @@ void initializeFeatures() {
 
         //Note: we override this method because we don't want to set feature manager to null if there are no FeatureInputs.
         //This is because we have at least 1 source of features (namely the driving dataset).
-        features = new FeatureManager(this, FEATURE_CACHE_LOOKAHEAD, cloudPrefetchBuffer, cloudIndexPrefetchBuffer,
+        features = new FeatureManager(this, FeatureDataSource.DEFAULT_QUERY_LOOKAHEAD_BASES, cloudPrefetchBuffer, cloudIndexPrefetchBuffer,
```
droazen: Add an explicit comment pointing out that you're using a different lookahead value for side inputs.

droazen: Also, you need to do some profiling on `VariantWalkers` with significant side inputs to prove that this doesn't harm performance.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -95,6 +94,15 @@ public final SAMSequenceDictionary getBestAvailableSequenceDictionary() {
     protected abstract Spliterator<VariantContext> getSpliteratorForDrivingVariants();
 
     /**
+     * When performing a query on the primary variant input how many overlapping records should be cached when querying.
+     * Subclasses can set this value by overriding this method.
+     * @return the number of bases ahead of a query to prefetch
+     */
+    protected int getVariantCacheLookAheadBases(){
```
droazen: Name of this method should make it clear that it applies only to the driving input (eg., `getDrivingVariantsCacheLookaheadBases()`)

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -148,6 +155,12 @@
     /** these are used when {@link #onlyOutputCallsStartingInIntervals) is true */
     private List<SimpleInterval> intervals;
 
+
+    @Override
+    protected int getVariantCacheLookAheadBases() {
+        return 1000;
```
droazen: Do heavy-duty profiling to determine the effect on both the memory use and runtime of `GenotypeGVCFs`.
(and bear in mind that for the 20k gnomAD run, it was the native heap memory use by GenomicsDB that was a problem (10+ GB in some cases), whereas the Java heap usage stayed within a fixed and relatively low `-Xmx` value (4-5 GB))

droazen:
@lbergelson Perhaps one way we could get this merged is to merge only the change that allows walkers to override the caching behavior, but keep `GenotypeGVCFs` set to the default for now. Then, you could open a separate ticket to profile different default values for that tool and select a new default.

droazen:
Re-assigning to @cmnbroad

cmnbroad:
I rebased this, and responded to code review comments:

- updated comments
- reverted GenotypeGVCFs change
- reverted changing the default lookahead for VariantWalker side inputs

I think changing the default lookahead for VariantWalker side inputs to the new, smaller value will hurt performance for tools like VQSR. I did some crude timing tests using the FeatureDataSource default (1000 bases) proposed in this branch, and the current default (100,000 bases). The following are total times as reported by Gradle for serial runs of the VQSR integration tests:

With 1000 base lookahead:
1m40s
1m29s
1m29s
1m25s

With 100,000 base lookahead:
1m29s
1m17s
1m16s
1m18s

Back to 1000 base lookahead again:
1m36s
1m26s
1m26s
1m29s

It seems pretty consistently slower with the smaller lookahead.

cmnbroad:
The more I think about it, the more I think we should consider increasing `FeatureDataSource .DEFAULT_QUERY_LOOKAHEAD_BASES` to something larger than 1000.

droazen:
I did a final rebase on this branch to account for the `VariantLocusWalker` and to make a few minor tweaks. Will merge once tests pass.

PR #3482, opened 2017-08-23
Move and rename MarkDuplicates/EstimateLibraryComplexity.

cmnbroad:
Move MarkDuplicates/EstimateLibraryComplexity, tests, and resources:
- from `tools.picard.sam` to `tools.walkers.markduplicates` package
- rename both tools to have a “2” suffix
- for now, these are still `PicardCommandLinePrograms`
- remaining related [classes](https://github.com/broadinstitute/gatk/tree/master/src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates) are not moved or renamed, though most have Picard analogs with the same simple name


droazen:
Closing, as @cmnbroad has incorporated this change into an upcoming branch.

PR #3485, opened 2017-08-23
Fix writing variants to GCS buckets

tomwhite:
Writing reads was fixed in https://github.com/broadinstitute/gatk/commit/73f2a62bee52518b57a985717770ed3a64d83243, but unfortunately the same problem occurs with variants.

This commit (https://github.com/broadinstitute/gatk/commit/a861a23b544012fb8f69358a3999dd587d343547) fixes the problem for variants, when deployed with a Hadoop-BAM fix (https://github.com/HadoopGenomics/Hadoop-BAM/pull/143).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSink.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSink.java
@@ -118,7 +118,7 @@ public final class VariantsSparkSink {
         }
 
         final JavaRDD<VariantContext> sortedVariants = sortVariants(variants, header, numReducers);
-        final String outputPartsDirectory = outputFile + ".parts";
+        final String outputPartsDirectory = outputFile + ".parts/";
         saveAsShardedHadoopFiles(ctx, conf, outputPartsDirectory, sortedVariants,  header, false);
         VCFFileMerger.mergeParts(outputPartsDirectory, outputFile, header);
     }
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSink.java
```diff
@@ -118,7 +118,7 @@ private static void writeVariantsSingle(
         }
 
         final JavaRDD<VariantContext> sortedVariants = sortVariants(variants, header, numReducers);
-        final String outputPartsDirectory = outputFile + ".parts";
```
lbergelson: Would you mind adding a test that writes variants to gcs so we don't break this again?  There's a `PrintReadsSparkIntegrationTest` that does it for bams, a `ReadSparkSinkUnit` test would be ideal.

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/PrintVariantsSparkIntegrationTest.java
```diff
@@ -0,0 +1,42 @@
...
+        };
+    }
+
+    /**
+     * Test the Spark code locally, including GCS access.
+     *
+     * For this to work, the settings in src/main/resources/core-site.xml must be correct,
+     * and the project name and credential file it points to must be present.
+     */
+    @Test(dataProvider = "gcsTestingData", groups = "bucket")
```
lbergelson: thanks, this is great

PR #3486, opened 2017-08-23
Add support for running tools with omitFromCommandLine=true

magicDGS:
Tools with omitFromCommandLine=true can be run after this commit, but
they do not appear in the CLI help (they are hidden).

Related with https://github.com/broadinstitute/barclay/issues/92

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -258,7 +258,16 @@ public class Main {
                 if (null == property) {
                     if (missingAnnotationClasses.isEmpty()) missingAnnotationClasses += clazz.getSimpleName();
                     else missingAnnotationClasses += ", " + clazz.getSimpleName();
-                } else if (!property.omitFromCommandLine()) { /** We should check for missing annotations later **/
+                } else if (property.omitFromCommandLine()) {
+                    // for classes that should be removed from the command line, run directly if is it called in the command line
+                    if (args.length > 1 && clazz.getSimpleName().equals(args[0])) {
+                        try {
+                            return (CommandLineProgram) clazz.newInstance();
+                        } catch (final InstantiationException | IllegalAccessException e) {
+                            throw new RuntimeException(e);
+                        }
+                    }
+                } else { /** We should check for missing annotations later **/
                     if (simpleNameToClass.containsKey(clazz.getSimpleName())) {
                         throw new RuntimeException("Simple class name collision: " + clazz.getSimpleName());
                     }
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -258,15 +258,6 @@ public class Main {
                 if (null == property) {
                     if (missingAnnotationClasses.isEmpty()) missingAnnotationClasses += clazz.getSimpleName();
                     else missingAnnotationClasses += ", " + clazz.getSimpleName();
-                } else if (property.omitFromCommandLine()) {
-                    // for classes that should be removed from the command line, run directly if is it called in the command line
-                    if (args.length > 1 && clazz.getSimpleName().equals(args[0])) {
-                        try {
-                            return (CommandLineProgram) clazz.newInstance();
-                        } catch (final InstantiationException | IllegalAccessException e) {
-                            throw new RuntimeException(e);
-                        }
-                    }
                 } else { /** We should check for missing annotations later **/
                     if (simpleNameToClass.containsKey(clazz.getSimpleName())) {
                         throw new RuntimeException("Simple class name collision: " + clazz.getSimpleName());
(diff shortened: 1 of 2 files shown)


magicDGS:
@SHuang-Broad, here you have the fix for being able to run tools that won't appear in the help. As suggested in the barclay issue, the correct way of annotate this tools are with `@BetaFeature` and do not include `@DocumentedFeature` to do not add them to the help.

magicDGS:
Maybe you can have a look to this one, @cmnbroad!

SHuang-Broad:
@magicDGS Thanks! I've picked the change in one of my branches and it is working perfect from my perspective!

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -258,7 +258,16 @@ private static CommandLineProgram extractCommandLineProgram(final String[] args,
...
-                } else if (!property.omitFromCommandLine()) { /** We should check for missing annotations later **/
+                } else if (property.omitFromCommandLine()) {
+                    // for classes that should be removed from the command line, run directly if is it called in the command line
+                    if (args.length > 1 && clazz.getSimpleName().equals(args[0])) {
+                        try {
+                            return (CommandLineProgram) clazz.newInstance();
+                        } catch (final InstantiationException | IllegalAccessException e) {
+                            throw new RuntimeException(e);
+                        }
+                    }
```
cmnbroad: Rather than short-circuiting all of the logic below (name collision tests, -h, etc.), and duplicating this code, it would be better to just restrict the specific behavior we want changed, which I think is only to not include the tool in printUsage output. I think it would be better to just have printUsage filter them out.

cmnbroad: Also, we should include a test to verify this behavior.

magicDGS:
Done - back to you @cmnbroad

src/test/java/org/broadinstitute/hellbender/MainTest.java
```diff
@@ -1,13 +1,65 @@
...
+
+        private SingleClpMain(final Class<? extends CommandLineProgram> clazz) {
+            this.clazz = clazz;
+        }
+
+        @Override
+        protected List<String> getPackageList() {
+            return Collections.emptyList();
+        }
+
```
cmnbroad: I think lines 25-35 are unnecessary if you make getClassList just return OmitFromCommanLineCLP.class.

cmnbroad: Also, please change OmitFromCommanLineCLP -> OmitFromCommandLineCLP, it has a missing "d" everywhere it appears.

src/test/java/org/broadinstitute/hellbender/MainTest.java
```diff
@@ -1,13 +1,65 @@
...
+    @CommandLineProgramProperties(
+            programGroup = TestProgramGroup.class,
+            summary = "OmitFromCommanLine test",
+            oneLineSummary = "OmitFromCommanLine test",
+            omitFromCommandLine = true)
+    public static final class OmitFromCommanLineCLP extends CommandLineProgram {
+
+        @Override
+        protected Object doWork() {
+            return 1;
```
cmnbroad: Can you make "1" into a public class constant named "OMIT_CLP_RETURN_VALUE" or something similar here, and then use that value in the assert below to check for equality with that same constant.

src/test/java/org/broadinstitute/hellbender/MainTest.java
```diff
@@ -1,13 +1,65 @@
...
+        }
+    }
+
+    @Test
+    public void testClpOmitFromCommandLine() {
+        final SingleClpMain main = new SingleClpMain(OmitFromCommanLineCLP.class);
+        // test that the tool can be run from main correctly (returns non-null)
+        Assert.assertNotNull(main.instanceMain(new String[]{"OmitFromCommanLineCLP"}));
```
cmnbroad: Can you factor out the "OmitFromCommanLineCLP "string into a constant, and also use that in the usage assert below. It reinforces that we're referring to the same value in both places. Also change this to assertEquals as mentioned above.

src/test/java/org/broadinstitute/hellbender/MainTest.java
magicDGS: I implemented the `SingleClpMain` as a helper class for increase the coverage of `Main` related with CLPs.  This allow to test in a more granular way different parts of the code paths handling CLPs, by implementing a new CLP with the required behaviour, and construct a `SingleClpMain` that is ensured to have only one CLP. For instance, this will allow to test the complete usage message. I am preparing another PR to increase the coverage of `Main` and I think that this will be helpful. Let me know what do you think about it.

Regarding the typo, it is fixed in the next commit.

src/test/java/org/broadinstitute/hellbender/MainTest.java
magicDGS: Done.

src/test/java/org/broadinstitute/hellbender/MainTest.java
magicDGS: Done.

src/test/java/org/broadinstitute/hellbender/MainTest.java
cmnbroad: If its going to be used in another PR, lets include/review it there rather than in this PR.

magicDGS: Ok, done.

src/test/java/org/broadinstitute/hellbender/MainTest.java
```diff
@@ -1,13 +1,53 @@
...
 
     @Test(expectedExceptions = UserException.class)
     public void testCommandNotFoundThrows(){
         this.runCommandLine(new String[]{"Brain"});
     }
+
+    @CommandLineProgramProperties(
+            programGroup = TestProgramGroup.class,
+            summary = "OmitFromCommanLine test",
+            oneLineSummary = "OmitFromCommanLine test",
```
cmnbroad: These two instances are still spelled wrong.

magicDGS: Sorry, I don't know what I did to miss this two... Done

magicDGS:
Thanks @cmnbroad!

PR #3495, opened 2017-08-24
Add out/ to .gitignore

magicDGS:
Building with Intellij 2017.2 generates an /out directory instead of using the build/ folder from gradle, that is show as untracked but can be committed by error.

This folder cannot be changed, even if the compiler output is changed (see CrazyCoder comment in https://stackoverflow.com/questions/45174989/building-with-intellij-2017-2-out-directory-duplicates-files-in-build-director).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ gsalib.tar.gz
 gradlew.bat
 tmp/
 src/main/java/local/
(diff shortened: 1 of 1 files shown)


magicDGS:
Thanks @droazen!

Issue #3506, opened 2017-08-24
Mutect2 generates untrimmed REF alleles


sooheelee:
For @davidbenjamin.

I am using the beta.3 stable jar. The data was generated in the Cloud using the WDL script commands from the repo and tool defaults. @vdauwera can share the workspace with you so you can see the actual workflow and download the callsets.

Here are the five calls with untrimmed REF allele representations from my WES callset:
```
WMCF9-CB5:8640dac1_waf_wpon shlee$ gzcat other_waf_wpon.vcf.gz | grep -v '#' | less
chr2    38856503        .       CT      TT      .       germline_risk;panel_of_normals  DP=78;ECNT=1;IN_PON;NLOD=2.87;N_ART_LOD=-2.323e-01;POP_AF=1.000e-03;P_GERMLINE=-1.812e-01;TLOD=6.41     GT:AD:AF:MBQ:MCL:MFRL:MMQ:MPOS:OBAM:OBAMRC:SA_MAP_AF:SA_POST_PROB       0/0:24,4:0.213:13,19:0,0:215,172:60,60:8,8:false:false  0/1:30,8:0.249:0,23:0,0:0,183:0,47:0,12:false:false:0.495,0.00,0.500:0.025,0.028,0.947
chr2    99390427        .       GT      TT      .       panel_of_normals;t_lod  DP=120;ECNT=2;IN_PON;NLOD=6.15;N_ART_LOD=-3.396e-01;POP_AF=2.719e-04;P_GERMLINE=-4.564e+00;TLOD=4.89    GT:AD:AF:MBQ:MCL:MFRL:MMQ:MPOS:OBAM:OBAMRC:SA_MAP_AF:SA_POST_PROB       0/0:58,5:0.136:25,31:0,0:211,240:60,60:22,20:false:false        0/1:39,6:0.186:26,31:0,0:235,216:60,60:22,26:false:false:0.202,0.00,0.200:0.011,0.053,0.936
[...]

PR #3509, opened 2017-08-24
clip M2 alleles before emitting in case some alleles were dropped

davidbenjamin:
@sooheelee This is the issue you brought to my attention, and the fix is the method you suggested.  Would you mind reviewing?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
@@ -32,6 +32,7 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import org.broadinstitute.hellbender.utils.MathUtils;
+import org.broadinstitute.hellbender.utils.variant.GATKVariantContextUtils;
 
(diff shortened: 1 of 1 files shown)


davidbenjamin:
@takutosato giving this to you.

takutosato:
@davidbenjamin could you describe the fix, perhaps with an example?

sooheelee:
I thought I gave an example, with data, in the original issue ticket?

sooheelee:
Here it is again:
https://github.com/broadinstitute/gatk/issues/3506

sooheelee:
And here is the VCF and index: 
[unclipped.tar.gz](https://github.com/broadinstitute/gatk/files/1290857/unclipped.tar.gz)

davidbenjamin:
@takutosato Here's the basic idea: suppose you have a homopolymer context AAA with two potential alt alleles (that is, after assembly), deletions of one and two bases.  Then the ref is AAA and the alts are A and AA.  Now suppose the two-base deletion has too low of a lod to emit.  Then we emit the ref AAA and the alt AA.

Because the A allele is no longer present, this representation is incorrect.  Instead, the ref should be AA and the alt A.  This is what the new clipping line of code does.

takutosato:
@davidbenjamin @sooheelee got it, thanks for the explanation.

PR #3515, opened 2017-08-25
Implemented watershed algorithm for finding local minima in 1D data based on topological persistence.

samuelklee:
This bit of code will be used for the new segmentation method, but could be more generally useful as well.  I put it in a utils package within the copynumber package for now, but we can put it somewhere else if more appropriate.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizer.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizer.java
@@ -1,0 +1,208 @@
+package org.broadinstitute.hellbender.tools.copynumber.utils.optimization;
+
+import org.broadinstitute.hellbender.utils.Utils;
+
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Given 1-dimensional data, finds all local minima sorted by decreasing topological persistence.
+ * This can be useful for finding the most significant local minima in noisy data and can be used
+ * to smooth out less significant local minima that may be induced by noise.
+ * Algorithm is adapted from
+ * <a href="https://people.mpi-inf.mpg.de/~weinkauf/notes/persistence1d.html">
+ *     https://people.mpi-inf.mpg.de/~weinkauf/notes/persistence1d.html</a>.
+ *
+ * @author Samuel Lee &lt;slee@broadinstitute.org&gt;
+ */
+public final class PersistenceOptimizer {
+    private static final int NO_COLOR = -1;
+
+    //represents a connected component bounded by leftIndex and rightIndex with minimum value minValue at minIndex
+    //(may also contain additional local minima)
+    private static final class Component {
+        private int leftIndex;
+        private int rightIndex;
+        private final int minIndex;
+        private final double minValue;
+
+        //Components always initially contain only a local minimum
+        private Component(final int minIndex,
+                          final double minValue) {
+            this.leftIndex = minIndex;
+            this.rightIndex = minIndex;
+            this.minIndex = minIndex;
+            this.minValue = minValue;
+        }
+    }
+
+    private static final class ExtremaPair {
+        private final int minIndex;
+        private final int maxIndex;                     //these could be used to find local maxima, if desired
+        private final double persistence;
+
+        private ExtremaPair(final double[] data,
+                            final int datumIndex1,
+                            final int datumIndex2) {
+            if (data[datumIndex1] > data[datumIndex2]) {
+                minIndex = datumIndex2;
+                maxIndex = datumIndex1;
+            } else if (data[datumIndex1] < data[datumIndex2]) {
+                minIndex = datumIndex1;
+                maxIndex = datumIndex2;
+            } else if (datumIndex1 < datumIndex2) {     //if values are equal, sort by index
+                minIndex = datumIndex1;
+                maxIndex = datumIndex2;
+            } else {
+                minIndex = datumIndex2;
+                maxIndex = datumIndex1;
+            }
+            this.persistence = data[maxIndex] - data[minIndex];
+        }
+    }
+
+    private final double[] data;
+    private final List<Integer> minimaIndices;
+    private final List<Double> persistences;
+
+    /**
+     * Identifies the local minima of {@code data} based on topological persistence upon construction.
+     */
+    public PersistenceOptimizer(final double[] data) {
+        Utils.nonNull(data);
+        Utils.validateArg(data.length > 0, "Data must contain at least one element.");
+
+        this.data = Arrays.copyOf(data, data.length);   //defensive copy
+        final List<Integer> sortedIndices = IntStream.range(0, this.data.length).boxed()
+                .sorted(Comparator.comparingDouble(i -> this.data[i]))
+                .collect(Collectors.toList());
+        final List<ExtremaPair> extremaPairs = findExtremaPairs(this.data, sortedIndices);
+        minimaIndices = extremaPairs.stream().map(p -> p.minIndex).collect(Collectors.toList());
+        minimaIndices.add(0, sortedIndices.get(0));     //add index of global minimum
+        persistences = extremaPairs.stream().map(p -> p.persistence).collect(Collectors.toList());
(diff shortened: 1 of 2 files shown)


samuelklee:
@asmirnov239 Please review!

src/test/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizerUnitTest.java
```diff
@@ -0,0 +1,87 @@
...
+                .toArray();
+        final List<Integer> minimaIndicesExpectedSinusoidal = Arrays.asList(
+                45, 94, 20, 68, 0, 82, 43, 22, 73, 31, 55, 79);                                         //from implementation at https://people.mpi-inf.mpg.de/~weinkauf/notes/persistence1d.html
+
+        rng.setSeed(RANDOM_SEED);
+        final double[] dataPolynomial = IntStream.range(0, numPoints).boxed()
+                .mapToDouble(i -> (double) i / numPoints - 0.5)
+                .map(x -> (x - 0.75) * (x - 0.5) * (x - 0.25) * (x - 0.1) * (x + 0.25) * (x + 0.5) + 0.0001 * rng.nextGaussian())
+                .toArray();
+        System.out.println(Arrays.stream(dataPolynomial).boxed().map(Object::toString).collect(Collectors.joining(", ")));
```
samuelklee: Oops!  Reminder to self to remove this.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizer.java
```diff
@@ -0,0 +1,208 @@
...
+                } else if (leftColor == NO_COLOR && rightColor != NO_COLOR) {   //extend component on right to include this point
+                    extendComponent(components, colors, rightColor, index);
+                } else {                                                        //local maximum, merge components
+                    if (components.get(rightColor).minValue < components.get(leftColor).minValue) {
+                        extremaPairs.add(new ExtremaPair(data, components.get(leftColor).minIndex, index));
+                    } else {
+                        extremaPairs.add(new ExtremaPair(data, components.get(rightColor).minIndex, index));
+                    }
+                    mergeComponents(components, colors, leftColor, rightColor);
+                    colors[index] = colors[index - 1];                          //local maxmimum takes color of component to left by convention
```
asmirnov239: maxmimum -> maximum

asmirnov239: I'm a little confused here. Isn't the local maximum a part of a larger component at this point in the code, color of which is determined by which of the merged components has the lowest minimum? Also can't we just skip recoloring of the local maximum as we skip recoloring of non-boundary component values in the `mergeComponents` method?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizer.java
```diff
@@ -0,0 +1,208 @@
...
+            indexToKeep = index1;
+            indexToMerge = index2;
+        } else {
+            indexToKeep = index2;
+            indexToMerge = index1;
+        }
+        colors[components.get(indexToMerge).leftIndex] = indexToKeep;
+        colors[components.get(indexToMerge).rightIndex] = indexToKeep;
+        if (components.get(indexToKeep).minIndex > components.get(indexToMerge).minIndex) {
+            components.get(indexToKeep).leftIndex = components.get(indexToMerge).leftIndex;
```
asmirnov239: Not removing the `components.get(indexToMerge)` component may introduce some space overhead, which could be linear in most degenerate examples, but I don't see a quick work-around..

src/test/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizerUnitTest.java
```diff
@@ -0,0 +1,87 @@
...
+                .mapToDouble(i -> i)
+                .toArray();
+        final List<Integer> minimaIndicesExpectedIdentity = Collections.singletonList(0);
+
+        final double[] dataIdentityNegative = IntStream.range(0, numPoints).boxed()
+                .mapToDouble(i -> -i)
+                .toArray();
+        final List<Integer> minimaIndicesExpectedIdentityNegative = Collections.singletonList(numPoints - 1);
+
+        final double[] dataFlat = new double[numPoints];
```
asmirnov239: Could you add an extra test case, which has a flat part, but also has a single maxima and minima?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizer.java
samuelklee: Technically true, but space considerations are probably not an issue in most cases (and I think actually removing "merged" components also complicates the bookkeeping a bit).

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizer.java
samuelklee: I don't think the maximum has been assigned a color yet, since we are traversing the points in ascending order, and both `createComponent` and `extendComponent` only assign colors to the point under consideration.

You are right in that assigning a color is not necessary just to find and order the local minima, but it's nice to have all points assigned a color at the end (and that's probably what would reasonably be expected from anyone looking to reuse this code in the future).

But perhaps it would be more consistent to also assign the color within `mergeComponents`; i.e., I'll change the signature of `mergeComponents` so that we call `mergeComponents(components, colors, leftColor, rightColor, index)` and will move the assignment of the color into the method accordingly.

samuelklee: Done.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizer.java
```diff
@@ -0,0 +1,208 @@
...
+     * Returns the corresponding list of the topological persistences of the local minima given by
+     * {@link PersistenceOptimizer#getMinimaIndices()}.
+     */
+    public List<Double> getPersistences() {
+        return Collections.unmodifiableList(persistences);
+    }
+
+    //find extrema pairs via a watershed algorithm
+    private static List<ExtremaPair> findExtremaPairs(final double[] data,
+                                                      final List<Integer> sortedIndices) {
```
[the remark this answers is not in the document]
samuelklee: Expanded the comment.  I have a test case where the entire function is constant, but I'll add one like you describe.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizer.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    private final double[] data;
+    private final List<Integer> minimaIndices;
+    private final List<Double> persistences;
+
+    /**
+     * Identifies the local minima of {@code data} based on topological persistence upon construction.
+     */
+    public PersistenceOptimizer(final double[] data) {
```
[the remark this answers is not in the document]
samuelklee: Done.

src/test/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizerUnitTest.java
samuelklee: Done.

src/test/java/org/broadinstitute/hellbender/tools/copynumber/utils/optimization/PersistenceOptimizerUnitTest.java
samuelklee: Good suggestion---the result of that test is actually a little non-intuitive!  Added some extra cases to boot.

samuelklee:
Addressed review comments, expanded documentation, added some test cases, and some minor renaming/cleanup.  Thanks, back to you @asmirnov239!

asmirnov239:
Looks great @samuelklee!

PR #3525, opened 2017-08-28
Repackage ReadFilter plugin tests

### LINKED ISSUE #2532: ReadFilter plugin tests does not live in the same package than the plugin itself
This comes from pre-Barclay code, but it is an issue when looking for tests for the plugin:

* `ReadFilterPluginUnitTest` lives in org.broadinstitute.hellbender.engine.filters while the plugin lives in `org.broadinstitute.hellbender.cmdline.GATKPlugin`
* In addition, the test is called in a different way that the plugin, which is `GATKReadFilterPluginDescriptor`.

magicDGS:
* Repackage and rename `org.broadinstitute.hellbender.engine.filters.ReadFilterPluginUnitTest` to `org.broadinstitute.hellbender.cmdline.GATKPlugin.GATKReadFilterPluginDescriptorTest`.
* Add `BaseTest` to `GATKReadFilterPluginDescriptorTest`.
* Make the filed `GATKReadFilterPluginDescriptor.userArgs` package-private now that it is possible. 

Closes #2532


magicDGS:
Mind to have a look to this, @cmnbroad? I'm asking you because you were involved in the `ReadFilter` implementation...

PR #3530, opened 2017-08-28
validatevariants -gvcf fix

jsotobroad:
fix issue with validatevariants when validatign a gvcf and a record is followed by a record that is fully encompassed by the first one.  The interval math was off in this case


src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/ValidateVariants.java
```diff
@@ -205,10 +205,13 @@ public void apply(final VariantContext vc, final ReadsContext readsContext, fina
             // if the current record is adjacent to the previous record and "overlap" them if they are so our set is as
             // small as possible while still containing the same bases.
             final int start = (previousInterval != null && previousInterval.overlapsWithMargin(refInterval, 1)) ?
-                    previousInterval.getEnd() : refInterval.getStart();
-            genomeLocSortedSet.add(genomeLocSortedSet.getGenomeLocParser().createGenomeLoc(refInterval.getContig(), start, vc.getEnd()), true);
+                    Math.min(previousInterval.getStart(), refInterval.getStart()) : refInterval.getStart();
```
yfarjoun: will this:
```
Math.min(previousInterval.getStart(), refInterval.getStart()) 
```
not always equal `previousInterval.getStart()` given the order of iteration?

jsotobroad: 👍

jsotobroad:
@lbergelson fix for an incorrect assumption I had made previously, if you or someone else could review thanks!

PR #3533, opened 2017-08-29
Migrate HaplotypeCallerSpark to AssemblyRegionIterator. Fixes #3523.

### LINKED ISSUE #3523: HaplotypeCallerSpark needs to migrate to AssemblyRegionIterator
`AssemblyRegionIterator` has replaced the old `AssemblyRegion.createFromReadShard()`, but `HaplotypeCallerSpark` is still using the latter. It should be migrated to `AssemblyRegionIterator`.

`AssemblyRegionIterator` holds many fewer reads in memory at a time, and also has a number of fixes that make the traversal more consistent with GATK3.x, such as its handling of uncovered loci. These fixes have a major effect on correctness of output.

Once the migration is complete, `AssemblyRegion.createFromReadShard()` can be removed.

tomwhite:


tomwhite:
This change degrades performance on Spark, so should not be merged.

droazen:
@tomwhite I'm looking into the performance issues now with the new code path -- it brings the output much closer to GATK3, but clearly needs some profiling work. Can you tell me what kind of difference you saw in the runtime on Spark? Eg., was it on the order of 20-30%, or was it worse than that?

tomwhite:
@droazen it was a lot worse - running the whole reads pipeline on an exome went from ~50 min to ~180 min.

droazen:
@tomwhite I see! Was this with or without a `-L` interval list?

tomwhite:
@droazen without an interval list.

droazen:
Closing -- this was done in https://github.com/broadinstitute/gatk/pull/4278

Issue #3501, opened 2017-08-24
Prototype a PythonScriptExecutor, and assess maintainability of an example tool that calls into a Python machine-learning library


droazen:
Implement a `PythonScriptExecutor` that is similar to the existing `RScriptExecutor` (invokes Python with a given set of arguments). Then ask the CNV team to prototype an example tool or two that use `PythonScriptExecutor` to call into a Python machine-learning library, and do an assessment of maintainability, etc.

`PythonScriptExecutor` will come with an attached set of conditions for its use, intended to address the most serious issues raised by the engine and support teams with having Python code in the GATK. We should document these conditions in the docs for `PythonScriptExecutor` when it's implemented:

1. All tools that use `PythonScriptExecutor` must have a Java-based front-end, with standard GATK (barclay-based) arguments. We put a lot of development effort into our arg parser and into striving for user-interface consistency across tools, and cannot afford to duplicate this effort in Python. Geraldine (CC'd) and the rest of the support team can back me up on this one!

2. An honest effort should be made to minimize the amount of code written in Python -- as much of each tool's work as possible should be done in Java. In particular, reading/writing final inputs and outputs should happen in Java. This is important for a number of reasons, including the engine team's goal of ensuring universal GCS support, consistent Google authentication handling, etc. Again, we really don't want to have to duplicate that work in Python, or for the tools that call into Python to be inconsistent with the rest of the toolkit.

3. All dependencies (Python and native) of Python libraries used will be clearly documented, and included in the default GATK docker image. I don't think I need to explain why this one is important :)   

4. Before we go any further down this path, we prototype one or two tools using `PythonScriptExecutor`, and do a fair assessment of maintainability and other concerns of the engine/support teams, such as whether it will even be possible to package all dependencies without conflicts.

5. Engine team will continue to search for Java-based solutions while this evaluation is ongoing, but this proposal at least unblocks the CNV team for now.

droazen:
@cmnbroad has volunteered to implement `PythonScriptExecutor` (should be quick), plus an example tool. Then he'll turn things over to @samuelklee and the CNV team to implement a prototype ML tool, after which we'll do the evaluation outlined above and decide whether this is the right path to go down.

mbabadi:
[...]

PR #3536, opened 2017-08-29
Simple PythonScriptExecutor.

cmnbroad:
There are two commits. The first one factores out code that can be shared between the R and Python executors, along with a few opportunistic changes in existing tests that have bad names.

The second has a simple PythonScriptExecutor in the spirit of the RScriptExecutor, along with unit tests, and an example tool and integration test.

First pass for https://github.com/broadinstitute/gatk/issues/3501.


src/main/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutor.java
```diff
@@ -0,0 +1,208 @@
...
+        this(python, ensureExists);
+    }
+
+    /**
+     * @param pythonExecutorName name of the python executable to start
+     * @param ensureScriptEngineExists throw if the python executor cannot be found
+     */
+    public PythonScriptExecutor(final String pythonExecutorName, final boolean ensureScriptEngineExists) {
+        super(pythonExecutorName);
+        if (!pythonExecutorName.equals(python) && !pythonExecutorName.equals(python3)) {
```
lbergelson: If these are the only two options we should have an enum for them.

src/main/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutor.java
```diff
@@ -0,0 +1,208 @@
...
+ * </li></ul>
+ *
+ * NOTE: Serial executions on the same PythonScriptExecutor are each run in a new process. No python state is retained
+ * between command/script/module execution. Using -i doesn't buy you anything (for this version of the executor, at
+ * least) since the process is terminated after each command completes.
+ */
+public class PythonScriptExecutor extends ScriptExecutor<PythonScriptExecutorException> {
+    private static final Logger logger = LogManager.getLogger(PythonScriptExecutor.class);
+
+    private static String python = "python";    // default executor name
```
lbergelson: Shouldn't the be constants at the least?  I think an enum would be better though if we're exposing them as a list of options.

src/main/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutor.java
```diff
@@ -0,0 +1,208 @@
...
+    private static String python3 = "python3";
+    public static String pyExtension = ".py";
+
+    private final List<String> curatedCommandLineArgs = new ArrayList<>();
+
+    /**
+     * @param ensureExists throw if the python executor cannot be located
+     */
+    public PythonScriptExecutor(boolean ensureExists) {
```
lbergelson: Why do we expose this parameter here but not for the RScriptExecutor?

src/main/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutor.java
```diff
@@ -0,0 +1,208 @@
...
+ * NOTE: Serial executions on the same PythonScriptExecutor are each run in a new process. No python state is retained
+ * between command/script/module execution. Using -i doesn't buy you anything (for this version of the executor, at
+ * least) since the process is terminated after each command completes.
+ */
+public class PythonScriptExecutor extends ScriptExecutor<PythonScriptExecutorException> {
+    private static final Logger logger = LogManager.getLogger(PythonScriptExecutor.class);
+
+    private static String python = "python";    // default executor name
+    private static String python3 = "python3";
+    public static String pyExtension = ".py";
```
lbergelson: this should be a constant

src/main/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutor.java
```diff
@@ -0,0 +1,208 @@
...
+     * @param rawArgs raw command line arguments to be passed to the Python process
+     * @return true if the command succeeds, otherwise false
+     */
+    public boolean executeArgs(final List<String> rawArgs) {
+        Utils.nonNull(rawArgs, "Raw args cannot be null");
+
+        // executor name first, followed by rawArgs
+        curatedCommandLineArgs.clear();
+        curatedCommandLineArgs.add(externalScriptExecutorName);
+        rawArgs.forEach(curatedCommandLineArgs::add);
```
lbergelson: probably clearer to use addAll instead

src/main/java/org/broadinstitute/hellbender/utils/runtime/ScriptExecutor.java
```diff
@@ -0,0 +1,142 @@
...
+    public ScriptExecutor(final String externalScriptExecutorName) {
+        Utils.nonNull(externalScriptExecutorName);
+        this.externalScriptExecutorName = externalScriptExecutorName;
+        this.externalScriptExecutorPath = RuntimeUtils.which(externalScriptExecutorName);
+    }
+
+    /**
+     * @return true if the executor exists and can be found on the path
+     */
+    public boolean getExternalExecutorExists() {
```
lbergelson: I think this would be clearer as just `externalExecutorExists()`

src/main/java/org/broadinstitute/hellbender/utils/runtime/ScriptExecutor.java
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Base class for executors that find and run scripts in an external script engine process (R, Python, etc).
+ *
+ * Subclasses must implement:
+ *
+ *   {@link #getApproximateCommandLine}
+ *   {@link #getScriptException}
+ */
+public abstract class ScriptExecutor<T extends ScriptExecutorException> {
```
lbergelson: As far as I can tell this type variable is unnecessary.  Subclasses can already override with a more specific return type without making it a parameterized type.  Maybe I'm missing something more subtle though.

src/test/java/org/broadinstitute/hellbender/tools/examples/ExampleLocusWalkerIntegrationTest.java
```diff
@@ -15,7 +15,7 @@
     private static final String TEST_OUTPUT_DIRECTORY = publicTestDir + "org/broadinstitute/hellbender/tools/examples/";
 
     @Test
-    public void testExampleVariantWalker() throws IOException {
+    public void testExampleLocusWalker() throws IOException {
```
lbergelson: heh

src/test/java/org/broadinstitute/hellbender/utils/R/RScriptExecutorUnitTest.java
```diff
@@ -34,17 +34,17 @@ public void testExistingScript() {
     }
 
     @Test(groups = {"R"}, dependsOnMethods = "testRscriptExists", expectedExceptions = RScriptExecutorException.class)
-    public void testNonExistantScriptException() {
+    public void testNonExistentScriptException() {
         RScriptExecutor executor = new RScriptExecutor();
-        executor.addScript(new File("does_not_exists.R"));
+        executor.addScript(BaseTest.getSafeNonExistentFile("does_not_exists.R"));
```
lbergelson: 👍

src/test/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutorUnitTest.java
```diff
@@ -0,0 +1,124 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public class PythonScriptExecutorUnitTest extends BaseTest {
+    final String HELLO_WORLD_SCRIPT = "print \"hello, world\"";
+
+    @Test(groups = {"Python"})
```
lbergelson: I think all of our other groups are lowercase maybe this one should be too?

src/test/java/org/broadinstitute/hellbender/tools/examples/ExampleLocusWalkerIntegrationTest.java
magicDGS: That was my fault, sorry...

src/main/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutor.java
cmnbroad: I'm not really sure if we even need both of these, but the gatk docker image seems to include both, so I'll leave it for now and make an enum.

src/main/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutor.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutorUnitTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutor.java
cmnbroad: It seemed like many/most of the RScriptExecutor use cases were for optional processing like post-traversal plotting, for which failure isn't catastrophic, but the Python use cases seem to have a hard requirement on Python. This was to make that dependency explicit and easy to express. Having said that, its makes sense to keep them as symmetric as possible, so I'll add a similar constructor to the RScriptExecutor.

PR #3538, opened 2017-08-31
Added `GetSampleName` as stopgap until we have named parameters 

LeeTL1220:
- Added new tool called `GetSampleName`, which will get the sample name from a bam file header.  
- If there is not exactly one sample name in the bam file, this tool will fail, **by design**.
- Altered the WDL to use this tool instead of a call to samtools in the M2 WDL.


LeeTL1220:
@davidbenjamin Just added an additional automated test thanks to @sooheelee

src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.stream.Collectors;
+
+@CommandLineProgramProperties(
+        oneLineSummary = "(EXPERIMENTAL) Emit a single sample name from the bam header into an output file.",
+        summary = "If the bam has zero or more than one sample names in the header, this tool will error, by design.\n" +
+                "  This tool has not been tested extensively.  Most options supported by the GATK are irrelevant for this tool.",
+        programGroup = QCProgramGroup.class
+)
+@BetaFeature
+final public class GetSampleName extends GATKTool{
```
davidbenjamin: Why not make this a `CommandLineProgram`?  BTW, if the answer is that emulating `getHeaderForReads` would be too cumbersome, I accept it with no further argument.

LeeTL1220: @davidbenjamin That would be my answer.  I'd be rewriting that code.

LeeTL1220: No action

src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
```diff
@@ -0,0 +1,65 @@
...
+    )
+    protected File outputSampleNameFile;
+
+    @Override
+    public void traverse() {
+        // Do nothing!
+    }
+
+    @Override
+    public boolean requiresReads() {return true;}
```
pshapiro4broad: Even trivial methods should be formatted on different lines. e.g.
```
public boolean requiresReads() {
    return true;
}
```

src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public void onTraversalStart() {
+
+        // Grab the header info
+        if ((getHeaderForReads() == null) || (getHeaderForReads().getReadGroups() == null)) {
+            throw new UserException.BadInput("The given input bam has no header or no read groups.  Cannot determine a sample name.");
+        }
+
+        final List<String> sampleNames = getHeaderForReads().getReadGroups().stream().map(s -> s.getSample()).distinct().collect(Collectors.toList());
```
pshapiro4broad: This could be simplified slightly if you collect as `Set` instead of `List`, which will let you avoid `distinct()`. Also I would prefer `SAMReadGroupRecord::getSample` over `s -> s.getSample()` for readability even though the lambda form is fewer characters.

src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
```diff
@@ -0,0 +1,65 @@
...
+        if (sampleNames.size() > 1) {
+            throw new UserException.BadInput("The given input bam has more than one unique sample name: " + StringUtils.join(sampleNames, ", "));
+        }
+
+        if (sampleNames.size() == 0) {
+            throw new UserException.BadInput("The given bam input has no sample names.");
+        }
+
+        try (final FileWriter fileWriter = new FileWriter(outputSampleNameFile, false)) {
```
pshapiro4broad: Using `Files` is more concise, e.g.
```
try {
    Files.write(outputSampleNameFile.toPath(), sampleNames.get(0).getBytes());
} catch (final IOException e) {
    ...
```
Changing `outputSampleNameFile` to be declared as `Path` would make this shorter as well.

LeeTL1220: @pshapiro4broad Thanks, but get me before I merge!

src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
LeeTL1220: @pshapiro4broad Note that I just let intelliJ write that...

PR #3539, opened 2017-08-31
Fix BucketUtils#dirSize on GCS. Fixes #3437.

### LINKED ISSUE #3437: fix spark performance regression on jenkins
We see a massive (18x) slowdown of our spark performance tests on jenkins.  The failures started on July 20th.

The last good build was of a21447f.  Which leaves one of:
 - 4c697e06ea33c9179840c81c843658442c82a951:  Move to google-cloud-java snapshot with more robust retries, and set …  
 - 1bc0bbfc5a2240e85fd4b9f9010673c7242552a0 Filter Mutect2 artifacts that arise from apparent-duplicate reads

as the culprit.  It seems more likely that the google cloud changes are causing the slowdown.
It seems like the slowdown is happening because of a change in the scattering, going from many partitions to fewer partitions which then all get scheduled on the same shard.
It's not immediately obvious what's causing this

tomwhite:


lbergelson:
@tomwhite 👍 The fix and test look good to me.

PR #3542, opened 2017-09-05
add external evidence stream for CNVs

tedsharpe:


cwhelan:
A quick high-level comment: what if the CNV interval is very large? Depending on the algorithm it could be 10s or 100s of kb. I'm worried that that could create a too-large assembly. One possible solution is that we might want to only create evidence intervals around the proposed breakpoints of the interval, based on adding some amount of slop around each breakpoint (since these inputs are from read-depth based CNV callers that won't be able to identify the breakpoint with single-nucleotide precision). It might take some analysis of the input parameters to figure out the right parameters for this type of approach, which we could do in a subsequent task if you want. Another option would be to offload that type of processing into a different tool depending on the CNV caller being used (ie a GenomeSTRipCNVCallPreprocessor or gCNVCallPreprocessor) if we think that the right approach might vary with the algorithm being used.

tedsharpe:
Maybe if the interval we're given is smaller than 500 bases we assume that the feature describes a CNV breakpoint, and just emit evidence for it.  Otherwise, we assume that the feature describes a CNV region, and emit evidence for each end + or - 100 bases.  How does that sound?

cwhelan:
I guess it depends on how general we want to keep this input path. I think most purely read-depth based callers won't really be able to discover events smaller than 800bp or so (maybe 500bp at the lower limit) with any accuracy. I also don't know of any tools that we're considering that will describe individual breakpoints. 

What about this for a rule: create two intervals for the start and end of the CNV interval + or - 151 bases (allowing a read length of slop). If the two intervals overlap, merge them together into a single evidence interval. 

We could also make the slop amount parameterizable per input file, since different tools might have different characteristics, although that would be a feature we could just make a ticket for until we need it.

tedsharpe:
Sounds good.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -219,6 +228,75 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+                final int featureStart = feature.getStart();
+                final SVLocation featureLocation = new SVLocation(contigID, featureStart);
+                if ( prevLocation.compareTo(featureLocation) > 0 ) {
+                    throw new UserException("Features in " + path + " are not coordinate sorted.");
+                }
+                prevLocation = featureLocation;
+                while ( featureLocation.compareTo(partitionBoundaries[partitionIdx+1]) >= 0 ) {
+                    partitionIdx += 1;
+                }
+                List<BreakpointEvidence> partitionEvidence = evidenceByPartition.get(partitionIdx);
```
cwhelan: Can this be final?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -219,6 +228,75 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+        final List<List<BreakpointEvidence>> evidenceByPartition = new ArrayList<>(nPartitions);
+        for ( int idx = 0; idx != nPartitions; ++idx ) {
+            evidenceByPartition.add(new ArrayList<>());
+        }
+
+        if ( path == null ) {
+            return evidenceByPartition;
+        }
+
+        SVLocation[] partitionBoundaries = new SVLocation[nPartitions+1];
```
cwhelan: Can this be final?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -618,11 +697,18 @@ public AlignedAssemblyOrExcuse apply( final Tuple2<Integer, List<SVFastqUtils.Fa
...
                         }, true)
-                    .mapPartitionsWithIndex( (idx, evidenceItr) -> {
+                    .mapPartitionsWithIndex( (idx, evidenceItr1) -> {
                             final ReadMetadata readMetadata = broadcastMetadata.value();
                             final PartitionCrossingChecker xChecker =
                                     new PartitionCrossingChecker(idx, readMetadata,
                                                                  readMetadata.getMaxMedianFragmentSize());
+                            final Iterator<BreakpointEvidence> evidenceItr2 =
+                                    broadcastExternalEvidenceByPartition.value().get(idx).iterator();
+                            List<Iterator<BreakpointEvidence>> evidenceItrItr = new ArrayList<>(2);
```
cwhelan: final? Also maybe call this `evidenceItrList`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
```diff
@@ -86,6 +91,32 @@ public BreakpointEvidence read( final Kryo kryo, final Input input, final Class<
         }
     }
 
+    @DefaultSerializer(ExternalEvidence.Serializer.class)
+    public final static class ExternalEvidence extends BreakpointEvidence {
+        public ExternalEvidence( final SVInterval interval, final int weight ) {
```
cwhelan: For the other evidence types, we're adding the uncertainty to their intervals here in their constructors in `BreakpointEvidence`, whereas you're adding the uncertainty prior to creating the object. Is that because we've parameterized this uncertainty but not the other types of uncertainty (which are still static constants)? If there's not a technical reason to do so it might be nice to have one place to look for this sort of logic. Not a really big deal, though.

tedsharpe: Yes, good idea.  I'll reuse that code.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -219,6 +228,75 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+                final int featureEnd = feature.getEnd();
+                if ( featureEnd - featureStart <= 2*externalEvidenceUncertainty ) {
+                    addExternalEvidence(partitionEvidence, contigID, featureStart, featureEnd, externalEvidenceWeight);
+                } else {
+                    final int start1 = Math.max(1, featureStart - externalEvidenceUncertainty);
+                    final int end1 = featureStart + externalEvidenceUncertainty;
+                    addExternalEvidence(partitionEvidence, contigID, start1, end1, externalEvidenceWeight);
+                    final int start2 = featureEnd - externalEvidenceUncertainty;
+                    final int end2 = featureEnd + externalEvidenceUncertainty;
+                    addExternalEvidence(partitionEvidence, contigID, start2, end2, externalEvidenceWeight);
```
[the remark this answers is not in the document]
tedsharpe: Good catch.  I'll add the code to make sure it goes into the correct partition's list.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -274,29 +275,35 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+                            BreakpointEvidence.fixedWidthInterval(contigID, featureEnd, externalEvidenceUncertainty);
+                    final SVLocation startLocation2 = interval2.getStartLocation();
+                    int partitionIdx2 = partitionIdx;
+                    if ( startLocation2.compareTo(partitionBoundaries[partitionIdx+1]) >= 0 ) {
+                        partitionIdx2 = Arrays.binarySearch(partitionBoundaries, interval2.getStartLocation());
+                        // Binary search returns the complement of the insert position when the exact key is not found.
+                        // I.e., ~partitionIdx2 will point to the smallest entry larger than the sought key.
+                        // We want the partition before that one, if it exists.
+                        if ( partitionIdx2 < 0 ) {
+                            partitionIdx2 = ~partitionIdx2;
```
cwhelan: Tricky!

PR #3545, opened 2017-09-05
Reduce number of output partitions in PathSeqPipelineSpark

mwalker174:
In PathSeqPipelineSpark, the reads are repartitioned to ~5k per partition (by default) just prior to the pathogen BWA alignment step (to ensure an even distribution of work). Currently, some samples with a lot of non-host reads cause 10,000's of sharded BAMs to be written at the end of the pipeline.

This PR reduces the number of partitions in the read RDD just before writing to disk in the PathSeqPipelineSpark tool. It exposes a command-line option for the number of reads per partition, with a default value that results in a much more reasonable number of sharded BAMs in even the worst cases.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqPipelineSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqPipelineSpark.java
@@ -56,6 +56,13 @@ public class PathSeqPipelineSpark extends GATKSparkTool {
             minValue = 100)
     public int readsPerPartition = 5000;
 
+    @Argument(doc = "Number of reads per partition for output.",
+            fullName = "readsPerPartitionOutput",
+            optional = true,
+            minValue = 100,
+            minRecommendedValue = 100000)
+    public int readsPerPartitionOutput = 1000000;
+
     /**
      * Reduces number of partitions of paired reads, keeping pairs together.
      */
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqPipelineSpark.java
```diff
@@ -56,6 +56,13 @@
             minValue = 100)
     public int readsPerPartition = 5000;
 
+    @Argument(doc = "Number of reads per partition for output.",
+            fullName = "readsPerPartitionOutput",
+            optional = true,
+            minValue = 100,
+            minRecommendedValue = 100000)
+    public int readsPerPartitionOutput = 1000000;
```
lbergelson: It seems like maybe this could be handled through the "numReducers" argument instead of adding a new argument?

SHuang-Broad: Not sure if I understood what @lbergelson is suggesting, but there's the `getRecommendedNumReducers()` method in `GATKSparkTool`.

But Mark's application is facing a non traditional-GATK domain so I trust his judgement.

lbergelson: I just wasn't sure if he knew about that argument.  If it doesn't work for his purposes that's fine, but if he just hadn't seen it maybe it's worth looking at since it does something very similar.

mwalker174: I was unsure whether to use numReducers but on second look I can't find anywhere it's used other than for a sort when writing a single BAM. If that's the case I might as well just use numReducers instead of creating a new option here.

PR #3547, opened 2017-09-06
Make directory name unique for BucketUtilsTest#testDirSizeGCS to avoi…

tomwhite:
…d unwanted test interaction.

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/utils/gcs/BucketUtilsTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/gcs/BucketUtilsTest.java
@@ -139,7 +139,8 @@ public final class BucketUtilsTest extends BaseTest {
     @Test(groups={"bucket"})
     public void testDirSizeGCS() throws IOException, GeneralSecurityException {
         final String src = publicTestDir + "empty.vcf";
-        final String intermediate = BucketUtils.randomRemotePath(getGCPTestStaging(), "dir/test-copy-empty", ".vcf");
+        final String gcsSubDir = BucketUtils.randomRemotePath(getGCPTestStaging(), "dir-", "/");
+        final String intermediate = BucketUtils.randomRemotePath(gcsSubDir, "test-copy-empty", ".vcf");
         BucketUtils.copyFile(src, intermediate);
         Assert.assertTrue(BucketUtils.fileExists(intermediate));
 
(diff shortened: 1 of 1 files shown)


src/test/java/org/broadinstitute/hellbender/utils/gcs/BucketUtilsTest.java
```diff
@@ -149,7 +150,7 @@ public void testDirSizeGCS() throws IOException, GeneralSecurityException {
         Assert.assertEquals(intermediateFileSize, srcFileSize);
         long intermediateDirSize = BucketUtils.dirSize(intermediate);
         Assert.assertEquals(intermediateDirSize, srcFileSize);
-        long intermediateParentDirSize = BucketUtils.dirSize(BucketUtils.getPathOnGcs(intermediate).getParent().toUri().toString());
+        long intermediateParentDirSize = BucketUtils.dirSize(gcsSubDir);
         Assert.assertEquals(intermediateParentDirSize, srcFileSize);
 
         BucketUtils.deleteFile(intermediate);
```
cmnbroad: Should this also delete gcsSubDir ?

tomwhite: Yes, it should. I'm struggling to run this locally (GCS permissions), which I need to do since the Travis turnaround time is too slow.

tomwhite: OK, I have managed to run locally now, and it's working there with this latest fix. It looks like you don't need to explicitly delete gcsSubDir, since when you delete the file in it, the directory disappears (because of the way GCS is not really a filesystem).

Thanks for taking a look @cmnbroad!

cmnbroad: I tried running it locally and was having the same problem. Temporary revert is always an option in the short term if you're time-bound.

tomwhite: I think our messages crossed. Travis is now running the latest fix, so let's see if that passes.

tomwhite:
This is passing now. Are you happy to approve this @cmnbroad?

cmnbroad:
:+1: Thanks @tomwhite.

tomwhite:
Thanks @cmnbroad

PR #3551, opened 2017-09-06
Fix @BeforeSuite usages to prevent excessive code execution

cmnbroad:
There is quite a bit of code annotated with @BeforeSuite that should probably just be @BeforeClass. @BeforeSuite code can run even when no methods in the class are selected in the current suite. Most of it is harmless, but some of it is painfully slow, i.e., using gradle to run a single test method always causes the BwaMemIntegrationTest code to create an index image for the large test reference, which takes 1-2 minutes, before the selected test even starts.

We should probably reserve @BeforeSuite for test-infrastructure setup code; most test classes should use @BeforeClass instead.

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
@@ -2,22 +2,20 @@ package org.broadinstitute.hellbender;
 
 import org.broadinstitute.hellbender.utils.bwa.BwaMemIndex;
 import org.broadinstitute.hellbender.utils.test.BaseTest;
-import org.testng.annotations.AfterSuite;
-import org.testng.annotations.BeforeSuite;
-import org.testng.annotations.Test;
+import org.testng.annotations.*;
 
 public class BwaMemIntegrationTest extends BaseTest {
 
     private BwaMemIndex index;
 
-    @BeforeSuite
+    @BeforeClass
     public void loadIndex() {
         final String imageFile = createTempFile(b37_reference_20_21, ".img").toString();
         BwaMemIndex.createIndexImageFromFastaFile(b37_reference_20_21, imageFile);
         index = new BwaMemIndex(imageFile);
     }
 
-    @AfterSuite
+    @AfterClass
     public void unloadIndex() {
         index.close();
         index = null;
--- a/src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeLikelihoodCalculatorsUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeLikelihoodCalculatorsUnitTest.java
@@ -2,7 +2,7 @@ package org.broadinstitute.hellbender.tools.walkers.genotyper;
 
 import org.broadinstitute.hellbender.utils.test.BaseTest;
 import org.testng.Assert;
-import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.BeforeClass;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
@@ -14,7 +14,7 @@ public final class GenotypeLikelihoodCalculatorsUnitTest extends BaseTest {
 
     GenotypeLikelihoodCalculators calcs;
 
-    @BeforeSuite
+    @BeforeClass
     public void init(){
         calcs = new GenotypeLikelihoodCalculators();
     }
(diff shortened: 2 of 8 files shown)


src/test/java/org/broadinstitute/hellbender/utils/recalibration/QualQuantizerUnitTest.java
```diff
@@ -16,7 +16,7 @@
 
 
 public final class QualQuantizerUnitTest extends BaseTest {
-    @BeforeSuite
+    @BeforeClass
     public void before() {
```
lbergelson: is this an empty method?

cmnbroad: Hm, yeah, that should probably just come out. Even better.

src/test/java/org/broadinstitute/hellbender/tools/exome/sexgenotyper/TargetCoverageSexGenotypeCalculatorUnitTest.java
```diff
@@ -58,34 +59,30 @@
...
-    private static final ReadCountCollection RANDOM_READ_COUNTS_WITHOUT_BAIT_BIAS;
-    private static final List<String> TRUTH_SEX_GENOTYPES_WITHOUT_BAIT_BIAS;
+    private static ReadCountCollection RANDOM_READ_COUNTS_WITHOUT_BAIT_BIAS;
+    private static List<String> TRUTH_SEX_GENOTYPES_WITHOUT_BAIT_BIAS;
 
-    static {
-        final ImmutablePair<List<String>, ReadCountCollection> randomData = generateRandomReadCountCollection(
+    @BeforeClass
+    public void before() {
+        HOMO_SAPIENS_GERMLINE_CONTIG_PLOIDY_ANNOTATIONS = ContigGermlinePloidyAnnotationTableReader
```
lbergelson: This seems worse, why change these?

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeLikelihoodCalculatorsUnitTest.java
```diff
@@ -14,7 +14,7 @@
 
     GenotypeLikelihoodCalculators calcs;
 
-    @BeforeSuite
+    @BeforeClass
     public void init(){
```
lbergelson: I feel like most of these before suite should just be initialized inline where they're declared.  Or replaced with a getter that returns a new instance of the thing on demand.  That's a bigger change though.

src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
```diff
@@ -2,22 +2,20 @@
...
-import org.testng.annotations.Test;
+import org.testng.annotations.*;
 
 public class BwaMemIntegrationTest extends BaseTest {
 
     private BwaMemIndex index;
 
-    @BeforeSuite
+    @BeforeClass
     public void loadIndex() {
```
lbergelson: Does this really take minutes?  that's gross... we should try to change it uses a smaller test file if that wouldn't reduce the power of the test unreasonably.

src/test/java/org/broadinstitute/hellbender/utils/recalibration/QualQuantizerUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/tools/exome/sexgenotyper/TargetCoverageSexGenotypeCalculatorUnitTest.java
cmnbroad: I did this because the static blocks always execute (AFAICT), even if you run a single test from another package, and it calls code that emits log output:

13:15:52.512 INFO  ContigGermlinePloidyAnnotationTableReader - Ploidy tags: SEX_XX, SEX_XY

which looks misplaced depending on the context. But that is a separate problem - my main goal with this PR was to eliminate the BWA indexing, since thats super slow and it happens every time. I'm happy to revert the changes in this file for now.

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypeLikelihoodCalculatorsUnitTest.java
cmnbroad: Agreed. I was trying to minimize the changes, but wanted to get rid of unnecessary @BeforeSuites.

src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
cmnbroad: It seems to take about 2 minutes on my laptop.

SHuang-Broad: I am suspecting this is leading to the new indexing step I see when running tests locally, and does increase the test time by ~3min.

Issue #3562, opened 2017-09-08
BwaMemIntegrationTest takes too long (3-4 minutes).


vruano:
Need to use a smaller reference so that it takes less time. Such a reference shall have more than on contig.

It could totally be made up (fasta created on the flight)

vruano:
Solution merged.

PR #3563, opened 2017-09-10
Changes in BwaMemIntegrationTest to avoid a 3-4 minutes runtime.

vruano:
Moves RandomDNA class into test from main as it is only used for testing.
Added a random reference creation to RandomDNA.
Added isConcrete method to Nucleotide.

Fixes issue #3562.


vruano:
Now it runs under a second at least when run from IntellJ.

Currently it creates a random fasta file with two chromosomes of 10,000bps and check around 1000 perfect + 1000 chimeric alignments.

src/main/java/org/broadinstitute/hellbender/utils/RandomDNA.java
```diff
@@ -1,90 +0,0 @@
...
-/**
- * Random DNA sequence generator.
- *
- * <p>
- *     Returned bases are always in upper case and one of the valid four nocleotides 'A', 'C', 'G' and 'T'.
- * </p>
- *
- * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
- */
-public final class RandomDNA {
```
magicDGS: I think that the `RandomDNA` could be useful for the testing framework packaged into the main jar (the same as other test classes)...

src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
```diff
@@ -1,34 +1,162 @@
...
+    public void loadIndex() throws IOException {
+        final RandomDNA rdnDNA = new RandomDNA(103);
+        try {
+            fastaFile = rdnDNA.nextFasta(TEST_DICTIONARY, 60);
+            fastaFile.deleteOnExit();
+            imageFile = createTempFile(fastaFile.getAbsolutePath(), ".img");
+            fastaIndex = FastaSequenceIndexCreator.buildFromFasta(fastaFile.toPath());
+            BwaMemIndex.createIndexImageFromFastaFile(fastaFile.getAbsolutePath(), imageFile.getAbsolutePath());
+            index = new BwaMemIndex(imageFile.getAbsolutePath());
+        } catch (final Throwable th) {
```
tedsharpe: In general, I think it's a bad idea to catch Throwable.  You can Google it to see if you agree.
You're working very hard to delete two files that are already marked for deleteOnExit.
But I'm not that concerned because this is test code.

src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
```diff
@@ -1,34 +1,162 @@
...
+            if (fastaFile != null) { try { fastaFile.delete(); } catch (final Throwable ex) {}; }
+            if (imageFile != null) { try { imageFile.delete(); } catch (final Throwable ex) {}; }
+            if (index != null) { try { index.close(); } catch (final Throwable ex) {}; }
+            Assert.fail("exception at initialization", th);
```
tedsharpe: I think you should rethrow the original throwable.

src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
```diff
@@ -1,34 +1,162 @@
...
+        }
     }
 
     @AfterClass
     public void unloadIndex() {
+        fastaFile.delete();
```
tedsharpe: These next three lines should probably be in the reverse order.  On Linux it doesn't matter that you're deleting a file that's currently memory-mapped.  But if anyone is running on Windows it'll be a problem, I think.  In general, it's best to tear down in the reverse order of the build up.

src/test/java/org/broadinstitute/hellbender/utils/RandomDNA.java
```diff
@@ -0,0 +1,162 @@
...
+    public File nextFasta(final SAMSequenceDictionary dict, final int basesPerLine)
+        throws IOException {
+        final File result = File.createTempFile("random-", ".fasta");
+        nextFasta(result, dict, basesPerLine);
+        return result;
+    }
+
+    public void nextFasta(final File out, final SAMSequenceDictionary dict, final int basesPerLine)
+            throws IOException {
+        final FileWriter writer = new FileWriter(out);
```
tedsharpe: Wrap in try() so that the close is done automatically?

SHuang-Broad:
I assume you no longer need me bugging you @vruano

vruano:
@SHuang-Broad please bug me any time.

src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
vruano: Changed it to RuntimeException.

src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
vruano: Done.

src/test/java/org/broadinstitute/hellbender/BwaMemIntegrationTest.java
vruano: Done.

src/test/java/org/broadinstitute/hellbender/utils/RandomDNA.java
vruano: Done.

src/main/java/org/broadinstitute/hellbender/utils/RandomDNA.java
vruano: Talked to our engine guys and  currently unless we know that someone it is using it downstream is best to remove it from the main jar. It seems that the fact that there are some test classes in the main is not ideal but result of dependencies issues when we had the project split in public and protected. 

I suspect that eventually we may create a separate artifact (a test jar) for those classes.

magicDGS: I think that this utilities could still leave in the main jar, under the `utils.test` package. The artifact with them is still not done and it may prevent downstream projects to use this useful testing classes for their own purpose. I've already requested this at some point to avoid packaging also TestNG and other testing dependencies, but removing completely some testing framework until it is package properly sounds like a bad idea to re-use the code in the GATK-framework...

Issue #3473, opened 2017-08-22
Improve long read alignment filtering tool


SHuang-Broad:
Now that #3432 is in, which provides a minimal solution to #3221, we need to improve up on the tool and decide how to deal with long reads that map to canonical and non-canonical chromosomes at the same time.

Issue #3431, opened 2017-08-10
Follow up with necessary tests for cpx SV PR series


SHuang-Broad:

SHuang-Broad:
now that tool is up and running. 
Need to provide tests for stability and this needs to be DONE BEFORE the prototyping tool hooked up into master

SHuang-Broad:
closed by #4330 sometime ago.

PR #3571, opened 2017-09-13
PR series for complex SV, part 5

SHuang-Broad:
This PR deals with long reads with exactly two alignments (no other equally good alignment configuration), mapped to 
* the same chromosome with reference order switch but without strand switch, or
* different chromosomes

This brings us (unfiltered) ~6000 mated BND records, half of which are on canonical chromosomes.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/BreakEndVariantType.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/BreakEndVariantType.java
@@ -1,28 +1,56 @@
 package org.broadinstitute.hellbender.tools.spark.sv.discovery;
 
+import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.util.SequenceUtil;
 import htsjdk.variant.variantcontext.Allele;
 import org.apache.commons.lang3.ArrayUtils;
 import org.broadinstitute.hellbender.engine.datasources.ReferenceMultiSource;
 import org.broadinstitute.hellbender.exceptions.GATKException;
-import org.broadinstitute.hellbender.tools.spark.sv.utils.GATKSVVCFConstants;
+import org.broadinstitute.hellbender.utils.IntervalUtils;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
+import scala.Tuple2;
 
 import java.io.IOException;
 import java.util.Collections;
 import java.util.Map;
 
+import static org.broadinstitute.hellbender.tools.spark.sv.utils.GATKSVVCFConstants.*;
+
 public abstract class BreakEndVariantType extends SvType {
 
+    /**
+     * Technically, a BND-formatted variant should have two VCF records, for mates, hence we also have this field.
+     */
     protected final boolean isTheUpstreamMate;
 
     public final boolean isTheUpstreamMate() {
         return isTheUpstreamMate;
     }
 
+    protected static SimpleInterval getMateRefLoc(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc) {
+        return forUpstreamLoc ? narl.leftJustifiedRightRefLoc : narl.leftJustifiedLeftRefLoc;
+    }
+
+    protected static String getIDString(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc) {
+        return BREAKEND_STR + INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
+                narl.leftJustifiedLeftRefLoc.getContig() + INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
+                narl.leftJustifiedLeftRefLoc.getEnd() + INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
+                narl.leftJustifiedRightRefLoc.getContig() + INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
+                narl.leftJustifiedRightRefLoc.getStart() + INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
+                (forUpstreamLoc ? "1" : "2");
+    }
+
+
     @Override
     public final String toString() {
-        return GATKSVVCFConstants.BREAKEND_STR;
+        return BREAKEND_STR;
+    }
+
+    private static Allele constructAltAllele(final byte[] bases, final boolean bracketPointsLeft, final SimpleInterval novelAdjRefLoc,
+                                             final boolean basesFirst) {
+        final String s = bracketPointsLeft ? "]" + novelAdjRefLoc.getContig() + ":" + novelAdjRefLoc.getStart() + "]"
+                : "[" + novelAdjRefLoc.getContig() + ":" + novelAdjRefLoc.getStart() + "[";
+        return Allele.create( basesFirst ? new String(bases) + s  : s + new String(bases) );
     }
 
     // see VCF spec 4.2 or 4.3 for BND format ALT allele field for SV
@@ -33,28 +61,59 @@ public abstract class BreakEndVariantType extends SvType {
         this.isTheUpstreamMate = isTheUpstreamMate;
     }
 
-    private static Allele constructAltAllele(final byte[] bases, final boolean bracketPointsLeft, final SimpleInterval novelAdjRefLoc,
-                                             final boolean basesFirst) {
-        final String s = bracketPointsLeft ? "]" + novelAdjRefLoc.getContig() + ":" + novelAdjRefLoc.getStart() + "]"
-                                           : "[" + novelAdjRefLoc.getContig() + ":" + novelAdjRefLoc.getStart() + "[";
-        return Allele.create( basesFirst ? new String(bases) + s  : s + new String(bases) );
-    }
+    //==================================================================================================================
 
-    static abstract class StrandSwitchBND extends BreakEndVariantType {
+    /**
+     * Breakend variant type for inversion suspects: those with novel adjacency between two reference locations
+     * on the same chromosome but the novel adjacency brings them together in a strand-switch fashion.
+     * This is to be distinguished from the more general "translocation" breakends, which are novel adjacency between
+     * reference locations without strand switch if the reference bases are from the same chromosome.
+     */
+    public static final class InvSuspectBND extends BreakEndVariantType {
+        /**
+         * for breakends, there's a concept of partner (see VCF spec) relationship between two reference bases.
+         * This indicates if the breakpoint, for an inversion suspect specifically, represents (one of the two) bases
+         * that is the left of the pair.
+         */
+        private static final Map<String, String> INV55_FLAG = Collections.singletonMap(INV55, "");
+        private static final Map<String, String> INV33_FLAG = Collections.singletonMap(INV33, "");
 
-        StrandSwitchBND(final String id, final Map<String, String> typeSpecificExtraAttributes,
-                        final byte[] bases, final boolean bracketPointsLeft, final SimpleInterval novelAdjRefLoc,
-                        final boolean basesFirst, boolean isTheUpstreamMate){
-            super(id, typeSpecificExtraAttributes, bases, bracketPointsLeft, novelAdjRefLoc, basesFirst, isTheUpstreamMate);
+        private InvSuspectBND(final String id, final byte[] bases, final SimpleInterval novelAdjRefLoc,
+                              final boolean bracketPointsLeft, final boolean basesFirst, boolean isTheUpstreamMate) {
+            super(id, (bracketPointsLeft && basesFirst) ? INV55_FLAG: INV33_FLAG,
+                    bases, bracketPointsLeft, novelAdjRefLoc, basesFirst, isTheUpstreamMate);
         }
 
+        public static Tuple2<BreakEndVariantType, BreakEndVariantType> getOrderedMates(final NovelAdjacencyReferenceLocations narl,
+                                                                                       final ReferenceMultiSource reference) {
+
+            // inversion breakend formatted records have "bracketPointsLeft" "basesFirst" taking the same value (see spec)
+            final boolean isInv55Suspect;
+            if (narl.strandSwitch == StrandSwitch.FORWARD_TO_REVERSE) { // INV55, leftHalfInPartnerPair
+                isInv55Suspect = true;
+            } else if (narl.strandSwitch == StrandSwitch.REVERSE_TO_FORWARD){
+                isInv55Suspect = false;
+            } else {
+                throw new GATKException("Wrong type of novel adjacency sent to wrong analysis pathway: " +
+                        "no strand-switch being sent to strand-switch path. \n" + narl.toString());
+            }
+            final BreakEndVariantType upstreamBreakpoint = new BreakEndVariantType.InvSuspectBND(getIDString(narl, true),
+                    extractBasesForAltAllele(narl, true, reference),
+                    getMateRefLoc(narl, true),
+                    isInv55Suspect, isInv55Suspect, true);
+            final BreakEndVariantType downstreamBreakpoint = new BreakEndVariantType.InvSuspectBND(getIDString(narl, false),
+                    extractBasesForAltAllele(narl, false, reference),
+                    getMateRefLoc(narl, false),
+                    isInv55Suspect, isInv55Suspect, false);
+            return new Tuple2<>(upstreamBreakpoint, downstreamBreakpoint);
+        }
 
         static byte[] extractBasesForAltAllele(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc,
                                                final ReferenceMultiSource reference) {
             try {
                 final byte[] ref = reference
                         .getReferenceBases(null, forUpstreamLoc ? narl.leftJustifiedLeftRefLoc :
-                                narl.leftJustifiedRightRefLoc)
+                                                                                narl.leftJustifiedRightRefLoc)
                         .getBases();
                 final String ins = narl.complication.getInsertedSequenceForwardStrandRep();
                 if (ins.isEmpty()) {
@@ -67,57 +126,70 @@ public abstract class BreakEndVariantType extends SvType {
                 throw new GATKException("Could not read reference for extracting reference bases.", ioex);
             }
         }
-
-        static SimpleInterval getTheOtherRefLoc(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc) {
-            return forUpstreamLoc ? narl.leftJustifiedRightRefLoc : narl.leftJustifiedLeftRefLoc;
-        }
     }
 
-    public static final class INV55BND extends StrandSwitchBND {
+    //==================================================================================================================
+    /**
+     * Generic variant type for suspected "translocations", including what could be a breakpoint for
+     * a dispersed duplication or an inter-chromosome novel adjacency.
+     */
+    public static final class TransLocBND extends BreakEndVariantType {
+        private static Map<String, String> emptyMap = Collections.emptyMap();
 
-        /**
-         * Technically, a strand switch breakpoint should have two VCF records, hence we also have {@code forUpstreamLoc}.
-         */
-        public INV55BND(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc,
-                        final ReferenceMultiSource reference) {
-            super(getIDString(narl, forUpstreamLoc),
-                    Collections.singletonMap(GATKSVVCFConstants.INV55, ""),
-                    extractBasesForAltAllele(narl, forUpstreamLoc, reference), true,
-                    getTheOtherRefLoc(narl, forUpstreamLoc), true, forUpstreamLoc);
+        private TransLocBND(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc,
+                            final ReferenceMultiSource reference, final SAMSequenceDictionary referenceDictionary,
+                            final boolean basesFirst, final boolean bracketPointsLeft) {
+            super(getIDString(narl, forUpstreamLoc), emptyMap,
+                    extractBasesForAltAllele(narl, forUpstreamLoc, reference, referenceDictionary), bracketPointsLeft,
+                    getMateRefLoc(narl, forUpstreamLoc), basesFirst, forUpstreamLoc);
         }
 
-        private static String getIDString(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc) {
-
-            return GATKSVVCFConstants.BREAKEND_STR + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
-                    GATKSVVCFConstants.INV55 + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
-                    narl.leftJustifiedLeftRefLoc.getContig() + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
-                    narl.leftJustifiedLeftRefLoc.getEnd() + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
-                    narl.leftJustifiedRightRefLoc.getStart() + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
-                    (forUpstreamLoc ? "1" : "2") ;
+        public static Tuple2<BreakEndVariantType, BreakEndVariantType> getOrderedMates(final NovelAdjacencyReferenceLocations narl,
+                                                                                       final ReferenceMultiSource reference,
+                                                                                       final SAMSequenceDictionary referenceDictionary) {
+            final boolean isSameChr = narl.leftJustifiedLeftRefLoc.getContig().equals(narl.leftJustifiedRightRefLoc.getContig());
+            final BreakEndVariantType bkpt_1, bkpt_2;
+            if (isSameChr) {
+                bkpt_1 = new BreakEndVariantType.TransLocBND(narl, true, reference, referenceDictionary, false, true);
+                bkpt_2 = new BreakEndVariantType.TransLocBND(narl, false, reference, referenceDictionary, true, false);
+            } else {
+                if (narl.strandSwitch == StrandSwitch.NO_SWITCH) {
+                    final boolean isFirstOfPartner = IntervalUtils.compareContigs(narl.leftJustifiedLeftRefLoc,
+                                                                                  narl.leftJustifiedRightRefLoc, referenceDictionary) < 0;
+                    bkpt_1 = new BreakEndVariantType.TransLocBND(narl, true, reference, referenceDictionary,
+                            isFirstOfPartner, !isFirstOfPartner);
+                    bkpt_2 = new BreakEndVariantType.TransLocBND(narl, false, reference, referenceDictionary,
+                            !isFirstOfPartner, isFirstOfPartner);
+                } else {
+                    bkpt_1 = new BreakEndVariantType.TransLocBND(narl, true, reference, referenceDictionary,
+                            narl.strandSwitch == StrandSwitch.FORWARD_TO_REVERSE, narl.strandSwitch == StrandSwitch.FORWARD_TO_REVERSE);
+                    bkpt_2 = new BreakEndVariantType.TransLocBND(narl, false, reference, referenceDictionary,
+                            narl.strandSwitch != StrandSwitch.FORWARD_TO_REVERSE, narl.strandSwitch != StrandSwitch.FORWARD_TO_REVERSE);
+                }
+            }
+            return new Tuple2<>(bkpt_1, bkpt_2);
         }
-    }
 
-    public static final class INV33BND extends StrandSwitchBND {
-
-        /**
-         * Technically, a strand switch breakpoint should have two VCF records, hence we also have {@code forUpstreamLoc}.
-         */
-        public INV33BND(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc,
-                        final ReferenceMultiSource reference) {
-            super(getIDString(narl, forUpstreamLoc),
-                    Collections.singletonMap(GATKSVVCFConstants.INV33, ""),
-                    extractBasesForAltAllele(narl, forUpstreamLoc, reference), false,
-                    getTheOtherRefLoc(narl, forUpstreamLoc), false, forUpstreamLoc);
-        }
-
-        private static String getIDString(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc) {
-
-            return GATKSVVCFConstants.BREAKEND_STR + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
-                    GATKSVVCFConstants.INV33 + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
-                    narl.leftJustifiedLeftRefLoc.getContig() + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
-                    narl.leftJustifiedLeftRefLoc.getEnd() + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
-                    narl.leftJustifiedRightRefLoc.getStart() + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
-                    (forUpstreamLoc ? "1" : "2") ;
+        static byte[] extractBasesForAltAllele(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc,
+                                               final ReferenceMultiSource reference, final SAMSequenceDictionary referenceDictionary) {
+            try {
+                final SimpleInterval refLoc = forUpstreamLoc ? narl.leftJustifiedLeftRefLoc : narl.leftJustifiedRightRefLoc;
+                final byte[] ref = reference.getReferenceBases(null, refLoc).getBases();
+                final String ins = narl.complication.getInsertedSequenceForwardStrandRep();
+                if (ins.isEmpty()) {
+                    return ref;
+                } else {
+                    if (narl.leftJustifiedLeftRefLoc.getContig().equals(narl.leftJustifiedRightRefLoc.getContig())
+                            || (narl.strandSwitch == StrandSwitch.NO_SWITCH && IntervalUtils.compareContigs(narl.leftJustifiedLeftRefLoc, narl.leftJustifiedRightRefLoc, referenceDictionary) > 0)
+                            || narl.strandSwitch == StrandSwitch.REVERSE_TO_FORWARD) {
+                        return forUpstreamLoc ? ArrayUtils.addAll(ins.getBytes(), ref) : ArrayUtils.addAll(ref, ins.getBytes());
+                    } else {
+                        return forUpstreamLoc ? ArrayUtils.addAll(ref, ins.getBytes()) : ArrayUtils.addAll(ins.getBytes(), ref);
+                    }
+                }
+            } catch (final IOException ioex) {
+                throw new GATKException("Could not read reference for extracting reference bases.", ioex);
+            }
         }
     }
 }
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
@@ -5,6 +5,8 @@ import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
 import com.google.common.annotations.VisibleForTesting;
+import htsjdk.samtools.SAMSequenceDictionary;
+import org.broadinstitute.hellbender.utils.IntervalUtils;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import scala.Tuple2;
 
@@ -71,6 +73,13 @@ public class ChimericAlignment {
         return Arrays.asList(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig);
     }
 
+    Tuple2<SimpleInterval, SimpleInterval> getCoordSortedReferenceSpans() {
+        if (involvesRefPositionSwitch(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig))
+            return new Tuple2<>(regionWithHigherCoordOnContig.referenceSpan, regionWithLowerCoordOnContig.referenceSpan);
+        else
+            return new Tuple2<>(regionWithLowerCoordOnContig.referenceSpan, regionWithHigherCoordOnContig.referenceSpan);
+    }
+
     protected ChimericAlignment(final Kryo kryo, final Input input) {
 
         this.sourceContigName = input.readString();
@@ -95,7 +104,8 @@ public class ChimericAlignment {
      */
     @VisibleForTesting
     public ChimericAlignment(final AlignmentInterval intervalWithLowerCoordOnContig, final AlignmentInterval intervalWithHigherCoordOnContig,
-                             final List<String> insertionMappings, final String sourceContigName) {
+                             final List<String> insertionMappings, final String sourceContigName,
+                             final SAMSequenceDictionary referenceDictionary) {
 
         this.sourceContigName = sourceContigName;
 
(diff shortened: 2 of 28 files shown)


SHuang-Broad:
detail explained [here](https://github.com/broadinstitute/dsde-methods-sv/pull/9)

SHuang-Broad:
@cwhelan I've finished the rebase now. Please take a look. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
```diff
@@ -77,50 +123,99 @@ protected NovelAdjacencyReferenceLocations(final Kryo kryo, final Input input) {
...
             if (ca.strandSwitch == StrandSwitch.NO_SWITCH) {
-                leftBreakpointCoord  = leftReferenceInterval.getEnd() - homologyLen;
-                rightBreakpointCoord = rightReferenceInterval.getStart() - 1;
-            } else if (ca.strandSwitch == StrandSwitch.FORWARD_TO_REVERSE){
-                leftBreakpointCoord  = leftReferenceInterval.getEnd() - homologyLen;
-                rightBreakpointCoord = rightReferenceInterval.getEnd();
+                if (ca.isForwardStrandRepresentation) { // there's no absolute difference between left and right here because they map to different chromosome, left/right could be later defined by dictionary when outputting VCF
+                    if (0 > IntervalUtils.compareContigs(ca.regionWithLowerCoordOnContig.referenceSpan,
+                            ca.regionWithHigherCoordOnContig.referenceSpan, referenceDictionary)) {
+                        leftBreakpointRefContig  = ca.regionWithLowerCoordOnContig.referenceSpan.getContig();
```
cwhelan: Again, having so many similar but subtly different sets of lines that set `leftBreakpointCoord`, etc. is quite difficult to read. Here's another suggestion for refactoring this code -- could you decide on a "default" ordering of the two CA regions, and then apply reversals or homology adjustments as operations on your coordinate set? Then it would look like:

coords = default coordinates
if (strand switch FR) reverse coords
etc..

That may or may not work in conjunction with my suggestion of refactoring to subclasses that I talked about above, so feel free to take one suggestion or the other.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ContigAlignmentsModifier.java
```diff
@@ -0,0 +1,208 @@
...
+                reconstructedTwo = clipAlignmentInterval(two, overlapOnRead, false);
+            }
+
+        }
+        return Arrays.asList(reconstructedOne, reconstructedTwo);
+    }
+
+    /**
+     * Given {@code clipLengthOnRead} to be clipped from an aligned contig's {@link AlignmentInterval} {@code input},
+     * return a modified {@link AlignmentInterval} with the requested number of bases clipped away and ref span recomputed.
```
cwhelan: I'd document here that the NM field will be wiped out.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ContigAlignmentsModifier.java
```diff
@@ -0,0 +1,208 @@
...
+        final int newTigStart, newTigEnd;
+        if (clipFrom3PrimeEnd) {
+            newTigStart = input.startInAssembledContig;
+            newTigEnd   = input.endInAssembledContig - clipLengthOnRead;
+        } else {
+            newTigStart = input.startInAssembledContig + clipLengthOnRead;
+            newTigEnd   = input.endInAssembledContig;
+        }
+        return new AlignmentInterval(result._1, newTigStart, newTigEnd, result._2, input.forwardStrand, input.mapQual,
+                StructuralVariationDiscoveryArgumentCollection.DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection.MISSING_NM,
```
cwhelan: What's the difference between this constant and `AlignmentInterval.NO_NM`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/FilterLongReadAlignmentsSAMSpark.java
```diff
@@ -248,7 +247,10 @@ private static boolean contigFilter(final AlignedContig contig) {
                 .orElseThrow(() -> new GATKException("Cannot find best-scoring configuration on alignments of contig: " + alignedContig.contigName));
 
         return IntStream.range(0, allConfigurations.size())
-                .filter(i -> scores.get(i) >= maxScore)
+                .filter(i -> {
+                    final Double s = scores.get(i);
+                    return s >= maxScore || maxScore - s <= Math.ulp(s); // sometimes, not sure what happened, two configurations with would-be-same-scores have differ by a Math.ulp(appear-to-be-lower)
```
cwhelan: I think there's a typo in your comment, maybe you mean "two configurations which should have the same scores differ by a Math.ulp".

I think it might be better to explicitly set a delta for tolerance of equal alignment scores rather than using `Math.ulp`, primarily because I don't quite understand where the difference is coming from and why Math.ulp is guaranteed to fix it. What do you think?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SuspectedTransLocDetector.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @SuppressWarnings("unchecked")
+    private static final List<String> EMPTY_INSERTION_MAPPINGS = Collections.EMPTY_LIST;
+
+    @Override
+    public void inferSvAndWriteVCF(final JavaRDD<AlignedContig> localAssemblyContigs, final String vcfOutputFileName,
+                                   final Broadcast<ReferenceMultiSource> broadcastReference,
+                                   final Broadcast<SAMSequenceDictionary> referenceDictionaryBroadcast,
+                                   final Logger toolLogger) {
+        localAssemblyContigs.cache();
```
cwhelan: Do you need this `cache` here? Where is it getting unpersisted? I think it would be better to have the calling method handle caching and unpersisting RDDs so that we don't lose track of all the things we've cached.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SuspectedTransLocDetector.java
```diff
@@ -0,0 +1,71 @@
...
+                                        .produceAnnotatedBNDmatesVcFromNovelAdjacency(noveltyTypeAndEvidence._1,
+                                                noveltyTypeAndEvidence._2._1, noveltyTypeAndEvidence._2._2, broadcastReference).iterator());
+
+        SVVCFWriter.writeVCF(annotatedBNDs.collect(), vcfOutputFileName.replace(".vcf", "_transBND.vcf"),
+                            referenceDictionaryBroadcast.getValue(), toolLogger);
+    }
+
+    private static Tuple2<ChimericAlignment, byte[]> convertAlignmentIntervalsToChimericAlignment(final AlignedContig contig,
+                                                                                                  final SAMSequenceDictionary referenceDictionary) {
+        // TODO: 9/9/17 this default empty insertion mapping treatment is temporary and should be fixed later
```
cwhelan: What needs to happen to fix this?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
```diff
@@ -109,14 +111,29 @@ protected void runTool(final JavaSparkContext ctx) {
...
-                .inferSvAndWriteVCF(contigsByPossibleRawTypes.get(RawTypes.Inv), outputDir+"/"+RawTypes.Inv.name()+".vcf",
-                        referenceMultiSourceBroadcast, discoverStageArgs.fastaReference, localLogger);
+                .inferSvAndWriteVCF(contigsByPossibleRawTypes.get(RawTypes.Inv), outputDir+"/"+ RawTypes.Inv.name()+".vcf",
+                        referenceMultiSourceBroadcast, sequenceDictionaryBroadcast, localLogger);
+
+        // here contigs supporting cpx having only 2 alignments could be handled easily together with ref block order switch ones
+        final JavaRDD<AlignedContig> diffChrTrans =
+                contigsByPossibleRawTypes.get(RawTypes.Cpx).filter(tig -> tig.alignmentIntervals.size() == 2);
+        new SuspectedTransLocDetector()
+                .inferSvAndWriteVCF(contigsByPossibleRawTypes.get(RawTypes.DispersedDupOrMEI).union(diffChrTrans),
```
cwhelan: Why not just put the two classes of alignment in the same bucket to begin with (in the "divertReads" method)? Might be more efficient, and they are likely all the same type of event.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/BreakEndVariantType.java
```diff
@@ -120,4 +137,77 @@ private static String getIDString(final NovelAdjacencyReferenceLocations narl, f
...
+
+        private TransLocBND(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc,
+                            final ReferenceMultiSource reference, final SAMSequenceDictionary referenceDictionary,
+                            final boolean basesFirst, final boolean bracketPointsLeft) {
+            super(getIDString(narl, forUpstreamLoc), emptyMap,
+                    extractBasesForAltAllele(narl, forUpstreamLoc, reference, referenceDictionary), bracketPointsLeft,
+                    getMateRefLoc(narl, forUpstreamLoc), basesFirst, forUpstreamLoc);
+        }
+
+        public static Tuple2<BreakEndVariantType, BreakEndVariantType> getOrderedMates(final NovelAdjacencyReferenceLocations narl,
```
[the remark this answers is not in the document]
SHuang-Broad: There is a documentation for that [here](https://github.com/broadinstitute/dsde-methods-sv/blob/1d8789605a542be0db65cb2666d0d2b1b4223fc4/docs/ideas/transLoc/transLocationVcfProposal.md) in our private repo. 
Admittedly it is accessible only for us. But I think that is OK for now. 

I've been refraining from adding tests for these new features through this whole complex SV PR series (see #3431) because of the nightmarish memory back in the days of adding tests for ins/del cases&mdash;getting the locations exactly right takes as much as, if not more, energy to get these logic right.
I've been keeping track of the test coverage in this prototyping package for a while, and yes, it is [VERY VERY BAD](https://codecov.io/gh/broadinstitute/gatk/tree/a23536ab54b723e35afb1f8012b5548d74664872/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype) right now. What I have in mind, is to have a pipeline in place, generate call set for evaluation (correctness, format, etc), and add tests based on feedback from at least one round of call set review. Rightnow, I'm the only one that is running this parallel program (not even a pipeline yet like `StructuralVariationDiscoveryPipelineSpark`). And before test coverage get above 75%, I have no intention of hooking up this path and replace the master pipeline.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/BreakEndVariantType.java
```diff
@@ -67,23 +74,26 @@ private static Allele constructAltAllele(final byte[] bases, final boolean brack
                 throw new GATKException("Could not read reference for extracting reference bases.", ioex);
             }
         }
-
-        static SimpleInterval getTheOtherRefLoc(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc) {
-            return forUpstreamLoc ? narl.leftJustifiedRightRefLoc : narl.leftJustifiedLeftRefLoc;
-        }
     }
 
     public static final class INV55BND extends StrandSwitchBND {
```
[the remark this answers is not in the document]
SHuang-Broad: Thanks! I've been wanting to do that but was too lazy. Made the change. Please see if it is better now.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/BreakEndVariantType.java
```diff
@@ -120,4 +137,77 @@ private static String getIDString(final NovelAdjacencyReferenceLocations narl, f
...
+        private static String getIDString(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc) {
+            return GATKSVVCFConstants.BREAKEND_STR + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
+                    narl.leftJustifiedLeftRefLoc.getContig() + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
+                    narl.leftJustifiedLeftRefLoc.getEnd() + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
+                    narl.leftJustifiedRightRefLoc.getContig() + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
+                    narl.leftJustifiedRightRefLoc.getStart() + GATKSVVCFConstants.INTERVAL_VARIANT_ID_FIELD_SEPARATOR +
+                    (forUpstreamLoc ? "1" : "2");
+        }
+
+        static byte[] extractBasesForAltAllele(final NovelAdjacencyReferenceLocations narl, final boolean forUpstreamLoc,
```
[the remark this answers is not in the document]
SHuang-Broad: Please see above.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/BreakEndVariantType.java
```diff
@@ -40,21 +61,51 @@ public final String toString() {
...
-    private static Allele constructAltAllele(final byte[] bases, final boolean bracketPointsLeft, final SimpleInterval novelAdjRefLoc,
-                                             final boolean basesFirst) {
-        final String s = bracketPointsLeft ? "]" + novelAdjRefLoc.getContig() + ":" + novelAdjRefLoc.getStart() + "]"
-                                           : "[" + novelAdjRefLoc.getContig() + ":" + novelAdjRefLoc.getStart() + "[";
-        return Allele.create( basesFirst ? new String(bases) + s  : s + new String(bases) );
-    }
-
-    static abstract class StrandSwitchBND extends BreakEndVariantType {
+    //==================================================================================================================
+    public static final class InvSuspectBND extends BreakEndVariantType {
```
cwhelan: Why go from `StrandSwitchBND` to `InvSuspectBND`? The old name seemed pretty clear to me.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/BreakEndVariantType.java
```diff
@@ -40,21 +61,51 @@ public final String toString() {
...
+            super(id, (bracketPointsLeft && basesFirst) ? INV55_FLAG: INV33_FLAG,
+                    bases, bracketPointsLeft, novelAdjRefLoc, basesFirst, isTheUpstreamMate);
         }
 
+        public static Tuple2<BreakEndVariantType, BreakEndVariantType> getOrderedMates(final NovelAdjacencyReferenceLocations narl,
+                                                                                       final ReferenceMultiSource reference) {
+
+            final BreakEndVariantType bkpt_1, bkpt_2;
+            if (narl.strandSwitch == StrandSwitch.FORWARD_TO_REVERSE) { // INV55, leftHalfInPartnerPair
+                bkpt_1 = new BreakEndVariantType.InvSuspectBND(getIDString(narl, true),
```
cwhelan: These two clauses are identical except for the three boolean parameters -- it seems like you could just have one set of lines with a boolean to control those.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/BreakEndVariantType.java
SHuang-Broad: This was to distinguish it from the "translocation BND" type below, because here I'm making the distinction upstream of this step:
* if a contig has two same-chr mappings with strand switch, I'm going to flag them later as inversion suspects (INV33 or INV55), because inversion is kind of special and to be backward compatible (when this experimental tool is hooked up into our main pipeline, inversion calls are not made directly by alignments, but needs other evidence for final call); 
* but for contigs with mappings to different chromosomes, I don't make a distinction between with or without strand switch&mdash;they are going to be output as simple BND formatted records.

So the old name `StrandSwitchBND` was not distinguishing between the inversion suspects (same chr event), and the different-chr events (called translocation here).

SHuang-Broad: In retrospect, I think there's a valid concern here: that the inversion breakpoint suspects may turn out to be insertions, and these should be merged with the translocation records, so the "inversion suspect" label should be revoked from such records.
What about dealing with this merge later, as this an issue that is IMO better treated in a different PR?

SHuang-Broad: But comments are added to this class for clarifying the distinction.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/BreakEndVariantType.java
SHuang-Broad: done. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ContigAlignmentsModifier.java
SHuang-Broad: done.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/ContigAlignmentsModifier.java
SHuang-Broad: removed and replaced with `NO_NM` everywhere

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/FilterLongReadAlignmentsSAMSpark.java
SHuang-Broad: i went back and digged a bit deeper.
I think the problem is an inherent sin of comparing floating point numbers that could potentially be very close to each other.
I initially discovered the problem by debugging for an unexpected explosion: two scores computed to be the same by the score computing unit, but when it is sent back to the calling code here, one of them is changed, slightly by the Math.ulp() amount. 

I've added an advanced parameter that defaults to zero (but floating point numbers are never exactly zero I guess), and it is argued [here](http://www.cygnus-software.com/papers/comparingfloats/Comparing%20floating%20point%20numbers.htm) that ULP is probably the least evil way to go.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SuspectedTransLocDetector.java
SHuang-Broad: agree and removed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SuspectedTransLocDetector.java
SHuang-Broad: removed the comment.
What I really was thinking about is that the alignment scoring and filtering step might decide an alignment brings in noise and filter it away,  think about the following scenario: 
a configuration with 3 alignments, with the middle one overlaps with other two, and has a mediocre MQ, this currently is treated as noise and filtered away in the alternative code path, but the master code path would note down it as inserted sequence mapping and skip it. 
The above scenario is definitely one place where the filtering step should be improved upon (#3473 and #3647 )

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
SHuang-Broad: There's a branch for this purpose, just not ready yet for a PR...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
```diff
@@ -204,6 +229,23 @@ public static StrandSwitch determineStrandSwitch(final AlignmentInterval first,
     }
 
     /**
+     * An SV event could be detected from a contig that seem originate from the forward or reverse strand of the reference,
+     * besides the annotation that the alignment flanking regions might flank either side of the two breakpoints.
+     */
+    @VisibleForTesting
+    static boolean isForwardStrandRepresentation(final AlignmentInterval regionWithLowerCoordOnContig,
```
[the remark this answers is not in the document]
SHuang-Broad: not quite understanding the commit. Could you explain a bit more?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
```diff
@@ -156,10 +179,12 @@ public ChimericAlignment(final AlignmentInterval intervalWithLowerCoordOnContig,
                 }
             }
 
-            final boolean isNotSimpleTranslocation = isNotSimpleTranslocation(current, next,
-                    determineStrandSwitch(current, next), involvesRefPositionSwitch(current, next));
-            if (isNotSimpleTranslocation)
-                results.add(new ChimericAlignment(current, next, insertionMappings, alignedContig.contigName));
+            final SAMSequenceDictionary referenceDictionary = null; // don't need this for its intended cases, which are all same CHR mapping
```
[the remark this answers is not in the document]
SHuang-Broad: refactored interface of this function by taking in a dictionary.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
```diff
@@ -238,35 +274,22 @@ public static boolean isNotSimpleTranslocation(final AlignmentInterval regionWit
     }
 
     /**
-     * An SV event could be detected from a contig that seem originate from the forward or reverse strand of the reference,
-     * besides the annotation that the alignment flanking regions might flank either side of the two breakpoints.
+     * Determine if the chimeric alignment indicates a simple translocation.
+     * Simple translocations are defined here and at this time as:
+     *  <li>inter-chromosomal translocations, i.e. novel adjacency between different reference chromosomes, or</li>
+     *  <li>intra-chromosomal translocation that DOES NOT involve a strand switch, i.e.
+     *      novel adjacency between reference locations on the same chromosome involving NO strand switch but reference interval order switch</li>
```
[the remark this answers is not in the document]
SHuang-Broad: Yes and Yes:
1. For intra-chromosome "translocations" with strand switch, I am not sure how to handle/confidently interpret that yet, but in this prototyping tool we are emitting them as BND format records with `INV33` or `INV55` flags, unlike what we used to emit as inversion calls. 
2. We definitely reinforce/discredit all our deletion calls with coverage information around the identified breakpoints, but that is a task of later stage (interpreting events based solely on alignment signatures seems to be less and less reliable as we gain more and more experience, and similarly for using coverage only)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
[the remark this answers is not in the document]
cwhelan: I was referring to what I said above:

>Also, the fact that there's a method called isForwardStrandRepresentation that's used in the other half of this if/else but doesn't handle the logic for this half bugs me a bit. Wouldn't it be possible to extend that method to handle this logic? Alternatively I'd rename that method to something that describes the scenario in which it's applicable, eg `isIntraContigMappingForwardStrandRepresentation`.

To clarify what I meant, the entire chunk of code in ChimericAlignment from lines 119 to 130 seems to be determining whether the alignment is in "forward strand representation." Part of that chunk of code delegates to this method called `isForwardStrandRepresentation`. I feel like the naming is confusing because this method is only handling the logic for a specific subset of inputs, but it would be a natural place to go for a code reader who wanted to see where "forward strand representation" is being set.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
```diff
@@ -104,14 +115,26 @@ public ChimericAlignment(final AlignmentInterval intervalWithLowerCoordOnContig,
...
+        final boolean mappedToSameChr = intervalWithLowerCoordOnContig.referenceSpan.getContig().equals(intervalWithHigherCoordOnContig.referenceSpan.getContig());
+        if (mappedToSameChr) {
+            final boolean involvesRefIntervalSwitch = involvesRefPositionSwitch(intervalWithLowerCoordOnContig, intervalWithHigherCoordOnContig);
+            this.isForwardStrandRepresentation = isForwardStrandRepresentation(intervalWithLowerCoordOnContig, intervalWithHigherCoordOnContig,
+                                                                                this.strandSwitch, involvesRefIntervalSwitch);
+        } else {
+            if (strandSwitch == StrandSwitch.NO_SWITCH) {
+                this.isForwardStrandRepresentation = regionWithLowerCoordOnContig.forwardStrand;
+            } else {
+                this.isForwardStrandRepresentation = IntervalUtils.compareContigs(intervalWithLowerCoordOnContig.referenceSpan,
```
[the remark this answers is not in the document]
SHuang-Broad: docs added. And the 2nd issue is actually handled in another branch that is lined up after this PR. I feel I am generating too many PRs...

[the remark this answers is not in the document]
SHuang-Broad: And the method `involvesRefPositionSwitch()` is going to be completely removed.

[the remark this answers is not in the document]
SHuang-Broad: diffchrwithss_neg2pos
diffchrwithss_pos2neg

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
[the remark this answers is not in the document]
SHuang-Broad: Ah, I see. Thanks! I just responded to the comment above with some docs in the method, and some figures.
The incoherence has been handled in a separate branch that also tries to deal with several other issues with this class. It will generate another small PR with tests.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
```diff
@@ -52,23 +57,64 @@ protected NovelAdjacencyReferenceLocations(final Kryo kryo, final Input input) {
...
 
         final int homologyLen = complication.getHomologyForwardStrandRep().length();
 
         final String leftBreakpointRefContig, rightBreakpointRefContig;
         final int leftBreakpointCoord, rightBreakpointCoord;
-        if (complication.hasDuplicationAnnotation()) {
-            if (complication.hasDupSeqButNoStrandSwitch()) { // simple duplication not involving inverted repeats
-                leftBreakpointRefContig = rightBreakpointRefContig = ca.regionWithLowerCoordOnContig.referenceSpan.getContig();
+        final boolean sameChromosome = ca.regionWithLowerCoordOnContig.referenceSpan.getContig().equals(ca.regionWithHigherCoordOnContig.referenceSpan.getContig());
+        if (sameChromosome) {
```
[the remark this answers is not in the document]
SHuang-Broad: Totally agree. I was too deep in the weeds that I forgot about this. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
```diff
@@ -52,23 +57,64 @@ protected NovelAdjacencyReferenceLocations(final Kryo kryo, final Input input) {
...
+                    }
+                    if (complication.getDupSeqRepeatNumOnCtg() > complication.getDupSeqRepeatNumOnRef()) {
+                        leftBreakpointCoord = leftReferenceInterval.getEnd() - homologyLen
+                                - (complication.getDupSeqRepeatNumOnCtg() - complication.getDupSeqRepeatNumOnRef()) * complication.getDupSeqRepeatUnitRefSpan().size();
+                    } else {
+                        leftBreakpointCoord = leftReferenceInterval.getEnd() - homologyLen;
+                    }
+                    rightBreakpointCoord = rightReferenceInterval.getStart() - 1;
+                } else {
+                    leftBreakpointCoord     = complication.getDupSeqRepeatUnitRefSpan().getStart() - 1;
```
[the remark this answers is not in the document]
SHuang-Broad: Because for inverted duplications, the ones we can currently catch is not as complicated as the tandem duplication (remember those heuristic optimization we need to do?). And for inverted duplications, there's no contraction here.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
SHuang-Broad: I refactored this 100+ line of branching into a class hierarchy in this new commit. Some refactoring still needs to be done for `ChimericAlignment`, but it is more suitable for the next planned refactoring PR.

SHuang-Broad:
@cwhelan I've made the suggested refactoring for breakpoint adjustment code.
Do you want to take a look again?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/BreakEndVariantType.java
cwhelan: What if you just called them `IntraChromStrandSwitchBND`? That way the logic isn't tied into the name -- IMO the name should just describe what the things _are_, not what you are going to do with them.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/FilterLongReadAlignmentsSAMSpark.java
```diff
@@ -61,6 +62,12 @@
     private static final long serialVersionUID = 1L;
     private final Logger localLogger = LogManager.getLogger(FilterLongReadAlignmentsSAMSpark.class);
 
+    @Advanced
+    @Argument(doc = "parameter (float) for specifying of two configurations differ by no more than this amount, the two are considered equally good",
```
cwhelan: How about, "Maximum difference between alignment configuration scores for which two configurations will be considered equally likely"?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/FilterLongReadAlignmentsSAMSpark.java
```diff
@@ -61,6 +62,12 @@
...
+            shortName = "cst",
+            fullName = "configScoreDiffTol", optional = true)
+    public final Double configScoreDiffTolerance = 0.0;
```
cwhelan: Do you want the default to be 0.0? I thought you'd want some small delta.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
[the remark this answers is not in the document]
cwhelan: I see.. For now maybe just rename that method to something like `isIntraContigMappingForwardStrandRepresentation` so the code isn't in a confusing state between your different changes?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
```diff
@@ -179,12 +182,14 @@ public ChimericAlignment(final AlignmentInterval intervalWithLowerCoordOnContig,
...
             }
 
-            final SAMSequenceDictionary referenceDictionary = null; // don't need this for its intended cases, which are all same CHR mapping
-            final ChimericAlignment chimericAlignment = new ChimericAlignment(current, next, insertionMappings, alignedContig.contigName, referenceDictionary);
-            if (!chimericAlignment.isNotSimpleTranslocation())
-                throw new GATKException.ShouldNeverReachHereException("Mapped assembled contigs are sent down the wrong path: " +
-                        "contig suggesting \"translocation\" is sent down the insert/deletion path.\n" + alignedContig.toString());
-            results.add(chimericAlignment);
+            // TODO: 10/18/17 this way of filtering CA based on not quality but alignment characteristics is
+            // an artifact of development and relies on the dark secret that this was initially developped for calling ins/del,
```
cwhelan: 'developed'.

Also I'm not sure it needs to be portrayed as a an 'artifact' and a 'dark secret', there's nothing to be ashamed about.. just say that they are sent down different code paths right now.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
```diff
@@ -57,64 +58,78 @@ protected NovelAdjacencyReferenceLocations(final Kryo kryo, final Input input) {
...
-    /**
-     * Returns the reference coordinates of the left and right breakpoints implied by this chimeric alignment.
-     * If there is homologous sequence represented in the alignments, it will be assigned to the side of the breakpoint
-     * with higher reference coordinates.
-     */
-    @VisibleForTesting
-    static Tuple2<SimpleInterval, SimpleInterval> leftJustifyBreakpoints(final ChimericAlignment ca,
-                                                                         final BreakpointComplications complication,
-                                                                         final SAMSequenceDictionary referenceDictionary) {
+    private abstract static class BreakpointLeftAligner {
```
cwhelan: Don't call these `Aligners` since that's a word with a specific meaning in our field (ie. actual string alignment). How about `ContigToReferenceBreakpointMapping` or something like that?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
```diff
@@ -57,64 +58,78 @@ protected NovelAdjacencyReferenceLocations(final Kryo kryo, final Input input) {
...
 
-        final int homologyLen = complication.getHomologyForwardStrandRep().length();
+        protected String upstreamBreakpointRefContig;
+        protected String dnstreamBreakpointRefContig;
```
cwhelan: i'd just write `downstreamBreakpointRefContig`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
```diff
@@ -123,79 +138,150 @@ protected NovelAdjacencyReferenceLocations(final Kryo kryo, final Input input) {
...
+     */
+    @VisibleForTesting
+    static Tuple2<SimpleInterval, SimpleInterval> leftJustifyBreakpoints(final ChimericAlignment ca,
+                                                                         final BreakpointComplications complication,
+                                                                         final SAMSequenceDictionary referenceDictionary) {
+
+        final boolean sameChromosome = ca.regionWithLowerCoordOnContig.referenceSpan.getContig()
+                .equals(ca.regionWithHigherCoordOnContig.referenceSpan.getContig());
+        final Tuple2<SimpleInterval, SimpleInterval> leftAdjustedBreakpoints;
+        if (sameChromosome) {
```
cwhelan: This refactoring is almost there, but I think you could be helped out a lot by a factory method pattern:

make a static method `getContigToReferenceBreakpointMapping(ChimericAlignment, BreakpointComplications, SAMSequenceDictionary)` on `BreakpointLeftAligner` (or whatever you end up renaming that to, like `ContigToReferenceBreakpointMapping`) that returns a `BreakpointLeftAligner` (or `ContigToReferenceBreakpointMapping`). Then this method only calls `getLeftJustifiedBreakpoints` on the result. That way there's only one invocation of `getLeftJustifiedBreakpoints`, one method handles choosing the logic path to go down, and you have separate, self-contained places where the different chunks of logic are implemented.

cwhelan:
@SHuang-Broad Added some comments on your changes; mostly I think they made the code a lot cleaner.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
SHuang-Broad: done. Learned factory pattern. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
SHuang-Broad: Roger. Renamed from `LeftAligner` to `Inference`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/FilterLongReadAlignmentsSAMSpark.java
SHuang-Broad: After reading the link I gave there, it seems there's no universally good answer to comparing floating point numbers.
What I am trying to fix here is that sometimes (we have no way of knowing when this happens as exact machine representation of floating point numbers depends on a lot of factors) two configurations that should be equally good while calculated by human with paper/pencil could have perfectly different machine representations. 
While a user can provide an epsilon, the code actually check and use which one is bigger&mdash;the user provided epsilon, and one ULP, in case a user provided literal 0 (that is really not desirable). So even though 0 is the default value here, it is actually not used. 
And this feature is marked as `@Advanced` so shouldn't be messed with unless the caller is experimenting with the filtering module itself.




To quote (I think we are facing the 2nd scenario)


>There is no silver bullet. You have to choose wisely.
>
>* If you are comparing against zero, then relative epsilons and ULPs based comparisons are usually meaningless. You’ll need to use an absolute epsilon, whose value might be some small multiple of FLT_EPSILON and the inputs to your calculation. Maybe.
>* If you are comparing against a non-zero number then relative epsilons or ULPs based comparisons are probably what you want. You’ll probably want some small multiple of FLT_EPSILON for your relative epsilon, or some small number of ULPs. An absolute epsilon could be used if you knew exactly what number you were comparing against.
>* If you are comparing two arbitrary numbers that could be zero or non-zero then you need the kitchen sink. Good luck and God speed.

Maybe Kronecker was right: _"God made the integers, all else is the work of man."_. :confused:

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/FilterLongReadAlignmentsSAMSpark.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
```diff
@@ -248,38 +258,44 @@ private static boolean isFirstInPartner(final ChimericAlignment ca, final SAMSeq
         }
     }
 
+    private static final class BreakpointsInferenceFactory {
```
cwhelan: Oh, sorry, I didn't notice this. I don't think you need a separate class to hold this method, you could just make this a static method on the containing class, unless you prefer it this way.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/FilterLongReadAlignmentsSAMSpark.java
cwhelan: Sure, but I was just wondering if there was some "effective" difference that's larger than the ULP-type differences but at which point it doesn't matter if the difference is due to floating point representations or the actual input values. For example, if there are two configurations made of five alignments and the only difference in score is due to one containing an alignment with mapq 51 and the other replaces it with an alignment with mapq 52, we might actually want to treat those as roughly equivalent and run them through your disambiguation procedure, no? So whatever that delta is might be an acceptable epsilon for this method. Anyway, this is not a big deal, so feel free to do whatever you want.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/NovelAdjacencyReferenceLocations.java
SHuang-Broad: Will be changed in the final merge. Thanks!

PR #3579, opened 2017-09-13
updates sv manager and cluster creation scripts to utilize dataproc

### LINKED ISSUE #3574: Update sv pipeline scripts for dataproc timed self-termination
Currently possible only if using `gcloud beta create ...`. Need to wait until out of beta phase.
Work is already done in branch `sh_dataproc_timeout`.

SHuang-Broad:
cluster timed self-termination (beta) feature.
Fixes #3574 considering that several people are suggesting beta from gcloud is safe enough.

### DIFF
--- a/scripts/sv/create_cluster.sh
+++ b/scripts/sv/create_cluster.sh
@@ -4,23 +4,26 @@
 
 set -eu
 
-if [[ "$#" -lt 6 ]]; then
+if [[ "$#" -lt 8 ]]; then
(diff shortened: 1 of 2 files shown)


SHuang-Broad:
Thanks @TedBrookings !

I did a creation and self-termination test run (without actually running the analysis to save time), and it worked as expected.

Also implemented the suggested changes:
1. documentation in help
2. upfront def instead of in the creation block

PR #3581, opened 2017-09-14
Upgrade Mockito from 1.10.19 -> 2.10.0.

cmnbroad:
I've had several Travis test failures (on my picard removal branch) that appear to be failures during kryo serialization of a mocked ReferenceMultiSource object (based on the failing class name, (org.broadinstitute.hellbender.engine.datasources.ReferenceMultiSource$$EnhancerByMockitoWithCGLIB$$b0dc631f, which looks like the CGLIB names mentioned [here](https://github.com/mockito/mockito/issues/319)).

We're on an ancient version of mockito anyway, and newer versions no longer use cglib, so it seemed like a good time to upgrade. To do so I also had to replace usage of the method getArgumentAt, which has been [deprecated](https://github.com/mockito/mockito/pull/373) in favor of getArgument.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -205,7 +205,7 @@ dependencies {
     //needed for DataflowAssert
     testCompile 'org.hamcrest:hamcrest-all:1.3'
     testCompile 'junit:junit:4.12'
-    testCompile "org.mockito:mockito-core:1.10.19"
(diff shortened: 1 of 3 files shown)


cmnbroad:
Note: it looks like this doesn't fix the ConcurrentModificationException during serialization while running the unit tests on my Picard branch.

PR #3584, opened 2017-09-15
Updating Mutect command line arguments and WDL

davidbenjamin:
@LeeTL1220 

* Updates command line arguments to conform to new GATK standard.
* Removes M2 wdl hacks no longer needed as of Cromwell 27


scripts/mutect2_wdl/README.md
```diff
@@ -195,7 +192,7 @@ gs://broad-dsde-methods/takuto/na12878-crsp-ice/SM-612V3.bam    gs://broad-dsde-
   "Mutect2_Multi.variants_for_contamination_index": "/data/m2/gnomad-common-biallelic-snps.vcf.idx",
   "Mutect2_Multi.is_run_orientation_bias_filter": true,
   "Mutect2_Multi.is_run_oncotator": true,
-  "Mutect2_Multi.m2_docker": "broadinstitute/gatk:1.0.0.0-alpha1.2.4",
+  "Mutect2_Multi.gatk_docker": "broadinstitute/gatk:1.0.0.0-alpha1.2.4",
```
LeeTL1220: Can we update the example value here?  It's way out of date.  How about `broadinstitute/gatk:4.beta.5`?

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -4,8 +4,8 @@
 #  gatk4_jar: java jar file containing gatk 4 (protected)
 #  intervals: genomic intervals
 #  ref_fasta, ref_fasta_index, ref_dict: reference genome, index, and dictionary
-#  tumor_bam, tumor_bam_index, and tumor_sample_name: self-explanatory
-#  normal_bam, normal_bam_index, and normal_sample_name: self-explanatory
+#  tumor_bam, tumor_bam_index: self-explanatory
```
LeeTL1220: I believe Geraldine has some good documentation for putting at the head of the M2 WDL.  We should probably bring that into the GATK repo.  Will it need edits, @vdauwera ?

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -190,10 +174,6 @@ workflow Mutect2 {
         File filtered_vcf = Filter.filtered_vcf
         File filtered_vcf_index = Filter.filtered_vcf_index
         File contamination_table = Filter.contamination_table
-        Array[String] tumor_bam_sample_names = M2.tumor_bam_sample_name
```
LeeTL1220: These are what are extracted from the bam file?  Are you sure we can delete these?  These are useful when calling from parent workflows....

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -232,85 +215,60 @@ task M2 {
...
-      preemptible: "${preemptible_attempts}"
+        docker: "${gatk_docker}"
+        memory: select_first([mem, 5]) + " GB"
+        disks: "local-disk " + select_first([disk_space_gb, 100]) + " HDD"
+        preemptible: select_first([preemptible_attempts, 2])
   }
 
   output {
-    File output_vcf = "${output_vcf_name}.vcf"
+    File output_vcf = glob("*.vcf")[0]
```
LeeTL1220: This seems a bit dangerous/fragile......

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -542,23 +520,23 @@ task oncotate_m2 {
...
         bootDiskSizeGb: 12
-        disks: "local-disk 100 HDD"
-        preemptible: "${preemptible_attempts}"
+        disks: "local-disk " + select_first([disk_space_gb, 100]) + " HDD"
+        preemptible: select_first([preemptible_attempts, 2])
     }
 
     output {
-        File oncotated_m2_maf="${case_id}.maf.annotated"
+        File oncotated_m2_maf=glob("*.maf.annotated")[0]
```
LeeTL1220: These globs[0] worry me.... I see what you are trying to do.  It's a bit fragile, but I admit the risk is low.  

And I'm generally okay with passing around entity_ids.... in fact, it can be useful when working in a data model such as FireCloud

scripts/mutect2_wdl/mutect2_multi_sample_template.json
```diff
@@ -14,9 +14,8 @@
   "Mutect2_Multi.variants_for_contamination_index": "$__variants_for_contamination_index__",
   "Mutect2_Multi.is_run_orientation_bias_filter": true,
   "Mutect2_Multi.is_run_oncotator": "$__is_run_oncotator__",
-  "Mutect2_Multi.m2_docker": "broadinstitute/gatk-protected:1.0.0.0-alpha1.2.4",
+  "Mutect2_Multi.gatk_docker": "broadinstitute/gatk-protected:1.0.0.0-alpha1.2.4",
```
LeeTL1220: Update example docker image, please.  This is way out of date.

scripts/mutect2_wdl/mutect2_template.json
```diff
@@ -19,9 +17,8 @@
   "Mutect2.variants_for_contamination_index": "$__variants_for_contamination_index__",
   "Mutect2.is_run_orientation_bias_filter": "$__is_run_orientation_bias_filter__",
   "Mutect2.is_run_oncotator": "$__is_run_oncotator__",
-  "Mutect2.m2_docker": "broadinstitute/gatk-protected:1.0.0.0-alpha1.2.4",
+  "Mutect2.gatk_docker": "broadinstitute/gatk-protected:1.0.0.0-alpha1.2.4",
```
LeeTL1220: Update docker image

scripts/mutect2_wdl/unsupported/README.md
```diff
@@ -22,8 +22,8 @@ The following resources:
 * A reference .fasta file, along with accompanying .fasta.fai and .dict files.
 * A gatk4 java .jar file.
 * A Picard java .jar file.
-* Three lists of .bam files -- one each for 5-plex, 10-plex and 20-plex replicates -- where each row has the format <bam_file.bam></TAB><bam_index.bai></TAB><sample_name>
-* A list of .bam files of the specificity validation's replicates.
+* Three lists of .bam files -- one each for 5-plex, 10-plex and 20-plex replicates -- where each row has the format <bam_file.bam></TAB><bam_index.bai>
```
LeeTL1220: Out of curiosity, could this cross pairing be done in the WDL with a `cross` call?

scripts/mutect2_wdl/unsupported/mutect2_opt.wdl
```diff
@@ -214,117 +193,93 @@ task M2 {
...
-      preemptible: "${preemptible_attempts}"
+        docker: "${gatk_docker}"
+        memory: select_first([mem, 5]) + " GB"
+        disks: "local-disk " + select_first([disk_space_gb, 100]) + " HDD"
+        preemptible: select_first([preemptible_attempts, 2])
   }
 
   output {
-    File output_vcf = "${output_vcf_name}.vcf"
+    File output_vcf = glob("*.vcf")[0]
```
LeeTL1220: Hey, @jsotobroad and @meganshand , how do you feel about this glob statement?

jsotobroad: We don't really like the globbed output for just one file but we can see why you would want to do it.  We had another way to get rid of all the workflow input sample names and no more preprocessing step.  We were thinking of taking out the `GetSampleName` calls out of M2 into its own task (because its awkward calling the same exact thing for every scatter of M2 and they all return the same output).  Use the output of this new GetSampleName task to get the vcf output name with something like

```
  String output_vcf = if defined(normal_bam) then GetSampleNameTask.normal_sample_name + "-vs-" + GetSampleNameTask.tumor_sample_name else GetSampleNameTask.tumor_sample_name + "-tumor-only"
```

and use this string as as the output vcf name as it was used in the original wdl.  Does that sound like a good solution to you guys?

davidbenjamin: @LeeTL1220 @jsotobroad I was trying to completely eliminate getting the sample name, thereby getting rid of one task entirely.  I've heard that there is overhead for every task on the cloud and I was perhaps being paranoid.  If this isn't a big deal I will do it Jose's way.

jsotobroad: Money wise it is very inexpensive deal to run and workflow wise, because the workflow has to run `SplitIntervals` before it can run M2 the total runtime of the workflow shouldn't be affected either.  It's definitely something that should be able to made nice with future cromwell/wdl features hopefully....

scripts/mutect2_wdl/README.md
davidbenjamin: done

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: I will restore them.

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: fixed

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: done

scripts/mutect2_wdl/mutect2_multi_sample_template.json
davidbenjamin: done

scripts/mutect2_wdl/mutect2_template.json
davidbenjamin: done

scripts/mutect2_wdl/unsupported/README.md
davidbenjamin: Seems to me the answer is yes, once wdl gets a composite bam + index data type, but for now it would probably be contrived.

scripts/mutect2_wdl/unsupported/mutect2_opt.wdl
davidbenjamin: I'm satisfied and will do it @jsotobroad's recommended way.

davidbenjamin:
@LeeTL1220 ready for re-review.  This was passing prior to squashing a few minutes ago.

scripts/mutect2_wdl/mutect2.wdl
LeeTL1220: @davidbenjamin Did you do this?

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -203,6 +216,40 @@ workflow Mutect2 {
...
+  File? gatk4_jar_override
+
+  # Runtime parameters
+  Int? mem
+  String gatk_docker
+  Int? preemptible_attempts
+  Int? disk_space_gb
+
+  command <<<
+      # Use GATK Jar override if specified
```
LeeTL1220: Do you really want a standalone task just to extract the sample name?

LeeTL1220: Especially, one that will localize a bam file in the cloud...?

davidbenjamin: > Did you do this?
I deleted "and tumor_sample_name"

davidbenjamin: > Do you really want a standalone task just to extract the sample name?

I'm neutral.  Whatever @jsotobroad thinks is best.

davidbenjamin:
@LeeTL1220 What do you think of this latest commit?  Getting sample names and the output vcf is now absorbed into the M2 task.

davidbenjamin:
@LeeTL1220 the ball is in your court.

LeeTL1220:
@davidbenjamin Sorry for the delay.  I just re-reviewed.  Merge away!

PR #3586, opened 2017-09-18
Use a handmade TestReferenceMultiSource in tests instead of a mock.

cmnbroad:
I've had ConcurrentModificationExceptions several times when mocked ReferenceMultiSource objects are both mocked and broadcast ([for example](https://travis-ci.org/broadinstitute/gatk/jobs/275982454)). Mocks appear to be mutated, even when the mocked object is immutable, so serialization can fail. This replaces the mock object with a faked one.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceMultiSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceMultiSource.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.hellbender.engine.datasources;
 
 import com.google.cloud.dataflow.sdk.options.PipelineOptions;
+import com.google.common.annotations.VisibleForTesting;
 import org.broadinstitute.hellbender.utils.SerializableFunction;
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.reference.ReferenceSequenceFileFactory;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceMultiSource.java
```diff
@@ -20,14 +20,21 @@
...
     private static final long serialVersionUID = 1L;
 
     private ReferenceSource referenceSource;
     private SerializableFunction<GATKRead, SimpleInterval> referenceWindowFunction;
 
+    // We can't use a real mock for this class in test code that requires an instance of this class,
+    // since in some cases they're used as a Spark broadcast variable. Mocks are mutated when they're
+    // accessed, which can result in ConcurrentModificationExceptions during serialization/broadcast.
+    // So we need to allow test code to create a subclass that provides a "fake" implementation of this
+    // class without actually being a mock.
```
droazen: This comment might be more appropriate in a test class -- here you should just document concisely why the default constructor is protected.

src/test/java/org/broadinstitute/hellbender/engine/spark/AddContextDataToReadSparkUnitTest.java
```diff
@@ -84,11 +80,37 @@ public void addContextDataTest(List<GATKRead> reads, List<GATKVariant> variantLi
...
+        public SAMSequenceDictionary getReferenceSequenceDictionary(final SAMSequenceDictionary optReadSequenceDictionaryToMatch) {
+            return sequenceDictionary;
+        }
+
+        @Override
+        public boolean isCompatibleWithSparkBroadcast(){
+            return true;
+        }
+
+        public SerializableFunction<GATKRead, SimpleInterval> getReferenceWindowFunction() {
```
droazen: `@Override`

PR #3588, opened 2017-09-18
updating to newer htsjdk snapshot

lbergelson:
updating to a new htsjdk snapshot and updating our VariantContextWriters to compile

test may fail on this....

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ repositories {
     mavenLocal()
 }
 
-final htsjdkVersion = System.getProperty('htsjdk.version','2.11.0-4-g958dc6e-SNAPSHOT')
+final htsjdkVersion = System.getProperty('htsjdk.version','2.11.0-18-ga35b420-SNAPSHOT')
 final sparkVersion = System.getProperty('spark.version', '2.0.2')
 final hadoopBamVersion = System.getProperty('hadoopBam.version','7.9.0')
 final genomicsdbVersion = System.getProperty('genomicsdb.version','0.7.0-proto-3.0.0-beta-1')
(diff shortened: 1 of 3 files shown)


lbergelson:
@jean-philippe-martin We're seeing the same errors here as in yours pr.  So it's something in the htsjdk update...

cmnbroad:
Since this is happening at the intersection of CRAM and indexing, I looked into it a bit. It can be readily reproduced if the failing test is run through gradle, but not from IntelliJ, which turns out to be because when it runs through gradle, its using asyncIO for BAM writing, but when run from IntelliJ its using sync io.

I'm not sure why this issue is showing up now, but in the cases that fail, the async writer thread is being interrupted by [this code](https://github.com/samtools/htsjdk/blob/912c28bec415c430b43515652ccaf13222b07e7b/src/main/java/htsjdk/samtools/util/AbstractAsyncWriter.java#L75) which is called from the main thread when the PrintReads closeTool method closes the writer (I added a trace statement to verify this). When I run in IntelliJ and force async io to be used, the interrupt call still happens, but there is no test failure.

cmnbroad:
One short term option would be to turn of async writing until we figure out the right fix.

lbergelson:
The regression was introduced in https://github.com/samtools/htsjdk/commit/1ade6c921628e540f1cab0dfadfbdc536f702d73

lbergelson:
the htsjdk regression was fixed in https://github.com/samtools/htsjdk/pull/992

lbergelson:
@jean-philippe-martin I've updated the htsjdk snapshot, you should be unblocked after a rebase.

jean-philippe-martin:
Thank you very much @lbergelson !

Issue #3591, opened 2017-09-19
jenkins spark tests failing with 404


lbergelson:
The jenkins spark tests are failing with the following error:

This seems to have been introduced in https://github.com/broadinstitute/gatk/pull/3576

```
code:      0
message:   Error code 404 trying to get security access token from Compute Engine metadata for the default service account. This may be because the virtual machine instance does not have permission scopes specified.
reason:    null
location:  null
retryable: false
com.google.cloud.storage.StorageException: Error code 404 trying to get security access token from Compute Engine metadata for the default service account. This may be because the virtual machine instance does not have permission scopes specified.
	at com.google.cloud.storage.spi.v1.HttpStorageRpc.translate(HttpStorageRpc.java:189)
	at com.google.cloud.storage.spi.v1.HttpStorageRpc.get(HttpStorageRpc.java:339)
	at com.google.cloud.storage.StorageImpl$5.call(StorageImpl.java:197)
	at com.google.cloud.storage.StorageImpl$5.call(StorageImpl.java:194)
	at shaded.cloud_nio.com.google.api.gax.retrying.DirectRetryingExecutor.submit(DirectRetryingExecutor.java:91)
	at com.google.cloud.RetryHelper.runWithRetries(RetryHelper.java:54)
	at com.google.cloud.storage.StorageImpl.get(StorageImpl.java:194)
	at com.google.cloud.storage.contrib.nio.CloudStorageFileSystemProvider.checkAccess(CloudStorageFileSystemProvider.java:614)
	at java.nio.file.Files.exists(Files.java:2385)
	at htsjdk.samtools.util.IOUtil.assertFileIsReadable(IOUtil.java:346)
	at org.broadinstitute.hellbender.engine.ReadsDataSource.<init>(ReadsDataSource.java:206)
	at org.broadinstitute.hellbender.engine.ReadsDataSource.<init>(ReadsDataSource.java:162)
	at org.broadinstitute.hellbender.engine.ReadsDataSource.<init>(ReadsDataSource.java:118)
	at org.broadinstitute.hellbender.engine.ReadsDataSource.<init>(ReadsDataSource.java:87)
	at org.broadinstitute.hellbender.engine.spark.datasources.ReadsSparkSource.getHeader(ReadsSparkSource.java:182)
	at org.broadinstitute.hellbender.engine.spark.GATKSparkTool.initializeReads(GATKSparkTool.java:390)
[...]

PR #3576, opened 2017-09-13
Move to google-cloud-java:0.23.1-alpha.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/3500.

PR #3592, opened 2017-09-19
Revert "Move to google-cloud-java:0.23.1-alpha. (#3576)"

lbergelson:
This reverts commit b47838c9a5fa172ed6669ed4872b04d91c962a85.
This commit introduced a major issue for spark https://github.com/broadinstitute/gatk/issues/3591.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -123,7 +123,7 @@ dependencies {
     // Using the shaded version to avoid conflicts between its protobuf dependency
     // and that of Hadoop/Spark (either the one we reference explicitly, or the one
     // provided by dataproc).
-    compile 'com.google.cloud:google-cloud-nio:0.23.1-alpha:shaded'
+    compile 'com.google.cloud:google-cloud-nio:0.20.4-alpha-20170727.190814-1:shaded'
     compile 'com.google.cloud.genomics:google-genomics-dataflow:v1beta2-0.15'
     compile 'com.google.cloud.genomics:gatk-tools-java:1.1'
     // this comes built-in when running on Google Dataproc, but the library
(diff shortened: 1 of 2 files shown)


lbergelson:
@mwalker174 Could you check if this fixes your spark problem?

droazen:
@lbergelson We really need to stay on the newest gcloud release -- it contains the official release of our NIO fixes! We should try to work through any issues rather than reverting to the previous (unreleased) snapshot.

cmnbroad:
FWIW there is already a newer version (0.24.0) than the (0.23.1) one we're on we could try.

cmnbroad:
PR is #3594 if we want to try moving forward one release.

droazen:
@cmnbroad Worth a try. Do we understand the underlying issue, though? If it's just the static initializer in `BaseTest`, perhaps we could refactor that into a `@BeforeSuite`?

cmnbroad:
I don't think the warning message is coming from BaseTest, its coming directly from SparkContextFactory, which is executing static code that I think it shouldn't (see the stack I just added to #3591). Its unclear to me if that is actually CAUSING #3591, or just causing the message to display.

mwalker174:
It turns out I was mistaken that setting the environment variables fixes the problem (stupid error on my part). It's possible the BaseTest message is unrelated. I haven't tested this branch out yet, but building from the commit immediately before the update works.

I am going to try the next version to see if it helps.

droazen:
@mwalker174 Can you try another run with the patch in https://github.com/broadinstitute/gatk/pull/3594, but this time comment out the call to `getGcsHadoopAdapterTestProperties()` in `SparkContextFactory` line 59?

mwalker174:
@droazen still giving error:

com.google.cloud.storage.StorageException: Error code 404 trying to get security access token from Compute Engine metadata for the default service account. This may be because the virtual machine instance does not have permission scopes specified.

droazen:
@mwalker174 Ok, I've asked our Google collaborator @jean-philippe-martin to comment on https://github.com/broadinstitute/gatk/issues/3591. It looks like there were some authentication-related changes in the newer gcloud releases that could explain the error. It may be that we just need to update our client code and/or project IAM settings.

droazen:
@lbergelson has also opened https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2453 to ask the `google-cloud-java` team for help.

droazen:
Since Google hasn't gotten back to us on this yet, it seems like we're forced to revert back to the older gcloud snapshot for now. Merging!

PR #3613, opened 2017-09-26
fix 3612 

SHuang-Broad:
by removing SGA associated classes, tools and tests, which are mostly skipped anyway


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/sga/ExternalCommandlineProgramModule.java
```diff
@@ -1,160 +0,0 @@
...
-import java.io.Serializable;
-import java.nio.charset.Charset;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.List;
-
-/**
- * A class for experimenting with cmd line programs (such as bwa and sga) that's not part of GATK (or has Java bindings yet).
- */
-abstract class ExternalCommandlineProgramModule {
```
cwhelan: Do you want to keep this one around? It's not SGA specific and seems like it could be useful sometime in the future.

SHuang-Broad: Do you think it would be reasonable, to keep around the `SGAModule` as well, for illustration purpose?

SHuang-Broad: I'll mark it `@Deprecated` as is already the case.

cwhelan:
👍

SHuang-Broad:
ping @lbergelson if you want to take a look, especially the leftover class for shelling out tasks to cmd line programs

Issue #3532, opened 2017-08-29
GKL  failures in MacOS


magicDGS:
## System

* GATK4 a1eee32e84c21c2f265d248c5f47789ae0ba2b37
* Mac OS X 10.11.6 x86_64
* machdep.cpu.features: FPU VME DE PSE TSC MSR PAE MCE CX8 APIC SEP MTRR PGE MCA CMOV PAT PSE36 CLFSH DS ACPI MMX FXSR SSE SSE2 SS HTT TM PBE SSE3 DTES64 MON DSCPL VMX SMX EST TM2 SSSE3 CX16 TPR PDCM SSE4.1 SSE4.2 POPCNT
* machdep.cpu.extfeatures: SYSCALL XD EM64T LAHF RDTSCP TSCI
* java version "1.8.0_60"
* Java(TM) SE Runtime Environment (build 1.8.0_60-b27)
* Java HotSpot(TM) 64-Bit Server VM (build 25.60-b23, mixed mode)

## Error

When updating a downstream project with the latest master and running the integration/unit tests with gradle, it generates the following error.

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGILL (0x4) at pc=0x00000001236427f4, pid=4010, tid=20739
#
# JRE version: Java(TM) SE Runtime Environment (8.0_60-b27) (build 1.8.0_60-b27)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.60-b23 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# C  [libgkl_compression7227189416687158431.dylib+0x17f4]  Java_com_intel_gkl_compression_IntelDeflater_resetNative+0x164
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/daniel/workspaces/ReadTools/hs_err_pid4010.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

Error report file: [hs_err_pid4010.log.txt](https://github.com/broadinstitute/gatk/files/1259963/hs_err_pid4010.log.txt)

## Forcing other GKL versions

* 0.5.2 (working)
* 0.5.3 (failing)
* 0.5.5 (failing)
* 0.5.6 (failing)
[...]

PR #3615, opened 2017-09-26
Upgrade to GKL 0.7

erniebrau:
Changes include: AVX-512 on non-Mac Skylake + ISA-L 2.19 for level 1 compression. AVX-512 support for OS X is planned for the near future.


erniebrau:
We think this release also addresses #3532, but have no way of testing ourselves. @magicDGS -- can you test if this is the case?

magicDGS:
This is working now @lbergelson - thanks @erniebrau!

PR #3617, opened 2017-09-26
Renames StructuralVariantContext to SVContext.

vruano:
This is more in-line with other SV class names.

Also SVContext.create is renamed to SVContext.of


vruano:
@SHuang-Broad just a simple class rename and a method rename, so basically this just about whether you agree.

SHuang-Broad:
I like it. 👍

PR #3620, opened 2017-09-27
Picard integration.

cmnbroad:
Notes:

- Classes in hellbender/tools/picard/analysis/artifacts are removed and replaced with Picard versions (except Transition, which is not public in Picard).
- GATK version of GatherVcfs is retained, and the Picard version is masked out - is this what we want ?
- The non-Spark GATK metrics tools have been removed and replaced with the Picard versions. The test data is retained (but moved) since its used by the Spark metrics tool tests.

Additional changes we'll want to make separately to minimize the complexity of this PR:
  - Eliminate the download of picard.jar from the GATK WDL tests and update the WDL to run Picard tools through GATK.
  - Unify and merge the Picard and GATK program groups. These are similar, but not identical, and the combined result has artificial/duplicate groups.
  - Normalize the confusing mix of Alpha/Beta/Experimental tags and comments.
  - Add unified doc and tab-completion tasks that include Picard.
  - Remove and replace SamComparison and Transition classes with the Picard versions.
  - Fix GATK CompareBaseQualities (its a PicardCommandLineProgram).


src/main/java/org/broadinstitute/hellbender/cmdline/PicardCommandLineProgramExecutor.java
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * This is the entry point for Picard tools that are called from GATK.
+     */
+    public Object instanceMain(final String[] argv) {
+        return picardCommandLineProgram.instanceMain(argv);
+    }
+
+    @Override
+    public Object doWork() {
+        // This method should never be called directly. Call instanceMain instead.
```
lbergelson: Why does this need to be public?

src/test/java/org/broadinstitute/hellbender/cmdline/PicardProgramExecutorIntegrationTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        // use GATK-style lower case names, which are rejected by Picard
+        final String[] args = {
+                "--input", input.getAbsolutePath(),
+                "--output", outfile.getAbsolutePath(),
+                "--TRUNCATE_SEQUENCE_NAMES_AT_WHITESPACE", "TRUE",
+                "--LINE_LENGTH", "5",
+        };
+
+        Assert.assertNotEquals(runCommandLine(args), 0);
```
lbergelson: Why do we get a bad return value here instead of an exception bubbling up?  If the args are no good shouldn't barclay be throwing an exception?

cmnbroad:
@LeeTL1220 Can you review, or appoint someone to review, the Mutect2 WDL and the tools/exome and tools/picard/analysis/artifacts packages part of this. Basically, I replaced most of the classes in the artifacts package with their Picard analogs, which also affected some embedded class names in metrics files (SequencingArtifactMetrics IIRC). The are multiple commits in here that delete various things, but the substantive changes are the last commit ("Integrate actual Picard tools via Picard"). Once this is in, we'll probably want to do [this](https://github.com/broadinstitute/gatk/issues/3625) separately as well.

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -241,16 +247,22 @@ private static boolean printStackTraceOnUserExceptions() {
      * Returns the command line program specified, or prints the usage and exits with exit code 1 *
      */
     private static CommandLineProgram extractCommandLineProgram(final String[] args, final List<String> packageList, final List<Class<? extends CommandLineProgram>> classList, final String commandLineName) {
+
         /** Get the set of classes that are our command line programs **/
         final ClassFinder classFinder = new ClassFinder();
         for (final String pkg : packageList) {
+            classFinder.find(pkg, picard.cmdline.CommandLineProgram.class);
```
magicDGS: This is getting really complicated for combining projects/toolkits and re-use this main. I guess that my idea of adding a really simple `CommandLineProgram` interface in Barclay is worth now and something simple to implement (just with a single method `instanceMain(String[] args)` to being able to combine toolkits without this big changes. I will submit a PR to add that simple layout to simplify things.

In addtion, a simple `Main` class could be implemented in Barclay to keep some of this logic and simplify things a bit, but that's a bigger change. I would submit a different PR for this to Barclay.

magicDGS: By the way, without the interface in Barclay, downstream projects are unable to specify single Picard classes in `getClassList()`, and it would require yet another method to return picard classess...which is kind of cumbersome...

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -294,6 +314,23 @@ private static CommandLineProgram extractCommandLineProgram(final String[] args,
         return null;
     }
 
+    // Return a list of (usually Picard) classes which we want to mask from the list of runnable tools,
+    // usually because want to favor the like-named GATK tool. Any CommandLineProgram that shares a simple class
+    // name with a Picard tool must be on this list. Failure to mask it out will result in name collisions of
+    // simple class names in Main.
+    private static Set<String> getMaskedToolClasses() {
```
magicDGS: Similarly to `getPackageList()` and `getClassList()`, this should be a protected non-static method, overridable but downstream projects, to provide a set of tools to remove from the command line. Downstream projects require more control on this.

On the other hand, I suggest to change the signature to return `List<Class<? extends CommandLineProgram>>`, but that should return a common interface for tools (both Picard and GATK).

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -241,16 +247,22 @@ private static boolean printStackTraceOnUserExceptions() {
...
+            classFinder.find(pkg, picard.cmdline.CommandLineProgram.class);
             classFinder.find(pkg, CommandLineProgram.class);
         }
         String missingAnnotationClasses = "";
         final Set<Class<?>> toCheck = classFinder.getClasses();
         toCheck.addAll(classList);
         final Map<String, Class<?>> simpleNameToClass = new LinkedHashMap<>();
+        final Set<String> maskedPicardToolClasses = getMaskedToolClasses();
         for (final Class<?> clazz : toCheck) {
+            if (maskedPicardToolClasses.contains(clazz.getName())) {
```
magicDGS: Instead of using `clazz.getName()`, here a direct comparison of classes will be better if there is a common interface....

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -282,7 +294,15 @@ private static CommandLineProgram extractCommandLineProgram(final String[] args,
                 if (simpleNameToClass.containsKey(args[0])) {
                     final Class<?> clazz = simpleNameToClass.get(args[0]);
                     try {
-                        return (CommandLineProgram) clazz.newInstance();
+                        if (clazz.getName().startsWith("picard.")) {
+                            // wrap the PicardCommandLinePrograms in a PicardCommandLineProgramExecutor
+                            @SuppressWarnings("unchecked")
+                            Class<? extends picard.cmdline.CommandLineProgram> picardClazz =
+                                    (Class<? extends picard.cmdline.CommandLineProgram>) clazz;
+                            return new PicardCommandLineProgramExecutor(picardClazz.newInstance());
```
magicDGS: The common interface in Barclay will reduce the complexity here and keep the previous code untouchable...

src/main/java/org/broadinstitute/hellbender/cmdline/PicardCommandLineProgramExecutor.java
```diff
@@ -0,0 +1,33 @@
+package org.broadinstitute.hellbender.cmdline;
+
+import org.broadinstitute.hellbender.exceptions.GATKException;
+
+/**
+ * Adapter shim for use within GATK to run Picard tools. Note that this class does not have
+ * it's own CommandLineProgramProperties annotation, and isn't intended to be run from the
+ * command line.
+ */
+public class PicardCommandLineProgramExecutor extends CommandLineProgram {
```
magicDGS: I do not like this hack at all...for downstream projects is going to be very difficult to include/exclude picard tools...

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -348,11 +348,7 @@ task CollectSequencingArtifactMetrics {
 
   command {
         set -e
-        java -Xmx4G -jar ${picard_jar} CollectSequencingArtifactMetrics I=${tumor_bam} O="metrics" R=${ref_fasta} VALIDATION_STRINGENCY=LENIENT
-
-        # Convert to GATK format
-        sed -r "s/picard\.analysis\.artifacts\.SequencingArtifactMetrics\\\$PreAdapterDetailMetrics/org\.broadinstitute\.hellbender\.tools\.picard\.analysis\.artifacts\.SequencingArtifactMetrics\$PreAdapterDetailMetrics/g" \
-            "metrics.pre_adapter_detail_metrics" > "gatk.pre_adapter_detail_metrics"
+        java -Xmx4G -jar ${picard_jar} CollectSequencingArtifactMetrics I=${tumor_bam} O="gatk" R=${ref_fasta} VALIDATION_STRINGENCY=LENIENT
```
LeeTL1220: This is great assuming it works.

LeeTL1220: And I confirmed that FilterByOrientationBias is being run.

src/main/java/org/broadinstitute/hellbender/utils/test/SamAssertionUtils.java
```diff
@@ -377,11 +377,13 @@ public static void assertCRAMContents(final File putativeCRAMFile) {
 
     private static void sortSam(final File input, final File output, final File reference, final ValidationStringency stringency) {
         final SortSam sort = new SortSam();
-        sort.INPUT = input;
-        sort.OUTPUT = output;
-        sort.REFERENCE_SEQUENCE = reference;
-        sort.SORT_ORDER = SAMFileHeader.SortOrder.coordinate;
-        sort.VALIDATION_STRINGENCY = stringency;
-        sort.runTool();
+        // this is running a Picard tools do don't use ArgumentBuilder
```
LeeTL1220: typo "tools" --> "tool"

src/main/java/org/broadinstitute/hellbender/cmdline/PicardCommandLineProgramExecutor.java
cmnbroad: It shouldn't be - good catch.

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -40,6 +41,10 @@
          * to think about number formatting issues.
          */
         Utils.forceJVMLocaleToUSEnglish();
+
+        // Turn off the Picard legacy parser and opt in to Barclay syntax for Picard tools. This should be replaced
+        // with a config setting once PR https://github.com/broadinstitute/gatk/pull/3447 is merged.
+        System.setProperty("picard.useLegacyParser", "false");
```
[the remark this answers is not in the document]
droazen: I don't think it hurts to allow this to be configured by users who know what they're doing.

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -282,7 +294,15 @@ private static CommandLineProgram extractCommandLineProgram(final String[] args,
                 if (simpleNameToClass.containsKey(args[0])) {
                     final Class<?> clazz = simpleNameToClass.get(args[0]);
                     try {
-                        return (CommandLineProgram) clazz.newInstance();
+                        if (clazz.getName().startsWith("picard.")) {
```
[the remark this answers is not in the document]
droazen: Yes, I agree that an `instanceof` check would be slightly better here.

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -294,6 +314,23 @@ private static CommandLineProgram extractCommandLineProgram(final String[] args,
...
+    // simple class names in Main.
+    private static Set<String> getMaskedToolClasses() {
+        final Set<String> maskedPicardToolClasses = new HashSet<>();
+
+        // First, add the GATK Picard executor shim, since it's only a shim, not a runnable tool, and has no
+        // CommandLineProgramProperties of it's own.
+        maskedPicardToolClasses.add("org.broadinstitute.hellbender.cmdline.PicardCommandLineProgramExecutor");
+
+        // Mask out picard GatherVcfs in favor of GATK GatherVcfs
+        maskedPicardToolClasses.add("picard.vcf.GatherVcfs");
```
droazen: I would eliminate the masking of Picard's `GatherVcfs` entirely, and just rename the GATK version. Suggested name for the GATK version: `GatherVcfsCloud`

src/main/java/org/broadinstitute/hellbender/cmdline/PicardCommandLineProgramExecutor.java
droazen: I think this class is fine -- it's just a simple adapter/shim, which is a pretty standard design pattern when integrating multiple codebases.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectBaseDistributionByCycleSpark.java
```diff
@@ -229,7 +230,7 @@ protected void saveResults(final MetricsFile<?, Integer> metrics, final SAMFileH
                 plotSubtitle = StringUtil.asEmptyIfNull(readGroups.get(0).getLibrary());
             }
             final RScriptExecutor executor = new RScriptExecutor();
-            executor.addScript(new Resource(CollectBaseDistributionByCycle.R_SCRIPT, CollectBaseDistributionByCycle.class));
+            executor.addScript(new Resource(R_SCRIPT, CollectBaseDistributionByCycleSpark.class));
```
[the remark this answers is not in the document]
droazen: It would be good to add unit tests for all of these R scripts that just assert that they are resolvable.

src/test/java/org/broadinstitute/hellbender/cmdline/PicardProgramExecutorIntegrationTest.java
```diff
@@ -1,53 +1,64 @@
...
                 "--input", input.getAbsolutePath(),
-                "--output", input.getAbsolutePath(),       //will blow up
+                "--output", outfile.getAbsolutePath(),
                 "--TRUNCATE_SEQUENCE_NAMES_AT_WHITESPACE", "TRUE",
                 "--LINE_LENGTH", "5",
         };
 
-        runCommandLine(args);
-
+        Assert.assertNotEquals(runCommandLine(args), 0);
```
droazen: Add a comment highlighting that you're expecting a non-zero return code here.

src/test/java/org/broadinstitute/hellbender/cmdline/PicardProgramExecutorIntegrationTest.java
```diff
@@ -1,53 +1,64 @@
...
     }
 
-    @Test(expectedExceptions = IllegalArgumentException.class)
-    public void testFailNormalize() {
+    @Test
+    public void testPicardNormalizeFastaWithBadArgs() throws IOException {
         final File input = new File(TEST_DATA_PATH, "testfasta.fasta");
+        final File outfile = createTempFile("normalized", ".fasta");
 
+        // use GATK-style lower case names, which are rejected by Picard
```
droazen: Need slightly clearer explanation here of why this fails (eg., mention explicitly that the Picard arg is named `--OUTPUT`)

src/test/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/PreAdapterOrientationScorerUnitTest.java
```diff
@@ -2,8 +2,9 @@
 
 import htsjdk.samtools.metrics.MetricsFile;
 import org.apache.commons.math3.linear.RealMatrix;
-import org.broadinstitute.hellbender.tools.picard.analysis.artifacts.SequencingArtifactMetrics;
-import org.broadinstitute.hellbender.tools.picard.analysis.artifacts.Transition;
+//import org.broadinstitute.hellbender.utils.artifacts.SequencingArtifactMetrics;
```
droazen: Remove commented-out import statement

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/metrics/CollectBaseDistributionByCycleSparkIntegrationTest.java
```diff
@@ -14,7 +14,7 @@
 import java.util.List;
 
 public final class CollectBaseDistributionByCycleSparkIntegrationTest extends CommandLineProgramTest {
-    private static final File TEST_DATA_DIR = new File(getTestDataDir(), "picard/analysis/CollectBaseDistributionByCycle");
+    private static final File TEST_DATA_DIR = new File(getTestDataDir(), "../metrics/analysis/CollectBaseDistributionByCycle");
```
droazen: I'd avoid use of `..` in test data paths

src/main/java/org/broadinstitute/hellbender/utils/test/SamAssertionUtils.java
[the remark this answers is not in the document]
cmnbroad: Yes, the comment is misleading and unnecessary.

src/test/java/org/broadinstitute/hellbender/cmdline/PicardProgramExecutorIntegrationTest.java
cmnbroad: Picard swallows Barclay exceptions and propagates them as a return code since it also has to work with the legacy parser, which uses return codes instead of exceptions.

src/main/java/org/broadinstitute/hellbender/cmdline/PicardCommandLineProgramExecutor.java
```diff
@@ -0,0 +1,33 @@
...
+    final private picard.cmdline.CommandLineProgram picardCommandLineProgram;
+
+    public PicardCommandLineProgramExecutor(final picard.cmdline.CommandLineProgram picardCommandLineProgram) {
+        this.picardCommandLineProgram = picardCommandLineProgram;
+    }
+
+    /**
+     * This is the entry point for Picard tools that are called from GATK.
+     */
+    public Object instanceMain(final String[] argv) {
```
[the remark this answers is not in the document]
cmnbroad: Yes.

cmnbroad:
@magicDGS I think the suggestion of using a common interface is a reasonable proposal, but we can't hold up this PR for that. We'll have to do that in a separate PR once the interface is propagated through the various projects.

src/main/java/org/broadinstitute/hellbender/Main.java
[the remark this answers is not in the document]

src/main/java/org/broadinstitute/hellbender/cmdline/PicardCommandLineProgramExecutor.java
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/artifacts/Transition.java
```diff
@@ -1,4 +1,4 @@
-package org.broadinstitute.hellbender.tools.picard.analysis.artifacts;
+package org.broadinstitute.hellbender.utils.artifacts;
 
 import htsjdk.samtools.util.SequenceUtil;
```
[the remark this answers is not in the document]
cmnbroad: Done. https://github.com/broadinstitute/gatk/issues/3632

src/main/java/org/broadinstitute/hellbender/cmdline/PicardCommandLineProgramExecutor.java
magicDGS: I agree, but it is impossible to include single Picard tools at the moment in `Main` due to this adapter class...

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -88,6 +93,7 @@ protected static void printDecoratedExceptionMessage(final PrintStream ps, final
     protected List<String> getPackageList() {
```
magicDGS: Maybe add to the documentation the packages that are included at the moment, for downstream projects being aware that Picard is also there...

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -241,16 +247,21 @@ private static boolean printStackTraceOnUserExceptions() {
...
+
         /** Get the set of classes that are our command line programs **/
         final ClassFinder classFinder = new ClassFinder();
         for (final String pkg : packageList) {
+            classFinder.find(pkg, picard.cmdline.CommandLineProgram.class);
             classFinder.find(pkg, CommandLineProgram.class);
         }
         String missingAnnotationClasses = "";
         final Set<Class<?>> toCheck = classFinder.getClasses();
         toCheck.addAll(classList);
```
magicDGS: For downstream projects, it may be useful until the CLP interface is included in Barclay (and used in Picard/GATK) to have the possibility to add single Picard classes instead of package-specific...similar to `getClassList`. Maybe the signature of `getClassList` can change in the meanwhile to just return `Class<?>` and check here if it is one of the accepted classes?

PR #3621, opened 2017-09-27
Created VariantAnnotationArgumentCollection and StandardM2Annotation classes

### LINKED ISSUE #3123: Update StandardSomaticAnnotation to reflect M2's default annotations, currently under M2ArgumentCollection as annotation

### LINKED ISSUE #3292: Bundle annotation-related arguments into an ArgumentCollection
Currently tools using the annotation engine are manually defining arguments to control which annotations should be enabled/disabled. See `HaplotypeCallerArgumentCollection` `annotationsToUse`/`annotationGroupsToUse` for an example. We should bundle these into a reusable argument collection, like we did for read filters, as part of the task of making annotations a barclay plugin.

### LINKED ISSUE #3618: Mutect2 should have a standard argument group
I noticed that in the `M2ArgumentCollection` there is a hardcoded list of annotations to include, namely `"Coverage", "DepthPerAlleleBySample", "TandemRepeat", "OxoGReadCounts", "ReadPosition", "BaseQuality", "MappingQuality", "FragmentLength", "StrandArtifact"` and no annotation groups specified. This means that currently if the user specifies any new annotations with the -A argument all of the above specified annotations will be blown away unless the user specifies each one individually. I propose that these annotations should be bundled into a group `StandardM2Annotations` on by default keeping with the pattern set by the Haplotype caller.

jamesemery:
Fixes #3618
Fixes #3292 
Fixes #3123

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/StandardArgumentDefinitions.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/StandardArgumentDefinitions.java
@@ -33,6 +33,9 @@ public final class StandardArgumentDefinitions {
     public static final String ADD_OUTPUT_SAM_PROGRAM_RECORD = "addOutputSAMProgramRecord";
     public static final String ADD_OUTPUT_VCF_COMMANDLINE = "addOutputVCFCommandLine";
     public static final String SEQUENCE_DICTIONARY_NAME = "sequenceDictionary";
+    public static final String ANNOTATION_LONG_NAME = "annotation";
+    public static final String ANNOTATION_GROUP_LONG_NAME = "annotationGroup";
+    public static final String ANNOTATIONS_TO_EXCLUDE_LONG_NAME = "annotationsToExclude";
 
     public static final String INPUT_SHORT_NAME = "I";
     public static final String OUTPUT_SHORT_NAME = "O";
@@ -65,6 +68,9 @@ public final class StandardArgumentDefinitions {
     public static final String CLOUD_PREFETCH_BUFFER_SHORT_NAME = "CPB";
     public static final String CLOUD_INDEX_PREFETCH_BUFFER_SHORT_NAME = "CIPB";
     public static final String DISABLE_BAM_INDEX_CACHING_SHORT_NAME = "DBIC";
+    public static final String ANNOTATION_SHORT_NAME = "A";
+    public static final String ANNOTATION_GROUP_SHORT_NAME = "G";
+    public static final String ANNOTATIONS_TO_EXCLUDE_SHORT_NAME = "AX";
 
     public static final String SPARK_PROPERTY_NAME = "conf";
 
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
@@ -1,0 +1,57 @@
+package org.broadinstitute.hellbender.cmdline.argumentcollections;
+
+import org.broadinstitute.barclay.argparser.Advanced;
+import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+import org.broadinstitute.hellbender.utils.Utils;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Arguments for requesting VariantContext annotations to be processed by {@link org.broadinstitute.hellbender.tools.walkers.annotator.VariantAnnotatorEngine}
+ * for tools that process variants objects.
+ */
+public class VariantAnnotationArgumentCollection implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Argument collection constructor which defines tool default annotation arguments when they are not overridden by the user
+     *
+     * @param defaultGroups                      List annotation group names to be used by default
+     * @param defaultAnnotations                 List of annotation class names to be used by default
+     * @param defaultAnnotationsToExclude        List of annotation class names to exclude by default. These override the default annotations and annotation groups.
+     */
+    public VariantAnnotationArgumentCollection(List<String> defaultGroups, List<String> defaultAnnotations, List<String> defaultAnnotationsToExclude) {
+        Utils.nonNull(defaultGroups);
+        Utils.nonNull(defaultAnnotations);
+        Utils.nonNull(defaultAnnotationsToExclude);
+
+        annotationGroupsToUse = new ArrayList<>(defaultGroups);
+        annotationsToUse = new ArrayList<>(defaultAnnotations);
+        annotationsToExclude = new ArrayList<>(defaultAnnotationsToExclude);
+    }
+
+    /**
+     * Which annotations to include in variant calls in the output. These supplement annotations provided by annotation groups.
+     */
+    @Argument(fullName=StandardArgumentDefinitions.ANNOTATION_LONG_NAME, shortName=StandardArgumentDefinitions.ANNOTATION_SHORT_NAME, doc="One or more specific annotations to add to variant calls", optional=true)
+    public List<String> annotationsToUse = new ArrayList<>();
+
+    /**
+     * Which annotations to exclude from output in the variant calls.  Note that this argument has higher priority than the
+     * -A or -G arguments, so these annotations will be excluded even if they are explicitly included with the other
+     * options.
+     */
+    @Argument(fullName=StandardArgumentDefinitions.ANNOTATIONS_TO_EXCLUDE_LONG_NAME, shortName=StandardArgumentDefinitions.ANNOTATIONS_TO_EXCLUDE_SHORT_NAME, doc="One or more specific annotations to exclude from variant calls", optional=true)
+    public List<String> annotationsToExclude = new ArrayList<>();
+
+    /**
+     * Which groups of annotations to add to the output variant calls.
+     * Any requirements that are not met (e.g. failing to provide a pedigree file for a pedigree-based annotation) may cause the run to fail.
+     */
+    @Argument(fullName= StandardArgumentDefinitions.ANNOTATION_GROUP_LONG_NAME, shortName=StandardArgumentDefinitions.ANNOTATION_GROUP_SHORT_NAME, doc="One or more groups of annotations to apply to variant calls", optional=true)
+    public List<String> annotationGroupsToUse = new ArrayList<>();
+
+}
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandArtifact.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandArtifact.java
@@ -21,7 +21,7 @@ import static org.broadinstitute.hellbender.tools.walkers.annotator.StrandArtifa
  *
  * Created by tsato on 4/19/17.
  */
-public class StrandArtifact extends GenotypeAnnotation implements StandardSomaticAnnotation {
+public class StrandArtifact extends GenotypeAnnotation implements StandardMutectAnnotation {
     public static final String POSTERIOR_PROBABILITIES_KEY = "SA_POST_PROB";
     public static final String MAP_ALLELE_FRACTIONS_KEY = "SA_MAP_AF";
 
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
@@ -4,6 +4,7 @@ import com.google.common.collect.Sets;
 import htsjdk.variant.variantcontext.*;
 import htsjdk.variant.vcf.*;
 import org.broadinstitute.barclay.argparser.CommandLineException;
+import org.broadinstitute.hellbender.cmdline.argumentcollections.VariantAnnotationArgumentCollection;
 import org.broadinstitute.hellbender.engine.FeatureContext;
 import org.broadinstitute.hellbender.engine.FeatureInput;
 import org.broadinstitute.hellbender.engine.ReferenceContext;
@@ -79,6 +80,26 @@ public final class VariantAnnotatorEngine {
         return new VariantAnnotatorEngine(AnnotationManager.ofSelectedMinusExcluded(annotationGroupsToUse, annotationsToUse, annotationsToExclude), dbSNPInput, comparisonFeatureInputs);
     }
 
+    /**
+     * An overload of {@link org.broadinstitute.hellbender.tools.walkers.annotator.VariantAnnotatorEngine#ofSelectedMinusExcluded ofSelectedMinusExcluded}
+     * except that it accepts a {@link org.broadinstitute.hellbender.cmdline.argumentcollections.VariantAnnotationArgumentCollection} as input.
+     * @param argumentCollection            VariantAnnotationArgumentCollection containing requested annotations.
+     * @param dbSNPInput                    input for variants from a known set from DbSNP or null if not provided.
+     *                   The annotation engine will mark variants overlapping anything in this set using {@link htsjdk.variant.vcf.VCFConstants#DBSNP_KEY}.
+     * @param comparisonFeatureInputs list of inputs with known variants.
+     *                   The annotation engine will mark variants overlapping anything in those sets using the name given by {@link FeatureInput#getName()}.
+     *                   Note: the DBSNP FeatureInput should be passed in separately, and not as part of this List - an GATKException will be thrown otherwise.
+     *                   Note: there are no non-DBSNP comparison FeatureInputs an empty List should be passed in here, rather than null.
+     * @return a VariantAnnotatorEngine initialized with the requested annotations
+     */
+    public static VariantAnnotatorEngine ofSelectedMinusExcluded(final VariantAnnotationArgumentCollection argumentCollection,
+                                                                 final FeatureInput<VariantContext> dbSNPInput,
+                                                                 final List<FeatureInput<VariantContext>> comparisonFeatureInputs) {
+        return ofSelectedMinusExcluded(argumentCollection.annotationGroupsToUse,
+                argumentCollection.annotationsToUse,
+                argumentCollection.annotationsToExclude,
+                dbSNPInput, comparisonFeatureInputs);
+    }
     private VariantOverlapAnnotator initializeOverlapAnnotator(final FeatureInput<VariantContext> dbSNPInput, final List<FeatureInput<VariantContext>> featureInputs) {
         final Map<FeatureInput<VariantContext>, String> overlaps = new LinkedHashMap<>();
         for ( final FeatureInput<VariantContext> fi : featureInputs) {
@@ -93,6 +114,7 @@ public final class VariantAnnotatorEngine {
 
         return new VariantOverlapAnnotator(dbSNPInput, overlaps);
     }
+
 
     /**
      * Returns the list of genotype annotations that will be applied.
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
@@ -1,15 +1,15 @@
 package org.broadinstitute.hellbender.tools.walkers.mutect;
 
 import htsjdk.variant.variantcontext.VariantContext;
-import org.broadinstitute.barclay.argparser.Advanced;
 import org.broadinstitute.barclay.argparser.Argument;
-import org.broadinstitute.barclay.argparser.Hidden;
+import org.broadinstitute.barclay.argparser.ArgumentCollection;
+import org.broadinstitute.hellbender.cmdline.argumentcollections.VariantAnnotationArgumentCollection;
 import org.broadinstitute.hellbender.engine.FeatureInput;
+import org.broadinstitute.hellbender.tools.walkers.annotator.StandardMutectAnnotation;
 import org.broadinstitute.hellbender.tools.walkers.haplotypecaller.AssemblyBasedCallerArgumentCollection;
 
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.List;
+import java.util.Collections;
 
 public class M2ArgumentCollection extends AssemblyBasedCallerArgumentCollection {
     private static final long serialVersionUID = 9341L;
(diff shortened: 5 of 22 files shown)


src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final long serialVersionUID = 1L;
+
+    public VariantAnnotationArgumentCollection(List<String> defaultGroups, List<String> defaultAnnotations, List<String> defaultExclude) {
+        annotationGroupsToUse = new ArrayList<>(defaultGroups);
+        annotationsToUse = new ArrayList<>(defaultAnnotations);
+        annotationsToExclude = new ArrayList<>(defaultExclude);
+    }
+
+    /**
+     * Which annotations to recompute for the combined output VCF file.
```
davidbenjamin: Why *re*-compute?

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,49 @@
...
+    public List<String> annotationsToExclude = new ArrayList<>();
+
+    /**
+     * Which groups of annotations to add to the output VCF file. The single value 'none' removes the default group. See
+     * the VariantAnnotator -list argument to view available groups. Note that this usage is not recommended because
+     * it obscures the specific requirements of individual annotations. Any requirements that are not met (e.g. failing
+     * to provide a pedigree file for a pedigree-based annotation) may cause the run to fail.
+     */
+    @Advanced
+    @Argument(fullName="group", shortName="G", doc="One or more classes/groups of annotations to apply to variant calls", optional=true)
```
davidbenjamin: How about "classes/groups" -> "groups"?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandArtifact.java
```diff
@@ -21,7 +21,7 @@
  *
  * Created by tsato on 4/19/17.
  */
-public class StrandArtifact extends GenotypeAnnotation implements StandardSomaticAnnotation {
+public class StrandArtifact extends GenotypeAnnotation implements StandardSomaticAnnotation, StandardM2Annotation {
```
davidbenjamin: You can delete `StandardSomaticAnnotation` entirely and absorb implementing classes into `StandardM2Annotation`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StandardM2Annotation.java
```diff
@@ -0,0 +1,7 @@
+package org.broadinstitute.hellbender.tools.walkers.annotator;
+
+/**
+ * This is a marker interface used to indicate which annotations are "Standard" for Mutect2 only.
+ */
+public interface StandardM2Annotation extends Annotation {
```
davidbenjamin: Could you call it `StandardMutectAnnotation`?

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,49 @@
...
+    public VariantAnnotationArgumentCollection(List<String> defaultGroups, List<String> defaultAnnotations, List<String> defaultExclude) {
+        annotationGroupsToUse = new ArrayList<>(defaultGroups);
+        annotationsToUse = new ArrayList<>(defaultAnnotations);
+        annotationsToExclude = new ArrayList<>(defaultExclude);
+    }
+
+    /**
+     * Which annotations to recompute for the combined output VCF file.
+     */
+    @Advanced
```
droazen: Should this really be marked as `@Advanced`? It doesn't strike me as an advanced argument. What do you think @davidbenjamin ?

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,49 @@
...
+import org.broadinstitute.barclay.argparser.Argument;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+public class VariantAnnotationArgumentCollection implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    public VariantAnnotationArgumentCollection(List<String> defaultGroups, List<String> defaultAnnotations, List<String> defaultExclude) {
```
droazen: `defaultExclude` -> `defaultAnnotationsToExclude` (otherwise could be misinterpreted)

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
```diff
@@ -93,6 +94,14 @@ private VariantOverlapAnnotator initializeOverlapAnnotator(final FeatureInput<Va
 
         return new VariantOverlapAnnotator(dbSNPInput, overlaps);
     }
+    public static VariantAnnotatorEngine ofSelectedMinusExcluded(final VariantAnnotationArgumentCollection argumentColleciton,
```
droazen: Add javadoc for this public method, and move it next to the other overload of `ofSelectedMinusExcluded()` in the source file.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -100,24 +105,17 @@
...
-            "TandemRepeat", "OxoGReadCounts", "ReadPosition", "BaseQuality", "MappingQuality",
-            "FragmentLength", "StrandArtifact"}));
 
     /**
-     * Which groups of annotations to add to the output VCF file. The single value 'none' removes the default group.
-     * Note that this usage is not recommended because it obscures the specific requirements of individual annotations.
+     * Set of annotation arguments to use.
      * Any requirements that are not met, e.g. failing to provide a pedigree file for a pedigree-based annotation, may cause the run to fail.
      * For somatic analyses, the StandardSomaticAnnotation group currently contains one annotations: StrandArtifact.
      * Note the latter is redundant to an annotation given by the --annotation argument default.
```
droazen: Remove mention of `StandardSomaticAnnotation`, if it's going to be deleted.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
davidbenjamin: I agree that this isn't advanced.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandArtifact.java
jamesemery: @davidbenjamin The only issue would seem to be that the UniqueAltReadCount appears to be a standard somatic annotation yet not included by default for M2, i'll add it to the standard M2 Annotations for now.

droazen: I wouldn't do that @jamesemery -- we don't want to be changing the default Mutect annotations as part of this PR

src/main/java/org/broadinstitute/hellbender/cmdline/StandardArgumentDefinitions.java
```diff
@@ -33,6 +33,9 @@ private StandardArgumentDefinitions(){}
     public static final String ADD_OUTPUT_SAM_PROGRAM_RECORD = "addOutputSAMProgramRecord";
     public static final String ADD_OUTPUT_VCF_COMMANDLINE = "addOutputVCFCommandLine";
     public static final String SEQUENCE_DICTIONARY_NAME = "sequenceDictionary";
+    public static final String ANNOTATION_LONG_NAME = "annotation";
+    public static final String ANNOTATION_GROUP_LONG_NAME = "group";
```
droazen: I think this should really be named `annotationGroup`, regardless of what GATK3.x had (though check the WDLs for usage of the old name)

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,57 @@
...
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+import org.broadinstitute.hellbender.utils.Utils;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Arguments for requesting VariantContext annotations to be processed by {@link org.broadinstitute.hellbender.tools.walkers.annotator.VariantAnnotatorEngine}
+ * for tools that process VariantContext objects.
```
droazen: "for tools that process variants"

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class VariantAnnotationArgumentCollection implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Argument collection constructor which defines tool default annotation arguments when they are not overridden by the user
+     *
+     * @param defaultGroups                      List annotation groups names to be used by default
```
droazen: "annotation groups names" -> "annotation group names"

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param defaultAnnotations                 List of annotation class names to be used by default
+     * @param defaultAnnotationsToExclude        List of default annotation classes to exclude from output, these override the previous two fields
```
droazen: "List of annotation class names to exclude by default. These override the default annotations and annotation groups."

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,57 @@
...
+
+        annotationGroupsToUse = new ArrayList<>(defaultGroups);
+        annotationsToUse = new ArrayList<>(defaultAnnotations);
+        annotationsToExclude = new ArrayList<>(defaultAnnotationsToExclude);
+    }
+
+    /**
+     * Which annotations to include in variant calls in the output.
+     */
+    @Argument(fullName=StandardArgumentDefinitions.ANNOTATION_LONG_NAME, shortName=StandardArgumentDefinitions.ANNOTATION_SHORT_NAME, doc="One or more specific annotations add to variant calls", optional=true)
```
droazen: "specific annotations add" -> "specific annotations to add"

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Which annotations to exclude from output in the variant calls.  Note that this argument has higher priority than the
+     * -A or -G arguments, so these annotations will be excluded even if they are explicitly included with the other
+     * options.
+     */
+    @Argument(fullName=StandardArgumentDefinitions.ANNOTATIONS_TO_EXCLUDE_LONG_NAME, shortName=StandardArgumentDefinitions.ANNOTATIONS_TO_EXCLUDE_SHORT_NAME, doc="One or more specific annotations to exclude from variant calls", optional=true)
+    public List<String> annotationsToExclude = new ArrayList<>();
+
+    /**
+     * Which groups of annotations to add to the output VCF file.
```
droazen: "output VCF file" -> "output variant calls"

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandArtifact.java
davidbenjamin: Regardless of that, `StandardSomaticAnnotation` doesn't seem to be very useful.  As far as I'm concerned you can delete it and leave `UniqueAltReadCount` without any extra annotation.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,57 @@
...
+        Utils.nonNull(defaultAnnotations);
+        Utils.nonNull(defaultAnnotationsToExclude);
+
+        annotationGroupsToUse = new ArrayList<>(defaultGroups);
+        annotationsToUse = new ArrayList<>(defaultAnnotations);
+        annotationsToExclude = new ArrayList<>(defaultAnnotationsToExclude);
+    }
+
+    /**
+     * Which annotations to include in variant calls in the output.
```
droazen: Add note that these supplement the annotations in the annotation groups specified via -G

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StandardMutectAnnotations.java
```diff
@@ -1,7 +1,7 @@
 package org.broadinstitute.hellbender.tools.walkers.annotator;
 
 /**
- * This is a marker interface used to indicate which annotations are "Standard" for MuTect only.
+ * This is a marker interface used to indicate which annotations are "Standard" for Mutect2 only.
  */
-public interface StandardSomaticAnnotation {
+public interface StandardMutectAnnotations extends Annotation {
```
droazen: Rename to `StandardMutectAnnotation` (without the plural), for grammatical consistency with the other annotation groups.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
```diff
@@ -79,6 +80,26 @@ public static VariantAnnotatorEngine ofSelectedMinusExcluded(final List<String>
...
+     * except that it accepts a {@link org.broadinstitute.hellbender.cmdline.argumentcollections.VariantAnnotationArgumentCollection} as input.
+     * @param argumentCollection            VariantAnnotationArgumentCollection containing requested annotations.
+     * @param dbSNPInput                    input for variants from a known set from DbSNP or null if not provided.
+     *                   The annotation engine will mark variants overlapping anything in this set using {@link htsjdk.variant.vcf.VCFConstants#DBSNP_KEY}.
+     * @param comparisonFeatureInputs list of inputs with known variants.
+     *                   The annotation engine will mark variants overlapping anything in those sets using the name given by {@link FeatureInput#getName()}.
+     *                   Note: the DBSNP FeatureInput should be passed in separately, and not as part of this List - an GATKException will be thrown otherwise.
+     *                   Note: there are no non-DBSNP comparison FeatureInputs an empty List should be passed in here, rather than null.
+     * @return
```
droazen: `@return a VariantAnnotatorEngine initialized with the requested annotations`

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -100,24 +100,16 @@
...
-            "TandemRepeat", "OxoGReadCounts", "ReadPosition", "BaseQuality", "MappingQuality",
-            "FragmentLength", "StrandArtifact"}));
 
     /**
-     * Which groups of annotations to add to the output VCF file. The single value 'none' removes the default group.
-     * Note that this usage is not recommended because it obscures the specific requirements of individual annotations.
+     * Set of annotation arguments to use.
      * Any requirements that are not met, e.g. failing to provide a pedigree file for a pedigree-based annotation, may cause the run to fail.
-     * For somatic analyses, the StandardSomaticAnnotation group currently contains one annotations: StrandArtifact.
      * Note the latter is redundant to an annotation given by the --annotation argument default.
```
droazen: This comment no longer makes sense: "Note the latter is redundant to an annotation given by the --annotation argument default"

src/main/java/org/broadinstitute/hellbender/cmdline/StandardArgumentDefinitions.java
jamesemery: Doesn't seem to be in any WDLs here or in dsde-pipelines

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StandardMutectAnnotations.java
jamesemery: done

droazen:
Note that this PR also closes https://github.com/broadinstitute/gatk/issues/3123

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/VariantAnnotationArgumentCollection.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Arguments for requesting VariantContext annotations to be processed by {@link org.broadinstitute.hellbender.tools.walkers.annotator.VariantAnnotatorEngine}
+ * for tools that process variants objects.
+ */
+public class VariantAnnotationArgumentCollection implements Serializable {
```
magicDGS: Please, consider to make this an abstract class to being able to hide some parameters from the command line. Then, a `DefaultGATKVariantAnnotationArgumentCollection` should be used in GATK with this implementation.

If the variant anotation engine is converted into a plugin, that will be useful for downstream projects: I was actually working on my own plugin, but I prefer to work with standards from the GATK framework, and if there are downstream projects that want to annotate variants with their own annotations, this will be required for example if they do not need groups to annotate together.

As an example, that will be similar to what I did for the `GATKReadFilterPluginDescriptor`  in https://github.com/broadinstitute/gatk/pull/2355. As an API user, I think that it is good to have all the arguments for the plugins as abstract classes to being able to hide some from the CLI...

jamesemery: @magicDGS This would be a useful feature but it sounds like the benefit is mostly hidden behind pending work on converting annotations into a plugin, which we have planned for the 4.0 release. This change would probably be incorporated as part of that branch.

magicDGS: I thought that this could be addressed here for using the `VariantAnnotationArgumentCollection` with the `VariantAnnotator` too, even if the plugin is not there. I cannot hide some parameters with the current implementation, to being able to pass the argument collection to the annotator...

Anyway, I can submit a patch if there is no planned timeplan for the plugin, to being able to use something similar in my code...

droazen: Maybe it would be easier to just add a generic/universal way to hide arguments in barclay, instead of having to complicate all of our argument collections with added abstract classes?

magicDGS: I think that the main reason of having argument collections is for tweak the arguments for 
make optional/required some arguments (such as `ReferenceInputArgumentCollection`) or have the ability to tweak some arguments to be hidden/removed from the user.

I believe that with the current system for parsing arguments and argument collections is going to be complicated to hide arguments. In addition, sometimes the requirement isn't hide the argument but provide a tool-specific documentation (e.g., the `ReadFilter` plugin specifies that are filters __before__ the analysis; in some cases, I have tools which perform the filtering __after__ processing, and it is important for the user to know that difference).

I think that having interfaces for a framework is the best way to encourage users with specific needs to use it, unless you really want to forbid the usage/tweak of some components. I see several use cases for having the abstract argument collections....

sooheelee:
Thanks for this.

PR #3628, opened 2017-09-27
Imprecise deletion evidence annotation and variant calling

cwhelan:
This PR adds logic to use imprecise evidence (gathered from discordant read pair mappings and split reads with supplementary alignments) to:

- Annotate events that we are calling from assembled breakpoints with the number of discordant read pairs and split reads that support those events.
- Call new variants when an imprecise evidence link supporting a deletion is supported by a minimum number of pieces of evidence (set to 7 by default in this PR; parameters can be tuned later).

In an initial run on the CHM mix this uses 3500 evidence-target clusters to annotate our existing deletion calls and makes an additional 670 IMPRECISE deletion calls. Initial visual evaluations against the CHM truth set looks good on visual inspection; more detailed evaluations will follow.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
@@ -19,19 +19,19 @@ import org.broadinstitute.hellbender.tools.spark.sv.discovery.*;
 import org.broadinstitute.hellbender.tools.spark.sv.evidence.AlignedAssemblyOrExcuse;
 import org.broadinstitute.hellbender.tools.spark.sv.evidence.EvidenceTargetLink;
 import org.broadinstitute.hellbender.tools.spark.sv.evidence.FindBreakpointEvidenceSpark;
+import org.broadinstitute.hellbender.tools.spark.sv.utils.PairedStrandedIntervalTree;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.bwa.BwaMemAlignment;
 import org.broadinstitute.hellbender.utils.bwa.BwaMemAlignmentUtils;
 import org.broadinstitute.hellbender.utils.fermi.FermiLiteAssembly;
 import scala.Serializable;
-import scala.Tuple2;
 
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
-import static org.broadinstitute.hellbender.tools.spark.sv.StructuralVariationDiscoveryArgumentCollection.FindBreakpointEvidenceSparkArgumentCollection;
 import static org.broadinstitute.hellbender.tools.spark.sv.StructuralVariationDiscoveryArgumentCollection.DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection;
+import static org.broadinstitute.hellbender.tools.spark.sv.StructuralVariationDiscoveryArgumentCollection.FindBreakpointEvidenceSparkArgumentCollection;
 
 /**
  * Tool to run the sv pipeline up to and including variant discovery
@@ -76,23 +76,49 @@ public class StructuralVariationDiscoveryPipelineSpark extends GATKSparkTool {
         final PipelineOptions pipelineOptions = getAuthenticatedGCSOptions();
 
         // gather evidence, run assembly, and align
-        final Tuple2<List<AlignedAssemblyOrExcuse>, List<EvidenceTargetLink>> alignedAssemblyOrExcuseAndLinkList =
+        final FindBreakpointEvidenceSpark.AssembledEvidenceResults assembledEvidenceResults =
                 FindBreakpointEvidenceSpark
                         .gatherEvidenceAndWriteContigSamFile(ctx, evidenceAndAssemblyArgs, header, getUnfilteredReads(),
                                 outputAssemblyAlignments, localLogger);
         // todo: when we call imprecise variants don't return here
-        if (alignedAssemblyOrExcuseAndLinkList._1().isEmpty()) return;
+        if (assembledEvidenceResults.getAlignedAssemblyOrExcuseList().isEmpty()) return;
 
         // parse the contig alignments and extract necessary information
         final JavaRDD<AlignedContig> parsedAlignments =
-                new InMemoryAlignmentParser(ctx, alignedAssemblyOrExcuseAndLinkList._1(), header, localLogger).getAlignedContigs();
+                new InMemoryAlignmentParser(ctx, assembledEvidenceResults.getAlignedAssemblyOrExcuseList(), header, localLogger).getAlignedContigs();
         // todo: when we call imprecise variants don't return here
         if(parsedAlignments.isEmpty()) return;
 
+        final List<EvidenceTargetLink> evidenceTargetLinks = assembledEvidenceResults.getEvidenceTargetLinks();
+        final PairedStrandedIntervalTree<EvidenceTargetLink> evidenceLinkTree = makeEvidenceLinkTree(evidenceTargetLinks);
+
         // discover variants and write to vcf
         DiscoverVariantsFromContigAlignmentsSAMSpark
-                .discoverVariantsAndWriteVCF(parsedAlignments, discoverStageArgs.fastaReference,
-                        ctx.broadcast(getReference()), pipelineOptions, vcfOutputFileName, localLogger);
+                .discoverVariantsAndWriteVCF(
+                        parsedAlignments,
+                        discoverStageArgs,
+                        ctx.broadcast(getReference()),
+                        vcfOutputFileName,
+                        localLogger,
+                        evidenceLinkTree,
+                        assembledEvidenceResults.getReadMetadata(),
+                        header.getSequenceDictionary());
+    }
+
+    /**
+     * Makes a PairedStrandedIntervalTree from a list of EvidenceTargetLinks. The value of each entry in the resulting tree
+     * will be the original EvidenceTargetLink. If the input list is null, returns a null tree.
+     */
+    private PairedStrandedIntervalTree<EvidenceTargetLink> makeEvidenceLinkTree(final List<EvidenceTargetLink> evidenceTargetLinks) {
+        final PairedStrandedIntervalTree<EvidenceTargetLink> evidenceLinkTree;
+
+        if (evidenceTargetLinks != null) {
+            evidenceLinkTree = new PairedStrandedIntervalTree<>();
+            evidenceTargetLinks.forEach(l -> evidenceLinkTree.put(l.getPairedStrandedIntervals(), l));
+        } else {
+            evidenceLinkTree = null;
+        }
+        return evidenceLinkTree;
     }
 
     public static final class InMemoryAlignmentParser extends AlignedContigGenerator implements Serializable {
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AnnotatedVariantProducer.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AnnotatedVariantProducer.java
@@ -1,17 +1,23 @@
 package org.broadinstitute.hellbender.tools.spark.sv.discovery;
 
 import com.google.common.annotations.VisibleForTesting;
+import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.variant.variantcontext.Allele;
+import htsjdk.variant.variantcontext.StructuralVariantType;
 import htsjdk.variant.variantcontext.VariantContext;
 import htsjdk.variant.variantcontext.VariantContextBuilder;
 import htsjdk.variant.vcf.VCFConstants;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.spark.broadcast.Broadcast;
 import org.broadinstitute.hellbender.engine.datasources.ReferenceMultiSource;
+import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.tools.spark.sv.StructuralVariationDiscoveryArgumentCollection;
-import org.broadinstitute.hellbender.tools.spark.sv.utils.GATKSVVCFConstants;
+import org.broadinstitute.hellbender.tools.spark.sv.evidence.EvidenceTargetLink;
+import org.broadinstitute.hellbender.tools.spark.sv.evidence.ReadMetadata;
+import org.broadinstitute.hellbender.tools.spark.sv.utils.*;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.Utils;
+import scala.Tuple2;
 
 import java.io.IOException;
 import java.io.Serializable;
@@ -105,6 +111,45 @@ public class AnnotatedVariantProducer implements Serializable {
         return vcBuilder.make();
     }
 
+    public static VariantContext produceAnnotatedVcFromEvidenceTargetLink(final EvidenceTargetLink e,
+                                                                          final SvType svType,
+                                                                          final SAMSequenceDictionary sequenceDictionary,
+                                                                          final ReferenceMultiSource reference) {
+        final String sequenceName = sequenceDictionary.getSequence(e.getPairedStrandedIntervals().getLeft().getInterval().getContig()).getSequenceName();
+        final int start = e.getPairedStrandedIntervals().getLeft().getInterval().midpoint();
+        final int end = e.getPairedStrandedIntervals().getRight().getInterval().midpoint();
+        try {
+            final VariantContextBuilder builder = new VariantContextBuilder()
+                    .chr(sequenceName)
+                    .start(start)
+                    .stop(end)
+                    .id(svType.variantId)
+                    .alleles(produceAlleles(new SimpleInterval(sequenceName, start, start), reference, svType))
+                    .attribute(VCFConstants.END_KEY, end)
+                    .attribute(GATKSVVCFConstants.SVLEN, svType.getSVLength())
+                    .attribute(GATKSVVCFConstants.SVTYPE, svType.toString())
+                    .attribute(GATKSVVCFConstants.IMPRECISE, true)
+                    .attribute(GATKSVVCFConstants.CIPOS, produceCIInterval(start, e.getPairedStrandedIntervals().getLeft().getInterval()))
+                    .attribute(GATKSVVCFConstants.CIEND, produceCIInterval(end, e.getPairedStrandedIntervals().getRight().getInterval()))
+                    .attribute(GATKSVVCFConstants.READ_PAIR_SUPPORT, e.getReadPairs())
+                    .attribute(GATKSVVCFConstants.SPLIT_READ_SUPPORT, e.getSplitReads());
+            return builder.make();
+        } catch (IOException e1) {
+            throw new GATKException("error reading reference base for variant context " + svType.variantId, e1);
+        }
+    }
+
+    /**
+     * Produces the string representation of a VCF 4.2-style SV CI interval centered around 'point'.
+     */
+    @VisibleForTesting
+    static String produceCIInterval(final int point, final SVInterval ciInterval) {
+        Utils.validate(ciInterval.getStart() <= point && ciInterval.getEnd() >= point, "Interval must contain point");
+        return String.join(",",
+                String.valueOf(ciInterval.getStart() - point),
+                String.valueOf(ciInterval.getEnd() - point));
+    }
+
     // TODO: 12/13/16 again ignoring translocation
     @VisibleForTesting
     static List<Allele> produceAlleles(final SimpleInterval refLoc,
@@ -147,6 +192,39 @@ public class AnnotatedVariantProducer implements Serializable {
             }
         }
         return attributeMap;
+    }
+
+    static VariantContext annotateWithImpreciseEvidenceLinks(final VariantContext variant,
+                                                             final PairedStrandedIntervalTree<EvidenceTargetLink> evidenceTargetLinks,
+                                                             final SAMSequenceDictionary referenceSequenceDictionary,
+                                                             final ReadMetadata metadata,
+                                                             final int defaultUncertainty) {
+        if (variant.getStructuralVariantType() == StructuralVariantType.DEL) {
+            SVContext svc = SVContext.of(variant);
+            final int padding = (metadata == null) ? defaultUncertainty : (metadata.getMaxMedianFragmentSize() / 2);
+            PairedStrandedIntervals svcIntervals = svc.getPairedStrandedIntervals(referenceSequenceDictionary, padding);
+
+            final Iterator<Tuple2<PairedStrandedIntervals, EvidenceTargetLink>> overlappers = evidenceTargetLinks.overlappers(svcIntervals);
+            int readPairs = 0;
+            int splitReads = 0;
+            while (overlappers.hasNext()) {
+                final Tuple2<PairedStrandedIntervals, EvidenceTargetLink> next = overlappers.next();
+                readPairs += next._2.getReadPairs();
+                splitReads += next._2.getSplitReads();
+                overlappers.remove();
+            }
+            final VariantContextBuilder variantContextBuilder = new VariantContextBuilder(variant);
+            if (readPairs > 0) {
+                variantContextBuilder.attribute(GATKSVVCFConstants.READ_PAIR_SUPPORT, readPairs);
+            }
+            if (splitReads > 0) {
+                variantContextBuilder.attribute(GATKSVVCFConstants.SPLIT_READ_SUPPORT, splitReads);
+            }
+
+            return variantContextBuilder.make();
+        } else {
+            return variant;
+        }
     }
 
     /**
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
@@ -1,10 +1,10 @@
 package org.broadinstitute.hellbender.tools.spark.sv.discovery;
 
-import com.google.cloud.dataflow.sdk.options.PipelineOptions;
 import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.CigarOperator;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.util.SequenceUtil;
 import htsjdk.variant.variantcontext.VariantContext;
 import org.apache.logging.log4j.LogManager;
@@ -24,8 +24,9 @@ import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.spark.GATKSparkTool;
 import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.tools.spark.sv.StructuralVariationDiscoveryArgumentCollection;
-import org.broadinstitute.hellbender.tools.spark.sv.utils.GATKSVVCFHeaderLines;
-import org.broadinstitute.hellbender.tools.spark.sv.utils.SVVCFWriter;
+import org.broadinstitute.hellbender.tools.spark.sv.evidence.EvidenceTargetLink;
+import org.broadinstitute.hellbender.tools.spark.sv.evidence.ReadMetadata;
+import org.broadinstitute.hellbender.tools.spark.sv.utils.*;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
 import scala.Tuple2;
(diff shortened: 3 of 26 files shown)


cwhelan:
@SHuang-Broad would you like to review this one?

SHuang-Broad:
Will be on it tomorrow.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
```diff
@@ -167,21 +180,119 @@ public static AlignedContig parseReadsAndOptionallySplitGappedAlignments(final I
...
+            final Iterator<Tuple2<PairedStrandedIntervals, EvidenceTargetLink>> overlappers = evidenceTargetLinks.overlappers(svcIntervals);
+            int readPairs = 0;
+            int splitReads = 0;
+            while (overlappers.hasNext()) {
+                final Tuple2<PairedStrandedIntervals, EvidenceTargetLink> next = overlappers.next();
+                readPairs += next._2.getReadPairs();
+                splitReads += next._2.getSplitReads();
+                overlappers.remove();
+            }
+            if (readPairs == 0 && splitReads == 0) {
```
SHuang-Broad: It feels more natural to me that these two annotations should exist for all (DEL for now) variants.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
```diff
@@ -167,21 +180,119 @@ public static AlignedContig parseReadsAndOptionallySplitGappedAlignments(final I
...
+            PairedStrandedIntervals svcIntervals = svc.getPairedStrandedIntervals(referenceSequenceDictionary, padding);
+
+            final Iterator<Tuple2<PairedStrandedIntervals, EvidenceTargetLink>> overlappers = evidenceTargetLinks.overlappers(svcIntervals);
+            int readPairs = 0;
+            int splitReads = 0;
+            while (overlappers.hasNext()) {
+                final Tuple2<PairedStrandedIntervals, EvidenceTargetLink> next = overlappers.next();
+                readPairs += next._2.getReadPairs();
+                splitReads += next._2.getSplitReads();
+                overlappers.remove();
```
SHuang-Broad: is this `remove()` necessary because you don't want to count an evidence twice for difference DEL records? Just to see if I'm understanding correctly.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSparkUnitTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+    private final Logger localLogger = LogManager.getLogger(DiscoverVariantsFromContigAlignmentsSAMSparkUnitTest.class);
+    private static String twoBitRefURL = publicTestDir + "large/human_g1k_v37.20.21.2bit";
+
+    @DataProvider(name = "evidenceTargetLinksAndVariants")
+    public Object[][] getEvidenceTargetLinksAndVariants() {
+        final VariantContext unAnnotatedVC = new VariantContextBuilder()
+                .id("TESTID")
+                .chr("20").start(200).stop(300)
+                .alleles("A", "<" + GATKSVVCFConstants.SYMB_ALT_ALLELE_DEL_IN_HEADER + ">")
```
SHuang-Broad: there's a `SimpleSVType.createBracketedSymbAlleleString()`

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSparkUnitTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+        PairedStrandedIntervalTree<EvidenceTargetLink> evidenceTree = new PairedStrandedIntervalTree<>();
+        etls.forEach(e -> evidenceTree.put(e.getPairedStrandedIntervals(), e));
+
+        final List<VariantContext> processedVariantContexts =
+                DiscoverVariantsFromContigAlignmentsSAMSpark.processEvidenceTargetLinks(params, localLogger, evidenceTree,
+                        metadata, inputVariants, referenceMultiSource);
+
+        Assert.assertEquals(processedVariantContexts.size(), expectedVariants.size());
+        for (int i = 0; i < processedVariantContexts.size(); i++) {
```
SHuang-Broad: Seems that there's the `VariantContextTestUtils.assertVariantContextsAreEqual()` for use if you have no attribute to ignore.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSparkUnitTest.java
```diff
@@ -0,0 +1,144 @@
...
+                                               final List<VariantContext> inputVariants,
+                                               final List<VariantContext> expectedVariants) throws Exception {
+        final StructuralVariationDiscoveryArgumentCollection.DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection params =
+                new StructuralVariationDiscoveryArgumentCollection.DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection();
+
+        final PipelineOptions options = null;
+        final ReferenceMultiSource referenceMultiSource = new ReferenceMultiSource(options, twoBitRefURL, ReferenceWindowFunctions.IDENTITY_FUNCTION);
+
+        ReadMetadata metadata = Mockito.mock(ReadMetadata.class);
+        when(metadata.getMaxMedianFragmentSize()).thenReturn(300);
```
SHuang-Broad: cool!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVContext.java
```diff
@@ -277,9 +280,12 @@ public int getStructuralVariantLength() {
                     composePaddedInterval(contigName, contigLength, start, start, padding));
         } else if (type == StructuralVariantType.DEL) {
             final int end = getEnd();
+            final int homologyPadding = (padForHomology ? getAttributeAsInt(GATKSVVCFConstants.HOMOLOGY_LENGTH, 0) : 0);
             return Arrays.asList(
-                    composePaddedInterval(contigName, contigLength, start, start, padding),
-                    composePaddedInterval(contigName, contigLength, end, end, padding));
+                    composePaddedInterval(contigName, contigLength, start + 1,
```
SHuang-Broad: I know this is probably my fault not asking this question before, but is there an off-by-1 error in `composePaddedInterval()`?  it seems to perform an asymmetrical padding to `start` and `end`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -209,7 +235,21 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+        return new EvidenceScanResults(readMetadata, intervals, intervalsAndEvidenceTargetLinks._2(), qNamesMultiMap);
+    }
+
+    static final class EvidenceScanResults {
+        final ReadMetadata readMetadata;
+        final List<SVInterval> intervals;
+        final List<EvidenceTargetLink> evidenceTargetLinks;
+        final HopscotchUniqueMultiMap<String, Integer, QNameAndInterval> qNamesForAssemblyMultiMap;
+
+        public EvidenceScanResults(final ReadMetadata readMetadata, final List<SVInterval> intervals, final List<EvidenceTargetLink> evidenceTargetLinks, final HopscotchUniqueMultiMap<String, Integer, QNameAndInterval> qNamesForAssemblyMultiMap) {
```
SHuang-Broad: break into multiple lines?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -76,23 +76,45 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+                        discoverStageArgs,
+                        ctx.broadcast(getReference()),
+                        vcfOutputFileName,
+                        localLogger,
+                        evidenceLinkTree,
+                        assembledEvidenceResults.getReadMetadata(),
+                        header.getSequenceDictionary());
+    }
+
+    private PairedStrandedIntervalTree<EvidenceTargetLink> makeEvidenceLinkTree(final List<EvidenceTargetLink> evidenceTargetLinks) {
```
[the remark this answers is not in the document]
cwhelan: Added documentation.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AnnotatedVariantProducer.java
```diff
@@ -105,6 +108,38 @@ static VariantContext produceAnnotatedVcFromInferredTypeAndRefLocations(final Si
...
+            final VariantContextBuilder builder = new VariantContextBuilder()
+                    .chr(sequenceName)
+                    .start(start)
+                    .stop(end)
+                    .id(svType.variantId)
+                    .alleles(produceAlleles(new SimpleInterval(sequenceName, start, start), reference, svType))
+                    .attribute(VCFConstants.END_KEY, end)
+                    .attribute(GATKSVVCFConstants.SVLEN, svType.getSVLength())
+                    .attribute(GATKSVVCFConstants.SVTYPE, svType.toString())
+                    .attribute(GATKSVVCFConstants.IMPRECISE, true)
```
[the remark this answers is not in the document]
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
```diff
@@ -66,6 +69,10 @@
             fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME)
     private String vcfOutputFileName;
 
+    private static boolean isImpreciseDeletion(Tuple2<PairedStrandedIntervals, EvidenceTargetLink> e) {
+        return e._1.getLeft().getInterval().getContig() == e._2().getPairedStrandedIntervals().getRight().getInterval().getContig() && e._1.getLeft().getStrand() && !e._1.getRight().getStrand();
```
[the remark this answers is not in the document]
cwhelan: Actually moved the method to `EvidenceTargetLink` since it seems to belong there. Added some documentation to `StrandedInterval`, which mostly just refers people to the definition of strand in `BreakpointEvidence`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
```diff
@@ -167,21 +180,119 @@ public static AlignedContig parseReadsAndOptionallySplitGappedAlignments(final I
...
+                    metadata,
+                    collectedAnnotatedVariants, broadcastReference.getValue());
+        }
+
+        SVVCFWriter.writeVCF(vcfOutputFileName, localLogger, collectedAnnotatedVariants,
+                sequenceDictionary);
+    }
+
+    @VisibleForTesting
+    static List<VariantContext> processEvidenceTargetLinks(final StructuralVariationDiscoveryArgumentCollection.DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection parameters,
```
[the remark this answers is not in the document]
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
```diff
@@ -167,21 +180,119 @@ public static AlignedContig parseReadsAndOptionallySplitGappedAlignments(final I
...
+                                                                   final Logger localLogger,
+                                                                   final PairedStrandedIntervalTree<EvidenceTargetLink> evidenceTargetLinks,
+                                                                   final ReadMetadata metadata,
+                                                                   List<VariantContext> collectedAnnotatedVariants,
```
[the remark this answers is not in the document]
cwhelan: Since we need to first compare the evidence links to the variants, and then the variants to the evidence target links, we'd either need to first broadcast one and then the other, or else come up with some sort of joining scheme. This seems much more efficient to just collect them all into local collections and compare on one machine.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
```diff
@@ -167,21 +180,119 @@ public static AlignedContig parseReadsAndOptionallySplitGappedAlignments(final I
...
+                                                                   final ReferenceMultiSource reference) {
+        final int originalEvidenceLinkSize = evidenceTargetLinks.size();
+        collectedAnnotatedVariants = collectedAnnotatedVariants
```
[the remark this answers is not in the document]
cwhelan: renamed several variables in this method

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
```diff
@@ -167,21 +180,119 @@ public static AlignedContig parseReadsAndOptionallySplitGappedAlignments(final I
...
+
+    private static VariantContext createImpreciseDeletionVariant(final EvidenceTargetLink e,
+                                                                 final SAMSequenceDictionary sequenceDictionary,
+                                                                 final ReferenceMultiSource reference) {
+        final SvType svType = new SimpleSVType.ImpreciseDeletion(e, sequenceDictionary);
+        return AnnotatedVariantProducer
+                .produceAnnotatedVcFromEvidenceTargetLink(e, svType, sequenceDictionary, reference);
+    }
+
+    private static VariantContext annotateWithImpreciseEvidenceLinks(final VariantContext variant,
```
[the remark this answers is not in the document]
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
cwhelan: If a high-scoring evidence link cluster matches an assembly variant, we just want to annotate the assembly variant with the number of read pairs and split reads, but not call a new IMPRECISE variant. So here the `EvidenceTargetLink` gets used up and removed from the collection when used in annotation. The remaining `EvidenceTargetLink` objects that weren't used for annotation will then be considered for IMPRECISE variant calling downstream. And yes, this will also cause `EvidenceTargetLinks` from being used to annotate more than one variant.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
cwhelan: OK. Removed this check. It feels like it might be more efficient to keep the check and maybe initialize all of the variants contexts to have these set to zero and not have to rebuild them at this step, but that seems less clean. What do you think?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
```diff
@@ -630,7 +630,8 @@ protected void serialize(final Kryo kryo, final Output output) {
 
         @Override
         public boolean hasDistalTargets(final ReadMetadata readMetadata, final int minEvidenceMapq) {
-            return getFragmentOrdinal() == TemplateFragmentOrdinal.PAIRED_FIRST && isTargetHighQuality(readMetadata, minEvidenceMapq);
+            return getLocation().isUpstreamOf(target)
```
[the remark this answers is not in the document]
cwhelan: renamed. 

The logic before was incorrect. This was meant to match the 'cannonicalization' of evidence target links that was suggested in review on the last PR, in which we attempt to only create a link for one of the reads in a location/target pair. The previous implementation had a bug in which one place expected canonical links to be those where the location was the first in pair read, and another method expected canonical links to have location upstream of start. This caused some links to be missed. Here I've fixed the problem, settling on the 'upstream' definition.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLink.java
```diff
@@ -72,23 +72,31 @@ protected void serialize(final Kryo kryo, final Output output ) {
 
     public String toBedpeString(ReadMetadata readMetadata) {
         final SVInterval sourceInterval = source.getInterval();
```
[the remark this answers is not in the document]
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/EvidenceTargetLinkClusterer.java
```diff
@@ -25,9 +25,6 @@ public EvidenceTargetLinkClusterer(final ReadMetadata readMetadata, final int mi
         while (breakpointEvidenceIterator.hasNext()) {
             final BreakpointEvidence nextEvidence = breakpointEvidenceIterator.next();
             if (nextEvidence.hasDistalTargets(readMetadata, minEvidenceMapq)) {
```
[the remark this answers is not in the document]
cwhelan: This was also part of the incorrect logic of link canonicalization that is being fixed in this PR. It actually resulted in missing links.

`instanceof SplitRead` reads about as clearly to me as `isSplitRead` so I think I'll keep it this way.

Incidentally, there seems to be some debate about this, but in my experience I would not call `instanceof` an instance of reflection in Java. I tend to think of reflection as using the Reflections API to do stuff like `object.getClass().getMethod("foo")` to simulate stuff like dynamic dispatch in weird ways. That type of reflection is slow and tends to break easily, but I think that just using `instanceof` to check the type of an object is more akin to straight polymorphism than reflection.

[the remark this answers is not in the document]
cwhelan: Added a little bit of documentation.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/GATKSVVCFHeaderLines.java
```diff
@@ -92,5 +96,9 @@ private static void addSymbAltAlleleLine(final VCFSimpleHeaderLine line) {
 
         addInfoLine(new VCFInfoHeaderLine(GATKSVVCFConstants.TANDUP_CONTRACTION_STRING, 0, VCFHeaderLineType.Flag, "Tandem repeats contraction compared to reference"));
         addInfoLine(new VCFInfoHeaderLine(GATKSVVCFConstants.TANDUP_EXPANSION_STRING, 0, VCFHeaderLineType.Flag, "Tandem repeats expansion compared to reference"));
+
```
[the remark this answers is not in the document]
cwhelan: My feeling is that each type of imprecise variant calling or annotation will need its own logic in terms of annotating/resolving new variants. I was thinking about adding tickets or opening tasks for inversions and tandem duplications next. I think that accurate resolution of imprecise evidence for those variant types will require read depth / CNV information, however. 

You're right that these should eventually go into the FORMAT, but right now we're not producing any genotype information at all. Eventually we should do that.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVContext.java
cwhelan: I think that it is right given that the start of the VariantContext is the position of the base before the event, and you are not allowed to have a zero-length `SimpleInterval`. But I could be wrong about that.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSparkUnitTest.java
cwhelan: changed

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSparkUnitTest.java
cwhelan: changed

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVContextUnitTest.java
```diff
@@ -153,7 +153,7 @@ public void testComposeAlternativeHaplotype(final VariantContext vc, @SuppressWa
 
 
     /**
-     * Tests {@link SVContext#getBreakPointIntervals(int, SAMSequenceDictionary)}.
+     * Tests {@link StructuralVariantContext#getBreakPointIntervals(int, SAMSequenceDictionary, boolean)}.
```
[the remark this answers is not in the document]
cwhelan: Fixed

src/test/resources/org/broadinstitute/hellbender/tools/spark/sv/integration/hg19_DEL.vcf
```diff
@@ -21,10 +21,12 @@
 ##INFO=<ID=INV55,Number=0,Type=Flag,Description="Whether the event represents a 5' to 3' inversion">
 ##INFO=<ID=MAPPING_QUALITIES,Number=.,Type=Integer,Description="Mapping qualities of the contig alignments that support the variant">
 ##INFO=<ID=MAX_ALIGN_LENGTH,Number=1,Type=Integer,Description="Maximum of the minimum aligned lengths of flanking regions from each contig alignment">
+##INFO=<ID=READ_PAIR_SUPPORT,Number=1,Type=Integer,Description="Number of discordant read pairs supporting the variant">
+##INFO=<ID=SPLIT_READ_SUPPORT,Number=1,Type=Integer,Description="Number of split read supplementary mappings supporting the variant">
```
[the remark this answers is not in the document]
cwhelan: Fixed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
SHuang-Broad: I like this idea. 
The change is most likely to happen in SVContext, with two over-loaded ctors, one with taking arguments of read pairs and split reads, one without and delegating to the other with default values. Would this work?

cwhelan: Since I'm annotating preexisting variants and don't currently have the evidence target links at VC creation time, it's hard to do things in constructors without greatly changing the code path (such that I would have to broadcast out all of the evidence target links so that they would be there at VC creation). I'm not sure it's worth it to make that change now.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVContext.java
SHuang-Broad: Ah, that's right!

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSparkUnitTest.java
SHuang-Broad: Sorry I gave the wrong name. But the `for` loop is no longer needed right?

SHuang-Broad:
@cwhelan . Thanks for adding the docs!

However, I still have concerns about three classes: `StrandedInterval`, `EvidenceTargetLink` and `BreakpointEvidence`. They seem to be tightly coupled, through an un-documented concept called "distal target" (not sure if people with more experience would immediately understand what it is, but it took me some time to grasp). 

* My understanding is that a target in the "distal target" is really an interval that is spanned and/or suggested by a piece of evidence. The "distal" part is for representing mate of discordant read pairs, and non primary locations of split reads. Hence the first mention to distal target, in base class `BreakpointEvidence`'s methods `hasDistalTargets()` and `getDistalTargets()`, should document it.

* By studying the possible arrangements of innies, outties, lefies and righties, the `strand` information stored in `StrandedInterval` could be understood, where upstream/downstream of possible event breakpoint location has 1-1 correspondence with the "target"'s strand. (BTW, the doc currently for the class refers to a method called "getStrand()" in BreakpointEvidence, which I couldn't find). But using the word "strand" is confusing I think, and some example in the documentation/example in `StrandedInterval` is needed.
* `StrandedInterval` at its current state, is almost synonymous to "distal target region".

I am not advocating that this to be done in this PR, but the BreakpointEvidence is so central to the logic (except the kmer acrobatics) that we should document it better.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSparkUnitTest.java
cwhelan: Whoops! Removed the loop.

cwhelan:
@SHuang-Broad 

I'm happy to update the documentation more and/or rename things to make them more clear, but I'm not sure what the best way to do it is. A couple of thoughts:

- Sorry about the erroneous mention of `BreakpointEvidence.getStrand` in the comment for `StrandedInterval`. This was renamed to `BreakpointEvidence.isEvidenceUpstreamOfBreakpoint` when the previous PR was being reviewed. I've updated the comment now. 

- The comment for `BreakpointEvidence.getDistalTargets` currently reads:

```
    /**
     * Returns the distal interval implicated as a candidate adjacency to the breakpoint by this piece of evidence.
     * For example, in the case of a discordant read pair, this would be the region adjacent to the mate of the current
     * read. Returns null if the evidence does not specify or support a possible targeted region (for example, the case
     * of an read with an unmapped mate). Strands of the intervals indicate whether the distal target intervals are
     * upstream or downstream of their proposed breakpoints: true indicates that the breakpoint is upstream of the interval
     * start position; false indicates that the breakpoint is downstream of the interval end position
     */
```

What else would you like to see documented there? 

- The use of the word strand in this case is largely driven by a mapping of these data structures to the BEDPE format, which is the older format for representing breakpoints implied by paired-end mapping data without assembly. If you only consider read pair mappings, strand has the natural interpretation of being the strand to which reads aligned. For example, a deletion's two intervals have strands `+` and `-` because the `+` reads align at left breakpoint and `-` reads align near the right breakpoint. Extending the concept to supplementary mappings of split reads muddies the concept a bit, which made me change the definition of strand to the existing one: whether the evidence suggests a breakpoint upstream of the interval start or downstream of the interval end. 

- I created `StrandedInterval` mostly just as a data container since I was often passing around an interval and an associated strand, and using them in conjunction with the `PairedStrandedIntervalTree` data structure. My goal with those was to have them be utility classes that could be used by anyone without regards to the particular mechanics of imprecise evidence clustering I've implemented here. I'd prefer to put the definition of how we're interpreting the interval and strand in our logic classes (`BreakpointEvidence`, `EvidenceTargetLink`, and EvidenceTargetLinkClusterer`). Does that make sense?

- A "distal target region" can be represented by a `StrandedInterval`. So can the original, proximal (non-distal) location of the breakpoint evidence. An `EvidenceTargetLink` has the two `StrandedInterval` objects representing the proximal and distal locations, and the count of evidence types in the link cluster. Does that make things more clear?

I've added some ASCII-art visual examples to `DiscordantReadPairEvidence.getDistalTargets` and `SplitRead.getDistalTargets`. Do those help?

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSparkUnitTest.java
```diff
@@ -136,9 +136,7 @@ public void testProcessEvidenceTargetLinks(final List<EvidenceTargetLink> etls,
                         metadata, inputVariants, referenceMultiSource);
 
         Assert.assertEquals(processedVariantContexts.size(), expectedVariants.size());
-        for (int i = 0; i < processedVariantContexts.size(); i++) {
-            VariantContextTestUtils.assertEqualVariants(processedVariantContexts, expectedVariants);
-        }
+        VariantContextTestUtils.assertEqualVariants(processedVariantContexts, expectedVariants);
```
SHuang-Broad: extremely trivial: I think `VariantContextTestUtils.assertEqualVariants` tests for size equality first, so the line above is not necessary any more.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
```diff
@@ -422,6 +424,27 @@ private boolean isHighQualityMapping(final ReadMetadata readMetadata, final int
                     && ! saInterval.overlaps(getLocation());
         }
 
+        /**
```
SHuang-Broad: Thanks for these examples. They make the concept much easier to understand!

PR #3629, opened 2017-09-27
add elapsed time to assembly info in intervals file

tedsharpe:
Here's a tiny commit that's mostly just code cleanup.  @SHuang-Broad -- I sprinkled some "final"s around.


src/main/java/org/broadinstitute/hellbender/tools/spark/utils/SparkUtils.java
```diff
@@ -0,0 +1,16 @@
+package org.broadinstitute.hellbender.tools.spark.utils;
+
+import org.apache.log4j.Logger;
+import org.apache.spark.broadcast.Broadcast;
+
+public class SparkUtils {
```
magicDGS: There's already a `SparkUtils.java` living in `org.broadinstitute.hellbender.utils.spark`. Maybe this can be moved there...

tedsharpe: Thanks.  Moved it there, as you suggested.

PR #3630, opened 2017-09-28
Adds supplementary alignment info into fastq output, also additional…

vruano:
… tags like NM and AS


vruano:
Conveniently divided in two commits where the second is the testing code that has the bulk of the changes.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AlignmentInterval.java
```diff
@@ -129,11 +171,54 @@ public AlignmentInterval(final String samSAtagString) {
...
+
+    private static int parseZeroOrPositiveInt(final String str, final int defaultValue, final String errorMessage) {
         try {
-            return Integer.parseInt(str);
-        } catch (NumberFormatException ex) {
-            throw new IllegalArgumentException("not a valid integer in: " + str, ex);
+            if (str.equals(NO_VALUE_STR)) {
+                return defaultValue;
+            } else {
+                return ParamUtils.isPositiveOrZero(Integer.parseInt(str), errorMessage + ": " + str);
```
tedsharpe: It would be clearer if the try...catch was around just this line rather than the whole method body.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFastqUtils.java
```diff
@@ -93,91 +94,159 @@ else if (ce == NEGATIVE.charEncoding)
...
-                final Matcher matcher = MAPPING_DESCRIPTION_PATTERN.matcher(mappingDescription);
-                if (!matcher.find()) {
-                    throw new IllegalArgumentException("invalid mapping description: '" + mappingDescription + "'");
+        public Mapping(final GATKRead read) {
+            Utils.nonNull(read);
+            if (read.isUnmapped()) {
+                intervals = Collections.emptyList();
+            } else if (!read.hasAttribute(SAMTag.SA.name())) {
+                if (read.isSupplementaryAlignment()) {
+                    throw new IllegalArgumentException("a supplementary alignment read record must supply a SA tag");
```
tedsharpe: While true for BWA, I'm not sure it's true in general that supplementary alignments must have SA tags.  Then there's also the confusion between XA and SA tags.  If it were me, I'd leave this check out entirely.
If you do leave this test in, maybe you could find a better exception to throw than an IllegalArgumentException, which doesn't seem to me very descriptive of a malformed read.

vruano: I think I will leave it for now indicating that this situation is unsupported.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AlignmentInterval.java
vruano: Ok.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFastqUtils.java
vruano: I'm using GATKException.

PR #2028, opened 2016-07-20
disabling htsjdk snappy

lbergelson:
adding snappy.disable=true to prevent htsjdk from using snappy
this is a temporary solution for #2026 until we can patch htsjdk to fix incompatibilities with more modern snappy

fixes #2026

Issue #2026, opened 2016-07-20
SplitNCigarReads fails with NoClassDefFoundError


lbergelson:
SplitNCigarReads is failing due to incompatibilities between htsjdk's version of snappy and Spark's version.

Temporary solution is to add system property 'disable.snappy' to force htsjdk to fallback to pure java.  Longer term solution likely involves patches to htsjdk and possibly snappy itself.

```
 ./gatk-launch SplitNCigarReads -I src/test/resources/large/CEUTrio.HiSeq.WGS.b37.NA12878.20.21.bam -O out.bam -R src/test/resources/large/human_g1k_v37.20.21.fasta

Running:
    /Users/louisb/Workspace/gatk/build/install/gatk/bin/gatk SplitNCigarReads -I src/test/resources/large/CEUTrio.HiSeq.WGS.b37.NA12878.20.21.bam -O out.bam -R src/test/resources/large/human_g1k_v37.20.21.fasta
15:31:00.516 INFO  IntelGKLUtils - Trying to load Intel GKL library from:
	jar:file:/Users/louisb/Workspace/gatk/build/install/gatk/lib/gkl-0.1.2.jar!/com/intel/gkl/native/libIntelGKL.dylib
15:31:00.552 INFO  IntelGKLUtils - Intel GKL library loaded from classpath.
[July 20, 2016 3:31:00 PM EDT] org.broadinstitute.hellbender.tools.walkers.rnaseq.SplitNCigarReads  --output out.bam --input src/test/resources/large/CEUTrio.HiSeq.WGS.b37.NA12878.20.21.bam --reference src/test/resources/large/human_g1k_v37.20.21.fasta  --refactor_NDN_cigar_string false --maxReadsInMemory 150000 --maxMismatchesInOverhang 1 --maxBasesInOverhang 40 --doNotFixOverhangs false --disable_all_read_filters false --interval_set_rule UNION --interval_padding 0 --readValidationStringency SILENT --secondsBetweenProgressUpdates 10.0 --disableSequenceDictionaryValidation false --createOutputBamIndex true --createOutputBamMD5 false --addOutputSAMProgramRecord true --help false --version false --verbosity INFO --QUIET false --use_jdk_deflater false
[...]

PR #3635, opened 2017-09-28
re-enable snappy use in htsjdk

lbergelson:
htsjdk has updated to a newer version of snappy (https://github.com/samtools/htsjdk/pull/872) which makes it compatible with the rest of the world's snappy

this means we can re-enable snappy usage in htsjdk which should give performance improvements in tools that use `SortingCollection`

this is a permanent solution to #2026 that replaces the changes we made in #2028


lbergelson:
@droazen Could you take a look at this quick one since chris is out today?

PR #3636, opened 2017-09-29
fixed bug in which F1R2 and F2R1 annotation kept discarded alleles

davidbenjamin:
@LeeTL1220 you were right.  Fortunately the fix is very simple.


LeeTL1220:
@davidbenjamin Can you make a test for this?

davidbenjamin:
@LeeTL1220 done.  I put it in the M2 integration test because it is part of a bigger potential issue, that of phantom alleles from the likelihoods that don't make it into the variant call.  Back to you.

LeeTL1220:
@davidbenjamin Looks good... merge away

PR #3639, opened 2017-09-29
PSUtils getMatchesLessDeletions() should not throw error if number of…

mwalker174:
… mismatches is too large

Removes an error that is thrown when the number of mismatches is greater than the number of mismatches/mismatches in the cigar. 

It appears that some aligners do this. I am seeing it frequently in soft-clipped reads in TCGA RNA-seq BAMs (maybe a STAR bug?). 

I would prefer to remove the thrown error rather than assume the dev will be aware of this issue and will catch it externally.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
@@ -101,7 +101,6 @@ public final class PSUtils {
                 numDeletions += e.getLength();
             }
         }
-        if (numMatches < 0) throw new IllegalArgumentException("numMismatches was greater than the number of matches/mismatches in the cigar");
         return numMatches - numDeletions;
     }
 }
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.hellbender.tools.spark.pathseq;
 
 import htsjdk.samtools.*;
+import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
```diff
@@ -16,6 +17,8 @@
  */
 public final class PSUtils {
 
+    protected static final Logger logger = LogManager.getLogger(PSUtils.class);
```
vruano: Make it private.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
```diff
@@ -101,6 +104,10 @@ public static int getMatchesLessDeletions(final Cigar cigar, final int numMismat
                 numDeletions += e.getLength();
             }
         }
+        if (numMatches < 0)  {
+            logger.warn("Invalid arguments passed to getMatchesLessDeletions(): numMismatches was greater than the number of matches/mismatches in the cigar. Returning 0.");
```
vruano: This is an infrequent occurrence, right? Otherwise I would place a max count of warning messages. Since it all be done inside a (infrequent-run) if body then you don't need to worry a bout using slower synchronization. Only if you notice that the amount of warning messages is large.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
mwalker174: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
mwalker174: It is relatively infrequent (a small proportion of reads). I will leave it as is.

PR #3640, opened 2017-09-29
Have PSFilter strip mate number from read names

mwalker174:
Adds a read transformer that removes the "/1" and "/2" from read names to the PathSeq filter phase. This seems to be the way TCGA RNA-seq BAMs are formatted.


SHuang-Broad:
Looks good to me, but voicing one concern/question: if the TCGA data have read names set this way, are the SAM flags set appropriately?

mwalker174:
Yes the pair flags are set properly.

PR #3643, opened 2017-10-02
Automatically distribute the BWA-MEM index image file to executors for BwaSpark

tomwhite:
This removes the need to use the Spark `--files` argument and the `BwaSpark` `--bwaMemIndexImage` argument, since instead the `.img` file will be distributed for the user based on the reference argument (`-R`).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
@@ -25,18 +25,11 @@ public final class BwaSpark extends GATKSparkTool {
 
     public static final String SINGLE_END_ALIGNMENT_FULL_NAME = "singleEndAlignment";
     public static final String SINGLE_END_ALIGNMENT_SHORT_NAME = "SE";
-    public static final String BWA_MEM_INDEX_IMAGE_FULL_NAME = "bwaMemIndexImage";
-    public static final String BWA_MEM_INDEX_IMAGE_SHORT_NAME = "image";
 
     @Argument(doc = "the output bam",
             shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
             fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME)
     private String output;
-
-    @Argument(doc = "the bwa mem index image file name that you've distributed to each executor",
-            fullName = BWA_MEM_INDEX_IMAGE_FULL_NAME,
-            shortName = BWA_MEM_INDEX_IMAGE_SHORT_NAME )
-    private String indexImageFile;
 
     @Argument(doc = "run single end instead of paired-end alignment",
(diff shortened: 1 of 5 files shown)


tedsharpe:
I have no problems whatsoever with the code, but I do have some concerns about the design:

In the SV group's pipeline, we distribute this multi-gig file from its home in the cloud once at cluster-creation time, and then reuse it for multiple client executions.  There are no superfluous copies lying about anywhere, and no redundant copying operations.  We can give it any name we wish, and put it anywhere we desire (except that the path must be the same on every worker).

This code, if I'm reading it correctly, will redistribute the file from a non-permanent home on the master's local file system or on the HDFS (to which it must be copied redundantly at least once per cluster instantiation), and then it will further be redundantly copied to a temporary location on each worker's local file system with every client execution.  I don't know if that's overhead that we can live with, or whether that might prevent us from writing clients with brief execution times.  I'm just opening the issue for discussion.

We also lose a little flexibility in that the image must live in the same directory as the reference, though I don't think that's a serious drawback -- it's a perfectly logical place for it.  However, since we're just appending a fixed extension (".img") to the reference name we can only have one image file per reference, which may be a problem because different images need to be created for different versions of bwa and for various options such as the list of alt contigs.  We can handle the first problem by insisting that all clients on a particular cluster stick to one version of bwa, which is probably a good idea, anyway, but I think we're stuck if clients need to specify various alt contig lists.  It might be better to provide a default path of "ref-name"+".img", but allow that default to be overridden.

Also, just to twist the knife a bit, it's too bad we never reviewed my PR for gatk-bwamem-jni, which version-stamped the images for safety.  It's now languished since July, and has suffered serious bit rot.

src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSpark.java
```diff
@@ -25,19 +25,12 @@
...
 
     @Argument(doc = "the output bam",
             shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
             fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME)
     private String output;
 
-    @Argument(doc = "the bwa mem index image file name that you've distributed to each executor",
-            fullName = BWA_MEM_INDEX_IMAGE_FULL_NAME,
-            shortName = BWA_MEM_INDEX_IMAGE_SHORT_NAME )
-    private String indexImageFile;
```
magicDGS: Maybe the indexImageFile can be an optional argument; if not present, the default img file will be used, otherwise use the one provided. Thus, this requires to move the code for detecting the image to the tool itself. Is that what you mean, @tedsharpe?

tomwhite:
Thanks for your response, @tedsharpe. I've reworked the design so that it supports the option of providing an image file that has already been distributed across the cluster, as suggested by @magicDGS. So it should be compatible with the existing way of doing things.

Does this look better?

tedsharpe:
Yes, that addresses all my concerns.  Looks great.  Thanks.

PR #3655, opened 2017-10-03
VQSR serialized model sets annotation order

lucidtronix:
This addresses the problem where serialized GMMs for VQSR assumed the annotation order would be the same between the commands that generated them and the commands that used them. VQSR no longer depends on the commandline order of the annotations.


src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantDataManager.java
```diff
@@ -56,7 +56,7 @@ public void setNormalization(final Map<String, Double> anMeans, final Map<String
         return data;
     }
 
-    public void normalizeData(final boolean calculateMeans) {
+    public void normalizeData(final boolean calculateMeans, List<Integer> theOrder) {
```
lbergelson: this needs javadoc

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -487,6 +486,25 @@ else if (null != sequenceDictionary) {
         }
     }
 
+
+    private void orderAndValidateAnnotations(final GATKReportTable annotationTable){
```
droazen: Can you please add an integration test that fails before this fix, and passes after it? Also should add a basic unit test for the new `orderAndValidateAnnotations()` method.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibrator.java
```diff
@@ -487,6 +486,25 @@ else if (null != sequenceDictionary) {
...
+        annotationOrder = new ArrayList<Integer>(dataManager.annotationKeys.size());
+
+        for (int i = 0; i < annotationTable.getNumRows(); i++){
+            String serialAnno = (String)annotationTable.get(i, "Annotation");
+            for (int j = 0; j < dataManager.annotationKeys.size(); j++) {
+                if (serialAnno.equals( dataManager.annotationKeys.get(j))){
```
droazen: Can you do a hash lookup against the `dataManager` to avoid an n^2 algorithm? (might not matter if these are small)

lucidtronix: Yeah I was being lazy, but I've never seen us use more than 7 annotations.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantDataManager.java
lucidtronix: Added

lucidtronix:
Added unit and integration tests.  Had to make a few things protected for testing and added a vcf with a single SNP (where @ldgauthier found this bug) for the integration test.  Back to you @droazen.

src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibratorModelOutputUnitTest.java
```diff
@@ -1,7 +1,9 @@
 package org.broadinstitute.hellbender.tools.walkers.vqsr;
```
magicDGS: After #3475, this should extend `GATKBaseTest`

ldgauthier:
@lucidtronix here's a blast from the past.  You're going to have to rebase and I hope you don't have too many merge conflicts.

@droazen while you're doing your Monday issue review, can you refresh your memory on this one?

lucidtronix:
@droazen back to you

lucidtronix:
@droazen bump

src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/VariantRecalibratorModelOutputUnitTest.java
```diff
@@ -1,7 +1,9 @@
...
 
+import avro.shaded.com.google.common.collect.Lists;
```
droazen: I think you want `com.google.common.collect.Lists` here.

PR #3663, opened 2017-10-04
Added note on updating sdk frequently

chandrans:
User brought up that Broad's google-cloud-sdk is not up to date, so I added a note telling people to upgrade regularly.


lbergelson:
@chandrans Thank you!

Issue #3664, opened 2017-10-05
SamAssertionUtils silently failing


tomwhite:
Since the Picard changes in #3620, SamAssertionUtils has been failing silently. See e.g. the Standard error tab for https://storage.googleapis.com/hellbender-test-logs/build_reports/13120.7/tests/test/classes/org.broadinstitute.hellbender.tools.spark.pipelines.ReadsPipelineSparkIntegrationTest.html:

```
USAGE: SortSam [arguments]
...
input is not a recognized option
```

tomwhite:
@cmnbroad any thoughts about what is happening here - something to do with case (input vs INPUT)?

cmnbroad:
@tomwhite Yeah - I know what it is. PR coming.

PR #3665, opened 2017-10-05
Fix SamAssertionUtils SortSam call.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/3664.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/test/SamAssertionUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/test/SamAssertionUtils.java
@@ -378,15 +378,25 @@ public final class SamAssertionUtils {
     private static void sortSam(final File input, final File output, final File reference, final ValidationStringency stringency) {
         final SortSam sort = new SortSam();
 
-        final ArgumentsBuilder args = new ArgumentsBuilder();
-
-        args.addInput(input);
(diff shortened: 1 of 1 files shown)


tomwhite:
+1 fixes the issue for the test I was running. Thanks @cmnbroad

src/main/java/org/broadinstitute/hellbender/utils/test/SamAssertionUtils.java
```diff
@@ -378,15 +378,25 @@ public static void assertCRAMContents(final File putativeCRAMFile) {
     private static void sortSam(final File input, final File output, final File reference, final ValidationStringency stringency) {
         final SortSam sort = new SortSam();
 
-        final ArgumentsBuilder args = new ArgumentsBuilder();
-
-        args.addInput(input);
-        args.addOutput(output);
+        // We can't use ArgumentsBuilder since it assumes GATK argument names, but we're running a Picard
```
lbergelson: maybe we should change ArgumentBuilder to use the short names which I think are shared between picard and gatk?

cmnbroad: I can do so if you want, but addVCF will still do the wrong thing. I think it might be better to just avoid the confusion altogether.

PR #3666, opened 2017-10-05
Add BWA-MEM as an option to ReadsPipelineSpark

tomwhite:


tedsharpe:
I think that there's too much flexibility to choose arguments that will result in a corrupt product in ReadsPipelineSpark:

Whether or not to align, and whether to align in single-ended or paired mode ought to be dictated by the input data, not by some data-jockey's choice of arguments.  I'm prepared to be corrected on this, but I think that paired alignment will only work on queryname sorted, paired input reads.  If the data has already been aligned and coordinate sorted, I'd expect things to blow up because the pairs won't be adjacent, and I don't think we do any input sorting.  I think we ought to be inferring whether to align, and whether to align singly or paired, from the input.  If we want to force realignment of an already-aligned input, we should just have the user run a tool that strips alignment info and produces unaligned, queryname-sorted input to this tool.  (Or we could incorporate that functionality into this tool, I suppose.)

Second crazy obscure issue:  if the input file hasn't been aligned, and therefore doesn't contain a dictionary, we're going to try to borrow one from the reference when we do alignment.  That's going to lead to insanity if the reference happens to be a 2bit file in which the order of the contigs is scrambled.  We can use such a source for the contig metadata, but the order of the contigs must be prescribed by the reference image file.  (The list of reference contig names in correct order is available from the bwa image.)

Third issue maybe isn't really an issue:  the BwaSparkEngine code doesn't copy tags from input to output, except for read group.  (That's because in the case that you're realigning, you'd want to strip all tags associated with the previous alignment, but there's no comprehensive list of tags that are associated with alignment vs. those that have to do with read itself, and there can't be due to the possibility of custom tags that might go into either category.)

tedsharpe:
Sorry for the delay in reviewing this PR.

tomwhite:
@tedsharpe thanks for your comments! Adding @droazen and @lbergelson for thoughts too.

1. Is it sufficient to look at the first two reads in the BAM file to determine whether the file is aligned or not? Scanning the whole file to see if it's aligned seems like overkill. I've updated this PR to look at the first two reads. Also, should we have a way of overriding these settings, to force alignment or single/paired mode? (BTW the single/paired read thing is an existing argument in BwaSpark.)

2. Is this an intrinsic problem with 2bit, or just the implementation? https://github.com/bigdatagenomics/adam/pull/1508 will fix the implementation (when ADAM 0.23 is out). Perhaps the best thing is to use the BWA image to sort the contig names correctly?

3. Not sure what is needed here.

tedsharpe:
I defer to @droazen and @lbergelson on what the design should actually be, but my recommendation would be as follows:
Have an argument that forces single-ended mode -- that's always a valid possibility regardless of whether the input data is single-ended or paired, and regardless of how it's sorted.
If that flag is set, then just pass the reads to the BwaEngine and align single-ended without further ado.
If that flag is not set, then check the first read to see if it's paired or not (GATKRead isPaired method).  We'll assume all the rest of the reads are similarly either paired or unpaired.
If the first read is unpaired, just pass the reads to the BwaEngine and align single-ended.
If the first read is paired:
1) check the sort order in the header to make sure it's queryname ordered.
2) check that the first read is the first read of the pair (read.isFirstOfPair).
3) get the 2nd read and check that it's also paired, that it's the 2nd read of the pair, and that it has the same name as the first read.
If any of these checks fail, whine and bail.  Otherwise, pass the reads to the BwaEngine and align in paired mode.

tedsharpe:
Oh, and on point 3 about attribute copying:  I don't know what we want to do, either.  One possibility would be to allow the user to specify a list of tag names that should be preserved.  Maybe it could default to RG and OQ?
But I'd definitely check with @lbergelson and or @droazen before implementing this.

magicDGS:
Would be nice to have also the `BC` tag included by default in point 3, to keep barcode information if present.

droazen:
@lbergelson & I think @tedsharpe 's proposal above sounds reasonable (including a whitelist of attributes to copy, optionally configurable via the command line).

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
```diff
@@ -116,42 +118,66 @@ protected void runTool(final JavaSparkContext ctx) {
         //TOOO: should this use getUnfilteredReads? getReads will apply default and command line filters
         final JavaRDD<GATKRead> initialReads = getReads();
 
-        final JavaRDD<GATKRead> markedReadsWithOD = MarkDuplicatesSpark.mark(initialReads, getHeaderForReads(), duplicatesScoringStrategy, new OpticalDuplicateFinder(), getRecommendedNumReducers());
+        List<GATKRead> firstReads = initialReads.take(2);
```
lbergelson: This seems like a kind of sketchy check.  It's very plausible that you'd just randomly have 2 unmapped reads at the start of your file.  

I think that if someone specifies an image file, then they are saying they want to align, and if they don't pass it in they don't want to align.

tomwhite: @lbergelson I agree that this isn't reliable: I added it as a demonstration really. The problem is that specifying an image file is optional - it's inferred from the name of the reference since #3643.

So perhaps we do need an `--align` flag, although @tedsharpe doesn't like that... Is there a reliable way to determine if the reads have been aligned by looking at the contents?

Similarly if unmapped reads can pop up anywhere, then determining single or paired alignment is unreliable.

Any suggestions on how to move forward on this?

tedsharpe:
I don’t have strong feelings about an —align flag.
If this is supposed to be a re-alignment tool as well as an alignment tool, I think we’ll need it.
The “paired read” flag is a property of the read, and not of the alignment. It should be present even in an unaligned BAM. The only assumption I think we need to make is that the pairedness of all reads is consistent with that of the first read.

tomwhite:
Thanks @tedsharpe. Perhaps the best thing to do at this point is have the existing flags (plus `--align`) and add some checks that look for inconsistencies and fail or warn as appropriate.

src/main/java/org/broadinstitute/hellbender/utils/test/BaseTest.java
```diff
@@ -70,6 +70,8 @@ public void setTestVerbosity(){
 
     public static final String b37_2bit_reference_20_21 = largeFileTestDir + "human_g1k_v37.20.21.2bit";
 
+    public static final String b37_reference_20_21_img = largeFileTestDir + "human_g1k_v37.20.21.fasta.img";
```
magicDGS: This should be moved to `GATKBaseTest`, because `BaseTest` should be clean from any repo-specific test resource...

tomwhite: Fixed - thanks for letting me know

tomwhite:
I've rebased the branch and re-instated the `--align` flag.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
```diff
@@ -116,42 +121,59 @@ protected void runTool(final JavaSparkContext ctx) {
         //TOOO: should this use getUnfilteredReads? getReads will apply default and command line filters
         final JavaRDD<GATKRead> initialReads = getReads();
 
-        final JavaRDD<GATKRead> markedReadsWithOD = MarkDuplicatesSpark.mark(initialReads, getHeaderForReads(), duplicatesScoringStrategy, new OpticalDuplicateFinder(), getRecommendedNumReducers());
+        final JavaRDD<GATKRead> alignedReads;
+        final SAMFileHeader header;
+        final BwaSparkEngine engine;
+        if (align) {
+                engine = new BwaSparkEngine(ctx, referenceArguments.getReferenceFileName(), bwaArgs.indexImageFile, getHeaderForReads(), getReferenceSequenceDictionary());
+                alignedReads = !bwaArgs.singleEndAlignment ? engine.alignPaired(getReads()) : engine.alignUnpaired(getReads());
```
droazen: You're calling `getReads()` again here -- can you just use `initialReads` above? Is this expensive?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
```diff
@@ -116,42 +121,59 @@ protected void runTool(final JavaSparkContext ctx) {
         //TOOO: should this use getUnfilteredReads? getReads will apply default and command line filters
         final JavaRDD<GATKRead> initialReads = getReads();
 
-        final JavaRDD<GATKRead> markedReadsWithOD = MarkDuplicatesSpark.mark(initialReads, getHeaderForReads(), duplicatesScoringStrategy, new OpticalDuplicateFinder(), getRecommendedNumReducers());
+        final JavaRDD<GATKRead> alignedReads;
+        final SAMFileHeader header;
+        final BwaSparkEngine engine;
```
droazen: Can you rename to `bwaEngine`?

src/test/java/org/broadinstitute/hellbender/GATKBaseTest.java
```diff
@@ -42,6 +42,8 @@
 
     public static final String b37_2bit_reference_20_21 = largeFileTestDir + "human_g1k_v37.20.21.2bit";
 
+    public static final String b37_reference_20_21_img = largeFileTestDir + "human_g1k_v37.20.21.fasta.img";
```
droazen: How large is this new file?

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
droazen: Also, `getReads()` does filtering -- did you want to pass in the unfiltered reads (`getUnfilteredReads()`) to BWA?

tomwhite: 1.  It doesn't make a difference, since it is lazy. But I've changed the code to use `initialReads` to make it less confusing.
2. Good question! What do you think is best, @droazen?

droazen: I would think that you'd want to pass BWA the unfiltered reads, and then do filtering after alignment. Do you agree, @tedsharpe ?

tedsharpe: On the one hand, it seems wasteful to spend time aligning reads that you're about to discard.  On the other hand, if you're doing pairwise alignment and your filters don't either pass or reject *both* members of the pair consistently, you'll have a royal mess.  Since filters are designed to operate on reads, not on read pairs, it would be hard (but not impossible) to guarantee correct filter behavior in this context.
All of which is a long-winded way of saying that you probably want to align unfiltered reads like @droazen said.

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
tomwhite: Done

src/main/java/org/broadinstitute/hellbender/tools/spark/pipelines/ReadsPipelineSpark.java
tomwhite: Thanks for the comments @tedsharpe and @droazen. I've updated the code to align on unfiltered reads then do the alignment. I've updated `BwaSpark` and `BwaAndMarkDuplicatesPipelineSpark` too. (BTW do we still need the latter do you think?)

tomwhite:
@droazen @lbergelson are you ok for this to go in now?

lbergelson:
@tomwhite This looks good to me.

PR #3667, opened 2017-10-05
fixing sample name reordering in GenomicsDBImport

lbergelson:
the sample swap issue seems to have been caused by the fact that the callset.json files had the sample names globally sorted
the actual data was entered in the order the sample names were input, but each bactch was sorted within the batch

this produced a mismatch between data and samplename mapping

fixing it by using sorted map everywhere sample names are handled
modifying our tests to use an out of order sample map file and multiple batches

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -199,7 +199,7 @@ public final class GenomicsDBImport extends GATKTool {
 
     // Sorted mapping between sample names and corresponding GVCF file name
     //
-    // This must be sorted or it will result in sample name swaps in the output database.
+    // IMPORTANT: This must be sorted or it will result in sample name swaps in the output database.
     // This happens because the callset json is generated independently from the import process
     // each imported batch is then sorted, so if we have an unsorted list we'll end up with different global vs batch
     // sorting.
@@ -351,7 +351,7 @@ public final class GenomicsDBImport extends GATKTool {
         logger.info("Callset Map JSON file will be written to " + callsetMapJSONFile);
         logger.info("Importing to array - " + workspace + "/" + GenomicsDBConstants.DEFAULT_ARRAY_NAME);
 
-        //use the given order so this is consistent with the batch importing
+        //Pass in true here to use the given ordering, since sampleNameToVcfPath is already sorted
         callsetMappingPB = GenomicsDBImporter.generateSortedCallSetMap(new ArrayList<>(sampleNameToVcfPath.keySet()), true);
         initializeInputPreloadExecutorService();
     }
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -198,6 +198,12 @@ public final class GenomicsDBImport extends GATKTool {
     private List<ChromosomeInterval> intervals;
 
     // Sorted mapping between sample names and corresponding GVCF file name
+    //
+    // This must be sorted or it will result in sample name swaps in the output database.
+    // This happens because the callset json is generated independently from the import process
+    // each imported batch is then sorted, so if we have an unsorted list we'll end up with different global vs batch
+    // sorting.
+    // We presumptively sort here so we will have consistent sorting.
     private SortedMap<String, Path> sampleNameToVcfPath = new TreeMap<>();
 
     // Needed as smartMergeHeaders() returns a set of VCF header lines
@@ -345,6 +351,7 @@ public final class GenomicsDBImport extends GATKTool {
         logger.info("Callset Map JSON file will be written to " + callsetMapJSONFile);
         logger.info("Importing to array - " + workspace + "/" + GenomicsDBConstants.DEFAULT_ARRAY_NAME);
 
+        //use the given order so this is consistent with the batch importing
         callsetMappingPB = GenomicsDBImporter.generateSortedCallSetMap(new ArrayList<>(sampleNameToVcfPath.keySet()), true);
         initializeInputPreloadExecutorService();
     }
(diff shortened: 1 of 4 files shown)


lbergelson:
@droazen I believe this should solve the issue, but you should take a careful look as well.

yfarjoun:
can we have two commits...one with failing test and one that fixes it?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -197,8 +197,8 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
     // Intervals from command line (singleton for now)
     private List<ChromosomeInterval> intervals;
 
-    // Linked hash map between sample names and corresponding GVCF file name
-    private LinkedHashMap<String, Path> sampleNameToVcfPath = new LinkedHashMap<>();
+    // Sorted mapping between sample names and corresponding GVCF file name
```
droazen: Add a prominent comment mentioning that maintaining the sample names in sorted order here is critical, and explain why!

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -345,10 +345,7 @@ public void onTraversalStart() {
         logger.info("Callset Map JSON file will be written to " + callsetMapJSONFile);
         logger.info("Importing to array - " + workspace + "/" + GenomicsDBConstants.DEFAULT_ARRAY_NAME);
 
-        // Passing in false here so that sample names will be sorted.
-        // This is needed for consistent ordering across partitions/machines
-        callsetMappingPB = GenomicsDBImporter.generateSortedCallSetMap(new ArrayList<>(sampleNameToVcfPath.keySet()), false);
-
+        callsetMappingPB = GenomicsDBImporter.generateSortedCallSetMap(new ArrayList<>(sampleNameToVcfPath.keySet()), true);
```
droazen: Add comment explaining why the final parameter to this method call is `true`.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -449,9 +446,9 @@ public Object onTraversalSuccess() {
      * @param lowerSampleIndex  0-based Lower bound of sample index -- inclusive
      * @return  Feature readers to be imported in the current batch
```
droazen: Document why this method returns a `SortedMap`

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -199,33 +199,35 @@ private static void checkGenomicsDBAgainstExpected(final String workspace, final
     }
 
     @Test
-    public void testSampleMappingFileInsteadOfVCFs() throws IOException {
-        final File sampleNameFile = createTempSampleMapFile();
+    public void testSampleMappingFileInsteadOfVCFsWithMultipleBatches() throws IOException {
+        final File sampleNameFile = createOutOfOrderTempSampleMapFile();
```
droazen: New test case(s) should supplement, not replace, existing ones. I think what you should do here is create a `DataProvider` that generates test cases with all combinations of:

-batch sizes 0, 1, 2, 3, 4
-samples in order vs. samples out of order
-sample map file vs. -V

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -197,8 +197,14 @@ public int getDefaultCloudIndexPrefetchBufferSize() {
     // Intervals from command line (singleton for now)
     private List<ChromosomeInterval> intervals;
 
-    // Linked hash map between sample names and corresponding GVCF file name
-    private LinkedHashMap<String, Path> sampleNameToVcfPath = new LinkedHashMap<>();
+    // Sorted mapping between sample names and corresponding GVCF file name
+    //
+    // This must be sorted or it will result in sample name swaps in the output database.
```
droazen: add "IMPORTANT:" here in all caps

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -345,10 +351,8 @@ public void onTraversalStart() {
         logger.info("Callset Map JSON file will be written to " + callsetMapJSONFile);
         logger.info("Importing to array - " + workspace + "/" + GenomicsDBConstants.DEFAULT_ARRAY_NAME);
 
-        // Passing in false here so that sample names will be sorted.
-        // This is needed for consistent ordering across partitions/machines
-        callsetMappingPB = GenomicsDBImporter.generateSortedCallSetMap(new ArrayList<>(sampleNameToVcfPath.keySet()), false);
-
+        //use the given order so this is consistent with the batch importing
```
droazen: "Pass in true here to use the given ordering, since `sampleNameToVcfPath` is already sorted"

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportUnitTest.java
```diff
@@ -10,13 +10,15 @@
 import org.testng.Assert;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
+import spire.math.Sort;
```
droazen: Remove stray import

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -198,34 +196,75 @@ private static void checkGenomicsDBAgainstExpected(final String workspace, final
...
+            results.add(new Object[] {new ArgumentsBuilder()
+                    .addArgument(GenomicsDBImport.BATCHSIZE_ARG_NAME, String.valueOf(batchSize))
+                    .addFileArgument(GenomicsDBImport.SAMPLE_NAME_MAP_LONG_NAME, createInOrderSampleMap())});
+
+            //out of order sample map
+            results.add(new Object[] {new ArgumentsBuilder()
+                    .addArgument(GenomicsDBImport.BATCHSIZE_ARG_NAME, String.valueOf(batchSize))
+                    .addFileArgument(GenomicsDBImport.SAMPLE_NAME_MAP_LONG_NAME, outOfOrderSampleMap)});
+        }
+        return results.iterator();
```
droazen: Add a least one additional test case that uses `--vcfInitializerThreads` with a value > 1, since that is a different code path that you also modified.

PR #3671, opened 2017-10-06
fix bug causing error finding ref base for imprecise deletions

cwhelan:
This PR fixes a critical error that was causing an "invalid interval" exception to be thrown while calling imprecise deletion intervals. The problem was a mismatch between the sequence dictionary in the reference and the read metadata's contig ID - contig name mapping. I've modified the code to always use the read metadata when translating the contig ID in `EvidenceTargetLink` intervals into contig names.


cwhelan:
@tedsharpe this should fix your error on NA12878; I just ran that sample through without hitting an exception with this code change. Want to take a look at the changes?

Issue #3569, opened 2017-09-12
BucketUtils does not make use of FileSystemProviders.installedProviders()


vruano:
It seems that you can find out what FileSystem corresponds to what URI/URL by using the return of FileSystemProviders.installedProviders. Instead BucketUtils has hard-wired scheme names and has many conditional (isCould... isHadoop) for operations that can be done by using the appropriate FileSystem implementation that in turn can be resolved looking at that list.

I guess that there are extant utility classes that process that list once for subsequent queries (e.g. scheme-name -> FS), but I have not confirmed that. Otherwise it is easy to write one.

Otherwise there is a reason what we need to do our own thing in BucketUtils?

lbergelson:
BucketUtils was a solution before we had Filesystem providers.  It's stuck around as a parallel set of code because we couldn't trust the providers at first.  In the long run it should be removed and replaced entirely by `Files`  operations.  We need to test that all the functionality exists / works as expected though, and it hasn't been a high priority to do so.  Particularly, I'm not sure we have a lot of faith in the HDFS NIO plugin, so we may need to keep around special cases for that.  It could definitely at least be simplified a lot though.

PR #3674, opened 2017-10-06
Cleanup sv utils

SHuang-Broad:
Periodic housekeeping.


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.*;
+
+public final class SVFileUtils {
+
+    private static final String REFERENCE_GAP_INTERVAL_FILE_COMMENT_LINE_PROMPT = "#";
+
+    /**
+     * Creates a directory, in local FS, HDFS, or Google buckets to write to.
+     */
+    public static boolean createDirToWriteTo(final String pathString) {
```
magicDGS: I think that this method, with some refactoring, could belong to `IOUtils`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+        try {
+            Utils.nonNull(pathString);
+            if ( java.nio.file.Files.exists(java.nio.file.Paths.get(pathString)) )
```
magicDGS: I think that the `java.nio.Path` should be created with `IOUtils` here, and outside to re-use afterwards. Then the method could be simplify, I guess...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+                final java.nio.file.Path dir = java.nio.file.Files.createDirectory(IOUtils.getPath(pathString));
+                isSuccessful = java.nio.file.Files.isDirectory(dir) && java.nio.file.Files.isWritable(dir);
+            }
+            return isSuccessful;
+        } catch (final IOException x) {
+            throw new GATKException("Could not create directory: " + pathString, x);
+        }
+    }
+
+    public static void writeLinesToSingleFile(final Iterator<String> linesToWrite, final String fileName) {
```
magicDGS: Cannot the code using this be changed to `Files.write(Path, Iterable<? extends CharSequence>, OpenOption...)`. If the `Path` is obtained with `IOUtils`, then it should be fine if it is in a bucket too...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Create an RDD from the reference sequences.
+     * The reference sequences are transformed into a single, large collection of byte arrays. The collection is then
+     * parallelized into an RDD.
+     * Each contig that exceeds a size given by REF_RECORD_LEN is broken into a series of REF_RECORD_LEN chunks with a
+     * K-1 base overlap between successive chunks. (I.e., for K=63, the last 62 bases in chunk n match the first 62
+     * bases in chunk n+1) so that we don't miss any kmers due to the chunking -- we can just kmerize each record
+     * independently.
+     */
+    public static JavaRDD<byte[]> getRefRDD(final JavaSparkContext ctx,
```
magicDGS: Maybe this could be moved to `SparkUtils`... may be useful outside the SV code...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SvCigarUtils.java
```diff
@@ -377,4 +377,11 @@ public static int computeAssociatedDistOnRead(final Cigar cigarAlong5To3DirOfRea
 
         return readWalkDist;
     }
+
+    public static int getTotalAlignmentLength(final Cigar cigar ) {
```
magicDGS: Maybe is it useful in `CigarUtils` to re-use outside SV code? I do not see specificity with SV here...

SHuang-Broad:
@vruano would you mind taking a look?

SHuang-Broad:
@magicDGS , I'll address your comments in the next commit. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
```diff
@@ -0,0 +1,116 @@
...
+
+        return ctx.parallelize(sequenceChunks, sequenceChunks.size()/ref_records_per_partition+1);
+    }
+
+    /**
+     * Do a map/reduce on an RDD of genomic sequences:
+     * Kmerize, mapping to a pair <kmer,1>, reduce by summing values by key, filter out <kmer,N> where
+     * N <= MAX_KMER_FREQ, and collect the high frequency kmers back in the driver.
+     */
+    @VisibleForTesting
```
vruano: This method is not just public visible for testing; is referenced from the main code from a different package so please remove this annotation.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
```diff
@@ -0,0 +1,116 @@
...
+    public static List<SVKmer> processRefRDD(final int kSize,
+                                             final int maxDUSTScore,
+                                             final int maxKmerFreq,
+                                             final JavaRDD<byte[]> refRDD) {
+        final int nPartitions = refRDD.getNumPartitions();
```
vruano: Check arguments are not null or otherwise invalid.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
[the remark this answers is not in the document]
vruano: Check arguments are not null or otherwise invalid.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
```diff
@@ -0,0 +1,116 @@
...
+        return ctx.parallelize(sequenceChunks, sequenceChunks.size()/ref_records_per_partition+1);
+    }
+
+    /**
+     * Do a map/reduce on an RDD of genomic sequences:
+     * Kmerize, mapping to a pair <kmer,1>, reduce by summing values by key, filter out <kmer,N> where
+     * N <= MAX_KMER_FREQ, and collect the high frequency kmers back in the driver.
+     */
+    @VisibleForTesting
+    public static List<SVKmer> processRefRDD(final int kSize,
```
vruano: I guess the name is again not informative enough...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
vruano: I think that getRefRDD and processRefRDD are doing a very specific kind of work on the reference genome and so I don't think they would qualify as utilities ... they seem to be only used by FindBadGenomeKmers, so I would suggest that these methods are moved to that class.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVUtils.java
```diff
@@ -165,6 +39,13 @@ public static int iteratorSize( final LongIterator itr ) {
         return Collections.singletonList(t).iterator();
     }
 
+    public static Collection<SVKmer> uniquify(final Collection<SVKmer> coll1, final Collection<SVKmer> coll2) {
+        final HopscotchSet<SVKmer> kmers = new HopscotchSet<>(coll1.size() + coll2.size());
```
vruano: Without looking at the uses... I guess you again simply added coll2 elements into coll1 and return coll1.

vruano: Since is public you must check arguments are legal (non null).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVUtils.java
```diff
@@ -207,6 +88,11 @@ private void advance() {
         return Collectors.toCollection( () -> new ArrayList<>(size));
     }
 
+    // =================================================================================================================
+
+    //Workaround for seed 14695981039346656037 that doesn't fit in a signed long
+    private static final long FNV64_DEFAULT_SEED = new BigInteger("14695981039346656037").longValue();
```
vruano: Curious what is the need to use a "problematic" seed number.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SvCigarUtils.java
vruano: I agree with @magicDGS, you can call it something like ```countAlignedBases```.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
```diff
@@ -0,0 +1,116 @@
...
+     * independently.
+     */
+    public static JavaRDD<byte[]> getRefRDD(final JavaSparkContext ctx,
+                                            final int kSize,
+                                            final ReferenceMultiSource ref,
+                                            final SAMSequenceDictionary readsDict,
+                                            final int ref_record_len,
+                                            final int ref_records_per_partition) {
+        final SAMSequenceDictionary dict = ref.getReferenceSequenceDictionary(readsDict);
+        if ( dict == null ) throw new GATKException("No reference dictionary available");
```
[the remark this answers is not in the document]
vruano: Does any subtype of reference source every going to return a null without failing first?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -161,7 +162,7 @@ protected void runTool( final JavaSparkContext ctx ) {
                                                                                    final Logger toolLogger) {
 
             final SAMFileHeader cleanHeader = new SAMFileHeader(header.getSequenceDictionary());
-            final List<String> refNames = AlignedAssemblyOrExcuse.getRefNames(header);
+            final List<String> refNames = SVReferenceUtils.getRefNames(header);
```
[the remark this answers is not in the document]

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/AlignedAssemblyOrExcuse.java
```diff
@@ -355,4 +241,97 @@ public AlignedAssemblyOrExcuse read( final Kryo kryo, final Input input, final C
             return new AlignedAssemblyOrExcuse(kryo, input);
         }
     }
+
+    public static Stream<SAMRecord> toSAMStreamForOneContig( final SAMFileHeader header, final List<String> refNames,
```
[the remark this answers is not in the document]
SHuang-Broad: done in a way mostly as suggested, except that kept the required argument of `refNames` for performance concerns (so we don't go through the streaming of getting the ref contig names for every read/assembly-contig).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/AlignedAssemblyOrExcuse.java
```diff
@@ -355,4 +241,97 @@ public AlignedAssemblyOrExcuse read( final Kryo kryo, final Input input, final C
...
+    }
+
+    /**
+     * write a file describing each interval
+     */
+    public static void writeIntervalFile( final String intervalFile,
+                                          final SAMFileHeader header,
+                                          final List<SVInterval> intervals,
+                                          final List<AlignedAssemblyOrExcuse> intervalDispositions ) {
+        final Map<Integer, AlignedAssemblyOrExcuse> resultsMap = new HashMap<>();
```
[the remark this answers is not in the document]
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/AlignedAssemblyOrExcuse.java
```diff
@@ -355,4 +241,97 @@ public AlignedAssemblyOrExcuse read( final Kryo kryo, final Input input, final C
...
+        intervalDispositions.forEach(alignedAssemblyOrExcuse ->
+                resultsMap.put(alignedAssemblyOrExcuse.getAssemblyId(), alignedAssemblyOrExcuse));
+
+        try ( final OutputStreamWriter writer =
+                      new OutputStreamWriter(new BufferedOutputStream(BucketUtils.createFile(intervalFile))) ) {
+            final List<SAMSequenceRecord> contigs = header.getSequenceDictionary().getSequences();
+            final int nIntervals = intervals.size();
+            for ( int intervalId = 0; intervalId != nIntervals; ++intervalId ) {
```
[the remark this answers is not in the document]
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/AlignedAssemblyOrExcuse.java
```diff
@@ -355,4 +241,97 @@ public AlignedAssemblyOrExcuse read( final Kryo kryo, final Input input, final C
...
+                final SVInterval interval = intervals.get(intervalId);
```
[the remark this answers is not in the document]
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/AlignedAssemblyOrExcuse.java
```diff
@@ -355,4 +241,97 @@ public AlignedAssemblyOrExcuse read( final Kryo kryo, final Input input, final C
...
+    }
+
+    /**
+     * write a SAM file containing records for each aligned contig
+     */
+    static void writeAssemblySAMFile( final List<AlignedAssemblyOrExcuse> alignedAssemblyOrExcuseList,
+                                      final SAMFileHeader header,
+                                      final SAMFileHeader.SortOrder assemblyAlnSortOrder,
+                                      final String outputAssembliesFile ) {
+        final SAMFileHeader cleanHeader = new SAMFileHeader(header.getSequenceDictionary());
```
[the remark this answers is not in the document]
SHuang-Broad: The header is requested several lines down.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/AlignedAssemblyOrExcuse.java
```diff
@@ -355,4 +241,97 @@ public AlignedAssemblyOrExcuse read( final Kryo kryo, final Input input, final C
...
+                } else {
+                    disposition = "produced " + alignedAssemblyOrExcuse.getAssembly().getNContigs() +
+                            " contigs in " + alignedAssemblyOrExcuse.getSecondsInAssembly() + " secs.";
+                }
+                writer.write(intervalId + "\t" +
+                        seqName + ":" + interval.getStart() + "-" + interval.getEnd() + "\t" +
+                        disposition + "\n");
+            }
+        } catch ( final IOException ioe ) {
+            throw new GATKException("Can't write intervals file " + intervalFile, ioe);
```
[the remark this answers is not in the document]
SHuang-Broad: No. This output file is for debugging or tuning only so user doesn't directly control that. So no action.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -141,13 +140,16 @@ public static AssembledEvidenceResults gatherEvidenceAndWriteContigSamFile(
...
-        // write the output file
-        final SAMFileHeader cleanHeader = new SAMFileHeader(header.getSequenceDictionary());
-        cleanHeader.setSortOrder(params.assembliesSortOrder);
-
-        AlignedAssemblyOrExcuse.writeSAMFile(outputAssembliesFile, cleanHeader, alignedAssemblyOrExcuseList,
-                params.assembliesSortOrder == SAMFileHeader.SortOrder.queryname);
+        // write alignments of the assembled contigs
+        AlignedAssemblyOrExcuse.writeAssemblySAMFile(alignedAssemblyOrExcuseList, header, params.assembliesSortOrder, outputAssemblyAlignments);
         log("Wrote SAM file of aligned contigs.", toolLogger);
+//        // write the output file
```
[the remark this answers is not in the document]
SHuang-Broad: :scream_cat: removed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/KmerCounter.java
```diff
@@ -25,7 +26,7 @@ public KmerCounter( final int kSize, final int kmersPerPartitionGuess,
         this.kmersPerPartitionGuess = kmersPerPartitionGuess;
     }
 
-    public Iterator<KmerAndCount> apply( final Iterator<GATKRead> readItr ) {
+    public Iterator<KmerAndCount> apply(final Iterator<GATKRead> readItr ) {
```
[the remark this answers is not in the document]
SHuang-Broad: added, as that seems to be TedS' style, who is the author as well.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -451,8 +453,8 @@ public EvidenceScanResults(final ReadMetadata readMetadata,
                         new ReadsForQNamesFinder(broadcastQNamesMultiMap.value(), nIntervals,
                                 includeMappingLocation, readItr, filter).iterator(), false)
                 .combineByKey(x -> x,
```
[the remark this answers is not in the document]
SHuang-Broad: I refactored/moved the `combineLists` from FindBreakpointEvidence to SVFastqUtils, so would rather let Ted handle this as he has many acrobatic tricks on how to avoid shuffles, and I refrain from messing with that.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFastqUtils.java
```diff
@@ -48,6 +48,15 @@
 
     private static final Pattern SA_TAG_ALN_INTERVAL_SEPARATOR_PATTERN = Pattern.compile(HEADER_FIELD_LIST_SEPARATOR_STR);
 
+    /** Concatenate two lists. */
```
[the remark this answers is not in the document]
SHuang-Broad: Again this is a simple move from another class, so would not mess with it right now.
Also, like you said this is not necessarily for Spark related jobs. So not messing with it yet in this PR.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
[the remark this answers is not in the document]
SHuang-Broad: What would you suggest for the name?
I am creating a dir to write some (SAM) files in a prototyping tool.

SHuang-Broad: Can you elaborate, specifically what refactoring is needed?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+public final class SVReferenceUtils {
+
+    public static final int REF_RECORD_LEN = 10000;
+    // assuming we have ~1Gb/core, we can process ~1M kmers per partition
+    public static final int REF_RECORDS_PER_PARTITION = 1024*1024 / REF_RECORD_LEN;
+
+    public static List<String> getRefNames(final SAMFileHeader header) {
```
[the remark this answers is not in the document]
SHuang-Broad: done in its new place

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVUtils.java
SHuang-Broad: done checking. If second is added to 1st, the 1st is modified, unlike what it is doing now (copying from the 1st).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVUtils.java
SHuang-Broad: not sure. That's a question for TedS.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SvCigarUtils.java
SHuang-Broad: done

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/integration/ExtractOriginalAlignmentRecordsByNameSparkIntegrationTest.java
```diff
@@ -4,28 +4,29 @@
...
 public final class ExtractOriginalAlignmentRecordsByNameSparkIntegrationTest extends CommandLineProgramTest {
 
     @Test(groups = "sv")
-    public void testExtractOriginalAlignmentRecordsByNameSparkRunnableLocal() throws IOException {
+    public void testExtractOriginalAlignmentRecordsByNameSparkRunnableLocal() {
 
         final File tempWorkingDir = BaseTest.createTempDir("extractOriginalAlignmentRecordsByNameSparkIntegrationTest");
 
-        FileUtils.writeLinesToSingleFile(Collections.singleton("asm013903:tig00002").iterator(), tempWorkingDir + "/names.txt");
+        SVFileUtils.writeLinesToSingleFile(Collections.singleton("asm013903:tig00002").iterator(), tempWorkingDir + "/names.txt");
```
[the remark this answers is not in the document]
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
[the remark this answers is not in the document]
SHuang-Broad: done

SHuang-Broad: This method is not really tested well I am hesitated to move it there. Beside I am not the author.
What usage do you have in mind?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
```diff
@@ -0,0 +1,116 @@
...
+     * Each contig that exceeds a size given by REF_RECORD_LEN is broken into a series of REF_RECORD_LEN chunks with a
+     * K-1 base overlap between successive chunks. (I.e., for K=63, the last 62 bases in chunk n match the first 62
+     * bases in chunk n+1) so that we don't miss any kmers due to the chunking -- we can just kmerize each record
+     * independently.
+     */
+    public static JavaRDD<byte[]> getRefRDD(final JavaSparkContext ctx,
+                                            final int kSize,
+                                            final ReferenceMultiSource ref,
+                                            final SAMSequenceDictionary readsDict,
+                                            final int ref_record_len,
```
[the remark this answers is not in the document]
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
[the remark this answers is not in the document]
vruano: ```createDirectory```? I just don't why would you create a directory if you are not going to write into it, you or something else later on so that part of the name is rather pointless.... That said it seems this method would better belong to BucketUtils pairing with ```createFile``` as I say below.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
```diff
@@ -0,0 +1,116 @@
...
+                                            final int ref_records_per_partition) {
+        final SAMSequenceDictionary dict = ref.getReferenceSequenceDictionary(readsDict);
```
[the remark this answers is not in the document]
SHuang-Broad: moved check over there

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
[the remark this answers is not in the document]
SHuang-Broad: not sure. This check was there before the refactoring.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
SHuang-Broad: removed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
SHuang-Broad: renamed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
SHuang-Broad: checked.

SHuang-Broad: `getRefRDD` was in `SVUtils` before so I moved it here thinking it is providing a special type of utility related to reference. So I'll leave it here.
Moved processRefRDD back to `FindBadGenomeKmers`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+        catch ( final IOException ioe ) {
+            throw new GATKException("Unable to read kmers from "+kmersFile, ioe);
+        }
+
+        return kmers;
+    }
+
+    /** Write kmers to file. */
+    public static <KType extends SVKmer> void writeKmersFile(final int kSize, final String kmersFile,
```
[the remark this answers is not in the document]
SHuang-Broad: added suffix "*Path"

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+                writer.write('\n');
+            }
+        }
+        catch ( final IOException ioe ) {
+            throw new GATKException("Unable to write kmers to "+kmersFile, ioe);
+        }
+    }
+
+    /** Read intervals from file. */
+    public static List<SVInterval> readIntervalsFile(final String intervalsFile,
```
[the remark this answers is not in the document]
SHuang-Broad: added suffix "*Path"

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
+        catch ( final IOException ioe ) {
+            throw new GATKException("Unable to read intervals from "+intervalsFile, ioe);
+        }
+        return intervals;
+    }
+
+    /** Write intervals to a file. */
+    public static void writeIntervalsFile( final String intervalsFile,
```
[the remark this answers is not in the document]
SHuang-Broad: added suffix "*Path"

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+
+    /** Read intervals from file. */
+    public static List<SVInterval> readIntervalsFile(final String intervalsFile,
+                                                     final Map<String, Integer> contigNameMap ) {
+        final List<SVInterval> intervals;
+        try ( final BufferedReader rdr =
+                      new BufferedReader(new InputStreamReader(BucketUtils.openFile(intervalsFile))) ) {
+            final int INTERVAL_FILE_LINE_LENGTH_GUESS = 25;
```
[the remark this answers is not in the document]
SHuang-Broad: local enough and used only once, so replace with commented number.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+            final long fileLength = BucketUtils.fileSize(kmersFile);
+            kmers = new HopscotchSet<>((int)(fileLength/(kSize+1)));
+            String line;
+            while ( (line = rdr.readLine()) != null ) {
+                if ( line.length() != kSize ) {
+                    throw new GATKException("SVKmer kill set contains a line of length " + line.length() +
+                            " but we were expecting K=" + kSize);
+                }
+
+                final SVKmerizer kmerizer = new SVKmerizer(line, kSize, 1, new SVKmerLong(kSize));
```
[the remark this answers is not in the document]
SHuang-Broad: Not really my code initially. So I'll leave it to the next round of clean up.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /** Write kmers to file. */
+    public static <KType extends SVKmer> void writeKmersFile(final int kSize, final String kmersFile,
+                                                             final Collection<KType> kmers ) {
+        try ( final Writer writer =
+                      new BufferedWriter(new OutputStreamWriter(BucketUtils.createFile(kmersFile))) ) {
+            for ( final KType kmer : kmers ) {
+                writer.write(kmer.toString(kSize));
+                writer.write('\n');
```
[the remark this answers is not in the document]
SHuang-Broad: available only to BufferedWriter but not Writer. BufferedWriter has `write(String, int, int )` but no `write(String)`, so no action considering trade-off.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+        final int idx = outputName.lastIndexOf(".");
+        if ( idx < 0) {
+            throw new IllegalArgumentException("Provided path doesn't have a proper extension: " + outputName);
+        }
+        final String fileExtension = outputName.substring(idx+1, outputName.length());
+        final SAMFileWriterFactory factory = new SAMFileWriterFactory()
+                .setCreateIndex(preOrdered && header.getSortOrder()== SAMFileHeader.SortOrder.coordinate);
+        final OutputStream outputStream = BucketUtils.createFile(outputName);
+
+        if (fileExtension.endsWith(SamReader.Type.BAM_TYPE.fileExtension())) {
```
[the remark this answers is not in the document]
SHuang-Broad: extension check pushed up.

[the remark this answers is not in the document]
SHuang-Broad: truth be told, the code that I moved and refactored from does this file creation. 
And I guess the `SAMFileWriterFactory` version doesn't take hdfs path, which this BucketUtils version does allow.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+            }
+            return isSuccessful;
+        } catch (final IOException x) {
+            throw new GATKException("Could not create directory: " + pathString, x);
+        }
+    }
+
+    public static void writeLinesToSingleFile(final Iterator<String> linesToWrite, final String fileName) {
+        try ( final OutputStream writer =
+                      new BufferedOutputStream(BucketUtils.createFile(fileName)) ) {
```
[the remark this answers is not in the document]
SHuang-Broad: changed to 
```java
final Writer writer =
   new BufferedWriter(new OutputStreamWriter(BucketUtils.createFile(fileName))
```
which doesn't have `newLine()` but does have `write(String)`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+            while (linesToWrite.hasNext()) {
+                writer.write(linesToWrite.next().getBytes()); writer.write('\n');
+            }
+        } catch ( final IOException ioe ) {
+            throw new GATKException("Can't write "+fileName, ioe);
+        }
+    }
+
+    public static void writeSAMFile(final Iterator<SAMRecord> alignments, final SAMFileHeader header, final String outputName,
```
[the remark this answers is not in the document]
SHuang-Broad: checking done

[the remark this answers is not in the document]
SHuang-Broad: done reordering parameters

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+                throw new IOException("Directory to be created already exists: " + pathString);
+
+            final boolean isSuccessful;
+            if (BucketUtils.isHadoopUrl(pathString)) {
+                isSuccessful = org.apache.hadoop.fs.FileSystem.get(new org.apache.hadoop.conf.Configuration()).mkdirs(new org.apache.hadoop.fs.Path(pathString));
+            } else {
+                final java.nio.file.Path dir = java.nio.file.Files.createDirectory(IOUtils.getPath(pathString));
+                isSuccessful = java.nio.file.Files.isDirectory(dir) && java.nio.file.Files.isWritable(dir);
+            }
+            return isSuccessful;
```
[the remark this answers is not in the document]
SHuang-Broad: changed to consistent behavior of returning boolean

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
magicDGS: I think that it would work by using the following code (but I haven't test it):

```java
// this should take care of GCS
final Path dir = IOUtils.getPath(Utils.nonNull(pathString));
if (Files.exists(dir)) {
    throw new IOException("Directory to be created already exists: " + dir.toUri());
}
// maybe some exceptions should be catch to return false (I am not sure about the usage of the boolean)
Files.createDirectory(dir);
return Files.isDirectory(dir) & Files.isWritable(dir);
```

In that case, it belongs to `IOUtils`, because it can be re-used easily...

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
magicDGS: I still thing that the usage of this method can be substituted by `Files.write(IOUtils.getPath(fileName), linesToWrite)`; there are also utils already implemented to convert between iterable, iterators and stream (see `org.broadinstitute.hellbender.utils.Utils`).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
magicDGS: I have an usage case where I need a tool iterating over kmers in the reference: although still in development, I have a reference walker by kmers, and at some point it would be nice to have also a spark-based reference walker.

Anyway, it was just a suggestion. For me, it is just simpler to look for useful methods in the utils classes related to the stuff I'm looking for (e.g., `CigarUtils` for cigar, `SparkUtils`for RDD related stuff, etc.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
SHuang-Broad: @magicDGS Thanks! I've replaced this method with the one you suggested and it is working well.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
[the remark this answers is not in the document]
SHuang-Broad: lied. Changed to throwing IOException and no return.

SHuang-Broad:
@vruano and @magicDGS  I made most of the changes you requested, with the following exceptions:

* methods that were complex enough and I am not the original author&mdash;I didn't change most of them except adding some arg checking
* `SVFileUtils.createDirectory()` not sure if the change is what you would like to see (most likely still needs some work) and it can be moved to `BucketUtils` (I need to add tests for it before that)
* `SVFileUtils.writeSAMFile()` maybe it is worth to move it to some engine utils, but I need to talk to some Engine guy first.

SHuang-Broad:
Regarding writing the SAM file, I now remember why it was created instead of using existing ones:  the SV pipeline needs to be able to write to HDFS, hence cannot rely on a `File` interface&mdash;which the `ReadUtils.createCommonSAMWriter(...)` exposes. 
So the current hand rolled version in `SVFileUtils.getSAMFileWriter()` calls into `BucketUtils.createFile(...)` for that HDFS compatibility, and then makes use of the `SAMFileWriterFactory.makeBAMWriter(final SAMFileHeader header, final boolean presorted, final OutputStream stream)`, unlike `ReadUtils.createCommonSAMWriter(...)` which calls into `SAMFileWriterFactory.makeSAMOrBAMWriter(final SAMFileHeader header, final boolean presorted, final Path outputPath)`.

So in short: HDFS compatibility.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
SHuang-Broad: I've done mostly what you suggested, can you take a look again?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
SHuang-Broad: @magicDGS in that case (considering your tool is still under dev), I intend to keep it here and get more test coverage when you think you are ready. Is that OK?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFastqUtils.java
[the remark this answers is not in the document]
SHuang-Broad: after 2nd thought, refactored into SVUtils.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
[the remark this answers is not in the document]
SHuang-Broad: renamed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBadGenomicKmersSpark.java
```diff
@@ -77,11 +84,63 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+        Utils.validateArg(maxDUSTScore > 0, "provided DUST filter score is non positive");
+        Utils.validateArg(maxKmerFreq > 0, "provided kmer frequency is non positive");
+
+        final int nPartitions = refRDD.getNumPartitions();
+        final int hashSize = 2*REF_RECORDS_PER_PARTITION;
+        return refRDD
+                .mapPartitions(seqItr -> {
+                    final HopscotchMap<SVKmer, Integer, KmerAndCount> kmerCounts = new HopscotchMap<>(hashSize);
+                    while ( seqItr.hasNext() ) {
+                        final byte[] seq = seqItr.next();
```
vruano: What about:
```
Utils.stream(seqItr)
       .flatMap(seq -> SVDUSTFilteredKmerizer.canonicalStream(seq, kSize, maxDUSTScore, new SVKmerLong())
       .forEach(kmer -> {...});
```

If one defines a ```Hopscotch.bumpAll(Hopscotch<K, V, T> other)``` that would add the counts from the other hop-scotch then one could use the ```Stream.reduce(Unit, Add, Combine)``` method instead than the ```forEach```.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBadGenomicKmersSpark.java
```diff
@@ -77,11 +84,63 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+                                    final KmerAndCount entry = kmerCounts.find(kmer);
+                                    if ( entry == null ) kmerCounts.add(new KmerAndCount((SVKmerLong)kmer));
+                                    else entry.bumpCount();
+                                });
+                    }
+                    return kmerCounts.iterator();
+                })
+                .mapToPair(entry -> new Tuple2<>(entry.getKey(), entry.getValue()))
+                .partitionBy(new HashPartitioner(nPartitions))
+                .mapPartitions(pairItr -> {
```
vruano: You could do something similar here.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/KmerAndCount.java
```diff
@@ -18,7 +18,7 @@
 
     public KmerAndCount( final SVKmerLong kmer ) { this(kmer,1); }
 
-    KmerAndCount( final SVKmerLong kmer, final int count ) {
+    public KmerAndCount( final SVKmerLong kmer, final int count ) {
```
vruano: Since is public, must it check that the ```kmer``` is not null and ```count``` is 0 or positive?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -108,35 +98,37 @@ private static SAMFileWriter getSAMFileWriter(final String outputName, final SAM
             }
         }
         catch ( final IOException ioe ) {
-            throw new GATKException("Unable to read kmers from "+kmersFile, ioe);
+            throw new GATKException("Unable to read kmers from "+kmersFilePath, ioe);
```
vruano: Spaces around "+", just bother if you are going to change this file for any other reason , not for this alone.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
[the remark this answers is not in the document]
vruano: Nothing is stopping you from change the type of ```writer``` to ```BufferedWriter```, but fair enough, is not your code.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -108,35 +98,37 @@ private static SAMFileWriter getSAMFileWriter(final String outputName, final SAM
...
-                      new BufferedWriter(new OutputStreamWriter(BucketUtils.createFile(kmersFile))) ) {
+                      new BufferedWriter(new OutputStreamWriter(BucketUtils.createFile(kmersFilePath))) ) {
             for ( final KType kmer : kmers ) {
                 writer.write(kmer.toString(kSize));
                 writer.write('\n');
             }
         }
         catch ( final IOException ioe ) {
-            throw new GATKException("Unable to write kmers to "+kmersFile, ioe);
+            throw new GATKException("Unable to write kmers to "+kmersFilePath, ioe);
```
vruano: same with the "+"

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -147,7 +139,7 @@ private static SAMFileWriter getSAMFileWriter(final String outputName, final SAM
                 }
                 final String[] tokens = line.split("\t");
                 if ( tokens.length != 3 ) {
-                    throw new GATKException("Interval file "+intervalsFile+" line "+
+                    throw new GATKException("Interval file "+intervalsFilePath+" line "+
```
vruano: Same.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -158,27 +150,27 @@ private static SAMFileWriter getSAMFileWriter(final String outputName, final SAM
                     intervals.add(new SVInterval(contigId, start, end));
                 }
                 catch ( final Exception e ) {
-                    throw new GATKException("Unable to parse interval file "+intervalsFile+" line "+lineNo+": "+line, e);
```
vruano: Won't repeat any more.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
```diff
@@ -39,16 +28,18 @@
...
-                                            final int kSize,
-                                            final ReferenceMultiSource ref,
-                                            final SAMSequenceDictionary readsDict,
-                                            final int ref_record_len,
-                                            final int ref_records_per_partition) {
-        final SAMSequenceDictionary dict = ref.getReferenceSequenceDictionary(readsDict);
-        if ( dict == null ) throw new GATKException("No reference dictionary available");
-
-        final int effectiveRecLen = ref_record_len - kSize + 1;
+    public static JavaRDD<byte[]> getReferenceBasesRDD(final JavaSparkContext ctx,
```
vruano: I still think this is not general enough to be a utility.... where do you think is it going to use? perhaps it could be made a bit more general if instead of returning chunks of arbitrary length it returns well defined kmer so you would hide the intermediary chanck size.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVUtils.java
vruano: My understanding is that you don't want to make the change because you think it will break it? I don't think it would but I guess I have to find the reference supporting my claim.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
```diff
@@ -130,4 +131,22 @@ public static String asTag( final BwaMemAlignment alignment, final List<String>
                 (SAMFlag.READ_REVERSE_STRAND.isSet(alignment.getSamFlag())?"-":"+")+","+
                 alignment.getCigar().replace('H','S')+","+alignment.getMapQual()+","+alignment.getNMismatches()+";";
     }
+
+    public static Stream<SAMRecord> toSAMStreamForRead(final String readName, final byte[] contigSequence,
+                                                       final List<BwaMemAlignment> alignments,
+                                                       final SAMFileHeader header, final List<String> refNames) {
+        if ( alignments.isEmpty() ) return Stream.empty();
```
vruano: Check arguments.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private static final String REFERENCE_GAP_INTERVAL_FILE_COMMENT_LINE_PROMPT = "#";
+
+    /**
+     * Creates a directory, in local FS, HDFS, or Google buckets to write individual files in.
+     */
+    public static void createDirectory(final String pathString) throws IOException {
+        Utils.nonNull(pathString);
+
+        if (BucketUtils.isHadoopUrl(pathString)) {
```
magicDGS: I still cannot understand why a Hadoop path should be treated different than another kind of path. I believe that the `IOUtils.getPath` will get properly the Hadoop path, and then `Files.createDirectory` should get the `HadoopFileSystemProvider` for the HDFS, which in turn uses the mkdirs that you are using here.

Sorry if I am missing something....

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -0,0 +1,176 @@
...
+        Utils.nonNull(outputName, "provided output name is null");
+
+        try (final SAMFileWriter writer = getSAMFileWriter(outputName, header, preOrdered)) {
+            alignments.forEachRemaining(writer::addAlignment);
+        } catch ( final UncheckedIOException ie) {
+            throw new GATKException("Can't write SAM file to the specified location: " + outputName, ie);
+        }
+    }
+
+    private static SAMFileWriter getSAMFileWriter(final String outputName, final SAMFileHeader header, final boolean preOrdered) {
```
magicDGS: The `ReadUtils` class will have support for HDFS and other `java.nio.Path` representation here: https://github.com/broadinstitute/gatk/pull/2558

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
magicDGS: It's perfectly fine, I'm just worried that I forget that this is here and implement my own method. Usually I look at the `SparkUtils` code for this kind of things, and I do not use to look at the SV code at all except in some PRs with functionality that I think that I can use as a third-party developer...

SHuang-Broad: @magicDGS . In that case, I'll try to add tests and move it to SparkUtils. In the meantime, if you find other methods in `SVUtils` and `SvCigarUtils` that are useful to you, let me know.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVUtils.java
SHuang-Broad: Not necessarily breaking it, but that would change the behavior (and of course the signature) of the function.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
```diff
@@ -19,6 +19,7 @@
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.spark.GATKSparkTool;
 import org.broadinstitute.hellbender.exceptions.GATKException;
```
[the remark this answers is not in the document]
SHuang-Broad: removed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBadGenomicKmersSpark.java
```diff
@@ -32,6 +36,9 @@
         programGroup = StructuralVariationSparkProgramGroup.class)
 @BetaFeature
 public final class FindBadGenomicKmersSpark extends GATKSparkTool {
+    public static final int REF_RECORD_LEN = 10000;
```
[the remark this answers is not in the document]
SHuang-Broad: I agree, but since this is the code already in place in master, and after I moved back the intended-to-be-refactored out method back here, it is not used by any other code, I'll say let's keep it as is.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/AlignedAssemblyOrExcuse.java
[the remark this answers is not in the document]
SHuang-Broad: You are right. Thanks for catch that!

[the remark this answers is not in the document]
SHuang-Broad: also changed the order of parameters so the 1st parameter is the file name as you suggested somewhere else.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBadGenomicKmersSpark.java
```diff
@@ -32,6 +36,9 @@
...
+    // assuming we have ~1Gb/core, we can process ~1M kmers per partition
+    public static final int REF_RECORDS_PER_PARTITION = 1024*1024 / REF_RECORD_LEN;
     private static final long serialVersionUID = 1L;
     @VisibleForTesting static final int MAX_KMER_FREQ = 3;
```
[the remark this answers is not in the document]
SHuang-Broad: changed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBadGenomicKmersSpark.java
```diff
@@ -77,11 +84,63 @@ protected void runTool( final JavaSparkContext ctx ) {
...
-        return SVReferenceUtils.processRefRDD(kSize, maxDUSTScore, MAX_KMER_FREQ, refRDD);
+        return collectUbiquitousKmersInReference(kSize, maxDUSTScore, MAX_KMER_FREQ, refRDD);
+    }
+
+    /**
+     * Do a map/reduce on an RDD of genomic sequences:
+     * Kmerize, mapping to a pair <kmer,1>, reduce by summing values by key, filter out <kmer,N> where
+     * N <= MAX_KMER_FREQ, and collect the high frequency kmers back in the driver.
+     */
+    public static List<SVKmer> collectUbiquitousKmersInReference(final int kSize,
```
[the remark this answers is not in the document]
SHuang-Broad: correct. Changed.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBadGenomicKmersSpark.java
SHuang-Broad: not going to change implementations in this PR, since this method is not being refactored into Utils any more. Trying to limit what this PR does.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/KmerAndCount.java
SHuang-Broad: added a check for non-negative number.
But for the null check, I don't quite understand yet what the benefit is for this specific one. It delegates to the parent class's ctor, which would immediately throw an NPE if it null. An explicit null check would pretty much to the same thing, right?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -15,42 +15,31 @@
...
-
-    public static void writeLinesToSingleFile(final Iterator<String> linesToWrite, final String fileName) {
-        try ( final OutputStream writer =
-                      new BufferedOutputStream(BucketUtils.createFile(fileName)) ) {
-            while (linesToWrite.hasNext()) {
-                writer.write(linesToWrite.next().getBytes()); writer.write('\n');
-            }
-        } catch ( final IOException ioe ) {
-            throw new GATKException("Can't write "+fileName, ioe);
+    public static void createDirectory(final String pathString) throws IOException {
```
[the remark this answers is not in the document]
SHuang-Broad: Yes. After the rename and rewrite I agree. Will move it there.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
SHuang-Broad: @magicDGS thinks he would have a use for that. 
I think this could be useful as well. 
I'll what improvements (test, representation) I can bring to this.

src/main/java/org/broadinstitute/hellbender/utils/bwa/BwaMemAlignmentUtils.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/utils/SequenceDictionaryUtils.java
```diff
@@ -466,6 +466,11 @@ private static boolean commonContigsAreAtSameIndices( final Set<String> commonCo
         return contigNames;
     }
 
+    public static List<String> getContigNamesList(final SAMSequenceDictionary refSeqDict) {
```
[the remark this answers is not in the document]
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
SHuang-Broad: done what's suggested here, moved to IOUtils and tested

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/KmerAndCount.java
vruano: As per GATK3  GATK4 policy NPE are not acceptable; you can argue that with @droazen but I think is a lost cause, right David?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
vruano: What most using code would find useful is ```JavaRDD<byte[]>``` of all the khmers of a given length rather than a collection of chunks of bases or different sizes.... I guess the reason for returning the later is because to create a JavaRDD it seems that you need to explicitly create a List of all the its elements (is not possible to provide a stream instead?) (so ```ctx.parallelize(kmerList)```... this obviously not practical for a large genome like the human genome. 

I think the best way to approach this by changing the actual RDD that is returned. 
Instead of returning the RDD of the DNA chunks as we do now we return a ```flatMap``` on it that would transform those chunks into the actual kmers so these are only instantiated when they are needed... then the parameters of this method except for the reference source and the kmerSize can be hidden given them reasonable defaults.

Perhaps instead of byte[] you could return SVKmers objects.

I leave it up to you but the chuck-of-DNA RDD return looks a bit nasty and complex as a return type and given its specific about k-1 overlapping ... it is obvious what the intention for it so why don't simply return a RDD to the kmers that is what people really want.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
vruano: @magicDGS, in theory you are right but in practice there seems to be some issues (see https://github.com/broadinstitute/gatk/issues/3569 https://github.com/broadinstitute/barclay/issues/106)
Perhaps now we could drop the handoop but as far as this PR is concern it should follow the current approach.

src/main/java/org/broadinstitute/hellbender/utils/read/CigarUtils.java
```diff
@@ -482,4 +482,11 @@ public static int countRightClippedBases(final Cigar cigar) {
             throw new IllegalArgumentException("the input cigar only have clipping operations");
         }
     }
+
+    public static int countAlignedBases(final Cigar cigar ) {
```
vruano: Add unit test.

vruano: Test Cigar is not Null with ```Utils.nonNull```

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
magicDGS: Actually, the problem is mainly with the GCS provider as far as I know (If I am wrong, maybe @lbergelson or @droazen can chime in for correct me).

The discussion in barclay is pointing to the method in `IOUtils`, as the hack for getting correctly the paths, which is working perfectly fine for hadoop (I have been using it in my own code properly).

In the case of this code, if the path is get by `IOUtils.getPath` the providers are loaded on demand (including the special case of GCS, which is handled with `BucketUtils`). Thus, using `Files` is possible once the providers are loaded. Again, as far as I know, the __jsr203-hadoop__ provider is working perfectly fine with `Files` (tested in my own toolkit) and that is the only special case handled here. I still do not fully understand why that's the case, while the GCS code is handled by `Files` in the implementation pointed by your comment. In principle, I think that most of the problems with providers are found with the GCS loading code...but again, @lbergelson or @droazen can correct me if I am missing something...

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFastqUtilsUnitTest.java
```diff
@@ -29,7 +30,7 @@
 /**
  * Unit tests for components in {@link SVFastqUtils}.
  */
-public class SVFastqUtilsUnitTest {
+public class SVFastqUtilsUnitTest extends BaseTest {
```
magicDGS: After https://github.com/broadinstitute/gatk/pull/3475, this should extend `GATKBaseTest`

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVUtilsUnitTest.java
```diff
@@ -4,12 +4,9 @@
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
-/**
- * Unit tests for SVUtils.
- */
 public class SVUtilsUnitTest extends BaseTest {
```
magicDGS: After https://github.com/broadinstitute/gatk/pull/3475, this should extend `GATKBaseTest`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/AlignedAssemblyOrExcuse.java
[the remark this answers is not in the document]
SHuang-Broad: did what you suggested on this round of cleanup.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/KmerAndCount.java
SHuang-Broad: added null check

src/main/java/org/broadinstitute/hellbender/utils/read/CigarUtils.java
SHuang-Broad: done and done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVReferenceUtils.java
SHuang-Broad: I've added a todo for this method, but don't think I have time for it this week or next.
Also cleaned up the documentation a bit.

SHuang-Broad:
@vruano I've addressed more questions. 
@magicDGS sorry but I don't think I have time for the `getRefRDD()` method this or next week. But I've added a todo for it, and will get back on it later after my complex SV work is submitted.

magicDGS:
@SHuang-Broad - thanks for taking it into consideration, even if cannot be soon.

PR #3680, opened 2017-10-10
Removed depth as an input to the allele-subsetting code. 

yfarjoun:
- Unclear what is its purpose.
- added and modified tests.


yfarjoun:
Thanks @davidbenjamin. Tests are failing but I think that they are wrong...Here's the deal. In the tests we have SelectVariants (with --trimAlleles) of a two-sample VCF to NA12891. It includes these two records:

```
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	NA12891	NA12892
20	9999900	.	G	<NON_REF>	.	.	.	GT:DP:GQ:MIN_DP:PL	./.:0:0:0:0,0,0	./.:2:6:2:0,6,69
20	9999996	.	A	ACT,<NON_REF>	.	.	BaseQRankSum=-7.585e+00;DP=154;MQ=60.17;MQRankSum=-7.653e+00;ReadPosRankSum=-6.232e+00	GT:AD:GQ:PL:SAC:SB	./.:0,73,0:99:3396,231,0,3396,231,3396:0,0,30,43,0,0:0,0,30,43	./.:39,40,0:99:1562,0,1880,1680,2000,3680:21,18,17,23,0,0:21,18,17,23
```

The "expected results" have the respective two lines:
```
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	NA12891
20	9999900	.	G	.	.	.	AN=0;DP=0	GT:DP:MIN_DP	./.:0:0
20	9999996	.	A	.	.	.	AN=0;BaseQRankSum=-7.585e+00;DP=154;MQ=60.17;MQRankSum=-7.653e+00;ReadPosRankSum=-6.232e+00	GT:AD:PL:SAC:SB	./.:0:0:0,0:0,0,30,43
```

So somehow, the first line has a non-informative PL omitted from the result, 
while the second line has the non-informative PL included (PL=[0]) in the result. My instinct is that PL=[0] should not appear in the VCF as it is completely uninformative: The length must be 1 since there's only one allele, and a PL or length 1 must be [0]. Do you agree? In that case I can change the expected results and move forward.

yfarjoun:
@vdauwera heads up! :-)

droazen:
This should not be merged before the next release.

yfarjoun:
OK.

yfarjoun:
Also, do we need to bring this up in a methods meeting?

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/AlleleSubsettingUtils.java
```diff
@@ -76,13 +74,12 @@ public static GenotypesContext subsetAlleles(final GenotypesContext originalGs,
                     final int PLindex = MathUtils.maxElementIndex(newLikelihoods);
                     newLog10GQ = GenotypeLikelihoods.getGQLog10FromLikelihoods(PLindex, newLikelihoods);
                 }
-
             }
 
-            final boolean useNewLikelihoods = newLikelihoods != null && (depth != 0 || GATKVariantContextUtils.isInformative(newLikelihoods));
+            final boolean useNewLikelihoods = newLikelihoods != null && newLikelihoods.length > 1;
```
lbergelson: This line was the solution to a big set of differences with gatk3 genotypegvcfs.  If we change it we will no longer match gatk3, which will make it more difficult to validate haplotype caller.  We may want to hold off on any change here until we're finished with that.

droazen: Yeah, this was a hard-fought line of code that we rescued from GATK3 after much agonized searching. I really don't think we can change this now -- I'd move that this PR be closed.

yfarjoun:
@droazen, just to make sure I understand...you are suggesting that we keep doing the wrong things in GATK4 just so the integration tests match their results in GATK3?

Wouldn't it be better to have a GATK3_EQUIVALENCE_MODE flag and create new tests that do the right thing?

droazen:
@yfarjoun I'm suggesting that while we're in the process of tieing out the GATK4 HaplotypeCaller against GATK3, we should not be making changes like this. After the tie-out is complete, then we can of course revisit this branch.

yfarjoun:
Ah, I see. the temporary nature of your suggestion was not clear (to me)

I accept your suggestion.

droazen:
Re-assigning to @jonn-smith and @jamesemery to have a second look at this old PR as part of their HaplotypeCaller work this quarter.

jonn-smith:
Closing PR.  It is quite old and it isn't clear that this is necssary for the current codebase.

PR #3684, opened 2017-10-10
Changed targets argument from optional to required in NormalizeSomaticReadCounts.

### LINKED ISSUE #3017: NormalizeSomaticReadCounts has ``targets`` parameter as optional, but it is required.
@LeeTL1220 commented on [Fri May 26 2017](https://github.com/broadinstitute/gatk-protected/issues/1106)

The code will actually run, but fail right before doing the normalization.

samuelklee:
Closes #3017.


samuelklee:
Actually not an issue, see comment in #3107.

PR #3690, opened 2017-10-11
Added implementation of fragment count collection

asmirnov239:
@mbabadi Could you please review?


src/main/java/org/broadinstitute/hellbender/tools/copynumber/coverage/readcount/ReadCountFileHeaderKey.java
```diff
@@ -0,0 +1,59 @@
+package org.broadinstitute.hellbender.tools.copynumber.coverage.readcount;
```
asmirnov239: Note that this class came from another branch, and is needed because IntegerReadCountFileComparator is dependent on it. It will be used in my future PR

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/CachedBinarySearchIntervalList.java
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Collection of {@link Locatable}s that supports cached binary search of {@link Locatable} objects. In particular,
+ * this class provides an interface to get a range of {@link Locatable}s that intersects with a given {@link Locatable}
+ * query location.
+ *
+ * The search is optimized based on the assumption that the incoming queries of {@link Locatable}s are in sorted order.
+ *
+ * @param <E> a locatable object
+ */
+public class CachedBinarySearchIntervalList<E extends Locatable> {
```
samuelklee: As we discussed, I would be curious what the actual performance hit would be if you simply used an OverlapDetector from htsjdk.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/CollectFragmentCounts.java
```diff
@@ -0,0 +1,223 @@
...
+import org.broadinstitute.hellbender.cmdline.programgroups.CopyNumberProgramGroup;
+import org.broadinstitute.hellbender.engine.FeatureContext;
+import org.broadinstitute.hellbender.engine.ReadWalker;
+import org.broadinstitute.hellbender.engine.ReferenceContext;
+import org.broadinstitute.hellbender.engine.filters.ReadFilter;
+import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
+import org.broadinstitute.hellbender.exceptions.GATKException;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.tools.copynumber.utils.CachedBinarySearchIntervalList;
+import org.broadinstitute.hellbender.tools.exome.SampleCollection;
```
samuelklee: I'd like to remove all dependencies from code in the exome package, if possible.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/CollectFragmentCounts.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    @Override
+    public void onTraversalStart() {
+        if (readArguments.getReadFilesNames().size() != 1) {
+            throw new UserException("This tool only accepts a single bam/sam/cram as input");
+        }
+
+        final SampleCollection sampleCollection = new SampleCollection(getHeaderForReads());
+        if (sampleCollection.sampleCount() > 1) {
```
samuelklee: There is some sample-name code in #3716 that we can use for things like this.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/CollectFragmentCounts.java
```diff
@@ -0,0 +1,223 @@
...
+    public Object onTraversalSuccess() {
+        logger.info("Finished collecting fragment files");
+        logger.info("Writing results to disk");
+        writeResults(outputCountsFile, sampleName);
+
+        logger.info("Fragment counts written to " + outputCountsFile.toString());
+        return "SUCCESS";
+    }
+
+    private void writeResults(final File outputFile, final String sampleName) {
```
samuelklee: @mbabadi Do you need these files to use the old target format?  If not, let's go ahead and switch over to something more reasonable.  If so, let's sequester this code in a copynumber/temporary package.

src/main/java/org/broadinstitute/hellbender/utils/test/IntegerReadCountFileComparator.java
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * Private constructor
+     *
+     * @param file input file
+     * @throws IOException if there is any problems while reading the file
+     * @throws UserException.BadInput if non-integer value is encountered below column names line or file
+     * is not formatted correctly
+     */
+    private IntegerReadCountFileComparator(final File file) throws IOException {
+        Utils.nonNull(file, "the file cannot be null");
```
samuelklee: Can you remove all non-null messages?  I think they are overkill, and it takes effort to make sure they are formatted consistently.

src/main/java/org/broadinstitute/hellbender/utils/test/IntegerReadCountFileComparator.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.IntStream;
+
+/**
+ * This is a testing utility class to check the correctness of coverage collection results for CNV analysis.
+ * It assumes that results come in a TSV format, and checks for exact string matches between two tables. Therefore,
+ * this class is NOT intended to be used with output formats containing floats, as it doesn't allow for rounding error.
+ */
+public class IntegerReadCountFileComparator {
```
samuelklee: I don't know about calling this a Comparator, since it doesn't impose an ordering?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/CollectFragmentCounts.java
```diff
@@ -0,0 +1,223 @@
...
+                        targetFromInterval.getName(),
+                        Integer.toString(intervalCounts[i]));
+                countsFileWriter.println(nextIntervalCountLine);
+            }
+        } catch (IOException exc) {
+            throw new UserException.CouldNotCreateOutputFile(outputFile, exc);
+        }
+    }
+
+    private static String composeHeader(final String commandLine, final String sampleName) {
```
samuelklee: Same thing with this style of header.  The sooner we can move towards consistent file formats, the better.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/CachedBinarySearchIntervalList.java
```diff
@@ -0,0 +1,161 @@
+package org.broadinstitute.hellbender.tools.copynumber.utils;
+
+import htsjdk.samtools.util.Locatable;
+import org.broadinstitute.hellbender.tools.exome.HashedListTargetCollection;
```
samuelklee: You don't use this import (which is good).

src/test/java/org/broadinstitute/hellbender/tools/copynumber/CollectFragmentCountsIntegrationTest.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Integration test for {@link CollectFragmentCounts}.
+ *
+ * @author Andrey Smirnov &lt;asmirnov@broadinstitute.org&gt;
+ */
+public class CollectFragmentCountsIntegrationTest extends CommandLineProgramTest {
+
+    private static final String TEST_SUB_DIR = publicTestDir + "org/broadinstitute/hellbender/tools/copynumber/collectfragmentcounts";
+
+    private static final File NA12878_BAM = new File(TEST_SUB_DIR, "exome-read-counts-NA12878.bam");
```
samuelklee: Let's also be more consistent in naming our test files.  All test files should start with the name of the tool/class, e.g., "collect-fragment-counts-*".  So perhaps something along the lines of "collect-fragment-counts-NA12878-exome.bam".

src/main/java/org/broadinstitute/hellbender/tools/copynumber/CollectFragmentCounts.java
samuelklee: However, note that even that code is a bit of a stopgap---we should eventually do this right with full BAM-style headers.

samuelklee:
@mbabadi @asmirnov239 Alternatively, we could get this PR in quick and I could make additions (e.g., HDF5 writing) to make it suitable to use in the new pipelines, at least for the time being.  We can substitute in the final version of the tool later.

samuelklee:
Added the changes in #3775, let's just go ahead and have a single PR.

PR #3691, opened 2017-10-13
Dump more evidence info for SV pipeline debugging

TedBrookings:
* Added new option --unfilteredBreakpointEvidenceDir
When set, this option dumps all evidence (even evidence that is
ultimately rejected) in an easy to parse text format. Some additional
info (cigarString, mappingQuality) is stored in ReadEvidence to output
information related to read quality.
* Updated option --readMetadata
When set, will additionaly output the map from contig number to contig
name. Added non-null ParitionBounds to readMetadata in
ReadMetadataTest::testEverything to prevent crash.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
@@ -15,6 +15,7 @@ import org.broadinstitute.hellbender.utils.read.GATKRead;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Various types of read anomalies that provide evidence of genomic breakpoints.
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
```diff
@@ -83,6 +84,23 @@ public String toString() {
...
+     * Child classes may extend stringRep() by appending class-specific tab-separated info
+     */
+    public String stringRep(final ReadMetadata readMetadata, final int minEvidenceMapq) {
+        final String dtString =  hasDistalTargets(readMetadata, minEvidenceMapq) ?
+                getDistalTargets(readMetadata, minEvidenceMapq).stream()
+                    .map(strandedInterval -> strandedInterval.getInterval().toString() + (strandedInterval.getStrand() ? "1" : "0"))
+                    .collect(Collectors.joining(";"))
+                : "";
+        return location.toString() + "\t" + weight + "\t" + (validated ? "1" : "0") + "\t" + this.getClass().getSimpleName()
+                + "\t" + dtString;
```
tedsharpe: This will print a trailing tab on evidence that has no distal targets.  Maybe move that tab up to line 96:
"\t" + getDistalTargets...
Yes, I'm being a fussbudget.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
```diff
@@ -329,6 +330,18 @@ public static void writeMetadata(final ReadMetadata readMetadata,
                 writer.write(idx + "\t" + bounds.firstContigID + "\t" + bounds.getFirstStart() + "\t" +
                         bounds.getLastContigID() + "\t" + bounds.getLastStart() + "\n");
             }
+            writer.write("contigs map:\n");
+            readMetadata.contigNameToID
```
tedsharpe: You could just use the contigIDToName array instead of the map.  It's already in the right order.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
TedBrookings: I actually want the extra tab because it makes parsing easy. I split the string by tabs and then the list of distal targets is the 5th word split by ';'. If that's empty there are no distal targets.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
TedBrookings: Ah, I didn't notice that member of ReadMetadata. I'll change it.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
tedsharpe: Fine with me.

TedBrookings:
Made change suggested by Ted Sharpe (write contigIDToName instead of contigNameToID). Will merge after tests pass.

PR #3693, opened 2017-10-15
Adding data.table R package.

samuelklee:
Adding an R dependency for a package that improves performance and memory usage for reading large TSV files.  This is convenient for generating CNV plots from WGS data.

Note: The current CNV plotting code has other issues that make it unsuitable for WGS data; these are addressed in the sl_wgs_acnv dev branch for the new pipeline by new versions of the plotting tools.  However, I do not plan on making these fixes to the old versions of the plotting tools.  The real purpose of this PR is just to get the additional R dependency merged into master so I can build a new docker for the dev branch.


vruano:
Being so trivial why can't this be added in the the dev branch where is needed right now and then it will find its way into master when that one is merged?

samuelklee:
The Docker tests will fail in the dev branch unless the image is updated, which I'd rather do off of master.  Since you agree that this is trivial,  I'm going to go ahead and merge!

vruano:
Sure.

PR #3696, opened 2017-10-16
Add omitFromCommandLine=true for example tools

magicDGS:
Example tools are not documented features, and are provided for being examples for developers to build their own traversals and for testing purposes. Thus, it is unnecessary to show them in the command line help, which already contains a huge list of tools (GATK, Picard, etc).

Although they can still be run from the command line if you know its existence, they won't be shown after #3486.


magicDGS:
Thanks @lbergelson!

PR #3712, opened 2017-10-18
Update data.table R dependency.

samuelklee:
Oops, looks like they just updated the URL last week.  Perhaps another reason why we should host these dependencies or have some simple contingencies for testing them other than manually building the base image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Using OpenJDK 8
-FROM broadinstitute/gatk:gatkbase-1.1
+FROM broadinstitute/gatk:gatkbase-1.2
 
 ADD . /gatk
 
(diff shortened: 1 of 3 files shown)


lbergelson:
@samuelklee 

In the non-docker integration test build we do run the R installation script, but a failure doesn't kill the build unless the tests fail because of it.  

Would it be less painful if we had a readme explaining how to update the r-dependencies?  That way it wouldn't be a surprise that you have to build a new docker image.  Typically the workflow when updating the R installation script should probably be to update the base image and update the Dockerfile in the same PR that updates the R dependencies.  That way we don't get out of sync.  

I do like the idea of hosting the cran dependencies somewhere ourselves.  Those cause us a lot of pain.

samuelklee:
Done, and verified that tests using the new package and base image are passing in my dev branch.  Thanks!

PR #3719, opened 2017-10-19
Removed "SNAPSHOT" from the gatk jar in the docker images

### LINKED ISSUE #3700: Docker build script should set -Drelease=true when building to prevent SNAPSHOT from appearing in the GATK version number

jamesemery:
This will not affect our nightly build naming patterns as that is generated manually by the script.

Fixes #3700

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,11 @@
 # Using OpenJDK 8
 FROM broadinstitute/gatk:gatkbase-1.2.1
+ARG DRELEASE
 
 ADD . /gatk
 
 WORKDIR /gatk
-RUN /gatk/gradlew clean compileTestJava installAll localJar
+RUN /gatk/gradlew clean compileTestJava installAll localJar -Drelease=$DRELEASE
 
 WORKDIR /root
 
--- a/build_docker.sh
+++ b/build_docker.sh
@@ -123,9 +123,7 @@ if [ -z "${IS_NOT_RUN_UNIT_TESTS}" ] ; then
 	fi
 
 	git lfs pull
-    chmod a+w ${STAGING_ABSOLUTE_PATH}/build/reports/tests
     chmod -R a+w ${STAGING_ABSOLUTE_PATH}/src/test/resources
-    chmod a+w ${STAGING_ABSOLUTE_PATH}/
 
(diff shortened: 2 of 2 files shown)


jamesemery:
I have tested the output by running building the docker on my machine, here is the command line output demonstrating that it works: 
```
12:31:13 emeryj@wma77-b9c:~/hellbender/gatk $ docker run -it 2d5bf2e7022f
root@98ca222ca7ae:/gatk# ls build/libs/
gatk-4.beta.6-6-g92fc3a2.jar                gatk-spark.jar                              hellbender.jar
gatk-package-4.beta.6-6-g92fc3a2-local.jar  gatk.jar                                    
gatk-package-4.beta.6-6-g92fc3a2-spark.jar  hellbender-spark.jar                        
root@98ca222ca7ae:/gatk# exit          
exit
12:31:43 emeryj@wma77-b9c:~/hellbender/gatk $ docker run -it ef280f7ce04f
root@d0413e8512b4:/gatk# ls build/libs/
gatk-4.beta.6-SNAPSHOT.jar                gatk-spark.jar                            hellbender.jar
gatk-package-4.beta.6-SNAPSHOT-local.jar  gatk.jar                                  
gatk-package-4.beta.6-SNAPSHOT-spark.jar  hellbender-spark.jar      
```

Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM broadinstitute/gatk:gatkbase-1.2
 ADD . /gatk
 
 WORKDIR /gatk
-RUN /gatk/gradlew clean compileTestJava installAll localJar
+RUN /gatk/gradlew clean compileTestJava installAll localJar -Drelease=true
```
droazen: Can you check whether the `build_docker.sh` script uses this `Dockerfile`? If not, you'll need to modify that as well.

jamesemery: I have tested using the build_docker.sh. It does indeed use this `Dockerfile` as that is how I generated the above command line output.

droazen: In that case, go ahead and merge when ready

droazen:
Ooh, actually, on second thought hold off on the merge @jamesemery. I think we might want to only want to run with `-Drelease=true` when publishing a docker image as part of an official GATK release. We might need to tie it to one or more of the arguments to the `build_docker.sh` script.

jamesemery:
@droazen I have updated it to use a parameter so it will only exclude the SNAPSHOT tag if it is set to push to dockerhub

Issue #2858, opened 2017-06-05
GATK CNV on WGS.


droazen:
@LeeTL1220 commented on [Mon May 23 2016](https://github.com/broadinstitute/gatk-protected/issues/524)

- [ ] File issues for evaluations (and make sure each evaluation is described, including ground truth).  Make sure that there is a milestone and assignee for each.

Need to show some measure of performance for the WGS evaluation (GATK CNV only.  Not ACNV)

samuelklee:
Reviving this.

[ ] Change HDF5 matrix writing to allow matrices with NxM > MAX_INT.  This can be done by chunking.  This will allow smaller bin sizes.
[ ] Check that CreatePanelOfNormals works correctly on Spark cluster.
[ ] Check that changing CBS alpha parameter reduces hypersegmentation.

samuelklee:
@vruano may also have some insight on the PCOV vs RAW issue.

samuelklee:
FYI @sooheelee I pointed you at the docs recently, but realized they're slightly out of date.  CreatePanelOfNormals currently expects proportional coverage, upon which it initially performs a series of filtering steps.  The docs state that these steps are performed on integer counts, which is incorrect.  The fact that filtering yields different results for the two types of input ultimately results in slightly different denoising.

sooheelee:
I'll keep an eye out for the "integer counts" passage in the docs and change this to reflect our recommendations.

samuelklee:
To be specific, I'm referring to Sec. IIB in https://github.com/broadinstitute/gatk-protected/blob/master/docs/CNVs/CNV-methods.pdf.

samuelklee:
Can someone explain this pseudoinverse business to me?  We calculate standardize the counts to give a matrix C, calculate the SVD to get the left-singular matrix U and the pseudoinverse of C, but then perform *another* SVD on U to get the pseudoinverse of U.  Why do we need these pseudoinverses?  @LeeTL1220 @davidbenjamin?

samuelklee:
The memory footprint used by ReadCountCollectionUtils.parse is extremely large compared to the size of the file.  I suspect there may be memory leaks in TableReader.

Speed is also an issue; pandas is 4x faster (taking ~10s on my desktop) on a 300bp-bin WGS read-count file with ~11.5 million rows if all columns are read, and 10x faster if the Target name field (which is pretty much useless) is ignored.

[...]

PR #3726, opened 2017-10-20
extract sample name from input read groups and create an output read …

cwhelan:
…group for the contig alignments

To do downstream stuff like correlate breakpoints in copy number calls that are in VCF format, and perhaps eventually put a genotype column in our output VCF, it would be helpful to keep track of the sample name. This PR tries to help do that by 1) validating that input read groups contain reads from only one sample, 2) extracting the sample name for future use, and 3) putting a constructed read group in our aligned assemblies output file that contains the sample name, and tagging all of the alignment records in that file with the read group id. 

As part of testing this I added an expected aligned contigs file test to `FindBreakpointEvidenceSparkIntegrationTest`.


samuelklee:
This is something we also want to address in the CNV pipeline soon.  However, we typically need to append sample metadata to TSVs (rather than SAMs) and TableReader/TableWriter currently only allows for comment lines starting with `#`.  Hence, in #2858 we tack on a `#SAMPLE_NAME=...` header, which is quite hacky.  Perhaps we can discuss unifying this sort of thing (along with appending sequence dictionaries when appropriate) across the tools?

cwhelan:
@samuelklee In a perfect world, I think I'd prefer VCF as our interchange format for CNV calls, and there the sample names are nicely baked into the format. I asked @mbabadi about this the other day and he thought that was the direction gCNV would eventually move in. In the case here, we want to infer the sample name from the input BAM file, where it's stored in the RG tags. The aligned assemblies output file I'm talking about here is really a temporary output sort of akin to a HaplotypeCaller bamout, and again there I think it's natural to add a readgroup with the sample name.

samuelklee:
We will certainly output VCFs for our final calls, but we also have useful intermediate file formats that would benefit from having the sample name (and sequence dictionary, if appropriate) attached---for example, our read-count and allelic-count files.  I see no reason that these shouldn't have a header with the appropriate tags generated from the input BAM and be read/written in the Tribble framework.

However, there are a few exceptions in the CNV pipeline for files that not collections of Features but are still associated with a sample, so I think it would be nice to extract out the sample-name related code across the various tools.  This is not only to avoid duplicating effort, but also to standardize how we store this information in our various file formats.

cwhelan:
@samuelklee Ah, I see what you are saying. Maybe something like a `SampleDataTable` in Tribble or something? Sounds like a good idea.

PR #3728, opened 2017-10-20
compare to truth data

tedsharpe:
for now, just prints a false positive rate for intervals, and a false negative and false positive rate for novel adjacencies.  we can beef up the reporting later if it seems useful.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
@@ -11,6 +11,7 @@ import htsjdk.variant.variantcontext.VariantContextBuilder;
 import htsjdk.variant.vcf.VCFHeader;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.apache.spark.broadcast.Broadcast;
@@ -26,10 +27,10 @@ import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.spark.GATKSparkTool;
 import org.broadinstitute.hellbender.exceptions.GATKException;
-import org.broadinstitute.hellbender.tools.spark.sv.StructuralVariationDiscoveryArgumentCollection;
 import org.broadinstitute.hellbender.tools.spark.sv.evidence.EvidenceTargetLink;
 import org.broadinstitute.hellbender.tools.spark.sv.evidence.ReadMetadata;
 import org.broadinstitute.hellbender.tools.spark.sv.utils.*;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
 import scala.Tuple2;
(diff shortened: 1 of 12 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
```diff
@@ -174,52 +170,118 @@ public static AlignedContig parseReadsAndOptionallySplitGappedAlignments(final I
...
                                         noveltyTypeAndEvidence._1,
                                         noveltyTypeAndEvidence._2._1,
                                         null,
                                         noveltyTypeAndEvidence._2._2,
-                                        broadcastReference));
+                                        broadcastReference))
+                        .collect();
 
-        List<VariantContext> collectedAnnotatedVariants = annotatedVariants.collect();
+        narlsAndSources.unpersist();
```
TedBrookings: Just since there was the Slack dust-up over .unpersist() vs .destroy(), I want to call to your attention to decide whether you wanted .unpersist() here. I'm still learning Spark and don't have a strong opinion, other than to note I've seen .destroy() crash in the wild, so I'd probably choose .unpersist() and accept the potential memory loss as you have here.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
```diff
@@ -174,52 +170,118 @@ public static AlignedContig parseReadsAndOptionallySplitGappedAlignments(final I
...
+        if ( parameters.truthVCF != null ) {
+            final SVIntervalTree<String> trueBreakpoints =
+                    SVVCFReader.readBreakpointsFromTruthVCF(parameters.truthVCF, sequenceDictionary, parameters.truthIntervalPadding);
+
+            if ( assembledIntervals != null ) {
+                evaluateIntervalsAgainstTruth(assembledIntervals, trueBreakpoints, localLogger);
+            }
+
+            final SVIntervalTree<String> narlyBreakpoints =
+                    readBreakpointsFromNARLs(narlsAndSources.map(Tuple2::_1).collect(), sequenceDictionary, parameters.truthIntervalPadding);
```
TedBrookings: Real minor nitpick: in readBreakpointsFromNARLs and evaluateNARLsAgainstTruth you deviate from camel case, whereas everywhere else you treat "narl" as a camel case word.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
tedsharpe: That had to do with broadcasts.  This is an RDD.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVVCFReader.java
```diff
@@ -0,0 +1,49 @@
...
+                final int start = vc.getStart();
+                switch ( svType ) {
+                    case DEL:
+                    case INV:
+                    case CNV:
+                        breakpoints.put(new SVInterval(contigID,start-padding, start+padding), eventName);
+                        final int end = vc.getEnd();
+                        breakpoints.put(new SVInterval(contigID,end-padding, end+padding), eventName);
+                        break;
+                    case INS:
```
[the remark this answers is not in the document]
tedsharpe: Thank you.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/DiscoverVariantsFromContigAlignmentsSAMSpark.java
tedsharpe: Fixed.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSparkUnitTest.java
```diff
@@ -177,23 +180,26 @@ public void readExternalEvidenceTest() {
...
             }
-            try (final BufferedReader actualReader = new BufferedReader(new FileReader(outputFile));
-                 final BufferedReader expectedReader = new BufferedReader(new FileReader(expectedFastqFilePrefix + intervalAndFastqBytes._1()))){
-                final List<String> actualLines = actualReader.lines().collect(Collectors.toList());
-                final List<String> expectedLines = expectedReader.lines().collect(Collectors.toList());
-                Assert.assertEquals(actualLines.size(), expectedLines.size(), "different number of lines");
-                for (int i = 0; i < actualLines.size(); i++) {
-                    Assert.assertEquals(actualLines.get(i), expectedLines.get(i), "difference in fastq line " + (i + 1));
+
+            try ( final BufferedReader actualReader = new BufferedReader(new FileReader(outputFile)) ) {
```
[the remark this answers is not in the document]
tedsharpe: The edge case is that when my brain malfunctions (which is not infrequent), I fail to recall that you can declare multiple resources in a single try-with-resources.  Fixed.

Issue #3735, opened 2017-10-23
Transient GCS Permission Errors when using 4.beta.6


jsotobroad:
We've been using 4.beta.6 to generate new callsets because it has the GenomicsDBImport batching fix and it seems to have introduced transient Auth errors that production was not seeing before.  This happens a maybe one shard at every task level and when rerun usually succeeds but as you can imagine is pretty annoying.  This happens across multiple tools (GenomicsDBImport, GatherVcfs)

Sometimes we get this as the only response from GATK when this happens

```
***********************************************************************

A USER ERROR has occurred: Couldn't read file. Error was: Failure while waiting for FeatureReader to initialize  with exception: com.google.cloud.storage.StorageException: 403 Forbidden
443301511749-compute@developer.gserviceaccount.com does not have storage.objects.get access to broad-gotc-prod-storage/pipeline/G101956/gvcfs/DDP_ATCP_42_1.4afb46bb-4009-47c4-9aa0-407e92de0db8.g.vcf.gz.

***********************************************************************
```

and other times we get a nice stacktrace for this issue

```
java.lang.RuntimeException: java.util.concurrent.ExecutionException: com.google.cloud.storage.StorageException: 403 Forbidden
443301511749-compute@developer.gserviceaccount.com does not have storage.objects.get access to broad-jg-dev-11k-call-set/JointGenotyping/0cb36821-b8bf-4e6d-a352-07b101f6b7d1/call-ApplyRecalibration/shard-1734/GMKF_Seidman_CHD_WGS_904.filtered.1734.vcf.gz.
	at org.broadinstitute.hellbender.utils.nio.SeekableByteChannelPrefetcher.read(SeekableByteChannelPrefetcher.java:309)
	at htsjdk.samtools.seekablestream.SeekablePathStream.read(SeekablePathStream.java:86)
	at htsjdk.samtools.util.IOUtil.transferByStream(IOUtil.java:141)
	at org.broadinstitute.hellbender.tools.GatherVcfsCloud.gatherWithBlockCopying(GatherVcfsCloud.java:394)
	at org.broadinstitute.hellbender.tools.GatherVcfsCloud.doWork(GatherVcfsCloud.java:143)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:119)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:176)
[...]

PR #3739, opened 2017-10-23
Add retries on auth failure

jean-philippe-martin:
...and test, naturally.

Issue #3735 reports that transient errors can cause the auth server to return "403 Forbidden" and, in turn, cause us to fail. This PR attempts to work around this situation by retrying a few times on these errors.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/nio/SeekableByteChannelPrefetcher.java
@@ -75,7 +75,7 @@ public final class SeekableByteChannelPrefetcher implements SeekableByteChannel 
     public long nbGoingBack = 0;
     // number of times the user asks for data past the end of the file.
     public long nbReadsPastEnd = 0;
-    // number of tries we retried in response to a "403: Forbidden" error.
+    // number of times we retried in response to a "403: Forbidden" error.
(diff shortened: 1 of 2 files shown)


droazen:
@jean-philippe-martin I don't think this should be implemented at the level of the prefetcher -- the prefetcher is not always on when we use NIO. Instead, I'd recommend a patch at the `google-cloud-java` level. If you open a PR against https://github.com/droazen/google-cloud-java/tree/dr_all_nio_fixes (which is the fork/branch our existing `google-cloud-java` snapshot is built from), we can easily move GATK to a modified snapshot for evaluation.

jean-philippe-martin:
I can't write to your fork; will have to fork it again or something but that'll be another day.

droazen:
@jean-philippe-martin Yeah, just fork my fork and open a PR against that branch, and I'll review.

lbergelson:
@droazen Should we just make an official broad fork?

droazen:
@lbergelson I don't think so -- the need for a fork will be eliminated once https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2453 is resolved.

lbergelson:
It seems likely we will need something similar in the future though.

jean-philippe-martin:
@droazen here's a version that changes the NIO channel directly:

https://github.com/droazen/google-cloud-java/pull/1

Once we get it to a shape we like, I have to remember to also send it upstream.

Issue #3724, opened 2017-10-20
com.github.fommil.netlib:all:jar:1.1.2


Vzzarr:
I am starting a Maven Project in which I would like to import your library; so I added this [dependency](http://search.maven.org/#artifactdetails%7Corg.broadinstitute%7Cgatk%7C4.beta.2%7C) to my pom.xml
```
<dependency>
    <groupId>org.broadinstitute</groupId>
    <artifactId>gatk</artifactId>
    <version>4.beta.2</version>
</dependency>
```
When I execute `mvn clear install` in my folder project, I receive this error: 
```
[ERROR] Failed to execute goal on project GATKpipe: 
Could not resolve dependencies for project uk.ac.ncl:GATKpipe:jar:0.0.1-SNAPSHOT: 
Could not find artifact com.github.fommil.netlib:all:jar:1.1.2 in 
all (https://mvnrepository.com/artifact/com.github.fommil.netlib/all) -> [Help 1]
```
and it seems that the problem is the dependency by com.github.fommil.netlib/all, indeed according to the output of `mvn clear install`, it attempt to download all-1.1.2.jar:
`Downloading: https://repo.maven.apache.org/maven2/com/github/fommil/netlib/all/1.1.2/all-1.1.2.jar`
but this jar is not available in the repository.

I noticed that even in other [projects](https://github.com/amplab/ml-matrix/issues/11) have the same issue.

How is possible to resolve this issue? 

Thanks for your time,
Nicholas

lbergelson:
Hi @Vzzarr ,

Thank you for reporting this.  I've never heard of this happening before, but I suspect the difference is that you are using maven and we've only ever used gradle.  @magicDGS @LeeTL1220 You both build downstream projects, have you ever seen this?

From what I can tell `com.github.fommil.netlib.all` is a POM only dependency, but it's not annotated as such in our generated POM.  Gradle resolves it correctly, and then resolves it's transitive dependencies correctly, but maven seems to resolve it differently.  

I've reproduced the issue locally, and it seems like it can be fixed by changing this line in our build file:

https://github.com/broadinstitute/gatk/blob/62e339fb9150db19f80d4b48aed8d47608461690/build.gradle#L220

to specify the exact dependencies we want instead of using the all artifact.

[...]

PR #3742, opened 2017-10-24
specifying com.github.fommil native dependencies explicitly

lbergelson:
previously we dependend on com.github.fommil.netlib:netlib:all which is a pom only dependency that includes a number of jars containing native code

this caused problems when importing GATK using maven, see #3724

to fix this, I've added the exact versions of the transitive dependencies that we want
an added bonus is that it means we are no longer including  native code for systems we don't support, like arm, 32 bit systems, and windows

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -216,8 +216,11 @@ dependencies {
     compile('org.objenesis:objenesis:1.2')
     testCompile('org.objenesis:objenesis:2.1')
 
-    // Comment the next line to disable native code proxies in Spark MLLib
-    compile('com.github.fommil.netlib:all:1.1.2')
+    // Comment the next lines to disable native code proxies in Spark MLLib
+    compile('com.github.fommil.netlib:netlib-native_ref-osx-x86_64:1.1:natives')
+    compile('com.github.fommil.netlib:netlib-native_ref-linux-x86_64:1.1:natives')
+    compile('com.github.fommil.netlib:netlib-native_system-linux-x86_64:1.1:natives')
+    compile('com.github.fommil.netlib:netlib-native_system-osx-x86_64:1.1:natives')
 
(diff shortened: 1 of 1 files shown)


lbergelson:
this should fix #3724, I've tested it locally by building a maven project with the following pom
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.broadinstitute</groupId>
  <artifactId>gatk-downstream-test</artifactId>
  <packaging>jar</packaging>
  <version>1.0-SNAPSHOT</version>
  <name>gatk-downstream-test</name>
  <url>http://maven.apache.org</url>
    <repositories>
       <repository>
          <snapshots />
          <id>snapshots</id>
          <name>libs-snapshot</name>
          <url>https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot</url>
        </repository>
    </repositories>
  <dependencies>
    <dependency>
      <groupId>org.broadinstitute</groupId>
      <artifactId>gatk</artifactId>
      <version>4.beta.6-15-g62e339f-SNAPSHOT</version>
    </dependency>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>3.8.1</version>
      <scope>test</scope>
    </dependency>
  </dependencies>
</project>
```

This didn't build correctly with the current gatk, but builds with this patch, (note that the snapshot version will be different if you download and build this yourself).

@Vzzarr Is it possible for you to build this locally and test it with your project?

PR #3743, opened 2017-10-24
Annotate SV calls with overlapping external CNV calls

cwhelan:
This PR annotates calls made by GATK-SV with any overlapping calls from the external CNV call set passed in.

The input CNV call set must be a single-sample VCF with a sample name that matches the sample name from the input reads and uses VCF-spec FORMAT tags to describe copy number (CN,CNQ).


src/test/java/org/broadinstitute/hellbender/tools/spark/sv/integration/StructuralVariationDiscoveryPipelineSparkIntegrationTest.java
```diff
@@ -74,8 +79,15 @@ public String toString() {
         final File tempDirNew = BaseTest.createTempDir("new");
         tempDirNew.deleteOnExit();
         Files.createDirectories(Paths.get(tempDirNew.getAbsolutePath()+"/fastq"));
-        tests.add(new Object[]{new StructuralVariationDiscoveryPipelineSparkIntegrationTest.StructuralVariationDiscoveryPipelineSparkIntegrationTestArgs(SVIntegrationTestDataProvider.TEST_BAM_NEW, SVIntegrationTestDataProvider.KMER_KILL_LIST, SVIntegrationTestDataProvider.ALIGNER_INDEX_IMG, tempDirNew.getAbsolutePath())});
-
+        tests.add(new Object[]{
+                new StructuralVariationDiscoveryPipelineSparkIntegrationTest.StructuralVariationDiscoveryPipelineSparkIntegrationTestArgs(
+                        SVIntegrationTestDataProvider.TEST_BAM_NEW,
```
SHuang-Broad: What was the reason the test bam is called "new"? I know we replaced the test bam but the old one is gone right? Can we remove the "new'?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -91,17 +111,45 @@ protected void runTool( final JavaSparkContext ctx ) {
...
+                        ctx.broadcast(header.getSequenceDictionary()),
+                        vcfOutputFileName,
+                        localLogger,
+                        evidenceLinkTree,
+                        assembledEvidenceResults.getReadMetadata(),
+                        broadcastCNVCalls
                 );
     }
 
+    public static String getSampleId(final SAMFileHeader header) {
```
SHuang-Broad: I think this method is better living in a "utils" class, like `SVUtils`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -91,17 +111,45 @@ protected void runTool( final JavaSparkContext ctx ) {
         final List<EvidenceTargetLink> evidenceTargetLinks = assembledEvidenceResults.getEvidenceTargetLinks();
         final PairedStrandedIntervalTree<EvidenceTargetLink> evidenceLinkTree = makeEvidenceLinkTree(evidenceTargetLinks);
 
+        final SVIntervalTree<VariantContext> cnvCalls;
```
SHuang-Broad: It looks like these lines are highly similar to those in `DiscoverVariantsFromContigAlignmentsSAMSpark.runTool()`. 
Could you refactor them into a method in `DiscoverVariantsFromContigAlignmentsSAMSpark` and try to have the `runTool()` in `DiscoverVariantsFromContigAlignmentsSAMSpark` thin (easier for the pipeline to call into)?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AnnotatedVariantProducer.java
```diff
@@ -71,16 +74,18 @@
      * @param inferredType                      inferred type of variant
      * @param altHaplotypeSeq                   alt haplotype sequence (could be null)
      * @param contigAlignments                  chimeric alignments from contigs used for generating this novel adjacency
-     * @param broadcastReference                broadcasted reference
-     *
-     * @throws IOException                      due to read operations on the reference
+     * @param broadcastReference                broadcast reference
+     * @param broadcastSequenceDictionary       broadcast reference sequence dictionary
+     *@param broadcastCNVCalls  @throws IOException                      due to read operations on the reference
```
SHuang-Broad: doc is off

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AnnotatedVariantProducer.java
```diff
@@ -108,9 +113,31 @@ static VariantContext produceAnnotatedVcFromInferredTypeAndRefLocations(final Si
...
             vcBuilder.attribute(GATKSVVCFConstants.SEQ_ALT_HAPLOTYPE, new String(altHaplotypeSeq));
 
+        if (broadcastCNVCalls != null && end > 0) {
+            annotateWithCNVCalls(refLoc, end, vcBuilder, broadcastSequenceDictionary, broadcastCNVCalls);
+        }
+
         return vcBuilder.make();
     }
 
+    private static void annotateWithCNVCalls(final SimpleInterval refLoc,
```
SHuang-Broad: a refactoring suggestion: can you make it return a `String` attribute and not taking (hence modifying) a VCBuilder?

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/integration/StructuralVariationDiscoveryPipelineSparkIntegrationTest.java
cwhelan: Will change, I think this was just an artifact of when we replaced it.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
cwhelan: Moved to `SVUtils`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/AnnotatedVariantProducer.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
cwhelan: It seemed like we were trying to leave operations involving the spark context, like actually broadcasting variables, within the `runTool` methods of the tools involved, which is why I didn't end up doing as you suggested. If I refactor this way I'll have to pass the spark context object in. What if I just make a static method to do the load and broadcast on `DiscoverVariantsFromContigAlignmentsSAMSpark` and call it from both places?

SHuang-Broad: Yep, that would be good.

PR #3746, opened 2017-10-25
Improve ChimericAlignment in SV pipeline

SHuang-Broad:
partially cleanup, partially reminiscent of PR series part 5.

Broke into 3 commits, with the first two having no logic change and the last with logic changes.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
@@ -15,8 +15,6 @@ import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 
-import static org.broadinstitute.hellbender.tools.spark.sv.StructuralVariationDiscoveryArgumentCollection.DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection.CHIMERIC_ALIGNMENTS_HIGHMQ_THRESHOLD;
-
 /**
  * Holds information about a split alignment of a contig, which may represent an SV breakpoint. Each ChimericAlignment
  * represents the junction on the contig of two aligned regions. For example, if a contig aligns to three different regions
@@ -73,13 +71,6 @@ public class ChimericAlignment {
         return Arrays.asList(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig);
     }
 
-    Tuple2<SimpleInterval, SimpleInterval> getCoordSortedReferenceSpans() {
-        if (involvesRefPositionSwitch(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig))
-            return new Tuple2<>(regionWithHigherCoordOnContig.referenceSpan, regionWithLowerCoordOnContig.referenceSpan);
-        else
-            return new Tuple2<>(regionWithLowerCoordOnContig.referenceSpan, regionWithHigherCoordOnContig.referenceSpan);
-    }
-
     protected ChimericAlignment(final Kryo kryo, final Input input) {
 
         this.sourceContigName = input.readString();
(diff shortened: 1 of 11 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
```diff
@@ -294,6 +294,24 @@ public boolean isNotSimpleTranslocation() {
                                         involvesRefPositionSwitch(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig));
     }
 
+    /**
+     * See {@link #isLikelyInvertedDuplication()}
+     */
+    @VisibleForTesting
+    public static boolean isLikelyInvertedDuplication(final AlignmentInterval one, final AlignmentInterval two) {
```
cwhelan: Now that this method is on `ChimericAlignment` could you make it non static and remove the parameters (since they're generally the instance variables `regionWithLowerCoordOnContig` and `regionWithHigherCoordOnContig`? Or at least you could create a non-static version that defers to the static one to be used when applicable?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
```diff
@@ -144,14 +142,20 @@ public ChimericAlignment(final AlignmentInterval intervalWithLowerCoordOnContig,
      * If the alignment region passes the above two filters and the next alignment region could be treated as potential inserted sequence,
      * note down the mapping & alignment information of that region and skip it
      * @param alignedContig          made of (sorted {alignmentIntervals}, sequence) of a potentially-signalling locally-assembled contig
-     * @param uniqueRefSpanThreshold for an alignment interval to be used to construct a ChimericAlignment,
-     *                               how long a unique--i.e. the same ref span is not covered by other alignment intervals--alignment on the reference must it have
      * @param referenceDictionary    reference sequence dictionary
+     * @param filterAlignmentByMqOrLength
```
cwhelan: A refactoring idea for maybe improving this code, but don't treat this as a mandatory change for the PR if you dont' think it's useful: What if you passed in a list of filters to be applied instead of having different flags for each filtering type? Each filter object could be a predicate that takes an alignment interval and decides whether to filter it. That would simplify a lot of the if statements in this method.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
```diff
@@ -300,28 +312,32 @@ static boolean isNotSimpleTranslocation(final AlignmentInterval regionWithLowerC
      * But to fully resolve this case, we need other types of evidence, hence should not be the task of this function.
      */
     public boolean isNotSimpleTranslocation() {
-        return isNotSimpleTranslocation(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig, strandSwitch,
-                                        involvesRefPositionSwitch(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig));
+        return !isLikelySimpleTranslocation(regionWithLowerCoordOnContig, regionWithHigherCoordOnContig, strandSwitch);
```
cwhelan: I'd remove this method now that's just a simple negation.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
SHuang-Broad: Yes, that is available several lines down. And the same case with `isLikelySimpleTranslocation()`: one static one for easier testing, and one non-static one for easier usage.

cwhelan: Oh, sorry I missed that. Carry on then.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
SHuang-Broad: this method exists for easier usage by other classes (mostly for filtering or branching logic). How about changing it to non-static version `isLikelySimpleTranslocation()`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
SHuang-Broad: Thanks! This is nice. I'll make it this way, but in the future because:

currently, the filtering is applied non-homogeneously&mdash;the first several alignments are filtered based on MQ only and later ones are filtered based on alignment length and MQ. To introduce as less logic changes possible to master pipeline right now, I'll leave it as it is right now, and bring in the suggested more principled way when the alternative path is merged to master.

PR #3752, opened 2017-10-30
Cpx SV PR series, part-6: Update prototyping SV discovery manager tool

SHuang-Broad:
As more and more branches are cleaned up, trunk needs to be updated.

This reflects updates to the tree described in #2703


src/test/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignmentUnitTest.java
```diff
@@ -7,8 +7,6 @@
 import htsjdk.samtools.SAMRecord;
```
magicDGS: This files might have conflicts after https://github.com/broadinstitute/gatk/pull/3475

SHuang-Broad: Yep. Thanks for pointing out.
I do plan to go over the SV tests when I have spare time over the weekend, and update them as appropriate.
Maybe I'll even name you as the reviewer/victim :smirk:

magicDGS: For the test code I'm happy to review as I did the refactor of the base test for my own benefit; I do not think that I will be useful for the rest of the SV code if you mean that!

SHuang-Broad: definitely not the sv code. just those 1-liner changes that change `BaseTest` to `GATKBaseTest`, if such changes are necessary (i.e. if not already done). 
Thanks!

cwhelan:
@SHuang-Broad can you resolve conflicts on this before review?

SHuang-Broad:
@cwhelan Here you go.
Just cleaned up and renamed this batch of series of PR that should enable us to call ALL types of SV based on alignment signature alone.
Now we just need to wait for feedback on if the call format makes sense/is intuitive or not, and go from there.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.Collections;
+import java.util.EnumMap;
+import java.util.List;
+
+final class AssemblyContigAlignmentSignatureClassifier {
+
+    enum RawTypes {
+        InsDel,                 // 2 alignments, indicating ins/del, simple duplication expansion/contractions
+        IntraChrStrandSwitch,   // intra-chromosome strand switch breakpoints
+        TandemDupOrMEIBkpt,     // intra-chromosome tandem duplications without strand switch OR inter-chromosome MEI (with or without strand switch) breakpoints
```
cwhelan: I'd refrain from calling things MEIs in the code until we know that's what they are. I can see why you might because most of the things that look like this will be MEIs in practice, but there will be other cases where it's not an MEI and it's confusing to call it an MEI before we've verified it (ie checked the insertion sequence against an MEI database or annotation set.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+
+        // split between the case where both alignments has unique ref span or not
+        final Tuple2<JavaRDD<AlignedContig>, JavaRDD<AlignedContig>> hasFullyContainedRefSpanOrNot =
+                RDDUtils.split(preprocessedContigs, AssemblyContigAlignmentSignatureClassifier::hasIncompletePictureDueToRefSpanContainment, false);
+        contigsByRawTypes.put(RawTypes.Incomplete, hasFullyContainedRefSpanOrNot._1);
+
+        // split between same chromosome mapping or not
+        final Tuple2<JavaRDD<AlignedContig>, JavaRDD<AlignedContig>> sameChrOrNot =
+                RDDUtils.split(hasFullyContainedRefSpanOrNot._2, AssemblyContigAlignmentSignatureClassifier::isSameChromosomeMapping, false);
+        contigsByRawTypes.put(RawTypes.TandemDupOrMEIBkpt, sameChrOrNot._2);
```
cwhelan: You put this set into the map here under the `TandemDupOrMEIBkpt` key, but then a few lines later you override it with a different set -- why? I'd recommend not putting collections into the map until they're complete for clarity.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+        contigsByRawTypes.put(RawTypes.InsDel, tandemDupBkptOrSimpleInsDel._2);
+        contigsByRawTypes.put(RawTypes.TandemDupOrMEIBkpt, tandemDupBkptOrSimpleInsDel._1);
+    }
+
+    //==================================================================================================================
+
+    // TODO: 10/27/17 add tests for these predicates
+
+    @VisibleForTesting
+    static boolean hasIncompletePictureDueToRefSpanContainment(final AlignedContig contigWithOnlyOneConfigAnd2Aln) {
```
cwhelan: I'd prefer a name like `doesNotSpanDuplicationBreakpoints` or something similar.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+                ||
+                two.referenceSpan.contains(one.referenceSpan);
+    }
+
+    /**
+     * @param contigWithOnlyOneConfigAnd2Aln assumed to have only two alignments
+     */
+    @VisibleForTesting
+    static boolean isSameChromosomeMapping(final AlignedContig contigWithOnlyOneConfigAnd2Aln) {
+        Utils.validateArg(!hasIncompletePictureDueToRefSpanContainment(contigWithOnlyOneConfigAnd2Aln),
```
cwhelan: This assumption seems weird given the method name and what it does. I'd remove it.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+        return contigWithOnlyOneConfigAnd2Aln.alignmentIntervals.get(0).referenceSpan.getContig()
+                .equals(contigWithOnlyOneConfigAnd2Aln.alignmentIntervals.get(1).referenceSpan.getContig());
+    }
+
+    /**
+     * @param contigWithOnlyOneConfigAnd2AlnToSameChr assumed to have only two alignments mapped to the same chromosome
+     */
+    @VisibleForTesting
+    static boolean indicatesIntraChrStrandSwitchBkpts(final AlignedContig contigWithOnlyOneConfigAnd2AlnToSameChr) {
+        Utils.validateArg(isSameChromosomeMapping(contigWithOnlyOneConfigAnd2AlnToSameChr),
```
cwhelan: Given the method names, maybe the right thing to do here is to just return false if the predicate doesn't hold? Ie return false if the mappings are on different chromosomes.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+                contigWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(1).forwardStrand;
+    }
+
+    /**
+     * @param contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch assumed to have only two alignments mapped to the same chromosome and the same strand
+     */
+    @VisibleForTesting
+    static boolean indicatesIntraChrTandemDupBkpts(final AlignedContig contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch) {
+        Utils.validateArg( !indicatesIntraChrStrandSwitchBkpts(contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch),
+                "assumption that input contig's 2 alignments map to the same chr is violated. \n" +
```
cwhelan: This message seems wrong, and again, it might be better to just return false if the condition doesn't hold rather than throwing an error, given how you've structured the code and named these methods.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+                        contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch.toString());
+
+        final AlignmentInterval one = contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch.alignmentIntervals.get(0),
+                                two = contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch.alignmentIntervals.get(1);
+        final ChimericAlignment tempCA = new ChimericAlignment(one, two, dummyInsertionMappingStringList,
+                "dummy", null); // passing null because alignments are guanranteed to be same-chr mapping
```
cwhelan: guaranteed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+     */
+    @VisibleForTesting
+    static boolean indicatesIntraChrTandemDupBkpts(final AlignedContig contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch) {
+        Utils.validateArg( !indicatesIntraChrStrandSwitchBkpts(contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch),
+                "assumption that input contig's 2 alignments map to the same chr is violated. \n" +
+                        contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch.toString());
+
+        final AlignmentInterval one = contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch.alignmentIntervals.get(0),
+                                two = contigWithOnlyOneConfigAnd2AlnToSameChrWithoutStrandSwitch.alignmentIntervals.get(1);
+        final ChimericAlignment tempCA = new ChimericAlignment(one, two, dummyInsertionMappingStringList,
```
cwhelan: Why do you need to create this dummy object? If it's just to get access to an instance method of `ChimericAlignment` I'd rather refactor that method to be static and just call it from here without needing to create dummy field values, etc.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+                "dummy", null); // passing null because alignments are guanranteed to be same-chr mapping
+        return !tempCA.isNotSimpleTranslocation();
+    }
+
+    private static final List<String> dummyInsertionMappingStringList = Collections.emptyList();
```
cwhelan: I don't think you need this as a separate static variable, I'd just inline it.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+
+    //==================================================================================================================
+
+    // TODO: 10/27/17 salvation on these reads, and flag accordingly
+    static EnumMap<RawTypes, JavaRDD<AlignedContig>> dumpsterDivingInAmbiguous() {
```
cwhelan: Maybe call this `ambiguousMappingsToBeRescued` or something like that?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -0,0 +1,171 @@
...
+        return new EnumMap<>(RawTypes.class);
+    }
+
+    static void dumpsterDivingInIncompletePicture(final JavaRDD<AlignedContig> incompletePictures) {
```
cwhelan: Remove this until there's an implementation.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/ChimericAlignment.java
```diff
@@ -303,7 +309,9 @@ static boolean isLikelySimpleTranslocation(final AlignmentInterval regionWithLow
      * <ul>
      *  <li>inter-chromosomal translocations, i.e. novel adjacency between different reference chromosomes, or</li>
      *  <li>intra-chromosomal translocation that DOES NOT involve a strand switch, i.e.
-     *      novel adjacency between reference locations on the same chromosome involving NO strand switch but reference interval order switch
+     *      novel adjacency between reference locations on the same chromosome involving NO strand switch,
+     *      but in the meantime, the two inducing alignments CANNOT overlap each other since that would point to
```
cwhelan: "but in the meantime," -> "with the condition that"

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -103,6 +104,12 @@ private static void decisionTreeForContigsWithTwoAlignments(final JavaRDD<Aligne
 
     // TODO: 10/27/17 add tests for these predicates
 
+    static boolean hasIncompletePicture(final AlignedContig contigWithOnlyOneConfigAnd2Aln) {
```
cwhelan: Again see my other comment about renaming this to be more clear about what's going on, ie something like `doesNotSpanDuplicationBreakpoints` or something

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -97,7 +97,7 @@ private static void decisionTreeForContigsWithTwoAlignments(final JavaRDD<Aligne
         final Tuple2<JavaRDD<AlignedContig>, JavaRDD<AlignedContig>> tandemDupBkptOrSimpleInsDel =
                 RDDUtils.split(strandSwitchOrNot._2, AssemblyContigAlignmentSignatureClassifier::indicatesIntraChrTandemDupBkpts, false);
         contigsByRawTypes.put(RawTypes.InsDel, tandemDupBkptOrSimpleInsDel._2);
-        contigsByRawTypes.put(RawTypes.TandemDupOrMEIBkpt, tandemDupBkptOrSimpleInsDel._1);
+        contigsByRawTypes.replace(RawTypes.TandemDupOrMEIBkpt, contigsByRawTypes.get(RawTypes.TandemDupOrMEIBkpt).union(tandemDupBkptOrSimpleInsDel._1));
```
cwhelan: Oh, I see, you fixed this later. I think it might be clearer to keep the first set of alignments in a local variable, and then then only add it to the map after joining the two groups together.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -136,6 +136,7 @@ static boolean hasIncompletePictureDueToOverlappingRefOrderSwitch(final AlignedC
         if (referenceSpanOne.contains(referenceSpanTwo) || referenceSpanTwo.contains(referenceSpanOne))
             return true;
 
+        // TODO: 10/29/17 this obsoletes the inverted duplication call code we have now, but those could be used to figure out how to annotate which known ref regions are invert duplicated
```
cwhelan: This TODO comment is also not very clear to me. Can you explicitly write down what needs to be done?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: removed this and the next method but kept TODO's.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: replaced with the now public method `ChimericAlignment.isLikelySimpleTranslocation(...)`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: removed

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: done as suggested

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: done

PR #3770, opened 2017-11-01
Cpx SV PR series, part-7: Finetune multiple/chimeric alignment of a single assembly contig

SHuang-Broad:
for multiple alignment fine tuning for 
* sending back some contigs back to the 2 alignment logic tree, as handled by #3752 , and 
* capturing more incomplete picture assemblies, and 
* sending the assembly contig seemingly has a complete picture down the multiple alignment logic tree, which is being finalized.

To be merged after #3752 is reviewed.

Code for hooking it up exists and is minimal.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
@@ -1,6 +1,5 @@
 package org.broadinstitute.hellbender.tools.spark.sv.discovery.prototype;
 
-import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.SAMSequenceDictionary;
 import org.apache.logging.log4j.Logger;
 import org.apache.spark.api.java.JavaRDD;
@@ -10,14 +9,16 @@ import org.broadinstitute.hellbender.tools.spark.sv.discovery.AlignmentInterval;
 import org.broadinstitute.hellbender.tools.spark.sv.discovery.ChimericAlignment;
 import org.broadinstitute.hellbender.tools.spark.sv.discovery.StrandSwitch;
 import org.broadinstitute.hellbender.tools.spark.sv.utils.RDDUtils;
-import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.Utils;
 import scala.Tuple2;
 
-import java.util.EnumMap;
-import java.util.List;
+import java.util.*;
+import java.util.stream.Collectors;
 
 final class AssemblyContigAlignmentSignatureClassifier {
+
+    public static final int ALIGNMENT_MAPQUAL_THREHOLD = 20;
+    public static final int ALIGNMENT_READSPAN_THRESHOLD = 10;
 
     enum RawTypes {
         InsDel,                 // 2 alignments, indicating ins/del, simple duplication expansion/contractions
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+        return new Tuple2<>(selectedAlignments, lowUniquenessMappings);
+    }
+
+    /**
+     * Each alignment has an entry of a tuple2, one for max overlap front, one for max overlap rear,
+     * max overlap front is a tuple2 registering the index and overlap bases count
+     */
+    private static List<Tuple2<Tuple2<Integer, Integer>, Tuple2<Integer, Integer>>> getMaxOverlapPairs(final List<AlignmentInterval> selectedAlignments) {
+
+        final List<Tuple2<Tuple2<Integer, Integer>, Tuple2<Integer, Integer>>> maxOverlapMap =
```
cwhelan: I'd make this be a `Map<AlignmentInterval, Tuple2<>>` instead of a `List`.

cwhelan: Actually, I think two maps might be better, one for "front" overlap and one for "rear". Then return a tuple2 of the two maps.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+                new ArrayList<>(Collections.nCopies(selectedAlignments.size(), new Tuple2<>(new Tuple2<>(-1, -1), new Tuple2<>(-1, -1))));
```
cwhelan: If you use maps you don't have to initialize stuff with fake values.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+
+        for(int i = 0; i < selectedAlignments.size() - 1; ++i) {
+            final AlignmentInterval cur = selectedAlignments.get(i);
+            int maxOverlap = -1;
+            int maxOverlapIdx = -1;
+            for (int j = i + 1; j < selectedAlignments.size(); ++j) {
+                final int overlap = AlignmentInterval.overlapOnContig(cur, selectedAlignments.get(j));
+                if (overlap > 0) {
+                    maxOverlap = Math.max(maxOverlap, overlap);
+                    maxOverlapIdx = j;
```
cwhelan: I might be confused about the logic but don't you only want to change the index if this overlap is in fact the max? Ie what if the previous statement left `maxOverlap` unchanged, do you still want to change `maxOverlapIdx`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+                }
+                else // following ones, as guaranteed by the ordering of alignments in the contig, cannot overlap
+                    break;
```
cwhelan: I think braces would be helpful

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+            final AlignmentInterval cur = selectedAlignments.get(i);
+            int maxOverlap = -1;
+            int maxOverlapIdx = -1;
+            for (int j = i + 1; j < selectedAlignments.size(); ++j) {
+                final int overlap = AlignmentInterval.overlapOnContig(cur, selectedAlignments.get(j));
+                if (overlap > 0) {
+                    maxOverlap = Math.max(maxOverlap, overlap);
+                    maxOverlapIdx = j;
+                }
+                else // following ones, as guaranteed by the ordering of alignments in the contig, cannot overlap
```
cwhelan: I'm also a little confused about this logic. What if the first alignment doesn't overlap this one (say this one is no. 4) but alignments no. 2 and 3 do?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * Preprocess provided {@code originalConfiguration} of a particular contig and return a configuration after operation.
+     * Note that "original" is meant to be possibly different from the returned configuration,
+     * but DOES NOT mean the alignments of the contig as given by the aligner, i.e. the configuration should be
+     * one of the best given by {@link FilterLongReadAlignmentsSAMSpark#pickBestConfigurations(AlignedContig, Set, Double)}.
+     *
+     * Note that the configuration after this preprocessing may not even have the same number of alignment as in input configuration.
+     */
+    static Tuple2<List<AlignmentInterval>, List<AlignmentInterval>> fineTuneAlignments(final List<AlignmentInterval> originalConfiguration,
```
cwhelan: I'd rename this method to something more specific -- `removeNonUniqueMappings`? -- and add some documentation on what the return value is.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+            if (alignment.mapQual >= mapQThresholdInclusive)
+                selectedAlignments.add(alignment);
+            else
+                lowUniquenessMappings.add(alignment);
+        }
+
+        // second pass, the slower one, is to remove alignments offering low READ UNIQUENESS,
+        // i.e. with only a very short part of the read being uniquely explained by this particular alignment;
+        // the steps are:
+        //      search bi-directionally until cannot find overlap any more, subtract from it all overlaps.
```
cwhelan: "subtract the overlap from the distance covered on the contig by the alignment"

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+        // max overlap front is a tuple2 registering the index and overlap bases count
+        final List<Tuple2<Tuple2<Integer, Integer>, Tuple2<Integer, Integer>>> maxOverlapMap =
+                getMaxOverlapPairs(selectedAlignments);
+
+        final List<Integer> idxToRemove = new ArrayList<>(selectedAlignments.size());
+        for (int i = 0; i < selectedAlignments.size(); ++i) {
+            final AlignmentInterval cur = selectedAlignments.get(i);
+            final Tuple2<Tuple2<Integer, Integer>, Tuple2<Integer, Integer>> maxOverlapFrontAndRear = maxOverlapMap.get(i);
+            final int maxOverlapFront = Math.max(0, maxOverlapFrontAndRear._1._2);
+            final int maxOverlapRead = Math.max(0, maxOverlapFrontAndRear._2._2);
```
cwhelan: should this be `maxOverlapRear`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+            if (uniqReadSpan < uniqReadLenInclusive)
+                idxToRemove.add(i);
+        }
+
+        if ( idxToRemove.isEmpty() )
+            return new Tuple2<>(selectedAlignments, lowUniquenessMappings);
+
+        // removing in reverse order so that iterators are not invalidated if we were to remove from start
+        final ListIterator<Integer> rit = idxToRemove.listIterator(idxToRemove.size());
+        while (rit.hasPrevious()) {
```
cwhelan: Do you want alignments being filtered at this step to be added to `lowUniquenessMappings`? It doesn't look like they are right now.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+
+            final int uniqReadSpan = cur.endInAssembledContig - cur.startInAssembledContig + 1 - maxOverlapFront - maxOverlapRead;
+            if (uniqReadSpan < uniqReadLenInclusive)
+                idxToRemove.add(i);
+        }
+
+        if ( idxToRemove.isEmpty() )
+            return new Tuple2<>(selectedAlignments, lowUniquenessMappings);
+
+        // removing in reverse order so that iterators are not invalidated if we were to remove from start
```
cwhelan: It's ok to remove via iterator if traversing forward. I think you could do this all in one loop instead of re-iterating over the list. Just iterate over selected alignments, test the overlap from the, and remove it via the iterator if the overlap is too high.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+            return true;
+
+        if (head.forwardStrand) {
+            return referenceSpanHead.getStart() >= referenceSpanTail.getStart();
+        } else {
+            return referenceSpanHead.getEnd() <= referenceSpanHead.getEnd();
+        }
+    }
+
+    private static boolean isNonInformative(final Tuple3<Boolean, AlignedContig, List<String>> tigInfo) {
```
cwhelan: I'd just make this `Tuple3` into a little class, then you don't need these extra methods to decode their entries.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+        final JavaPairRDD<AlignedContig, List<String>> multipleAlignmentsIncompletePicture =
+                multipleAlignments.filter(pair -> hasIncomePicture(pair._1));
+        final JavaPairRDD<AlignedContig, List<String>> multipleAlignmentsCompletePicture =
+                multipleAlignments.filter(pair -> !hasIncomePicture(pair._1));
+
+        return new Tuple4<>(garbage, multipleAlignmentsIncompletePicture.keys(),
+                twoAlignments, multipleAlignmentsCompletePicture);
+    }
+
+    private static boolean hasIncomePicture(final AlignedContig contigWithMultipleAlignments) {
```
cwhelan: `hasIncompletePicture`? Or as I suggested before maybe something more explicit like `doesNotContainDuplicationBreakpoints`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+                        contigWithMultipleAlignments.toString());
+
+        final AlignmentInterval head = contigWithMultipleAlignments.alignmentIntervals.get(0),
+                                tail = contigWithMultipleAlignments.alignmentIntervals.get(contigWithMultipleAlignments.alignmentIntervals.size()-1);
+
+        // tail not resuming the "direction of flow" started by head
+        if ( !head.referenceSpan.getContig().equals(tail.referenceSpan.getContig()) || head.forwardStrand != tail.forwardStrand)
+            return true;
+
+        // reference span anomaly
```
cwhelan: Can you expand this comment more? Especially since the next block of code is based on a variable named `refSpanAnomaly`.. how does that differ from this check?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
```diff
@@ -0,0 +1,211 @@
...
+                             referenceSpanTail = tail.referenceSpan;
+        if (referenceSpanHead.contains(referenceSpanTail) || referenceSpanTail.contains(referenceSpanHead))
+            return true;
+
+        final boolean refSpanAnomaly =
+                contigWithMultipleAlignments.alignmentIntervals
+                        .subList(1, contigWithMultipleAlignments.alignmentIntervals.size() - 1)
+                        .stream().map(ai -> ai.referenceSpan)
+                        .anyMatch( middle -> { // middle alignments' ref span either disjoint from head/tail, or completely contained in head/tail ref span
+                            final boolean badHead = middle.overlaps(referenceSpanHead) && !referenceSpanHead.contains(middle);
```
cwhelan: you're saying a middle alignment is bad if it overlaps the head and isn't contained by it? That's not what I was expecting. Is that what you want (or am I misreading something)? Can you explain further what the reference span anomaly is in this case?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/TempMultipleAlignmentsReclassifier.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -10,174 +9,413 @@
...
+    // below are dummy predicates
+    private static boolean isInformative(final AssemblyContigWithFineTunedAlignments contig) {
+        return contig.isInformative();
+    }
+
+    private static boolean hasOnly2GoodAlignments(final AssemblyContigWithFineTunedAlignments contig) {
+        return contig.hasOnly2GoodAlignments();
+    }
+
+    private static boolean hasIncomePicture(final AssemblyContigWithFineTunedAlignments contig) {
```
cwhelan: `hasIncompletePicture`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -10,174 +9,413 @@
...
+
+        return ChimericAlignment.isLikelySimpleTranslocation(first, last, StrandSwitch.NO_SWITCH);
+    }
+
+    //==================================================================================================================
+    // FOR ASSEMBLY CONTIGS WITH MORE THAN TWO ALIGNMENTS
+
+    /**
+     * convenience struct holding 4 classes:
+     *      1) non-informative contigs who after fine tuning has 0 or 1 good alignment left
```
cwhelan: Just for ease of reading it'd be nice to order the lines of this comment in the same order as the fields and constructor params in the code.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -10,174 +9,413 @@
...
+        // i.e. with only a very short part of the read being uniquely explained by this particular alignment;
+        // the steps are:
+        //      search bi-directionally until cannot find overlap any more,
+        //      subtract the overlap from the distance covered on the contig by the alignment.
+        //      This gives unique read region it explains.
+        //      If this unique read region is "short": shorter than {@code uniqReadLenInclusive}), drop it.
+
+        // each alignment has an entry of a tuple2, one for max overlap maxFront, one for max overlap maxRear,
+        // max overlap maxFront is a tuple2 registering the index and overlap bases count
+        final List<MaxOverlapInfo> maxOverlapMap = getMaxOverlapPairs(selectedAlignments);
```
cwhelan: I still think this would be nicer as a `Map<AlignmentInterval, MaxOverlapInfo>` -- then you wouldn't need your `k` index variable -- but I won't insist if you like it this way better.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
```diff
@@ -10,174 +9,413 @@
...
-                        contigWithOnlyOneConfigAnd2AlnToSameChr.toString());
-        return contigWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(0).forwardStrand
-                ^
-                contigWithOnlyOneConfigAnd2AlnToSameChr.alignmentIntervals.get(1).forwardStrand;
+    private static final class MaxOverlapInfo {
+        final Tuple2<Integer, Integer> maxFront; // 1st holds index pointing to another alignment before this, 2nd holds the count of overlapping bases
+        final Tuple2<Integer, Integer> maxRear;  // same intention as above, but for alignments after this
+
+        MaxOverlapInfo() {
+            maxFront = new Tuple2<>(-1, -1);
```
cwhelan: Again, this is just for clarity and not mandatory, but what if you made `MaxOverlapInfo` have the following fields:

`AlignmentInterval maxFrontOverlapper`
`int maxFrontOverlap`
`AlignmentInterval maxRearOverlapper`
`int maxRearOverlap`

Then you don't have int indices..

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: done with some reordering. The old order was carried-over from the original bad tuple implementation.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: done. I wish Java have something similar to the `zip` function in Python....

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/AssemblyContigAlignmentSignatureClassifier.java
SHuang-Broad: The index is actually used in the implementation that uses it to retrieve from the old list implementation. Again, I hope there's a `zip`...

PR #3776, opened 2017-11-02
Added Utils::split (and tests).  Added performance test for string splitting.

### LINKED ISSUE #3759: Create a unit test to compare various String.split methods in the GATK
There are several methods that perform the `String.split` operation in the GATK, two of which claim to be faster than the base Java's `String.split`.

Make a unit test to compare the performance of the baseline Java `String.split`, HTSJDK's `ParsingUtils::split` and GATK3's `Utils::split`.

jonn-smith:
Based on timing results (see below), we should do two things:
	1) Switch to the HTSJDK `ParsingUtils::split` method for all
	   cases where we split a string using a single character.
	2) Switch to the `Utils::split` method for all cases where
	   we split a string by another string of length > 1.

A quick stopgap that will reduce splitting time by ~1/2 is to just
replace all calls to Java's `String::split` with `Utils::split` (since
they both take two Strings as arguments, it should be very easy).

Also, I have disabled the test so that it doesn't slow down testing cycles.

Fixes #3759

| Method | Benchmark | Total Time (ns) | Total Time (ms) | Time Per Split Operation (ns) | Time Per Split Operation (ms) |   
| --- | --- | --- | --- | --- | --- |
| Java String::split | Split on Words | 131867865203 | 131867.865203 | 6048.98464233945 | 0.00604898464233945 |
| Java String::split | Split on Chars | 12917243085 | 12917.243085 | 3004.010019767442 | 0.003004010019767442 |
| HTSJDK ParsingUtils::split | Split on Words | N/A | N/A | N/A | N/A | 
| HTSJDK ParsingUtils::split | Split on Chars | 5882790859 | 5882.790859 | 1368.0908974418605 | 0.0013680908974418606 | 
| GATK Utils::split | Split on Words | 38734463275 | 38734.463275 | 1776.8102419724771 | 0.0017768102419724772 |
| GATK Utils::split | Split on Chars | 7120052467 | 7120.052467 | 1655.826155116279 | 0.0016558261551162792 |

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
@@ -4,6 +4,7 @@ import com.google.common.collect.AbstractIterator;
 import com.google.common.collect.Iterators;
 import com.google.common.primitives.Ints;
 import htsjdk.samtools.SAMFileHeader;
+import htsjdk.tribble.util.ParsingUtils;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -1143,9 +1144,9 @@ public final class Utils {
      *
      * @param str the string to split.
      * @param delimiter the delimiter used to split the string.
-     * @return an array of tokens.
+     * @return A {@link String} Array of tokens.
      */
-    public static ArrayList<String> split(final String str, final String delimiter) {
+    public static String[] split(final String str, final String delimiter) {
         // This is 10 because the ArrayList default capacity is 10 (but private).
         return split(str, delimiter, 10);
     }
@@ -1155,9 +1156,9 @@ public final class Utils {
      * @param str The {@link String} to split.
      * @param delimiter The delimiter used to split the {@link String}.
      * @param expectedNumTokens The number of tokens expected (used to initialize the capacity of the {@link ArrayList}).
-     * @return An {@link ArrayList} of tokens.
+     * @return A {@link String} Array of tokens.
      */
-    public static ArrayList<String> split(final String str, final String delimiter, final int expectedNumTokens) {
+    public static String[] split(final String str, final String delimiter, final int expectedNumTokens) {
         final ArrayList<String> result =  new ArrayList<>(expectedNumTokens);
 
         if ( str.length() == 0) {
@@ -1167,6 +1168,9 @@ public final class Utils {
             for ( int i = 0; i < str.length(); ++i ) {
                 result.add( str.substring(i, i+1) );
             }
+        }
+        else if ( delimiter.length() == 1) {
+            return ParsingUtils.split(str, delimiter.charAt(0)).toArray(new String[]{});
         }
         else {
             int delimiterIdx = -1;
@@ -1178,6 +1182,6 @@ public final class Utils {
             } while (delimiterIdx != -1);
         }
 
-        return result;
+        return result.toArray(new String[]{});
     }
 }
--- a/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
@@ -6,6 +6,7 @@ import com.google.common.primitives.Ints;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.tribble.util.ParsingUtils;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.jexl2.JexlException;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.math3.random.RandomDataGenerator;
@@ -1141,47 +1142,82 @@ public final class Utils {
 
     /**
      * Splits a String using indexOf instead of regex to speed things up.
-     *
+     * If given an empty delimiter, will return each character in the string as a token.
      * @param str the string to split.
      * @param delimiter the delimiter used to split the string.
-     * @return A {@link String} Array of tokens.
+     * @return A {@link List} of {@link String} tokens.
      */
-    public static String[] split(final String str, final String delimiter) {
+    public static List<String> split(final String str, final String delimiter) {
         // This is 10 because the ArrayList default capacity is 10 (but private).
         return split(str, delimiter, 10);
     }
 
     /**
+     * Splits a String using indexOf instead of regex to speed things up.
+     *
+     * @param str the string to split.
+     * @param delimiter the delimiter used to split the string.
+     * @return A {@link List} of {@link String} tokens.
+     */
+    public static List<String> split(final String str, final char delimiter) {
+        List<String> tokens = ParsingUtils.split( str, delimiter );
+
+        // We must adjust for splitting on the last character so that the results here will be the same as
+        // the results of String.split:
+        if ( delimiter == str.charAt(str.length() - 1)) {
+            tokens.remove( tokens.size() - 1 );
+        }
+
+        return tokens;
+    }
+
+
+    /**
      * Splits a given {@link String} using {@link String#indexOf} instead of regex to speed things up.
+     * If given an empty delimiter, will return each character in the string as a token.
      * @param str The {@link String} to split.
      * @param delimiter The delimiter used to split the {@link String}.
      * @param expectedNumTokens The number of tokens expected (used to initialize the capacity of the {@link ArrayList}).
-     * @return A {@link String} Array of tokens.
+     * @return A {@link List} of {@link String} tokens.
      */
-    public static String[] split(final String str, final String delimiter, final int expectedNumTokens) {
-        final ArrayList<String> result =  new ArrayList<>(expectedNumTokens);
+    public static List<String> split(final String str, final String delimiter, final int expectedNumTokens) {
+        List<String> result =  new ArrayList<>(expectedNumTokens);
 
-        if ( str.length() == 0) {
+        if ( str.isEmpty() ) {
             result.add("");
         }
-        else if ( delimiter.length() == 0) {
+        else if ( delimiter.isEmpty() ) {
             for ( int i = 0; i < str.length(); ++i ) {
                 result.add( str.substring(i, i+1) );
             }
         }
         else if ( delimiter.length() == 1) {
-            return ParsingUtils.split(str, delimiter.charAt(0)).toArray(new String[]{});
+            result = ParsingUtils.split(str, delimiter.charAt(0));
+
+            // We must adjust for splitting on the last character so that the results here will be the same as
+            // the results of String.split:
+            if ( delimiter.charAt(0) == str.charAt(str.length() - 1)) {
+                result.remove( result.size() - 1 );
+            }
         }
         else {
             int delimiterIdx = -1;
+            int tokenStartIdx = delimiterIdx + 1;
             do {
-                final int tokenStartIdx = delimiterIdx + 1;
                 delimiterIdx = str.indexOf(delimiter, tokenStartIdx);
                 final String token = (delimiterIdx != -1 ? str.substring(tokenStartIdx, delimiterIdx) : str.substring(tokenStartIdx));
                 result.add(token);
+                tokenStartIdx = delimiterIdx + delimiter.length();
             } while (delimiterIdx != -1);
+
+            // Check to see if the last entry is empty.
+            // String.split doesn't give us empty last elements of the token list, so we remove it.
+            if ( result.get(result.size() - 1).isEmpty() ) {
+                result.remove( result.size() - 1 );
+            }
+
         }
 
-        return result.toArray(new String[]{});
+        return result;
     }
 }
--- a/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
@@ -1137,4 +1137,47 @@ public final class Utils {
         return sorted.get(sorted.size() / 2);
     }
 
+
+    /**
+     * Splits a String using indexOf instead of regex to speed things up.
+     *
+     * @param str the string to split.
+     * @param delimiter the delimiter used to split the string.
+     * @return an array of tokens.
+     */
+    public static ArrayList<String> split(final String str, final String delimiter) {
+        // This is 10 because the ArrayList default capacity is 10 (but private).
+        return split(str, delimiter, 10);
+    }
+
+    /**
+     * Splits a given {@link String} using {@link String#indexOf} instead of regex to speed things up.
+     * @param str The {@link String} to split.
+     * @param delimiter The delimiter used to split the {@link String}.
+     * @param expectedNumTokens The number of tokens expected (used to initialize the capacity of the {@link ArrayList}).
+     * @return An {@link ArrayList} of tokens.
+     */
+    public static ArrayList<String> split(final String str, final String delimiter, final int expectedNumTokens) {
+        final ArrayList<String> result =  new ArrayList<>(expectedNumTokens);
+
+        if ( str.length() == 0) {
+            result.add("");
+        }
+        else if ( delimiter.length() == 0) {
+            for ( int i = 0; i < str.length(); ++i ) {
+                result.add( str.substring(i, i+1) );
+            }
+        }
+        else {
+            int delimiterIdx = -1;
+            do {
+                final int tokenStartIdx = delimiterIdx + 1;
+                delimiterIdx = str.indexOf(delimiter, tokenStartIdx);
+                final String token = (delimiterIdx != -1 ? str.substring(tokenStartIdx, delimiterIdx) : str.substring(tokenStartIdx));
+                result.add(token);
+            } while (delimiterIdx != -1);
+        }
+
+        return result;
+    }
 }
--- a/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
@@ -1190,7 +1190,7 @@ public final class Utils {
      * @param expectedNumTokens The number of tokens expected (used to initialize the capacity of the {@link ArrayList}).
      * @return A {@link List} of {@link String} tokens.
      */
-    public static List<String> split(final String str, final String delimiter, final int expectedNumTokens) {
+    private static List<String> split(final String str, final String delimiter, final int expectedNumTokens) {
         final List<String> result;
 
         if ( str.isEmpty() ) {
--- a/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
@@ -6,7 +6,6 @@ import com.google.common.primitives.Ints;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.tribble.util.ParsingUtils;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.jexl2.JexlException;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.math3.random.RandomDataGenerator;
@@ -1160,7 +1159,7 @@ public final class Utils {
      * @return A {@link List} of {@link String} tokens.
      */
     public static List<String> split(final String str, final char delimiter) {
-        List<String> tokens = ParsingUtils.split( str, delimiter );
+        final List<String> tokens = ParsingUtils.split( str, delimiter );
 
         // We must adjust for splitting on the last character so that the results here will be the same as
         // the results of String.split:
@@ -1181,26 +1180,24 @@ public final class Utils {
      * @return A {@link List} of {@link String} tokens.
      */
     public static List<String> split(final String str, final String delimiter, final int expectedNumTokens) {
-        List<String> result =  new ArrayList<>(expectedNumTokens);
+        final List<String> result;
 
         if ( str.isEmpty() ) {
+            result = new ArrayList<>(1);
             result.add("");
         }
         else if ( delimiter.isEmpty() ) {
+            result = new ArrayList<>( str.length() );
             for ( int i = 0; i < str.length(); ++i ) {
                 result.add( str.substring(i, i+1) );
             }
         }
         else if ( delimiter.length() == 1) {
-            result = ParsingUtils.split(str, delimiter.charAt(0));
-
-            // We must adjust for splitting on the last character so that the results here will be the same as
-            // the results of String.split:
-            if ( delimiter.charAt(0) == str.charAt(str.length() - 1)) {
-                result.remove( result.size() - 1 );
-            }
+            result = split(str, delimiter.charAt(0));
         }
         else {
+            result = new ArrayList<>(expectedNumTokens);
+
             int delimiterIdx = -1;
             int tokenStartIdx = delimiterIdx + 1;
             do {
--- a/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/Utils.java
@@ -1141,8 +1141,36 @@ public final class Utils {
 
     /**
      * Splits a String using indexOf instead of regex to speed things up.
+     * This method produces the same results as {@link String#split(String)} and {@code String.split(String, 0)},
+     * but has been measured to be ~2x faster (see {@code StringSplitSpeedUnitTest} for details).
+     *
+     * @param str       the string to split.
+     * @param delimiter the delimiter used to split the string.
+     * @return A {@link List} of {@link String} tokens.
+     */
+    public static List<String> split(final String str, final char delimiter) {
+
+        final List<String> tokens;
+
+        if ( str.isEmpty() ) {
+            tokens = new ArrayList<>(1);
+            tokens.add("");
+        }
+        else {
+            tokens = ParsingUtils.split(str, delimiter);
+            removeTrailingEmptyStringsFromEnd(tokens);
+        }
+
+        return tokens;
+    }
+
+    /**
+     * Splits a String using indexOf instead of regex to speed things up.
      * If given an empty delimiter, will return each character in the string as a token.
-     * @param str the string to split.
+     * This method produces the same results as {@link String#split(String)} and {@code String.split(String, 0)},
+     * but has been measured to be ~2x faster (see {@code StringSplitSpeedUnitTest} for details).
+     *
+     * @param str       the string to split.
      * @param delimiter the delimiter used to split the string.
      * @return A {@link List} of {@link String} tokens.
      */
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1138,50 @@ public static void forceJVMLocaleToUSEnglish() {
...
+    }
+
+    /**
+     * Splits a given {@link String} using {@link String#indexOf} instead of regex to speed things up.
+     * @param str The {@link String} to split.
+     * @param delimiter The delimiter used to split the {@link String}.
+     * @param expectedNumTokens The number of tokens expected (used to initialize the capacity of the {@link ArrayList}).
+     * @return A {@link String} Array of tokens.
+     */
+    public static String[] split(final String str, final String delimiter, final int expectedNumTokens) {
```
droazen: Add an overload that takes a char and delegates directly to `ParsingUtils.split()`

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1138,50 @@ public static void forceJVMLocaleToUSEnglish() {
...
+        final ArrayList<String> result =  new ArrayList<>(expectedNumTokens);
+
+        if ( str.length() == 0) {
```
droazen: `str.isEmpty()`

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1138,50 @@ public static void forceJVMLocaleToUSEnglish() {
...
+     */
+    public static String[] split(final String str, final String delimiter) {
+        // This is 10 because the ArrayList default capacity is 10 (but private).
+        return split(str, delimiter, 10);
+    }
+
+    /**
+     * Splits a given {@link String} using {@link String#indexOf} instead of regex to speed things up.
+     * @param str The {@link String} to split.
+     * @param delimiter The delimiter used to split the {@link String}.
```
droazen: Document what an empty delimiter means

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1138,50 @@ public static void forceJVMLocaleToUSEnglish() {
...
+        if ( str.length() == 0) {
+            result.add("");
+        }
+        else if ( delimiter.length() == 0) {
+            for ( int i = 0; i < str.length(); ++i ) {
+                result.add( str.substring(i, i+1) );
+            }
+        }
+        else if ( delimiter.length() == 1) {
+            return ParsingUtils.split(str, delimiter.charAt(0)).toArray(new String[]{});
```
droazen: Return a List if possible, to avoid this wasteful conversion. If this would cause problems for client code, then have two versions: an array version and a List version.

src/test/java/org/broadinstitute/hellbender/utils/StringSplitSpeedUnitTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+import htsjdk.tribble.util.ParsingUtils;
+import org.broadinstitute.hellbender.GATKBaseTest;
+import org.testng.annotations.Test;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * A class to test the speed of different string.split implementations.
```
droazen: Add a note that these tests are disabled by default.

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -731,4 +731,50 @@ public void testGetDuplicatedItems(final Collection<?> collection, final Set<?>
...
+                { s, "r",   s.split("r") },
+                { s, "s",   s.split("s") },
+                { s, "t",   s.split("t") },
+                { s, "u",   s.split("u") },
+                { s, "v",   s.split("v") },
+                { s, "w",   s.split("w") },
+                { s, "x",   s.split("x") },
+                { s, "y",   s.split("y") },
+                { s, "z",   s.split("z") },
+                { s, " ",   s.split(" ") },
```
droazen: Add test cases for:

-splitting on the final char (`.`)
-splitting on the first char (`T`)
-delimiters of more than 1 char (multiple cases)

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Fixed!

src/test/java/org/broadinstitute/hellbender/utils/StringSplitSpeedUnitTest.java
jonn-smith: Fixed!

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
jonn-smith: I've added more text to the test string and added in multiple char delimiters.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -4,7 +4,9 @@
 import com.google.common.collect.Iterators;
 import com.google.common.primitives.Ints;
 import htsjdk.samtools.SAMFileHeader;
+import htsjdk.tribble.util.ParsingUtils;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.jexl2.JexlException;
```
magicDGS: Unused import

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1139,85 @@ public static void forceJVMLocaleToUSEnglish() {
...
+    /**
+     * Splits a given {@link String} using {@link String#indexOf} instead of regex to speed things up.
+     * If given an empty delimiter, will return each character in the string as a token.
+     * @param str The {@link String} to split.
+     * @param delimiter The delimiter used to split the {@link String}.
+     * @param expectedNumTokens The number of tokens expected (used to initialize the capacity of the {@link ArrayList}).
+     * @return A {@link List} of {@link String} tokens.
+     */
+    public static List<String> split(final String str, final String delimiter, final int expectedNumTokens) {
+        List<String> result =  new ArrayList<>(expectedNumTokens);
```
magicDGS: Maybe it's better to instantiate only in the else branch; if there's an empty delimiter, the better guess is `str.length` (and it'll be more efficient) and for the case of a singke char, there's no need for instantiate here.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1139,85 @@ public static void forceJVMLocaleToUSEnglish() {
...
+        if ( str.isEmpty() ) {
+            result.add("");
+        }
+        else if ( delimiter.isEmpty() ) {
+            for ( int i = 0; i < str.length(); ++i ) {
+                result.add( str.substring(i, i+1) );
+            }
+        }
+        else if ( delimiter.length() == 1) {
+            result = ParsingUtils.split(str, delimiter.charAt(0));
```
magicDGS: This branch is repeating the code in tje char method. Using it will avoid mismatch between the methods...

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: That's a good point.  I've moved the instantiation into each if / else case with initial capacities tailored to each case.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Sounds good.  Fixed!

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1137,83 @@ public static void forceJVMLocaleToUSEnglish() {
...
+     * @param str the string to split.
+     * @param delimiter the delimiter used to split the string.
+     * @return A {@link List} of {@link String} tokens.
+     */
+    public static List<String> split(final String str, final char delimiter) {
+        final List<String> tokens = ParsingUtils.split( str, delimiter );
+
+        // We must adjust for splitting on the last character so that the results here will be the same as
+        // the results of String.split:
+        if ( delimiter == str.charAt(str.length() - 1)) {
```
pshapiro4broad: I'm not sure if there are strict rules on this, but the formatting of the code in this comment differs from the most of the code I've seen in GATK:

- most GATK code doesn't have spaces inside parens (I see that spaces inside parens is not done consistently here as well, e.g. on this line there is a space before `delimiter` but not after `1)`)
- most GATK code formats `else` as `} else {` or `} else if (...) {`

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1137,83 @@ public static void forceJVMLocaleToUSEnglish() {
...
+     *
+     * @param str the string to split.
+     * @param delimiter the delimiter used to split the string.
+     * @return A {@link List} of {@link String} tokens.
+     */
+    public static List<String> split(final String str, final char delimiter) {
+        final List<String> tokens = ParsingUtils.split( str, delimiter );
+
+        // We must adjust for splitting on the last character so that the results here will be the same as
+        // the results of String.split:
```
droazen: Document behavior with respect to trailing empty tokens in the method javadoc. Note that `String.split()` can be configured to return trailing empty tokens if you call the two-argument overload and pass in a negative value for the limit parameter, so explicit docs are required here (eg., "this method matches the behavior of `String.split()` with no `limit` argument")

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1137,83 @@ public static void forceJVMLocaleToUSEnglish() {
...
+            int tokenStartIdx = delimiterIdx + 1;
+            do {
+                delimiterIdx = str.indexOf(delimiter, tokenStartIdx);
+                final String token = (delimiterIdx != -1 ? str.substring(tokenStartIdx, delimiterIdx) : str.substring(tokenStartIdx));
+                result.add(token);
+                tokenStartIdx = delimiterIdx + delimiter.length();
+            } while (delimiterIdx != -1);
+
+            // Check to see if the last entry is empty.
+            // String.split doesn't give us empty last elements of the token list, so we remove it.
```
droazen: Document behavior with respect to trailing empty tokens in the method javadoc. Note that `String.split()` can be configured to return trailing empty tokens if you call the two-argument overload and pass in a negative value for the limit parameter, so explicit docs are required here (eg., "this method matches the behavior of `String.split()` with no `limit` argument")

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1137,83 @@ public static void forceJVMLocaleToUSEnglish() {
...
+     * @param delimiter The delimiter used to split the {@link String}.
+     * @param expectedNumTokens The number of tokens expected (used to initialize the capacity of the {@link ArrayList}).
+     * @return A {@link List} of {@link String} tokens.
+     */
+    public static List<String> split(final String str, final String delimiter, final int expectedNumTokens) {
+        final List<String> result;
+
+        if ( str.isEmpty() ) {
+            result = new ArrayList<>(1);
+            result.add("");
```
pshapiro4broad: If the result of this operation doesn't need to be modifiable, it would be more efficient to use `Collections.singletonList("");` here.

droazen: We don't want the method to sometimes return an immutable collection and sometimes to return a mutable one -- unexpected immutable collections often cause us problems.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1137,83 @@ public static void forceJVMLocaleToUSEnglish() {
...
+     * @param delimiter the delimiter used to split the string.
+     * @return A {@link List} of {@link String} tokens.
+     */
+    public static List<String> split(final String str, final char delimiter) {
+        final List<String> tokens = ParsingUtils.split( str, delimiter );
+
+        // We must adjust for splitting on the last character so that the results here will be the same as
+        // the results of String.split:
+        if ( delimiter == str.charAt(str.length() - 1)) {
+            tokens.remove( tokens.size() - 1 );
```
droazen: If you are splitting a single-character String containing only the delimiter (eg., the String `":"` with a delimiter of `':'`) this will (I believe) cause an empty tokens array to be returned. Does this match the behavior of `String.split()` with no limit argument?

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1137,83 @@ public static void forceJVMLocaleToUSEnglish() {
...
+        final List<String> result;
+
+        if ( str.isEmpty() ) {
+            result = new ArrayList<>(1);
+            result.add("");
+        }
+        else if ( delimiter.isEmpty() ) {
+            result = new ArrayList<>( str.length() );
+            for ( int i = 0; i < str.length(); ++i ) {
+                result.add( str.substring(i, i+1) );
```
droazen: Does `String.split()` do the same thing with an empty delimiter?

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -731,4 +731,84 @@ public void testGetDuplicatedItems(final Collection<?> collection, final Set<?>
...
+        for ( final String delim : wordsToSplitOn ) {
+            testCases.add( new Object[] { stringData, delim, Arrays.asList( stringData.split(delim) ) } );
+        }
+
+        // Create test cases for repeated substrings:
+        for ( final String delim : repeatedSubstringsToSplitOn ) {
+            testCases.add( new Object[] { stringData, delim, Arrays.asList( stringData.split(delim) ) } );
+        }
+
+        return testCases.toArray( new Object[][] {} );
```
pshapiro4broad: You can avoid copying here by making the dataprovider return `Iterator` and return `testCases.iterator` instead.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1137,83 @@ public static void forceJVMLocaleToUSEnglish() {
...
+                delimiterIdx = str.indexOf(delimiter, tokenStartIdx);
+                final String token = (delimiterIdx != -1 ? str.substring(tokenStartIdx, delimiterIdx) : str.substring(tokenStartIdx));
+                result.add(token);
+                tokenStartIdx = delimiterIdx + delimiter.length();
+            } while (delimiterIdx != -1);
+
+            // Check to see if the last entry is empty.
+            // String.split doesn't give us empty last elements of the token list, so we remove it.
+            if ( result.get(result.size() - 1).isEmpty() ) {
+                result.remove( result.size() - 1 );
```
droazen: What if the String ends with 2 (or more) consecutive delimiters? Couldn't you then have multiple trailing empty tokens?

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -731,4 +731,84 @@ public void testGetDuplicatedItems(final Collection<?> collection, final Set<?>
...
+                    new Object[] { stringData, "t",   Arrays.asList( stringData.split("t"  ) ) },
+                    new Object[] { stringData, "u",   Arrays.asList( stringData.split("u"  ) ) },
+                    new Object[] { stringData, "v",   Arrays.asList( stringData.split("v"  ) ) },
+                    new Object[] { stringData, "w",   Arrays.asList( stringData.split("w"  ) ) },
+                    new Object[] { stringData, "x",   Arrays.asList( stringData.split("x"  ) ) },
+                    new Object[] { stringData, "y",   Arrays.asList( stringData.split("y"  ) ) },
+                    new Object[] { stringData, "z",   Arrays.asList( stringData.split("z"  ) ) },
+                    new Object[] { stringData, " ",   Arrays.asList( stringData.split(" "  ) ) },
+                    new Object[] { stringData, "T",   Arrays.asList( stringData.split("T"  ) ) },
+                    new Object[] { stringData, "9",   Arrays.asList( stringData.split("9") ) }
```
droazen: You are still missing a number of important edge cases in your test data. Eg.,

-A String consisting only of delimiters: ":::::" with a delimiter of ":"
-Multiple consecutive delimiters at the beginning, middle, and/or end of the String ("::foo", "foo::", "foo::bar" with a delimiter of ":")

Try to enumerate all meaningfully-different arrangements of delimiter and non-delimiter in short Strings.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1139,85 @@ public static void forceJVMLocaleToUSEnglish() {
         final List<T> sorted = values.stream().sorted().collect(Collectors.toList());
         return sorted.get(sorted.size() / 2);
     }
+
+    /**
+     * Splits a String using indexOf instead of regex to speed things up.
```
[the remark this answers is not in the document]
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Yeah - I have a test case for it:

`{ stringData, "",    Arrays.asList( stringData.split(""   ) ) }`

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
jonn-smith: I added a few more test cases that I think handle this.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: I prefer the extra whitespace (the space inconsistency in the parens was a typo).

I've had to maintain a lot of poorly formatted code and having some extra line breaks makes it easier to visually differentiate between different blocks.

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
jonn-smith: Interesting. 

Fixed!

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Added in the test case.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Good point.  Fixed!

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Good point.  Fixed!

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
jonn-smith: Yes.  I added some more tests and modified the code to preserve the behavior of `String.split`, though now performance is somewhat degraded in the case where there are trailing repeated delimiters in the string to split.

src/test/java/org/broadinstitute/hellbender/utils/StringSplitSpeedUnitTest.java
```diff
@@ -0,0 +1,254 @@
...
+        printTimingString("\"Chars\"", javaSplitSingleCharStringTotalTime_ns, singleCharStringsToSplitOn.size());
+        System.out.println("HTSJDK ParsingUtils::split");
+        printTimingString("\"Chars\"", htsjdkSplitSingleCharTotalTime_ns, singleCharStringsToSplitOn.size());
+        System.out.println("GATK Utils::split");
+        printTimingString("Words", gatkSplitWordTotalTime_ns, wordsToSplitOn.size());
+        printTimingString("\"Chars\"", gatkSplitSingleCharStringTotalTime_ns, singleCharStringsToSplitOn.size());
+        System.out.println("================================================================================");
+    }
+
+    private static void printTimingTableMarkup(final long   javaSplitWordTotalTime_ns,
```
lbergelson: markdown

jonn-smith: Fixed!

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -731,4 +731,153 @@ public void testGetDuplicatedItems(final Collection<?> collection, final Set<?>
...
+        final int maxStringLength = 7;
+
+        // Create single character delimiter strings:
+        final String singleCharDelimiter = "o";
+        final List<String> singleCharTestStrings = Arrays.asList("X", singleCharDelimiter);
+        final List<List<String>> exhaustiveListsForSingleChar = Utils.makePermutations( singleCharTestStrings, maxStringLength, true );
+
+        // Create multi-character delimiter strings:
+        final String multiCharDelimiter = "oz";
+        // Must include the individual characters of the multi-char delimiter here:
```
droazen: Can you explain why?

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -731,4 +731,153 @@ public void testGetDuplicatedItems(final Collection<?> collection, final Set<?>
...
+        final List<Object[]> testCases = new ArrayList<>();
+
+        // Add single-char cases:
+        for ( final List<String> testCase : exhaustiveListsForSingleChar ) {
+            testCases.add( new Object[] { String.join( "", testCase ), singleCharDelimiter, Arrays.asList( String.join( "", testCase ).split(singleCharDelimiter)) } );
+        }
+
+        // Add multi-char cases:
+        for ( final List<String> testCase : exhaustiveListsForMultiChar ) {
+            testCases.add( new Object[] { String.join( "", testCase ), multiCharDelimiter, Arrays.asList( String.join( "", testCase ).split(multiCharDelimiter)) } );
```
droazen: Same comment as above applies to this `DataProvider` -- it would be easier to generate the expected output using `String.split()` within the `@Test` method itself.

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -731,4 +731,153 @@ public void testGetDuplicatedItems(final Collection<?> collection, final Set<?>
...
+        final String frontDelimiterTestString = "::::::::::1:23::::::::::456:7890";
+        final String middleDelimiterTestString = "1:23::::::::::456:7890";
+        final String backDelimiterTestString = "1:23::::::::::456:7890::::::::::";
+        final String fullDelimiterTestString = "::::::::::1:23::::::::::456:7890::::::::::";
+
+        final List<Object[]> testCases = new ArrayList<>();
+        testCases.addAll(
+                Arrays.asList(
+                    new Object[] { "", "",   Arrays.asList( "".split(""  ) ) },
+                    new Object[] { "", "1",  Arrays.asList( "".split("1" ) ) },
```
droazen: You are missing test cases with a *non-empty* string plus an *empty* delimiter. This would exercise the following code path, which I think might be entirely uncovered by tests:

```
        else if ( delimiter.isEmpty() ) {
            result = new ArrayList<>(str.length());
            for ( int i = 0; i < str.length(); ++i ) {
                result.add(str.substring(i, i + 1));
            }
        }
```

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1137,4 +1138,99 @@ public static void forceJVMLocaleToUSEnglish() {
...
+     * If given an empty delimiter, will return each character in the string as a token.
+     * This method produces the same results as {@link String#split(String)} and {@code String.split(String, 0)},
+     * but has been measured to be ~2x faster (see {@code StringSplitSpeedUnitTest} for details).
+     *
+     * @param str               The {@link String} to split.
+     * @param delimiter         The delimiter used to split the {@link String}.
+     * @param expectedNumTokens The number of tokens expected (used to initialize the capacity of the {@link ArrayList}).
+     * @return A {@link List} of {@link String} tokens.
+     */
+    public static List<String> split(final String str, final String delimiter, final int expectedNumTokens) {
```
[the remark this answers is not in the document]
magicDGS: I think that the `expectedNumTokens` is useful because initializing the array to the correct number of tokens might speed-up parsing of delimited files with fixed number of columns (e.g. gff/gtf). I understand your concern about the `limit` arg in the `String.split()`, but a clear javadoc should minimize confusion...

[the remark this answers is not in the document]
jonn-smith: I agree with @droazen - it confuses the user because the interface becomes the same as the original `String::split`.

I'll take it out and default it.

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
jonn-smith: Added it in!

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
```diff
@@ -731,4 +731,153 @@ public void testGetDuplicatedItems(final Collection<?> collection, final Set<?>
...
+                    new Object[] { ":", ":",   Arrays.asList( ":".split(":"  ) ) },
+                    new Object[] { stringData, "",    Arrays.asList( stringData.split(""   ) ) },
+                    new Object[] { stringData, "1",   Arrays.asList( stringData.split("1"  ) ) },
+                    new Object[] { stringData, "a",   Arrays.asList( stringData.split("a"  ) ) },
+                    new Object[] { stringData, "b",   Arrays.asList( stringData.split("b"  ) ) },
+                    new Object[] { stringData, "c",   Arrays.asList( stringData.split("c"  ) ) },
```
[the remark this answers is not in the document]
jonn-smith: Sure.  Fixed!

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
jonn-smith: Fixed!

src/test/java/org/broadinstitute/hellbender/utils/UtilsUnitTest.java
jonn-smith: The call to `Utils.makePermutations` takes a list of objects and permutes them, so I'm passing individual characters in to create the permuted string.

I've added a comment as well.

Issue #3725, opened 2017-10-20
GATK4.beta.6 HaplotypeCallerSpark does not output bgzipped VCFs


chapmanb:
I was testing the latest beta.6 release and ran into an issue that HaplotypeCallerSpark no longer outputs bgzipped VCF outputs. Specifying `--output outfile.vcf.gz` produces a plain unzipped VCF. This worked in the beta.5 release so appears to be due to recent changes. Here is a small self-contained test case that reproduces the issue:

https://s3.amazonaws.com/chapmanb/testcases/gatk/gatk4_spark_output.tar.gz

Please let me know if I can provide anything else that would help. Thanks as always for all the improvements and work on GATK.

lbergelson:
@chapmanb Thanks for the report.  That's unexpected.  Hopefully it was something simple that we can fix quickly.

tomwhite:
The change is because we started using VariantsSparkSink to write VCFs on the cluster, rather than writing them in a single thread from the driver (which doesn't scale).

VariantsSparkSink only supports .bgz extensions currently, not .gz. So if you change the output extension to .bgz it will work.

Gzip is not splittable so if possible we'd avoid outputting it at all. Hail for example will not load .gz unless the force option is used. (There are actually two force options, one to read it as regular gzip, the other to read it as bgzip, so you have to know which flavour of gzip it is...)

We could do one of the following:

1. Throw an error if the extension is .gz.
2. Write bgzipped output to the .gz file.
3. Write regular gzipped output to the .gz file.

Thoughts?

chapmanb:
Tom -- thanks so much for looking at this. My preference would be 2, writing bgzipped output to the .gz file. `vcf.gz` is pretty widely used to mean bgzipped (then tabixed). I personally never use plain gzip for this. It would be cool if we actually used a different extension for the two to not confuse ourselves but dealing with it the way people expect is also important. Thanks again.

tomwhite:
Thanks Brad. Here's a PR that implements 2: #3791

PR #3791, opened 2017-11-03
Write VCF files with .gz extension in block gzip format.

tomwhite:
Addresses #3725


src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSink.java
```diff
@@ -178,7 +178,7 @@ private static void writeVariantsSingle(
             final VCFHeader header, final boolean writeGvcf, final List<Integer> gqPartitions, final int defaultPloidy, final int numReducers) throws IOException {
 
         final Configuration conf = ctx.hadoopConfiguration();
-        if (outputFile.endsWith(BGZFCodec.DEFAULT_EXTENSION)) {
+        if (outputFile.endsWith(BGZFCodec.DEFAULT_EXTENSION) || outputFile.endsWith(".gz")) {
```
lbergelson: It's crazy that we don't have a consistent way we check this.  The best list seems to be in `AbstractFeatureReader.hasBlockCompressedExtensions` but there are a whole bunch of places that only accept .gz, or some combination of gz other extensions.

lbergelson:
@tomwhite Thanks for tracking this one down!

PR #3794, opened 2017-11-03
adds arg validation to make sure margin is non-negative in SimpleInte…

SHuang-Broad:
…rval.overlapsWithMargin(...)

I find myself being tempted to use negative -1 for that, but it shouldn't make sense for most people, unless there's other strange people out there like me.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
@@ -225,10 +225,15 @@ public final class SimpleInterval implements Locatable, Serializable {
       * This is the same as plain overlaps if margin=0.
       *
       * @param other interval to check
-      * @param margin how many bases may be between the two interval for us to still consider them overlapping.
+      * @param margin how many bases may be between the two interval for us to still consider them overlapping; must be non-negative
       * @return true if this interval overlaps other, otherwise false
+      * @throws IllegalArgumentException if margin is negative
       */
      public boolean overlapsWithMargin(final Locatable other, final int margin) {
+         if ( margin < 0 ) {
+             throw new IllegalArgumentException("given margin is negative: " + margin +
+                     "\tfor this: " + toString() + "\tand that: " + (other == null ? "other is null" : other.toString()));
+         }
          if ( other == null || other.getContig() == null ) {
              return false;
          }
--- a/src/test/java/org/broadinstitute/hellbender/utils/SimpleIntervalUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/SimpleIntervalUnitTest.java
@@ -208,6 +208,25 @@ public final class SimpleIntervalUnitTest extends GATKBaseTest {
                 "overlap() returned incorrect result for intervals " + firstInterval + " and " + secondInterval);
     }
 
+    @DataProvider()
+    private Object[][] overlapsWithMarginExpectingException() {
+        final SimpleInterval standardInterval = new SimpleInterval("1", 10, 20);
+        final SimpleInterval middleInterval = new SimpleInterval("1", 100, 200);
+
+        return new Object[][] {
+                { standardInterval, new SimpleInterval("2", 10, 20), -100 },
+                { standardInterval, new SimpleInterval("1", 30, 50), -9 },
+                { standardInterval, new SimpleInterval("1", 30, 50), -10 },
+                { middleInterval, new SimpleInterval("1", 50, 90), -9 },
+                { middleInterval, new SimpleInterval("1", 50, 90), -10 },
+        };
+    }
+    @Test(dataProvider = "overlapsWithMarginExpectingException", expectedExceptions = IllegalArgumentException.class)
+    public void testOverlapWithMarginExpectingException( final SimpleInterval firstInterval, final SimpleInterval secondInterval,
+                                                         int margin) {
+        firstInterval.overlapsWithMargin(secondInterval, margin);
+    }
+
     @DataProvider(name = "IntervalContainsData")
     public Object[][] getIntervalContainsData() {
         final SimpleInterval containingInterval = new SimpleInterval("1", 10, 20);


src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
```diff
@@ -229,6 +230,8 @@ public boolean overlaps( final Locatable other ) {
       * @return true if this interval overlaps other, otherwise false
       */
      public boolean overlapsWithMargin(final Locatable other, final int margin) {
+         ParamUtils.isPositiveOrZero(margin,
+                 "given margin is negative: " + margin + "\tfor this: " + toString() + "\tand that: " + other.toString());
```
droazen: This will end up calling `toString()` on both intervals on every call to this method, even if margin is >= 0. Since this is the kind of method that is called millions of times, this could be a performance issue.

 You should refactor this change so that the `toString() calls only happen if there's an actual error and you are going to throw.

SHuang-Broad:
changed. @droazen not sure if this is what you want

src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
```diff
@@ -225,10 +225,14 @@ public boolean overlaps( final Locatable other ) {
       * This is the same as plain overlaps if margin=0.
       *
       * @param other interval to check
-      * @param margin how many bases may be between the two interval for us to still consider them overlapping.
+      * @param margin how many bases may be between the two interval for us to still consider them overlapping; assumed non-negative
```
droazen: `assumed` -> `must be`

src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
```diff
@@ -225,10 +225,14 @@ public boolean overlaps( final Locatable other ) {
...
       * @return true if this interval overlaps other, otherwise false
+      * @throws IllegalArgumentException if margin is negative
       */
      public boolean overlapsWithMargin(final Locatable other, final int margin) {
+         if ( margin < 0 )
```
droazen: Add curly braces around if body -- omitting curly braces for single-line ifs can lead to bugs later down the line.

droazen: Also, this check should come after the checks for null, otherwise `other.toString()` could cause a `NullPointerException`

src/test/java/org/broadinstitute/hellbender/utils/SimpleIntervalUnitTest.java
```diff
@@ -208,6 +208,26 @@ public void testOverlapWithMargin( final SimpleInterval firstInterval, final Sim
                 "overlap() returned incorrect result for intervals " + firstInterval + " and " + secondInterval);
     }
 
+    @DataProvider(name = "overlapsWithMarginExpectingException")
```
droazen: If you omit an explicit name for your `DataProvider`, it defaults to the method name (which is slightly less typo-prone and usually what you want anyway)

src/test/java/org/broadinstitute/hellbender/utils/SimpleIntervalUnitTest.java
```diff
@@ -208,6 +208,26 @@ public void testOverlapWithMargin( final SimpleInterval firstInterval, final Sim
...
+    private Object[][] overlapsWithMarginExpectingException() {
+        final SimpleInterval standardInterval = new SimpleInterval("1", 10, 20);
+        final SimpleInterval middleInterval = new SimpleInterval("1", 100, 200);
+
+        return new Object[][] {
+                { standardInterval, new SimpleInterval("2", 10, 20), -100, false },
+                { standardInterval, new SimpleInterval("1", 30, 50), -9, false },
+                { standardInterval, new SimpleInterval("1", 30, 50), -10, true },
+                { middleInterval, new SimpleInterval("1", 50, 90), -9, false },
+                { middleInterval, new SimpleInterval("1", 50, 90), -10, true },
```
droazen: There isn't actually an expected result for these test cases, since you expect them all to throw, so you can omit the final `expectedOverlapResult` boolean, and just call `firstInterval.overlapsWithMargin(secondInterval, margin);` without the assert in the body of the test method.

SHuang-Broad:
@droazen I've change to using the suggested simple `if`.
Please take a look again. Thanks!

src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
SHuang-Broad: done with braces

SHuang-Broad: for moving checking after null.
I thought about it, but did the margin check before null because I think if a margin is provided with negative value, that indicates a logic error has been made upstream so should warn about that.
But I now check for null on `other` in the error message. 
Do you agree?

src/test/java/org/broadinstitute/hellbender/utils/SimpleIntervalUnitTest.java
SHuang-Broad: done. Thanks for catching that quick copy-and-paste error.

src/test/java/org/broadinstitute/hellbender/utils/SimpleIntervalUnitTest.java
SHuang-Broad: ahh! learned something new today. Thanks!

SHuang-Broad:
@droazen back to you.

src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
```diff
@@ -225,10 +225,15 @@ public boolean overlaps( final Locatable other ) {
       * This is the same as plain overlaps if margin=0.
       *
       * @param other interval to check
-      * @param margin how many bases may be between the two interval for us to still consider them overlapping.
+      * @param margin how many bases may be between the two interval for us to still consider them overlapping; must be non-negative
       * @return true if this interval overlaps other, otherwise false
+      * @throws IllegalArgumentException if margin is negative
       */
      public boolean overlapsWithMargin(final Locatable other, final int margin) {
+         if ( margin < 0 ) {
```
lbergelson: Why not use `ParamUtils.isPositiveOrZero`?

src/test/java/org/broadinstitute/hellbender/utils/SimpleIntervalUnitTest.java
```diff
@@ -208,6 +208,25 @@ public void testOverlapWithMargin( final SimpleInterval firstInterval, final Sim
                 "overlap() returned incorrect result for intervals " + firstInterval + " and " + secondInterval);
     }
 
+    @DataProvider()
+    private Object[][] overlapsWithMarginExpectingException() {
+        final SimpleInterval standardInterval = new SimpleInterval("1", 10, 20);
+        final SimpleInterval middleInterval = new SimpleInterval("1", 100, 200);
+
+        return new Object[][] {
+                { standardInterval, new SimpleInterval("2", 10, 20), -100 },
```
lbergelson: This seems like a lot of tests for a pretty simple feature, It's unclear to me what all these cases are, and if they're all needed.

lbergelson:
Oh, heh, I opened this tab this morning and then reviewed it after it was already merged.  Sorry, please ignore.

src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
SHuang-Broad: That was the initial implementation, but there's a very reasonable performance concern for string construction in the error message, so changed to `if`

PR #3795, opened 2017-11-03
Fixing contig ordering issue when calling CombineBreakpoints

LeeTL1220:
- expanding unit tests to actually replicate the error if we have a regression.
- applied fix that sorts the output.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/IntervalUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/IntervalUtils.java
@@ -1115,9 +1115,6 @@ public final class IntervalUtils {
     // (end of shard-related code)
 
     /**
-     *
-     * For most uses, developers will want to call {@link IntervalUtils#combineBreakpointsWithSorting(List, List, SAMSequenceDictionary)} rather than this method.  This method has unsorted output.
-     *
      * Combine the breakpoints of multiple intervals and return a list of locatables based on the updated breakpoints.
      *
      * Suppose we have two lists of locatables:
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/utils/IntervalUtils.java
```diff
@@ -1293,7 +1298,7 @@ static public int endOfShard(int shardIndex, int shardSize) {
         final List<T> sortedLocatables1 = sortLocatablesBySequenceDictionary(locatables1, dictionary);
         final List<T> sortedLocatables2 = sortLocatablesBySequenceDictionary(locatables2, dictionary);
 
-        return combineBreakpoints(sortedLocatables1, sortedLocatables2);
+        return  sortLocatablesBySequenceDictionary(combineBreakpoints(sortedLocatables1, sortedLocatables2), dictionary);
```
samuelklee: I actually think that just one method named `combineAndSortBreakpoints` that always returns a sorted result (even if only a single list is passed) would suffice.  I can't see a scenario in which you'd want a list without a guaranteed order?

Also, is sorting the two lists of locatables redundant?

LeeTL1220: Not redundant.  Since the combineBreakpoints makes assumptions about sorting.

Done

LeeTL1220:
@samuelklee Can you tell me if okay to merge?

src/main/java/org/broadinstitute/hellbender/utils/IntervalUtils.java
```diff
@@ -1154,15 +1154,16 @@ static public int endOfShard(int shardIndex, int shardSize) {
      *
      * Intervals are assumed to include the start and end bases.
      *
-     * Output contigs will be unordered.
-     *
      * @param locatables1 list of locatables
      * @param locatables2 list of locatables
+     * @param dictionary Sequence dictionary to base the sort.  The order of contigs/sequences in the dictionary is the order of the sorting here.
```
samuelklee: The first line of the javadoc above needs to be updated.

src/main/java/org/broadinstitute/hellbender/utils/IntervalUtils.java
```diff
@@ -1280,16 +1281,16 @@ static public int endOfShard(int shardIndex, int shardSize) {
     }
 
     /**
-     *  Same as {@link IntervalUtils::combineBreakpoints}, but sorts the inputs first.  Sorted versions are kept in a new list.
+     *  Same as {@link IntervalUtils::combineAndSortBreakpoints}, but sorts the inputs first.  Sorted versions are kept in a new list.
```
samuelklee: Sorry, maybe I was unclear in the last review---I don't think this method is necessary, right?  Isn't it pretty much equivalent to `combineAndSortBreakpoints`?  (And if not, the method names don't make clear the difference between the two.)  You will need to make that method public, though.

src/main/java/org/broadinstitute/hellbender/utils/IntervalUtils.java
samuelklee: I removed this method and replaced it everywhere with your `combineAndSortBreakpoints`, and all tests passed.  Sorting the copies of the two input lists is pretty much redundant if you are going to sort the final list, right?

LeeTL1220:
@samuelklee Take 2 is pushed

samuelklee:
Thanks, looks good to me now!

Issue #3788, opened 2017-11-02
Huge -L seems to take longer than it should


davidbenjamin:
I'm running SelectVariants with an interval list L of 600,000 loci and an excluded interval list XL of 160 loci.  Based on restricting L and XL it seems that the runtime scales as |L|x|XL|, which comes out to over an hour.  The runtime does not depend at all on the size of the vcf in the -V argument.  That is, once traversal begins it's basically instantaneous, but processing the intervals takes forever.

Naively, one could make this scale as |L| + |XL| by querying each variant in the input vcf against L and then against XL.  For reasons @cmnbroad points out this is not a great solution, but it seems like because (i) L and XL are ordered and (ii) each L overlaps with only a sparse subset of XL and vice versa that this quadratic scaling is not necessary.

For my case I can get around this easily by preprocessing V with -XL, then running SelectVariants with -L.

@lbergelson was also in this conversation.

droazen:
To make sure I understand, are you saying that it took over an hour for traversal to even start in this case?

lbergelson:
@davidbenjamin Could you upload your interval files somewhere so we can reproduce this.

davidbenjamin:
@droazen That's right -- the whole thing is over an hour but traversal itself is a few seconds.

davidbenjamin:
@lbergelson @cmnbroad @droazen I investigated further and to my surprise the excluded intervals are a red herring.  The time is linear in |L| and I just assumed it was linear in |XL| as well, but the command took an hour *without* any `-XL` at all.  So the issue is just that processing a large interval list seems to take longer than it ought to.  Here's the full command, with paths:

```bash
input_vcf=/dsde/working/davidben/mutect/speed/realign/subsetted_truth.vcf
intervals=/dsde/working/davidben/mutect/speed/realign/inactiveRegions.list

#same behavior for master builds on 5-16, 7-10, and 9-6
gatk=/dsde/working/davidben/gatk_builds/master-11-1.jar
java -jar $gatk SelectVariants -V $input_vcf -L $intervals -selectType SNP -O out.vcf
```

The output is
```bash
16:04:11.274 INFO  SelectVariants - HTSJDK Defaults.COMPRESSION_LEVEL : 5
16:04:11.275 INFO  SelectVariants - HTSJDK Defaults.USE_ASYNC_IO_READ_FOR_SAMTOOLS : false
[...]

Issue #3555, opened 2017-09-07
Picard interval list files with .list extension can't be used with -L


cmnbroad:
This Barclay [feature](https://github.com/broadinstitute/barclay/issues/25) automatically expands the contents of a file ending in ".list" whenever the target argument is a collection. This precludes the use of Picard interval list files ending in ".list" with -L in GATK, since they contain a sam header. The raw sam header lines wind up getting added as interval strings, which then fails parsing: A USER ERROR has occurred: Badly formed genome unclippedLoc: Failed to parse Genome Location string: @HD	VN:1.5: Problem parsing start/end value in interval string. Value was: 1.5

A short term GATK workaround is to use a file ending in one of the other known Picard interval list extensions (.interval_list, .intervals, or .picard) instead, but we should find a better fix for this since .list seems to be commonly used.

Tools such as GetHetCoverage, which take an interval list in an argument typed as a File (--snpIntervals), are able to consume the interval file because the target argument is not a collection, so the auto-expansion is not triggered.

I expect this issue could cause more problems in Picard as well once Barclay is the default parser there.

cmnbroad:
@sooheelee This is what I mentioned in slack.

magicDGS:
I've already suggested in https://github.com/broadinstitute/barclay/pull/28#discussion_r98629000), when the feature was implemented, that this can be an issue in downstream projects. Maybe it is still not late to change the extension to ".arg_list" to be sure that it is what the user requested.

cmnbroad:
After some discussion, we're planning to change the built-in extension used by Barclay for automatic expansion from ".list" to  ".args".

sooheelee:
Thanks for letting me know Chris. I'm on vacay and just viewing the email thread that starts with mention of my name and not the whole github thread. If I can help out with something specific, do let me know.

magicDGS:
I added a PR to Barclay (https://github.com/broadinstitute/barclay/pull/95) to both change the extension and to switch off the behaviour if need it. That may solve other errors in the future too, if other downstream project want to use the extension for other purposes.

cmnbroad:
This is fixed by the Barclay upgrade in https://github.com/broadinstitute/gatk/pull/3804, but we really should add an explicit integration test that uses a Picard interval list with a .list extension that includes a header (i.e., src/test/resources/small_unmerged_picard_intervals.list, but this is only used in a unit test).

droazen:
We fixed this.

PR #3804, opened 2017-11-07
Upgrade to Barclay 1.2.2.

cmnbroad:
This changes the argument expansion file extension from ".list" to ".args." Fixes https://github.com/broadinstitute/gatk/issues/3710, https://github.com/broadinstitute/gatk/issues/3555, and https://github.com/broadinstitute/gatk/issues/3788.


README.md
```diff
@@ -135,6 +135,9 @@ You can download and run pre-built versions of GATK4 from the following places:
 
 * To run a non-Spark tool, or to run a Spark tool locally, the syntax is: **`./gatk-launch ToolName toolArguments`**.
 
+* Tool arguments that allow multiple values, such as -I, can be supplied on the command line using a file with the extension “.args”. Each line of the file should contain a
```
pshapiro4broad: It looks like straight quotes are used elsewhere in this document (in both the text and in code blocks). Was it intentional to use curly quotes here?

src/test/java/org/broadinstitute/hellbender/utils/read/ReadUtilsIntegrationTest.java
```diff
@@ -48,8 +50,9 @@ public void testCreatePathSAMWriter(
       final boolean createMD5,
       final boolean expectIndex) throws Exception {
 
-    final Path outputPath = BucketUtils.getPathOnGcs(getGCPTestStaging() + "samWriterTest" + outputExtension);
-    final Path md5Path = BucketUtils.getPathOnGcs(getGCPTestStaging() + "samWriterTest" + outputExtension + ".md5");
+    final String randomizedOutputPathName = getGCPTestStaging() + "samWriterTest" + outputExtension + Integer.toString(new Random().nextInt());
```
pshapiro4broad: Unnecessary call to `Integer.toString()`, since `+` will automatically convert any object to a string in this context. (In general, I think `String.valueOf()` is preferable over `X.toString()` since it has the same behavior and doesn't need to be changed if the type of its argument ever changes.)

I'm not sure what this code is doing, but it seems like you'd want `outputExtension` after the random number, not before it.

README.md
cmnbroad: Thats what I get for pasting from TextEdit. Fixed.

PR #3810, opened 2017-11-08
Change tests to not clobber GCS files being used by other tests.

cmnbroad:
The GCS ReadUtils tests are failing intermittently on the Barclay upgrade branch, probably due to filename collision when tests are running in parallel on Travis because the tests don't use unique temporary filenames.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
@@ -177,11 +177,11 @@ public final class BucketUtils {
 
     /**
      * Get a temporary file path based on the prefix and extension provided.
-     * This file (and possible indexes associated with it will be scheduled for deleton on shutdown
+     * This file (and possible indexes associated with it) will be scheduled for deletion on shutdown
      *
      * @param prefix a prefix for the file name
      *               for remote paths this should be a valid URI to root the temporary file in (ie. gcs://hellbender/staging/)
-     *               there is no guarantee that this will be used as the root of the tmp file name, a local prefix may be placed in the tmp folder for exapmle
+     *               there is no guarantee that this will be used as the root of the tmp file name, a local prefix may be placed in the tmp folder for example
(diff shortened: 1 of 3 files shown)


src/test/java/org/broadinstitute/hellbender/utils/nio/GcsNioIntegrationTest.java
```diff
@@ -47,7 +47,8 @@ public void openPublicFile() throws IOException {
     @Test(groups={"bucket"})
     public void writePrivateFile() throws IOException {
         try {
-            final String dest = getGCPTestStaging() + "GcsNioIntegrationTest-writePrivateFile-test" + new Random().nextInt();
+            final String dest = BucketUtils.randomRemotePath(
```
lbergelson: why not use `BucketUtils.getTempFilePath()`?  That will autodelete as well.

cmnbroad: @lbergelson God idea. I also changed the ReadUtils test (the other place in this PR that I had modified) to use BucketUtils.getTempFilePath() as well, which in turn allowed me to remove some of the local file deletion code and a try/finally block from that file.

Also, I notice that ".md5" is not it included in the list of file extensions that BucketUtils.getTempFilePath() auto-deletes. So I added that. Can  you please take a look the new commit ? Thx.

src/main/java/org/broadinstitute/hellbender/utils/gcs/BucketUtils.java
```diff
@@ -177,11 +177,11 @@ public static void deleteFile(String pathToDelete) throws IOException {
 
     /**
      * Get a temporary file path based on the prefix and extension provided.
-     * This file (and possible indexes associated with it will be scheduled for deleton on shutdown
+     * This file (and possible indexes associated with it will be scheduled for deletion on shutdown)
```
lbergelson: I think the `)` should go after *associated with it*

Issue #3816, opened 2017-11-09
gatk-launch --list output should go to stdout, not stderr


droazen:
gatk-launch --list output currently goes to stderr. It should go to stdout instead to make it easier to grep for tools.

Reported by @sooheelee

PR #3818, opened 2017-11-09
Send command line output to the correct stream.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/3816.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -222,15 +222,15 @@ public class Main {
      */
     protected void handleStorageException(final StorageException exception) {
         // HTTP error code
-        System.out.println("code:      " + exception.getCode());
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -285,10 +285,10 @@ private static CommandLineProgram extractCommandLineProgram(final String[] args,
         classes.addAll(simpleNameToClass.values());
 
         if (args.length < 1) {
-            printUsage(classes, commandLineName);
+            printUsage(System.err, classes, commandLineName);
```
droazen: Does this case actually represent an error?

lbergelson: lets move this into the normal help output case, so we don't have a division between them.

cmnbroad: Done. So now it goes to stdout.

PR #3819, opened 2017-11-09
make kmer counts depend on coverage

### LINKED ISSUE #3607: SV pipeline run on NA12878 hg19 has a very slow laggard assembly task
When running the current version of the SV pipeline against the NA12878_PCR-_30X bam file aligned to hg19/b37, stage 10 is held up by a single assembly task that takes much longer than the others. Unfortunately I don't have exact timings yet, and haven't identified the assembly issue that's causing  the problem, but I wanted to log this as an issue so we don't forget about it. It's possible that other samples will have similar outlier assemblies so it would be good to track down cases like this so that we can try to identify and remove pathological intervals.

@tedsharpe if you find yourself with free time could you take a look at this? Whatever is going wrong with it might be fixed or relevant to your work on debugging assemblies and kmer gathering (if we're lucky perhaps your 'diffuse k-mer' fix will solve this issue).

tedsharpe:
Closes #3607 
In cleaning rare or ubiquitous kmers from the list for local assemblies, we now make our min and max counts depend on the measured coverage.
We also count the number of partitions in which a kmer appears, and discard those that appear in too many dispersed locations.  This is a big win for keeping the assemblies small and local, and allows us to do more of them (there are fewer "too bigs").
Finally, we reorder the assemblies so that we do the big ones first.  This keeps us from being hung up on stragglers.
This reduces by a percent or so the number of events called, but preliminary results suggest that these were mostly false positives caused by assembling homologous regions together.


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
```diff
@@ -93,14 +107,16 @@ public ReadMetadata( final Set<Integer> crossContigIgnoreSet,
...
+        contigNameToID = buildContigNameToIDMap(header.getSequenceDictionary());
         contigIDToName = buildContigIDToNameArray(contigNameToID);
         readGroupToLibrary = buildGroupToLibMap(header);
         this.nReads = nReads;
         nRefBases = header.getSequenceDictionary().getSequences()
                 .stream().mapToLong(SAMSequenceRecord::getSequenceLength).sum();
+        avgReadLen = (int)(coverage * nRefBases / nReads);
         this.maxReadsInPartition = maxReadsInPartition;
         this.coverage = coverage;
+        meanBaseQuality = 30.f;
```
cwhelan: Extract this as a constant, ie DEFAULT_MEAN_BASE_QUALITY_FOR_TESTING

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
```diff
@@ -225,26 +245,38 @@ public float getZishScore( final String readGroup, final int fragmentSize ) {
...
     }
+    public float getMeanBaseQuality() { return meanBaseQuality; }
+    public float getAccurateKmerCoverage( final int kSize ) {
+        // i.e., p = (1 - 10**(-Q/10))**K
+        final float probabilityOfAnAccurateKmer =
+                (float)Math.exp(kSize * Math.log1p(-Math.pow(10.,meanBaseQuality/-10.)));
+        return coverage * probabilityOfAnAccurateKmer;
+    }
+
+    public int getMeanPartitionSpan() {
```
cwhelan: It looks like you're returning the median here (which is probably better) but maybe forgot to rename the method.

cwhelan: There's `MathUtils.median()` that also handles cases like having an odd number of values properly in case you want to consider using that.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/experimental/CalcMetadata.java
```diff
@@ -0,0 +1,36 @@
...
+import org.broadinstitute.hellbender.tools.spark.sv.evidence.ReadMetadata;
+import org.broadinstitute.hellbender.tools.spark.sv.evidence.SVReadFilter;
+
+import java.util.Collections;
+
+@CommandLineProgramProperties(summary="Dump some data about the reads.",
+        oneLineSummary="Dump some data about the reads.",
+        programGroup=StructuralVariationSparkProgramGroup.class)
+@BetaFeature
+public class CalcMetadata extends GATKSparkTool {
```
cwhelan: I know this is only for testing / interactive debugging right now, but it would be great to flesh this out into a full tool that calculates and writes out read metadata and then provide a `readMetadata` method that can parse the output file into a full `ReadMetadata` object. That would let us write and/or downstream tools that would like to have the read metadata but not have to recompute it from the reads. Could you maybe add a ticket to do that in the future?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/experimental/CalcMetadata.java
```diff
@@ -0,0 +1,36 @@
...
+    @Override protected void runTool( final JavaSparkContext ctx ) {
+        final StructuralVariationDiscoveryArgumentCollection.FindBreakpointEvidenceSparkArgumentCollection params =
+                new StructuralVariationDiscoveryArgumentCollection.FindBreakpointEvidenceSparkArgumentCollection();
+        final ReadMetadata readMetadata =
+                new ReadMetadata(Collections.emptySet(), getHeaderForReads(),
+                        10000, getUnfilteredReads(),
+                        new SVReadFilter(params), LogManager.getLogger(CalcMetadata.class));
+        System.out.println("Effective Kmer coverage for K=" + params.kSize + " = " +
+                readMetadata.getAccurateKmerCoverage(params.kSize));
+        ReadMetadata.writeMetadata(readMetadata, "/dev/stdout");
```
cwhelan: `/dev/stdout` isn't really portable, is it? even though this is a debugging/test tool, I'd rather have it write to a proper file.

build.gradle
```diff
@@ -238,7 +238,7 @@ dependencies {
     }
 
     compile 'org.broadinstitute:gatk-bwamem-jni:1.0.3'
-    compile 'org.broadinstitute:gatk-fermilite-jni:1.0.0'
+    compile 'org.broadinstitute:gatk-fermilite-jni:1.0.0-rc6'
```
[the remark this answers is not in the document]
tedsharpe: Careful error checking of all system calls in the c-language part, and a bug fix (source contig was marked RC when it was not, and vice versa).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FermiLiteAssemblyHandler.java
```diff
@@ -0,0 +1,73 @@
...
+    public AlignedAssemblyOrExcuse apply( final Tuple2<Integer, List<SVFastqUtils.FastqRead>> intervalAndReads ) {
+        final int intervalID = intervalAndReads._1();
+        final List<SVFastqUtils.FastqRead> readsList = intervalAndReads._2();
+
+        final int fastqSize = readsList.stream().mapToInt(FastqRead -> FastqRead.getBases().length).sum();
+        if (fastqSize > maxFastqSize) {
+            return new AlignedAssemblyOrExcuse(intervalID, "no assembly -- too big (" + fastqSize + " bytes).");
+        }
+
+        final String fastqName = String.format("%s/%s.fastq",fastqDir, AlignedAssemblyOrExcuse.formatAssemblyID(intervalID));
```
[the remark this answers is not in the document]
tedsharpe: @cwhelan I can restore this argument as optional, but:
1) It isn't optional in our launch script -- we always specify it.
2) I thought that Valentin's genotyper relied on the fastq's, and so we'd always want to produce them.
Would you prefer to retain it as optional?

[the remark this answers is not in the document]
cwhelan: Those are both valid points.. I just wasn't sure why the behavior was being changed here. Since we don't really know the performance characteristics of @vruano's genotyper yet (and depending on how that goes we might need to look for a more efficient solution to data transfer) I thought we might want to keep the flexibility for now. If it's not a pain to revert I think I'd rather keep it optional.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
```diff
@@ -71,17 +75,27 @@ public ReadMetadata( final Set<Integer> crossContigIgnoreSet,
...
         nReads = combinedMaps.values().stream().mapToLong(LibraryRawStatistics::getNReads).sum();
         final long nReadBases = combinedMaps.values().stream().mapToLong(LibraryRawStatistics::getNBases).sum();
+        avgReadLen = (int)(nReadBases / nReads);
         nRefBases = header.getSequenceDictionary().getSequences()
                 .stream().mapToLong(SAMSequenceRecord::getSequenceLength).sum();
-        coverage = (int)((nReadBases + nRefBases/2) / nRefBases); // rounding the coverage number
+        final float cov = (float)nReadBases / nRefBases;
+        if ( cov >= MIN_COVERAGE ) coverage = cov;
+        else {
+            logger.warn("Apparent coverage (" + cov + ") too low.  Pretending it's 10x.");
```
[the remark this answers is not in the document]
cwhelan: It's true we don't have a great place to put this sort of documentation yet. I guess I was thinking just the javadoc for `FindBreakpointEvidenceSpark` and `StructuralVariationDiscoveryPipelineSpark` for now, just so that it's somewhere.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
tedsharpe: I realize that efficiency doesn't much matter here -- it's a small array.  But I just can't bring myself to box a mess of ints into a Collection, convert that to an array of doubles, and do an elaborate sampling procedure when I can just sort a small array of primitives.
It's just me.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FermiLiteAssemblyHandler.java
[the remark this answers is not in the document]
tedsharpe: roger

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/experimental/CalcMetadata.java
tedsharpe: I hacked in a solution that uses Kryo serialization to dump a file that could be deserialized by some other tool.  Better than nuthin'.

PR #3820, opened 2017-11-09
 Added denoising tools for ModelSegments CNV pipeline.

### LINKED ISSUE #3153: Refactor WGS coverage collection and GC annotation.
This should be more parallel to the WES workflow and should take an interval list of precomputed bins as input.  Consider using ReadWalkerSpark.

### LINKED ISSUE #3246: Eliminate Targets from new/legacy CNV tools.
The use of Targets to refer to genomic intervals is unnecessary and confusing.  It obfuscates the fact that most of the tools and code can be applied to not only counts from WES targets, but also counts from WES baits, WGS bins, etc.  Requiring that Targets be named also adds unnecessary storage and memory burden.  We should just use SimpleIntervals everywhere.  We should also get rid of the target file format.

In terms of external visibility, we can just rename tools and edit javadoc.  Internally, there will be many classes that need to be both renamed and refactored.  I instead suggest that we rebuild new versions of the classes and tools as necessary in the tools/copynumber package.

- [ ] Rename tools: AnnotateTargets -> AnnotateIntervals, TargetCoverageSexGenotyper -> ReadCountSexGenotyper.  
- [ ] Deprecate tools: CalculateTargetCoverage, ConvertBedToTargetFile, and PadTargets will be replaced by @asmirnov239's new CollectReadCounts tool and on-the-fly padding specified by --interval_padding parameters.
- [ ] Deprecate target file format and change all other affected file formats.
- [ ] Refactor/rename/rebuild classes.

### LINKED ISSUE #3349: Have SparkGenomeReadCounts produce a HDF5 for coverage in addition to the tsv and make it readable during PoN creation.
This would probably yield a non-trivial increase in speed, since much of the time spent is I/O.

### LINKED ISSUE #3356: Expose more optional parameters for CreatePanelOfNormals in CNV WDL.

### LINKED ISSUE #3570: Add option to run somatic pair workflow without a PoN but with GC correction.
This will be easier once the new binning is in.

samuelklee:
First commit:
-Added CreateReadCountPanelOfNormals tool.  This is an update of CreatePanelOfNormals.  Related code is written from scratch.
-Added DenoiseReadCounts tool.  This is an update of NormalizeSomaticReadCounts.  Related code is written from scratch.
-Added AnnotateIntervals tool.  This is an update of AnnotateTargets.  Related code (e.g., GCBiasCorrector) is mostly ported and does not have to be closely re-reviewed.  I naively introduced RecordCollection and LocatableCollection classes that are analogous to SampleRecordCollection and SampleLocatableCollection, respectively, for collections that are not tied to a sample (e.g., GC-content annotations); we can go back and refactor these classes later.
-SVDDenoisingUtils contains many package-private helper methods for filtering and denoising without unit tests.  This is intentional.  I have verified that this code exactly reproduces the old PoN results down to the 1E-16 level (with the discrepancy coming from the removal of redundant pseudoinverse operations).  Rather than writing or porting unit tests for this code, I think it is best if we simply do not reuse this code or make non-trivial changes to it going forward.  We can add unit tests later if we have extra time on our hands...
-SparkGenomeReadCounts now outputs TSV and HDF5.
-Added some tests for SimpleCountCollection, HDF5SimpleCountCollection, and some disabled tests for HDF5Utils.
-Miscellaneous cleanup and boy scout activities.

Second commit:
-Updated coverage collection in germline and legacy somatic CNV WDLs to use only integer read counts and account for changes to SparkGenomeReadCounts.
-Added tasks for PreprocessIntevals, AnnotateIntervals, and CollectFragmentCounts.
-Renamed and moved some files.

Closes #3570.
Closes #3356.
Closes #3349.
Closes #3246.
Closes #3153.


samuelklee:
@LeeTL1220 do you have some bandwidth to review this?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/CreateReadCountPanelOfNormals.java
```diff
@@ -0,0 +1,262 @@
...
+
+    private static List<SimpleInterval> getIntervalsFromFirstReadCountFile(final Logger logger,
+                                                                           final List<File> inputReadCountFiles) {
+        final File firstReadCountFile = inputReadCountFiles.get(0);
+        logger.info(String.format("Retrieving intervals from first read-count file (%s)...", firstReadCountFile));
+        final SimpleCountCollection readCounts = SimpleCountCollection.read(firstReadCountFile);
+        return readCounts.getIntervals();
+    }
+
+    private RealMatrix constructReadCountMatrix(final Logger logger,
```
samuelklee: Remove logger argument.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/CreateReadCountPanelOfNormals.java
```diff
@@ -0,0 +1,262 @@
...
+                "List of input read-count files cannot contain duplicates.");
+        inputReadCountFiles.forEach(IOUtils::canReadFile);
+        if (numEigensamplesRequested > inputReadCountFiles.size()) {
+            logger.warn(String.format("Number of eigensamples (%d) is greater than the number of input samples (%d); " +
+                            "the number of samples retained after filtering will be used instead.",
+                    numEigensamplesRequested, inputReadCountFiles.size()));
+        }
+    }
+
+    private static List<SimpleInterval> getIntervalsFromFirstReadCountFile(final Logger logger,
```
samuelklee: Remove logger argument and make non-static.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/LocatableCollection.java
```diff
@@ -0,0 +1,89 @@
...
+                               final Function<DataLine, RECORD> recordFromDataLineDecoder,
+                               final BiConsumer<RECORD, DataLine> recordToDataLineEncoder) {
+        super(inputFile, mandatoryColumns, recordFromDataLineDecoder, recordToDataLineEncoder);
+        validateIntervals(getRecords());
+    }
+
+    //check for ordering, duplicates, and overlaps
+    private static <T extends Locatable> void validateIntervals(final List<T> records) {
+        if (!Ordering.from(LEXICOGRAPHICAL_ORDER_COMPARATOR).isStrictlyOrdered(records)) {
+            throw new IllegalArgumentException("Records fwere not strictly sorted in lexicographical order.");
```
samuelklee: fwere -> were

src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/LocatableCollection.java
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @author Samuel Lee &lt;slee@broadinstitute.org&gt;
+ */
+public abstract class LocatableCollection<RECORD extends Locatable> extends RecordCollection<RECORD> {
+    public static final Comparator<Locatable> LEXICOGRAPHICAL_ORDER_COMPARATOR = IntervalUtils.LEXICOGRAPHICAL_ORDER_COMPARATOR;
+
+    /**
+     * Records are sorted using {@code LEXICOGRAPHICAL_ORDER_COMPARATOR}.
+     */
+    public LocatableCollection(final List<RECORD> records,
```
samuelklee: Make the corresponding constructors in the other collection classes public.

samuelklee:
@LeeTL1220 What's the status of the review?

LeeTL1220:
@samuelklee Resuming review.  Not quite done yet.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/annotation/AnnotatedInterval.java
```diff
@@ -0,0 +1,73 @@
+package org.broadinstitute.hellbender.tools.copynumber.annotation;
+
+import htsjdk.samtools.util.Locatable;
```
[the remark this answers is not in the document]
LeeTL1220: Or can we reconcile the two (not in this PR)?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/coverage/denoising/svd/SVDDenoisingUtils.java
```diff
@@ -0,0 +1,482 @@
...
+            public double visit(int sampleIndex, int intervalIndex, double value) {
+                return value - sampleLog2Medians[sampleIndex];
+            }
+        });
+        logger.info("Sample read counts standardized.");
+
+        return result;
+    }
+
+    private static PreprocessedStandardizedResult preprocessPanel(final RealMatrix readCounts,
```
LeeTL1220: This method is really long.  Can it be simplified or broken into sub-methods?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/LocatableCollection.java
```diff
@@ -0,0 +1,89 @@
...
+    public LocatableCollection(final File inputFile,
+                               final TableColumnCollection mandatoryColumns,
+                               final Function<DataLine, RECORD> recordFromDataLineDecoder,
+                               final BiConsumer<RECORD, DataLine> recordToDataLineEncoder) {
+        super(inputFile, mandatoryColumns, recordFromDataLineDecoder, recordToDataLineEncoder);
+        validateIntervals(getRecords());
+    }
+
+    //check for ordering, duplicates, and overlaps
+    private static <T extends Locatable> void validateIntervals(final List<T> records) {
```
LeeTL1220: Can you move the overlap detection (only) to IntervalUtils?  Others (myself included) could use that code.

.travis.yml
```diff
@@ -11,7 +11,7 @@ env:
   - TEST_TYPE=cloud UPLOAD=true
   - TEST_TYPE=integration TEST_DOCKER=true TEST_VERBOSITY=minimal
   - TEST_TYPE=unit TEST_DOCKER=true TEST_VERBOSITY=minimal
-  - RUN_CNV_SOMATIC_WDL=true
+  - RUN_CNV_SOMATIC_LEGACY_WDL=true
```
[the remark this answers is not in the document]
samuelklee: It's already in the PR to follow.

scripts/cnv_wdl/cnv_common_tasks.wdl
```diff
@@ -18,13 +18,12 @@ task PadTargets {
     String filename = select_first([targets, ""])
     String base_filename = basename(filename, ".tsv")
 
-    command <<<
-        echo ${filename}; \
+    command {
```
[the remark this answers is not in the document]
samuelklee: OK.  Will fix this across all WDLs in the next PR, if that's OK with you.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/CreateReadCountPanelOfNormals.java
```diff
@@ -0,0 +1,262 @@
...
+ *     will be performed by {@link GCBiasCorrector} before filtering and SVD.  GC-content information for the
+ *     intervals will be stored in the PoN and used to perform explicit GC-bias correction identically when denoising
+ *     case samples.  Note that if annotated intervals are not provided, it is still likely that GC-bias correction
+ *     is implicitly performed by the SVD denoising (i.e., some of the principal components arise from GC bias).
+ * </p>
+ *
+ * <h3>Examples</h3>
+ *
+ * <pre>
+ * gatk-launch --javaOptions "-Xmx4g" CreateReadCountPanelOfNormals \
```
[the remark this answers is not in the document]
samuelklee: Typically I just use a bash for loop, but I agree that this could be nice.  However, this seems like something we should standardize across all tools that accept multiple `-I`, rather than just for this tool.  Is this something that we already do?  If not, maybe @vdauwera or @cmnbroad have some thoughts?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/annotation/AnnotatedInterval.java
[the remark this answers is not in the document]
samuelklee: Because I wrote this code before that class existed!  We can reconcile later, will file an issue.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/LocatableCollection.java
samuelklee: This code is somewhat cribbed from `checkForOverlaps` in `HashedListTargetCollection`.  The naive linear check for overlaps is only valid if the list is sorted, so there's no real way to usefully decouple the checks...

src/main/java/org/broadinstitute/hellbender/tools/copynumber/coverage/denoising/svd/SVDDenoisingUtils.java
samuelklee: Unfortunately, refactoring this code so that it is memory efficient (using the filter arrays) and broken up into methods makes it much less readable and not very reusable, in my opinion (I actually wrote the code that way originally).

.travis.yml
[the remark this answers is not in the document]
LeeTL1220: Great.

scripts/cnv_wdl/cnv_common_tasks.wdl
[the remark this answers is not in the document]
LeeTL1220: Totally OK.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/annotation/AnnotatedInterval.java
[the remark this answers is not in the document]
LeeTL1220: Great.  That is fine for this PR.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/coverage/denoising/svd/SVDDenoisingUtils.java
LeeTL1220: If methods are passed by reference, how can you not even break it up into sub-methods?

Regardless, I am okay with this long method, if you add a comment as to why.

samuelklee: The signatures of some (but not all) of the sub-methods must then include both filter arrays, which makes them inconsistent and awkward to reuse.

I'll add some comments about why this code is long and a warning not to split it up or refactor without good reason.

samuelklee:
Thanks, @LeeTL1220!  Made some minor updates, will merge when tests pass.

LeeTL1220:
@samuelklee Sounds great.

Issue #3809, opened 2017-11-08
Add documentedfeature tag to annotation modules and make content consistent


sooheelee:
E.g. SOR annotation does not show up in GATK4 versioned docs currently. Code at
 <https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandOddsRatio.java> needs `@DocumentedFeature` tag at top.

magicDGS:
I think that no `VariantAnnotation` is documented yet, because there is no even a plugin. It will be nice to have them anyway...

sooheelee:
Hey @magicDGS, since I'm on the forum documentation side of things, I'm not familiar with a plugin. Is this something @cmnbroad typically takes care of?

cmnbroad:
@magicDGS is correct that no annotations are tagged as `@DocumentedFeature` yet, but thats just because nobody has done it. The plugin (which I think @jamesemery is planning to implement soon), is a completely separate issue - the `@DocumentedFeature` annotations can be added either way.

magicDGS:
I think that the annotation plugin is assigned to @jamesemery here: https://github.com/broadinstitute/gatk/issues/3287

But of course it can be documented before it is a plugin. I just wanted to point out that all of them should be annotated properly...

sooheelee:
I'll ask @jamesemery to explain to me what he is doing fits in with what we need to do on the documntation side.

Is there anything else besides the `@DocumentedFeature` that I should be adding to each annotation document to annotate properly?

sooheelee:
Also, I'll remove the annotation epic label.

cmnbroad:
@sooheelee You'll need to add new string constants for the group name and group summary for annotations to HelpConstants.java, and then use them in each `@DocumentedFeature`, along with a summary specific to each annotation. Take a look at how [read filters](https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/engine/filters/ReadFilterLibrary.java#L32) are done as a model. Let me know if you need help with that part.

sooheelee:
## SOP 

[1] Add feature tag and define elements to all annotation modules in <https://github.com/broadinstitute/gatk/tree/master/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator>. E.g. the tag looks thus:
```
[...]

PR #3835, opened 2017-11-14
Add ANNOTATORS group under Utilities and tag annotation modules for documentation

sooheelee:
I'll be adding documented feature tags to the 57 annotation modules. Before I get to these, I need a new ANNOTATORS category to exist in HelpConstants.java. 

I've taken the liberty to name the category ANNOTATORS. Here is the relevant info I added to HelpConstants.java:

- group name variable and descriptor: DOC_CAT_ANNOTATORS = "Annotation Modules"
- group summary variable and descriptor: DOC_CAT_ANNOTATORS_SUMMARY = "Annotations available to HaplotypeCaller, Mutect2 and VariantAnnotator"
- super category: Utilities (same group as read filters)

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
@@ -66,7 +66,7 @@ public final class HelpConstants {
     public final static String DOC_CAT_VAREVAL = "Variant Evaluation Tools";
     public final static String DOC_CAT_VARMANIP = "Variant Manipulation Tools";
     
-    public final static String DOC_CAT_ANNOTATORS = "Annotation Modules";
+    public final static String DOC_CAT_ANNOTATORS = "Variant Annotations";
     public final static String DOC_CAT_ANNOTATORS_SUMMARY = "Annotations available to HaplotypeCaller, Mutect2 and VariantAnnotator";
     
     public final static String DOC_CAT_EXAMPLE = "Example Tools";
--- a/src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
@@ -65,7 +65,10 @@ public final class HelpConstants {
     public final static String DOC_CAT_VARDISC = "Variant Discovery Tools";
     public final static String DOC_CAT_VAREVAL = "Variant Evaluation Tools";
     public final static String DOC_CAT_VARMANIP = "Variant Manipulation Tools";
-
+    
+    public final static String DOC_CAT_ANNOTATORS = "Annotation Modules";
+    public final static String DOC_CAT_ANNOTATORS_SUMMARY = "Annotations available to HaplotypeCaller, Mutect2 and VariantAnnotator";
+    
     public final static String DOC_CAT_EXAMPLE = "Example Tools";
     public final static String DOC_CAT_EXAMPLE_SUMMARY = "Example tools that show developers how to implement new tools";
 
@@ -98,6 +101,7 @@ public final class HelpConstants {
 
             // supercat Utilities
             groupToSuperCategory.put(DOC_CAT_READFILTERS, DOC_SUPERCAT_UTILITIES);
+            groupToSuperCategory.put(DOC_CAT_ANNOTATORS, DOC_SUPERCAT_UTILITIES);
 
             // supercat Exclude
             groupToSuperCategory.put(DOC_CAT_EXAMPLE, DOC_SUPERCAT_EXCLUDE);
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQuality.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQuality.java
@@ -2,8 +2,10 @@ package org.broadinstitute.hellbender.tools.walkers.annotator;
 
 import com.google.common.primitives.Ints;
 import htsjdk.variant.variantcontext.VariantContext;
+import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hellbender.utils.GATKProtectedMathUtils;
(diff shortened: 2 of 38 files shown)


sooheelee:
Since you showed me the ropes, would you also be willing to review my changes, @cmnbroad? Much appreciated.

sooheelee:
Actually, I think I can add all my changes to this one pull request and tests should pass.

cmnbroad:
@sooheelee Looks like you're on the right track, but yes, you should probably add the actual tags into this same PR and then we'll review.

sooheelee:
Will do.

sooheelee:
@cmnbroad I've tagged the docs that I determined needed tagging. The rest are listed in <https://github.com/broadinstitute/gatk/issues/3809> as unchecked. I hope the checks go over all the changes and not each commit one-by-one.

cmnbroad:
@sooheelee A couple things:

The builds are failing because there are no DocumentedFeature import statements in these files. Each one needs to have an "import org.broadinstitute.barclay.help.DocumentedFeature;"  statement with the other imports at the top of the file (they're usually listed alphabetically). IntelliJ would do this for you automatically, so I'm assuming you didn't use IntelliJ to make these changes; doing so would also call your attention to any syntax errors that may have been introduced.

One minor point - we try to be consistent about use of whitespace; either use spaces around the "=" everywhere, or don't use spaces. Most of these changes use a mixture:

`@DocumentedFeature(groupName=HelpConstants.DOC_CAT_ANNOTATORS, groupSummary=HelpConstants.DOC_CAT_ANNOTATORS_SUMMARY, summary = "Median base quality of bases supporting each allele (MBQ)")`

Finally, I'm not sure there is value in using separate commits for every change like this. Its fine if its you're preferred work style, but its not necessary.

sooheelee:
Thanks for the comments @cmnbroad.

> The builds are failing because there are no DocumentedFeature import statements in these files. Each one needs to have an "import org.broadinstitute.barclay.help.DocumentedFeature;" statement with the other imports at the top of the file (they're usually listed alphabetically). IntelliJ would do this for you automatically, so I'm assuming you didn't use IntelliJ to make these changes; doing so would also call your attention to any syntax errors that may have been introduced.

I made changes via a browser, directly through Github. Making changes this way is actually faster for such minor changes than using IntelliJ. As I mentioned last week, I need to set up my IntelliJ again, starting with our institutional account key.

> One minor point - we try to be consistent about use of whitespace; either use spaces around the "=" everywhere, or don't use spaces. Most of these changes use a mixture:

I copied the spacing from a filters document. I assumed there was some significance to the differential spacing, but if there is not, I would be happy to be consistent.

> Finally, I'm not sure there is value in using separate commits for every change like this. Its fine if its you're preferred work style, but its not necessary.
Merge state

This is a side effect of making changes directly through Github. I'll search and see if there is something I can do so all the changes get saved to the same commit.

src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
```diff
@@ -65,7 +65,10 @@ public static String forumPost(String post) {
     public final static String DOC_CAT_VARDISC = "Variant Discovery Tools";
     public final static String DOC_CAT_VAREVAL = "Variant Evaluation Tools";
     public final static String DOC_CAT_VARMANIP = "Variant Manipulation Tools";
-
+    
+    public final static String DOC_CAT_ANNOTATORS = "Annotation Modules";
```
cmnbroad: I know we called these "Annotation Modules" in GATK3, but the word "module" seems out of place to me. I think everywhere else in the doc we just refer to them as annotations. My suggestion would be to just call them "Variant Annotations".

src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
```diff
@@ -65,7 +65,10 @@ public static String forumPost(String post) {
...
+    public final static String DOC_CAT_ANNOTATORS_SUMMARY = "Annotations available to HaplotypeCaller, Mutect2 and VariantAnnotator";
```
cmnbroad: I wonder if we should say "Variant Callers" and "VariantAnnotator" rather than enumerating here, so we don't cause confusion by leaving out Sparkk tools (HaplotypeCallerSpark).

sooheelee:
@cmnbroad this is ready to review. All other changes, e.g. content and pulling additional elements with code to populate the docs, can be a separate issue.

src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
sooheelee: I think that the annotations extend to Spark versions should be understood.

src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
sooheelee: Let's ask @vdauwera to chime in  here.

vdauwera: works for me

sooheelee: Making this change. Thanks for the suggestion!

sooheelee:
Incorporated Chris' suggestion with Geraldine's approval. Chris already reviewed so this one is ready to merge.

src/main/java/org/broadinstitute/hellbender/utils/help/HelpConstants.java
vdauwera: I'd like to make these more future-proof by removing the tool names (a dev could add a new tool that annotates things at any time). Can't make it about variant callers only (VA is not a caller, and again, want to be future proof) so should try  to describe this in a more generic way.

magicDGS: I support @vdauwera idea - I will use the annotators for some of my downstream tools, and this summary is tool-specific although other tools might use them...

PR #3836, opened 2017-11-14
fix to allow sample renaming to work correctly in GenomicsDBImport

lbergelson:
it seems like the only change necessary was in our code
removing pointless asserts
adding tests the deal with renaming

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -144,9 +144,9 @@ public final class GenomicsDBImport extends GATKTool {
             shortName = SAMPLE_NAME_MAP_LONG_NAME,
             doc = "Path to file containing a mapping of sample name to file uri in tab delimited format.  If this is " +
                     "specified then the header from the first sample will be treated as the merged header rather than " +
-                    "merging the headers, and the sample names will be taken from this file.  This is a performance optimization " +
-                    "that relaxes the normal checks for consistent headers.  Using vcfs with incompatible headers may result " +
-                    "in silent data corruption.",
+                    "merging the headers, and the sample names will be taken from this file.  This may be used to rename " +
+                    "input samples. This is a performance optimization that relaxes the normal checks for consistent " +
+                    "headers.  Using vcfs with incompatible headers may result in silent data corruption.",
             optional = true,
             mutex = {StandardArgumentDefinitions.VARIANT_LONG_NAME})
     private String sampleNameMapFile;
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -494,11 +494,11 @@ public final class GenomicsDBImport extends GATKTool {
                                                                                    final int batchSize, final int lowerSampleIndex) {
         final SortedMap<String, FeatureReader<VariantContext>> sampleToReaderMap = new TreeMap<>();
         logger.info("Starting batch input file preload");
-        final List<Future<FeatureReader<VariantContext>>> futures = new ArrayList<>();
+        final Map<String, Future<FeatureReader<VariantContext>>> futures = new LinkedHashMap<>();
         final List<String> sampleNames = new ArrayList<>(sampleNametoPath.keySet());
         for(int i = lowerSampleIndex; i < sampleNametoPath.size() && i < lowerSampleIndex+batchSize; ++i) {
             final String sampleName = sampleNames.get(i);
-            futures.add(inputPreloadExecutorService.submit(() -> {
+            futures.put(sampleName, inputPreloadExecutorService.submit(() -> {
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -492,11 +492,11 @@ public Object onTraversalSuccess() {
                                                                                    final int batchSize, final int lowerSampleIndex) {
         final SortedMap<String, FeatureReader<VariantContext>> sampleToReaderMap = new TreeMap<>();
```
droazen: Add explicit docs to the sample name map argument clarifying whether the sample names must match the names in the header.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -261,12 +269,131 @@ private static File createInOrderSampleMap() {
...
+                .map( pair -> pair.getKey() + "\t" + pair.getValue())
+                .collect(Collectors.joining("\n")));
+    }
+
+    @DataProvider
+    public static Iterator<Object[]> getRenameCombinations() {
+        final Map<String,String> noRemapping = new LinkedHashMap<>();
+        noRemapping.put("s1", "s1");
+        noRemapping.put("s2", "s2");
+        noRemapping.put("s3", "s3");
```
droazen: We really could use a test case with more than a handful of samples -- say, 100. You could write a method to permute the values in a sample -> sample mapping.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -261,12 +269,131 @@ private static File createInOrderSampleMap() {
...
+        try(final GenomicsDBFeatureReader<VariantContext, PositionalBufferedStream> reader = getGenomicsDBFeatureReader(workspace, b37_reference_20_21)) {
+            final CloseableTribbleIterator<VariantContext> iterator = reader.iterator();
+            Assert.assertTrue(iterator.hasNext(), "expected to see a variant");
+            Assert.assertTrue(expectedSampleNames.size() > 0);
+            Assert.assertEquals(expectedSampleNames.size(), renamingMap.size());
+            iterator.forEachRemaining(vc -> {
+                   Assert.assertEquals(vc.getSampleNames(), expectedSampleNames);
+                expectedSampleNames.forEach( sample -> {
+                   Assert.assertEquals(vc.getGenotype(sample).getAnyAttribute(SAMPLE_NAME_KEY), renamingMap.get(sample));
+                   Assert.assertEquals(vc.getGenotype(sample).getAnyAttribute(ANOTHER_ATTRIBUTE_KEY), 10); //check another attribute just to make sure we're not mangling things
```
droazen: Have you confirmed that this test case fails without the fix to the tool?

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -261,12 +269,131 @@ private static File createInOrderSampleMap() {
...
+        final Map<String,String> sameInput = new LinkedHashMap<>();
+        sameInput.put("s1", "s1");
+        sameInput.put("s2", "s1");
+        sameInput.put("s3", "s1");
+
+
+        final Map<String,String> sameInputWeirdOrder = new LinkedHashMap<>();
+        sameInputWeirdOrder.put("s3", "s1");
+        sameInputWeirdOrder.put("s1", "s1");
+        sameInputWeirdOrder.put("s2", "s1");
```
droazen: What currently happens if there are duplicate sample names in the sample name map (ie., on the left side of the mapping)? Add a separate test case for this with (I would hope) an expected exception.

lbergelson: We have a test for this in `GenomicsDBImportUnitTest.testBadInputFiles`

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
lbergelson: A subset of the test cases fail without the fix.  It only applies to 1 code path and it depends on the batching, so many pass with / without the fix.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
lbergelson: added

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
lbergelson: opened #3861

Issue #3692, opened 2017-10-13
Load a conda environment from GATK


lucidtronix:
The Python scripts called by the PythonScriptExecutor will require python dependencies which can be managed within a conda environment.  Is there a way to load the appropriate conda environment from GATK so that users and unit tests can run the PythonScriptExecutor without worrying about wrangling python libraries. @samuelklee @mbabadi @cmnbroad Any ideas?

droazen:
@cmnbroad Could you look into our options here, when you get a chance?

cmnbroad:
@lucidtronix @mbabadi Update on this: There is a branch [here](https://github.com/broadinstitute/gatk/commits/cn_python_environment) that updates the docker image that we build on travis to include a full anaconda installation. That may be overkill in the long run but for now gives us a base environment for running integration tests on Travis. The environment managers I've looked at (conda and virtualenv) both appear to work by activating a shell environment. Ideally, we wouldn't need an intermediate shell between the Java code and the Python code, but thats still TBD. In parallel, I'm working on a streaming I/O solution for using input/output streams over standard I/O as the next incarnation of the PythonScriptExecutor for https://github.com/broadinstitute/gatk/issues/3698.

cmnbroad:
@lucidtronix @mbabadi @samuelklee I think the best solution would be to establish a single, common Python environment, with a single set of dependencies, that all GATK Python tools depend on. We would establish a single docker image that has all of these dependencies pip installed, and could also include a conda env for the GATK environment for users who don't want to use the docker image.

If we could do that, it would eliminate the need load per-tool conda environments. From what I've seen so far based on existing branches, the two environments we need (gCNV and CNN-VQSR) don't look that far apart in terms of dependencies. gCNV is using Theano, and CNN Tensorflow, but the rest looks [pretty close](https://docs.google.com/a/broadinstitute.org/spreadsheets/d/1RV7--uBQ0ctlXzMH09cmr0VimpZYIU68DdxJzE60y-c/edit?usp=sharing).

So a strawman proposal for the main components for a common environment would be:

Python           3.6
Numpy           >= 1.13.1
Scipy             1.0.0
Theano         .0.9.0
Tensorflow    1.4.0
Pymc3           3.1
Keras             2.1.1

Can you all chime on on whether you think we can converge in a single environment ? If so, it would greatly simplify things, and we can start with getting a docker image built for running travis tests.

samuelklee:
[...]

PR #3838, opened 2017-11-15
gCNV python computational kernel

mbabadi:
Code is polished and is ready for review.

Partial TODO list:
- [ ] (ISSUE TO BE MADE) generate denoising sample summary statistics output (distribution of bias factors, HMM likelihoods, distribution of unexplained variance)
- [ ] (ISSUE TO BE MADE) generate plots (training history, GC curves, histogram of unexplained variance, histogram of mean bias)
- [ ] (ISSUE TO BE MADE) prior optimization based on real data

In addition, I will also make issues for the TODO comments in the code.


mbabadi:
@samuelklee @lucidtronix @asmirnov239 @davidbenjamin This is a good time for all of us to decide on GATK-style python coding (how much type hinting? how much documentation? etc.). I am adding you to this thread so that we can all have a say.

**Code style**: The standard practice is to strictly follow [PEP-0008](https://www.python.org/dev/peps/pep-0008/). The python plugin for IntelliJ IDEA (=PyCharm) bark loudly when one deviates, which is pretty useful.

**Docstring style**: I have written most of it in reST-style. In retrospect, perhaps [NumPyStyle](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) would have been a better choice. If I have time now, I will switch. Definitely at some point.

You may want to wait until I check the I tick off "fill in missing docs" and "further code cleanup" boxes before you take a look at the code.

samuelklee:
This is great!  We can solicit feedback on style at the Methods meeting as well.  Feel free to spin off low-priority TODOs into issues that we can leave for after release.

lbergelson:
Are you going to use the built in `unittest` or something like nose2?

droazen:
@cmnbroad Can you have a look a this when you get a chance and provide high-level feedback related to eventual integration with the `PythonScriptExecutor` and any dependency-related issues? Thanks!

cmnbroad:
A couple of high-level things to start:

- I think the main short-term issue is getting some tests with inout/output files, and a test strategy, both for Python unit testing and for GATK integration testing. @mbabadi Do you have any tests for exercise this functionality ?
- We're probably going to want to standardize on a way to integrate tool logging (both the level and the stream) with Python logging. Since the stdout/stderr streams of the Python process will be redirected back to GATK via anonymous pipes, we need to ensure that verbose logging doesn’t block Python, which will happen when the buffering capability of the pipe is exceeded.

src/main/python/org/broadinstitute/hellbender/tools/gcnv/gcnvkernel/structs/metadata.py
```diff
@@ -265,21 +266,22 @@ def read_sample_coverage_metadata(self, input_file: str) -> List[str]:
             contig_list = []
             for row in reader:
                 row_num += 1
+                num_header_elems = None
```
samuelklee: I think this line breaks things.

samuelklee:
I managed to 1) wrap the cohort_determine_ploidy_and_depth.py script in a CLI tool (see the sl_gcnv_ploidy_cli branch), and 2) build the gcnvkernel package within a conda environment on gsa6 and run the tool, both without too much trouble.  A few things:

1) I don't see any logging from the python script (but then again, I didn't try anything out of the ordinary to get it working---should I have had to?)

2) I had to copy the cohort_determine_ploidy_and_depth.py script to the appropriate place in src/main/resources.  Perhaps that's a better place for python scripts/packages, rather than src/main/python?

src/main/python/org/broadinstitute/hellbender/tools/gcnv/gcnvkernel/structs/metadata.py
mbabadi: huh, I had fixed this earlier... weirdness.

mbabadi:
@samuelklee this is fantastic!

1. @cmnbroad any ideas about the absence of log messages from python? by default, the logger writes to `sys.stderr`. Perhaps I should set it to `sys.stdout`?

2. I think it makes sense to move python packages and scripts (or at least, the scripts) to src/main/resources.

mbabadi:
@cmnbroad 

1. We could write tests for some parts of the python package, e.g. those relating to I/O. The code also contains a number of stand-alone compiled theano functions that could be tested (e.g. forward-backward algorithm). Beyond that, most of the code juggle symbolic tensors that ultimately get compiled by theano into one giant _step function_. It is going to be hard to write surgical unit tests for such a code (and possibly unnecessary/wasteful). The ultimate requirement from the computational core is to (1) make correct inferences on simulated data, (2) pass certain sensitivity/specificity requirements on real data.

I vote for _not_ unit testing the python package (except for I/O), and to treat it as a black box with certain required specifications. The java-side integration tests can test most of the relevant cases.

2. Regarding logging, the scripts log both to console and optionally to a logfile at specified verbosity levels. At the moment, I am using the following formatter `%(asctime)s %(name)-12s %(levelname)-8s %(message)s`. We can easily change it to match GATK-style logs.

cmnbroad:
@mbabadi @samuelklee The PythonScriptExecutor, which is based on the same underlying code as the RScriptExecutor, redirects the child process's stdout/stderr to an anonymous pipe, and then swallows anything written there without propagating it, unless the process exits with a non-zero exit code.

There is a new [StreamingPythonExecutor](https://github.com/broadinstitute/gatk/commit/e59e41a0c90a169291be2317f96a1b31337ae9d9) (not yet merged) that allows you to keep the Python process alive; send and retrieve multiple commands, and retrieve the output. But it's more complicated to use, since its not one-shot, synchronous command execution.

If you're going to use the PythonScriptExecutor, then we should probably add a setting that causes it to redirect Python's stderr output to GATK's log output.

cmnbroad:
Also, I think we're going to need to establish an environment for running all GATK Python tools, like a single Docker image that doesn't depend on a per-tool Conda environment already being activated (unless there is a programmatic activation mechanism that is independent of the shell environment - I'm not aware of any ?). Ideally, we would agree on a set of "GATK Python" dependencies across all Python-dependent tools, establish a Docker image to reflect that, and then all of the Python tools would run on it. If we can't do that, I think we have a much more complicated problem, since activating a per-tool Conda (or virtualenv) environment from Java adds to the complexity.

cmnbroad:
@mbabadi On testing, I think what you suggest is workable. I thought you guys would want unit tests for your own purposes - but for travis we're going to have to depend on java-driven tests anyway, so as long as we have good tests that should work.

magicDGS:
Hello everyone: I am realizing that the GATK framework is going to have a lot of dependencies from java and python even if the simpler framework is the one need it. Maybe it is a good idea to start thinking about sub-modules within the same repository for the engine (maybe even separate the Spark framework), CNV...and create an independent artifact for every of them, and one combined one. Does it sound reasonable?

src/main/python/org/broadinstitute/hellbender/tools/gcnv/gcnvkernel/models/model_denoising_calling.py
```diff
@@ -0,0 +1,1027 @@
...
+                              type=float,
+                              help="Initial value of ARD prior precision relative to the typical interval-specific "
+                                   "unexplained variance scale")
+
+        process_and_maybe_add("num_gc_bins",
+                              type=int,
+                              help="Number of knobs on the GC curve")
+
+        process_and_maybe_add("gc_curve_sd",
+                              type=int,
```
samuelklee: int -> float

mbabadi: Done.

mbabadi:
@samuelklee The module can now save and load everything, including the state of the optimizer. This allows to making interesting inference pipelines. Here's a decent strategy for obtaining the global optimum (it works flawless on simulated data every time):

- In the first pass, one disables annealing and obtains the variational parameters in a thermal state. The temperature needs to be _high enough_ to allow most/all local minima to merge, though, not too high to allow copy numbers to travel too far away from baseline copy numbers. If this occurs, one must anneal very slowly in the next stage (see below). The results are checkpointed once converged.

- In the second pass, one makes another call to the CLI tool, this time w/ annealing enabled (starting from the same temperature) and starting from the checkpointed thermal results (model params, posteriors, adam(ax) state). The annealing rate must be slow enough to prevent thermal fluctuations from getting quenched (i.e. the evolution must be quasi-isothermal). One must look for a steady and linear rise of ELBO, such that when the annealing protocol ends, SNR quickly drops to values below 1.

In both runs, the learning rate must be very small (in the rate 0.01-0.05) such that we wouldn't have to worry about controlling stochastic noise. Adam(ax) quickly adjusts its moment estimates and compensates for the small learning rate, so this doesn't increase the training time significantly.

src/main/python/org/broadinstitute/hellbender/tools/gcnv/bin/case_denoising_calling.py
```diff
@@ -0,0 +1,213 @@
...
+                   default=argparse.SUPPRESS,
+                   help="(advanced) Path to exported optimizer state to take as the starting point")
+
+# add denoising config args
+# Note: we are hiding parameters that are either set by the model or are irrelevant to the case calling task
+gcnvkernel.DenoisingModelConfig.expose_args(
+    parser,
+    hide={
+        "--max_bias_factors",
+        "--psi_j_scale",
```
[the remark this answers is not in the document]
mbabadi: Yes, these are provided by the imported model.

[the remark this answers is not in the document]
mbabadi: oh wait... it must be psi_t_scale. i'll look into it when i get in.

samuelklee:
@lbergelson Do you have an opinion on the best way to pip install the gcnvkernel python package and dependencies for Travis testing?  I've verified that the pip install works within a basic conda environment with python=3.6.  We'll need to load this environment both for unit/integration tests as well as WDL tests.  As long as this is the only python environment we need, I think we can simply use the base environment in the Docker.  If more environments are required (e.g., for @lucidtronix), then maybe we'll need to be more clever for unit/integration tests, but we can still load them manually in the scripts that kick off the WDL tests.

cmnbroad:
@samuelklee Timely question - can you take a look at https://github.com/broadinstitute/gatk/issues/3692#issuecomment-348188451 and let me know what you think.

samuelklee:
I'm rebasing my sl_gcnv_ploidy branch on top of sl_wgs_acnv_headers and your branch.  Running into a new warning in the same conda environment:

    WARNING (theano.configdefaults): g++ not detected ! Theano will be unable to execute optimized C-implementations (for both CPU and GPU) and will default to Python implementations. Performance will be severely degraded. To remove this warning, set Theano flags cxx to an empty string.

MartonKN:
I went through the python code and done with my review. 
Thanks, Mehrtash, it is a really awesome code.

samuelklee:
@MartonKN Just so you know, this is the old copy of the branch.  The version that got merged was #3925.

PR #3854, opened 2017-11-17
upgrade htsjdk to 2.13.1

lbergelson:
upgrading to htsjdk 2.13.1
4 commits
1. is the bare upgrade to 2.13.0 with deprecations fixed
2. replacing our custom NON_REF_SYMBOLIC_ALLELE with the newly introduced one in HTSJDK Allele
3. replacing our ClassFinder with the newly introduced one in htsjdk
4. update to 2.13.1 which has an important bug fix


src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -1,10 +1,10 @@
 package org.broadinstitute.hellbender;
 
+import htsjdk.utils.ClassFinder;
```
droazen: Can we use the barclay version of `ClassFinder` for now?

lbergelson:
@droazen Changed to use the barclay ClassFinder and opened https://github.com/broadinstitute/barclay/issues/113

PR #3856, opened 2017-11-20
changing log level of SmithWaterman timing output

lbergelson:
going from debug -> info so that we'll always get timing information


lbergelson:
@gspowley This addresses your comment on the smith waterman pr.

PR #3863, opened 2017-11-20
Added in GC Content calculation and annotation to Funcotator.

### LINKED ISSUE #3859: Funcotator - implement GC content annotation.
Need to implement the GC content annotation.

This is calculated by taking the region consisting of a window 200 bases to either side of a variant locus and calculating the percentage of bases in that window that are G or C.

jonn-smith:
Added in GC Content calculation and annotation to Funcotator.

Fixes #3859.


jonn-smith:
PR #3952 obviates this PR.

PR #3872, opened 2017-11-27
Python streaming process executor.

cmnbroad:
This contains the streaming Python executor, implemented by a (Python-independent) StreamingProcessController. The StreamingProcessController, and the existing ProcessController are refactored to use a shared ProcessControllerBase, and changed from using raw Threads and Runnables to using an ExecutorService with Futures and Callables.


src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
```diff
@@ -0,0 +1,122 @@
...
+            previousBatch = null;
+        }
+        return lastCompleteBatch;
+    }
+
+    /**
+     * Terminate the async writer, cancelling any outstanding work.
+     * @return true
+     */
+    public boolean terminate() {
```
lbergelson: Should this be close instead?  Why do we want to cancel outstanding work instead of joining and waiting for it to complete?

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
```diff
@@ -0,0 +1,122 @@
...
+                throw new RuntimeException(e);
+            }
+            previousBatch = null;
+        }
+        return lastCompleteBatch;
+    }
+
+    /**
+     * Terminate the async writer, cancelling any outstanding work.
+     * @return true
```
lbergelson: The description of return is wrong, although I don't actually know why you care about the return value here or what the caller is supposed to do with the information.

src/main/java/org/broadinstitute/hellbender/utils/runtime/CapturedStreamOutputSnapshot.java
```diff
@@ -0,0 +1,50 @@
+package org.broadinstitute.hellbender.utils.runtime;
+
+import org.broadinstitute.hellbender.utils.Utils;
+
+import java.io.*;
+
+/**
+ * Stream output captured from a streaming stream.
```
lbergelson: stream stream stream stream :)

src/main/java/org/broadinstitute/hellbender/utils/runtime/CapturedStreamOutputSnapshot.java
```diff
@@ -0,0 +1,50 @@
...
+        super(settings, processStream, standardStream);
+
+        // strictly speaking, we could support these, but their use in streaming applications seems questionable
+        Utils.validate(!settings.getStreamLocations().contains(StreamLocation.Standard), "Stream snapshots can't go to standard out");
+        Utils.validate(!settings.getStreamLocations().contains(StreamLocation.File), "Stream snapshots can't go to a file");
+    }
+
+    /**
+     * Drain the input stream to keep the process from backing up until it's empty.
+     * File streams will be closed automatically when this method returns.
```
lbergelson: is this true?  I don't see the close here anymore.

lbergelson: this should describe what the return value means

src/main/java/org/broadinstitute/hellbender/utils/runtime/CapturedStreamOutput.java
```diff
@@ -89,10 +92,11 @@ public boolean isBufferTruncated() {
      *
      * @throws IOException When unable to read or write.
      */
-    public void readAndClose() throws IOException {
+    public boolean read() throws IOException {
```
lbergelson: what's the return value mean?

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessControllerBase.java
```diff
@@ -0,0 +1,157 @@
...
+{
+    private static final Logger logger = LogManager.getLogger(ProcessControllerBase.class);
+
+    protected enum ProcessStream {Stdout, Stderr}
+
+    // Tracks running controllers.
+    protected static final Set<ProcessControllerBase<?>> running = Collections.synchronizedSet(new LinkedHashSet<>());
+
+    // We can't use a fixed thread pool size, since we could run out if multiple controllers were to make blocking calls
+    final static protected ExecutorService executorService = Executors.newCachedThreadPool();
```
lbergelson: we should probably provide a thread factory here, that A: sets the threads as daemon threads, B) gives them some sane name,

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessControllerBase.java
```diff
@@ -0,0 +1,157 @@
...
+        } catch (Exception e) {
+            logger.error(e);
+        }
+        super.finalize();
+    }
+
+    protected class OutputCapture implements Callable<CAPTURE_POLICY> {
+        private final ProcessStream key;
+
+        //TODO: use controllerContextName
```
lbergelson: what's the deal with this todo?

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessSettings.java
```diff
@@ -48,6 +49,10 @@ public ProcessSettings(String[] command, boolean redirectErrorStream, File direc
         return command;
     }
 
+    public String getCommandString() {
+        return StringUtils.join(command, " ");
```
lbergelson: why not use String.join?

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessControllerBase.java
```diff
@@ -0,0 +1,157 @@
...
+
+        // Start the process running.
+        try {
+            process = builder.start();
+            running.add(this);
+        } catch (IOException e) {
+            final String message = String.format("Unable to start command: %s\nReason: %s",
+                    StringUtils.join(builder.command(), " "),
+                    e.getMessage());
+            throw new GATKException(message);
```
lbergelson: you should propagate the underlying exception here

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessControllerBase.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+public abstract class ProcessControllerBase<CAPTURE_POLICY extends CapturedStreamOutput>
+{
+    private static final Logger logger = LogManager.getLogger(ProcessControllerBase.class);
+
+    protected enum ProcessStream {Stdout, Stderr}
```
lbergelson: aren't enums usual ALL_CAPS?

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Example ReadWalker program that uses a Python streaming executor to stream summary data from a BAM
+ * input file to a Python process through a FIFO. The read data is accumulated in a List until a batch
+ * size threshold is reached, at which point the batch is handed off to an asynchronous write service,
+ * which writes the batch to the FIFO stream on a background thread. The Python process in turn just
+ * writes the data to an output file.
+ *
+ * <ol>
+ * <li>Opens a FIFO for writing.</li>
+ * <li>Creates an AsyncWriterervice</li>
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+ * <li>Writes a string of attributes for each read to the List until the batchSize theshold is reavched.</li>
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+ * </ol>
+ */
+@CommandLineProgramProperties(
+        summary = "Example/toy program that uses a Python script.",
+        oneLineSummary = "Example/toy program that uses a Python script.",
+        programGroup = ExampleProgramGroup.class,
+        omitFromCommandLine = true
+)
+public class ExampleStreamingPythonExecutor extends ReadWalker {
+    private final static String NL = String.format("%n");
```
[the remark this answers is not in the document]
cmnbroad: I don't think it matters as a practical thing, but it does seem clearer to use lineSeparator. Done.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+            doc = "Output file")
+    private File outputFile; // output file produced by Python code
+
+    @Argument(fullName = "batchSize",
+            doc = "Size of a batch for writing")
+    private int batchSize = 1000;
+
+    // Create the Python executor. This doesn't actually start the Python process, but verifies that
+    // the requestedPython executable exists and can be located.
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+    public void closeTool() {
+        if (asyncWriter != null) {
+            asyncWriter.terminate();
+        }
+        try {
+            fifoWriter.close();
+        } catch (IOException e) {
+            throw new GATKException("IOException closeing fifo writer", e);
+        }
+        pythonExecutor.terminate();
```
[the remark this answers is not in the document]
cmnbroad: Yes, for completeness it probably should. Done.

src/main/java/org/broadinstitute/hellbender/utils/python/PythonExecutorBase.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    /**
+     * Return an exception specific to this executor type, to be thrown on error conditions.
+     * @param message
+     */
+    @Override
+    public ScriptExecutorException getScriptException(final String message) {
+        return new PythonScriptExecutorException(message.toString());
```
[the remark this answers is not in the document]
cmnbroad: Yes. Fixed.

src/main/java/org/broadinstitute/hellbender/utils/python/StreamingPythonScriptExecutor.java
```diff
@@ -0,0 +1,274 @@
...
+     */
+    public void sendAsynchronousCommand(final String line) {
+        if (!line.endsWith(NL)) {
+            throw new IllegalArgumentException("Python commands must be newline-terminated");
+        }
+        spController.writeProcessInput(line);
+    }
+
+    /**
+     * See if any output is currently available. This is non-blocking, and can be used to determin if a blocking
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/python/StreamingPythonScriptExecutor.java
```diff
@@ -0,0 +1,274 @@
...
+     * Note that the output returned is somewhat non-deterministic, in that the only guaranty is that a prompt
+     * was detected on either stdout or stderr. It is possible for the remote process to produce the prompt on
+     * one stream (stderr or stdout), and additional output on the other; this method may detect the prompt before
+     * detecting the additional output on the other stream. Such output will be retained, and returned as part of
+     * the payload the next time output is retrieved.
+     *
+     * For this reason, program correctness should not rely on consuming data written by Python to standard streams.
+     *
+     * This should only be used for short, synchronous commands that produce output quickly. If no data has been
+     * sent from the process, this call blocks waiting for data, or the timeout (default 5 seconds) to be reached.
```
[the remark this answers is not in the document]
cmnbroad: Not for this class. I generally tried to discourage depending on timeouts (the only exception is in the AsynchronousStreamWriter class, which has a  blocking method with a timeout argument that waits for a batch to complete), and instead provide a non-blocking polling method "isOutputAvailable" see if there is any output available.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+    private List<String> batchList = new ArrayList<>(batchSize);
+    private int batchCount = 0;
+
+    @Override
+    public void onTraversalStart() {
+
+        // Start the Python process, and get a FIFO from the executor to use to send data to Python. The lifetime
+        // of the FIFO is managed by the executor; the FIFO will be destroyed when the executor is terminated.
+        pythonExecutor.start(Collections.emptyList());
+        final File fifoFile = pythonExecutor.getFIFOForWrite();
```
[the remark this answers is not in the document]
cmnbroad: Yeah, I have mixed feelings. I agree that its subtle and somewhat non-intuitive. I had a higher-level API in a previous incarnation, but making it general purpose got complicated, and it wound up obscuring the orchestration, which the caller needs to understand because the python code has be structured to conform to it. I'd lobby to wait until we build a couple of real tools before introducing something higher level.

src/main/java/org/broadinstitute/hellbender/utils/python/StreamingPythonScriptExecutor.java
```diff
@@ -0,0 +1,274 @@
...
+     *
+     * Longer-running, blocking commands can be executed using {@link #sendAsynchronousCommand}, in combination
+     * with {@link #isOutputAvailable}.
+     *
+     * @return ProcessOutput containing all accumulated output from stdout/stderr
+     * @throws UserException if a timeout occurs waiting for output
+     * @throws PythonScriptExecutorException if a traceback is detected in the outpu
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/python/StreamingPythonScriptExecutor.java
```diff
@@ -0,0 +1,274 @@
...
+            throw new UserException("A timeout ocurred waiting for output from the remote Python command.", e);
+        }
+    }
+
+    /**
+     * Terminate the remote process, closing the fifo if any.
+     *
+     * @return true if the process has been successfully terminated
+     */
+    public boolean terminate() {
```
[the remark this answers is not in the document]
cmnbroad: I'm not sure what possibilities could cause this to fail - it basically means the remote process is still alive. I think the close/closeable idea is a good one - I'll do the same thing in the controller as well.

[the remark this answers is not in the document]
cmnbroad: Changed to throw an exception - let me know if you still think it should be close, and/or Closable.

src/main/java/org/broadinstitute/hellbender/utils/python/StreamingPythonScriptExecutor.java
```diff
@@ -0,0 +1,274 @@
...
+     */
+    public <T> AsynchronousStreamWriterService<T> getAsynchronousStreamWriterService(final OutputStream streamWriter) {
+        Utils.nonNull(streamWriter);
+        return spController.getAsynchronousStreamWriterService(streamWriter);
+    }
+
+    /**
+     * Return a (not necessarily executable) string representing the current command line for this executor
+     * for error reporting purposes.
+     * @return Command line string.
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
```diff
@@ -0,0 +1,122 @@
...
+        previousBatch = null;
+    }
+
+    /**
+     * Request that a batch of items be written to the stream on a background thread. Any previously requested batch
+     * must have already been completed and retrieved via {@link #waitForPreviousBatchCompletion}.
+     *
+     * @param batchList a list of items to be written
+     * @param batchSize number of items to write
+     * @return {@code Future} representing this batch
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Waits for a batch that was previously initiated via {@link #startAsynchronousBatchWrite(List, int)}}
+     * to complete, flushes the target stream and returns the corresponding completed Future. The Future representing
+     * a given batch can only be obtained via this method once. If no work is outstanding, and/or the previous batch
+     * has already been retrieved, null is returned.
+     * @param timeout maximum time to wait for a response
+     * @param timeoutUnit units of {@code timeout}
+     * @return returns null if no previous work to complete, otherwise a completed Future
+     */
+    public Future<Integer> waitForPreviousBatchCompletion(long timeout, TimeUnit timeoutUnit) {
```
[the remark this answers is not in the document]
cmnbroad: Yes, this weirdness you're noticing is an artifact of a previous implementation, where this class had a queue of these. Only this method needs to return the Future.

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Request that a batch of items be written to the stream on a background thread. Any previously requested batch
+     * must have already been completed and retrieved via {@link #waitForPreviousBatchCompletion}.
+     *
+     * @param batchList a list of items to be written
+     * @param batchSize number of items to write
+     * @return {@code Future} representing this batch
+     */
+    public Future<Integer> startAsynchronousBatchWrite(final List<T> batchList, int batchSize) {
```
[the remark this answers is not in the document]
cmnbroad: Yes - its an artifact from a previous implementation. Fixed.

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
```diff
@@ -0,0 +1,122 @@
...
+        Utils.nonEmpty(batchList);
+
+        if (previousBatch != null) {
+            throw new IllegalStateException("Previous batch not yet complete");
+        }
+
+        previousBatch = executorService.submit(() -> {
+            try {
+                for (int i = 0; i < batchSize; i++) {
+                    T element = batchList.get(i);
```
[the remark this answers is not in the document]
cmnbroad: Right -fixed.

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
```diff
@@ -0,0 +1,122 @@
...
+        boolean isCancelled = true;
+        if (previousBatch != null) {
+            logger.warn("Cancelling outstanding asynchronous writing");
+            isCancelled = previousBatch.cancel(true);
+        }
+        previousBatch = null;
+        return isCancelled;
+    }
+
+    protected void finalize() throws Throwable {
```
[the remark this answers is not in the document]
cmnbroad: Agreed. Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
cmnbroad: I think if the caller is terminating without having waited for the last batch to complete, they're on a bad path or did something wrong. Maybe this should cancel and then throw ?

src/main/java/org/broadinstitute/hellbender/utils/runtime/CapturedStreamOutput.java
cmnbroad: Another artifact from a previous implementation. Removed.

src/main/java/org/broadinstitute/hellbender/utils/runtime/CapturedStreamOutputSnapshot.java
cmnbroad: Yes this should be removed as its not true of this implementation.

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessControllerBase.java
cmnbroad: Yes; this was like this when I started but there aren't too many references to it so I'll change it.

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessControllerBase.java
cmnbroad: Good idea - done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessControllerBase.java
cmnbroad: Yup. Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessSettings.java
cmnbroad: command is a String[], which matches the utils signature exactly.

lbergelson: oh, good point

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+     * @param line data to be written to the remote process
+     */
+    public void writeProcessInput(final String line) {
+        try {
+            // Its possible that we already have completed futures from output from a previous command that
+            // weren't consumed before we found the prompt at the last synchronization point. If so, drop them
+            // before we issue a new command (since retaining them could confound synchronization on output from
+            // this command), and warn.
+            if (stdErrFuture != null && stdErrFuture.isDone()) {
+                //logger.warn("Dropping stale stderr output: {}", stdErrFuture.get().getBufferString());
```
lbergelson: lets remove these commented out lines or make them debug level or something

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+                processJournal.writeLogMessage("Dropping stale stderr output: " + stdErrFuture.get().getBufferString());
+                stdErrFuture = null;
+            }
+            if (stdOutFuture != null && stdOutFuture.isDone()) {
+                //logger.warn("Dropping stale stdout output: {}", stdOutFuture.get().getBufferString());
+                processJournal.writeLogMessage("Dropping stale stdout output: " + stdOutFuture.get().getBufferString());
+                stdOutFuture = null;
+            }
+        } catch (InterruptedException e) {
+            throw new RuntimeException(String.format("Interrupted retrieving stale future: " + line, e));
```
lbergelson: I'm curious why you're using RuntimeExceptions, it seems like either a GatkException or a specific ProcessControllerException would be better?

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+        final ProcessSettings mkFIFOSettings = new ProcessSettings(new String[]{"mkfifo", fifoTempFileName});
+        mkFIFOSettings.getStdoutSettings().setBufferSize(-1);
+        mkFIFOSettings.setRedirectErrorStream(true);
+        final ProcessController mkFIFOController = new ProcessController();
+        final ProcessOutput result = mkFIFOController.exec(mkFIFOSettings);
+        final int exitValue = result.getExitValue();
+
+        fifoFile = new File(fifoTempFileName);
+        if (exitValue != 0) {
+            throw new GATKException(String.format("Failure creating FIFO (%s) got exit code (%d)", fifoTempFileName, exitValue));
```
lbergelson: If this triggers it would be really helpful to dump the stderr/ stdout from the process as part of the error.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+     * destroying it if necessary.
+     *
+     * @return true if the process was successfully destroyed
+     */
+    public boolean terminate() {
+        closeFIFO();
+        tryDestroy();
+        boolean exited = false;
+        try {
+            exited = process.waitFor(10, TimeUnit.SECONDS);
```
lbergelson: Should this be extracted to a constant somewhere?

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+    }
+
+    /**
+     * Stops the process from running and tries to ensure the process is cleaned up properly.
+     * NOTE: sub-processes started by process may be zombied with their parents set to pid 1.
+     * NOTE: capture threads may block on read.
+     */
+    @Override
+    protected void tryDestroy() {
+        if (stdErrFuture != null && !stdErrFuture.isDone()) {
```
lbergelson: It doesn't look like this is actually killing the process anywhere?  Just closing the various outputs?

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+        tryDestroy();
+        boolean exited = false;
+        try {
+            exited = process.waitFor(10, TimeUnit.SECONDS);
+            processJournal.close();
+            if (!exited) {
+                // we timed out waiting for the process to exit; it may be in a blocking call, so just force
+                // it to shutdown
+                process.destroy();
+                process.waitFor();
```
lbergelson: Won't this wait forever if destroy failed somehow?  Can destroy fail?  It doesn't seem like it can.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+    private final ProcessSettings settings;
+    private final String promptForSynchronization;
+
+    private File fifoTempDir = null;
+    private File fifoFile = null;
+
+    // Timeout used when retrieving output from the remote process to prevent the GATK main tool thread from\
+    // excessive blocking. {@link #isOutputAvailable} can be used to to check for output before making a
+    // blocking call in order to avoid exceeding timeouts.
+    private final int timeOutMillis = 5000;
```
lbergelson: this thing should be a static constant

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+     * @param synchronizationString string to synchronize on
+     * @return ProcessOutput containing a {@code synchronizationString} terminated string in either std or stderr
+     * @throws TimeoutException
+     */
+    private ProcessOutput getOutputSynchronizedBy(final String synchronizationString) throws TimeoutException {
+        boolean gotPrompt = false;
+        try (final ByteArrayOutputStream stdOutAccumulator = new ByteArrayOutputStream(CapturedStreamOutput.STREAM_BLOCK_TRANSFER_SIZE);
+             final ByteArrayOutputStream stdErrAccumulator = new ByteArrayOutputStream(CapturedStreamOutput.STREAM_BLOCK_TRANSFER_SIZE))
+        {
+            while (gotPrompt == false) {
```
lbergelson: why not just `!gotPrompt`?

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+                final StreamOutput stdOut = processOutput.getStdout();
+                if (stdOut != null) {
+                    stdOutAccumulator.write(stdOut.getBufferBytes());
+                }
+                final StreamOutput stderr = processOutput.getStderr();
+                if (stderr != null) {
+                    stdErrAccumulator.write(stderr.getBufferBytes());
+                }
+            }
+            stdOutAccumulator.flush();
```
lbergelson: I think flushing a ByteArrayOutputStream is a noop.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+                }
+            }
+        }
+
+        /**
+         * Close the journal file writer.
+         */
+        public void close() {
+            try {
+                if (journalingFileWriter != null) {
```
lbergelson: I might log a message here that the journal was closed so you can tell if it ended due to an expected closing or if it crashed.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -0,0 +1,598 @@
...
+            }
+        } catch (InterruptedException e) {
+            logger.error(String.format("Interrupt exception waiting for process (%s) to terminate", settings.getCommandString()));
+        }
+        return !process.isAlive();
+    }
+
+    // Stream output class used to aggregate output pulled fom the process stream while waiting for
+    // a synchronization point.
+    private class AggregateStreamOutput extends StreamOutput {
```
lbergelson: I want the naming of this to be different, I was expecting it to perform some sort of aggregation operation, but really it just take s a byte array and returns it... Maybe something like `ByteArrayBackedStreamOutput`?

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
```diff
@@ -0,0 +1,122 @@
...
+
+        if (previousBatch != null) {
+            throw new IllegalStateException("Previous batch not yet complete");
+        }
+
+        previousBatch = executorService.submit(() -> {
+            try {
+                for (int i = 0; i < batchSize; i++) {
+                    T element = batchList.get(i);
+                    streamWriter.write(element.toString().getBytes());
```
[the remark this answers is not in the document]
cmnbroad: Yeah, good catch, I meant to go back and change that. I don't want to restrict this to Strings, or rely on java serialization as a cross-process format, so I added a Function<T, ByteArrayOutputStream> argument so a serializer can be provided, and threaded it through the calls. I also added a static default member variable that can be used for the common String case as a convenience .

src/test/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutorUnitTest.java
```diff
@@ -16,23 +16,23 @@
 public class PythonScriptExecutorUnitTest extends GATKBaseTest {
     final String HELLO_WORLD_SCRIPT = "print (\"hello, world\")";
 
-    @Test(groups = {"python"})
```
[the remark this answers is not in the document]
cmnbroad: Good point. There is still a "python" group (though it probably should be called "conda" because its used for things that depend on the conda env). There is also a travis matrix entry for it, and tests in the "python" group only run in that one travis job. All of the VQSR-CNN and gCNV tests will go in that group soon as well .

There isn't anything about these tests that requires conda, but I could go either way. Maybe that does make the most sense though. If I leave these in the python group, then I would probably also put all of the StreamingPythonExecutor tests in that group as well. Let me know what you think.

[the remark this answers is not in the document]
cmnbroad: Going put all of these in the python group.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
cmnbroad: Yup. Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
cmnbroad: Yes, those both should have been removed.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
cmnbroad: Removed.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
cmnbroad: I think its correct; this should probably use the same timeout as above. Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
cmnbroad: Makes sense. Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
cmnbroad: Yeah. It used to aggregate, but now its just a container. Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
cmnbroad: It just tells the caller if an outstanding batch existed and cancelling it failed. javadoc updated.

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriterService.java
```diff
@@ -0,0 +1,122 @@
...
+                }
+                // this can block, waiting for the stream to be consumed if its buffered
+                streamWriter.flush();
+                return batchSize; // return the number of items this batch was asked to write
+            } catch (IOException e) {
+                throw new RuntimeException(e);
```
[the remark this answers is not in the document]
cmnbroad: yes, replaced with GATKException.

cmnbroad:
@lbergelson Thanks for the review - I think I've addressed everything, and tests are running now. Let me know if you have any more comments/responses - I'd like to get this merged today if possible to unblock VQSR-CNN.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
```diff
@@ -27,8 +27,8 @@
  *
  * <ol>
  * <li>Opens a FIFO for writing.</li>
- * <li>Creates an AsyncWriterervice</li>
- * <li>Writes a string of attributes for each read to the List until the batchSize theshold is reavched.</li>
+ * <li>Creates an AsynchronousWriterServiceto allow writing to the FIFO in batches on a background thread</li>
+ * <li>Writes a string of attributes for each read to the List until the batchSize threshold is reacched.</li>
```
lbergelson: I'm not sure this typo fix was quite correct :)

PR #3876, opened 2017-11-27
Fix getMatchesLessDeletions() so it accounts for indels in NM tag

mwalker174:
Previous behavior assumed that the NM tag was the number of substitutions (as would be implied by "matches or mismatches" in CIGAR lingo). This patches makes the function correctly account for indels as well.


src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
```diff
@@ -91,21 +91,31 @@ public static SAMFileHeader checkAndClearHeaderSequences(final SAMFileHeader inp
         return header;
     }
 
-    public static int getMatchesLessDeletions(final Cigar cigar, final int numMismatches) {
+    public static int getMatchesLessDeletions(final Cigar cigar, final int nmTagValue) {
         Utils.nonNull(cigar, "Cannot get match score for null cigar");
```
vruano: You can simply this code a bit, I don't think you need to track the number of inserted-bases. You could change th ```switch``` for ```if - else - if``` if you like.

```
int alignedBases = 0;
int deletedBases = 0;
for (final CigarElement ce : cigar.getCigarElements()) {`
     switch (ce.getOperator()) {
        case M:
        case EQ:
        case X:  alignedBases += ce.getLength(); break;
        case D:  deletedBases += ce.getLength(); break;
        default:
     }
}
return alignedBases - numMismatches + 2 * deletedBases;
```
... plus checking/logging code.

src/test/resources/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqScoreSpark/expected_paired.txt
```diff
@@ -1,766 +1,766 @@
 tax_id	taxonomy	type	name	kingdom	score	score_normalized	reads	unambiguous	reference_length
-10240	root|Viruses|dsDNA_viruses,_no_RNA_stage|Poxviridae	family	Poxviridae	Viruses	56.0	4.643449419568822	56	56	0
```
vruano: No change needed now but I suggest that in the future you format output float numbers to a fix maximum number of decimals to be shown (e.g 4 most significant numbers) with ```String.format```. On the one hand is not visually appealing but also it might mean that some tribal changes in the code (e.g change in the order of float point operations) could break the test due to a very unimportant decimal value flip.

src/main/java/org/broadinstitute/hellbender/tools/spark/pathseq/PSUtils.java
mwalker174: I do think your suggestion is cleaner, but I want to keep the current code for clarity's sake. It's not a performance bottleneck.

src/test/resources/org/broadinstitute/hellbender/tools/spark/pathseq/PathSeqScoreSpark/expected_paired.txt
mwalker174: Yes I've had this problem and currently in tests check if the value is equal within 6 digits. Truncating would be preferable, though.

PR #3877, opened 2017-11-27
Making -A become -AM in FilterByOrientationBias.

### LINKED ISSUE #3875: FilterByOrientationBias reuses the '-A' argument for '--artifactModes'
It seems that the tool `FilterByOrientationBias` has the argument `--artifactModes` which has the command line abbreviation `-A`. Unfortunately this abbreviation collides with the enshrined `-A` abbreviation for `--annotation` for tools that deal with variants. I would propose we change the argument to `-AM` to avoid naming collisions. If we want to make this change, we should probably do it soon before it becomes too painful to reverse.

LeeTL1220:
So as to avoid a conflict with -A for annotations.

closes #3875

### DIFF
--- a/scripts/mutect2_wdl/mutect2.wdl
+++ b/scripts/mutect2_wdl/mutect2.wdl
@@ -380,7 +380,7 @@ task Filter {
 
     # FilterByOrientationBias must come after all of the other filtering.
     if [[ ! -z "${pre_adapter_metrics}" ]]; then
-        java -Xmx4g -jar $GATK_JAR FilterByOrientationBias -A ${sep=" -A " artifact_modes} \
+        java -Xmx4g -jar $GATK_JAR FilterByOrientationBias -AM ${sep=" -AM " artifact_modes} \
             -V filtered.vcf -P ${pre_adapter_metrics} --output ${filtered_vcf_name}
     else
         mv filtered.vcf ${filtered_vcf_name}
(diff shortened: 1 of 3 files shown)


scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -380,7 +380,7 @@ task Filter {
 
     # FilterByOrientationBias must come after all of the other filtering.
     if [[ ! -z "${pre_adapter_metrics}" ]]; then
-        java -Xmx4g -jar $GATK_JAR FilterByOrientationBias -A ${sep=" -A " artifact_modes} \
+        java -Xmx4g -jar $GATK_JAR FilterByOrientationBias -AM ${sep=" -AM " artifact_modes} \
```
jamesemery: Are there any other WDLs out in the wild (or production) that we have to worry about breaking with this change? Who else is using this tool?

LeeTL1220: I could not find any others in the repo.  

There are some out in the wild.  I warned CGA, since they are using a custom WDL.  Once there is support in FireCloud, they intend to use our WDL as a subworkflow.

LeeTL1220: Most are using our WDL in conjunction with the GATK jar, so this change will be invisible to them.

PR #3892, opened 2017-11-30
putting a link to travis results at the top

lbergelson:
I'm sick of scrolling to the bottom to find the test report uri.
This saves the pain of having to scroll all the way down.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,10 @@ cache:
     - $HOME/gcloud/
     - $HOME/site-library
 before_install:
+- REPORT_PATH=${TRAVIS_BRANCH}_${TRAVIS_JOB_NUMBER};
+- if [ $TRAVIS_SECURE_ENV_VARS == true && $TRAVIS_EVENT_TYPE != cron ]; then
+    echo "Test report will be written to  https://storage.googleapis.com${HELLBENDER_TEST_LOGS}${REPORT_PATH}/tests/test/index.html";
+  fi
 #setup google cloud and github authentication
 - if [ $TRAVIS_SECURE_ENV_VARS == true ]; then
     export BOTO_CONFIG=/dev/null;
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -52,6 +52,8 @@ cache:
     - $HOME/gcloud/
     - $HOME/site-library
 before_install:
+- REPORT_PATH=${TRAVIS_BRANCH}_$T{RAVIS_JOB_NUMBER};
```
jamesemery: $T{RAVIS_JOB_NUMBER} -> ${TRAVIS_JOB_NUMBER}

jamesemery: Also perhaps we should hide this text behind $TRAVIS_SECURE_ENV_VARS == true  so its not pointing to some non-existant bucket if someone external submits a PR.

lbergelson: That's a a good idea.

lbergelson: oh, and a good typo catch, this is why I should never make "minor changes" after checking that something is working....

src/main/java/org/broadinstitute/hellbender/utils/downsampling/AlleleBiasedDownsamplingUtils.java
```diff
@@ -6,11 +6,8 @@
 import org.apache.commons.collections4.map.DefaultedMap;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.hellbender.exceptions.UserException;
-import org.broadinstitute.hellbender.utils.BaseUtils;
```
droazen: You  might want to clean up the diff here though...

.travis.yml
```diff
@@ -52,6 +52,10 @@ cache:
     - $HOME/gcloud/
     - $HOME/site-library
 before_install:
+- REPORT_PATH=${TRAVIS_BRANCH}_${TRAVIS_JOB_NUMBER};
+- if [ $TRAVIS_SECURE_ENV_VARS == true && $TRAVIS_EVENT_TYPE != cron ]; then
```
jamesemery: It looks like you may have needed a second set of brackets. If you check the travis logs this line is currently erroring with the following message and failing to print the path correctly: 
`0.01s$ if [ $TRAVIS_SECURE_ENV_VARS == true && $TRAVIS_EVENT_TYPE != cron ]; then echo "Test report will be written to  https://storage.googleapis.com${HELLBENDER_TEST_LOGS}${REPORT_PATH}/tests/test/index.html"; fi
/home/travis/.travis/job_stages: line 57: [: missing `]'`

jamesemery:
The most recent run seems to have fixed it, 👍

PR #3893, opened 2017-11-30
merge prototyping SV breakpoint and type inference tool into SV spark pipeline

### LINKED ISSUE #3890: merge prototyping SV breakpoint and type inference tool into SV spark pipeline
First with a flag in the tool that allows user to choose which code path to use, the existing version or the prototyping version.

But the test coverage is still low and should be improved.

SHuang-Broad:
resolves #3890
Broken into commits for easier review

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
@@ -4,21 +4,23 @@ import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMFlag;
 import htsjdk.samtools.SAMReadGroupRecord;
+import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.variant.variantcontext.VariantContext;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.apache.spark.broadcast.Broadcast;
-import org.broadinstitute.barclay.argparser.Argument;
-import org.broadinstitute.barclay.argparser.ArgumentCollection;
-import org.broadinstitute.barclay.argparser.BetaFeature;
-import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
+import org.broadinstitute.barclay.argparser.*;
 import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 import org.broadinstitute.hellbender.cmdline.programgroups.StructuralVariationSparkProgramGroup;
+import org.broadinstitute.hellbender.engine.datasources.ReferenceMultiSource;
 import org.broadinstitute.hellbender.engine.spark.GATKSparkTool;
 import org.broadinstitute.hellbender.tools.spark.sv.discovery.*;
+import org.broadinstitute.hellbender.tools.spark.sv.discovery.prototype.AssemblyContigAlignmentSignatureClassifier;
+import org.broadinstitute.hellbender.tools.spark.sv.discovery.prototype.AssemblyContigWithFineTunedAlignments;
+import org.broadinstitute.hellbender.tools.spark.sv.discovery.prototype.SvDiscoverFromLocalAssemblyContigAlignmentsSpark;
 import org.broadinstitute.hellbender.tools.spark.sv.evidence.AlignedAssemblyOrExcuse;
 import org.broadinstitute.hellbender.tools.spark.sv.evidence.EvidenceTargetLink;
 import org.broadinstitute.hellbender.tools.spark.sv.evidence.FindBreakpointEvidenceSpark;
@@ -30,8 +32,11 @@ import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.bwa.BwaMemAlignment;
 import org.broadinstitute.hellbender.utils.bwa.BwaMemAlignmentUtils;
 import org.broadinstitute.hellbender.utils.fermi.FermiLiteAssembly;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import org.broadinstitute.hellbender.utils.read.SAMRecordToGATKReadAdapter;
 import scala.Serializable;
 
+import java.util.EnumMap;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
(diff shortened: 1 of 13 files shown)


src/test/java/org/broadinstitute/hellbender/tools/spark/sv/integration/StructuralVariationDiscoveryPipelineSparkIntegrationTest.java
```diff
@@ -151,16 +155,50 @@ public void testSVDiscoverPipelineRunnableMiniCluster(final StructuralVariationD
...
+                        .resolve(StructuralVariationDiscoveryPipelineSpark.EXPERIMENTAL_INTERPRETATION_OUTPUT_DIR_NAME.concat("/InsDel.vcf"))
+                        .toUri().toString();
+            } else {
+                experimentalInsDelVcf = IOUtils.getPath(generatedVCFPath).getParent().toAbsolutePath().toString() +
+                        "/" + StructuralVariationDiscoveryPipelineSpark.EXPERIMENTAL_INTERPRETATION_OUTPUT_DIR_NAME + "/InsDel.vcf";
+            }
+
+            actualVcs = extractActualVCs(experimentalInsDelVcf, onHDFS);
+
+            // TODO: 11/30/17 temporary solution to ignore these attributes before they can be brought back
```
SHuang-Broad: See #3894

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -169,9 +173,11 @@ private void experimentalInterpretation(final JavaSparkContext ctx,
                                             final Broadcast<SAMSequenceDictionary> broadcastSequenceDictionary,
                                             final String sampleId,
                                             final FindBreakpointEvidenceSpark.AssembledEvidenceResults assembledEvidenceResults) {
-        final String expOutputDir =
-                IOUtils.getPath(vcfOutputFileName).getParent().toAbsolutePath().toString() +
-                        "/experimentalVariantInterpretations";
+        String expOutputDir =
```
cwhelan: It seems like it might be a lot cleaner to just replace the output file parameter with an output directory, and then write the requested files into that directory. Is that too much of a hassle? Alternatively you could just make the method of requesting the new variants be to provide a new file parameter, instead of a flag (for now, we can change this when we fully switch over).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -169,9 +173,11 @@ private void experimentalInterpretation(final JavaSparkContext ctx,
...
+                IOUtils.getPath(vcfOutputFileName).getParent().toAbsolutePath().resolve(EXPERIMENTAL_INTERPRETATION_OUTPUT_DIR_NAME)
+                .toUri().toString();
+        if ( !BucketUtils.isRemoteStorageUrl(expOutputDir) )
+            expOutputDir = expOutputDir.replaceFirst("file://", "");
```
cwhelan: At the very least I'd like to get rid of having to do this, it seems terribly hacky. You could at least query the URI above for the path minus the scheme, right?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -68,6 +65,11 @@
     private final DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection discoverStageArgs
             = new DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection();
 
+    @Hidden
+    @Argument(doc = "to use prototyping breakpoint and type inference tool or not, in addition to master version",
```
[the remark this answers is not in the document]

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
```diff
@@ -132,24 +132,25 @@ protected void runTool(final JavaSparkContext ctx) {
...
         new SimpleStrandSwitchVariantDetector()
                 .inferSvAndWriteVCF(contigsByPossibleRawTypes.get(RawTypes.IntraChrStrandSwitch).map(decoratedTig -> decoratedTig.contig),
-                        outputDir+"/"+ RawTypes.IntraChrStrandSwitch.name()+".vcf",
+                        outputDir +"/"+ RawTypes.IntraChrStrandSwitch.name()+".vcf",
                         referenceMultiSourceBroadcast, broadcastSequenceDictionary, localLogger, sampleId);
 
         new SuspectedTransLocDetector()
                 .inferSvAndWriteVCF(contigsByPossibleRawTypes.get(RawTypes.MappedInsertionBkpt).map(decoratedTig -> decoratedTig.contig),
-                        outputDir+"/"+ RawTypes.MappedInsertionBkpt.name()+".vcf",
+                        outputDir +"/"+ RawTypes.MappedInsertionBkpt.name()+".vcf",
```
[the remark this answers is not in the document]
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
SHuang-Broad: totally. Rewrote how the path is resolved in the new commit.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/prototype/SvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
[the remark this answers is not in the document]
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
SHuang-Broad: I actually thought about it, but eventually went the way it is currently implemented basically based on preference of "adding" over "replacing":
since it is still experimental, it might be better to clearly indicate that "in addition to what we have clearly evaluated/tested, we are intending to output more in the future, so here are both if you specifically know what you are requesting, with them separated into a single VCF and a directory, and please compare them if necessary"

For the alternative solution you are suggesting, is my understanding correct:
instead of a flag in the CLI, have a (possibly null) string, and the method `experimentalInterpretation(...)` could test if the string is null and call into the experimental code, so that user has the option of specifying the experimental output dir?

cwhelan: Yes, that's what I was suggesting as an alternative. It seemed to match with how we handle some of our optional outputs, eg. you request writing of the assembly sam file or the target links file by passing in an output path for it. Does that make sense to you or does it seem weird?  This isn't really a big deal, feel free to choose whatever feels right to you. Hopefully we can get the old version of the tool replaced with the new version soon anyway.

SHuang-Broad: Yes, it totally makes sense. I'll make that change and merge afterwards.

PR #3897, opened 2017-11-30
Funcotator - created parser for non-locatable "separated value" files

### LINKED ISSUE #3757: Need simple parser for non-locatable files
We need a parser that can read in non-Locatable files.

These files will be *sv formatted (e.g. tsv / csv / etc.).

This should be based off of the simple `Path` reader in #3756

jonn-smith:
Adding in a parser that handles text files that are delimited with some known separator (i.e. commas, tabs, words).  The new class is called `SimpleKeyXsvFuncotationFactory` and it is usable with the rest of `Funcotator` using new command-line arguments.

Added new integration tests that include this new data source factory.

Fixes #3757


jonn-smith:
PR #3952 obviates this PR.

Issue #3144, opened 2017-06-22
Add/Change arguments in LibraryReadFilter


magicDGS:
`LibraryReadFilter`should have support for:

* Exclude libraries
* Select several libraries and not only one

PR #3901, opened 2017-12-01
Support several libraries in LibraryReadFilter

magicDGS:
Part of https://github.com/broadinstitute/gatk/issues/3144, maintaining the usage as before but allowing multiple `--library` arguments to provide more than one library...

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/filters/LibraryReadFilter.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/filters/LibraryReadFilter.java
@@ -34,6 +34,6 @@ public final class LibraryReadFilter extends ReadFilter implements Serializable 
             return false;
         }
 
-        return libraryToKeep.stream().anyMatch(library::equals);
+        return libraryToKeep.contains(library);
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/engine/filters/LibraryReadFilter.java
```diff
@@ -28,6 +30,10 @@ public LibraryReadFilter( final SAMFileHeader header ) {
     @Override
     public boolean test( final GATKRead read ) {
         final String library = ReadUtils.getLibrary(read, samHeader);
-        return library != null && library.equals(libraryToKeep);
+        if (library == null) {
+            return false;
+        }
+
+        return libraryToKeep.stream().anyMatch(library::equals);
```
lbergelson: @magicDGS Shouldn't this use `libraryToKeep.contains(library)`  instead a streaming match?  That seems like it would be more efficient.

magicDGS: True! My bad! Changed in next commit.

lbergelson:
@magicDGS Thank you!

PR #3912, opened 2017-12-04
Create a conda-based Python environment.

cmnbroad:
- Adds the “gatk” conda environment, with dependencies defined by the file scripts/gatkcondaenv.yml.
- Updates the docker image to include the activated conda environment.
- Adds a new entry to the travis test matrix for running tests that depend on Python and the conda environment.
- Adds a “python” test group. Any tests for tools or functionality that are dependent on Python should be put into this group. Tests in this group will be executed in a docker container on travis in the python build matrix entry only.
- The existing WDL tests are unchanged; so although they execute in the context of the docker container and conda environment, there are no tests (yet) that are actually dependent on the conda environment and run through cromwell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,16 @@ releases of the toolkit.
     * Gradle 3.1 or greater, needed for building the GATK. We recommend using the `./gradlew` script which will
       download and use an appropriate gradle version automatically (see examples below).
     * Python 2.6 or greater (needed for running the `gatk-launch` frontend script)
+    * Python 3.6.2, along with a set of additional Python packages, are required to run some tools and workflows.
+      GATK uses the [Conda](https://conda.io/docs/index.html) package manager to
+      establish and manage the environment and dependencies required by these tools. The GATK Docker image comes
+      with this environment pre-configured. In order to establish an environment suitable to run these tools
+      outside of the Docker image, the conda [gatkcondaenv.yml](https://github.com/broadinstitute/gatk/blob/master/scripts/gatkcondaenv.yml)
+      file is provided. To establish the conda environment locally, [Conda](https://conda.io/docs/index.html) must first
+      be installed. Then, create the gatk environment by running the command ```conda env -n gatk -f gatkcondaenv.yml```.
+      To activate the environment once its been created, run the command ```source activate gatk```. See the
+      [Conda](https://conda.io/docs/user-guide/tasks/manage-environments.html) documentation for
+      additional information about using and managing Conda environments.
     * R 3.1.3 (needed for producing plots in certain tools, and for running the test suite)
     * [git-lfs](https://git-lfs.github.com/) 1.1.0 or greater (needed to download large files for the complete test suite).
       Run `git lfs install` after downloading, followed by `git lfs pull` from the root of your git clone to download the large files. The download is several hundred megabytes.
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
   - TEST_TYPE=cloud UPLOAD=true
   - TEST_TYPE=integration TEST_DOCKER=true TEST_VERBOSITY=minimal
   - TEST_TYPE=unit TEST_DOCKER=true TEST_VERBOSITY=minimal
+  - TEST_TYPE=python TEST_DOCKER=true TEST_VERBOSITY=minimal
   - RUN_CNV_SOMATIC_LEGACY_WDL=true
   - RUN_M2_WDL=true
   global:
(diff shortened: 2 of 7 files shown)


cmnbroad:
@samuelklee @lucidtronix @mbabadi I still need to update the readme, but please review/try this out. I wound up just creating the conda environment right in the docker, so we use the same environment we give to non-docker users. The environment includes the common dependencies from https://github.com/broadinstitute/gatk/issues/3692, but not the actual GATK (CNN-vqs or gCNV packages).

samuelklee:
Excellent, thanks @cmnbroad!  I have one branch, sl_gcnv_ploidy_cli, which has WDL Travis tests that require the environment.  I've verified that these tests pass locally within my gcnv conda environment.

This branch is a few rebases deep (sl_wgs_acnv:f49669d -> mb_gcnv_python_kernel:54ba864) at the moment.  I'll spend some time tomorrow rebasing everything on top of your branch and will see if tests pass on Travis---fingers crossed!

mbabadi:
@cmnbroad this is an excellent PR!

mbabadi:
By the way, `theano` needs `g++` for compiling the computational graph. In your opinion, what/where is a good strategy/place to assert this @cmnbroad?

cmnbroad:
@mbabadi Not sure - we'll need to discuss how to do handle this. It looks like the docker image already has g++ installed, which is a start. Having said that, the docker image is getting pretty huge rapidly... BTW, do you know if there is way to programmatically query theano to determine whether it will compile the graph, vs running python/numpy  ? I know it prints out a message, but it would be nice if we could detect that from theano directly.

README.md
```diff
@@ -58,7 +58,7 @@ releases of the toolkit.
 * Optional, but recommended:
     * Gradle 3.1 or greater, needed for building the GATK. We recommend using the `./gradlew` script which will
       download and use an appropriate gradle version automatically (see examples below).
-    * Python 2.6 or greater (needed for running the `gatk-launch` frontend script)
+    * [THIS SECTION NEEDS TO BE UPDATED WITH CONDA INFO] Python 2.6 or greater (needed for running the `gatk-launch` frontend script)
```
samuelklee: Is this happening in this PR or a later one?  Do we need to file an issue?

cmnbroad: I want to get at least something in here as part of this PR.

mbabadi:
@cmnbroad The path to the c++ compiler can be specifically provided to theano by setting `theano.config.cxx` in python scripts, or by creating a `.theanorc` in the home directory, or by setting the environmental variable `THEANO_FLAGS=cxx=<path_to_g++>,...`.

If a working c++ compiler exists and provided to theano, it is fair to assume that the graph _will_ compile. If the c++ compiler is not explicitly specified, theano will try to discover it. It first tries to execute `g++ -v` in the present environment and if it succeeds, it resolves the absolute path to the executable. On darwin, it further searches for `clang++` and on Win32, it looks for a working mingw gcc setup.

We could _enforce_ the presence of a c++ compiler at the beginning of all python scripts and throw an exception and an informative message instead of numpy/python fallback. If we do so, the integration tests (and all gCNV CLI tools) will fail and will force the user to install a c++ compiler. In your opinion, is this fail-fast strategy a better approach, given that python fallback runs 2~3 orders of magnitude slower?

mbabadi:
@cmnbroad Chris, could you update `Miniconda2` to `Mininconda3` in the docker config? here's the sh bundle + MD5:
```
ENV CONDA_URL https://repo.continuum.io/miniconda/Miniconda3-4.3.30-Linux-x86_64.sh
ENV CONDA_MD5 = "0b80a152332a4ce5250f3c09589c7a81"```

mbabadi:
@cmnbroad on a related note -- it might be worthwhile to setup the Docker to include a dynamic BLAS library and pass it to theano. I will test how it affects the performance. NumPy is usually either linked against MKL or OpenBLAS. If theano has no dynamic BLAS lib available to link the compiled graph against, it will fall back to NumPy for linalg ops. It is not too bad since the only cost is the c++/python communication overhead.

mbabadi:
@cmnbroad ignore the previous message -- I managed to add MKL to the docker.

cmnbroad:
@mbabdi I've updated my PR to use miniconda3.

@mbabdi @lucidtronix @samuelklee I think we should aim for tools that at least run out-of-box, without depending on any out-of-band configuration other than the conda env. On top of that we can provide guidance/configs for users on how to enable further optimizations, like g++. Does that sound like an achievable goal ?

As for the docker, we're going to have strike the right balance between image bloat and performance(including test performance). I think we're around 4+ gig now, and counting. Before the Python integration we were at 1.9G, and trying to find ways to reduce it. So lets see where we wind up but keep that in mind.

Finally, we need to find a way to install the (GATK) python package(s) without depending on access to the GATK repo. Right now I think the gCNV branch has a "pip install from source" added to the conda env .yml. That will work on the docker at the moment (and thus on travis), but that won't work for non-docker users how don't have source/repo access. Also, one of the proposals to reduce the size of the docker is to remove the repo clone that is currently there.

My proposal is that we change the gradle build to create an archive/zip of the python source (this would include the VQSR-CNN package code as well as gCNV kernal). We can then copy that on to the docker image, and pip-install it from the copy. That would retain the ability to always run travis tests based on the code in the repo, and also keep the nightly docker image in sync. We'll also have deliver the archive as an artifact for non-docker users.

cmnbroad:
@lucidtronix Do you have any comments on the conda env defined in this branch ? The .yml file here is based in part on dependencies I took from your conda env (minus the actual VQSR-CNN package, since we want our travis tests to run on the code and env defined in the GATK source repo - see comment above), but I haven't actually tried  running your code with it yet. I want to make sure we're converging on a unified conda env.

samuelklee:
@lucidtronix Mind taking a look at this as soon as you get a chance?  It'd be great to get it merged so we can rebase and review some of the gCNV branches.

cmnbroad:
I did a smoke test and ran a slightly out of date version of the VQSR_CNN NeuralNetStreamingExecutor in this environment, so I'm merging this.

Issue #3709, opened 2017-10-17
Add GCS support for fasta and interval files via NIO


droazen:

jean-philippe-martin:
This issue could use a description. Support for reading? Writing? In Spark or outside of Spark?

droazen:
Let's make it easy: read support only, and in walkers only.

droazen:
Basically the walker arguments -I and -V already support GCS, but the arguments -R and -L don't, so this ticket is to enable GCS support for the -R and -L arguments in walkers.

PR #3921, opened 2017-12-05
Use Path for references instead of File

jean-philippe-martin:
Update multiple tools to use a Path instead of File.

Updated: FeatureWalker, ReadWalker, IntervalWalker, MultiVariantWalker, AnnotateIntervals, BaseRecalibrator, SelectVariants, SplitNCigarReads.

Some tools will need a change to ReferenceSequenceFileWalker in htsjdk, so I couldn't change those. Same for the writers.

Part of addressing issue #3709

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReferenceInputArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReferenceInputArgumentCollection.java
@@ -13,11 +13,6 @@ public abstract class ReferenceInputArgumentCollection implements Serializable {
     private static final long serialVersionUID = 1L;
 
     /**
-     * Get the reference file specified at the command line, creating the File object first if necessary.
-     */
-    public abstract File getReferenceFile();
-
-    /**
      * Get the name of the reference file specified at the command line.
      */
     public abstract String getReferenceFileName();
--- a/src/main/java/org/broadinstitute/hellbender/engine/ReferenceFileSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/ReferenceFileSource.java
@@ -2,6 +2,7 @@ package org.broadinstitute.hellbender.engine;
 
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.reference.ReferenceSequence;
+import java.nio.file.Path;
 import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.fasta.CachingIndexedFastaSequenceFile;
@@ -30,11 +31,11 @@ public final class ReferenceFileSource implements ReferenceDataSource {
      *
      * The provided fasta file must have companion .fai and .dict files.
      *
-     * @param fastaFile reference fasta file
+     * @param fastaPath reference fasta file
      */
-    public ReferenceFileSource(final File fastaFile) {
+    public ReferenceFileSource(final Path fastaPath) {
         // Will throw a UserException if the .fai and/or .dict are missing
-        reference = CachingIndexedFastaSequenceFile.checkAndCreate(Utils.nonNull(fastaFile));
+        reference = CachingIndexedFastaSequenceFile.checkAndCreate(Utils.nonNull(fastaPath));
     }
 
     /**
--- a/src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceFileSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceFileSource.java
@@ -4,8 +4,13 @@ import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.reference.ReferenceSequenceFile;
 import htsjdk.samtools.reference.ReferenceSequenceFileFactory;
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.io.IOUtils;
 import org.broadinstitute.hellbender.utils.reference.ReferenceBases;
 
 import java.io.File;
@@ -15,33 +20,52 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 
 /**
- * Class to load a reference sequence from a local fasta file.
+ * Class to load a reference sequence from a fasta file.
  */
 public class ReferenceFileSource implements ReferenceSource, Serializable {
     private static final long serialVersionUID = 1L;
 
-    private final String referencePath;
+    private final URI referenceUri;
+    // ReferenceFileSource is serializable (for running across a cluster), so we re-make the path
+    // if necessary. Of course if you give us a Path to a RAM filesystem, don't expect
+    // it to survive serialization.
+    private transient Path referencePath;
 
     /**
-     * @param referencePath the local path to the reference file
+     * @param referenceUri the path to the reference file
      */
-    public ReferenceFileSource(final String referencePath) {
-        if (!new File(referencePath).exists()) {
+    public ReferenceFileSource(final String referenceUri) {
+        this(IOUtils.getPath(referenceUri));
+    }
+
+    /**
+     * @param referencePath the path to the reference file
+     */
+    public ReferenceFileSource(final Path referencePath) {
+        this.referencePath = referencePath;
+        this.referenceUri = referencePath.toUri();
+        if (!Files.exists(this.referencePath)) {
             throw new UserException.MissingReference("The specified fasta file (" + referencePath + ") does not exist.");
         }
-        this.referencePath = referencePath;
+    }
+
+    private synchronized Path getReferencePath() {
+        if (null == referencePath) {
+            this.referencePath = Paths.get(referenceUri);
+        }
+        return referencePath;
     }
 
     @Override
     public ReferenceBases getReferenceBases(final SimpleInterval interval) throws IOException {
-        try ( ReferenceSequenceFile referenceSequenceFile = ReferenceSequenceFileFactory.getReferenceSequenceFile(new File(referencePath)) ) {
+        try ( ReferenceSequenceFile referenceSequenceFile = ReferenceSequenceFileFactory.getReferenceSequenceFile(getReferencePath()) ) {
             ReferenceSequence sequence = referenceSequenceFile.getSubsequenceAt(interval.getContig(), interval.getStart(), interval.getEnd());
             return new ReferenceBases(sequence.getBases(), interval);
         }
     }
 
     public Map<String, ReferenceBases> getAllReferenceBases() throws IOException {
-        try ( ReferenceSequenceFile referenceSequenceFile = ReferenceSequenceFileFactory.getReferenceSequenceFile(new File(referencePath)) ) {
+        try ( ReferenceSequenceFile referenceSequenceFile = ReferenceSequenceFileFactory.getReferenceSequenceFile(getReferencePath()) ) {
             Map<String, ReferenceBases> bases = new LinkedHashMap<>();
             ReferenceSequence seq;
             while ( (seq = referenceSequenceFile.nextSequence()) != null ) {
@@ -54,7 +78,7 @@ public class ReferenceFileSource implements ReferenceSource, Serializable {
 
     @Override
     public SAMSequenceDictionary getReferenceSequenceDictionary(final SAMSequenceDictionary optReadSequenceDictionaryToMatch) throws IOException {
-        try ( ReferenceSequenceFile referenceSequenceFile = ReferenceSequenceFileFactory.getReferenceSequenceFile(new File(referencePath)) ) {
+        try ( ReferenceSequenceFile referenceSequenceFile = ReferenceSequenceFileFactory.getReferenceSequenceFile(getReferencePath()) ) {
             return referenceSequenceFile.getSequenceDictionary();
         }
     }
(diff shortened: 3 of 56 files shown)


src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReferenceInputArgumentCollection.java
```diff
@@ -14,6 +14,8 @@
 
     /**
      * Get the reference file specified at the command line, creating the File object first if necessary.
+     *
+     * Use getReferencePath() instead.
```
magicDGS: Why not just remove it and use wherever it appears the `getReferencePath().toFile()`?

src/main/java/org/broadinstitute/hellbender/engine/ReferenceFileSource.java
```diff
@@ -35,8 +48,7 @@
      * @param fastaFile reference fasta file
      */
     public ReferenceFileSource(final File fastaFile) {
-        // Will throw a UserException if the .fai and/or .dict are missing
-        reference = CachingIndexedFastaSequenceFile.checkAndCreate(Utils.nonNull(fastaFile));
+        this(fastaFile.toPath());
```
magicDGS: This will throw an NPE if the `fastaFile`is null. Better: `this(Utils.nonNull(fastaFile).toPath())`

src/main/java/org/broadinstitute/hellbender/tools/exome/pulldown/BayesianHetPulldownCalculator.java
```diff
@@ -9,6 +9,7 @@
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker;
 import htsjdk.samtools.util.IntervalList;
 import htsjdk.samtools.util.SamLocusIterator;
+import java.nio.file.Path;
```
magicDGS: Unused import.

jean-philippe-martin: Removed.

src/main/java/org/broadinstitute/hellbender/engine/ReferenceFileSource.java
jean-philippe-martin: Good call, done.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReferenceInputArgumentCollection.java
jean-philippe-martin: That's not a bad idea, but leaving it in makes it super-easy to find the places I still have to update (using the IDE's "find calls"). The latter syntax would require a full-text search, which is slower and less convenient. Also, it's public so removing it might break some code that depends on GATK. What I was thinking instead was to mark it as @deprecated once we're done cleaning house.

src/main/java/org/broadinstitute/hellbender/engine/ReferenceFileSource.java
```diff
@@ -35,8 +48,7 @@
      * @param fastaFile reference fasta file
      */
     public ReferenceFileSource(final File fastaFile) {
```
lbergelson: I think we can just get rid of this method completely and switch to path in the tests that use it.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReferenceInputArgumentCollection.java
lbergelson: That's a good point.  We should definitely fix those usages as well, but we could do it in another PR if you don't want to fit it into this one.  It looks like it needs some changes to the spark tools / one of the CNV tools.

src/main/java/org/broadinstitute/hellbender/utils/fasta/CachingIndexedFastaSequenceFile.java
```diff
@@ -134,53 +138,53 @@ public CachingIndexedFastaSequenceFile(final File fasta) throws FileNotFoundExce
...
         }
 
-        final boolean isGzipped = fastaFile.getAbsolutePath().endsWith(".gz");
+        final boolean isGzipped = fastaPath.toString().endsWith(".gz");
         if ( isGzipped ) {
             throw new UserException.CannotHandleGzippedRef();
         }
 
-        final File indexFile = new File(fastaFile.getAbsolutePath() + ".fai");
+        final Path indexPath = Paths.get(URI.create(fastaPath.toUri().toString() + ".fai"));
```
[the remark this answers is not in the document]
jean-philippe-martin: good call!

src/main/java/org/broadinstitute/hellbender/engine/ReferenceFileSource.java
jean-philippe-martin: Agreed, and done.

src/main/java/org/broadinstitute/hellbender/utils/fasta/CachingIndexedFastaSequenceFile.java
```diff
@@ -134,53 +138,53 @@ public CachingIndexedFastaSequenceFile(final File fasta) throws FileNotFoundExce
...
 
         // determine the name for the dict file
         // TODO: use the htsjdk method implemented in https://github.com/samtools/htsjdk/pull/774
```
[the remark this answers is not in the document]
jean-philippe-martin: Thanks! Done.

src/main/java/org/broadinstitute/hellbender/utils/fasta/CachingIndexedFastaSequenceFile.java
```diff
@@ -134,53 +138,53 @@ public CachingIndexedFastaSequenceFile(final File fasta) throws FileNotFoundExce
...
-        final String fastaExt = fastaFile.getAbsolutePath().endsWith("fa") ? "\\.fa$" : "\\.fasta$";
-        final File dictFile = new File(fastaFile.getAbsolutePath().replaceAll(fastaExt, IOUtil.DICT_FILE_EXTENSION));
+        final String fastaExt = fastaPath.toString().endsWith("fa") ? "\\.fa$" : "\\.fasta$";
+        final Path dictPath = Paths.get(URI.create(fastaPath.toUri().toString().replaceAll(fastaExt, IOUtil.DICT_FILE_EXTENSION)));
 
         // It's an error if either the fai or dict file does not exist. The user is now responsible
         // for creating these files.
-        if (!indexFile.exists()) {
-            throw new UserException.MissingReferenceFaiFile(indexFile, fastaFile);
+        if (!Files.exists(indexPath)) {
```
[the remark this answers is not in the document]
jean-philippe-martin: Agreed.

jean-philippe-martin:
@lbergelson I added a jimfs test to ReferenceFileSourceUnitTest. While I was there I also discovered there's not just `enging.ReferenceFileSource` but also `engine.datasources.ReferenceFileSource`. I updated the latter a little bit so it would work with `Path`s.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReferenceInputArgumentCollection.java
magicDGS: Dependent code on GATK should be aware that using a beta version might change the API - I think that I am one of the affected by changes in API and I am find with them, because I accepted it from the beginning...

magicDGS: And I agree that it can wait, but before the release to do not have support for deprecated things...

src/test/java/org/broadinstitute/hellbender/engine/ReferenceFileSourceUnitTest.java
```diff
@@ -12,7 +16,61 @@
 
     @Test(expectedExceptions = UserException.MissingReference.class)
     public void testMissingReferenceFile() throws IOException {
-        new ReferenceFileSource(GATKBaseTest.getSafeNonExistentFile("NonExistentReference.fasta").getAbsolutePath());
+        new org.broadinstitute.hellbender.engine.datasources.ReferenceFileSource(
```
lbergelson: Wow, I didn't realize we had no direct tests for this before...

src/test/java/org/broadinstitute/hellbender/engine/ReferenceFileSourceUnitTest.java
```diff
@@ -12,7 +16,61 @@
...
+    public void testReferenceInPath() throws IOException {
+        try (FileSystem jimfs = Jimfs.newFileSystem(Configuration.unix())) {
+            final Path refPath = jimfs.getPath("reference.fasta");
+            Files.createFile(refPath);
+            final Path faiFile = jimfs.getPath("reference.fasta.fai");
+            Files.createFile(faiFile);
+            final Path dictPath = jimfs.getPath("reference.dict");
+            Files.createFile(dictPath);
+
+            new org.broadinstitute.hellbender.engine.ReferenceFileSource(refPath);
```
lbergelson: Could you copy a real fasta into jimfs and then assert that we can actually get correct data out of it instead of just that it doesn't crash?

src/test/java/org/broadinstitute/hellbender/engine/ReferenceFileSourceUnitTest.java
```diff
@@ -12,7 +16,61 @@
...
+    public void testDatasourcesReferenceInPath() throws IOException {
+        try (FileSystem jimfs = Jimfs.newFileSystem(Configuration.unix())) {
+            final Path refPath = jimfs.getPath("reference.fasta");
+            Files.createFile(refPath);
+            final Path faiFile = jimfs.getPath("reference.fasta.fai");
+            Files.createFile(faiFile);
+            final Path dictPath = jimfs.getPath("reference.dict");
+            Files.createFile(dictPath);
+
+            new org.broadinstitute.hellbender.engine.datasources.ReferenceFileSource(refPath);
```
lbergelson: This is so confusing... We should fix this naming collision..  I opened a ticket for it #3926.

I think this test should probably be moved to a new test class ReferenceFileSourceUnitTest in the same package as that.  If it could assert that we can really read data as well that would be great.

src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceFileSource.java
```diff
@@ -16,33 +19,40 @@
...
+    public ReferenceFileSource(final Path referencePath) {
+        this.referencePath = referencePath;
+        if (!Files.exists(this.referencePath)) {
             throw new UserException.MissingReference("The specified fasta file (" + referencePath + ") does not exist.");
         }
-        this.referencePath = referencePath;
     }
 
-    @Override
+        @Override
```
lbergelson: an extra tab got added here somehow

lbergelson:
@jean-philippe-martin Thanks, and thanks for catching that other class, it's bizarre that we have 2... I have a few more minor comments, and then 👍

src/test/java/org/broadinstitute/hellbender/engine/ReferenceFileSourceUnitTest.java
jean-philippe-martin: Is there a reference test that reads from disk that I can use as a model?

src/test/java/org/broadinstitute/hellbender/engine/ReferenceFileSourceUnitTest.java
jean-philippe-martin: Agreed; I split the class in two accordingly. Remains to actually read the data.

src/main/java/org/broadinstitute/hellbender/engine/datasources/ReferenceFileSource.java
jean-philippe-martin: Got it, fixed.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReferenceInputArgumentCollection.java
jean-philippe-martin: Good points. I went ahead and deleted getReferenceFile. The Spark usages look a bit wonky now, someone may want to revisit in a later PR and perhaps simplify the logic there.

src/main/java/org/broadinstitute/hellbender/tools/exome/pulldown/BayesianHetPulldownCalculator.java
jean-philippe-martin: It's back now, but also it's used. I was able to switch `BayesianHetPulldownCalculator` to `Path`.

lbergelson:
@jean-philippe-martin It looks like `ReferenceDataSourceUnitTest` is what's covering the `engine.ReferenceFileSource` functionality.  It should be easy to adapt one of those tests to run in jmfs.

jean-philippe-martin:
Ah yes indeed, I see. Would you like me to move these tests to ReferenceFileSourceUnitTest? That seems like a better home for them.

jean-philippe-martin:
Aha, one interesting wrinkle that came when trying to remove `getReferenceFile()` entirely in favor of `getReferencePath()` : It is valid to have references of the form `gg://foobar` (referencing the Google Genomics Reference API). You can apparently put that string in a `File` just fine, but a `Path` will try to find the matching provider, and will fail since there is no NIO provider for "gg" (it doesn't behave like a filesystem).

The work around is to use getReferenceFileName(). The risk is that some places that allowed a relative path (relying on the call to getAbsolutePath) may now require an absolute path.

jean-philippe-martin:
OK, Travis passes. I think I've addressed all the comments, so the ball is in your camp!

lbergelson:
@jean-philippe-martin I think this seems good now.  There were some merge conflicts so I resolved those.  Hopefully tests will pass on the new version and we can merge.

lbergelson:
Aaah, there are more conflicts now

droazen:
I'll rebase this branch, since there are conflicts once again.

PR #3931, opened 2017-12-06
Removed SUDO from the docker build script

### LINKED ISSUE #3798: Remove the necessity for SUDO on the build_docker.sh script
The script is dangerous in its current state because it uses rm -Rf ${dir} arguments which can result in unwanted deletion if something goes wrong with the input arguments. We should either fix those arguments or remove the necessity for the script to be run with root permissions altogether to avoid any future problems that might arise and make the script safer for users. Additionally, the whole script could be refactored to be cleaner.

jamesemery:
resolves #3798

### DIFF
--- a/build_docker.sh
+++ b/build_docker.sh
@@ -42,11 +42,6 @@ Optional arguments:  \n \
 	exit 1
 fi
 
-# Make sure sudo or root was used.
-if [ "$(whoami)" != "root" ]; then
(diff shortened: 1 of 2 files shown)


jamesemery:
I have built several docker images and showed that it will build without SUDO. The only worry is pushing which I have not tested.

jamesemery:
@cmnbroad Can I go ahead and merge this? I know it might conflict with some of your branches

Issue #3932, opened 2017-12-06
Picard to GATK command example conversion.


vruano:
GATK exposes Picard tools thru its command line interface. The usage syntax is different between these two tool kits and so picard code examples in javadocs don't make that much sense when displayed in GATK's own docs.

For now code examples embeded in the Picard repo should stick to picard syntax as a matter of principle. So the solution is to transform those snippets on-the-fly when we generate the gatk-docs. 

The is a good opportunity to do so in ```GATKHelpDocWorkUnitHandler``` by overloading the ```getDescription(DocWorkUnit)``` so that any occurrence of a picard.jar example is translated into gatk's version:
```

<pre>java -jar picard.jar toolName \
              ARG1=VAL1 \
              ARG2=VAL2 
</pre>
```

to this:

```
<pre>gatk toolName \
           -ARG1 VAL1 \
           --ARG2 VAL2
</pre>
```

The difficult part is to find-out the mapping of picard argument names and gatk's exposed picard argument names and whether these need single or double dash.

To this end we can use some reflection.

cmnbroad:
@vruano If you implement this, you shouldn't have to do any reflection. Just override `addCustomBindings` instead. By that point, the DownWorkUnit property map will already be populated with all of the information about the command line args that was generated by the parser, including names, with dashes and all (you'll have to inspect the map to see the structure - drill down to args, there are "name" and "synonym" properties). I think you should be able to reset the transformed description based on those.

vruano:
@cmnbroad thanks... that only would work if the example is located in the javadoc on that tool that implement that class. However it is possible that some tool documentation contains examples of other tools command since that tool and the current tool are closely related?

vruano:
I guess it is reasonable to require all examples to make no reference to any other tool except the current one.

vruano:
Got a PR on a solution.

vdauwera:
@cmnbroad Sorry if we've already discussed how we were going to address this -- do we have a path forward?

cmnbroad:
Well, my preference would be to move forward with https://github.com/broadinstitute/picard/issues/1120. Then this issue reduces to just the jar name (well, once the Picard tool doc is updated), which is trivial. Barring that, this PR could be resurrected and completed.

PR #3942, opened 2017-12-09
Addresses issue #3932

vruano:
This PR makes GATKHelpDoclet change picard command examples into GATK's equivalent. See Test cases for examples.


vruano:
@cmnbroad, can you take a look?

droazen:
@cmnbroad Should we close this one?

cmnbroad:
I think we should - the need for this should be obsoleted soon.

Issue #3853, opened 2017-11-17
All hands on deck: tool doc updates


sooheelee:
Thank you everyone for your contributions towards this documentation effort. 
Instructions from @vdauwera ~~to follow~~ at [this Google doc](https://docs.google.com/a/broadinstitute.org/document/d/1r1AV4yWP4_vNmniUDR5LojihuggMDI2OnEpfRiYyvdk/edit?usp=sharing)
Favorite tool doc examples from @vdauwera NOW in her SOP doc.
Spreadsheet from @sooheelee ~~to be~~ posted [here](https://docs.google.com/a/broadinstitute.org/spreadsheets/d/19SvP6DHyXewm8Cd47WsM3NUku_czP2rkh4L_6fd-Nac/edit?usp=sharing)

sooheelee:
I've tentatively categorized the tools and they are listed in speadsheet format at:
### https://docs.google.com/a/broadinstitute.org/spreadsheets/d/19SvP6DHyXewm8Cd47WsM3NUku_czP2rkh4L_6fd-Nac/edit?usp=sharing

- [1]   GATK4 and Picard tool categories are up for discussion. They are meant to be functional and will  be used at  <https://software.broadinstitute.org/gatk/documentation/tooldocs/current/>.   First pass by Soo Hee. If you have a better idea, please write to this issue ticket.
- [2] We can do better than minimum. At minimum, each tool has a summary description and example command. Authorship should not be picked up   by gatkDocs (but can remain in javaDoc portion of code so long as masked).
- [3] Picard tool docs are not a  priority but categorization of   Picard tools is. In the forum tool list, Picard tools will be mixed with GATK   tools alphabetically, with the PICARD label coming after the tool name.

To view docs, build with `./gradlew clean gatkDoc`, then view local index in browser.

sooheelee:
@vdauwera The tools are categorized and listed in the Google Spreadsheet above. It is waiting for you to assign tech leads to tools for documentation.

[...]

PR #3948, opened 2017-12-12
use approved program groups

tedsharpe:


sooheelee:
Thanks for checking with the docgen @tedsharpe. Looks like five of the ten tooldocs here do not show up in gatkDoc because they are missing the `@DocumentedFeature` tag:

FilterLongReadAlignmentsSAMSpark.java
SvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
CalcMetadata.java
FindSmallIndelRegions.java
ExtractOriginalAlignmentRecordsByNameSpark.java

Is this intentional? We want all tools to show up. If they are in some pre-beta status, then ChrisN was implementing an `Experimental` feature tag, much like the `Beta` feature tag, that could be used.

tedsharpe:
Deleted FindSmallIndelRegions.  The rest have been modified to emit documentation:
screen shot 2017-12-12 at 3 54 08 pm

sooheelee:
Thanks for exposing the hidden tools and deleting one Ted. Here are some comments:

- Can we use a different word than `Dump` to describe what CalcMetadataSpark does? How about "Collects statistics on reads". Also, this tool sounds like a DxQC tool. I think perhaps I should come talk to you about each of these previously hidden tools so we can correctly assign them to functional categories.
- No periods at the end of the one-line summaries
- "Find" --> "Finds", "Tool to" --> "", "discover" --> "Discovers", "run"-->"runs" etc for consistency accross tools.

sooheelee:
Okay, I've had time to sit down and go through each tool. Sorry, but I'm WFH today so I've no paper proofs to hand you.

For Jan 9 release, we are aiming for:
- Meaningful one-line summaries that convey the tool functionality
- Functional categorization of tools
- Example commands that are representative and of course that work, i.e. uses updated kebab syntax

--- 

## CalcMetadataSpark 

1. Revise one-line summary to something like:
Collects read metrics relevant to structural variant discovery

    - Notice the lack of a period at the end above.
    - Not statistics but metrics?

2. Overview and Notes could use finessing but let's leave this for next year. One thing to do now is move this statement up top:
This tool is used in development and should not be of interest to most researchers.

3. I think this tool fits under the DiagnosticsAndQCProgramGroup.java.
4. The tool takes a SAM/BAM/CRAM and calculates fragment length statistics...
5. "This is the first step in the workflow"--> makes it sound like this tool is necessary in the SV workflow but you say otherwise in the debugging sentence. I find this confusing.

6. I'm noticing that the example command does not have spark options despite the tool being a Spark tool. For such cases, it would be helpful to state, e.g. "This tool can run in both Spark and non-Spark modes, depending on if --sparkMaster is set." Then include a second example command that shows how to utilize Spark. There is an example from ChrisW in <https://github.com/broadinstitute/gatk/issues/3853>:

```
	-- \
    --sparkRunner GCS \
    --cluster my-dataproc-spark-cluster
```


---
## DiscoverVariantsFromContigAlignmentsSAMSpark

1. "Parse" is vague. How about: 
Parses aligned contig assemblies of genomic breakpoints and calls structural variants
6. 

---
## ExtractOriginalAlignmentRecordsByNameSpark

1. Subsets reads by names
2. I think you mean FilterSamReads (Picard) and not PrintReads. AFAIK, PrintReads cannot subset based on a list of read names. Rather FilterSamReads can do so as long as the reads are queryname-sorted. So then it would be good to distinguish this tool from FilterSamReads by saying (assuming true) "Unlike FilterSamReads, this tool can take any sort-order, e.g. unsorted, to subset target reads."
3. ReadDataProgramGroup.java
6.

---
## FilterLongReadAlignmentsSAMSpark

1. In the one-line summary, I'm not clear on what is meant by "Filters". Based on the result file, seems like it collects metrics on each contig alignment.
2. 
3. If metrics, then DiagnosticsAndQCProgramGroup.java
6.

---
## FindBadGenomicKmersSpark

1. The term "copy number" should be reserved in reference to CNV analyses. So instead, how about:
Identify sequence contexts that occur at high frequency in a reference
2. Please define a kmer. If only a reference fasta is required (as listed under Inputs) great. But if the tool also depends on a FAI index and DICT dictionary, please do include them. Also, it would be good to provide an example of how such information is used in SV discovery, e.g. "the resulting file can be given to FindBreakpointEvidenceSpark, which will then ignore such sequence contexts during analysis." Also would be good to mention that the default kmer size (--k-size 51) is optimized for human if indeed this is the case.
3. ReferenceProgramGroup.java
6.

---
## FindBreakpointEvidenceSpark

1. Assembles and aligns contigs of genomic breakpoint regions associated with structural variants 
2. Overview and Notes could use finessing but let's leave this for next year. One thing to include is a reference to FermiLite for those seeking more information. A publication would be best.
6. 


---
## StructuralVariationDiscoveryPipelineSpark

1. Runs the structural variant discovery workflow on a single sample in Spark 
2. Fyi we sanction a "Caveats" section, which is likely more appropriate for the PE expectation and the fact that low coverage data less than 30x will give suboptimal results. Also, should mention this workflow is meant only for WGS. Or is it the case one case use exome data? Second note on BwaMemIndexImageCreator could be consolidated with the same under Inputs. Same with third note.
6. 


---
## SvDiscoverFromLocalAssemblyContigAlignmentsSpark

1. "Parse" is vague. Please clarify one-line summary.
2. Again place up top "This tool is used in development and should not be of interest to most researchers."

---
## ParallelCopyGCSDirectoryIntoHDFSSpark

1. Let's explain the acronyms or their use context, e.g.
Parallel copy a file or directory from Google Cloud Storage into the HDFS format used in Spark.

2. GCS refers to Google Cloud Storage and HDFS to Hadoop Distributed File System. The latter is used in Spark ...

- What is the difference between RDD (resilient distributed datasets) and HDFS? 
- Can I use globbing? 
- Why do I need this tool in the SV pipeline? 
- Can the tool run in Spark and nonSpark modes?

6. 
gatk ParallelCopyGCSDirectoryIntoHDFSSpark \
          --input-gcs-path gs://my-bucket/my-data-directory/ \
          --output-hdfs-directory hdfs://my-dataproc-spark-cluster-m:8020/my-data \
          -- \
          --sparkRunner GCS \
          --cluster my-dataproc-spark-cluster

- Can we update the example command so it is more concrete, e.g. takes a BAM or multiple BAMs?

sooheelee:
@tedsharpe Seems you've already merged your changes. I've just reviewed the contents of the tooldocs and posted my comments above. Since this issue ticket is about program groups and my commnets relate to program group assignments plus doc-related content, do you want me to post these to another issue/PR ticket?

tedsharpe:
No, this is fine.  I'll make a new PR with the suggested revisions.

sooheelee:
@tedsharpe asks that we not require each Spark tool to have a Spark-specific command and @vdauwera agrees that we can refer folks to a common document on setting Spark arguments.

If example commands for Spark tools do not utilize Spark, then I think we need this statement in each tool doc (to which it applies): 
> This tool can be run without explicitly specifying Spark options. That is to say, the given example command without Spark options will run locally. See <a href ="https://software.broadinstitute.org/gatk/documentation/article?id=10060">Tutorial#10060</a> for an example of how to set up and run a Spark tool on a cloud Spark cluster.

This tutorial isn't the best, as it only focuses on Google Dataproc, but it's what we have currently to get people started. Let's just say it's a placeholder until we get something better up.

tedsharpe:
Perfect.  Do you want that language appended to the Usage Example section?  (This would be my vote.)  Or as part of the main description?

sooheelee:
Usage example section sounds good.

PR #3964, opened 2017-12-13
Install GATK Python packages.

cmnbroad:
We need a way to install GATK Python modules onto the docker image, from repo source, in a way that doesn't assume a repo clone is present on the docker image (there currently is one, but we want to remove it to recover space), and that also doesn't make the conda environment dependent on a repo clone.

This PR adds a build task that creates a zip archive of the GATK Python source; propagates that to the docker image, and then pip installs the contents of the archive into the conda environment on the docker. Since we don't have an actual python module in the repo at the moment, there is a second, temporary, commit that contains a dummy python module used only to trigger and test that the installation works.

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/utils/python/PythonEnvironmentIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/python/PythonEnvironmentIntegrationTest.java
@@ -25,6 +25,7 @@ public class PythonEnvironmentIntegrationTest {
                 { "keras" },
                 { "pymc3" },
                 { "argparse" },
+                { "gatkpython" }, // termporary test of the install-from-archive option
         };
     }
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ARG DRELEASE
 ADD . /gatk
 
 WORKDIR /gatk
-RUN /gatk/gradlew clean compileTestJava installAll localJar -Drelease=$DRELEASE
+RUN /gatk/gradlew clean compileTestJava installAll localJar createPythonPackageArchive -Drelease=$DRELEASE
 
 WORKDIR /root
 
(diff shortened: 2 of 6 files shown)


cmnbroad:
@samuelklee @mbabadi @lucidtronix Please take a look. This still doesn't address the issue of how we distribute the GATK packages to end users who are not using the docker, but there are lots of ways to do that (like PyPI, or even having the conda env install from the zip archive created by the new build target, etc.).

src/test/java/org/broadinstitute/hellbender/utils/python/PythonEnvironmentIntegrationTest.java
```diff
@@ -25,6 +25,7 @@
                 { "keras" },
                 { "pymc3" },
                 { "argparse" },
+                { "gatkpython" }, // termporary test of the install-from-archive option
```
lucidtronix: temporary

mbabadi:
@cmnbroad PyPI sound like a reasonable solution with the caveat that we need to make sure the github repo and PyPI are in sync at all times. Perhaps a less error-prone solution for git cloning users is putting together an official setup script @lbergelson and encouraging the users to install GATK via the setup script. The setup script could offer the benefits of `gradlew` while taking care of setting up the conda env and installing GATK python modules at the same time. We must also discourage the users from downloading JARs and make git clone + setup and the docker as the only two official methods of installing and using GATK.

cmnbroad:
I was thinking that if we relied on PyPI for distribution, it would only be for released builds, not a release for every repo merge commit. But, I'm increasingly inclined to think that in the short term we should just include the python archive/zip file right in the gatk distribution zip, and modify the env .yml to install from that. Then every configuration (docker image, git clone user, and end user) could use exactly the same method to establish the environment. That seems like the simplest solution for now.

cmnbroad:
I'm merging this as is (retaining 2 separate commits) so we can start merging in python branches. For now we need to retain the second commit with the dummy python package, since otherwise this would fail in the interim. Once we have some real installable python packages in place, that commit can be reverted.

PR #3966, opened 2017-12-14
GetPileupSummaries throws exception if AF field missing from vcf. Closes #3955.

### LINKED ISSUE #3955: GetPileupSummaries silently fails if AF isn't present in the VCF?
After banging my head on GetPileupSummaries for a bit, I think I've figured out a problem with the tool: it assumes that the AF INFO annotation is always present in an input VCF.  There are several ways that it could proceed with a record/VCF that doesn't have AF, but the way it currently does (silently skip all of those records) is not a good one.  I had zero indication as to why the output was empty.
FWIW, I would suggest doing the following:
1. Warn the user the first time it encounters a record with no AF that such a record will be ignored.
2. If no records contain AF then throw a User error at the end of the traversal.
3. Provide an option to the tool that tells it to assume that the AF is less than 0.2 if not present.

We should probably fix this before the release.

davidbenjamin:
@LeeTL1220 can you review this?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/GetPileupSummaries.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/GetPileupSummaries.java
@@ -16,6 +16,7 @@ import org.broadinstitute.hellbender.engine.filters.MappingQualityReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.filters.WellformedReadFilter;
+import org.broadinstitute.hellbender.exceptions.UserException;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/GetPileupSummaries.java
```diff
@@ -81,6 +82,9 @@
 
     private VariantContext lastVariant = null;
 
+    private boolean sawVariantsWithoutAlleleFrequency = false;
```
LeeTL1220: Can you update the tool docs to state that AF must appear in every variant in the VCF?

Are you sure you want to throw an exception if a variant is missing allele frequency?  Not simply output some NAs?  You know the use case better than I do, so this suggestion can be ignored.

davidbenjamin: @LeeTL1220 If only some AFs are missing no error is thrown.  I added a test that verifies this.

Issue #3979, opened 2017-12-15
GenomicsDBImport: add support for samples with spaces in the name


dianekaplan:
We are unable to run joint genotyping on the cloud when it includes a sample with a space in the name; it fails when calling GATK's GenomicsDBImport functionality (see error log below). 

This practice (samples with spaces in the name) is unfortunately not uncommon, so we've had to modify our various workflows one by one to add support for this case. Please update GenomicsDBImport to handle samples with a space in the name. 

Example error log: 
gsutil cat gs://broad-jg-dev-cromwell-execution/JointGenotyping/6918095f-ca06-4883-bcb5-f5c2e343bb6d/call-ImportGVCFs/shard-0/ImportGVCFs-0-stderr.log

Using GATK jar /usr/gitc/gatk-package-4.beta.6-local.jar
Running:
    java -Dsamjdk.use_async_io_read_samtools=false -Dsamjdk.use_async_io_write_samtools=true -Dsamjdk.use_async_io_write_tribble=false -Dsamjdk.compression_level=1 -Xmx4g -Xms4g -jar /usr/gitc/gatk-package-4.beta.6-local.jar GenomicsDBImport --genomicsDBWorkspace genomicsdb --batchSize 50 -L chr1:1-391754 --sampleNameMap /cromwell_root/broad-jg-dev-storage/freimer_dutch_fin_wgs_v1/v1/sample_map --readerThreads 5 -ip 500
Picked up _JAVA_OPTIONS: -Djava.io.tmpdir=/cromwell_root/tmp.H9t5pC
[...]

Issue #3815, opened 2017-11-09
GenomicsDBImport should throw by default if the target GenomicsDB already exists, not continue silently


droazen:
@lbergelson reports that `GenomicsDBImport` by default will continue silently when the target GenomicsDB already exists, instead of throwing, resulting in a corrupt GenomicsDB. The default should be to throw.

droazen:
This was done

PR #3982, opened 2017-12-15
Allow sample names with whitespace in sample name <--> map file

kgururaj:
Fixes:
* https://github.com/broadinstitute/gatk/issues/3815
* https://github.com/broadinstitute/gatk/issues/3979
  * Samples can have leading and trailing whitespace, thus \<space\>Sample0 is different from Sample0. The VCF spec does not appear to place any restriction on the sample name.


droazen:
@kgururaj Can you please add a good integration test for this branch?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -332,11 +332,11 @@ private static void assertGVCFHasOnlyOneSample(final String variantPath, final V
             final LinkedHashMap<String, Path> sampleToFilename = new LinkedHashMap<>();
             for ( final String line : lines) {
                 final String[] split = line.split("\\t",-1);
-                if (split.length != 2
-                        || split[0].isEmpty() || containsWhitespace(split[0])
-                        || split[1].isEmpty() || containsWhitespace(split[1])) {
+                if (split.length != 2)
+                    throw new UserException.BadInput("Expected a file of format\nSample\tFile\n but found line: " + line);
+                if(split[0].trim().isEmpty()
```
lbergelson: Could you ensure that the names don't start / end with whitespace here as well?

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportUnitTest.java
```diff
@@ -41,8 +41,7 @@
                 {"Sample1\tfile1\t"},           // extra tab
                 {"Sample1\nfile"},              // newline instead of tab
                 {"\t"},                         // only tab
-                {"Sample1 file1"},              // non-tab whitespace
-                {"Sample 1\tfile1"},            // white space in a token
+                {"Sample1 file1"}               // non-tab whitespace
```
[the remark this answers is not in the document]
lbergelson: Could you add an additional test case where the sample is only whitespace and show that that fails? 
ex 
```
{    \file1}
```

Also a case with whitespace starting the name and one with whitespace ending the sample name.  I think we should fail in those cases as well.

droazen:
Reassigning to @jamesemery to add the needed tests.

jamesemery:
@lbergelson @droazen Rebased and added a test asserting that spaces work as expected.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportUnitTest.java
```diff
@@ -41,8 +41,10 @@
                 {"Sample1\tfile1\t"},           // extra tab
                 {"Sample1\nfile"},              // newline instead of tab
                 {"\t"},                         // only tab
-                {"Sample1 file1"},              // non-tab whitespace
-                {"Sample 1\tfile1"},            // white space in a token
+                {"Sample1 file1"},              // 1 column
+                {" name\tfile1"},               // preceding whitespace
+                {"name \tfile1"},               // trailing whitespace
+                {"name\tfile1 "},               // trailing whitespace second column
```
droazen: In addition to these negative test cases, you should add some positive test cases for `GenomicsDBImport.loadSampleNameMapFile()` to prove that that method can handle sample names with internal spaces.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -394,11 +394,11 @@ private static void assertGVCFHasOnlyOneSample(final String variantPath, final V
             final LinkedHashMap<String, Path> sampleToFilename = new LinkedHashMap<>();
             for ( final String line : lines) {
                 final String[] split = line.split("\\t",-1);
-                if (split.length != 2
-                        || split[0].isEmpty() || containsWhitespace(split[0])
-                        || split[1].isEmpty() || containsWhitespace(split[1])) {
+                if (split.length != 2)
+                    throw new UserException.BadInput("Expected a file of format\nSample\tFile\n but found line: " + line);
```
droazen: Add curly braces around if statement body.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -394,11 +394,11 @@ private static void assertGVCFHasOnlyOneSample(final String variantPath, final V
...
+                if ( !split[0].trim().equals(split[0]) || split[0].trim().isEmpty()
+                        || !split[1].trim().equals(split[1]) || split[1].trim().isEmpty())
                     throw new UserException.BadInput("Expected a file of format\nSample\tFile\n but found line: " + line);
```
droazen: Add curly braces around if statement body.

droazen: Also update the tool documentation to describe the requirements for the sample name map file (eg., "Sample names in the sample name map file may contain internal non-tab spaces, but may not begin or end with whitespace.")

droazen: Also, this error message here could be better -- this is the case where either the sample name or the VCF path had leading/trailing whitespace, or was all whitespace. Error message should state that.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -499,8 +490,9 @@ public void traverse() {
             final long variantContextBufferSize = vcfBufferSizePerSample * sampleToReaderMap.size();
             final GenomicsDBImportConfiguration.ImportConfiguration importConfiguration =
                     createImportConfiguration(workspace, GenomicsDBConstants.DEFAULT_ARRAY_NAME,
-                                              variantContextBufferSize, segmentSize,
-                                              i, (i+updatedBatchSize-1));
+                            variantContextBufferSize, segmentSize,
+                            i, (i+updatedBatchSize-1),
+                            (batchCount == 1)); //Fail if array exists and this is the first batch
```
droazen: Fix indentation here (there may be literal tabs here that need converting to spaces).

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -669,6 +662,7 @@ public Object onTraversalSuccess() {
                 .setGatk4IntegrationParameters(gatk4Parameters)
                 .setSizePerColumnPartition(variantContextBufferSize)
                 .setSegmentSize(segmentSize)
+                .setFailIfUpdating(failIfArrayExists)
```
droazen: Can you explain why this change is part of this PR, and why there are no new tests to cover this change? It's unrelated to the sample-with-spaces fix. If it was to address https://github.com/broadinstitute/gatk/issues/3815, wasn't that already addressed by patching `overwriteOrCreateWorkspace()` to throw if the workspace already exists, or was that fix incomplete? @kgururaj @lbergelson

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -539,7 +531,7 @@ public Object onTraversalSuccess() {
         } catch (final FileNotFoundException fe) {
             throw new UserException("Unable to write callset map JSON file " + callsetMapJSONFile.getAbsolutePath(), fe);
         }
-        try {
+	try {
```
droazen: Fix indentation here (there may be literal tabs here that need converting to spaces).

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -638,7 +630,8 @@ public Object onTraversalSuccess() {
             final long variantContextBufferSize,
             final long segmentSize,
             final long lbSampleIndex,
-            final long ubSampleIndex) {
+            final long ubSampleIndex,
+        final boolean failIfArrayExists) {
```
droazen: Fix indentation here (there may be literal tabs here that need converting to spaces).

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -47,13 +48,16 @@
     private static final String MULTIPLOID_DATA_HG37 = largeFileTestDir + "gvcfs/HapMap5plex.ploidy10.b37.g.vcf";
     private static final String NA12878_HG37 = " src/test/resources/org/broadinstitute/hellbender/tools/haplotypecaller/expected.testGVCFMode.gatk4.g.vcf";
     private static final String ARTIFICIAL_PHASED = getTestDataDir() + "/ArtificalPhasedData.1.g.vcf";
+    private static final String HG_00268_withSpaces = largeFileTestDir + "gvcfs/HG00268.spaceInSampleName.g.vcf";
```
droazen: Use THIS_STYLE for names of constants. 

Also, why does this need to be a large file? Can't we test this functionality with a smaller file?

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -47,13 +48,16 @@
...
     private static final List<String> LOCAL_GVCFS = Arrays.asList(HG_00096, HG_00268, NA_19625);
     private static final String GENOMICSDB_TEST_DIR = publicTestDir + "org/broadinstitute/hellbender/tools/GenomicsDBImport/";
     private static final String COMBINEGVCFS_TEST_DIR = publicTestDir + "org/broadinstitute/hellbender/tools/walkers/CombineGVCFs/";
 
     private static final String COMBINED = largeFileTestDir + "gvcfs/combined.gatk3.7.g.vcf.gz";
+    private static final String TRUNCATED_COMBINED = largeFileTestDir + "gvcfs/combined.gatk3.7.truncated.g.vcf";
```
droazen: What does this name mean, `TRUNCATED_COMBINED`? How is this file truncated?

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -344,6 +348,25 @@ private static void checkGenomicsDBAgainstExpected(final String workspace, final
...
+                        "HG00096\t" + HG_00096 );
+
+        final String workspace = createTempDir("gendbtest").getAbsolutePath() + "/workspace";
+
+        ArgumentsBuilder args = new ArgumentsBuilder()
+                .addArgument(GenomicsDBImport.BATCHSIZE_ARG_LONG_NAME, String.valueOf(2))
+                .addFileArgument(GenomicsDBImport.SAMPLE_NAME_MAP_LONG_NAME, outOfOrderSampleMap).addArgument("L", IntervalUtils.locatableToString(TRUNCATED_INTERVAL))
+                .addArgument(GenomicsDBImport.WORKSPACE_ARG_LONG_NAME, workspace);
+
+        runCommandLine(args);
```
droazen: Is there a reason you can't just use the pre-existing `writeToGenomicsDB()` method here?

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
jamesemery: It doesn't need to be large, indeed it isn't a particularly large file, I just figured it made sense to keep it with the other files (considering that this is just a subsection of one of the other files)

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
jamesemery: Because that doesn't take a sample name map. Thus to test the support for whitespaces in the sample name map I must use this codepath.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
jamesemery: Its a subsection of the longer file, I shortened it to be a much smaller portion of the larger file to cut down on size, especially since it was really just a test asserting that it didn't crash.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
jamesemery: done

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
droazen: "Truncated" implies that the file is incomplete or corrupt -- recommend using "SMALLER_INTERVAL_COMBINED" and "SMALLER_INTERVAL" instead.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
jamesemery: done, I also removed the whitespace check from the path (because I don't think there should be a failure condition for that come to think of it (because the path API handles it as far as I can tell))

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: I guess there is duplication here - GenomicsDB had a flag in the C++ modules to fail if writing to an existing array. The current PR just passes the correct value to the C++ side.

I think Louis wrote a patch to handle this in the GATK code as well.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -394,14 +395,15 @@ private static void assertGVCFHasOnlyOneSample(final String variantPath, final V
             final LinkedHashMap<String, Path> sampleToFilename = new LinkedHashMap<>();
             for ( final String line : lines) {
                 final String[] split = line.split("\\t",-1);
-                if (split.length != 2
-                        || split[0].isEmpty() || containsWhitespace(split[0])
-                        || split[1].isEmpty() || containsWhitespace(split[1])) {
+                if (split.length != 2) {
                     throw new UserException.BadInput("Expected a file of format\nSample\tFile\n but found line: " + line);
```
droazen: Error message should explicitly say "expected 2 fields but found N fields"

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -638,7 +632,8 @@ public Object onTraversalSuccess() {
             final long variantContextBufferSize,
             final long segmentSize,
             final long lbSampleIndex,
-            final long ubSampleIndex) {
+            final long ubSampleIndex,
+        final boolean failIfArrayExists) {
```
droazen: Indentation here still not fixed (replace literal tab with spaces)

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -47,13 +47,16 @@
     private static final String MULTIPLOID_DATA_HG37 = largeFileTestDir + "gvcfs/HapMap5plex.ploidy10.b37.g.vcf";
     private static final String NA12878_HG37 = " src/test/resources/org/broadinstitute/hellbender/tools/haplotypecaller/expected.testGVCFMode.gatk4.g.vcf";
     private static final String ARTIFICIAL_PHASED = getTestDataDir() + "/ArtificalPhasedData.1.g.vcf";
+    private static final String HG_00268_WITH_SPACES = largeFileTestDir + "gvcfs/HG00268.spaceInSampleName.g.vcf";
     private static final List<String> LOCAL_GVCFS = Arrays.asList(HG_00096, HG_00268, NA_19625);
     private static final String GENOMICSDB_TEST_DIR = publicTestDir + "org/broadinstitute/hellbender/tools/GenomicsDBImport/";
     private static final String COMBINEGVCFS_TEST_DIR = publicTestDir + "org/broadinstitute/hellbender/tools/walkers/CombineGVCFs/";
 
     private static final String COMBINED = largeFileTestDir + "gvcfs/combined.gatk3.7.g.vcf.gz";
+    private static final String COMBINED_WITHSPACES = largeFileTestDir + "gvcfs/combined.gatk3.7.truncated.g.vcf";
```
droazen: Remove "truncated" from this file name -- truncated implies malformed, which this is not. Use a term like "smaller interval" instead.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -47,13 +47,16 @@
...
     private static final SimpleInterval INTERVAL = new SimpleInterval("chr20", 17960187, 17981445);
     private static final SimpleInterval INTERVAL_NONDIPLOID = new SimpleInterval("20", 10000000, 10100000);
+    private static final SimpleInterval TRUNCATED_INTERVAL = new SimpleInterval("chr20", 17960187, 17961973);
```
droazen: truncated -> smaller

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportUnitTest.java
```diff
@@ -41,17 +41,36 @@
...
+                {"Sample1 001\tFile", 1},          // sample names with whitespace
+                {"name name\tfile1 ", 1},          // trailing whitespace second column
+                {"name name\t file1 ", 1},         // leading and trailing whitespace second column
+        };
+    }
+
+    @Test(dataProvider = "getGoodSampleNameMapFileSyntax")
+    public void testValidSampleFiles(final String text, final int numEntries){
+        final File sampleFile = IOUtils.writeTempFile(text, "badSampleMapping", ".txt");
+        Assert.assertEquals(GenomicsDBImport.loadSampleNameMapFile(sampleFile.toPath()).size(),numEntries);
```
droazen: Also check the sample names and file names themselves to ensure that they were parsed correctly.

droazen:
Karthik addressed most of my comments in his commit, but my comment about `GenomicsDBUnitTest` is still not addressed: "Also check the sample names and file names themselves to ensure that they were parsed correctly." Once that's done we can merge this.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportUnitTest.java
```diff
@@ -41,17 +41,40 @@
...
         final File sampleFile = IOUtils.writeTempFile(text, "badSampleMapping", ".txt");
         GenomicsDBImport.loadSampleNameMapFile(sampleFile.toPath()  );
     }
 
+    @DataProvider
+    public Object[][] getGoodSampleNameMapFileSyntax(){
+        return new Object[][]{
+                // Note: none of these files are real, these are just valid files syntactically
+                {"Sample1\tsamplePath \n"
+                +"Sample2\tsamplePath", new String[][] {{"Sample1","samplePath"},{"Sample1","samplePath"}}},     // normal sample names
```
droazen: This expected value is wrong -- you check Sample1 twice, but don't check Sample2. Also, you should give Sample1 and Sample2 difference paths in the fake map.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportUnitTest.java
```diff
@@ -41,17 +41,40 @@
...
+                {"Sample1 001\tFile", new String[][] {{"Sample1 001","File"}}},          // sample names with whitespace
+                {"name name\tfile1 ", new String[][] {{"name name","file1"}}},          // trailing whitespace second column
+                {"name name\t file1 ", new String[][] {{"name name","file1"}}}        // leading and trailing whitespace second colum
+                };
+    }
+
+    @Test(dataProvider = "getGoodSampleNameMapFileSyntax")
+    public void testValidSampleFiles(final String text, final String[][] expectedEntries){
+        final File sampleFile = IOUtils.writeTempFile(text, "badSampleMapping", ".txt");
```
droazen: `badSampleMapping` -> `goodSampleMapping`

PR #3986, opened 2017-12-16
Restricting Funcotator to use `basic` transcripts only.

### LINKED ISSUE #3920: Funcotator should only use `Basic` transcripts
Funcotator should only create annotations of GENCODE transcripts with the basic tag by default.

jonn-smith:
Restricting `Funcotator` to use only `basic` `Gencode` transcripts for annotations.

Fixes #3920


LeeTL1220:
@jonn-smith this looks good to me.  Merge away.

PR #3988, opened 2017-12-18
Mutect downsampler that recognizes and skips bad mapping regions

davidbenjamin:
This greatly reduces wall clock time in M2 scatters without affecting sensitivity.  It is decoupled from HaplotypeCaller's downsampling.


LeeTL1220:
How do you know "without affecting sensitivity" ?

davidbenjamin:
> How do you know "without affecting sensitivity" ?

I ran our validations.

LeeTL1220:
@davidbenjamin I'm fine with merging as is.

Issue #3659, opened 2017-10-04
HaplotypeCallerSpark, SparkgenomeReadCounts fail reference name lookup


chapmanb:
I'm running HaplotypeCallerSpark with GATK4 beta 5 and running into an issue where it complains about a reference name lookup, reporting that a strange reference name is not found:
```
java.lang.IllegalArgumentException: Reference name for '1145119298' not found in sequence dictionary.
```
This happens with HaplotypeCallerSpark independent of the number of cores or memory requested but works fine with HaplotypeCaller using the same BED and BAM inputs. The BED file has a number of regions on chr15, chr16 and chr17 and I'm happy to share anything that would be useful. I'm still working on generating a minimal example that demonstrates the issue.

The command we're running is:
```
[...]

PR #3991, opened 2017-12-18
Upgrade to Hadoop-BAM 7.9.1

tomwhite:


tomwhite:
The main one is improved split guessing, which will fix #3659 and #2113. There's also a fix for hg38 contig names.

Changes are listed here: https://github.com/HadoopGenomics/Hadoop-BAM/blob/master/CHANGELOG.txt

PR #3992, opened 2017-12-18
Removed outdated link to CNV forum post from README.

samuelklee:


samuelklee:
@droazen another quick one for you, if you don't mind!

samuelklee:
Replacement post/tutorial/etc. is not available just yet.  In any case, it seems idiosyncratic to have such a link only for the CNV somatic workflow...I think it might've just been carried over from the gatk-protected README at some point.

PR #3998, opened 2017-12-20
Proposal: argument collection for CLP options

magicDGS:
I have three main reasons to propose to move the arguments in CLP to an argument collection that is configurable by downstream tools/projects:

1. Support hiding some arguments for downstream projects. For example, I do not want to support a config file by the user, but rather decide the settings for the framework and expose only some configuration.
1. Set custom defaults for some downstream tools (including GATK). For example, a concrete tool might want to force the temp directory to be specified to avoid failures due to no space (and specify that in the documentation).
1. Support old-style arguments (not kebab-case) for downstream projects that rely on the current argument definitions. I am specially affected by this one, because updating GATK to the 4.0.0 release of January will be a breaking change that will cause some nightmares for my users - and I don't want to do a major version bump yet (I have to re-work a bit my own framework before it)

Thus, the first commit of this PR holds the proposal for the new argument collection. As I know that the team is also trying to normalize arguments and documentation, I included two more commits to help with the task (they can be removed if you think that it is better after the argument collection):
* Use `java.nio.Path` for temp directories (to support temp directories in HDFS, for example)
* Change arguments moved to the collection to kebab-case (to help with #3853)

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -142,20 +142,25 @@ public abstract class CommandLineProgram implements CommandLinePluginProvider {
                 }
             }
             try {
-                // TODO - not sure if this will work as before, because using File.setReadable(true, false) and
-                // TODO - f.setWritable(true, false) was done in two steps, so if one failed the other was done
-                // TODO - I don't know if the setPosixFilePermissions try to set all and then throws an exception
-                // TODO - if any of them fails, or just has an early termination
-                // TODO - an option to avoid this problem is iterate over a list of the permissions and set them
-                // TODO - one by one, ignoring possible errors (this was done in the original commit/PR)
-                Files.setPosixFilePermissions(p, new HashSet<>(Arrays.asList(
+                // add the READ/WRITE permissions to the actual permissions of the file
+                final Set<PosixFilePermission> permissions = Files.getPosixFilePermissions(p);
+                // only set the permissions if they change
+                if (permissions.addAll(Arrays.asList(
                         PosixFilePermission.OWNER_READ, PosixFilePermission.GROUP_READ, PosixFilePermission.OTHERS_READ,
-                        PosixFilePermission.OWNER_WRITE, PosixFilePermission.GROUP_WRITE, PosixFilePermission.OTHERS_WRITE)
-                ));
-            } catch (UnsupportedOperationException | IOException e) {
+                        PosixFilePermission.OWNER_WRITE, PosixFilePermission.GROUP_WRITE, PosixFilePermission.OTHERS_WRITE))) {
+                    Files.setPosixFilePermissions(p, permissions);
+                }
+            } catch (final UnsupportedOperationException | IOException e) {
                 // intentionally ignoring
+                // TODO - logging can be removed, but I think that it might be informative
+                // TODO - maybe it should be debug
+                logger.warn("Temp directory {}: unable to set permissions due to {}", p, e.getMessage());
             }
-            System.setProperty("java.io.tmpdir", p.toAbsolutePath().toUri().toString()); // in loop so that last one takes effect
+
+            // TODO - this should be p.toAbsolutePath().toUri().toString() to allow other FileSystems to be used
+            // TODO - but it did not work with the default FileSystem because it appends the file:// scheme
+            // TODO - maybe a method in IOUtils for converting a Path to String is required for handling this (and can be re-used in other places when it is required)
+            System.setProperty("java.io.tmpdir", p.toAbsolutePath().toString()); // in loop so that last one takes effect
         }
 
         //Set defaults (note: setting them here means they are not controllable by the user)
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -9,24 +9,22 @@ import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.metrics.StringHeader;
 import htsjdk.samtools.util.BlockCompressedOutputStream;
 import htsjdk.samtools.util.BlockGunzipper;
-import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.Log;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.barclay.argparser.*;
+import org.broadinstitute.hellbender.cmdline.argumentcollections.CLPConfigurationArgumentCollection;
+import org.broadinstitute.hellbender.cmdline.argumentcollections.GATKDefaultCLPConfigurationArgumentCollection;
 import org.broadinstitute.hellbender.utils.LoggingUtils;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.config.ConfigFactory;
 import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
 import org.broadinstitute.hellbender.utils.help.HelpConstants;
 
-import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetAddress;
-import java.net.MalformedURLException;
 import java.net.URL;
-import java.net.URLClassLoader;
 import java.text.DecimalFormat;
 import java.time.Duration;
 import java.time.ZonedDateTime;
@@ -36,6 +34,10 @@ import java.util.Collections;
 import java.util.List;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.*;
 import java.util.stream.Collectors;
 
 /**
@@ -61,38 +63,12 @@ public abstract class CommandLineProgram implements CommandLinePluginProvider {
     // abstract, this is fine (as long as no logging has to happen statically in this class).
     protected final Logger logger = LogManager.getLogger(this.getClass());
 
-    @Argument(common=true, optional=true)
-    public List<File> TMP_DIR = new ArrayList<>();
-
     @ArgumentCollection(doc="Special Arguments that have meaning to the argument parsing system.  " +
             "It is unlikely these will ever need to be accessed by the command line program")
     public SpecialArgumentsCollection specialArgumentsCollection = new SpecialArgumentsCollection();
 
-    @Argument(fullName = StandardArgumentDefinitions.VERBOSITY_NAME, shortName = StandardArgumentDefinitions.VERBOSITY_NAME, doc = "Control verbosity of logging.", common = true, optional = true)
-    public Log.LogLevel VERBOSITY = Log.LogLevel.INFO;
-
-    @Argument(doc = "Whether to suppress job-summary info on System.err.", common=true)
-    public Boolean QUIET = false;
-
-    @Argument(fullName = "use-jdk-deflater", shortName = "jdk-deflater", doc = "Whether to use the JdkDeflater (as opposed to IntelDeflater)", common=true)
-    public boolean useJdkDeflater = false;
-
-    @Argument(fullName = "use-jdk-inflater", shortName = "jdk-inflater", doc = "Whether to use the JdkInflater (as opposed to IntelInflater)", common=true)
-    public boolean useJdkInflater = false;
-
-    @Argument(fullName = "gcs-max-retries", shortName = "gcs-retries", doc = "If the GCS bucket channel errors out, how many times it will attempt to re-initiate the connection", optional = true)
-    public int NIO_MAX_REOPENS = ConfigFactory.getInstance().getGATKConfig().gcsMaxRetries();
-
-    // This option is here for documentation completeness.
-    // This is actually parsed out in Main to initialize configuration files because
-    // we need to have the configuration completely set up before we create our CommandLinePrograms.
-    // (Some of the CommandLinePrograms have default values set to config values, and these are loaded
-    // at class load time as static initializers).
-    @Argument(fullName = StandardArgumentDefinitions.GATK_CONFIG_FILE_OPTION,
-              doc = "A configuration file to use with the GATK.",
-                common = true,
-                optional = true)
-    public String GATK_CONFIG_FILE = null;
+    @ArgumentCollection
+    public CLPConfigurationArgumentCollection configArgs = getClpConfigurationArgumentCollection();
 
     private CommandLineParser commandLineParser;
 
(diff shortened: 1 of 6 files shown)


magicDGS:
Can you have a look to this proposal, @droazen? I really need to have this in before the release of GATK4 to be able to update my dependency for the release one. Otherwise, I will need a version bump or a hacked CLP class (which I prefer to avoid).

Thank you very much in advance!

lbergelson:
@magicDGS

It looks like there's some sort of compilation error here:

```
/gatk/src/test/java/org/broadinstitute/hellbender/tools/walkers/markduplicates/MarkDuplicatesGATKIntegrationTest.java:180: error: incompatible types: inference variable T has incompatible bounds
        ((GATKDefaultCLPConfigurationArgumentCollection) markDuplicatesGATK.configArgs).TMP_DIR = CollectionUtil.makeList(outputDir);
                                                                                                                         ^
    equality constraints: String
    lower bounds: File
  where T is a type-variable:
    T extends Object declared in method <T>makeList(T...)
/gatk/src/test/java/org/broadinstitute/hellbender/tools/walkers/markduplicates/MarkDuplicatesGATKIntegrationTest.java:256: error: incompatible types: inference variable T has incompatible bounds
        ((GATKDefaultCLPConfigurationArgumentCollection) markDuplicatesGATK.configArgs).TMP_DIR = CollectionUtil.makeList(outputDir);
                                                                                                                         ^
    equality constraints: String
    lower bounds: File
  where T is a type-variable:
    T extends Object declared in method <T>makeList(T...)
```

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -133,37 +120,46 @@ public final Object runTool(){
...
+                    Files.createDirectories(p);
+                } catch (IOException e) {
+                    // intentionally ignoring
+                }
+            }
+            for (final PosixFilePermission perm: Arrays.asList(
+                    PosixFilePermission.OWNER_READ, PosixFilePermission.GROUP_READ, PosixFilePermission.OTHERS_READ,
+                    PosixFilePermission.OWNER_WRITE, PosixFilePermission.GROUP_WRITE, PosixFilePermission.OTHERS_WRITE)) {
+                try {
+                    Files.setPosixFilePermissions(p, Collections.singleton(perm));
```
lbergelson: Why loop through the permissions one at a time?  Wouldn't it make sense to pass them all in as a to a single call of Files.setPosixFilePermissions?

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/CLPConfigurationArgumentCollection.java
```diff
@@ -0,0 +1,37 @@
...
+ * Arguments used to configure the {@link org.broadinstitute.hellbender.cmdline.CommandLineProgram}.
+ *
+ * <p>The purpose of this abstract class is to allow downstream projects to decide which configuration arguments they
+ * want to expose to the CLP.
+ *
+ * <p>In addition, this class might contain arguments for documentation-only (e.g., arguments from a wrapper script).
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public abstract class CLPConfigurationArgumentCollection implements Serializable {
```
lbergelson: Why not make this an interface instead?

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/CLPConfigurationArgumentCollection.java
```diff
@@ -0,0 +1,37 @@
...
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Return a non-empty list of temp directories.
+     */
+    public abstract List<Path> getResolvedTmpDirectories();
```
lbergelson: I would just call this `getTmpDirectories`.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/GATKDefaultCLPConfigurationArgumentCollection.java
```diff
@@ -0,0 +1,101 @@
...
+    public static final String QUIET_NAME = "quiet";
+    public static final String USE_JDK_DEFLATER_LONG_NAME = "use-jdk-deflater";
+    public static final String USE_JDK_DEFLATER_SHORT_NAME = "jdk-deflater";
+    public static final String USE_JDK_INFLATER_LONG_NAME = "use-jdk-inflater";
+    public static final String USE_JDK_INFLATER_SHORT_NAME = "jdk-inflater";
+    public static final String NIO_MAX_REOPENS_LONG_NAME = "gcs-max-retries";
+    public static final String NIO_MAX_REOPENS_SHORT_NAME = "gcs-retries";
+
+    @Argument(fullName = TMP_DIR_NAME, doc = "List of temp directories to use.", common=true, optional=true)
+    public List<String> TMP_DIR = new ArrayList<>();
```
lbergelson: While we're making these changes, could you camel case these variable names?  They've always been weirdly inconsistent with the rest of gatk.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: There are two reasons to implement each permission one by one:
* In the `File` implementation, `File.setReadable` and `File.setWritable`were separated. If one of them failed, the returned value was ignored and the next was applied. With `Files.setPosixFilePermissions` it might fail earlier and thus the same behavior won't be always the case. For example, if the `FileSystem` does not implement readable permissions but just writable ones, perhaps it does not set the writting permissions...
* In the file implementation, the second param (ownerOnly) is false, which determines that the permissions are not only for the user but for everybody. If one of the permissions fail, I am not sure how `Files.setPosixFilePermissions` would behave. The same if the `FileSystem` implementation does throw an error if not owner/group/others is allowed.

I will add at the end of this PR a commit with your solution. Feel free to revert that commit to get this merge as soon as possible to enter in the release.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/CLPConfigurationArgumentCollection.java
magicDGS: I had default implementations before, but I remove them before the PR. Made an interface in the commit for addressing the comments.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/CLPConfigurationArgumentCollection.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/GATKDefaultCLPConfigurationArgumentCollection.java
magicDGS: Done.

magicDGS:
@lbergelson and @droazen - back to you and thank you for considering this when I know that you are about to release.

Please, feel free to modify the PR, rebase to test if there is any incompatibility and/or revert commits. I would like to have this for the released version to be able to remove the configuration argument, which might be confusing for my users, and to been able to use `java.nio.Path` as a temp directory. Thanks in advance!

magicDGS:
@lbergelson and @droazen - any chance to get this in before release? I am afraid that the next point-release will be too late for me to update my software and use deprecated code...

droazen:
Sorry @magicDGS -- since this is failing tests and needs a rebase/review, and we're extremely pressed for time this morning, this is going to have to wait until the next point release. But don't worry, I think I can safely say that we plan to do point releases frequently -- I'd expect the first one within a couple of weeks.

lbergelson:
@magicDGS There's some test failures here still:

```
        at org.broadinstitute.hellbender.tools.funcotator.Funcotator.closeTool(Funcotator.java:174)
        at org.broadinstitute.hellbender.engine.GATKTool.doWork(GATKTool.java:893)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:110)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:175)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:194)
        at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:152)
        at org.broadinstitute.hellbender.Main.instanceMain(Main.java:138)
        at org.broadinstitute.hellbender.Main.instanceMain(Main.java:179)
        at org.broadinstitute.hellbender.CommandLineProgramTest.runCommandLine(CommandLineProgramTest.java:32)
        at 
```

And it looks like it needs another rebase to account for the argument renaming that happened right before the 4.0.0.0 release.  Once those are resolved I think we should be good to merge and get this in 4.0.1.0

droazen:
After the rebase and tests are passing, I want to do a quick review pass on this.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -141,45 +124,59 @@ public final Object runTool(){
...
+                // TODO - an option to avoid this problem is iterate over a list of the permissions and set them
+                // TODO - one by one, ignoring possible errors (this was done in the original commit/PR)
+                Files.setPosixFilePermissions(p, new HashSet<>(Arrays.asList(
+                        PosixFilePermission.OWNER_READ, PosixFilePermission.GROUP_READ, PosixFilePermission.OTHERS_READ,
+                        PosixFilePermission.OWNER_WRITE, PosixFilePermission.GROUP_WRITE, PosixFilePermission.OTHERS_WRITE)
+                ));
+            } catch (UnsupportedOperationException | IOException e) {
+                // intentionally ignoring
+            }
+            System.setProperty("java.io.tmpdir", p.toAbsolutePath().toUri().toString()); // in loop so that last one takes effect
```
magicDGS: It looks like this line is causing the failure, because `java.io.tmpdir` is using the full URI even for the default FS. In my local computer it appears a folder called `file://` with the sub-folder structure following the `$TMPDIR` variable set in my system.

I would try locally if the `p.toAbsolutePath().toString()` works in this case, but I am not sure if this will work for `java.nio.Path` in other filesystems. Any idea @droazen and/or @lbergelson?

src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/GATKDefaultCLPConfigurationArgumentCollection.java
```diff
@@ -0,0 +1,102 @@
...
+    public String gatkConfigFile = null;
+
+    private final List<Path> resolvedTmpDirectories = new ArrayList<>();
+
+    @Override
+    public List<Path> getTmpDirectories() {
+        if (resolvedTmpDirectories.isEmpty()) {
+            // Provide one temp directory if the caller didn't
+            if (tmpDir == null || tmpDir.isEmpty()) {
+                resolvedTmpDirectories.add(IOUtil.getDefaultTmpDirPath());
```
magicDGS: In addition, it looks like the `IOUtil.getDefaultTmpDirPath` is broken in HTSJDK, because it keeps generating in sub-subsequent calls the `urer.dir`: in my local computer, there are some tests where there are at least 50 `daniel` sub-folders!

magicDGS:
@droazen - this is prepared for a review. All the test pass, but there are several an annoying hacks:

* Using the java property `"java.io.tmpdir` for default tmp directory instead of HTSJDK method (it looks somehow broken, and it is not used anymore)
* Using `java.nio.Path.toString` directly, instead of using the URI form. Otherwise, the temp directory won't work for the default FileSystem. I guess that a better option is check if that path is associated with the default FS and if so do not use the URI and otherwise use it. What do you think?

Waiting for your first pass once test show up to do not fail. Thank you!

magicDGS:
@droazen - can you have a look to this one? I will need it to update to the latest master (or point release if it happen soon).

magicDGS:
Friendly ping here, @droazen and @lbergelson - Thank you!

droazen:
I'm not sure that we want to go down this path of having to actually change an interface every time we want to add a new top-level argument to the GATK. Let me discuss this PR with the other GATK devs to get their thoughts -- we'll get back to you on this soon.

magicDGS:
I think that it is necessary to have a way for downstream projects to override some of the top-level arguments in the base CLP class. For example, the config file is for documentation purposes, but I don't want to expose users to that argument because I will set the defaults programmatically. Another example is the GCS retries, which might not be useful for a software that is not planning to support GCS even if it is already implemented (or does not want to expose).

As a downstream developer, for me it is important to being able to configure arguments and expose/hide them to my final users; with the current implementation, my main issue is to have an argument that are irrelevant for the toolkit user and that I get questions about why and how to use them (the most clear example, the config file).

If the main problem is to change an interface, a default value for new methods can be added to keep the same behavior.

magicDGS:
@droazen - any news about thi one? I would like to have some control over the CLP class arguments, such as the config-file one. Thank you!

droazen:
@magicDGS Sorry for the delay on this one -- I'll try to talk it over with the other GATK devs soon. One question: would a generic "hide/disable this list of arguments" feature in barclay be a workable solution for you? That seems like a viable alternate approach that could spare GATK from having to wrap all of our arguments in interfaces.

magicDGS:
I don't think that hiding/disable arguments would work in every case: sometimes, an argument shouldn't be exposed but still available to set programmatically, or maybe just reduce visibility making it `@Hidden` and/or `@Advance`.

What is the problem of making an interface for the top-level argument to the GATK? Changing the interface or the `CommadnLineProgram` has the same effect, but the API user can still behave the same as before. It is much more extensible and downstream-friendly.

What's about making the `CLPConfigurationArgumentCollection` an interface always returning defaults to be able to change it in a proper way? The cycle of development of a new argument will be: 1) add a new method to the interface with a default returning what will be expected from the previous behaviour, 2) add and return by the argument in the GATK implementation, 3) use the getter in the CLP for perform the operation. This only adds the first point, and operating in 3 classes instead of 3. For API user it is really easy to maintain the previous behavior when upgrading the dependency by just using their own implementation of the class, or include the top-level new arguments by using the GATK implementation. It is much more flexible and extensible (I always think about GATK also as a library).

In addition, I think that this approach is also important for evolving GATK. For example, if a new top-level argument is tagged as experimental (still not supported but requested in Barclay), removing it would allow to keep the interface (no version bump) the same and final users can still operate with the experimental argument.

The same applies to the `GATKTool` base class (https://github.com/broadinstitute/gatk/issues/4341), and for downstream projects the aim should be to be able to extend safely the `CommandLineProgram` directly to implement their own toolkit using the powerful GATK framework.

magicDGS:
@droazen - in the meantime that this is considered, I open 2 PRs that includes some minor changes to the base code compared to this one, but are directly related:

- #4468 for extracting constants for the arguments (that would be a temporary solution for my higher priority use case)
- #4469 for support NIO in `--TMP_DIR`

Can you or @lbergelson have a look to any of them?

magicDGS:
Perfect @cmnbroad - I'll wait until #4469 gets merged, and then I will come back to this.

magicDGS:
Closing in favor of https://github.com/broadinstitute/gatk/pull/5137, now that #4469 is in.

PR #4000, opened 2017-12-20
Added tests to GenomicsDBImport testing for non-diploid genotyping

### LINKED ISSUE #3342: GenomicsDB does not support non-diploid data
As discussed in https://github.com/broadinstitute/dsde-docs/issues/2303, `GenomicsDB` does not yet support non-diploid data. We should remedy this!

### LINKED ISSUE #3345: GenomicsDB should preserve phasing information in the GT field

### LINKED ISSUE #3411: GenomicsDB instances must always have the same absolute file path
It has recently come to our (Hail team's) attention that GenomicsDB instances [cannot be moved from the creation location](https://github.com/Intel-HLS/GenomicsDB/issues/135).

Either GenomicsDB will need to generalize to work on instances stored at absolute paths differing from their creation location, or GATK will need to provide the absolute path of the creation location so that Hail can construct the appropriate directory structure to read from a GenomicsDB instance.

Please advise as to GATK's plan with respect to delivering GenomicsDB instances and their absolute creation paths.

cc: @cseed @jackgoldsmith4

jamesemery:
Also added a (Deactivated) test for moving a genomicsDB

fixes #3342 
fixes #3345 
fixes #3411


src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -42,18 +43,23 @@
     private static final String HG_00096 = largeFileTestDir + "gvcfs/HG00096.g.vcf.gz";
     private static final String HG_00268 = largeFileTestDir + "gvcfs/HG00268.g.vcf.gz";
     private static final String NA_19625 = largeFileTestDir + "gvcfs/NA19625.g.vcf.gz";
+    private static final String NA_12878_PHASED = largeFileTestDir + "NA12878.phasedData.Chr20.vcf";
```
ldgauthier: Can you note that this is a VCF with HaplotypeCaller PGT:PID phasing info, which isn't really what you wanted for phasing, but it's nice to test it anyway.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -455,14 +551,52 @@ public void testPreserveContigOrderingInHeader() throws IOException {
...
+                                                     String referenceGenome, boolean produceGTField) throws IOException {
+        //Produce temporary JSON query config file
+        String indentString = "    ";
+        String queryJSON = "{\n";
+        queryJSON += indentString + "\"scan_full\": true,\n";
+        queryJSON += indentString + "\"workspace\": \""+tiledbWorkspace+"\",\n";
+        queryJSON += indentString + "\"array\": \""+GenomicsDBConstants.DEFAULT_ARRAY_NAME+"\",\n";
+        queryJSON += indentString + "\"vid_mapping_file\": \""+vidMappingFile+"\",\n";
+        queryJSON += indentString + "\"callset_mapping_file\": \""+callsetMappingFile+"\",\n";
+        queryJSON += indentString + "\"produce_GT_field\": true,\n";
```
ldgauthier: This piqued my interest for a totally unrelated reason -- is this how you get the GDB to echo back the GTs that went in instead of converting to ./.?  Or is it an option to produce sites-only output?  Either one would be useful.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -186,8 +265,9 @@ private static void checkJSONFilesAreWritten(final String workspace) {
 	Assert.assertTrue(new File(workspace, GenomicsDBConstants.DEFAULT_VCFHEADER_FILE_NAME).exists());
     }
 
-    private static void checkGenomicsDBAgainstExpected(final String workspace, final SimpleInterval interval, final String expectedCombinedVCF) throws IOException {
+    private static void checkGenomicsDBAgainstExpected(final String workspace, final SimpleInterval interval, final String expectedCombinedVCF, final String referenceFile, final boolean testAll) throws IOException {
         final GenomicsDBFeatureReader<VariantContext, PositionalBufferedStream> genomicsDBFeatureReader =
+                getGenomicsDBFeatureReader(workspace, referenceFile, !testAll);
 		getGenomicsDBFeatureReader(workspace, b38_reference_20_21);
```
ldgauthier: I think this is a spurious line from the old code.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
jamesemery: Its what you need to get the GTs back. Unfortunately for the time being it has to be done with the query json file, though we have asked for a easier api solution so that might eventually show up.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
jamesemery: Oh yeah, thats right. I will mark it. I ended up adding the artifical phased data  file for this purpose.

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
jamesemery: removed

PR #4001, opened 2017-12-20
Better hom alt vs het discrimination in CalculateContamination

davidbenjamin:
@LeeTL1220 I have tested this on:

* 5 FFPE bams from Chip that were having problems before
* mixtures of HCC1143 with NA12878 from contamination of 0 to 15%
* mixtures of HCC1143 with NA12891 and NA12892 from total contamination of 0 to 15%
* NA12878 ice
* NA12878 PCR-free wgs samples with high and low coverage
* synthetic LoH data

It is working very nicely on everything.


src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
```diff
@@ -57,17 +63,30 @@
```
LeeTL1220: Can we have some docs (does not have to be here) about how the algorithm works?

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
```diff
@@ -89,73 +108,144 @@
...
-                : filterSites(PileupSummary.readPileupSummaries(matchedPileupSummariesTable));
+        final List<PileupSummary> sites = filterSites(PileupSummary.readFromFile(inputPileupSummariesTable));
 
         // used the matched normal to genotype if available
-        final List<PileupSummary> pileupSummariesForGenotyping = matchedPileupSummaries == null ? pileupSummaries : matchedPileupSummaries;
-        final List<PileupSummary> homAltSites = findHomAltSites(pileupSummariesForGenotyping);
+        final List<PileupSummary> genotypingSites = matchedPileupSummariesTable == null ? sites :
+                filterSites(PileupSummary.readFromFile(matchedPileupSummariesTable));
+
+        final List<List<PileupSummary>> genotypingSegments = findSegments(genotypingSites);
```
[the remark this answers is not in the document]
LeeTL1220: I might be missing something here.  Can you just add a comment explaining (in little amount of text) why you segment the genotypingSites?

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
```diff
@@ -89,73 +108,144 @@
...
 
-        final List<PileupSummary> homAltPileups = matchedPileupSummaries == null ? homAltSites : getCorrespondingHomAltSites(homAltSites, pileupSummaries);
 
-        final Pair<Double, Double> contaminationAndError = calculateContamination(homAltPileups);
-        ContaminationRecord.writeContaminationTable(Arrays.asList(new ContaminationRecord(ContaminationRecord.Level.WHOLE_BAM.toString(), contaminationAndError.getLeft(), contaminationAndError.getRight())), outputTable);
+        List<PileupSummary> homAltGenotypingSites = new ArrayList<>();
+        final MutableDouble genotypingContamination = new MutableDouble(INITIAL_CONTAMINATION_GUESS);
+
+        for (int iteration = 0; iteration < MAX_ITERATIONS; iteration++) {
```
LeeTL1220: There is no code already for handling max iterations and a convergence criteria?  If not, or this would take too long to integrate then disregard this comment.

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
davidbenjamin: Done in mutect2.pdf

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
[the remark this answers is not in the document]
davidbenjamin: > Would you rather read in the output of a CNV run on the normal? Or maybe that is an option in the future?

That would be nice for the future.  There's more to coordinate than just this.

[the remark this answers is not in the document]
davidbenjamin: > I might be missing something here. Can you just add a comment explaining (in little amount of text) why you segment the genotypingSites?

Comment added.

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
davidbenjamin: There is no code

davidbenjamin:
Back to you, @LeeTL1220

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
LeeTL1220: Right, forgot...

LeeTL1220:
@davidbenjamin If it passes testing and the conflict resolution is trivial, feel free to merge

Issue #3981, opened 2017-12-15
Restore PadTargets padding/merging behavior to PreprocessIntervals.


samuelklee:
Currently in master, intervals from the IAC that have been modified by the engine are further padded, overlapping intervals are merged, and bins are created.  WES should be run by specifying bin_length = 0, and the total number of intervals (targets, in this case) should remain fixed.  However, current behavior does not prevent intervals that overlap after padding from being merged.

Note also that in sl_gcnv_ploidy_cli, bins that contain only Ns are also dropped.

We should add validation of the IAC to check for minimal modification by the engine (as is done by other CNV CLIs), and then pad without introducing overlaps or merging if bin_length = 0.  If bin length != 0, then the current behavior is fine (since it is conceivable that one might want to bin when running WES).

samuelklee:
@LeeTL1220 Not sure if you will absolutely need this for HCC1143 WES validation, but just be aware that this change is coming soon.

LeeTL1220:
Sure.  Sounds good that it will be fixed.  I doubt that it will affect results too much.

PR #3925, opened 2017-12-06
Added CLIs and WDL for python gCNV pipeline.

samuelklee:

PR #4010, opened 2017-12-23
Added documentation and kebab-case updates for ModelSegments pipeline (excluding PreprocessIntervals), along with some WDL updates.

samuelklee:
@sooheelee Ready for your review (after the break, of course).  Documentation for the CLIs should be considered a rough first cut; I tried to highlight all of the possible workflows in a non-technical manner, while linking to the technical details in the Javadoc for helper classes for those who are interested.  Feel free to make changes directly to the branch.

@LeeTL1220 This includes that change to kernel-variance-allele-fraction that we discussed.

I need to make some non-trivial code changes to PreprocessIntervals (see #3981), so I'll make some further updates to the docs for that tool in a separate PR.


samuelklee:
Also, a reminder to myself to update the program groups when they're available.

samuelklee:
A reminder to myself: I forgot to set —interval-merging-rule on some of the examples.   Will go back and check all of the examples with test files at some point.

samuelklee:
OK, went back to set --interval-merging-rule and check examples.  @sooheelee this is ready for you.

scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl
```diff
@@ -386,30 +381,30 @@ task ModelSegments {
...
-            --smoothingThresholdAlleleFraction ${default="2.0" smoothing_threshold_allele_fraction} \
-            --maxNumSmoothingIterations ${default=10 max_num_smoothing_iterations} \
-            --numSmoothingIterationsPerFit ${default=0 num_smoothing_iterations_per_fit} \
+            --denoised-copy-ratios ${denoised_copy_ratios} \
+            --allelic-counts ${allelic_counts} \
+            ${"--normal-allelic-counts " + normal_allelic_counts} \
+            --minimum-total-allele-count ${default="30" min_total_allele_count} \
+            --genotyping-homozygous-log-ratio-threshold ${default="-10.0" genotyping_homozygous_log_ratio_threshold} \
+            --genotyping-base-error-rate ${default="0.05" genotyping_base_error_rate} \
+            --maximum-number-of-segments-per-chromosome ${default="500" max_num_segments_per_chromosome} \
```
samuelklee: The default value for maximum-number-of-segments-per-chromosome here needs to be synced with the default in the jar.

samuelklee:
@sooheelee Just to let you know, we have a couple of other branches that are stacked on top of this PR (and also need to be reviewed), so we should aim to get this in first.  Next in the stack is sl_gcnv_ploidy_cli (#3925), followed by sl_preprocess (#4027).

sooheelee:
Fyi, I updated the program groups. Back to you @samuelklee.

samuelklee:
Great, thanks, I’ll take a look soon.  However, just one note: as mentioned above, I didn’t make any changes to PreprocessIntervals here.  I instead made them in #4027.  Sorry for the confusion, but I think it’ll be easiest if I revert your changes here and you instead take a look at the docs in that PR.

sooheelee:
Sure, no problem. I only read the paragraph I was tagged in and can look at the separate PR for PreprocessIntervals. Thanks for all this work.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/AnnotateIntervals.java
```diff
@@ -44,14 +45,14 @@
  *
  * <ul>
  *     <li>
- *         GC-content annotated-intervals file.
- *         This is a TSV with a SAM-style header containing a sequence dictionary,
- *         a row specifying the column headers contained in {@link AnnotatedIntervalCollection.AnnotatedIntervalTableColumn},
+ *         GC-content annotated-intervals file in TSV format.
+ *         This is a text-based file with tab-separated column values, a SAM-style header containing a sequence dictionary,
+ *         a row specifying the column headers, e.g. CONTIG, START, END, GC_CONTENT,
```
samuelklee: @sooheelee If possible, I'd rather not include the column headers explicitly in the Javadoc here, since it's just one more place things can get out of sync if they are ever changed.  The way I originally had it, you'll see that I added extra Javadoc to AnnotatedIntervalCollection.AnnotatedIntervalTableColumn so that the headers show up there.

Unless you have strong feelings about the matter (or unless the link doesn't work properly for some reason when the html documentation is generated---I didn't check this), I'll revert.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/CollectAllelicCounts.java
```diff
@@ -28,22 +28,22 @@
...
- *         BAM file.
+ *         SAM format read data
  *     </li>
  *     <li>
- *         Reference file.
+ *         Reference FASTA file
  *     </li>
  *     <li>
- *         Sites at which allelic counts will be collected.
+ *         Sites at which allelic counts will be collected, e.g. sites of common germline variation
```
samuelklee: Since we are putting this tool under CoverageAnalysisProgramGroup, you could imagine it might be used more generally to collect counts at sites.  No reason to point out sites of common germline variation specifically.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/AnnotateIntervals.java
samuelklee: Also, it appears this is the only place you added the column headers?

samuelklee: And finally, I think you may have modified the copy-and-pasted text "This is a TSV with a SAM-style header..." differently throughout the PR.  I'll try to make these more consistent.

samuelklee:
@sooheelee  To answer your question about --interval-merging-rule, the CNV tools that take -L also check to make sure that --interval-merging-rule is set to OVERLAPPING_ONLY (rather than the default of ALL), otherwise an exception is thrown.  If we did not fail, then accidentally using ALL would merge directly adjacent intervals; these occur often in CNV analyses (e.g., bins for coverage collection in WGS), and it is difficult to imagine a scenario in which we'd want to merge them, so we explicitly forbid it.

Perhaps better would be to add an --interval-merging-rule NONE option and make that the default for all tools, but that is a discussion for engine team and @droazen.

Will revert a few changes, resolve conflicts (from the `gatk-launch -> gatk` PR), and merge if that's OK with you.  Thanks for cleaning things up!

PR #4027, opened 2018-01-02
Changed PreprocessIntervals behavior to not merge intervals that overlap after padding and updated docs.

### LINKED ISSUE #3981: Restore PadTargets padding/merging behavior to PreprocessIntervals.

samuelklee:
Just the last commit.  Rebased on top of #3925.
  
Closes #3981.


samuelklee:
@MartonKN please review code changes, @sooheelee please review docs.  You can ignore the fact that tests are failing for the time being; these are just the gCNV WDL tests, see note in #3925.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
+ * <p>
+ *     The input intervals are first checked for overlapping intervals, which are merged.
+ *     The resulting intervals are then padded.  The padded intervals are then split into bins.
+ *     Finally, bins that contain only Ns are filtered out.
+ * </p>
  *
- * <p>Using the -P flag, the user can specify the amount of padding (in bp) added to each side of the intervals.
- * This padding is in addition to the padding added by the -ip (or --interval-padding) argument of
- * IntervalArgumentCollection. However, we encourage using only the -P flag.
+ * <h3>Input</h3>
```
sooheelee: --> Inputs

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
  *
- * <p>The user can also specify the length of the bins (in bp) using the -BL option. If this is not commensurate with
- * the length of the padded intervals, then the last bin will be of different length than the others.  If zero is
- * specified, then no binning will be performed.</p>
+ * <ul>
+ *     <li>
+ *         Reference file.
```
sooheelee: --> FASTA reference

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
 import htsjdk.samtools.util.Interval;
 import htsjdk.samtools.util.IntervalList;
+import htsjdk.samtools.util.IntervalUtil;
 import org.apache.commons.math3.util.FastMath;
 import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.barclay.argparser.BetaFeature;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 import org.broadinstitute.hellbender.cmdline.programgroups.CopyNumberProgramGroup;
```
sooheelee: Should be: 
```
import org.broadinstitute.hellbender.cmdline.programgroups.IntervalsProgramGroup;
```

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
+ *     <li>
+ *         Reference file.
+ *     </li>
+ *     <li>
+ *         Intervals to be preprocessed.
+ *         The argument {@code interval-merging-rule} must be set to {@link IntervalMergingRule#OVERLAPPING_ONLY}
+ *         and all other common arguments for interval padding or merging must be set to their defaults.
+ *         If no intervals are specified, then each contig will be assumed to be a single interval and binned accordingly;
+ *         this produces bins appropriate for whole genome sequencing analyses.
+ *     </li>
```
sooheelee: Thanks for clarifying the use-cases.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
+ *     <li>
+ *         Padding length (in bp).
+ *         Use {@code padding} to specify the size of each of the regions added to both ends of the intervals that result
+ *         after overlapping intervals have been merged.  Do not use the common {@code interval-padding} argument, which must
+ *         instead be set to zero.  Intervals that would overlap after padding by the specified amount are instead only
```
sooheelee: omit `, which must instead be set to zero` as `-ip`'s default is already zero.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
- * <p> The -O argument specifies a filename for the output bins, stored as a Picard interval list. </p>
+ * <h3>Output</h3>
+ *
+ * <ul>
+ *     <li>
+ *         Preprocessed Picard interval-list file.
+ *     </li>
+ * </ul>
+ *
+ * <h3>Examples</h3>
```
sooheelee: --> Usage examples

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
+ *
```
sooheelee: Add "To pad intervals by the default 250 bases and to turn off binning:"

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
+ * <pre>
+ *     gatk PreprocessIntervals \
+ *          -R reference.fa \
+ *          -L intervals.interval_list \
+ *          --bin-length 0 \
+ *          --padding 250 \
```
sooheelee: remove `--padding 250 \`

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
+ *          -O preprocessed_intervals.interval_list
+ * </pre>
  *
- * <h3>Example</h3>
```
sooheelee: Add "To generate consecutive bins of 20k bases from the reference:"

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
  * <pre>
- * gatk-launch --javaOptions "-Xmx4g" PreprocessIntervals \
- *   -R ref_fasta.fa \
- *   -L intervals.list \
- *   -BL 10000 \
- *   -P 500 \
- *   -O preprocessed-intervals.interval_list
+ *     gatk PreprocessIntervals \
+ *          -R reference.fa \
+ *          --bin-length 1000 \
```
sooheelee: If the 1000 length is arbitrary, then may I suggest instead using 20k bins? It is what production is using for joint calling.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
+ *          --padding 0 \
```
sooheelee: Does it make a difference to turn off padding in this instance? I should think the bins would take precedence. If so, then let's remove the `--padding 0 \`.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
```diff
@@ -3,73 +3,108 @@
...
-        summary = "Prepare intervals for binning, then create bins that cover them. "
-                + "The intervals are first padded, the overlapping ones are merged "
-                + "and bins are created that cover the intervals. "
-                + "If the bin length is incommensurate with the length of an interval, "
-                + "the last bin of that interval will be of a different length. "
-                + "The length of the padding regions at both sides of the intervals (-P) and "
-                + "the length of the bins (-BL) can be specified. ",
-        oneLineSummary = "Prepare intervals for binning then create bins that cover them.",
+        summary = "Prepare bins for coverage collection.",
+        oneLineSummary = "Prepare bins for coverage collection.",
```
sooheelee: I'd prefer "Prepares bins for ...". Please omit the period at the end.

src/main/python/org/broadinstitute/hellbender/tools/gcnv/README.txt
```diff
@@ -0,0 +1,43 @@
...
+
+    `gcnvkernel.tasks`: Inference task executors for different models and their use cases.
+        This module additionally provides general-purpose templates for setting up arbitrary
+        ADVI inference tasks for models with mixed discrete and continuous latent variables.
+
+    `gcnvkernel.utils`: Contains utility classes for CLI tools, as well as an implementation
+        of recursive least squares (RLS).
+
+Even though `gcnvkernel` can be used as a standalone tool, its intended use case is to be
+privately called by GATK CLI tools. As such, the outputs of `gcnvkernel` are considered
```
sooheelee: I think @vdauwera and @droazen need to know about this gcnvkernal tool if they do not already. Currently, it is not something we document and having this readme in the repo will ultimately lead to questions from the users. I mean, I have questions--namely, what is this and how do I use it?

sooheelee:
Back to you @samuelklee.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
samuelklee: This bin size is not really related to that bin size.  This bin size determines the resolution of CNV breakpoints.  The theoretical limit depends on coverage depth and the insert-size distribution, but typically we can expect that bin sizes on the order of the read length would give reasonable results.  For example, we've done several WGS runs where the bin size is 250bp, which is probably as low as we'd want to go for typical data.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PreprocessIntervals.java
samuelklee: You could imagine preprocessing WES targets that you'd want to both pad and break up into smaller bins (if your targets were very long, this might buy you a bit of resolution), which is why we allow for both arguments to be specified simultaneously.

src/main/python/org/broadinstitute/hellbender/tools/gcnv/README.txt
samuelklee: I'm pretty sure @droazen is aware of it, but @vdauwera might not be.  `gcnvkernel` is the name of the python package that provides the machine-learning backend for the new gCNV tools (DetermineGermlineContigPloidy and GermlineCNVCaller).  It relies heavily on the pymc3 package (https://github.com/pymc-devs/pymc3) to specify and perform inference for the gCNV models.  DetermineGermlineContigPloidy and GermlineCNVCaller are essentially Java wrappers that simply pass arguments to the respective python scripts in the gcnvkernel package; one might imagine that advanced users could directly use the python scripts, so gcnvkernel is essentially standalone.  This is not fundamentally different from how some tools in the GATK are simply wrappers of various R scripts, it's just that the thing being wrapped is sufficiently complex to warrant the development of an entire standalone package.

sooheelee: Would I invoke the gcnvkernal tool much like we do CatVariants?

- CatVariants doc: https://software.broadinstitute.org/gatk/documentation/tooldocs/current/org_broadinstitute_gatk_tools_CatVariants.php

If so, gcnvkernal will need its own tool doc, like CatVariants.

samuelklee: No, a typical user would simply use the DetermineGermlineContigPloidy and GermlineCNVCaller tools, which are already documented.

samuelklee:
@MartonKN This is going in soon unless you have any comments.

MartonKN:
Thanks, I have no more comments.

PR #4031, opened 2018-01-02
HaplotypeCaller: major performance improvements and fixes

droazen:
This set of optimizations brings the GATK4 HaplotypeCaller performance into line
with GATK3.x performance.

Note that HaplotypeCallerSpark is not touched by this PR (that is for a future PR).

Summary of changes:

* AssemblyRegionWalker: query all intervals on each contig simultaneously, rather than individually
* GATKRead: Cache adaptor boundary, soft start/end, and cigar length
* GATKRead: add getBasesNoCopy() / getBaseQualitiesNoCopy()
* ReadPileup: speed up stratified constructor
* LIBS.lazyLoadNextAlignmentContext(): don't keep pileup elements unnecessarily separated by sample during pileup creation
* Restore faster GATK3 version of ReferenceConfidenceModel.sumMismatchingQualities()
* RefVsAnyResult: nest within ReferenceConfidenceModel, and allow direct field access
* Remove redundant getBases() call in ReadThreadingGraph
* Fix BaseGraph Utils.validateArg() call
* ReadPileup: replace Collections.unmodifiableList(pileupElements).iterator() with direct return of an iterator that forbids removal
* Kill expensive bounds checking in GATKRead getBase()/getBaseQuality()/getCigarElement()
* Kill nonNull checks in PileupElement
* Kill expensive PileupElement and ReadPileup arg validation
* GATKRead adapter: clear cached values upon mutation


droazen:
@lbergelson please review

droazen:
@davidbenjamin take note: the changes to `AssemblyRegionWalker` here have a major effect on `HaplotypeCaller` performance when run with a large (eg., whole-exome) interval list. I would expect there to also be a dramatic effect on `Mutect2` performance when run with such an interval list -- would you have time to do a quick check on this?

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
```diff
@@ -207,6 +252,14 @@ default boolean isEmpty() {
      */
     byte[] getBases();
 
+    /**
+     * @return The read sequence as ASCII bytes ACGTN=, or an empty byte[] if no sequence is present.
+     *
+     * This method DOES NOT make a defensive copy of the bases array before returning it, so modifying the
+     * returned array WILL alter the bases in the read. CALLER BEWARE!
+     */
+    byte[] getBasesNoCopy();
```
lbergelson: I would strongly consider renaming these to getBasesUnsafe() and changing the description to make it clear that editing the bases only MAY alter the bases in the read, not WILL alter the bases in the read, since half of the implementations do not.  Then I would provide a default implementation that's just getBases()

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
```diff
@@ -175,6 +175,21 @@ public int getUnclippedEnd() {
...
+    }
+
+    @Override
+    public int getSoftEnd() {
+        return ReadUtils.getSoftEnd(this);
+    }
+
+    @Override
+    public int getAdaptorBoundary() {
+        return ReadUtils.getAdaptorBoundary(this);
```
lbergelson: These should be moved up as defaults.

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
```diff
@@ -291,6 +312,12 @@ public void setBases( final byte[] bases ) {
         return convertedBaseQualities;
     }
 
+    @Override
+    public byte[] getBaseQualitiesNoCopy() {
+        // Not possible to have a no-copy implementation for Google reads
+        return getBaseQualities();
```
lbergelson: these should either throw as not implemented or the description of the semantics needs to change in GATKRead

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
```diff
@@ -124,6 +124,51 @@ default boolean isEmpty() {
...
+     *
+     *   |---------------------->   *
+     *     <----------------------|
+     *
+     *   in these cases the adaptor boundary is at the start of the read plus the inferred insert size (plus one)
+     *
+     * @return the reference coordinate for the adaptor boundary (effectively the first base IN the adaptor, closest to the read).
+     * CANNOT_COMPUTE_ADAPTOR_BOUNDARY if the read is unmapped or the mate is mapped to another contig.
+     */
+    int getAdaptorBoundary();
```
lbergelson: I think this may be problematic to include on this interface because I think it relies on the our definition of fragment size which is neither universally accepted or universally used.

src/main/java/org/broadinstitute/hellbender/utils/read/SAMRecordToGATKReadAdapter.java
```diff
@@ -28,6 +28,18 @@
 
     private final SAMRecord samRecord;
 
+    private transient Integer cachedSoftStart = null;
+    private transient Integer cachedSoftEnd = null;
+    private transient Integer cachedAdaptorBoundary = null;
+    private transient Integer cachedCigarLength = null;
+
+    private void clearCachedValues() {
```
lbergelson: I'm a bit worried that repeated calls to clearCachedValues() are going to introduce performance regressions in other places.  I'd sort of lean towards putting the caching in an additional wrapper layer that's only used in haplotype caller, although I understand the concern about that introducing more performance issues.

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
lbergelson: It's also super specific to our sequencing technology and just bizarre to have on GATKRead.  I don't like it.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionIterator.java
```diff
@@ -10,17 +10,19 @@
...
 import org.broadinstitute.hellbender.utils.read.GATKRead;
 import org.broadinstitute.hellbender.utils.read.ReadUtils;
 
 import java.util.*;
 
 /**
- * Given a {@link Shard} of {@link GATKRead}, iterates over each {@link AssemblyRegion} within that shard, using the
- * provided {@link AssemblyRegionEvaluator} to determine the boundaries between assembly regions.
+ * Given a {@link MultiIntervalShard} of {@link GATKRead}, iterates over each {@link AssemblyRegion} within
+ * that shard, using theprovided {@link AssemblyRegionEvaluator} to determine the boundaries between assembly
```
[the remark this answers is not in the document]
droazen: Fixed

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionIterator.java
```diff
@@ -137,11 +142,6 @@ private AssemblyRegion loadNextAssemblyRegion() {
         while ( locusIterator.hasNext() && nextRegion == null ) {
             final AlignmentContext pileup = locusIterator.next();
 
-            // Ignore sites that are not contained within this shard's interval
```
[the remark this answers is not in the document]
droazen: We haven't needed it for a while. The `IntervalAlignmentContextIterator` will only give us pileups that overlap our intervals.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -181,24 +149,20 @@
     protected final void onStartup() {
         super.onStartup();
 
-        if ( readShardSize <= 0 && readShardSize != NO_INTERVAL_SHARDING ) {
-            throw new CommandLineException.BadArgumentValue("read shard size must be > 0 or " + NO_INTERVAL_SHARDING + " for no sharding");
-        }
-
-        if ( readShardPadding < 0 ) {
-            throw new CommandLineException.BadArgumentValue("read shard padding must be >= 0");
+        if ( minAssemblyRegionSize <= 0 || maxAssemblyRegionSize <= 0 ) {
```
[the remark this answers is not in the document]
droazen: I leave it to a future PR to migrate the codebase to the barclay `minValue` functionality.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/ReferenceConfidenceModel.java
```diff
@@ -89,6 +91,66 @@
...
+         *                           genotypes given the number of alleles (always 2), ploidy (arbitrary) less the genotyping
+         *                           model non-sense genotype count if applies.
+         * @throws IllegalArgumentException if {@code likelihoodCapacity} is negative.
+         */
+        public RefVsAnyResult(final int likelihoodCapacity) {
+            ParamUtils.isPositiveOrZero(likelihoodCapacity, "likelihood capacity is negative");
+            genotypeLikelihoods = new double[likelihoodCapacity];
+        }
+
+        double[] getGenotypeLikelihoods() {
```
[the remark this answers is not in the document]
droazen: Comment added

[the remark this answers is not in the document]
droazen: Method deleted

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/ReferenceConfidenceModel.java
```diff
@@ -408,32 +470,35 @@ VariantContext getOverlappingVariantContext(final Locatable curPos, final Collec
...
+     * @param readQuals non-null quals of the read
      * @param readStart the starting position of the read (i.e., that aligns it to a position in the reference)
      * @param refBases the reference bases
      * @param refStart the offset into refBases that aligns to the readStart position in readBases
      * @param maxSum if the sum goes over this value, return immediately
      * @return the sum of quality scores for readBases that mismatch their corresponding ref bases
      */
     @VisibleForTesting
-    int sumMismatchingQualities(final GATKRead read,
+    int sumMismatchingQualities(final byte[] readBases,
```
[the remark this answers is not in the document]
droazen: We'd have to re-profile to make that change, since this bit of code is extremely performance-sensitive to the smallest change -- not worth doing now.

src/main/java/org/broadinstitute/hellbender/utils/IntervalUtils.java
```diff
@@ -956,6 +956,51 @@ public static long intervalSize(final List<GenomeLoc> locs) {
...
+    /**
+     * Pads the provided intervals by the specified amount, sorts the resulting intervals, and merges intervals
+     * that are adjacent/overlapping after padding.
+     *
+     * @param intervals intervals to pad
+     * @param basePairs number of bases of padding to add to each side of each interval
+     * @param dictionary sequence dictionary used to restrict padded intervals to the bounds of their contig
+     * @return the provided intervals padded by the specified amount, sorted, with adjacent/overlapping intervals merged
+     */
+    public static List<SimpleInterval> getIntervalsWithFlanks(final List<SimpleInterval> intervals, final int basePairs, final SAMSequenceDictionary dictionary) {
```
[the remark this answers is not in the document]
droazen: It's an overload to an existing method of the same name.

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -105,7 +109,6 @@ PileupElement getElementForRead(final GATKRead read) {
      * Helper routine for converting reads and a single offset to a PileupElement list.
      */
     private static List<PileupElement> readsOffsetsToPileup(final List<GATKRead> reads, final int offset) {
-        Utils.nonNull(reads, "Illegal null read list");
         return reads.stream().map(r -> PileupElement.createPileupForReadAndOffset(r, offset)).collect(Collectors.toList());
```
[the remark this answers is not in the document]
droazen: In a future profiling pass, perhaps :)

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -201,7 +203,26 @@ public ReadPileup getPileupForSample(final String sample, final SAMFileHeader he
      */
     @Override
     public Iterator<PileupElement> iterator() {
-        return Collections.unmodifiableList(pileupElements).iterator();
+        // Profiling has determined that returning a custom unmodifiable iterator is faster than
+        // Collections.unmodifiableList(pileupElements).iterator()
+        return new Iterator<PileupElement>() {
```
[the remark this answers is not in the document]
droazen: Would require re-profiling, but yes we probably could

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
droazen: Well, it's called `GATKRead` after all, not `Read` :) The benefits of caching this value per read are so huge that I think we have to live with the awkwardness.

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
```diff
@@ -124,6 +124,51 @@ default boolean isEmpty() {
...
     int getUnclippedEnd();
 
+    /**
+     * Calculates the reference coordinate for the beginning of the read taking into account soft clips but not hard clips.
+     *
+     * Note: {@link #getUnclippedStart} adds soft and hard clips, this method only adds soft clips.
+     *
+     * @return the unclipped start of the read taking soft clips (but not hard clips) into account
+     */
+    int getSoftStart();
```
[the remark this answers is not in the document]
droazen: Added a default implementation

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
```diff
@@ -124,6 +124,51 @@ default boolean isEmpty() {
...
+
+    /**
+     * Calculates the reference coordinate for the end of the read taking into account soft clips but not hard clips.
+     *
+     * Note: {@link #getUnclippedEnd} adds soft and hard clips, this method only adds soft clips.
+     *
+     * @return the unclipped end of the read taking soft clips (but not hard clips) into account
+     */
+    int getSoftEnd();
```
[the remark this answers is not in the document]
droazen: Added

src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java
droazen: I don't like the name `getBasesUnsafe()`, since it scare people without explaining why they should be scared. I think it's better to have the name explain how it differs from `getBases()`.

I will change the description, however, to indicate that implementations will avoid a copy only if it's possible to do so.

droazen: Also added default implementations.

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
droazen: Done

src/main/java/org/broadinstitute/hellbender/utils/read/GoogleGenomicsReadToGATKReadAdapter.java
droazen: Description in `GATKRead` changed

src/main/java/org/broadinstitute/hellbender/utils/read/SAMRecordToGATKReadAdapter.java
droazen: Well, setters are not generally called repeatedly in hotspots, while getters are. So I think it's ok to slightly penalize the setters in order to greatly speed up the getters in critical regions. Putting the caching in a separate layer could harm the performance gains I've seen with this approach, so I'd rather not.

droazen:
Comments addressed -- will merge once tests pass.

src/main/java/org/broadinstitute/hellbender/utils/locusiterator/LocusIteratorByState.java
```diff
@@ -315,12 +315,16 @@ private void lazyLoadNextAlignmentContext() {
             readStates.collectPendingReads();
 
             final Locatable location = getLocation();
-            final Map<String, ReadPileup> fullPileupPerSample = new LinkedHashMap<>();
+
+            // We don't need to keep the pileup elements separated by sample within this method,
+            // since they are just going to get combined into one monolithic pileup anyway
+            // when we construct the final ReadPileup below. This optimization speeds up the
+            // HaplotypeCaller by quite a bit!
```
magicDGS: Maybe this will be important for #2321, which relies on the separation of the stratified pileup here. Maybe a new option for the LIBS will be good for improving performance...

src/main/java/org/broadinstitute/hellbender/utils/pileup/ReadPileup.java
```diff
@@ -48,7 +46,15 @@ public ReadPileup(final Locatable loc, final List<PileupElement> pileup) {
      * Note: the current implementation of ReadPileup does not efficiently retrieve the stratified pileup
      */
     public ReadPileup(final Locatable loc, final Map<String, ReadPileup> stratifiedPileup) {
-        this(loc, stratifiedPileup.values().stream().flatMap(ReadPileup::getElementStream).collect(Collectors.toList()));
+        // NOTE: this bit of code is a potential performance hotspot for the HaplotypeCaller.
+        // This straightforward loop outperforms the equivalent streaming expression by over 2x.
```
magicDGS: Maybe this will be improved too with #2321, but we should check if there is a regression there.

PR #4035, opened 2018-01-03
print reads doc

takutosato:


src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
```diff
@@ -13,7 +13,33 @@
...
+ * <p>
+ *     This tool copies reads from a SAM/BAM/CRAM file to a SAM file. If you are interested in printing reads in
+ *     certain intervals only, use the -L argument.
+ *
+ * </p>
+ *
+ * <h3> Input </h3>
+ * <ul>
+ *     <li> SAM/BAM/CRAM file to be copied to a SAM file </li>
+ *     <li> (optional) interval file with the -L argument </li>
```
ldgauthier: I wouldn't call an interval file an input.  Yes, it's a file the user supplies, but it's more of an argument.

src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
```diff
@@ -13,7 +13,33 @@
 import org.broadinstitute.hellbender.utils.read.SAMFileGATKReadWriter;
 
 /**
- * Print reads from a SAM/BAM/CRAM file.
+ * Copy reads in a SAM/BAM/CRAM file verbatim to a SAM file
+ *
+ * <p>
+ *     This tool copies reads from a SAM/BAM/CRAM file to a SAM file. If you are interested in printing reads in
+ *     certain intervals only, use the -L argument.
```
ldgauthier: @sooheelee Is it worth mentioning that this gets used with BQSR or that it can be used with read filters?  Otherwise it sounds like a pretty pointless tool.

cmnbroad: @ldgauthier I think ApplyBQSR is intended to be the replacement for using PrintReads for BQSR. So I'm not sure this has much real analytical value, other than filtering (and maybe format conversion, and index generation, but since it has a default filter WellFormed its not necessarily full fidelity by default).

ldgauthier: I forgot about ApplyBQSR.  Fair point.

src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
```diff
@@ -13,7 +13,33 @@
 import org.broadinstitute.hellbender.utils.read.SAMFileGATKReadWriter;
 
 /**
- * Print reads from a SAM/BAM/CRAM file.
+ * Copy reads in a SAM/BAM/CRAM file verbatim to a SAM file
```
lbergelson: This makes it sound like it only outputs sam files.  Shouldn't it say to a SAM/BAM/CRAM file?

src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
```diff
@@ -13,7 +13,33 @@
...
+ *
+ * <p>
+ *     This tool copies reads from a SAM/BAM/CRAM file to a SAM file. If you are interested in printing reads in
```
lbergelson: It might be useful to explain what this tool is useful for, I.e. selecting parts intervals from a bam or applying filters to a bam in order to produce a new bam that doesn't have certain classes of reads.

src/main/java/org/broadinstitute/hellbender/tools/PrintReads.java
```diff
@@ -13,7 +13,33 @@
...
+ * SAM file with reads copied from the source
+ *
+ *
+ * <h3>Examples</h3>
+ * <pre>
+ *  gatk PrintReads \
+ *  -R reference.fasta \
+ *  -I input.bam \
+ *  -L 20 \
+ *  -O chr20.sam
```
lbergelson: I'd make the example output a bam probably.

sooheelee:
I'm going to go ahead and make changes directly to the branch @takutosato and try to incorporate @lbergelson's suggestions to get this going.

sooheelee:
Once the tests come through, since this only changes the doc portion, I will merge this.

sooheelee:
Test keep failing despite restart. Help? @lbergelson?

lbergelson:
@sooheelee Don't merge this yet.  Something funny has happened to the history.

sooheelee:
Rebase fail.

lbergelson:
I don't think you need to close it, we can probably fix it.

sooheelee:
I just made a new one at https://github.com/broadinstitute/gatk/pull/4084

PR #4037, opened 2018-01-03
Getting rid of '@' from lines when generating the oncotated outputs.

LeeTL1220:
This WDL change is needed for two reasons:
- keep formats the same as previous CNV oncotator outputs
- remove any sequence dictionary that might be prepended to the incoming seg file, which oncotator cannot handle.


samuelklee:
:+1:

PR #4045, opened 2018-01-04
Adding boot disk for CNV WDL.

LeeTL1220:
- Error that was missed in testing.


samuelklee:
:+1:  Is this an attribute that we should be setting in other tasks?  I only see one other use of it in the M2 WDL.

LeeTL1220:
Nope, just oncotator.  Not only that, only the CNV oncotator really needs it, since it uses a big docker image with some datasources baked-in.  We could probably remove it from the M2 WDL, since that uses the typical oncotator docker image and downloads the datasources on the fly.

PR #4046, opened 2018-01-04
Changed default values for ModelSegments segmentation parameters.

### LINKED ISSUE #4032: Decide on similar-segment merging behavior and preliminary values for relevant parameters for ModelSegments release.
Particularly the smoothing thresholds for both CR and AF and the kernel variance for AF.

samuelklee:
Placeholders for now.  We can tweak the actual values once @LeeTL1220 checks effect on validation.

Closes #4032.

### DIFF
--- a/scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl
+++ b/scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl
@@ -387,9 +387,9 @@ task ModelSegments {
             --minimum-total-allele-count ${default="30" min_total_allele_count} \
             --genotyping-homozygous-log-ratio-threshold ${default="-10.0" genotyping_homozygous_log_ratio_threshold} \
             --genotyping-base-error-rate ${default="0.05" genotyping_base_error_rate} \
-            --maximum-number-of-segments-per-chromosome ${default="500" max_num_segments_per_chromosome} \
+            --maximum-number-of-segments-per-chromosome ${default="1000" max_num_segments_per_chromosome} \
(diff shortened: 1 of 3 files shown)


LeeTL1220:
I have a run going now

samuelklee:
Remember to sync defaults in the WDL task as well (which I didn't do yet).  I think maximum-number-of-segments-per-chromosome is also out of sync, might be a few others as well.

samuelklee:
@LeeTL1220 Latest commit includes the rollback.  Looking at it again, I initially described the change to you incorrectly.  I thought it was "similar CR || similar AF -> merge" to "similar CR && similar AF -> merge", but that's not actually the case; it's instead "similar according to credible interval 1 || similar according to credible interval 2 -> merge" to "similar according to credible interval 1 && similar according to credible interval 2 -> merge".  Probably the `&&` behavior is way too conservative, so I think rolling back to the `||` behavior would be fine for release.  Let's double check with the validation just to be sure.

samuelklee:
@LeeTL1220 OK, see the sl_change_model_segments_defaults_rebased branch.  For validation, I'd say that sweeping the following should suffice:

Array[Float] kernel_variance_allele_fractions = [0.025, 0.05, 0.25]
Array[Float] smoothing_thresholds_allele_fraction = [2.0, 10.0, 50.0]
Array[Float] smoothing_thresholds_copy_ratio = [2.0, 10.0, 50.0]

samuelklee:
Merging this according to discussion with @LeeTL1220.  We still need to investigate some possible regressions in the CRSP validation, but we should be good for prerelease on the CNV somatic side.  There is one last branch (#4061) on the germline side, but I think we didn't plan on having gCNV ready for prerelease.

PR #4047, opened 2018-01-05
DON'T MERGE - Allele specific annotations support in GenomicsDB

kgururaj:


kgururaj:
@jamesemery would it be possible to run a few checks to see if the allele specific annotations output of GenomicsDB matches your expectation? I have used the GATK Travis inputs and things look OK.

kgururaj:
Some clarification questions about the field AS_RAW_MQ. In the examples provided, the header line corresponding to this field is

    ##INFO=<ID=AS_RAW_MQ,Number=A,Type=Float,Description="Allele-specfic raw data for RMS Mapping Quality">

The data lines contain entries similar to `AS_RAW_MQ=123769.00|3600.00|46800.00|0.00`

Is the AS_RAW_MQ field simply a vector of float?
- If yes
  - Why is the '|' used as the delimiter? Why not simply use ','?
  - Based on the entries, shouldn't the Number descriptor in the header line be 'R' instead of 'A'
- If no, i.e. AS_RAW_MQ can be a vector of vector of float (entries such as  `AS_RAW_MQ=1,2|3,4,5|6|0.00` where `[1, 2]` corresponds to allele 0 (reference allele) etc.)
    - Shouldn't the Type descriptor in the header line be 'String'  instead of 'Float'? (similar to the AS_RAW_\*RankSum fields)

ldgauthier:
Yes, AS_RAW_MQ is effectively a vector of floats.  I used the pipe delimiter for consistency with the other annotations, which may contains lists for each allele and thus needed two delimiters.

That header doesn't seem to agree with the example.  For most of the raw annotations, I treated them as strings and then did the parsing in the GATK code since the VCF spec (and thus htsjdk) doesn't support lists of lists.

kgururaj:
Thanks Laura, I picked the header and data from files used in GATK tests:
https://github.com/broadinstitute/gatk/blob/master/src/test/resources/org/broadinstitute/hellbender/tools/walkers/CombineGVCFs/NA12878.AS.chr20snippet.g.vcf

https://github.com/broadinstitute/gatk/blob/master/src/test/resources/org/broadinstitute/hellbender/tools/walkers/CombineGVCFs/testAlleleSpecificAnnotations.vcf

However, a different file in the GATK repo specifies AS_RAW_MQ as a String:
https://github.com/broadinstitute/gatk/blob/master/src/test/resources/org/broadinstitute/hellbender/tools/haplotypecaller/expected.testGVCFMode.gatk4.alleleSpecific.g.vcf

Currently, GenomicsDB can handle both cases (with a slight hack in the internal code), but wanted to check if I'm missing anything.

ldgauthier:
I may have given James some dubious test data for CombineGVCFs.  The header line in GATKVCFHeaderLines.java describes AS_RAW_MQ as having one string value.  That's the version I would like to move forward with.  I can fix the CombineGVCFs test data if that's causing a problem.

jamesemery:
Yea, I believe the header mismatching has to do with the canonical header lines in the gatk being updated in the middle of this work. htsjdk doesn't discriminate between bad header line count fields so there are a number of sloppy header lines like this. I will update my genomicsDB branch to perform a proper check of the allele specific annotations based on the new version now.

kgururaj:
I also switched the format used to transfer data between the Java and C++ modules of GenomicsDB to VCF (from BCF2) since the String fields are causing some issues in htsjdk. I filed an issue here: https://github.com/samtools/htsjdk/issues/1063

kgururaj:
I forgot to test with the branch after Patrick's changes - using the pre-generated merged VCF header caused issues with BCF2Codec. I have now implemented a workaround.

jamesemery:
Right, this seems to have fixed the scrambling annotations problem. The primary Issue I see now is that you are still dropping empty/NaN annotations. If we are concerned with copying gatk3 behavior it seems necessary to at least preserve the fact that those annotations were present but uninformative.

(GenomicsDB): `[VC null @ 20:10433312-10433319 Q. of type=MIXED alleles=[CAAAAAAA*, <NON_REF>, C, CAAAAAA] attr={AS_RAW_MQ=7.659800e+04|0.000000e+00|1.682000e+03|0.000000e+00, AS_SB_TABLE=19,20|0,0|1,1|0,0, BaseQRankSum=1.5880000591278076, DP=80, ExcessHet=3.0102999210357666, MQRankSum=-1.2589999437332153, RAW_MQ=179123.0, ReadPosRankSum=0.3330000042915344}` 

(CmbineGVCFs)L [VC Unknown @ 20:10433312-10433319 Q. of type=MIXED alleles=[CAAAAAAA*, <NON_REF>, C, CAAAAAA] attr={AS_RAW_BaseQRankSum=|||, AS_RAW_MQ=76598.00|0.00|1682.00|0.00, AS_RAW_MQRankSum=|||, AS_RAW_ReadPosRankSum=|||, AS_SB_TABLE=[19, 20|0, 0|1, 1|0, 0], BaseQRankSum=1.59, DP=80, ExcessHet=3.01, MQRankSum=-1.259e+00, RAW_MQ=179123.00, ReadPosRankSum=0.333}

It seems like it is preserving (correctly it seems) raw annotations in the case where there is data to speak of. Would it be possible to emit the empty annotations if they were present in the original merged variants?

kgururaj:
Prints out empty fields now

jamesemery:
Insofar as its matching all of the tests that currently exist for allele specific annotations it appears to be working this time around (with a few changes to the vcf test suite to handle the the precision issues). I'm thinking this might be good.

jamesemery:
It has been decided to merge this branch in after the release so it can be more substantially tested. Since it is being pushed off, we should take a look at the precision issue for annotations. Would it be possible to limit the precision of the output annotations to 3 decimal points? It seems to be outputting a lot of annotations that end up being excessively long.

kgururaj:
That's disappointing.

How would you like to print FP values?
- Fixed point or scientific
- Should the decimal point always be printed?
- Should the total number of digits (integer+fraction) be constant (overall precision)? Or should a constant number of fraction digits be printed?

ldgauthier:
I'm a big fan of the htsjdk precision code: https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/vcf/VCFEncoder.java  

That being said, fixed point is fine.  We don't really need the precision of scientific for the standard GATK annotations.  I would like the decimal point to always be present (somewhere I made an assumption about that in order to avoid a bunch of `instanceof` calls) and total number of digits being constant is good.

kgururaj:
Closing since @psfoley has opened https://github.com/broadinstitute/gatk/pull/4261

PR #4052, opened 2018-01-05
Fixed a bug with CombineGVCFs making invalid intervals for wole genomes.

jamesemery:


src/test/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFsIntegrationTest.java
```diff
@@ -365,4 +365,34 @@ public void testOneHasDeletionAndTwoHasRefBlock() throws Exception {
...
+    @Test
+    public void testStartChromosome() throws Exception {
+        final File output = createTempFile("genotypegvcf", ".vcf");
+
+        final ArgumentsBuilder args = new ArgumentsBuilder();
+        args.addReference(new File(b37_reference_20_21))
+                .addOutput(output);
+        args.addVCF(getTestFile("gvcfExample1.fullchrom.2021.vcf"));
+        args.addVCF(getTestFile("gvcfExample2.vcf"));
+        args.add("--" + CombineGVCFs.IGNORE_VARIANTS_THAT_START_OUTSIDE_INTERVAL );
```
ldgauthier: What's this for?  I don't see any provided intervals.

jamesemery: In reality? Nothing it was just copied from above. I will remove it. The important piece of the test is that I'm checking a vcf file containing variants that start at position 1 of the chromosome, `CombineGVCFs.IGNORE_VARIANTS_THAT_START_OUTSIDE_INTERVAL` only applies in the case where there is an interval provided, I will remove it

jamesemery: Though perhaps it should throw if there is no interval provided?

ldgauthier:
I'll leave the exception up to you.  I don't have a preference one way or the other. Otherwise looks good to me.

ldgauthier:
:+1:

PR #4067, opened 2018-01-07
Add Python package archive and Conda env .yml to distribution bundle.

cmnbroad:


build.gradle
```diff
@@ -459,6 +459,8 @@ task bundle(type: Zip) {
     from("gatk")
     from("README.md")
     from("build/docs/tabCompletion/gatk-completion.sh")
+    from("scripts/gatkcondaenv.yml")
+    from("$buildDir/gatkPythonPackageArchive.zip")
```
droazen: By how much does this increase the size of the zip? I ask because we post this zip on github as the standard GATK binary download.

cmnbroad: @droazen Its insignificant. The Python archive zip file is 74k. 360,288,393 to 360,358,438.

PR #4070, opened 2018-01-07
Upgrade to Barclay 2.0.0 and Picard 2.17.2.

cmnbroad:
There are 4 separate commits:
- Upgrade to Barclay 2.0.0 and Picard 2.17.2.
- Changes for CommandLinePluginDescriptor updates (required for the Barclay upgrade)
- Updates for Experimental tag (dependent on Barclay upgrade).
- Remove placeholder and obsolete program groups - Part 1 (dependent on Picard upgrade).There are still 3 obsolete program groups (ReadProgramGroup, VariantProgramGroup, and SparkProgramGroup) who's tools need to be redistributed to the new program groups. But thats can be a separate PR since it will be a lot of files.


src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -371,20 +399,8 @@ public final CountingReadFilter getMergedCountingReadFilter(final SAMFileHeader
...
-                                .map(e -> e.getValue())
-                                .collect(Collectors.toList());
-
-        // now add in any additional filters enabled on the command line (preserving order)
-        final List<ReadFilter> clFilters = getAllInstances();
+        final List<ReadFilter> finalFilters = new ArrayList<>();
+        final List<ReadFilter> clFilters = getResolvedInstances();
         if (clFilters != null) {
             clFilters.stream()
                     .filter(f -> !finalFilters.contains(f)) // remove redundant filters
```
droazen: Is `finalFilters` redundant here now? Seems like you always just end up with `finalFilters` equal to `clFilters` after this patch.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
```diff
@@ -166,16 +163,41 @@ public boolean isDependentArgumentAllowed(final Class<?> predecessorClass) {
     }
 
     /**
-     * Pass back the list of ReadFilter instances that were actually seen on the command line in the same
-     * order they were specified. Its possible for this to return a filter that was originally included
-     * in the list of tool defaults only in the case where the user also specifies it on the command line.
+     * @return an ordered list ReadFilter instances that result from resolving all command line
```
droazen: `list` -> `list of`

src/main/java/org/broadinstitute/hellbender/tools/ConvertHeaderlessHadoopBamShardToBam.java
```diff
@@ -27,7 +27,7 @@
 @CommandLineProgramProperties(
         summary = "Convert a headerless hadoop bam shard into a readable bam",
         oneLineSummary = "Convert a headerless hadoop bam shard into a readable bam",
-        programGroup = TestSparkProgramGroup.class
+        programGroup = OtherProgramGroup.class
```
droazen: Why is this in `OtherProgramGroup`? Did comms just not understand what this tool does? It is a diagnostic/debugging tool for allowing intermediate reads data on Spark to be read using conventional programs like `samtools`

droazen: Should also probably be marked as `Experimental`

src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
```diff
@@ -17,13 +17,13 @@
 import java.util.stream.Collectors;
 
 @DocumentedFeature
+@ExperimentalFeature
 @CommandLineProgramProperties(
-        oneLineSummary = "(EXPERIMENTAL) Emit a single sample name from the bam header into an output file.",
```
droazen: Did you go through the codebase and replace all instances of hardcoded `(EXPERIMENTAL)` text with the new annotation? I only see this one, and I know there were more. If you haven't, please do so in this PR.

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
```diff
@@ -39,7 +37,7 @@
 @CommandLineProgramProperties(
         summary = "Creates an index for a feature file, e.g. VCF or BED file.",
         oneLineSummary = "Creates an index for a feature file, e.g. VCF or BED file.",
-        programGroup = VCFManipulationProgramGroup.class
+        programGroup = VariantManipulationProgramGroup.class
```
droazen: This tool is also mis-categorized I think. The files that this tool can index do not necessarily contain variants. It can index any file format that is supported by Tribble, including things like BED files.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/FilterMutectCalls.java
```diff
@@ -11,7 +11,8 @@
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
-import org.broadinstitute.hellbender.cmdline.programgroups.VariantFilteringProgramGroup;
+import picard.cmdline.programgroups.VariantFilteringProgramGroup;
+import picard.cmdline.programgroups.VariantFilteringProgramGroup;
```
droazen: Duplicated import

src/main/java/org/broadinstitute/hellbender/tools/ConvertHeaderlessHadoopBamShardToBam.java
sooheelee: The doc portion is cryptic and no, it's not clear what this tool does from the current documentation. I asked that the description be updated a few days ago.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/GATKReadFilterPluginDescriptor.java
cmnbroad: You're right - redundancies are already removed, so this is all for nothing and can be simplified.

src/main/java/org/broadinstitute/hellbender/tools/ConvertHeaderlessHadoopBamShardToBam.java
cmnbroad: I'll mark as experimental. The only other group that makes sense is reads, unless we invent something new.  I'll move to the reads group for now.

src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
cmnbroad: No, I only did the ones I was already touching. I'll search for any others and repair.

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
cmnbroad: True, though "Other" is the only logical choice given the ones we have. @sooheelee this needs a category ?

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/FilterMutectCalls.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/tools/ConvertHeaderlessHadoopBamShardToBam.java
sooheelee: I don't like this move. We can discuss after the release. My quip is that then CreateHadoopBamSplittingIndex needs to move with ConvertHeaderlessHadoopBamShardToBam.

src/main/java/org/broadinstitute/hellbender/tools/IndexFeatureFile.java
sooheelee: Other is the only one that makes sense.

PR #4078, opened 2018-01-08
deleting unused SparkToggleCommandLineProgram

lbergelson:
I noticed that this was completely unused and untested


LeeTL1220:
Used to be used.

Used to be tested through an integration test that is no longer, I guess.

Merge away.

@lbergelson

PR #4080, opened 2018-01-08
ReadClipper: when hardclipping results in an empty read, avoid trying to set a negative start position

### LINKED ISSUE #3466: IllegalArgumentException during clipping: contig must be non-null and not equal to *, and start must be >= 1
We're running into an error with GATK 4.0beta3 on some read pairs that appears related to processing clipping on a read. The BAM file itself gets processed fine on GATK 3.8 and the problematic read appears okay on a manual inspection. This is a self-contained reproducible test case that demonstrates the issue:

https://s3.amazonaws.com/chapmanb/testcases/gatk_cigar_error.tar.gz

The error is:
```
java.lang.IllegalArgumentException: contig must be non-null and not equal to *, and start must be >= 1
        at org.broadinstitute.hellbender.utils.read.SAMRecordToGATKReadAdapter.setPosition(SAMRecordToGATKReadAdapter.java:92)
        at org.broadinstitute.hellbender.utils.clipping.ClippingOp.applyHARDCLIP_BASES(ClippingOp.java:381)
        at org.broadinstitute.hellbender.utils.clipping.ClippingOp.apply(ClippingOp.java:73)
        at org.broadinstitute.hellbender.utils.clipping.ReadClipper.clipRead(ReadClipper.java:147)
        at org.broadinstitute.hellbender.utils.clipping.ReadClipper.clipRead(ReadClipper.java:128)
        at org.broadinstitute.hellbender.utils.clipping.ReadClipper.hardClipSoftClippedBases(ReadClipper.java:332)
        at org.broadinstitute.hellbender.utils.clipping.ReadClipper.hardClipSoftClippedBases(ReadClipper.java:335)
        at org.broadinstitute.hellbender.tools.walkers.haplotypecaller.AssemblyBasedCallerUtils.finalizeRegion(AssemblyBasedCallerUtils.java:84)
        at org.broadinstitute.hellbender.tools.walkers.haplotypecaller.AssemblyBasedCallerUtils.assembleReads(AssemblyBasedCallerUtils.java:238)
        at org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCallerEngine.callRegion(HaplotypeCallerEngine.java:480)
        at org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCaller.apply(HaplotypeCaller.java:221)
        at org.broadinstitute.hellbender.engine.AssemblyRegionWalker.processReadShard(AssemblyRegionWalker.java:244)
        at org.broadinstitute.hellbender.engine.AssemblyRegionWalker.traverse(AssemblyRegionWalker.java:217)
        at org.broadinstitute.hellbender.engine.GATKTool.doWork(GATKTool.java:838)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:116)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:173)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:192)
        at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:131)
        at org.broadinstitute.hellbender.Main.mainEntry(Main.java:152)
        at org.broadinstitute.hellbender.Main.main(Main.java:233)
```
and I believe this is the problematic read:
```
SRR826467.159276265     113     MT      18      52      108S62M80S      *       6996084 0       AAAATTGAGCTTTATCCGAAAGCTGCCTCCTGTTTCAGGGTCACAAGGCCTAAGTAGCCCACGCGTTCCCCTTAAATATGCCATCACGATTGACCACAGTTCTCTCGACCTATTAACCTCTCACGGGGGCTCTCCATGCATTTGGTATTTTCGTCTGGGGGGTATGCACGTCATTTTTTTTTTTTTTCAAGCAGAAGACGGCATACGAGATCGTGATGTGACTGGAGTTCAGACGTGTGCTCTTCCGATC ##################################################################################################################################################################################DDDDFFEC>GGGIEJIGGF8GHCDFD?HFD;FGHGGHGFDDIGJJJIIJJJIIGGGIIIHHHFHFFFFFCCC        RG:Z:DRAGEN_RGID        AS:i:52 XS:i:30 NM:i:2
```
Thanks much for looking at this and please let me know I can provide any other information.

droazen:
When hard clipping a read resulted in an empty read, and the read was sufficiently
close to the start of a contig, we could end up trying to set the read to a negative
start position and blowing up. This patch causes us to return an empty read earlier,
preventing us from setting an invalid start.

Resolves the "IllegalArgumentException during clipping: contig must be non-null and not
equal to *, and start must be >= 1" error reported by many users of the GATK4
HaplotypeCaller

Resolves #3466


droazen:
@jamesemery please review

src/main/java/org/broadinstitute/hellbender/utils/clipping/ClippingOp.java
```diff
@@ -365,6 +366,14 @@ private GATKRead applyHARDCLIP_BASES(final GATKRead read, final int start, final
         // the cigar may force a shift left or right (or both) in case we are left with insertions
         // starting or ending the read after applying the hard clip on start/stop.
         final int newLength = read.getLength() - (stop - start + 1) - cigarShift.shiftFromStart - cigarShift.shiftFromEnd;
+
+        // If the new read is going to be empty, return an empty read now. This avoids initializing the new
+        // read with invalid values below in certain cases (such as a negative alignment start).
+        // See https://github.com/broadinstitute/gatk/issues/3466
+        if ( newLength == 0 ) {
+            return ReadUtils.emptyRead(read);
```
jamesemery: Right, it seems like this bug really was only for empty reads which caused the calculation of the new offset to break.

src/test/java/org/broadinstitute/hellbender/utils/clipping/ReadClipperUnitTest.java
```diff
@@ -444,4 +442,22 @@ public void testSoftClipByReferenceCoordinates() {
             }
         }
     }
+
+    // Test fix for https://github.com/broadinstitute/gatk/issues/3466
+    @Test
+    public void testHardClipSoftClippedBasesResultsInEmptyReadDontSetNegativeStartPosition() {
+        final GATKRead originalRead = ArtificialReadUtils.createArtificialRead(TextCigarCodec.decode("170H70S"));
+        // It's important that the read be near the start of the contig for this test, to test
```
jamesemery: Looking at the bug, this is technically not correct. The code was always producing bogus read start positions for empty reads before, its just that they never got error checked unless they fell off the contig. Either way this looks like a good test.

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
```diff
@@ -458,6 +458,23 @@ public void testReadsThatConsumeZeroReferenceReads() throws Exception {
         runCommandLine(args);
     }
 
+    // test fix for https://github.com/broadinstitute/gatk/issues/3466
+    @Test
+    public void testCompletelyClippedReadNearStartOfContig() throws Exception {
+        final File testCaseFilesDir = new File(TEST_FILES_DIR, "issue3466_gatk_cigar_error");
```
jamesemery: You should probably move these test files into some other folder, like the  HaplotypeCaller folder rather than making its own folder for this one commit

src/test/java/org/broadinstitute/hellbender/utils/clipping/ReadClipperUnitTest.java
droazen: What the comment means is that the test case only fails before the patch if the read is near the start of the contig. Otherwise you don't end up with a negative start, and the corrupt read gets replaced by a new "empty" read.

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
droazen: I'd rather keep them separated, since they are very specific to this bug.

PR #4082, opened 2018-01-08
Added missing requiredReference() argument to CombineGVCFs

jamesemery:


lbergelson:
@jamesemery Some test failures here https://storage.googleapis.com/hellbender-test-logs/build_reports/master_16064.3/tests/test/index.html

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerGroupedOnStartUnitTest.java
```diff
@@ -91,7 +91,7 @@ public void apply(List<VariantContext> variantContexts, ReferenceContext referen
     @SuppressWarnings({"unchecked"})
     public void testSimpleCaseDuplicateFile(File file, File file2) {
         final DummyExampleGroupingMultiVariantWalker tool = new DummyExampleGroupingMultiVariantWalker();
-        final String[] args = {"--variant", file.getAbsolutePath(), "--variant", file2.getAbsolutePath()};
+        final String[] args = {"--variant", file.getAbsolutePath(), "--variant", file2.getAbsolutePath(), "-R", b37_reference_20_21};
```
droazen: How were these tests passing without -R?

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerGroupedOnStart.java
```diff
@@ -114,6 +114,11 @@ public final void traverse() {
         afterTraverse();
     }
 
+    @Override
+    public boolean requiresReference() {
+        return true;
+    }
```
droazen: Move this override to the top of the class to make it more prominent.

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerGroupedOnStartUnitTest.java
```diff
@@ -133,7 +133,7 @@ public void testMergingTwoFilesCorrectBehavior() {
             List<String> inputList1 = ((List<VariantContext>)IteratorUtils.toList(variantContextFeatureDataSource1.iterator())).stream().map(VariantContext::toString).collect(Collectors.toList());
             List<String> inputList2 = ((List<VariantContext>)IteratorUtils.toList(variantContextFeatureDataSource2.iterator())).stream().map(VariantContext::toString).collect(Collectors.toList());
 
-            Assert.assertEquals(tool.seenVariants.size(), 14);
+            Assert.assertEquals(tool.seenVariants.size(), 13);
```
droazen: So the previous results without -R were just silently incorrect?

jamesemery: No, I changed the test file which had a read on chromosome 19

src/test/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerGroupedOnStartUnitTest.java
jamesemery: Because the walker actually handles the case where the reference context is null and will go on doing its grouping behavior properly. The reference context is needed for some of the functionality (importantly much of the functionality a tool cares about), but not the functionality I was testing here.

PR #4087, opened 2018-01-08
Python environment on OSX module dependency fix (libgcc-ng)

mbabadi:
Just removed it -- unnecessary module.

### DIFF
--- a/scripts/gatkcondaenv.yml
+++ b/scripts/gatkcondaenv.yml
@@ -6,7 +6,6 @@ channels:
 dependencies:
 - certifi=2016.2.28=py36_0
 - intel-openmp=2018.0.0
-- libgcc-ng=7.2.0
 - mkl=2018.0.1
 - mkl-service=1.1.2
 - openssl=1.0.2l=0


cmnbroad:
@mbabadi Thx. I think we need to take this, but have you actually run the tests on a Mac using this environment ? I can now create the environment with this change, but some of the python group tests fail when run from within it. If I just start Python with it activated and do "import gcnvkernel" or even "import pymc3", it fails. It might be a local issue -  it would be comforting to know that we've actually solved the full problem.

>>> import gcnvkernel
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/cnorman/miniconda3/envs/gatk/lib/python3.6/site-packages/gcnvkernel/__init__.py", line 1, in <module>
    from pymc3 import __version__ as pymc3_version
  File "/Users/cnorman/miniconda3/envs/gatk/lib/python3.6/site-packages/pymc3/__init__.py", line 12, in <module>
    from .sampling import *
  File "/Users/cnorman/miniconda3/envs/gatk/lib/python3.6/site-packages/pymc3/sampling.py", line 14, in <module>
    from .plots.traceplot import traceplot
  File "/Users/cnorman/miniconda3/envs/gatk/lib/python3.6/site-packages/pymc3/plots/__init__.py", line 1, in <module>
    from .autocorrplot import autocorrplot
  File "/Users/cnorman/miniconda3/envs/gatk/lib/python3.6/site-packages/pymc3/plots/autocorrplot.py", line 2, in <module>
    import matplotlib.pyplot as plt
  File "/Users/cnorman/miniconda3/envs/gatk/lib/python3.6/site-packages/matplotlib/pyplot.py", line 116, in <module>
    _backend_mod, new_figure_manager, draw_if_interactive, _show = pylab_setup()
  File "/Users/cnorman/miniconda3/envs/gatk/lib/python3.6/site-packages/matplotlib/backends/__init__.py", line 60, in pylab_setup
    [backend_name], 0)
  File "/Users/cnorman/miniconda3/envs/gatk/lib/python3.6/site-packages/matplotlib/backends/backend_macosx.py", line 17, in <module>
    from matplotlib.backends import _macosx
RuntimeError: Python is not installed as a framework. The Mac OS X backend will not be able to function correctly if Python is not installed as a framework. See the Python documentation for more information on installing Python as a framework on Mac OS X. Please either reinstall Python as a framework, or try one of the other backends. If you are using (Ana)Conda please install python.app and replace the use of 'python' with 'pythonw'. See 'Working with Matplotlib on OSX' in the Matplotlib FAQ for more information.

mbabadi:
@cmnbroad it turns out that this is a known issue of `matplotlib` on osx and conda environments: https://matplotlib.org/faq/osx_framework.html

On my MacBook, I used the base anaconda3 installation (which is "installed as a framework") and installed the required packages. When creating a virtual environment, however, the python is not installed as a framework on Mac and makes matplotlib complain.

The simplest workaround is to (1) require the user to have an mini(ana)conda3 installation, and (2) on osx, use `pythonw` instead of `python` to execute python codes.

cmnbroad:
Yeah, there are a few stackoverflow and github threads about this issue.
https://stackoverflow.com/questions/21784641/installation-issue-with-matplotlib-python

I tried the suggestion there, and it resolved it. I wish I knew what introduced it though, because I previously was able to import gcnvkernel without doing this (though I don't recall when the last time I did so was):

Create a file ~/.matplotlib/matplotlibrc there and add the following code: backend: TkAgg

cmnbroad:
Ok, so I was able to run all the python tests using  ~/.matplotlib/matplotlibrc. @mbabadi Wondering why you changed this to DO NOT MERGE. Do you have an alternative proposal ? Removing the libgcc-ng dependency doesn't solve the whole mac problem, but at least if we do remove it the workaround for the matplotlib part is easily conveyed.

Also, @samuelklee @mbabadi, is there any visibility for end-users that they need to establish the conda env to run the tool (like in the doc summary, etc.). I'm guessing the failure mode for just running without the environment will be pretty cryptic.

mbabadi:
@cmnbroad I'm testing the alternative of using `pythonw` instead of `python`. It works on Mac. If also works on Linux (Docker), isn't it an easier fix?

cmnbroad:
I don't think pythonw is on the docker image ?

cmnbroad:
Looks like this failed on travis. I think given that given the lateness of the hour (release wise), we might want to take the original change that removes the libgcc-ng dependency, since that passed on travis, and rely on the simple workarounds for osx, which we'll have to convey out-of-band.

Anything that requires changing the docker image seems risky at this point, not to mention that the image is already at 5.2 gig, which is way over our desired target.

@samuelklee Any thoughts on this ?

samuelklee:
@cmnbroad @mbabadi Let’s go with the matplotlibrc workaround and add some documentation to the README and Javadoc for the tools.  We can take a closer look after release.

cmnbroad:
Great thx.

Issue #1438, opened 2016-01-16
parse strange contig names correctly in the commandline -L argument  


yfarjoun:
Charlotte recently stepped on a bug in GATK: It interprets -L argument that ends in the regex ':[0-9]*' as indicating a single site in the contig that precedes it and then barfs if it cannot find that contig in the dictionary. In hg38 we have contig names like 'HLA:01:01:01' and when used on the command-line (as in CreateRealignerTargets) it barfs as in the following workflow: https://picard.broadinstitute.org/pipeline/workflows/viewWorkflow/8536444

given that the SAM spec allows any printed character ! through ~ in the ending of contig names (yikes!!) https://github.com/samtools/hts-specs/issues/124 it seems that some more "smarts" needs to be put into the parsing of this argument.

droazen:
This will be tricky, as GATK4, unlike GATK3, does not require that a sequence dictionary be present (eg., tools can take just a vcf, for example). It is a bit crazy that these problematic characters are forbidden in the VCF spec but not the SAM spec...

cmnbroad:
#4093 detects ambiguities, but throws when it finds them. Reopening this to keep the history, since we should still probably invent some kind of quoting mechanism to allow the user to resolve ambiguities.

droazen:
This is done well enough for hg38 purposes -- we can open a separate ticket if we want to go further with a quoting mechanism.

PR #4093, opened 2018-01-08
Handle interval queries against hg38 contigs. Detect and reject ambiguous query intervals.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/1438. It is still in theory possible to have queries that are ambiguous that will be rejected (though not with hg38). Longer term we'll need to implement a mechanism for quoting or otherwise allowing the user to handle how to resolve those.


src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
```diff
@@ -4,10 +4,16 @@
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.SAMSequenceRecord;
 import htsjdk.samtools.util.Locatable;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.read.mergealignment.SamAlignmentMerger;
```
droazen: Remove stray import

src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
```diff
@@ -135,12 +142,124 @@ public SimpleInterval(final String str){
...
+      * {@code sequenceDictionary}. If the list is empty, the query doesn't match any contig in the sequence
+      * dictionary. If the list contains more than one interval, the query string is ambiguous and should be
+      * rejected. If the list contains a single interval, the query is unambiguous and can be safely used to
+      * conduct a query.
+      * @thows IllegalArgumentException if the query only matches a single contig in the dictionary, but the start
+      * and end positions are not valis
+      * @throws NumberFormatException if the query only matches a single contig in the dictionary, but the query
+      * interval paramaters (start, end) cannot be parsed
+      */
+     public static List<SimpleInterval> getResolvedIntervals(
```
droazen: Can this be moved to `IntervalUtils`? Have this huge utility method directly in `SimpleInterval` like this is a little gross...  (if it would be non-trivial we can move it in a separate PR).

src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
```diff
@@ -135,12 +142,124 @@ public SimpleInterval(final String str){
...
+             final String intervalQueryString,
+             final SAMSequenceDictionary sequenceDictionary) {
+         Utils.nonNull(intervalQueryString);
+         Utils.validateArg(!intervalQueryString.isEmpty(), "intervalQueryString should not be empty");
```
droazen: Sequence dictionary must also be non-null

src/test/java/org/broadinstitute/hellbender/utils/GenomeLocParserUnitTest.java
```diff
@@ -504,4 +505,21 @@ public void testGenomeLocPadding(final GenomeLoc input, final int pad) {
         Assert.assertEquals(padded.getStop(), Math.min(input.getStop() + pad, contigLength));
     }
 
+    @Test
+    public void testQueryAllHG38Intervals() {
+        SAMSequenceDictionary sd;
+        final File testFile = new File ("src/test/resources/org/broadinstitute/hellbender/engine/Homo_sapiens_assembly38.headerOnly.vcf.gz");
```
droazen: Use inherited `publicTestDir` constant

src/test/java/org/broadinstitute/hellbender/utils/GenomeLocParserUnitTest.java
```diff
@@ -504,4 +505,21 @@ public void testGenomeLocPadding(final GenomeLoc input, final int pad) {
...
+        try (VCFFileReader vcfReader = new VCFFileReader(testFile, false)) {
+            sd = vcfReader.getFileHeader().getSequenceDictionary();
+        }
+
+        // Test that we can use any contig from hg38 as a query against a VCF with an hg38 sequence dictionary without
+        // having an ambiguity.
+        final GenomeLocParser localGenomeLocParser = new GenomeLocParser(sd);
+        sd.getSequences().stream().forEach(hg38Contig -> localGenomeLocParser.parseGenomeLoc(hg38Contig.getSequenceName()));
+        sd.getSequences().stream().forEach(hg38Contig -> localGenomeLocParser.parseGenomeLoc(hg38Contig.getSequenceName() + ":1-1"));
+        sd.getSequences().stream().forEach(hg38Contig -> localGenomeLocParser.parseGenomeLoc(hg38Contig.getSequenceName() + ":1+"));
```
droazen: Add a test for the "contig:1" case here. Also, assert that the return value of `parseGenomeLoc()` is as expected

src/test/java/org/broadinstitute/hellbender/utils/IntervalUtilsUnitTest.java
```diff
@@ -439,6 +440,149 @@ public void testParseIntervalArguments() {
...
+                    ambiguousIntervals = SimpleInterval.getResolvedIntervals(
+                            new SimpleInterval(hg38Contig.getSequenceName(), 1, hg38Contig.getSequenceLength()).toString(), sd);
+                    Assert.assertNotNull(ambiguousIntervals);
+                    Assert.assertEquals(ambiguousIntervals.size(), 1);
+
+                    // query using an interval with the format "name:1+"
+                    ambiguousIntervals = SimpleInterval.getResolvedIntervals(hg38Contig.getSequenceName() + ":1+", sd);
+                    Assert.assertNotNull(ambiguousIntervals);
+                    Assert.assertEquals(ambiguousIntervals.size(), 1);
+                }
```
droazen: Also test the ":1" case, and check the contents of `ambiguousIntervals` to make sure the interval it contains is correct.

src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/utils/GenomeLocParserUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/utils/IntervalUtilsUnitTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/utils/GenomeLocParserUnitTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/SimpleInterval.java
cmnbroad: Done.

cmnbroad:
Also rebased.

jonn-smith:
Nothing jumps out at me (modulo the changes droazen suggested).  I think it should be fine.

droazen:
Go ahead and merge @cmnbroad :+1:

PR #4095, opened 2018-01-08
Added statement to the stderr output stating the HTSJDK and Picard versions

jamesemery:
As of yet, this only works if you launch gatk from the jar or the gatk launch script.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -20,7 +20,12 @@ import org.broadinstitute.hellbender.utils.config.ConfigFactory;
 import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
 
 import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.InetAddress;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
 import java.text.DecimalFormat;
 import java.time.Duration;
 import java.time.ZonedDateTime;
@@ -28,6 +33,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -263,6 +270,27 @@ public final String getVersion() {
                 "Unavailable";
     }
 
+    public String getLibraryVersions() {
```
magicDGS: Any chance of making this, somehow, configurable? For example, using a provided map of "manifest entry name" and "dsplay name": "htsjdk-Version" -> "HTSJDK" and "Picard" -> "Picard-version", and then printing for each entry:
```
final String version = manifestAttributes.getValue(entry.getKey());
String.format("%s Version: %s", entry.getValue(), version != null ? version : "unknown");
```

Another option is to name everything with "${LibraryName}-Version" and pull out every entry in the MANIFEST... It is mainly to do not repeat code all around...

droazen: You should be able to configure it by overriding `getLibraryVersions()` with your own implementation.

magicDGS: I know, but the code here is quite useful: in addition, it requires to overwrite in every walker/extended class. Maybe extracting an utility method somewhere might be useful for donwstream projects, that's what I meant!

droazen: Sure, we can certainly extract a utility method for retrieving attributes from the manifest.

It sounds like you might want a base class like `GATKTool` for your toolkit that extends `CommandLineProgram` and that clients of your toolkit can extend. That way, you don't have to override methods like this in every tool as you would if all tools extend `CommandLineProgram` directly.

droazen: Though it occurs to me that this solution would be awkward if you want to extend a mix of built-in and custom walker classes in your toolkit. We'll try to provide a better mechanism of customizing the `CommandLineProgram` output in the future, but it will have to wait until after the 4.0 release unfortunately.

magicDGS: Yes, that's why I was thinking about: e.g., `GATKTool` and `ReadWalker` should be extended in custom toolkits to have both the same information. I can wait until release, but I was thinking that it might be useful to have a common `CommandLineProgram` in Barclay with some of this functionality made easier for any toolkit (extended or not).

Thank you for the discussion!

PR #4100, opened 2018-01-09
Read filters: kebabify args

droazen:


magicDGS:
Sorry, I know that I am late because it is already merged. Could it be possible to extract all the arguments in the `ReadFilter` classes to a `ReadFilterArguments` class with static values? That will make easier to figure out which incompatibilities my users will have, and to override the whole class before migrating my toolkit to kebab-case.

I can open a quick PR for that. Thank you in advance!

droazen:
@magicDGS We're getting close to release time, so I'm afraid this will have to wait. Feel free to open a PR after 4.0 is out.

magicDGS:
With which frequency will you make point-releases? I'm afraid that incompatibilities will make me stick to an unreleased GATK4 dependency!

I am preparing a PR at this moment to try to get it in!

droazen:
Don't worry @magicDGS, I expect point releases will be pretty frequent, at least in the early days after the 4.0 release. I'd be very surprised if we don't do one within the next few weeks. Unfortunately our plate is full this morning and we can't accept any more patches before the release.

PR #4107, opened 2018-01-09
adding a more prominent warning to HaplotypeCallerSpark

lbergelson:


src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -126,6 +129,9 @@ public boolean requiresReference(){
 
     @Override
     protected void runTool(final JavaSparkContext ctx) {
+        logger.info("The output of this tool DOES NOT match the output of HaplotypeCaller. " +
```
droazen: Can you add some asterisks for emphasis? :)

PR #4110, opened 2018-01-09
build_docker.sh: don't rebuild the image when tagging latest

droazen:
This fixes an issue in build_docker.sh where the script was
rebuilding the image from scratch when creating the "latest"
tag.


droazen:
@jamesemery please review

jamesemery:
👍 Yea, I guess that fell through the cracks for a while because we don't release much

PR #4118, opened 2018-01-10
Remove -SNAPSHOT from online doc name.

cmnbroad:
This retains the commit hash and commit number delta part in the version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -548,7 +548,7 @@ task gatkDoc(type: Javadoc, dependsOn: classes) {
         options.addStringOption("output-file-extension", phpExtension)
         options.addStringOption("index-file-extension", phpExtension)
     }
-    options.addStringOption("absolute-version", getVersion())
+    options.addStringOption("absolute-version", gitVersion().replaceAll("-SNAPSHOT", "").replaceAll(".dirty", ""))
     options.addStringOption("build-timestamp", new Date().format("dd-mm-yyyy hh:mm:ss"))
 }
 


samuelklee:
@lbergelson @droazen Just noticed the M2 WDL test is failing here---is this similar to the CNV WDL failures you were seeing just before release?

````
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
````

If so, this is definitely something I've seen more frequently lately.  Not sure what the remedy is, but would probably be worth looking into soon so we don't lose too much time to intermittent failures.

**Comment by cmnbroad** [comment]
@samuelklee Yes, this looks similar to what happened in master yesterday. In that case it looked like a transient remote access issue during the docker build. Having said that, its strange that the two data points we have were both on M2 WDL build.

**Comment by lbergelson** [comment]
I opened https://github.com/broadinstitute/gatk/issues/4129 to keep track of this.  If we can save test logs there when they happen maybe it will help shed light on the problem.

File: build.gradle [hunk 1]
```diff
@@ -548,7 +548,7 @@ task gatkDoc(type: Javadoc, dependsOn: classes) {
         options.addStringOption("output-file-extension", phpExtension)
         options.addStringOption("index-file-extension", phpExtension)
     }
-    options.addStringOption("absolute-version", getVersion())
+    options.addStringOption("absolute-version", gitVersion().replaceAll("-SNAPSHOT", "").replaceAll(".dirty", ""))
```
droazen: If the GATK version is a SNAPSHOT as opposed to a release, shouldn't that be clearly reflected in the docs? I'm not sure I understand the motivation behind this change.

droazen:
Closing this one -- we don't want to strip out SNAPSHOT from the version in docs generated from an actual snapshot. 

If docs for an actual release end up with SNAPSHOT in the version, you probably need to run with `-Drelease=true`.

PR #4119, opened 2018-01-10
Removed references to ACNV.

samuelklee:
Just some minor instances that slipped through during the rewrite.


mbabadi:
Good to merge.

PR #4128, opened 2018-01-11
made min-base-quality-score argument actually do something. Closes #4126.

### LINKED ISSUE #4126: min-base-quality-score argument doesn't do anything
`HaplotypeCaller` and `Mutect2` share an argument `min-base-quality-score` that restricts low-quality bases from assembly.  This argument gets passed to `ReadThreadingAssembler`, but take a look at the following code in that class:

```java
private static final byte MIN_BASE_QUALITY_TO_USE_IN_ASSEMBLY = DEFAULT_MIN_BASE_QUALITY_TO_USE;
protected byte minBaseQualityToUseInAssembly = DEFAULT_MIN_BASE_QUALITY_TO_USE;
```
The latter variable is set from the command line argument, but only the `static` constant is ever used, in particular in line 447:

```java
final ReadThreadingGraph rtgraph = new ReadThreadingGraph(kmerSize, debugGraphTransformations, MIN_BASE_QUALITY_TO_USE_IN_ASSEMBLY, numPruningSamples);
```

The fix is extremely simple: just replace `MIN_BASE_QUALITY_TO_USE_IN_ASSEMBLY` with `minBaseQualityToUseInAssembly`.

davidbenjamin:
@lbergelson Does this look reasonable to you?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/readthreading/ReadThreadingAssembler.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/readthreading/ReadThreadingAssembler.java
@@ -57,9 +57,7 @@ public final class ReadThreadingAssembler {
     private boolean debugGraphTransformations = false;
     private boolean recoverDanglingBranches = true;
     private int minDanglingBranchLength = 0;
-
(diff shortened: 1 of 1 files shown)


droazen:
@davidbenjamin Can you implement a simple integration test for this arg, to ensure it doesn't break again?

PR #4137, opened 2018-01-11
Sh sv cluster image ver upgrade and custom name

SHuang-Broad:
does two small things:
1. bump the image version so that we don't get the annoying "org.bdgenomics.adam.serialization.ADAMKryoRegistrator: Did not find Spark internal class. This is expected for Spark 1." message
2. name of cluster created with manage_sv_pipeline.sh now post-fixed with "master" or "feature" depending on if running master or a feature branch.

@TedBrookings can you please review?

### DIFF
--- a/scripts/sv/manage_sv_pipeline.sh
+++ b/scripts/sv/manage_sv_pipeline.sh
@@ -176,10 +176,16 @@ export GATK_GCS_STAGING=${GATK_GCS_STAGING:-"gs://${PROJECT_NAME}/${GCS_USER}/st
 
 echo
 
-# set cluster name based on user and target bam file
+# set cluster name based on user and target bam file and branch name
 # (NOTE: can override by defining SV_CLUSTER_NAME)
 SANITIZED_BAM=$(basename "${GCS_BAM}" | awk '{print tolower($0)}' | sed 's/[^a-z0-9]/-/g')
-CLUSTER_NAME=${SV_CLUSTER_NAME:-"${GCS_USER}-${SANITIZED_BAM}"}
+GATK_GIT_HASH=$(readlink ${GATK_DIR}/build/libs/gatk-spark.jar | awk 'BEGIN {FS="-g"} {print $2}' | cut -d- -f 1)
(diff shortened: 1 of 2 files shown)


scripts/sv/manage_sv_pipeline.sh
```diff
@@ -176,10 +176,16 @@ export GATK_GCS_STAGING=${GATK_GCS_STAGING:-"gs://${PROJECT_NAME}/${GCS_USER}/st
...
 echo
 
-# set cluster name based on user and target bam file
+# set cluster name based on user and target bam file and branch name
 # (NOTE: can override by defining SV_CLUSTER_NAME)
 SANITIZED_BAM=$(basename "${GCS_BAM}" | awk '{print tolower($0)}' | sed 's/[^a-z0-9]/-/g')
-CLUSTER_NAME=${SV_CLUSTER_NAME:-"${GCS_USER}-${SANITIZED_BAM}"}
+GATK_GIT_HASH=$(readlink ${GATK_DIR}/build/libs/gatk-spark.jar | awk 'BEGIN {FS="-g"} {print $2}' | cut -d- -f 1)
+GIT_BRANCH=$(git -C ${GATK_DIR} branch --contains ${GATK_GIT_HASH} | rev | cut -d" " -f 1 | rev)
+SANITIZED_GIT_BRANCH="master"
```
TedBrookings: Another alternative would be
SANITIZED_GIT_BRANCH=$(echo "${GIT_BRANCH}" | awk  '{print tolower($0)}' | sed 's/[^a-z0-9]/-/g')
If a user is on master, SANITIZED_GIT_BRANCH would still be master, but otherwise it should closely resemble the branch name unless they go out of their way to be difficult.

SHuang-Broad: Thanks for the suggestion!
I thought about it, but the branch name's length cannot be controlled and 
we probably will endup exceed the length limit of a cluster's name. 
Besides, we are likely only working on 1 feature branch at a time.

PR #4139, opened 2018-01-12
Moved HC-only arguments out of AssemblyBasedCallerArgumentCollection

### LINKED ISSUE #4134: Move args used in HC but not M2 out of shared ArgumentCollection
AssemblyBasedCallerArgumentCollection contains arguments that Mutect2 doesn't use, which is misleading.

davidbenjamin:
Closes #4134.

@droazen Do you want someone on the engine team to review this?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
@@ -66,8 +66,6 @@ public class HaplotypeCallerArgumentCollection extends AssemblyBasedCallerArgume
     /**
      * The reference confidence mode makes it possible to emit a per-bp or summarized confidence estimate for a site being strictly homozygous-reference.
      * See http://www.broadinstitute.org/gatk/guide/article?id=2940 for more details of how this works.
-     * Note that if you set -ERC GVCF, you also need to set -variant_index_type LINEAR and -variant_index_parameter 128000 (with those exact values!).
-     * This requirement is a temporary workaround for an issue with index compression.
      */
     @Advanced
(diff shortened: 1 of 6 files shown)


droazen:
Yes, I'll review it

davidbenjamin:
@droazen Note that taking `dbSNP` and `comps` out of M2 is intentional.  That's just not how we use M2.  I would argue that dbSNP should be removed from HapotypeCaller, but that's for another day.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
```diff
@@ -39,10 +43,36 @@
...
+     */
+    @Advanced
+    @Argument(fullName = "comp", shortName = "comp", doc = "Comparison VCF file(s)", optional = true)
+    public List<FeatureInput<VariantContext>> comps = Collections.emptyList();
+
+    /**
+     * The reference confidence mode makes it possible to emit a per-bp or summarized confidence estimate for a site being strictly homozygous-reference.
+     * See http://www.broadinstitute.org/gatk/guide/article?id=2940 for more details of how this works.
+     * Note that if you set -ERC GVCF, you also need to set -variant_index_type LINEAR and -variant_index_parameter 128000 (with those exact values!).
+     * This requirement is a temporary workaround for an issue with index compression.
```
droazen: While you're here, would you mind deleting this obsolete section of the docs:

```
Note that if you set -ERC GVCF, you also need to set -variant_index_type LINEAR and -variant_index_parameter 128000 (with those exact values!).
This requirement is a temporary workaround for an issue with index compression.
```

davidbenjamin: done

PR #4144, opened 2018-01-12
PathSeq postprocessing scripts

mwalker174:
Adds two small Python scripts for processing output from the PathSeq pipeline:

1. aggregate_results.py : combines abundance scores and metrics across multiple samples. It outputs several files, one for metrics, one for scores, etc. each containing a table of values across all the samples.

2. extract_unmapped_reads.py : reads the output SAM from the pipeline from stdin and writes unmapped records (as determined by the absence of a YZ tag) to stdout.

### DIFF
--- a/scripts/pathseq/postprocessing/aggregate_results.py
+++ b/scripts/pathseq/postprocessing/aggregate_results.py
@@ -1,0 +1,147 @@
+#!/bin/python
+
+# This script aggregates score tables and metrics files from multiple PathSeq runs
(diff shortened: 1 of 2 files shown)


lbergelson:
If users are meant to make use of these, they should not be python scripts.

droazen:
@mwalker174 Do you still want to get this ancient PR in?

mwalker174:
@droazen Nope I can find a better home for this.

PR #4147, opened 2018-01-12
prevent log message from triggering incorrectly

lbergelson:
* moved a log message about creating the python bundle during build so it only triggers when the python bundle is actually being built
* previously it was emitted on every build


lbergelson:
@cmnbroad I moved this message so it runs during the actual task execution rather than at configuration time.

PR #4154, opened 2018-01-14
Replaced IntervalsSkipList with OverlapDetector

### LINKED ISSUE #3541: Would be better that IntervalsSkipList.getOverlapping would accept a Locatable rather than a SimpleInterval.
I can see the reassurance of knowing that the input Locatable is constant and with a non-null contig... yet as a result we are often creating redundant simpleIntervals instances when our objects of interest are some other type of Locatable.

### LINKED ISSUE #3608: deprecate and remove IntervalSkipList
I believe we discovered that `IntervalSkipList` had worse performance than htsjdk's `OverlapDetector`.  It's probably best to remove IntervalSkipList in this case and prefer the htsjdk version.

davidbenjamin:
Closes #3541.  Closes #3608.

@lbergelson May I assign you by virtue of your github activity on this and related issues?

Note that `OverlapDetector::getOverlaps` returns an unsorted `Set` so I sorted its output in a few places.  I don't know if this was the right solution.


droazen:
@davidbenjamin Have you actually profiled this change on a Spark/dataproc cluster?

lbergelson:
@droazen Adam did profiling that showed that overlaps detector was strictly better.  We also had some suspicion that there was some bug lurking in the skip list implementation because of weird non-deterministic results of something that relied on it.

droazen:
@lbergelson Adam did the profiling? That must have been years ago. Someone definitely needs to profile this change using the latest master to see if his ancient findings still hold up.

davidbenjamin:
@lbergelson @droazen I can do it but I would need a lot of hand-holding for both the profiling and the Spark.

droazen:
We're happy to help -- I think it's worth doing even if this is a no-brainer substitution, as it's easy to break Spark functionality on a cluster by silly things like introducing a non-serializable class in the wrong place.

davidbenjamin:
@droazen When we left off you showed me how to use JProfiler (which I now use regularly) and had pointed me to instructions for running GATK on a Spark cluster.  I realize I'm going to need another round of hand-holding, because I'm not sure which Spark cluster to test on, and I'm not sure which command to test.  And beyond measuring total runtime and looking for any new hotspots in JProfiler I don't know what else to measure.  Could I get some more help from someone on the engine team?

droazen:
@jamesemery Can you make a decision on what to do with this one after we merge https://github.com/broadinstitute/gatk/pull/5292 ?

lbergelson:
This is easily resolvable without performance profiling since the only remaining use of the `IntervalSkipList` was in `ContextShard` which is now also unused / untested.  Removing context shard and rebasing.

lbergelson:
@davidbenjamin I've stolen this branch from you.  I hope you don't mind.

davidbenjamin:
@lbergelson No objection.  It's in better hands than mine now.

lbergelson:
@davidbenjamin Thank you for this one!

PR #4159, opened 2018-01-15
Fix 4141

SHuang-Broad:
Separated into two commits, only the 2nd is relevant (1st commit is the squashed commit from PR #3934 ).

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.hellbender.tools.spark.sv.discovery.alignment;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMSequenceDictionary;
@@ -192,7 +193,7 @@ public class AssemblyContigAlignmentsConfigPicker {
                                               final Set<String> canonicalChromosomes,
                                               final int maxCanonicalChrAlignerScore) {
 
-        final double tigExplainQual = computeTigExplainQualOfOneAlignment(configuration, canonicalChromosomes, maxCanonicalChrAlignerScore);
+        final double tigExplainQual = computeTigExplainQualOfOneConfiguration(configuration, canonicalChromosomes, maxCanonicalChrAlignerScore);
 
         int redundancy = 0;
         for (int i = 0; i < configuration.size() -1 ; ++i) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
```diff
@@ -266,13 +266,62 @@ private static double computeTigExplainQualOfOneAlignment(final List<AlignmentIn
         return numFirst.thenComparing(mismatchSecond);
     }
 
-    private static List<AlignmentInterval> splitGaps(final List<AlignmentInterval> configuration) {
-        return configuration.stream()
+    @VisibleForTesting
+    static List<AlignmentInterval> splitGaps(final List<AlignmentInterval> configuration) {
+
+        final List<Integer> blackList = new ArrayList<>(configuration.size());
```
cwhelan: It seems inefficient to do the splitting twice. I'm also not a huge fan of this blacklist list-of-indices. How about a different approach, something like:

1) Make a map of input AI's that can be split to their split forms.
2) Make a copy of the input AI list to hold the results.
3) Check each of the split alignments from 1) against the input list. Replace the entries in the output list corresponding to their input alignments, and remove any entries from the output list that fail as a result of these checks.

Would that work?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
```diff
@@ -266,13 +266,62 @@ private static double computeTigExplainQualOfOneAlignment(final List<AlignmentIn
...
+                            blackList.add(i);
+                        }
+                    }
+                }
+            }
+        }
+
+        final List<AlignmentInterval> result = new ArrayList<>(configuration.size());
+        for (int i = 0; i < configuration.size(); ++i) {
+            if (blackList.indexOf(i) == -1)
```
cwhelan: I think `contains(i)` would be easier to read.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
SHuang-Broad: changed to am implementation where a list holds the split alignments (or original if no large gaps), for alignments that fails the check, its corresponding split is invalidated (set to null), then the list is collected on those whose split is still valid after the filter.

Just out of curiosity: are you concerned more about the blacklist approach in general or the fact that the blacklist was storing indices?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
```diff
@@ -192,7 +192,7 @@ static double computeScoreOfConfiguration(final List<AlignmentInterval> configur
                                               final Set<String> canonicalChromosomes,
                                               final int maxCanonicalChrAlignerScore) {
 
-        final double tigExplainQual = computeTigExplainQualOfOneAlignment(configuration, canonicalChromosomes, maxCanonicalChrAlignerScore);
+        final double tigExplainQual = computeTigExplainQualOfOnConfiguration(configuration, canonicalChromosomes, maxCanonicalChrAlignerScore);
```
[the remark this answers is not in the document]
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
SHuang-Broad: irrelevant now as switched to the suggested logic

Issue #4135, opened 2018-01-11
Piping picard tools in GATK fails because of logging to STDOUT


lbergelson:
As reported in https://github.com/broadinstitute/gatk/issues/4133 piped commands that work in Picard, don't work when running with GATK.  The immediate culprit is the line:
```
15:29:29.595 INFO  NativeLibraryLoader - Loading libgkl_compression.dylib from jar:file:/Users/louisb/Workspace/gatk/build/install/gatk/lib/gkl-0.8.2.jar!/com/intel/gkl/native/libgkl_compression.dylib
```
Which is emitted to stdout in GATK, a similar warning is instead output to STDERR in picard.  

There are likely other problematic logging lines.

Issue #4133, opened 2018-01-11
picard 2.10.7 fails pipelining SortSam and SetNmAndUqTags


sooheelee:
Bug on 4.0.0.0 reported on forum -- please advise how to proceed @droazen @lbergelson.

---

Ok now things are getting more interesting. Disabling intelflaters seems to solve the problem. But why?

Same flaters work with standalone picard as well and they don't have any issues. Compression level does not change anything. Is this a memory allocation error? I don't think so because it does not segfault. The error message says filename not found. 

I am wondering if this has anything to do with meltdown patches. Both of my systems (Ubuntu and macOS) are patched against meltdown bug. And Intel flaters are causing this issue. But this won't explain why standalone picard is still working fine with them. The only explanation that I have is the difference between HTSJDK and GKL versions between standalone picard 2.17.2 and embedded picard 2.17.2 in gatk. The latest HTSJDK update for standalone picard mentions 2.13.1 version but the one in gatk 4.0 is 2.13.2 I believe. I don't have the info about GKL version used in them. 

This Issue was generated from your [forums] 
[forums]: https://gatkforums.broadinstitute.org/gatk/discussion/comment/44985#Comment_44985

sooheelee:
For @droazen to delegate.

lbergelson:
I commented on the message itself.  The problem seems to be that we output some log messages to STDOUT that break piping.  Picard puts them on STDERR.

PR #4167, opened 2018-01-16
Changed logging output to stderr

jamesemery:
Tested that piping seems to work by running `$ ./gatk SortSam -I file.bam -O /dev/stdout -SO coordinate | ./gatk SetNmMdAndUqTags -I /dev/stdin -O testSorted.bam --CREATE_INDEX true -R reference.fasta`

Fixes https://github.com/broadinstitute/gatk/issues/4135
Fixes https://github.com/broadinstitute/gatk/issues/4133

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/engine/PipelineSupportIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/engine/PipelineSupportIntegrationTest.java
@@ -3,19 +3,31 @@ package org.broadinstitute.hellbender.engine;
 import org.broadinstitute.hellbender.GATKBaseTest;
 import org.broadinstitute.hellbender.utils.runtime.ProcessController;
 import org.broadinstitute.hellbender.utils.test.BaseTest;
+import org.testng.Assert;
 import org.testng.annotations.Test;
 
 import java.io.File;
 
-public class PipelineSupportIntegrationTest extends GATKBaseTest{
+public class PipelineSupportIntegrationTest extends GATKBaseTest {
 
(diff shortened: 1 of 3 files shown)


droazen:
@jamesemery Could you also add an integration test with the pipeline above? You'll need to run bash using `ProcessController` or similar

src/test/java/org/broadinstitute/hellbender/engine/PipelineSupportIntegrationTest.java
```diff
@@ -0,0 +1,21 @@
...
+public class PipelineSupportIntegrationTest extends GATKBaseTest{
+
+    private static final String INPUT_BAM = toolsTestDir + "print_reads.sorted.chr1_1.bam";
+
+    @Test
+    // Asserting that Picard programs are able to pipe their output without errors
+    public void testPipeForPicardTools() {
+        File output = createTempFile("testOutput",".bam");
+        BaseTest.runProcess(ProcessController.getThreadLocal(), ("./gatk SortSam -I "+INPUT_BAM+" -O /dev/stdout -SO coordinate " +
+                "| ./gatk SetNmMdAndUqTags -I /dev/stdin -O "+ output.getAbsolutePath()+ " --CREATE_INDEX true -R "+hg19_chr1_1M_Reference).split(" "));
```
droazen: Can you add an assert that after this pipeline runs, the output bam exists, and has an expected number of records in it?

jamesemery: Sure, I considered this sufficient for the piping bug because it would exit with an error otherwise but for added security I will add proper asserts.

jamesemery:
@droazen Updated the test so it's more stringent

PR #4176, opened 2018-01-16
Funcotator - Added map-style accessor for fields.

### LINKED ISSUE #3919: Funcotator - Refactor Funcotation class to use a HashMap for each field
It would be better if in addition to / instead of the getter/setter scheme that exists with current `Funcotation` classes we used a `HashMap` to be able to directly get each field by name.  This would allow for more programmatic manipulation of `Funcotation` objects.

jonn-smith:
Added map-style accessors to all concrete Funcotation classes (Funcotation.`getField`).

Fixes #3919


LeeTL1220:
@jonn-smith This looks okay.  Merge away.

PR #4183, opened 2018-01-17
Use BED files as a way to resolve ambiguous queries.

cmnbroad:
Add a test that validates than an ambiguous interval query can be disambiguated by the user by providing the interval in a bed file; changes the error message to recommend this alternative; fixes an issue where the error message was displaying the entire interval query multiple times, rather than the specific contigs which make the query ambiguous.

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/utils/IntervalUtilsUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/IntervalUtilsUnitTest.java
@@ -26,6 +26,7 @@ import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 import java.io.File;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.util.*;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


src/test/java/org/broadinstitute/hellbender/utils/IntervalUtilsUnitTest.java
```diff
@@ -577,6 +578,43 @@ public void testUseAllHG38ContigsInIntervalQuery() {
...
+        }
+
+        // now use the bed file
+        final List<GenomeLoc> intervalList = IntervalUtils.parseIntervalArguments(genomeLocParser, bedFile.getAbsolutePath());
+
+        // validate the (translated from Bed) interval coordinates
+        Assert.assertEquals(intervalList.size(), 1);
+        Assert.assertEquals(intervalList.get(0).getContig(), "HLA-A*01:02:03:04");
+        Assert.assertEquals(intervalList.get(0).getStart(), bedStart + 1);
+        Assert.assertEquals(intervalList.get(0).getEnd(), 99);
```
cmnbroad: Just noticed this should use bedEnd. Will fix after review.

droazen:
@lbergelson please review

src/test/java/org/broadinstitute/hellbender/utils/IntervalUtilsUnitTest.java
```diff
@@ -577,6 +578,44 @@ public void testUseAllHG38ContigsInIntervalQuery() {
...
+        final int BED_END = 99;
+        final String TARGET_CONTIG = "HLA-A*01:02:03:04";
+        try (FileWriter bedWriter = new FileWriter(bedFile)) {
+            bedWriter.write(String.format("%s\t%d\t%d", TARGET_CONTIG, BED_START, BED_END));
+        }
+        final GenomeLocParser genomeLocParser = new GenomeLocParser(ambiguousSequenceDictionary);
+
+        // first, make sure we're using a query that is actually ambiguous
+        try {
+            IntervalUtils.parseIntervalArguments(genomeLocParser, TARGET_CONTIG);
```
lbergelson: I think this check would be cleaner as:
```
Assert.assertThrows(UserException.MalformedBAM.class, () ->  IntervalUtils.parseIntervalArguments(genomeLocParser, TARGET_CONTIG));
 ```

cmnbroad: Yeah, done.

PR #4187, opened 2018-01-17
IndexFeatureFile: more informative error message when trying to index a malformed file

### LINKED ISSUE #4184: IndexFeatureFile can't index .vcf.gz using ProgressReportingDelegatingCodec
When trying to index a gzipped VCF at `/humgen/gsa-hpprojects/dev/gauthier/reblockGVCF/hybrid.m38.vcf.gz`I get the error `A USER ERROR has occurred: This tool does not supports indexing of block-compressed files for ProgressReportingDelegatingCodec`

Indexing this file with tabix is speedy and successful.

droazen:
Resolves #4184

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
@@ -169,7 +169,7 @@ public final class IndexFeatureFileIntegrationTest extends CommandLineProgramTes
         checkIndex(index, Arrays.asList("1", "2", "3", "4"));
     }
 
-    @Test(expectedExceptions = UserException.MalformedFile.class)
+    @Test(expectedExceptions = UserException.CouldNotIndexFile.class)
     public void testVCFGZIPIndex() throws IOException {
         //This tests blows up because the input file is not blocked gzipped
         final File ORIG_FILE = getTestFile("test_variants_for_index.vcf.gzip.gz"); //made by gzip
@@ -181,7 +181,7 @@ public final class IndexFeatureFileIntegrationTest extends CommandLineProgramTes
         final Object res = this.runCommandLine(args);
     }
 
-    @Test(expectedExceptions = UserException.MalformedFile.class)
+    @Test(expectedExceptions = UserException.CouldNotIndexFile.class)
     public void testVCFGZIPIndex_inferredName() throws IOException {
         //This tests blows up because the input file is not blocked gzipped
         final File ORIG_FILE = getTestFile("test_variants_for_index.vcf.gzip.gz"); //made by gzip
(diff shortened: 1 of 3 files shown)


droazen:
@ldgauthier please review

src/test/java/org/broadinstitute/hellbender/tools/IndexFeatureFileIntegrationTest.java
```diff
@@ -169,7 +169,7 @@ public void testVCFGZIndex_inferredName(){
         checkIndex(index, Arrays.asList("1", "2", "3", "4"));
     }
 
-    @Test(expectedExceptions = UserException.MalformedFile.class)
```
ldgauthier: @lbergelson and I have lamented how many of our tests expect non-subclassed UserExceptions and then get called with missing input files and everything passes.  I'd be happier if IndexFeatureFile threw a more specific UserException, but I understand you're lumping together a few different cases.  Maybe add a UserException.TribbleParseError or something similarly named?

droazen: Unfortunately the underlying cause here can be anything from "file is malformed" to "codec does not support tabix", and it all comes out of htsjdk as a `TribbleException`. I thought about using `UserException.MalformedFile` here, which will be accurate 90%+ of the time, but there is that other 10%...

lbergelson: Over time every test that catches raw UserException breaks and starts to pass incorrectly due to changes in the command line inputs.

lbergelson: Yeah, it's hard when all we know is "Something is wrong"

droazen: But I can switch to `UserException.MalformedFile` all the same, if you prefer.

ldgauthier: I would prefer UserException.MalformedFile to vanilla UserException.

droazen: @lbergelson Again, the problem is in htsjdk in lumping all tribble errors together. Our choices in GATK are to throw a more-specific but sometimes-completely-inaccurate exception, or a more general exception with exposure of the underlying error message.

lbergelson: Fixing htsjdk errors would be a nice benefit of redoing things in htsjdk, although it's going to be a big undertaking I think.

droazen: How about I introduce a `UserException.FailureToIndexFile` exception here -- that would be accurate and address your concerns about tests erroneously passing.

lbergelson: I'm actually more in favor of throwing the more general exception with the exposure of the underlying issue.  It's just nasty for testing.

lbergelson: Good idea!

ldgauthier: I like it.

droazen:
@ldgauthier Introduced a suitably-vague `UserException.CouldNotIndexFile` in response to your comments. Ready for a last look.

ldgauthier:
:+1:

PR #4191, opened 2018-01-17
Restored ability to make single-sample ModelSegments PoN.

samuelklee:
Original behavior during development relied on Spark mllib's computeSVD to return something sensible even if a PoN was built with only one sample; typically this would yield a singular value of 0 and an eigenvector of NaNs (although it can also result in miscellaneous exceptions, depending on how many zeros are contained in the single-row matrix on which we perform SVD).  At one point, I added a check that at least one non-negligible singular value was returned (even when building a PoN with multiple samples).  This spoiled the ability to build single-sample PoNs (which is useful if you only have a matched pair).

I've added more explicit logic about how single-sample PoNs are built.  Now we explicitly set both the singular value and eigenvector to zero.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/HDF5SVDReadCountPanelOfNormals.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/HDF5SVDReadCountPanelOfNormals.java
@@ -305,32 +305,41 @@ public final class HDF5SVDReadCountPanelOfNormals implements SVDReadCountPanelOf
             }
             logger.info(String.format("Performing SVD (truncated at %d eigensamples) of standardized counts (transposed to %d x %d)...",
                     numEigensamples, numPanelIntervals, numPanelSamples));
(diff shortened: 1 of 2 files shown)


samuelklee:
@LeeTL1220 A short one for you.  Thanks!

src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/HDF5SVDReadCountPanelOfNormals.java
```diff
@@ -305,32 +305,41 @@ public static void create(final File outFile,
             }
             logger.info(String.format("Performing SVD (truncated at %d eigensamples) of standardized counts (transposed to %d x %d)...",
                     numEigensamples, numPanelIntervals, numPanelSamples));
```
LeeTL1220: Is there a test here for a single sample PoN?  Apologies if I missed it.

samuelklee: Yup, I added single-sample tests, but also simplified how the GC/no-GC tests are run---so you might have missed it.

PR #4193, opened 2018-01-17
Standardized use of mem parameters in CNV WDLs.

### LINKED ISSUE #4092: Standardize memory units and use of machine_mem and command_mem in CNV WDLs.

samuelklee:
We use the `machine_mem`/`command_mem` framework for most tasks; others are unlikely to need any special memory considerations.

There were some tasks for which `machine_mem` and `command_mem` seemed to be switched in the original WDL.  @jsotobroad was there any reason for this?  I'm guessing they were just typos.  I'm also not sure that some of the tasks would've actually run even if they were switched, since they would've resulted in non-integer -Xmx arguments.  I changed everything over to MB to avoid this.

Closes #4092.


samuelklee:
@davidbenjamin mind reviewing?  @LeeTL1220 take note for the style guide, if necessary.

scripts/cnv_wdl/cnv_common_tasks.wdl
```diff
@@ -92,6 +92,9 @@ task CollectCounts {
     Int? preemptible_attempts
     Int? disk_space_gb
 
+    Int machine_mem = select_first([mem, 8]) * 1000
```
jsotobroad: As far as small optimizations go, if these tasks that are asking for `8000 MB` by default can make due with `7500 MB` then you can save 25% on GCP compute - https://cloud.google.com/compute/pricing#machinetype - `n1-standard-2` vs`n1-highmem-2`

scripts/cnv_wdl/cnv_common_tasks.wdl
```diff
@@ -137,8 +140,7 @@ task CollectAllelicCounts {
     Int? preemptible_attempts
     Int? disk_space_gb
 
-    # Mem is in units of GB but our command and memory runtime values are in MB
```
jsotobroad: Somewhere it should made clear that the user provided `mem` input should be defined in units of GB

scripts/cnv_wdl/cnv_common_tasks.wdl
```diff
@@ -110,7 +113,7 @@ task CollectCounts {
 
     runtime {
         docker: "${gatk_docker}"
```
jsotobroad: As per @LeeTL1220's template, we should be putting in values for `cpu` even in the single threaded case.

scripts/cnv_wdl/cnv_common_tasks.wdl
samuelklee: Any objection to simply using `*_mem_gb` and `*_mem_mb` everywhere?  I'd rather not duplicate this comment everywhere. 
 @LeeTL1220 @davidbenjamin?

LeeTL1220: @samuelklee Fine by me...

scripts/cnv_wdl/cnv_common_tasks.wdl
samuelklee: I didn't check to make sure all of the tasks adhered to the template just yet, since I'm assuming that it might still change.  (Actually, I don't think any of the somatic CNV tasks specify cpu.)  I'll focus on the mem changes in this PR and overhaul the rest later (perhaps once tasks are automatically generated), if you don't mind!

scripts/cnv_wdl/cnv_common_tasks.wdl
samuelklee: OK, good to know!  No particular reason these need Xmx8G, so I'll change them.

scripts/cnv_wdl/cnv_common_tasks.wdl
samuelklee: Done.

samuelklee:
Will merge when tests pass, unless there are further objections.

scripts/cnv_wdl/cnv_common_tasks.wdl
LeeTL1220: @samuelklee Fine by me.

LeeTL1220:
@samuelklee From my perspective, merge away.  Assuming that you are doing the cpu change later.

scripts/cnv_wdl/cnv_common_tasks.wdl
```diff
@@ -87,11 +87,14 @@ task CollectCounts {
...
 
     # Runtime parameters
-    Int? mem
+    Int? mem_gb
     String gatk_docker
     Int? preemptible_attempts
     Int? disk_space_gb
 
+    Int machine_mem_mb = select_first([mem_gb  * 1000, 7500])
+    Int command_mem_mb = machine_mem_mb - 1000
```
davidbenjamin: Why don't you do this in the tasks above?

scripts/cnv_wdl/cnv_common_tasks.wdl
```diff
@@ -69,7 +69,7 @@ task AnnotateIntervals {
 
     runtime {
         docker: "${gatk_docker}"
-        memory: select_first([mem, 5]) + " GB"
+        memory: select_first([mem_gb, 5]) + " GB"
```
davidbenjamin: This is strange because if `mem_gb` is supplied the machine memory and command memory are the same, but if it's not supplied they use different defaults.

scripts/cnv_wdl/cnv_common_tasks.wdl
```diff
@@ -21,7 +21,7 @@ task PreprocessIntervals {
         set -e
         export GATK_LOCAL_JAR=${default="/root/gatk.jar" gatk4_jar_override}
 
-        gatk --java-options "-Xmx${default="2" mem}g" PreprocessIntervals \
+        gatk --java-options "-Xmx${default="2" mem_gb}g" PreprocessIntervals \
```
davidbenjamin: I don't feel great about using Xmx_g in some places and Xmx_m in others.

scripts/cnv_wdl/cnv_common_tasks.wdl
davidbenjamin: > Any objection to simply using *_mem_gb and *_mem_mb everywhere?

That is sufficiently self-documenting for my tastes.

scripts/cnv_wdl/germline/cnv_germline_cohort_workflow.wdl
```diff
@@ -349,8 +349,8 @@ task GermlineCNVCallerCohortMode {
     Float? caller_admixing_rate
     Boolean? disable_annealing
 
-    Int machine_mem = if defined(mem) then select_first([mem]) else 8
-    Float command_mem = machine_mem - 0.5
+    Int machine_mem_mb = select_first([mem_gb, 7]) * 1000
+    Int command_mem_mb = machine_mem_mb - 500
```
davidbenjamin: Elsewhere you have a 1GB difference.

scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl
```diff
@@ -366,15 +365,14 @@ task ModelSegments {
...
     String gatk_docker
     Int? preemptible_attempts
     Int disk_space_gb
 
-    # Mem is in units of GB but our command and memory runtime values are in MB
-    Int machine_mem = if defined(mem) then mem * 1000 else 13000
+    Int machine_mem_mb = select_first([mem_gb, 13]) * 1000
     # ModelSegments seems to need at least 3GB of overhead to run
-    Int command_mem = machine_mem - 3000
+    Int command_mem_mb = machine_mem_mb - 3000
```
davidbenjamin: Does this make sense to you?  I mean, shouldn't the difference between command and machine memory just be the OS, hence independent of the GATK command?

samuelklee: Not sure, this was from @jsotobroad.  I'd be fine with standardizing the difference everywhere if that works.

scripts/cnv_wdl/cnv_common_tasks.wdl
samuelklee: Oops, good catch.  I think for some of these more minor tasks we didn't do the whole machine_mem/command_mem thing.  I'll go back and standardize.

scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl
LeeTL1220: @jsotobroad can answer this, @davidbenjamin .

scripts/cnv_wdl/cnv_common_tasks.wdl
samuelklee: Thanks for catching this.  Actually, I think this makes the tests fail---you can't perform the multiply operation inside select_first, apparently.

samuelklee:
Responded to @davidbenjamin.  Still have differences for machine_mem_mb - command_mem_mb = 500, 1000, and 3000---@jsotobroad any reason for this?  Can we make everything the same?

LeeTL1220:
I'm good with this if @jsotobroad is good too.

davidbenjamin:
Me too.

samuelklee:
@jsotobroad I'm going to go ahead and merge this and file an issue for the difference in memory overhead.

PR #4205, opened 2018-01-18
Somatic CNV tagging germline events, bringing SimpleAnnotatedGenomicRegion in line with the collection conventions, and region merging, 

LeeTL1220:
Most of these changes are to support automated evaluation of GATK CNV.

- Updates SimpleAnnotatedGenomicRegion to use the collection framework used in other GATK CNV CLIs.

CLIs (both experimental quality): 
- `TagGermlineEvents` is a simple tool that attempts to identify events in a tumor seg file that correspond to a germline events.  
  - This is done purely with concordance on the breakpoints of the events (within some padding).  
  - Input germline segments must have calls. 
  - If a germline call is broken into multiple segments, this tool will handle that appropriately (ditto if there are multiple tumor segments overlapping the germline call)
  

- `MergeAnnotatedRegions` will merge all overlapping regions and resolve annotation value conflicts.


src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/AbstractLocatableCollection.java
```diff
@@ -31,7 +31,7 @@
     /**
      * @param metadata records are sorted using the contained {@link SAMSequenceDictionary}
      */
-    AbstractLocatableCollection(final METADATA metadata,
+    protected AbstractLocatableCollection(final METADATA metadata,
```
samuelklee: These are package-private so that subclasses must be placed in the collections package.  Easier to keep everything in one place.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/AbstractRecordCollection.java
```diff
@@ -184,7 +184,7 @@ private SAMFileHeader getHeader() throws FileNotFoundException {
 
         @Override
         protected boolean isCommentLine(final String[] line) {
-            return line.length > 0 && line[0].startsWith(COMMENT_PREFIX);
+            return line.length > 0 && (line[0].startsWith(COMMENT_PREFIX) || line[0].startsWith("#"));
```
samuelklee: Also going to veto this.  The `AbstractRecordCollection` should be a very strictly defined format.  If you have external GT files that we need to remove comment lines from a single time, let's do that externally rather than relax the format.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/AbstractRecordCollection.java
```diff
@@ -200,4 +200,8 @@ protected void composeLine(final RECORD record, final DataLine dataLine) {
             recordToDataLineEncoder.accept(record, dataLine);
         }
     }
+
+    public TableColumnCollection getMandatoryColumns() {
```
samuelklee: This is not used.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/CalledCopyRatioSegmentCollection.java
```diff
@@ -22,7 +22,7 @@
     /**
      * CONTIG, START, END, NUM_POINTS_COPY_RATIO, MEAN_LOG2_COPY_RATIO, CALL
      */
-    enum CalledCopyRatioSegmentTableColumn {
+    public enum CalledCopyRatioSegmentTableColumn {
```
samuelklee: This should also not be exposed.  See below.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/CombineSegmentBreakpoints.java
```diff
@@ -14,12 +15,13 @@
...
 import org.broadinstitute.hellbender.tools.copynumber.arguments.CopyNumberStandardArgument;
 import org.broadinstitute.hellbender.tools.copynumber.utils.annotatedregion.SimpleAnnotatedGenomicRegion;
+import org.broadinstitute.hellbender.tools.copynumber.utils.annotatedregion.SimpleAnnotatedGenomicRegionCollection;
 import org.broadinstitute.hellbender.utils.IntervalUtils;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
-import org.testng.collections.Lists;
 
 import java.io.File;
 import java.util.*;
+import java.util.function.Consumer;
```
samuelklee: Not in this PR, but we removed all short names from CNV tools as part of the kebab-casing effort.  Let's do the same for all tools in this PR.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/MergeAnnotatedRegions.java
```diff
@@ -0,0 +1,64 @@
...
+@CommandLineProgramProperties(
+        oneLineSummary = "(EXPERIMENTAL) Merge annotated genomic regions based entirely on contig and annotation value.",
+        summary = "(EXPERIMENTAL) Merge annotated genomic regions based entirely on contig and annotation value.  Column header order is not guaranteed to be preserved. Reference is required and will superseded any sequence dictionary in the given seg/region files.  Sequence dictionary is optional on the input file, but will be included on the output.  Conflicts of annotations is resolved by sorting the values and inserting a delimiter.  THIS TOOL IS TOTALLY UNSUPPORTED",
+        programGroup = CopyNumberProgramGroup.class)
+
+@BetaFeature
+public class MergeAnnotatedRegions extends GATKTool {
+    @Argument(
+            doc = "Input segment file or annotated segment file -- touching segments will be merged in the output.",
+            fullName = ExomeStandardArgumentDefinitions.SEGMENT_FILE_LONG_NAME,
```
samuelklee: Oops, guess I forgot to purge `ExomeStandardArgumentDefinitions`.  Can you delete and switch over to `CopyNumberStandardArgument`?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/TagGermlineEvents.java
```diff
@@ -0,0 +1,81 @@
...
+    @Argument(
+            doc = "Output TSV file identical to the tumor segment file, but with additional germline tag column (" + GERMLINE_TAG_HEADER + ").",
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+            fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME)
+    private File outputFile;
+
+    @Argument(
+            doc = "Amount of padding to allow a breakpoint match.",
+            shortName = "pbp",
+            fullName = "paddingInBp")
```
samuelklee: Extract `fullName` and kebab case.  Just use `padding` like we do elsewhere and instead specify that it is in bp in the `doc` string.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/TagGermlineEvents.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Argument(
+            doc = "Output TSV file identical to the tumor segment file, but with additional germline tag column (" + GERMLINE_TAG_HEADER + ").",
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+            fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME)
+    private File outputFile;
+
+    @Argument(
+            doc = "Amount of padding to allow a breakpoint match.",
+            shortName = "pbp",
```
samuelklee: Remove shortName.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/TagGermlineEvents.java
```diff
@@ -0,0 +1,81 @@
...
+            doc = "Matched normal called segment file.",
+            fullName = MATCHED_NORMAL_SEGMENT_FILE_LONG_NAME,
+            shortName = MATCHED_NORMAL_SEGMENT_FILE_SHORT_NAME
+    )
+    private File calledNormalSegmentFile;
+
+    @Argument(
+            doc = "Output TSV file identical to the tumor segment file, but with additional germline tag column (" + GERMLINE_TAG_HEADER + ").",
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+            fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME)
```
samuelklee: `fullName` should precede `shortName`.

src/main/java/org/broadinstitute/hellbender/utils/tsv/DataLine.java
```diff
@@ -428,7 +428,7 @@ public String get(final String columnName, final String defaultValue) {
     private int columnIndex(final String columnName) {
         final int index = columns.indexOf(columnName);
         if (index < 0) {
-            throw new IllegalArgumentException("there is no such a column: " + columnName);
+            throw new IllegalArgumentException("there is no such column: " + columnName);
```
samuelklee: Thanks for fixing this!

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
```diff
@@ -28,7 +25,7 @@
     public final static String START_HEADER = "START";
     public final static String END_HEADER = "END";
 
-    private final SimpleInterval interval;
+    private SimpleInterval interval;
```
samuelklee: Not in this diff, but `final public` -> `public final` above.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
```diff
@@ -85,23 +110,29 @@ public String toString() {
...
+     *  See {@link SimpleAnnotatedGenomicRegion#readAnnotatedRegions(File, Set)}, except that all columns (except those to define the interval)
+     *   are considered headers of interest.
      *
      * @param tsvRegionFile -- File containing tsv of genomic regions and annotations per line.  E.g. a segment file.
-     * @param headersOfInterest -- should not include any headers that are used to define the region (e.g. contig, start, end)
-     * @return annotated regions with one line in the input file for each entry of the list.  Never {@code null}
+     * @return annotated regions with one line in the input file for each entry of the list.  Never {<@UNVERIFIED|@code> null}
      */
-    public static List<SimpleAnnotatedGenomicRegion> readAnnotatedRegions(final File tsvRegionFile, final Set<String> headersOfInterest) {
+    public static List<SimpleAnnotatedGenomicRegion> readAnnotatedRegions(final File tsvRegionFile) {
```
samuelklee: Also vetoing this.  Put sequence dictionaries in your files if you want to obey the CNV collections format.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
```diff
@@ -113,35 +144,44 @@ protected SimpleAnnotatedGenomicRegion createRecord(final DataLine dataLine) {
...
+     *                          not include any headers that are used to define the region (e.g. contig, start, end)
+     * @return annotated regions with one line in the input file for each entry of the list.  Never {<@UNVERIFIED|@code> null}
      */
-    public static void writeAnnotatedRegionsAsTsv(final List<SimpleAnnotatedGenomicRegion> regions, final File outputFile) {
-
-        // Can't do Arrays.asList, since we will be adding to this list.
-        final List<String> columnHeaders = Lists.newArrayList(CONTIG_HEADER, START_HEADER, END_HEADER);
-        final List<String> otherHeaders = new ArrayList<>(regions.get(0).getAnnotations().keySet());
-        otherHeaders.sort(String::compareTo);
+    public static List<SimpleAnnotatedGenomicRegion> readAnnotatedRegions(final File tsvRegionFile, final Set<String> headersOfInterest) {
```
samuelklee: Same here (and this is essentially only used in test code, as well).

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
```diff
@@ -59,6 +56,34 @@ public int getEnd() {
         return interval.getEnd();
     }
 
+    public void setInterval(final Locatable interval) {
```
samuelklee: Why does this need to be mutable?  Perhaps I will discover why later, but I think most other records (if not all) in the CNV collections are essentially immutable.

Also, wouldn't it make more sense to parameterize as `SimpleAnnotatedGenomicRegion<Annotations>` (or something along those lines)?  Is there anything stopping me from trying to make a `SimpleAnnotatedGenomicRegionCollection` out of a list of `SimpleAnnotatedGenomicRegion`s that have maps with different keys?

I would guess that refactoring along these lines would simplify the code.  Let me read on and see.

samuelklee:
@LeeTL1220 Just noticed integration tests are failing...perhaps I should continue reviewing, and you can address comments and the tests at the same time?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/TagGermlineEvents.java
```diff
@@ -0,0 +1,81 @@
...
+import org.broadinstitute.hellbender.tools.copynumber.utils.annotatedregion.SimpleAnnotatedGenomicRegion;
+import org.broadinstitute.hellbender.tools.copynumber.utils.annotatedregion.SimpleAnnotatedGenomicRegionCollection;
+import org.broadinstitute.hellbender.tools.copynumber.utils.germlinetagging.SimpleGermlineTagger;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+
+@CommandLineProgramProperties(
+        oneLineSummary = "(EXPERIMENTAL) Do a simplistic tagging of germline events in a tumor segment file.",
```
samuelklee: `summary` should precede `oneLineSummary`, and neither should end in a period.  I think it's better to limit them both to one line if possible.  Also, statements like "The algorithm used is very simplistic" are probably not necessary in the summaries.  Further description can go in the Javadoc for the tools.

samuelklee: By the way, none of your tools have Javadoc or command-line examples.  @sooheelee Are we requiring this for experimental tools?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
```diff
@@ -59,6 +56,34 @@ public int getEnd() {
...
+        this.interval = new SimpleInterval(interval);
+    }
+
+    public void setEnd(final int end) {
```
samuelklee: Again, it seems unnecessary to have setters and make this mutable.  Also, `setInterval` is not actually used anywhere and `setAnnotation` unnecessarily returns a `String`.

To give you an idea of how idiosyncratic this is: If we merged this code, these would actually be the only setter methods in the entire CNV codebase!

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/TagGermlineEvents.java
LeeTL1220: @sooheelee I'm fine with writing docs and CLI examples.  These are only two tools.  I just started these so long ago that I forgot to add the docs and CLI examples later.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
LeeTL1220: I'm not even sure that I need the setters.  This might be forgotten leftovers.  I'll delete.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
samuelklee: After looking through the code, I think it needs to be refactored.  The collections framework isn't really used as intended---and there are many places where you actually modify the framework itself to accommodate your new collection.

As stated in the above comment, the use of a mutable map to represent each record in a TSV seems unnecessary and may be problematic if no checks are performed on the keys.  I think making `AnnotatedInterval` more flexible is the better approach.  Although the `AnnotationSet` that is used there currently only contains `GC_CONTENT`, we could perhaps modify it to contain a set of immutable annotations dynamically generated from input columns of interest.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegionCollection.java
```diff
@@ -0,0 +1,165 @@
...
+
+import static org.broadinstitute.hellbender.tools.copynumber.utils.annotatedregion.SimpleAnnotatedGenomicRegion.*;
+
+/**
+ * Represents a collection of annotated regions.  The annotations do not need to be known ahead of time, if reading from a file.
+ * Though {@link SimpleAnnotatedGenomicRegion::CONTIG_HEADER},
+ *  {@link SimpleAnnotatedGenomicRegion::START_HEADER}, and
+ *  {@link SimpleAnnotatedGenomicRegion::END_HEADER} are always expected for defining the genomic region.
+ */
+public class SimpleAnnotatedGenomicRegionCollection extends AbstractLocatableCollection<LocatableMetadata, SimpleAnnotatedGenomicRegion> {
```
samuelklee: This class looks very different from the other CNV collections classes.  All other classes are essentially an enum of column headers, codecs specified within the class (either statically or dynamically), and constructors that take either a file or metadata + list and simply call the superclass's constructor.  They contain no custom read/write methods, as these should all be handled by the superclass.

In contrast, your constructors include mandatory columns and codecs and you introduce several custom read/write methods.

I also think the use of Javadoc actually makes the class harder to read (and maintain).  These classes should be simple and self documenting.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/CombineSegmentBreakpoints.java
```diff
@@ -78,8 +80,11 @@
     @Override
     public void traverse() {
 
-        final List<SimpleAnnotatedGenomicRegion> segments1 = SimpleAnnotatedGenomicRegion.readAnnotatedRegions(segmentFiles.get(0), columnsOfInterest);
-        final List<SimpleAnnotatedGenomicRegion> segments2 = SimpleAnnotatedGenomicRegion.readAnnotatedRegions(segmentFiles.get(1), columnsOfInterest);
+        final SimpleAnnotatedGenomicRegionCollection simpleAnnotatedGenomicRegionCollection1 = SimpleAnnotatedGenomicRegionCollection.readAnnotatedRegions(segmentFiles.get(0), columnsOfInterest);
```
samuelklee: I think I made the same comment in a previous PR, but I think you will be able to use the collections framework much easier if you just specify the columns you want from each file separately, i.e., `columns-of-interest-for-first-input` and `columns-of-interest-for-second-input` or something along those lines.

You then create an AnnotatedIntervalCollection (after modifying that class appropriately) for each column set, dynamically generating codecs, and read in each corresponding file.  Incorrectly specified or missing columns will be handled by TableReader.

You then create an AnnotatedIntervalCollection for the combination of the two column sets.  You will have to take care to add suffixes when columns are duplicated.  You could extract this logic to some method that combines two AnnotatedIntervalCollections.

Your CLI should then be much simpler---just read in the files, create a new collection by calling the combine method, and then write the new collection.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/AbstractRecordCollection.java
LeeTL1220: Left in by accident.  Done

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/CombineSegmentBreakpoints.java
LeeTL1220: Done.  Also updated `ValidateBasicSomaticShortMutations`

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/MergeAnnotatedRegions.java
LeeTL1220: Done.  No short name there?  Seems like segment files could have a short name

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/TagGermlineEvents.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/TagGermlineEvents.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/TagGermlineEvents.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/tsv/DataLine.java
LeeTL1220: No problem.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
LeeTL1220: It's now required everywhere.

Done.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
LeeTL1220: This method was deleted.
Done

src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/CalledCopyRatioSegmentCollection.java
LeeTL1220: Left in by accident.  Done.

LeeTL1220: So how do I get the column header for a call?

LeeTL1220: Never mind.  Forcing user to specify.

Done.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/formats/collections/AbstractRecordCollection.java
LeeTL1220: Removed it.  I'll modify the validation pipeline.  I worry that this will be a problem later on, when we want to support comments and sequence dictionaries.

Done.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
LeeTL1220: I originally did the setters for performance reasons.  But I figured out a way around that.  So I was able to remove setters for the locatable portion.

The discordance of annotation names is still an issue.  I can remove the annotation setter, though.  That would make in-place adjustments impossible, though.  This would hurt performance, since we would be making a lot of copies.  I could also enforce uniform keys in the Collection class, but that might be messy.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/TagGermlineEvents.java
LeeTL1220: Also, these are not listed as `@DocumentedFeature`

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/CombineSegmentBreakpoints.java
LeeTL1220: I will look into Annotated Intervals and get back to you.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegionCollection.java
LeeTL1220: Let me look into AnnotatedInterval.  Then I might be able to get what I need there.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/annotatedregion/SimpleAnnotatedGenomicRegion.java
LeeTL1220: The setters are deleted.  `setAnnotation()` no longer returns a string.

Not quite done yet.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/utils/CombineSegmentBreakpoints.java
samuelklee: To be clear, you will have to make some modifications to AnnotatedIntervalCollection and its associated classes.

If you do things right, we could probably retool most of the other CNV collections to be AnnotatedIntervalCollections, since that is essentially what they are.  We can do this later, though.

LeeTL1220:
Closing in favor of upcoming PR that uses tribble, instead.

PR #4210, opened 2018-01-19
updating to htsjdk 2.14.1

lbergelson:
htsjdk 2.13.2 -> 2.14.1

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,7 +56,7 @@ repositories {
     mavenLocal()
 }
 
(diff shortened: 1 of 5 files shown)


cmnbroad:
@lbergelson The old File SequenceDictionaryExtractor methods are still there right ? If not we should upgrade picard as well.

lbergelson:
They're still there, just deprecated.

cmnbroad:
Looks like these were just deprecated, so they should be ok. :+1: when tests pass.

Issue #4204, opened 2018-01-18
Audit ReadClipper code (and clients) to make sure that it can handle empty reads


droazen:
There are two differences in GATK4 relative to GATK3 that are causing bugs in the read clipping code (see #3466 and #3845):

* `GATKRead` does not allow you to set mapped reads to a negative start position or a null contig, whereas the GATK3 clipping code allowed reads to enter such states.

* We return a `ReadUtils.emptyRead()` when we completely clip away a read, as opposed to (eg.,) a read with a negative start position and no bases like we did in GATK3.

We should audit all usages of the `ReadClipper` (as well as direct usages of `ReadUtils.emptyRead()` itself) to make sure that all client code can handle the return value of `ReadUtils.emptyRead()`, which returns an unmapped read with no bases. I've already audited the usages in `HaplotypeCaller`.

PR #4212, opened 2018-01-19
Added documentation warnings about empty clipped reads from readClipper where they can occur

jamesemery:
* Fixed an issue where OverhangFixingManger did not handle empty reads correctly resulting in an NPE
* Checked every codepath that could result in an empty read in readClipper and ensured that it handles empty reads.

Resolves: https://github.com/broadinstitute/gatk/issues/4204

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSpark.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.hellbender.tools.spark.transforms.markduplicates;
 
+import htsjdk.samtools.Defaults;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.metrics.MetricsFile;
 import org.apache.spark.Partitioner;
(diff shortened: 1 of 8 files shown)


droazen:
@jamesemery This is failing with a compiler error:

```
error: package org.testng.annotations does not exist
import org.testng.annotations.Test
```

jamesemery:
@droazen Fixed the error

src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSpark.java
```diff
@@ -106,6 +107,8 @@
 
         // Here we combine the original bam with the repartitioned unmarked readnames to produce our marked reads
         return sortedReadsForMarking.zipPartitions(repartitionedReadNames, (readsIter, readNamesIter)  -> {
+            System.out.println("system property says: " + System.getProperty("samjdk.compression_level"));
+            System.out.println("htsjdk says: " + Defaults.COMPRESSION_LEVEL);
```
droazen: @jamesemery These debug print statements seem to have snuck into your latest commit -- can you remove them?

jamesemery: whoops, yeah they are gone

jamesemery:
@droazen back to you

PR #4213, opened 2018-01-22
Update CommonSuffixSplitter.java

githublet:
preciseness


droazen:
Before we can accept a change like this, we need a good unit test to illustrate the problem that this patches fixes. The test should fail without the fix, and pass with it.

droazen:
Closing this one, as we can't accept a change like this without a good test to illustrate the problem.

PR #4214, opened 2018-01-22
*ON HOLD*: Update SharedSequenceMerger.java

githublet:
https://gatkforums.broadinstitute.org/gatk/discussion/11236/a-logical-problem-with-splitcommonsuffices-and-mergecommonsuffices#latest

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/SharedSequenceMerger.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/SharedSequenceMerger.java
@@ -65,7 +65,7 @@ public final class SharedSequenceMerger {
      * @return true if we can safely merge incomingVertices
      */
     private static boolean canMerge(final SeqGraph graph, final SeqVertex v, final Collection<SeqVertex> incomingVertices) {
-        if ( incomingVertices.isEmpty() ) {
+        if ( incomingVertices.size()<=1 ) { 
             return false;
         }
 


vruano:
Please keep this PR on hold until we can clarify what is trying to fix.

droazen:
Before we can accept a change like this, we need a good unit test to illustrate the problem that this patch fixes. The test should fail without the fix, and pass with it.

githublet:
@droazen @vruano 
change code just because in **CommonSuffixSplitter.java** some code like this:
                **for** ( final SeqVertex mid : **toSplit** ) {
                    // create my own copy of the suffix
                    final SeqVertex suffixV = new SeqVertex(**suffixVTemplate.getSequence()**);
                    graph.addVertex(suffixV);
so I think in **SharedSequenceMerger.java**,better merge vertices incomingVertices.size()>1

githublet:
I recode java to c (just haplotypecaller step 2), and will merge CommonSuffixSplitter.java SharedSequenceMerger.java in to one file, 
I'll just add vertex suffixVTemplate(graph.addVertex(suffixVTemplate)),then add in_edge to suffixVTemplate, last do something like this(this is not c code): final SeqVertex newV = new SeqVertex(ArrayUtils.addAll(suffixVTemplate.getSequence(), v.getSequence())).
for some test ,also the algorithm does,after process removePathsNotConnectedToRef(),
MergeTails() does not work!May be we can change MergeTails.java where logic like SharedSequenceMerger.java.

droazen:
Closing this one, since we still don't have a test illustrating the problem, and @vruano doesn't see the problem after considerable back-and-forth on the forum.

PR #4218, opened 2018-01-22
Python script executor updates required for NeuralNetInference branch.

cmnbroad:
These are changes we made in order to get the NeuralNetInference branch integration tests to pass, and some example program updates..

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
@@ -27,10 +27,13 @@ import java.util.concurrent.TimeUnit;
  *
  * <ol>
  * <li>Opens a FIFO for writing.</li>
- * <li>Creates an AsynchronousWriterServiceto allow writing to the FIFO in batches on a background thread</li>
+ * <li>Creates an AsynchronousWriterService to allow writing to the FIFO in batches on a background thread</li>
  * <li>Writes a string of attributes for each read to the List until the batchSize threshold is reached.</li>
  * <li>Uses Python to read each attribute line from the FIFO, and write it to the output file.</li>
  * </ol>
+ *
+ * See https://github.com/broadinstitute/gatk/wiki/Writing-GATK-Tools-that-use-Python for more information
+ * on using Python with GATK.
  */
 @CommandLineProgramProperties(
         summary = "Example/toy program that uses a Python script.",
@@ -78,6 +81,8 @@ public class ExampleStreamingPythonExecutor extends ReadWalker {
         } catch ( IOException e ) {
             throw new GATKException("Failure opening FIFO for writing", e);
         }
+        // synchronize on the output prompt before executing the next statement
+        pythonExecutor.getAccumulatedOutput();
 
         // Also, ask Python to open our output file, where it will write the contents of everything it reads
         // from the FIFO. <code sendSynchronousCommand/>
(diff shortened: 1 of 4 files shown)


src/test/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessControllerUnitTest.java
```diff
@@ -97,7 +97,8 @@ public void testMultipleParallelStreamingControllers() throws TimeoutException {
         Assert.assertFalse(teeController.getProcess().isAlive());
     }
 
-    @Test(groups = "python", timeOut = 10000)
+    // The timeout value here needs to exceed the timeout value used by StreamingProcessControllerUnitTest.
```
cmnbroad: This comment should say "...used by StreamingProcessController", not "StreamingProcessControllerUnitTest.".

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
```diff
@@ -78,6 +81,8 @@ public void onTraversalStart() {
         } catch ( IOException e ) {
             throw new GATKException("Failure opening FIFO for writing", e);
         }
+        // synchronize on the output prompt before executing the next statement
+        pythonExecutor.getAccumulatedOutput();
```
jonn-smith: Would you want to add a call to `getAccumulatedOutput` in `apply` as well?

jonn-smith:
@cmnbroad one quick question and I think the rest of the changes are OK here.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
cmnbroad: Thats an excellent question. Not in this case, but I see why you would think so. You only need to call getAccumulatedOutput and "clear out" the accumulated output when you're about to execute a synchronous command. But you have to know for sure that you've previously executed an asynchronous command, otherwise getAccumulatedOutput would time out. (Put another way, waitForPreviousBatchCompletion is idempotent, but getAccumulatedOutput is not). This tool calls `getAccumulatedOutput` once at the end in onTraversalSuccess, right before executing the next synchronous call, so it works.

But this example should illustrate what you're supposed to do, not what you can get away with...

In the NeuralNetInference tool, we added a boolean variable to address this - it keeps track of whether there is an outstanding command or not - and uses that to determine if its safe to make the call to getAccumulatedOutput. Thats what this example tool should do. I'm going to make another ticket for that change though, rather than holding up this PR, since we need these changes to get the NeuralNetInference branch merged, and I want to update some of the executor javadoc as part of that change as well.

cmnbroad: @jonn-smith I'll wait to hear if you're ok with the answer before I merge.

jonn-smith: @cmnbroad That sounds good.  Merge away.

I wonder if you can create a flag in the `pythonExecutor` itself that does the check for asynchronous output so that you don't need to pass around that boolean.

cmnbroad: Maybe we can talk that through some more tomorrow, as it would be great to find a way to simplify it.

PR #4223, opened 2018-01-22
make sv pipeline script copy_sv_results.sh able to take bucket path t…

SHuang-Broad:
make sv pipeline script copy_sv_results.sh able to take bucket path that starts with gs://

currently the result-copying step would fail if I run the script by providing the `GCS_SAVE_PATH` argument that starts with `gs://`.


SHuang-Broad:
@TedBrookings can you take a look? It wouldn't take long. Thanks!

Issue #4209, opened 2018-01-19
Conda environment creation fails when executed from the release distribution


byoo:
First of all, thank you so much for your hard work! I am excited to explore GATK 4.0.0.0.
I would like to report minor issues in setting up optional dependencies for GATK4.

First, `gatkcondaenv.yml` defines `gatkPythonPackageArchive.zip` location as
`build/gatkPythonPackageArchive.zip`
The path doesn't seem to work seamlessly with the release archive, though the path makes sense in the root directory of source repo. Unzipping the package file locates both files in the same directory and, to me, it is natural to create the conda environment in the directory.

Second, it would be more convenient to include `install_R_packages.R` in the release like gatkcondaenv.yml for python dependencies. Also, I had an issue running `install_R_packages.R` for fresh compiled R-3.2.5. The way `install_R_packages.R` install `optparse` prevents its dependency (the getopt package) from being automatically installed, because (I guess) repos argument is set to NULL. I think the helper script needs to install the getopt package as well.
```
optparseUrl="http://cran.r-project.org/src/contrib/Archive/optparse/optparse_1.3.2.tar.gz"
if (!("optparse" %in% rownames(installed.packages()))) {
  install.packages(optparseUrl, repos=NULL, type="source")
}
```

Finally, it would be convenient to depend on a R version available in conda, so that R dependency can be a part of the gatk conda environment. I wasn't able to get the recommended R-3.2.5 from conda.

droazen:
Thanks for raising these issues @byoo -- @cmnbroad would you be able to address them?

cmnbroad:
@byoo Thanks for reporting these issues. I'm going to move the R issue into a separate ticket, since its a bigger change than the python fix.

I'm not sure I understand this part, though:

> Unzipping the package file locates both files in the same directory and, to me, it is natural to create the conda environment in the directory.

There should be no need to manually unzip the package archive (maybe you did this as part of the workaround for bad path in the .yml ?). And you should be able to direct conda to create the env wherever you want it - we just suggested the simplest possible default case. Let me know if I'm misunderstanding your suggestion.

byoo:
@cmnbroad Sorry I should have been clearer. I meant unzipping the gatk release file. Unzipping the gatk release file places both files in the sample directory as below. I tried to create the gatk conda environment in the directory, which was failed due to the path issue.
```
$ unzip src/gatk-4.0.0.0.zip
Archive:  src/gatk-4.0.0.0.zip
[...]

PR #4233, opened 2018-01-23
Change conda environment creation to work with release distribution.

cmnbroad:
Fixes the Python part of https://github.com/broadinstitute/gatk/issues/4209. Also updated the README with more specific instructions based on previous feedback from another user. This was tested manually.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,6 +66,8 @@ final testNGVersion = '6.11'
 final baseJarName = 'gatk'
 final secondaryBaseJarName = 'hellbender'
 
+final pythonPackageArchiveName = 'gatkPythonPackageArchive.zip'
+
 configurations.all {
     resolutionStrategy {
         // the snapshot folder contains a dev version of guava, we don't want to use that.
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -455,9 +455,14 @@ task bundle(type: Zip) {
     from("gatk")
     from("README.md")
     from("build/docs/tabCompletion/gatk-completion.sh")
-    from("scripts/gatkcondaenv.yml")
-    from("$buildDir/gatkPythonPackageArchive.zip")
+    from("${buildDir}/${pythonPackageArchiveName}")
+    into(baseName)
```
lbergelson: Is this second `into(baseName)` necessary?  I'm not totally sure how the from/ into interacts I guess, but it looks like this might be an unintentional addition?

cmnbroad: It was intentional, but you're right that its not needed. I had previously been doing

`from("${buildDir}/${pythonPackageArchiveName}").filter(...)`

which corrupted the entire zip file. Using the `from` overload that takes the closure fixes that, but now I can remove the superfluous `into`.

lbergelson: I find this stuff very confusing in gradle...  I wish they had a big list of examples of using some of these things.

Issue #4188, opened 2018-01-17
M2 WDL script wishlist


sooheelee:
For @davidbenjamin. I've been asked to take the user-perspective and report things I'd want to make our WDL scripts more extensible. So here goes.

<img width="824" alt="screenshot 2018-01-17 13 22 27" src="https://user-images.githubusercontent.com/11543866/35060379-501cb8c8-fb8c-11e7-845e-a146fc2ced94.png">

## Major wants
- The is_bamout Boolian appears to be hardcoded to `false` in the script. Users need to be able to understand that this option can be changed without ambiguity. So this should become a proper  optional variable. @LeeTL1220 tells me this can be overwritten. However, why leave this as misinterpretable to newbie WDL-scriptors? Especially since `wdltools inputs` doesn't include it as a variable at all in the generated inputs list. Please can we make this a proper optional argument that `wdltools inputs` will generate a variable for.
- [ "${variants_for_contamination}" == *.vcf ] does not allow *.vcf.gz files. It should accept either.
- Outputs should allow either .vcf or .vcf.gz compression by user-specification. Alternatively, if we want to keep it simple and hardcode, then the preference is for compressed files. Some of us prefer to save on storage.
- Need to be able to specify optional string args for SplitIntervals. I would like to be able to use the BALANCING_WITHOUT_INTERVAL_SUBDIVISION mode. Furthermore, I'd like for the tool to automatically interpret this mode, when not given an -L intervals list, to not split reference contigs. I.e. a contig is an interval. (Perhaps already the tool behavior?)
- The version of Oncotator is not compatible with GRCh38. Please, can we have an option to switch this out with Funcotator? 

## Minor wants
- The JSON template in the repo should show the optional variables.
- Script calls for a Picard jar. I don't mind specifying this because I like controlling for the Picard version I use. However, users may want to call the Picard version within the GATK jar. I cannot fathom a simple way to allow switching this out in the script, but perhaps something like the gatk_override option could work. The goal would be to call the Picard tool from a Docker. This better enables provenance.

samuelklee:
Would appreciate similar insights on the CNV WDLs when you get a chance, @sooheelee!  Just one minor comment for @davidbenjamin: I think we agreed to remove the CNV JSON templates from the repo at some point, any reason why we can't do the same for the M2 template?  These templates don't buy you anything over just using `wdltools inputs`, but have the drawback of having to be kept up to date.

sooheelee:
[...]

PR #4238, opened 2018-01-23
Addressing a few of Soo Hee's requests for the Mutect2 wdl

davidbenjamin:
@sooheelee Once this is merged the remaining requests from issue #4188 will be compressed vcf output and Funcotator.  In this PR, the bamout becomes an *explicitly* optional `Boolean?` and `SplitIntervals` gets optional extra arguments, and we remove the template json as discussed with @samuelklee.  We already got rid of the picard jar by invoking picard tools from the gatk and enabled vcf.gz input for all resource files.

@takutosato This ought to be very quick.


davidbenjamin:
@sooheelee This PR also uses the matched normal for CalculateContamination as you mentioned later in #4188.

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -71,15 +72,19 @@ workflow Mutect2 {
     File? gnomad_index
     File? variants_for_contamination
     File? variants_for_contamination_index
-    Boolean is_run_orientation_bias_filter
+    Boolean? is_run_orientation_bias_filter
```
takutosato: While we're at it, can we rename all of the "is_xxx" variables? I prefer `orientation_bias_filter` to `is_run_orientation_bias_filter`, for example. If you think of a better name I'll support that too, though it's difficult since we have to avoid the name collision with `run_orientation_bias_filter.` Don't worry about it if you disagree. It's just a pet peeve of mine...

sooheelee:
Awesome. Thank you!

davidbenjamin:
> While we're at it, can we rename all of the "is_xxx" variables?

@takutosato Agreed.  I renamed them all.

sooheelee:
One question @davidbenjamin about SplitIntervals. If I use a mode that does not SUBDIVIDE, sans intervals list, then will it create an intervals list of unsplit-contigs? If so, I'd like to clarify this in the tooldoc. If not, can I get this feature? Thanks.

davidbenjamin:
@sooheelee That's right -- it implicitly creates an interval list of one interval per contig and then scatters into sublists containing unsplit contigs.  If the scatter count is greater than the number of contig if will use a smaller scatter count.  

For example, if your reference contains contigs 1, 2, 3, and 4 and you set `scatter = 5` you will get four lists, each with one interval containing one contig.  If you set `scatter = 3` you will get three interval lists, one of which has two contig intervals.

sooheelee:
Cool. Thanks for the clarification!

Issue #4224, opened 2018-01-22
Tribble segfault reading .vcf.gz without index


ldgauthier:
My struggles with indexes in GATK4 continue.  I forgot to pull down the corresponding *.tbi index for my .vcf.gz, but SelectVariants just toodled along until chr14:
```
htsjdk.tribble.TribbleException: Line 3889836: there aren't enough columns for line chr14       24737838        rs1101636       C       T       .       PASS    AC=2;AF=1.00;AN=2;DB;DP=38;ExcessHet=0.7420;FS=0.000;InbreedingCoeff=0.0357;MQ=59.98;MQRankSum=0.026;MQ_DP=24332;POSITIVE_TRAIN_SITE;QD=22.41;QUALapprox=537 (we expected 9 tokens, and saw 8 ), for input source: file:///humgen/gsa-hpprojects/dev/gauthier/reblockGVCF/gnomADaccuracyTest.noMQinSNPVQSR.SynDip.vcf.gz
        at htsjdk.variant.vcf.AbstractVCFCodec.decodeLine(AbstractVCFCodec.java:281)
        at htsjdk.variant.vcf.AbstractVCFCodec.decode(AbstractVCFCodec.java:262)
        at htsjdk.variant.vcf.AbstractVCFCodec.decode(AbstractVCFCodec.java:64)
        at htsjdk.tribble.AsciiFeatureCodec.decode(AsciiFeatureCodec.java:70)
        at htsjdk.tribble.AsciiFeatureCodec.decode(AsciiFeatureCodec.java:37)
        at htsjdk.tribble.TribbleIndexedFeatureReader$WFIterator.readNextRecord(TribbleIndexedFeatureReader.java:365)
        at htsjdk.tribble.TribbleIndexedFeatureReader$WFIterator.next(TribbleIndexedFeatureReader.java:346)
        at htsjdk.tribble.TribbleIndexedFeatureReader$WFIterator.next(TribbleIndexedFeatureReader.java:307)
        at java.util.Iterator.forEachRemaining(Iterator.java:116)
        at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
[...]

PR #4240, opened 2018-01-23
Disabling unindexed tabix files as input to FeatureDataSource

jamesemery:
Relates to https://github.com/broadinstitute/gatk/issues/4224


src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
```diff
@@ -253,6 +253,9 @@ public FeatureDataSource(final FeatureInput<T> featureInput, final int queryLook
         } else {
             throw new GATKException("Found a feature input that was neither GenomicsDB or a Tribble AbstractFeatureReader.  Input was " + featureInput.toString() + ".");
         }
+        if (!hasIndex && TribbleIndexedFeatureReader.hasBlockCompressedExtension(featureInput.getFeaturePath())) {
+            throw new GATKException("Found a feature input that was block compressed but lacked an index file.  Input was " + featureInput.toString() + ".");
```
magicDGS: I think that this is a temporary fix, no? If so, maybe it is worth to point that this restriction might be removed in the future.

In addition, shouldn't it be an `UserException` if the input should be indexed by the user?

src/test/java/org/broadinstitute/hellbender/engine/FeatureDataSourceUnitTest.java
```diff
@@ -71,6 +72,13 @@ public void testGetSequenceDictionary() {
         }
     }
 
+    @Test (expectedExceptions = GATKException.class)
+    // this test asserts that a helpful exception is thrown for blockZipped files lacking an index as they may not be fully supported
+    //TODO this is a temporary fix until https://github.com/broadinstitute/gatk/issues/4224 has been resolved
```
magicDGS: Maybe it is a good idea to add a test that it's failing with unindexed bgzip files, providing wrong results or parsing errors. That test should be disabled and used to test if the issue is solved. Just an idea...

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
droazen: I agree with @magicDGS on both points. This should be a `UserException.MissingIndex` (which you should create), and include a note that "The requirement that all block-compressed inputs have an index may be relaxed in the future." The error message should also mention that the tool `IndexFeatureFile` can be used to create the necessary index.

droazen: You should also add a comment above this if statement with a link to the relevant bug on github, and a brief explanation of why this constraint exists.

src/test/java/org/broadinstitute/hellbender/engine/FeatureDataSourceUnitTest.java
```diff
@@ -71,6 +72,13 @@ public void testGetSequenceDictionary() {
         }
     }
 
+    @Test (expectedExceptions = GATKException.class)
```
droazen: `UserException.MissingIndex`

src/test/java/org/broadinstitute/hellbender/engine/FeatureDataSourceUnitTest.java
```diff
@@ -71,6 +72,13 @@ public void testGetSequenceDictionary() {
...
+    // this test asserts that a helpful exception is thrown for blockZipped files lacking an index as they may not be fully supported
+    //TODO this is a temporary fix until https://github.com/broadinstitute/gatk/issues/4224 has been resolved
+    public void testUnindexedBZippedFile() {
+        new FeatureDataSource<>(new File(toolsTestDir + "IndexFeatureFile/4featuresHG38Header.unindexed.vcf.gz"));
```
droazen: In addition to this unit test, add an integration test to `FeatureSupportIntegrationTest` as well.

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
```diff
@@ -1,26 +1,37 @@
...
 import org.broadinstitute.hellbender.engine.FeatureDataSource;
+import org.broadinstitute.hellbender.engine.FeatureInput;
+import org.broadinstitute.hellbender.engine.FeatureManager;
 import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
+import org.broadinstitute.hellbender.utils.io.IOUtils;
 import org.broadinstitute.hellbender.utils.test.ArgumentsBuilder;
 import org.broadinstitute.hellbender.utils.test.VariantContextTestUtils;
+import org.broadinstitute.hellbender.utils.test.VariantContextTestUtilsUnitTest;
```
droazen: Clean up unused imports

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
```diff
@@ -38,36 +49,40 @@
...
         final ArgumentsBuilder args = new ArgumentsBuilder();
         inputs.forEach(args::addInput);
         args.addOutput(output)
                 .addArgument(GatherVcfsCloud.GATHER_TYPE_LONG_NAME, gatherType.toString());
         runCommandLine(args);
 
-        try (  final FeatureDataSource<VariantContext> actualReader = new FeatureDataSource<>(output);
-               final FeatureDataSource<VariantContext> expectedReader = new FeatureDataSource<>(expected);){
+        try (final AbstractFeatureReader<? extends Feature, ?> actualReader = AbstractFeatureReader.getFeatureReader(output.getAbsolutePath(), null, FeatureManager.getCodecForFile(output, VariantContext.class), false, Function.identity(), Function.identity());
+             final AbstractFeatureReader<? extends Feature, ?> expectedReader = AbstractFeatureReader.getFeatureReader(expected.getAbsolutePath(), null, FeatureManager.getCodecForFile(expected, VariantContext.class), false, Function.identity(), Function.identity())){
```
droazen: Type these as `FeatureReader<VariantContext>` instead. This should allow you to remove the gross typecasts below.

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
```diff
@@ -38,36 +49,40 @@
...
 
-            final List<VariantContext> actualVariants = Lists.newArrayList(actualReader);
-            final List<VariantContext> expectedVariants = Lists.newArrayList(expectedReader);
+
+            final List<VariantContext> actualVariants = new ArrayList<>();
+            ((Iterator<VariantContext>)actualReader.iterator()).forEachRemaining(actualVariants::add);
+            final List<VariantContext> expectedVariants = new ArrayList<>();
+            ((Iterator<VariantContext>)expectedReader.iterator()).forEachRemaining(expectedVariants::add);
```
droazen: You should be able to use `Lists.newArrayList()` to create these lists directly instead of doing this, provided you change the types of the readers as suggested above.

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
```diff
@@ -92,7 +107,7 @@ private void assertGatherProducesCorrectVariants(GatherVcfsCloud.GatherType gath
     }
 
     @Test(dataProvider = "getHeaderOnlyCases")
-    public void testVcfsWithOnlyAHeader(List<File> inputs, final GatherVcfsCloud.GatherType gatherType, File expected){
+    public void testVcfsWithOnlyAHeader(List<File> inputs, final GatherVcfsCloud.GatherType gatherType, File expected) throws IOException {
         assertGatherProducesCorrectVariants(gatherType, expected, inputs);
     }
```
droazen: Add a test for `IndexFeatureFile` proving that the tool can still index an unindexed `.vcf.gz` (if we don't already have a good test that does that)

src/test/java/org/broadinstitute/hellbender/engine/FeatureDataSourceUnitTest.java
jamesemery: Unfortunately implementing that test has problems two-fold, the first is that I can't assert this problem anymore now that I have restricted featureDataSource (well maybe not technically but functionally), and the second is that it's difficult to trigger this for small files. I will put some effort into making some files that trigger the bug, but its inconsistent between machines so I couldn't really enforce much.

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
jamesemery: done

src/test/java/org/broadinstitute/hellbender/engine/FeatureDataSourceUnitTest.java
jamesemery: done

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
jamesemery: done

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
jamesemery: `IndexFeatureFileIntegrationTest.testVCFGZIndex_tabix()`

src/test/java/org/broadinstitute/hellbender/engine/FeatureDataSourceUnitTest.java
jamesemery: done

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
jamesemery: I don't think I can type the feature readers that way without casting them by using AbstractFeatureReader.

jamesemery:
@droazen Responded to comments

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
```diff
@@ -253,6 +253,11 @@ public FeatureDataSource(final FeatureInput<T> featureInput, final int queryLook
         } else {
             throw new GATKException("Found a feature input that was neither GenomicsDB or a Tribble AbstractFeatureReader.  Input was " + featureInput.toString() + ".");
         }
+        // Due to a bug in HTSJDK, unindexed block compressed input files may fail to parse completely. For safety,
+        // these files have been disabled. See https://github.com/broadinstitute/gatk/issues/4224 for discussion
+        if (!hasIndex && TribbleIndexedFeatureReader.hasBlockCompressedExtension(featureInput.getFeaturePath())) {
+            throw new UserException.MissingIndex("Found a feature input that was block compressed but lacked an index file, support for unindexed feature files has temporarily been removed.  Input was " + featureInput.toString() + ".");
```
droazen: Include instructions on how to index the file in the error message ("You can use the IndexFeatureFile tool to index this file")

src/main/java/org/broadinstitute/hellbender/exceptions/UserException.java
```diff
@@ -103,6 +103,12 @@ public CouldNotReadInputFile(String message) {
         public MissingReference(String message) { super(message); }
     }
 
+    public static class MissingIndex extends UserException {
+        private static final long serialVersionUID = 0L;
+
+        public MissingIndex(String message) { super(message); }
```
droazen: When adding a new exception subclass of `UserException`, it's good practice to insert a standard error message in the constructor. Eg.,

```
public MissingIndex(String file, String message) {
    super(String.format("An index is required but was not found for file %s. %s", file, message)); 
}
```

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
```diff
@@ -31,47 +35,49 @@
...
 
-        try (  final FeatureDataSource<VariantContext> actualReader = new FeatureDataSource<>(output);
-               final FeatureDataSource<VariantContext> expectedReader = new FeatureDataSource<>(expected);){
+        try (final AbstractFeatureReader<VariantContext, ?> actualReader = AbstractFeatureReader.getFeatureReader(output.getAbsolutePath(), null, new VCFCodec(), false, Function.identity(), Function.identity());
+             final AbstractFeatureReader<VariantContext, ?> expectedReader = AbstractFeatureReader.getFeatureReader(expected.getAbsolutePath(), null, new VCFCodec(), false, Function.identity(), Function.identity())) {
 
-            final List<VariantContext> actualVariants = Lists.newArrayList(actualReader);
-            final List<VariantContext> expectedVariants = Lists.newArrayList(expectedReader);
+            final List<VariantContext> actualVariants = new ArrayList<>();
+            (actualReader.iterator()).forEachRemaining(actualVariants::add);
```
droazen: You should be able to do `final List<VariantContext> actualVariants = Lists.newArrayList(actualReader);` here (and for `expectedVariants` as well).

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
```diff
@@ -143,7 +149,7 @@ public void testBlockGather(final File vcf, final int numShards) throws IOExcept
                 VariantContextTestUtils.assertEqualVariants(actual, expected);
 
                 Assert.assertEquals(((VCFHeader) outputDataSource.getHeader()).getMetaDataInInputOrder(),
-                                    ((VCFHeader) input.getHeader()).getMetaDataInInputOrder());
+                        ((VCFHeader) input.getHeader()).getMetaDataInInputOrder());
```
droazen: Your IDE seems to have reformatted a lot of the whitespace in this file...you should try to avoid introducing whitespace changes into your diffs unless they're to fix actual formatting issues.

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
jamesemery: done

src/main/java/org/broadinstitute/hellbender/exceptions/UserException.java
jamesemery: done

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
jamesemery: Yea, that was accidental, though i'm reasonably sure all of those cases involved reformatting whitespace from tabs to spaces which is a good thing.

jamesemery:
@droazen responded

src/main/java/org/broadinstitute/hellbender/exceptions/UserException.java
```diff
@@ -103,6 +103,16 @@ public CouldNotReadInputFile(String message) {
...
     }
 
+    public static class MissingIndex extends UserException {
+        private static final long serialVersionUID = 0L;
+
+        public MissingIndex(String file, String message) {
+            super(String.format("An index is required but was not found for file %s. %s", file, message));
+        }
+
+        public MissingIndex(String message) { super(message); }
```
droazen: Use the constructor with the baked-in error text, and delete this constructor

src/test/java/org/broadinstitute/hellbender/tools/GatherVcfsCloudIntegrationTest.java
```diff
@@ -31,47 +35,48 @@
...
 
-        try (  final FeatureDataSource<VariantContext> actualReader = new FeatureDataSource<>(output);
-               final FeatureDataSource<VariantContext> expectedReader = new FeatureDataSource<>(expected);){
+        try (final AbstractFeatureReader<VariantContext, ?> actualReader = AbstractFeatureReader.getFeatureReader(output.getAbsolutePath(), null, new VCFCodec(), false, Function.identity(), Function.identity());
+             final AbstractFeatureReader<VariantContext, ?> expectedReader = AbstractFeatureReader.getFeatureReader(expected.getAbsolutePath(), null, new VCFCodec(), false, Function.identity(), Function.identity())) {
 
-            final List<VariantContext> actualVariants = Lists.newArrayList(actualReader);
-            final List<VariantContext> expectedVariants = Lists.newArrayList(expectedReader);
+            final List<VariantContext> actualVariants = Lists.newArrayList((Iterable<? extends VariantContext>) actualReader);
+            final List<VariantContext> expectedVariants = Lists.newArrayList((Iterable<? extends VariantContext>) actualReader);
```
droazen: Looks like you've accidentally clobbered this test -- now you're comparing the actual output against itself.

Also, given that `AbstractFeatureReader` does not implement `Iterable`, you should instead call the overload that takes an `Iterator` (by passing in `reader.iterator()`), and remove the typecast.

PR #4247, opened 2018-01-24
Clean up and improve re-usability of help classes

magicDGS:
There are several changes in this PR:

* Re-use Picard's supercategory map and supercategory constants. This will keep in line Picard an GATK supercategories without the need of modifying the code in `HelpConstants`
* Re-use Picard's Testing program group. The `TestProgramGroup` is a duplicate of Picard's `Testing` program group. This duplicated also some constants from the Picard's code. Again, this will maintain in sync GATK.
* Remove the unused **"RNA-Specific Tools"** category constants, which was unused.
* Make public the `getSuperCategoryMap` as in Picard's code, to allow downstream projects to re-use supercategory assignation in their own documentation, and include their own.


magicDGS:
I don't think that there is any breaking change here. Please, could you review the code here, @cmnbroad?

Maybe also someone from the communications team can also have a look to check if everything is alright from the GATK's documentation side. @sooheelee / @vdauwera ?

Thank you!

cmnbroad:
@magigDGS There are 4 separate issues in here, and I have slightly different feelings about each of them. Some comments:

- Removing the RNA string seems fine.
- I deliberately left the GATK test program group in because even though there is a Picard one, its harmless, and easier for people to find.
- I have reservations about exposing and sharing the super category maps. The Picard docgen process is pretty much unused at this point. The GATK supercategory map "truth" is, and should be, defined by GATK.
- I'm reluctant to make the GATK `getSuperCategoryMap` public, because I don't think it can have any kind of useful contract. Its tied to the GATK doc templates and doc process, which we need to be able to change freely. It seems much safer for ReadTools to define it's own set categories (the only overlap would probably be something like ReadFilters, right ?).

magicDGS:
At the moment, ReadTools only documents and use `ReadFilters` but I am planning to probably add pack a couple of tools from the GATK/Picard tools at some point. In addition, I am working on another toolkit based on the GATK code, and it will include also annotations for variants (and probably some VCF tools).

I just thought that it will be useful to been able to pull out the super-category map to re-use the GATK docgen code. If a downstream project with extra-categories wants to use the GATK templates and DocGen code might get into troubles without being able to access that. I am still working on how to document better my toolkits, but it is not a problem yet.

Anyway, I don't really have any strong feeling about this; I just wanted to reduce a bit the complexity of the GATK code and do the same with my downstream projects. If it is something that you anticipate that it is going to change the contract often, feel free to close (the RNA Strings can be removed in other PR).

cmnbroad:
Closing per discussion above.

PR #4259, opened 2018-01-25
Update variable names in mutect2_pon.wdl

byoo:
Update variable names in mutect2_pon.wdl to reflect changes in mutect2.wdl


davidbenjamin:
At some point we deleted the lines in question because the variables are `false` by default.

PR #4263, opened 2018-01-25
Exposed bounds for copy-neutral segments in CallCopyRatioSegments.

samuelklee:
Although identification of the copy-neutral state is still relatively manual, this will at least make the tool usable for samples with high ploidy until @MartonKN finishes up the new caller.


LeeTL1220:
@samuelklee Looks reasonable to me.  Feel free to merge.

samuelklee:
Thanks, @LeeTL1220!

PR #4271, opened 2018-01-26
Compressed vcf output and Funcotator options in Mutect2 wdl

### LINKED ISSUE #4188: M2 WDL script wishlist

davidbenjamin:
Closes #4188.

@sooheelee This wraps up the wish list in #4188.

@takutosato Although funcotator is off in the Travis test, I *did* test it locally. Also, I have tested with and without compressed vcf output.  The funcotator command is simply copied from @jonn-smith's funcotator.wdl.  Subworkflows are a pain in Firecloud, which is why I don't `import` it.

### DIFF
--- a/scripts/m2_cromwell_tests/test_m2_wdl_multi.json
+++ b/scripts/m2_cromwell_tests/test_m2_wdl_multi.json
@@ -9,6 +9,7 @@
   "Mutect2_Multi.scatter_count": 2,
   "Mutect2_Multi.run_orientation_bias_filter": true,
   "Mutect2_Multi.run_oncotator": true,
+  "Mutect2_Multi.run_funcotator": false,
   "Mutect2_Multi.preemptible_attempts": 2,
   "Mutect2_Multi.artifact_modes": ["G/T", "C/T"],
   "Mutect2_Multi.compress_vcfs": false,
--- a/scripts/m2_cromwell_tests/test_m2_wdl_multi.json
+++ b/scripts/m2_cromwell_tests/test_m2_wdl_multi.json
@@ -7,11 +7,10 @@
   "Mutect2_Multi.ref_dict": "/home/travis/build/broadinstitute/gatk/src/test/resources/large/human_g1k_v37.20.21.dict",
   "Mutect2_Multi.pair_list": "/home/travis/build/broadinstitute/gatk/scripts/m2_cromwell_tests/pair_list",
   "Mutect2_Multi.scatter_count": 2,
-  "Mutect2_Multi.dbsnp": "/home/travis/build/broadinstitute/gatk/src/test/resources/large/dbsnp_138.b37.20.21.vcf",
-  "Mutect2_Multi.dbsnp_index": "/home/travis/build/broadinstitute/gatk/src/test/resources/large/dbsnp_138.b37.20.21.vcf.idx",
-  "Mutect2_Multi.is_run_orientation_bias_filter": true,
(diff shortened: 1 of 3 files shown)


samuelklee:
@davidbenjamin What is the issue with subworkflows?  We import some subworkflows in the CNV WDLs.  I think sub-subworkflows are when things get hairy...

@jonn-smith @LeeTL1220 Let me know at what point we should switch over to Funcotator (or make it an option) in the CNV WDLs as well.  I'd also feel more comfortable if the Funcotator/Oncotator tests actually ran on Travis.

davidbenjamin:
@samuelklee It's quite possible I just plain misunderstood about subworkflows.  There may still be something to be said for making the M2 wdl totally self-contained, but that's a call for Lee and Geraldine.

LeeTL1220:
@davidbenjamin Oncotator runs in the M2 WDL travis tests (last I checked).  It just uses no datasources.  Oncotator is disabled for the CNV WDL travis tests.  I did this because the CNV Oncotator uses a different docker image, which is fairly large.

@davidbenjamin @samuelklee Life is only a little bit easier if the WDL is self-contained.    The sub-...-workflows gets a little hairy for setting optional parameters (particularly runtime parameters).

davidbenjamin:
@LeeTL1220 So what's the verdict -- import the funcotator workflow or make the M2 wdl self-contained with a Funcotator task?  If the former, what do I need to do to un-break M2 in Firecloud?

LeeTL1220:
@davidbenjamin I would make the Funcotator inline in the M2 WDL for now.  As sub-workflows get easier to manage, maybe we re-assess.  I'm punting on CNV since Funcotator does not support CNV yet.

davidbenjamin:
@takutosato Lee's offline verdict was to keep Funcotator inlined in the M2 wdl for now, so this PR can stay as-is.

Issue #4275, opened 2018-01-26
prevent running HaplotypeCallerSpark with broken output formats


lbergelson:
Currently trying to create a `g.vcf.gz` with HaplotypeCallerSpark silently produces a vcf without blocking.  See #4274  for more details.

We should temporarily prevent this from happening until we can fix the underlying issue, as well as identifying any other bad cases.

Issue #4274, opened 2018-01-26
HaplotypeCallerSpark doesn't write g.vcf.gz files


lbergelson:
If you ask HaplotypeCallerSpark for a gvcf.gz it outputs a base pair resolution GVCF with no blocking.  This is due to confusion in hadoop-bam / VariantSparkSink.

It works fine if you write an uncompressed g.vcf.

This is due to a conditional statement in `KeyIgnoringVCFOutputFormat.getRecordWriter(askAttemptContext ctx)`

```
		if (!isCompressed) {
			return getRecordWriter(ctx, file);
		} else {
			FileSystem fs = file.getFileSystem(conf);
			return getRecordWriter(ctx, codec.createOutputStream(fs.create(file)));
		}
```

The two branches call two different overloads of `getRecordWriter`

```
getRecordWriter(TaskAttemptContext ctx, Path out)

getRecordWriter(TaskAttemptContext ctx, OutputStream outputStream)
```

The first is public, and overriden to provide GVCF writers in our code, the second is private and doesn't know about our GVCF writer.   We could override `getRecordWriter(ctx)` but we need access to a constructor for `VCFRecordWriter` that takes a stream and propagates the ctx which doesn't exist.

lbergelson:
see #4275 for the temporary workaround

lbergelson:
This will likely require a fix in hadoop-bam unless we either copy most of the the hadoop-bam code into gatk or someone comes up with a more clever solution than I.

lbergelson:
similarly, see #4303

droazen:
@tomwhite Should be a fairly easy fix in Hadoop-BAM, we think.

PR #4277, opened 2018-01-26
prevent users from requesting g.vcf.gz in Spark

lbergelson:
* this is currently broken, see https://github.com/broadinstitute/gatk/issues/4274
* add a check to HaplotypeCallerSpark and VariantSparkSink and throw a clear exception in this case
* added test for GVCF writing in VariantSparkSink which previously didn't exist
* added new UserException.UnimplementedFeature class
* closes https://github.com/broadinstitute/gatk/issues/4275

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSink.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSink.java
@@ -2,8 +2,8 @@ package org.broadinstitute.hellbender.engine.spark.datasources;
 
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
-import htsjdk.samtools.SAMFileHeader;
-import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.util.IOUtil;
+import htsjdk.tribble.AbstractFeatureReader;
 import htsjdk.variant.variantcontext.VariantContext;
 import htsjdk.variant.variantcontext.writer.VariantContextWriter;
 import htsjdk.variant.vcf.VCFHeader;
@@ -21,6 +21,7 @@ import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.apache.spark.broadcast.Broadcast;
+import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.gcs.BucketUtils;
 import org.broadinstitute.hellbender.utils.variant.HomoSapiensConstants;
 import org.broadinstitute.hellbender.utils.variant.writers.GVCFWriter;
@@ -29,7 +30,6 @@ import org.seqdoop.hadoop_bam.util.BGZFCodec;
 import org.seqdoop.hadoop_bam.util.VCFFileMerger;
 import scala.Tuple2;
 
-import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Comparator;
@@ -100,7 +100,7 @@ public final class VariantsSparkSink {
                     WRITE_HEADER_PROPERTY, true);
 
             switch (format) {
-                case BCF: return new KeyIgnoringBCFRecordWriter<NullWritable>(out,header,wh,ctx);
+                case BCF: return new KeyIgnoringBCFRecordWriter<NullWritable>(out, header, wh, ctx);
                 case VCF: return new GvcfKeyIgnoringVCFRecordWriter<NullWritable>(out,header,wh,ctx);
                 default: throw new IllegalStateException("Unrecognized variant format: " + format);
             }
@@ -178,6 +178,12 @@ public final class VariantsSparkSink {
             final VCFHeader header, final boolean writeGvcf, final List<Integer> gqPartitions, final int defaultPloidy, final int numReducers) throws IOException {
 
         final Configuration conf = ctx.hadoopConfiguration();
+
+        //TODO remove me when https://github.com/broadinstitute/gatk/issues/4274 is fixed
+        if (writeGvcf && (AbstractFeatureReader.hasBlockCompressedExtension(outputFile) || outputFile.endsWith(IOUtil.BCF_FILE_EXTENSION))) {
+            throw new UserException.UnimplementedFeature("It is currently not possible to write a compressed g.vcf or bcf.gz on spark.  See https://github.com/broadinstitute/gatk/issues/4274 for more details.");
+        }
+
         if (outputFile.endsWith(BGZFCodec.DEFAULT_EXTENSION) || outputFile.endsWith(".gz")) {
             conf.setBoolean(FileOutputFormat.COMPRESS, true);
             conf.setClass(FileOutputFormat.COMPRESS_CODEC, BGZFCodec.class, CompressionCodec.class);
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSink.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSink.java
@@ -179,9 +179,9 @@ public final class VariantsSparkSink {
 
         final Configuration conf = ctx.hadoopConfiguration();
 
-        //TODO remove me when https://github.com/broadinstitute/gatk/issues/4274 is fixed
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSink.java
```diff
@@ -178,6 +178,12 @@ private static void writeVariantsSingle(
             final VCFHeader header, final boolean writeGvcf, final List<Integer> gqPartitions, final int defaultPloidy, final int numReducers) throws IOException {
 
         final Configuration conf = ctx.hadoopConfiguration();
+
+        //TODO remove me when https://github.com/broadinstitute/gatk/issues/4274 is fixed
+        if (writeGvcf && (AbstractFeatureReader.hasBlockCompressedExtension(outputFile) || outputFile.endsWith(IOUtil.BCF_FILE_EXTENSION))) {
+            throw new UserException.UnimplementedFeature("It is currently not possible to write a compressed g.vcf or bcf.gz on spark.  See https://github.com/broadinstitute/gatk/issues/4274 for more details.");
```
droazen: Doesn't this forbid writing any BCFs when in GVCF mode (not just `bcf.gz`)? Was this the intent? Both the error message and the ticket https://github.com/broadinstitute/gatk/issues/4274 seem incomplete, since they don't mention disallowing `g.bcf`

src/main/java/org/broadinstitute/hellbender/exceptions/UserException.java
```diff
@@ -427,4 +427,12 @@ public CouldNotIndexFile(final File file, final Exception e) {
                     file.getAbsolutePath(), e.getClass().getCanonicalName(), e.getMessage()), e);
         }
     }
+
+    public static final class UnimplementedFeature extends UserException {
+        private static final long serialVersionUID = 0L;
+
+        public UnimplementedFeature(String message){
+            super(message);
```
droazen: New `UserException` subclasses should generally prepend some boilerplate text to the error message in their constructors (eg, "Feature is not currently implemented: %s")

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -133,6 +135,13 @@ public boolean requiresReference(){
 
     @Override
     protected void runTool(final JavaSparkContext ctx) {
+        //TODO remove me when https://github.com/broadinstitute/gatk/issues/4274 is fixed
+        if(hcArgs.emitReferenceConfidence == ReferenceConfidenceMode.GVCF
+                && (AbstractFeatureReader.hasBlockCompressedExtension(output) || output.endsWith(IOUtil.BCF_FILE_EXTENSION))) {
+            throw new UserException.UnimplementedFeature("It is currently not possible to write a compressed g.vcf or bcf.gz from HaplotypeCallerSpark.  " +
+                                            "See https://github.com/broadinstitute/gatk/issues/4274 for more details.");
```
droazen: Are you repeating this check so that the tool can fail early, or would the check in `VariantSparkSink` not catch this?

Either way, my comment above also applies here: `Doesn't this forbid writing any BCFs when in GVCF mode (not just bcf.gz)? Was this the intent? Both the error message and the ticket #4274 seem incomplete, since they don't mention disallowing g.bcf`

src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSinkUnitTest.java
```diff
@@ -84,6 +94,72 @@ public void testWritingToFileURL(String vcf, String outputFileExtension) throws
         assertSingleShardedWritingWorks(vcf, outputUrl);
     }
 
+    @Test(expectedExceptions = UserException.UnimplementedFeature.class)
+    public void testGVCF_GZ_isDisallowed() throws IOException {
+        JavaSparkContext ctx = SparkContextFactory.getTestSparkContext();
+        VariantsSparkSink.writeVariants(ctx, createTempFile("test",".g.vcf.gz").toString(), null,
```
droazen: Also test that `.g.bcf` and `.g.bcf.gz` are disallowed.

src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSinkUnitTest.java
```diff
@@ -84,6 +94,72 @@ public void testWritingToFileURL(String vcf, String outputFileExtension) throws
...
+        final JavaSparkContext ctx = SparkContextFactory.getTestSparkContext();
+        final File output = createTempFile(outputFileName, extension);
+        VariantsSparkSink.writeVariants(ctx, output.toString(), ctx.parallelize(vcs), getHeader(), writeGvcf, Arrays.asList(100), 2, 1 );
+
+        new CommandLineProgramTest(){
+            @Override
+            public String getTestedToolName(){
+                return IndexFeatureFile.class.getSimpleName();
+            }
+        }.runCommandLine(new String[]{"-F", output.getAbsolutePath()});
```
droazen: Isn't `new Main().instanceMain(args)` simpler here?

src/test/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSparkIntegrationTest.java
```diff
@@ -127,6 +127,18 @@ public void testGVCFModeIsConcordantWithGATK3_8Results() throws Exception {
         Assert.assertTrue(concordance >= 0.99, "Concordance with GATK 3.8 in GVCF mode is < 99% (" +  concordance + ")");
     }
 
+    @Test(expectedExceptions = UserException.UnimplementedFeature.class)
+    public void testGVCF_GZ_isDisallowed() {
+
+        final String[] args = {
+                "-I", NA12878_20_21_WGS_bam,
+                "-R", b37_2bit_reference_20_21,
+                "-O", createTempFile("testGVCF_GZ_throw_exception", ".g.vcf.gz").getAbsolutePath(),
```
droazen: Also test that `.g.bcf` and `.g.bcf.gz` are disabled.

src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSink.java
lbergelson: That was the intent, it's currently not possible to write g.bcf at all, compressed or not.  I think we are actually compressing bcf though automatically on spark, but it interacts strangely with the .gz codepaths.  I've updated the comment and opened a new issue https://github.com/broadinstitute/gatk/issues/4303

src/main/java/org/broadinstitute/hellbender/exceptions/UserException.java
lbergelson: I feel like this generally leads to more confusing error messages since you have to fight the existing text to get it to say what you want.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: Repeating it to catch it early, otherwise it will run the entire haplotype caller and then fail.  Updated here as well.

src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSinkUnitTest.java
lbergelson: Added

src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/VariantsSparkSinkUnitTest.java
lbergelson: It is but it misses out on some things like setting the default test verbosity.

src/test/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSparkIntegrationTest.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/exceptions/UserException.java
droazen: It leads to more consistent error messages. You don't have to "fight" the existing text so much as just look at it before adding your message.

PR #4280, opened 2018-01-27
Added WDL for PostprocessGermlineCNVCalls.

### LINKED ISSUE #4281: WDL for somatic CNV PoN creation looks broken
The workflow in cnv_somatic_panel_workflow.wdl defines:
" Int? mem_gb_for_create_read_count_pon"

But when it's used in the call to CreateReadCountPanelOfNormals, the variable is referred to as "mem_for_create_read_count_pon" (i.e. no "_gb").

This causes the workflow to fail...

samuelklee:
Note that I renamed PostProcessGermlineCNVCalls -> PostprocessGermlineCNVCalls.  I also fixed up some minor WDL typos and added some outputs.

Closes #4281.

### DIFF
--- a/scripts/cnv_wdl/cnv_common_tasks.wdl
+++ b/scripts/cnv_wdl/cnv_common_tasks.wdl
@@ -237,6 +237,7 @@ task PostprocessGermlineCNVCalls {
     String vcf_filename = "${entity_id}.vcf.gz"
 
     String dollar = "$" #WDL workaround for using array[@], see https://github.com/broadinstitute/cromwell/issues/1819
+
     command <<<
         set -e
         export GATK_LOCAL_JAR=${default="/root/gatk.jar" gatk4_jar_override}
--- a/scripts/cnv_wdl/cnv_common_tasks.wdl
+++ b/scripts/cnv_wdl/cnv_common_tasks.wdl
@@ -217,3 +217,54 @@ task ScatterIntervals {
         Array[File] scattered_interval_lists = glob("${base_filename}.scattered.*.interval_list")
     }
 }
+
+task PostprocessGermlineCNVCalls {
+    String entity_id
+    Array[File] chunk_path_tars
+    String sample_index
+    File? gatk4_jar_override
+
+    # Runtime parameters
+    Int? mem_gb
+    String gatk_docker
+    Int? preemptible_attempts
+    Int? disk_space_gb
+
+    Int machine_mem_mb = select_first([mem_gb, 7]) * 1000
+    Int command_mem_mb = machine_mem_mb - 1000
+
+    String sample_directory = "SAMPLE_${sample_index}"  #this is a hardcoded convention in gcnvkernel
+    String vcf_filename = "${entity_id}.vcf.gz"
+
+    String dollar = "$" #WDL workaround for using array[@], see https://github.com/broadinstitute/cromwell/issues/1819
+    command <<<
+        set -e
+        export GATK_LOCAL_JAR=${default="/root/gatk.jar" gatk4_jar_override}
+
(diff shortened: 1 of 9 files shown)


scripts/cnv_wdl/cnv_common_tasks.wdl
```diff
@@ -216,4 +216,55 @@ task ScatterIntervals {
...
+    Int? disk_space_gb
+
+    Int machine_mem_mb = select_first([mem_gb, 7]) * 1000
+    Int command_mem_mb = machine_mem_mb - 1000
+
+    String sample_directory = "SAMPLE_${sample_index}"  #this is a hardcoded convention in gcnvkernel
+    String vcf_filename = "${entity_id}.vcf.gz"
+
+    String dollar = "$" #WDL workaround for using array[@], see https://github.com/broadinstitute/cromwell/issues/1819
+    command <<<
```
asmirnov239: Space before `command`

scripts/cnv_wdl/germline/cnv_germline_case_workflow.wdl
```diff
@@ -177,12 +177,26 @@ workflow CNVGermlineCaseWorkflow {
...
+            sample_index = 0,
+            gatk4_jar_override = gatk4_jar_override,
+            gatk_docker = gatk_docker,
+            preemptible_attempts = preemptible_attempts
+    }
+
     output {
         File preprocessed_intervals = PreprocessIntervals.preprocessed_intervals
-        File read_counts = CollectCounts.counts
+
```
asmirnov239: Can we get rid of the spaces here?

scripts/cnv_wdl/germline/cnv_germline_cohort_workflow.wdl
```diff
@@ -220,12 +220,31 @@ workflow CNVGermlineCohortWorkflow {
...
+                sample_index = sample_index,
+                gatk4_jar_override = gatk4_jar_override,
+                gatk_docker = gatk_docker,
+                preemptible_attempts = preemptible_attempts
+        }
+    }
+
     output {
+        File preprocessed_intervals = PreprocessIntervals.preprocessed_intervals
+
```
asmirnov239: And here too?

scripts/cnv_wdl/cnv_common_tasks.wdl
```diff
@@ -216,4 +216,55 @@ task ScatterIntervals {
     output {
         Array[File] scattered_interval_lists = glob("${base_filename}.scattered.*.interval_list")
     }
+}
+
+task PostprocessGermlineCNVCalls {
+    String entity_id
+    Array[File] chunk_path_tars
```
asmirnov239: Will this always pass chunks in the sorted order?

samuelklee: Only if the output of PreprocessIntervals is sorted.  Actually, where do you check that the intervals are sorted in PostprocessGermlineCNVCalls?  I'll file an issue for us to double check.

asmirnov239: I meant are the files in chunk_path_tars passed according to the indices order(i.e. same order they were processed in the scatter block upstream)? The intervals ordering is checked now on the fly for each posterior record passed to the postprocessor, so it checks that next interval is in the right order compared to the previous one according to the SAMSequenceDictionary found in interval files in chunk directories

samuelklee: OK, great.  I don't think we require PreprocessIntervals output to be sorted, so we should fix that.

This task is called in the workflow with `chunk_path_tars = GermlineCNVCallerCohortMode.gcnv_calls_tar`, so I think we are OK with respect to order of the scatter indices.

Issue #4287, opened 2018-01-29
Revert exposure of optional parameters in CNV WDLs when Cromwell support is available.


samuelklee:
See #3980.

samuelklee:
@ruchim is there a status update on this?

PR #4288, opened 2018-01-29
Exposed optional task-level parameters and standardized runtime parameters in CNV WDLs.

### LINKED ISSUE #3980: Expose or hide all task-level parameters in CNV WDLs consistently.
Somatic WDLs have them hidden, gCNV WDLs (in sl_gcnv_ploidy_cli) have them exposed.  (EDIT: Actually, gCNV WDLs only have them exposed for gCNV-specific tasks.  Common tasks such as PreprocessIntervals are also not exposed.)

The former makes for cleaner `wdltool inputs` JSONs that contain only the bare minimum inputs, but it is unclear whether FC will allow for task-level parameters to be set.  However, this can still be done via JSON, as long as the task is at the main workflow level (although this may change with a C30 hotfix?)

The latter makes for messier JSONs and requires more upkeep to make sure everything stays exposed, but should work in FC (unless the workflow is used as a subworkflow?)

samuelklee:
Just the last commit.  As requested by @LeeTL1220 to enable Intel to access task-level parameters for subworkflows.

Note that this adds an unnecessary amount of boilerplate and will quickly become untenable if there are many identically named parameters.  As discussed in #3980, this sort of thing really should be handled by Cromwell, otherwise there is not much benefit to using optional parameters.  I've filed #4287 to revert when we can.

Closes #3980.


samuelklee:
@LeeTL1220 Note that I've validated with womtool, but as we've seen in the past, changes of this sort (which deal with optional parameters, etc.) may slip through even if tests pass.  You should take a careful look to make sure everything is in order!

LeeTL1220:
@samuelklee Should I be worried that the tests don't pass?

samuelklee:
I think I fixed it---we'll see!

LeeTL1220:
The code that handles the parameter exposure is in womtool.  I'll start talking to the engineers.

samuelklee:
Tests appear to be passing.  Will rebase on master to remove commits from #4280 and merge so we don't delay the release.

PR #4292, opened 2018-01-29
Added check for non-finite copy ratios in ModelSegments pipeline.

samuelklee:
@LeeTL1220 This should fail earlier in situations like the one you ran into with the low coverage test data.  I don't *think* there should be any unintended side effects.  I'm fine if this doesn't make it in before the point release if you want to run some sanity checks on real data with it (since users should be able to figure out what is going on relatively quickly if they run into the issue), but I'll leave it up to you.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/SVDDenoisingUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/SVDDenoisingUtils.java
@@ -473,6 +473,11 @@ public final class SVDDenoisingUtils {
     private static void divideBySampleMedianAndTransformToLog2(final RealMatrix matrix) {
         logger.info("Dividing by sample medians and transforming to log2 space...");
         final double[] sampleMedians = MatrixSummaryUtils.getRowMedians(matrix);
+        IntStream.range(0, sampleMedians.length).forEach(sampleIndex ->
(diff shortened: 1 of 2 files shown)


LeeTL1220:
@samuelklee I'm fine without this one in next release.

samuelklee:
@LeeTL1220 OK, tweaked the message a bit.  I think I'm OK with this going in for the next point release.  This is the sort of thing for which it will be nice to have the automatic validations, as a sanity check.

LeeTL1220:
@samuelklee Is it possible to expand the error message a bit so that users get something actionable.  For example, "Does the interval list match what was covered in the bam file?  Does this bam have coverage in all intervals?"

samuelklee:
I'd rather keep the message more generic, and think of the check as simply defining what a valid `CopyRatio` object can be: an interval associated with a finite double value.  One might imagine that someone would try to create such an object that does not originate from a BAM (perhaps for test data, or for imputing missing values in pre-existing data, etc.).  This check says that they must create it with some finite value.

A more appropriate place for the sort of message you suggest is in the relevant denoising method.  In the edge case you encountered, you used a BAM that was almost completely uncovered in all bins at the specified resolution, resulting in a sample median of zero.  Since the first step in standardization is dividing by the sample median, this results in a divide by zero.  I've added the corresponding check.

samuelklee:
@LeeTL1220 Is this good to go?

LeeTL1220:
Go for it.

PR #4297, opened 2018-01-29
m2 wdl works with cram

### LINKED ISSUE #4285: M2 task will not work on cram files
Similar to #4253 

The M2 task calls out to `GetSampleName`, twice, but does not supply the input reference.  Therefore, it will not work with cram.

Proposed solution:  Simply add `-R ${ref_fasta}` to each invocation of `GetSampleName` in the M2 wdl

davidbenjamin:
Closes #4285.

@LeeTL1220 Here you go.  Note the additional changes besides those mentioned in #4285.  I'm going to head out before finding out whether tests pass.  If they fail I will try to address tonight.

### DIFF
--- a/scripts/mutect2_wdl/mutect2.wdl
+++ b/scripts/mutect2_wdl/mutect2.wdl
@@ -224,6 +224,9 @@ workflow Mutect2 {
             input:
                 gatk_override = gatk_override,
                 intervals = intervals,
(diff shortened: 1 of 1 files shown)


LeeTL1220:
@davidbenjamin Feel free to merge if tests pass.

PR #4309, opened 2018-01-30
replacing inappropriate overrides of makeReadFilter

lbergelson:
replacing overrides of `makeReadFilter` with `getDefaultReadFilters`
This way they these tools will obey read filter command line options correctly.

@SHuang-Broad I think you accidentally used the wrong method here, which is very reasonable because it's super confusing.  I think this is an improvement, but feel free to reject this is you disagree, I may be misunderstanding.


droazen:
@SHuang-Broad We didn't make `makeReadFilter()` final because in rare cases a tool or traversal might need to override it. The docs for `makeReadFilter()` state that you shouldn't override it in general, however: "Most tools will not need to override the method, and should only do so in order to provide custom behavior or processing of the final merged read filter."

SHuang-Broad:
I see, thanks for explaining!

PR #4314, opened 2018-01-31
upgrading to spark 2.2.0

### LINKED ISSUE #2555: upgrade to spark 2.2
We should try to keep on top of the latest spark version while we're still not released yet.  

When we upgrade we should verify that #2545 is resolved.

### LINKED ISSUE #4186: get rid of ADAMSparkRegistrator spam
```
18/01/17 08:24:55 INFO org.bdgenomics.adam.serialization.ADAMKryoRegistrator: Did not find Spark internal class. This is expected for Spark 1.
``` 

Appears hundreds of times in the log output since we upgraded ADAM.  This is despite the fact that we're not using spark 1.  Maybe there's a way to silence a specific logger?

lbergelson:
This bumped some transitive dependencies which required a minor update in unrelated classes

We shouldn't merge this until we get a 👍 from the SV team as well as running the jenkins spark tests.

I think the SV team is already using 2.2.0 since they've gone to dataproc image 1.2.

This will prevent the annoying adam log spam, closes #4186 
closes #2555

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,7 @@ repositories {
 final htsjdkVersion = System.getProperty('htsjdk.version','2.14.1')
 final picardVersion = System.getProperty('picard.version','2.17.2')
 final barclayVersion = System.getProperty('barclay.version','2.1.0')
-final sparkVersion = System.getProperty('spark.version', '2.0.2')
+final sparkVersion = System.getProperty('spark.version', '2.2.0')
 final hadoopBamVersion = System.getProperty('hadoopBam.version','7.9.1')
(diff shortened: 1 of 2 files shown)


lbergelson:
This is exploding due to netty conflicts between version 4.0.4x and 4.1.x.  I'm not sure why the move to spark 2.2.0 changed this exactly, but the version that causes problem is imported by google's gatk-java library.  Removing that removes the conflict.  I rebased this ontop of the branch I have in PR that removes that library and hopefully it won't fail...

droazen:
@lbergelson How's this one looking? Can it be merged yet?

droazen:
@lbergelson Are you comfortable merging this one yet?

droazen:
@lbergelson As part of your current set of Spark test runs, can you also test whether this PR appears to work on dataproc?

Issue #4319, opened 2018-01-31
CNV Somatic Pair Workflow (v4.0.1.0) not compatible with latest Cromwell


ruchim:
The CNV Somatic Pair workflow is missing workflow level inputs for "additional_args" and "mem_gb_for_oncotator" (both of which are fed into the Oncotator sub-workflow) and won't run on the latest Cromwell (v30.2).

samuelklee:
Grr, looks like this slipped by us in #4288.  See my comments there---why does this validate in womtool?  We got bit by a similar issue in #4281.

ruchim:
Agreed this is certainly a bug with womtool. Has this issue already been filed?

samuelklee:
Also, the sooner Cromwell handles exposure of subworkflow task-level parameters, the better.  I had to make these changes to bubble up and expose all optional parameters for all subworkflows, a process which adds an enormous amount of boilerplate and is (obviously) prone to errors.  I plan to revert the changes when Cromwell is ready (#4287), so let me know!

I did not file an issue yet.  @LeeTL1220 may have?

samuelklee:
Also, @LeeTL1220 @davidbenjamin @jonn-smith  I think we need to have some plan for running Oncotator WDL tests on Travis and/or moving everything over to a single Funcotator WDL in the near future.

LeeTL1220:
@ruchim would you be able to run the centaur tests on an arbitrary hash?   That way we don't release bad WDL.

LeeTL1220:
@ruchim @samuelklee I can file an issue once I know who womtool PO is.

LeeTL1220:
Fixing womtool is a bit more complicated, since it affects FireCloud method configuration parameters as well.

LeeTL1220:
@katevoss Here is the issue that was filed when we discovered https://github.com/broadinstitute/cromwell/issues/3230

PR #4321, opened 2018-01-31
fixing memory GB/MB inconsistency in M2 wdl

davidbenjamin:
@bshifaw found this bug in the M2 wdl where it requests eg 3500 GB of RAM instead of 3500 MB, causing disastrous "no machines available" errors.  @LeeTL1220 any way to get this into a release?


samuelklee:
Let's fix #4319 at the same time.

LeeTL1220:
Can you coordinate with @droazen?

davidbenjamin:
@LeeTL1220 I just double-checked the template and this PR conforms.

PR #4330, opened 2018-02-01
Add tests to CPX variants in SV pipeline and fixing bugs

SHuang-Broad:
E.g. The `CTG_NAMES` and `TOTAL_MAPPINGS`.

Also grouping the equivalent variant from different contigs into one, i.e. avoiding duplicates.


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantDetector.java
```diff
@@ -44,7 +44,8 @@
  * paint the full picture we could decide to emit all BND records,
  * but that could be dealt with later.
  */
-public final class CpxVariantDetector {
+public final class CpxVariantDetector implements Serializable {
```
cwhelan: Why is this `java.io.Serializable`? I don't see where this class get serialized.. I always get nervous when we mix java serialization with kryo serialization.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantDetector.java
```diff
@@ -879,6 +852,153 @@ private static boolean alignmentIsDisjointFromAlphaOmega(final SimpleInterval al
...
+     *         segments
+     *     </li>
+     *     <li>
+     *         alt arrangements
+     *     </li>
+     * </ul>
+     * @return
+     */
+    @DefaultSerializer(EquivKey.Serializer.class)
+    private static final class EquivKey {
```
cwhelan: Can you give this class a name that is more descriptive of what it represents? I'm thinking something like `CpxVariantCanonicalRepresentation` or something like that. Make sense to you?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantDetector.java
```diff
@@ -59,8 +60,17 @@ public void inferSvAndWriteVCF(final JavaRDD<AssemblyContigWithFineTunedAlignmen
         final JavaRDD<AnnotatedContig> annotatedContigs =
                 assemblyContigs.map(tig -> new AnnotatedContig(tig, referenceSequenceDictionaryBroadcast.getValue()));
 
-        SVVCFWriter.writeVCF(annotatedContigs.map(tig -> tig.toVariantContext(referenceBroadcast.getValue())).collect(),
-                outputPath, referenceSequenceDictionaryBroadcast.getValue(), toolLogger);
+        final List<VariantContext> complexVariants =
+                annotatedContigs
+                        .groupBy(tig -> new EquivKey(tig.getAffectedRefRegion(), tig.getAltSeq(), tig.getSegmentsAndDescription()))
+                        .map(pair -> {
+                            final VariantContext rawVariantContext = pair._1.toVariantContext(referenceBroadcast.getValue());
```
cwhelan: I think it might be nice to extract this into a method called `produceCpxVariantContext(Tuple2<>...)`. Then it would be great to write a little unit test demonstrating the functionality (ie getting one variant context back with the right annotations set).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantDetector.java
cwhelan: If it's for some reason necessary because you're serializing the inner class, might be better to bump the inner class to a top-level class.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantDetector.java
```diff
@@ -879,6 +852,153 @@ private static boolean alignmentIsDisjointFromAlphaOmega(final SimpleInterval al
...
+    private static Map<String, String> standardAttributes(final Iterable<AnnotatedContig> annotatedContigs) {
+        final List<AnnotatedContig> sorted = Utils.stream(annotatedContigs)
+                .sorted(Comparator.comparing(tig -> tig.tigWithInsMappings.getSourceContig().contigName))
+                .collect(Collectors.toList());
+
+        final Map<String, String> attributeMap = new HashMap<>();
+        attributeMap.put(GATKSVVCFConstants.TOTAL_MAPPINGS,    String.valueOf(sorted.size()));
+        attributeMap.put(GATKSVVCFConstants.CONTIG_NAMES,      sorted.stream().map(annotatedContig -> annotatedContig.tigWithInsMappings.getSourceContig().contigName).collect(Collectors.joining(VCFConstants.INFO_FIELD_ARRAY_SEPARATOR)));
+
+        // TODO: 2/1/18 question: show we introduce new annotations for them?
```
cwhelan: What do you mean by "introduce new annotations"? I'd add these in so that your calls are equivalent to the calls from the current tool. However, to make the annotations for these complex calls equivalent to those for deletions, etc, I'd look only at the flanking alignments that anchor the variant to the reference -- ie. the head and tail alignments from the contig. Does that make sense?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantDetector.java
```diff
@@ -879,6 +852,153 @@ private static boolean alignmentIsDisjointFromAlphaOmega(final SimpleInterval al
...
+//        attributeMap.put(GATKSVVCFConstants.HQ_MAPPINGS,       tigWithInsMappings.getSourceContig().alignmentIntervals.stream().mapToInt(ai -> ai.mapQual).anyMatch( mq -> mq < CHIMERIC_ALIGNMENTS_HIGHMQ_THRESHOLD ) ? "0" : "1");
+//        attributeMap.put(GATKSVVCFConstants.MAPPING_QUALITIES, tigWithInsMappings.getSourceContig().alignmentIntervals.stream().map(ai -> String.valueOf(ai.mapQual)).collect(Collectors.joining(VCFConstants.INFO_FIELD_ARRAY_SEPARATOR)));
+//        attributeMap.put(GATKSVVCFConstants.ALIGN_LENGTHS,     tigWithInsMappings.getSourceContig().alignmentIntervals.stream().map(ai -> String.valueOf(ai.getSizeOnRead())).collect(Collectors.joining(VCFConstants.INFO_FIELD_ARRAY_SEPARATOR)));
+//        attributeMap.put(GATKSVVCFConstants.MAX_ALIGN_LENGTH,  String.valueOf(tigWithInsMappings.getSourceContig().alignmentIntervals.stream().mapToInt(AlignmentInterval::getSizeOnRead).max().orElse(0)));
+
+        // TODO: 12/11/17 integrate these with those that survived the alignment filtering step?
```
cwhelan: I'm a little confused about what would go into this annotation for a complex variant. Don't you already cover this with your segments representation? Maybe you could write a unit test with some examples?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantDetector.java
```diff
@@ -879,6 +852,153 @@ private static boolean alignmentIsDisjointFromAlphaOmega(final SimpleInterval al
...
+            if ( ! segmentsAndAltArrange.referenceSegments.isEmpty() ) {
+                attributes.put(CPX_SV_REF_SEGMENTS,
+                        String.join(VCFConstants.INFO_FIELD_ARRAY_SEPARATOR,
+                                segmentsAndAltArrange.referenceSegments.stream()
+                                        .map(SimpleInterval::toString).collect(Collectors.toList())));
+            }
+            return attributes;
+        }
+
+        VariantContext toVariantContext(final ReferenceMultiSource reference) throws IOException {
```
cwhelan: What if you made this return a `VariantContextBuilder` instead of a `VariantContext`, then you wouldn't have to make it, then make a new builder from the new `VariantContext`, and then make that?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantDetector.java
```diff
@@ -879,6 +852,153 @@ private static boolean alignmentIsDisjointFromAlphaOmega(final SimpleInterval al
...
+
+            final CpxVariantType cpxVariant = new CpxVariantType(affectedRefRegion, typeSpecificExtraAttributes());
+            final SimpleInterval pos = new SimpleInterval(affectedRefRegion.getContig(), affectedRefRegion.getStart(), affectedRefRegion.getStart());
```
cwhelan: Are we sure we want the end to be equal to the start for these?

SHuang-Broad:
Hi @cwhelan , I've expanded this PR to do more than what it originally was trying to fix, and separated the patches by commits as usual:

* the originally proposed fix, which brings back the annotation that are available to simple variants but go missing due to a careless bug, is now done in commit 50f1b640a31ddb528dc763b83b26a9d98dce8556; this commit also accordingly refactors the giant class `CpxVariantDetector` into three new classes
* in the 2nd commit 734516383fb665a79796de76535560fc03cb754b, I did more refactoring on how we group the descriptions for the annotation keys, and updated the test VCF files accordingly.
* because of the refactoring, the review comments were gone, so I added them back in the 3rd commit b7619c45a949dfba21d65a5ed876bc72e832aa77, which contains the comments and my replies. They come in as TODO's but are going to be removed ultimately
* in the following commits, I added tests for the CPX code path, selecting three representative cases (there's no limit how complex the scenario can go). One particular commit 224c97c7b736e94ed6b4d8b067ec830a9f8f2403 is large but most of it is for adding a flat file that contains the chromosome names in hg38 and their lengths for building a bare bone sequence dictionary used in building test data.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
```diff
@@ -348,10 +362,33 @@ static SimpleInterval getAffectedReferenceRegion(final List<SimpleInterval> even
...
+        /**
+         * REVIEW COMMENT:
+         * Are we sure we want the end to be equal to the start for these?
+         *
+         * REPLY:
+         * Not sure I understand the comment.
+         * the variable "pos" below is going to be the POS column of the VCF record,
+         * and several lines down below is where we populate the END INFO field.
+         */
         final SimpleInterval pos = new SimpleInterval(affectedRefRegion.getContig(), affectedRefRegion.getStart(), affectedRefRegion.getStart());
```
cwhelan: Oh I see. I was confused about what this interval is being used for, sorry. If the variable was named 'refAlleleInterval' or something maybe it would help?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantInterpreter.java
```diff
@@ -113,18 +121,153 @@ private static AssemblyContigWithFineTunedAlignments furtherPreprocess(final Ass
...
-            result.add(deoverlapped.get(0));
-            one = deoverlapped.get(1);
+            final int overlapOnContig = AlignmentInterval.overlapOnContig(one, two);
+            if ( overlapOnContig == 0 ) { // nothing to remove
+                result.add(one);
+                one = two;
+            } else {
+                Boolean hintRequestYieldingOverlapToAlignmentTwo = null;
+                if (one.equals(originalAlignments.get(0))) {
+                    hintRequestYieldingOverlapToAlignmentTwo = Boolean.FALSE;
```
cwhelan: just use `false` or `true`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantInterpreter.java
```diff
@@ -113,18 +121,153 @@ private static AssemblyContigWithFineTunedAlignments furtherPreprocess(final Ass
...
             }
-            final List<AlignmentInterval> deoverlapped = ContigAlignmentsModifier.removeOverlap(one, two, refSequenceDictionary);
-            result.add(deoverlapped.get(0));
-            one = deoverlapped.get(1);
+            final int overlapOnContig = AlignmentInterval.overlapOnContig(one, two);
+            if ( overlapOnContig == 0 ) { // nothing to remove
+                result.add(one);
+                one = two;
+            } else {
+                Boolean hintRequestYieldingOverlapToAlignmentTwo = null;
```
cwhelan: Rather than a three value variable (`null`/`true`/`false`) which is effectively how you're using this, I think it would be more explicit if you just had two booleans, eg, `firstIsAlignmentHead`, `secondIsAlignmentTail`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantInterpreter.java
```diff
@@ -113,18 +121,153 @@ private static AssemblyContigWithFineTunedAlignments furtherPreprocess(final Ass
...
+                        removeOverlap(one, two, overlapOnContig, refSequenceDictionary, hintRequestYieldingOverlapToAlignmentTwo);
+                result.add(deoverlapped._1);
+                one = deoverlapped._2;
+            }
         }
         result.add(one);
+
+        final AlignmentInterval head = result.get(0);
+        final AlignmentInterval tail = result.get(result.size() - 1);
+        final int ovp = AlignmentInterval.overlapOnContig(head, tail);
```
cwhelan: spell out `overlap`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
```diff
@@ -100,6 +98,31 @@
...
+        this.affectedRefRegion = affectedRefRegion;
+        this.referenceSegments = referenceSegments;
+        this.eventDescriptions = eventDescriptions;
+        this.altSeq = altSeq;
+    }
+
+    @VisibleForTesting
+    static CpxVariantCanonicalRepresentation makeInterpretation(final CpxVariantInducingAssemblyContig cpxVariantInducingAssemblyContig) {
+
+        if (cpxVariantInducingAssemblyContig.getEventPrimaryChromosomeSegmentingLocations().size() == 1) {
```
cwhelan: Why have two different code paths here and two different constructors? All this method is doing is looking at data in the `cpxVariantInducingAssemblyContig`, so why not just have one constructor that takes that? Then this method can be super simple or just go away.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
```diff
@@ -447,8 +467,12 @@ private CpxVariantCanonicalRepresentation(final Kryo kryo, final Input input) {
 
         final int numSegments = input.readInt();
         referenceSegments = new ArrayList<>(numSegments);
-        for (int i = 0; i < numSegments; ++i)
-            referenceSegments.add(kryo.readObject(input, SimpleInterval.class));
+        for (int i = 0; i < numSegments; ++i) {
```
cwhelan: Why did you have to re-do this serialization instead of delegating to `SimpleInterval.Serializer`?

SHuang-Broad: There was no Kryo serializer for `SimpleInterval` since it is Java serializable, and the class is not registered either.

The serialization tests (which are new in this PR) fail if I did the `kryo.readObject(input, SimpleInterval.class)`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantInterpreter.java
SHuang-Broad: changed as suggested.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
SHuang-Broad: changed as suggested. now much cleaner. Thanks!

PR #4333, opened 2018-02-02
Remove min crazy zish score for template size evidence

cwhelan:
This PR removes the cap on insert size for calling an 'innie' read pair SV evidence. Previously this was set to a max Z score of 100, which worked out to be around 10kb for some libraries. This was causing us to not gather read-pair based evidence around deletions of over 10kb. The motivation for this limit was to exclude very long read pairs as evidence since they are predominantly mapping errors. I believe that a) this limit was set too low, causing us to miss real events and b) the reads with very long insert sizes likely cover many instances of real variation, even if the actual mappings locations are incorrect, and that the assembly process, variant filters, and correlation with coverage data should be able to filter out assemblies triggered by reads with long mappings. 

Removing the limit on insert size does cause the imprecise deletion-calling code to produce many more spurious calls. In practice, I think that we should still evaluate this evidence but not make a call unless it is supported by coverage information, and a long deletion event should be very visible from depth-based calling. Therefore, I've added a limit on the size of imprecise deletions that will be called right now, with the expectation that this hard limit will be removed in favor of depth and other filtering in the future.

Manual evaluation of the results shows a small increase in incorrect imprecise variants (mostly in centromeric regions) that were just small enough to slip under the hard threshold, a very small number of incorrect small insertions caused by assembling messy regions, and a number of real 10-25kb deletions that we now assemble and either call an imprecise or precise deletion variant at, so I think this is a small net gain in callset quality.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryArgumentCollection.java
@@ -182,8 +182,9 @@ public class StructuralVariationDiscoveryArgumentCollection implements Serializa
         public static final int CHIMERIC_ALIGNMENTS_HIGHMQ_THRESHOLD = 60;
         public static final int DEFAULT_MIN_ALIGNMENT_LENGTH = 50; // Minimum flanking alignment length filters used when going through contig alignments.
         public static final int DEFAULT_ASSEMBLED_IMPRECISE_EVIDENCE_OVERLAP_UNCERTAINTY = 100;
-        public static final int DEFAULT_IMPRECISE_EVIDENCE_VARIANT_CALLING_THRESHOLD = 7;
+        public static final int DEFAULT_IMPRECISE_VARIANT_EVIDENCE_THRESHOLD = 7;
         public static final int DEFAULT_TRUTH_INTERVAL_PADDING = 50;
+        public static final int DEFAULT_MAX_CALLABLE_IMPRECISE_DELETION_SIZE = 15000;
 
         @Argument(doc = "Minimum flanking alignment length", fullName = "min-align-length")
         public Integer minAlignLength = DEFAULT_MIN_ALIGNMENT_LENGTH;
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/ImpreciseVariantDetector.java
```diff
@@ -20,13 +21,16 @@
...
-                                                                              final int impreciseEvidenceVariantCallingThreshold,
+                                                                              final int impreciseVariantEvidenceThreshold,
+                                                                              final int maxCallableImpreciseVariantDeletionSize,
                                                                               final Logger localLogger) {
         final List<VariantContext> impreciseVariants =
                 Utils.stream(evidenceTargetLinks)
                         .map(p -> p._2)
                         .filter(EvidenceTargetLink::isImpreciseDeletion)
-                        .filter(e -> e.getReadPairs() + e.getSplitReads() > impreciseEvidenceVariantCallingThreshold)
+                        // todo: for now, we set a limit on the size of deletion we can call solely based on imprecise evidence
```
SHuang-Broad: Can you create a ticket for this, just so that we don't forgot it like what happened in this z-score filtering?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryArgumentCollection.java
```diff
@@ -198,8 +199,8 @@
         public int assemblyImpreciseEvidenceOverlapUncertainty = DEFAULT_ASSEMBLED_IMPRECISE_EVIDENCE_OVERLAP_UNCERTAINTY;
 
         @Argument(doc = "Number of pieces of imprecise evidence necessary to call a variant in the absence of an assembled breakpoint.",
-                fullName = "imprecise-evidence-variant-calling-threshold")
-        public int impreciseEvidenceVariantCallingThreshold = DEFAULT_IMPRECISE_EVIDENCE_VARIANT_CALLING_THRESHOLD;
+                fullName = "imprecise-variant-evidence-threshold")
```
SHuang-Broad: Alternatively, can we still call such variants, and filter out the really large ones? Is performance a concern if we go this way?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadClassifier.java
```diff
@@ -17,7 +17,6 @@
     private static final byte MIN_QUALITY = 15; // minimum acceptable quality in a soft-clip window
     private static final int MAX_LOW_QUALITY_SCORES = 3; // maximum # of low quality base calls in soft-clip window
     private static final float MAX_ZISH_SCORE = 6.f; // maximum fragment-length "z" score for a normal fragment
-    private static final float MIN_CRAZY_ZISH_SCORE = 100.f; // "z" score that's probably associated with a mapping error
```
SHuang-Broad: Can we rename `MAX_ZISH_SCORE` to something like `MAX_REGULAR_ZISH_SCORE`? (I'm trying to avoid terms like normal since distributions are involved here).

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/ImpreciseVariantDetector.java
cwhelan: Added https://github.com/broadinstitute/gatk/issues/4372

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadClassifier.java
cwhelan: Renamed to `MAX_NON_OUTLIER_ZISH_SCORE`

PR #4339, opened 2018-02-05
Variable depth bins in Mutect2 autoval

davidbenjamin:
@skwalker @madduran This puts in the variable depth bins that make the sensitivity part of M2 autoval viable for WGS.  Would one of you care to review?

@LeeTL1220 Looping you in.


madduran:
@davidbenjamin looks good to me

PR #4352, opened 2018-02-05
fixing bugs in CalculateGenotypePosteriers

### LINKED ISSUE #4346: Duplicate records in VCF produced by CalculateGenotypePosteriors
### Affected tool(s)
CalculateGenotypePosteriors

### Affected version(s)
4.0.1.1

### Description 
The input VCF only contains 1 record for the position, but after running CalculateGenotypePosteriors, the output VCF contains 2 records for the position.

This Issue was generated from your [forums] 
[forums]: https://gatkforums.broadinstitute.org/gatk/discussion/11278/duplicate-records-after-calculategenotypeposteriors/p1

lbergelson:
fixing 3 issues in CalclualteGenotypePosteriers
* no longer emit duplicate variants if you have overlapping VCF records in the input
* specifying vcf.gz output will produce a vcf.gz instead of an unzipped file called vcf.gz
* fix potential NPE during shutdown

* fixes #4346


lbergelson:
@ldgauthier Could you take a look at this? I changed a bunch of logic to bring it inline with modern tool design.  I think it's right, but maybe I'm breaking something that I didn't understand?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
```diff
@@ -223,21 +223,14 @@
...
 
     private FamilyLikelihoods famUtils;
-    private SampleDB sampleDB = null;
 
     private VariantContextWriter vcfWriter;
 
     @Override
     public void onTraversalStart() {
-        final VariantContextWriterBuilder builder = new VariantContextWriterBuilder().setOutputFile(out).setOutputFileType(VariantContextWriterBuilder.OutputType.VCF);
-        if (hasReference()){
```
ldgauthier: I already wasn't happy with the reference checking in this tool and then you took out more of it?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
```diff
@@ -257,17 +250,6 @@ public void onTraversalStart() {
             throw new UserException("VCF has no genotypes");
         }
 
-        if ( header.hasInfoLine(GATKVCFConstants.MLE_ALLELE_COUNT_KEY) ) {
-            final VCFInfoHeaderLine mleLine = header.getInfoHeaderLine(GATKVCFConstants.MLE_ALLELE_COUNT_KEY);
-            if ( mleLine.getCountType() != VCFHeaderLineCount.A ) {
-                throw new UserException("VCF does not have a properly formatted MLEAC field: the count type should be \"A\"");
```
ldgauthier: There's no regression test for this, but it's from a particular issue: https://github.com/broadinstitute/gsa-unstable/issues/845
The issue was caused by a 4.0 VCF as the supplemental VCF.

lbergelson: Hmn, I think I thought this was being checked somewhere else now, but I can't find any evidence that it is.  Not sure why I took that out...

lbergelson:
@ldgauthier  I undid that change that made no sense and added some new tests so it doesn't happen again.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
```diff
@@ -300,40 +296,39 @@ public void apply(final VariantContext variant,
...
+            builder.genotypes(gc);
+        }
+        VariantContextUtils.calculateChromosomeCounts(builder, false);
+        vc_familyPriors = builder.make();
+
+        if (!skipPopulationPriors) {
+            vc_bothPriors = PosteriorProbabilitiesUtils.calculatePosteriorProbs(vc_familyPriors, otherVCs, missing * numRefIfMissing, globalPrior, !ignoreInputSamples, defaultToAC, useACoff);
+        } else {
+            final VariantContextBuilder builder2 = new VariantContextBuilder(vc_familyPriors);
             VariantContextUtils.calculateChromosomeCounts(builder, false);
```
ldgauthier: Oops, this should probably be builder2, right?  All this time and no one ever said anything.

ldgauthier:
Take a look at the builder2 issue, but otherwise LGTM! :+1:

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
lbergelson: It definitely looks like it should be.  It's effectively the same thing, but it's very confusing and if anyone ever make changes it could lead to problems... I might start a separate PR to address that.

lbergelson: wait, what am I talking about... the thing that gets output should be the the result of modifying builder2 and we modify builder.  That seems super wrong.

lbergelson: no, i'm back to "it's dumb but fine"

lbergelson: new pr to fix this https://github.com/broadinstitute/gatk/pull/4431

PR #4354, opened 2018-02-06
Removed debugging line from CNV plotting script.

samuelklee:


samuelklee:
@mbabadi Can we get this in for the next point release?  (I know it's a very long PR...)  Thanks!

mbabadi:
Well, I think I'm gonna sleep on this one and let the main idea sink in a bit before I can start with my review.

PR #4356, opened 2018-02-06
Pv sw enable av xcode default

pnvaidya:
Default for smithwaterman will use optimized AVX code


droazen:
This PR will have to wait until we have continuous testing up and running for the AVX2 implementation of `SmithWaterman`

droazen:
I think we can finally unblock this using PAPIv2's ability to request machine types on Google Cloud. We just have to configure our jenkins server to use PAPIv2.

PR #4360, opened 2018-02-06
M2 wdl handles sample names with space

davidbenjamin:
@LeeTL1220 This fixes the issue.  I tested it on SGE and the M2 integration tests now have a DREAM bam where I switched the sample to "tumor sample".

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
@@ -13,6 +13,9 @@ import org.broadinstitute.hellbender.exceptions.UserException;
 import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -51,12 +54,24 @@ import java.util.stream.Collectors;
 )
 final public class GetSampleName extends GATKTool {
 
+    public static final String STANDARD_ENCODING = "UTF-8";
+
     @Argument(
             doc = "Output file with only the sample name in it.",
             fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
             shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME
     )
     protected File outputSampleNameFile;
+
+    public static final String URL_ENCODING_LONG_NAME = "use-url-encoding";
+    public static final String URL_ENCODING_SHORT_NAME = "encode";
+
+    @Argument(
+            doc = "Apply URL encoding to convert spaces and other special characters in sample name.",
+            fullName = URL_ENCODING_LONG_NAME,
+            shortName = URL_ENCODING_SHORT_NAME
+    )
+    protected boolean urlEncode;
 
     @Override
     public void traverse() {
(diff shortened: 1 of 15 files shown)


LeeTL1220:
Tests are failing.   Should I still review?

samuelklee:
Can I ask why you guys still have GetSampleName as a separate tool?  At some point I cribbed the code there into a CNV utility method, MetadataUtils.readSampleName(samFileHeader).  Is the point to support multi-sample BAMs as input to M2 (the CNV pipelines currently do not, should they)?

davidbenjamin:
@LeeTL1220 Whoops, surprise side effects.  I think I resolved them but you may as well wait for tests to pass.

src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
```diff
@@ -84,9 +99,27 @@ public void onTraversalStart() {
...
-            fileWriter.write(sampleNames.get(0));
+            final String rawSample = sampleNames.get(0);
+            final String outputSample = urlEncode ? encode(rawSample) : rawSample;
+            fileWriter.write(outputSample);
         } catch (final IOException ioe) {
             throw new UserException("Could not write file.", ioe);
         }
     }
+
+    public static final String encode(final String string) {
```
LeeTL1220: Please move to a utility class. An internal class (Mutect2Engine) should not call into a CLI (see below).  That violates separation of concerns.

src/main/java/org/broadinstitute/hellbender/tools/GetSampleName.java
```diff
@@ -84,9 +99,27 @@ public void onTraversalStart() {
...
+        try {
+            return URLEncoder.encode(string, STANDARD_ENCODING);
+        } catch (final UnsupportedEncodingException ex) {
+            throw new UserException("Could not encode sample name", ex);
+        }
+    }
+
+    public static final String decode(final String string) {
+        try {
```
LeeTL1220: Ditto

davidbenjamin:
@LeeTL1220 Did it.

LeeTL1220:
Merge away

samuelklee:
@davidbenjamin @LeeTL1220 I will assume that I am off the hook for multi-sample BAMs, then!

davidbenjamin:
@samuelklee We need a way to get the sample name within a wdl where only the tumor bam is input.

samuelklee:
Wait, now I am confused---does M2 (the CLI, not the WDL) even support multi-sample BAMs?  That is, can I pass a single BAM containing multiple tumor samples and pull out the one I want by specifying the sample name via the command line?

If not, and all incoming BAMs are required to only contain a single sample, then can you just extract the method that GetSampleName uses (for reading the sample name and checking that there is only one), call that method within M2, and get rid of the use of GetSampleName in the M2 WDL?  That way you don't have to put the name in a file or pass it to the M2 command line.

davidbenjamin:
@samuelklee You are rightfully confused.  I think the `GetSampleName` CLI is still necessary but would be glad to find out that it's not.  Here's the story: GATK4 does not have tagged arguments yet, so unlike GATK3 you can't do `gatk Mutect2 -I:tumor tumor.bam -I:normal normal.bam`.  Since we need to pass bams via `-I` in order to be a `GATKTool`, w.
e're left with `-I tumor.bam -I normal.bam` and we need some other way to say which bam is the tumor and which is the normal.  The most convenient way is to do that via the sample name, because `GATKRead`s know their sample.

Although this is a workaround, it does give the support for multi-sample bams that you mentioned,  not that we have cared about that so far.

samuelklee:
Gotcha.  I was thinking that the order of the -I could be used, but that probably has its own downsides---I guess it's better to fail earlier than to accidentally run with the tumor and normal swapped.

davidbenjamin:
I suspect Geraldine would execute me for that.

samuelklee:
Well, the problem just gets bumped up one level to the WDL, I suppose...!

PR #4364, opened 2018-02-07
Exposed more mem_gb parameters and fixed mem_gb_for_model_segments.

### LINKED ISSUE #4353: mem_gb_for_model_segments not passed through in CNV WDL.
Another parameter slipped by in #4288...

@ruchim Can we get an ETA for when exposing subworkflow task-level parameters will be handled by Cromwell?

samuelklee:
Closes #4353.


samuelklee:
@LeeTL1220 I went ahead and exposed more mem_gb parameters, which is convenient when we want to go below 250bp and the pair WDL is used as a subworkflow.  Please review carefully!

samuelklee:
Thanks @LeeTL1220, I also added another commit to make some minor comment tweaks.  Will merge after tests pass!

LeeTL1220:
@samuelklee Still looks good.

Issue #4345, opened 2018-02-05
Add GATKdocs to GATK bundle zip


droazen:

droazen:
For @cmnbroad

PR #4377, opened 2018-02-08
Include user docs in distribution bundle.

cmnbroad:
Adds a folder called gatkdoc to the bundle. Fixes https://github.com/broadinstitute/gatk/issues/4345.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,7 +66,7 @@ final testNGVersion = '6.11'
 
 final baseJarName = 'gatk'
 final secondaryBaseJarName = 'hellbender'
-
+final docBuildDir = "$buildDir/docs"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -66,7 +66,7 @@ final testNGVersion = '6.11'
 
 final baseJarName = 'gatk'
 final secondaryBaseJarName = 'hellbender'
-
+final docBuildDir = 'build/docs'
```
cmnbroad: this should be using $buildDir.

cmnbroad: Had to change some single quote to doubles to enable interpolation. Unrelated note: is there any downside to standardizing on double quotes everywhere ?

cmnbroad: Done.

PR #4381, opened 2018-02-08
Passing vcf indices correctly in M2 wdls

davidbenjamin:
This is the bug fix for @eitanbanks's pon.  @LeeTL1220 can you review if tests pass?  If they fail I will fix ASAP.

### DIFF
--- a/scripts/mutect2_wdl/unsupported/mutect2_opt.wdl
+++ b/scripts/mutect2_wdl/unsupported/mutect2_opt.wdl
@@ -182,6 +182,7 @@ workflow Mutect2 {
                 preemptible_attempts = preemptible_attempts,
                 m2_extra_args = m2_extra_args,
                 make_bamout = make_bamout_or_default,
+                compress = compress,
                 gatk_override = gatk_override,
                 gatk_docker = gatk_docker,
                 preemptible_attempts = preemptible_attempts,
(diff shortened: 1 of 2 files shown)


LeeTL1220:
@davidbenjamin I think this affects more than the PoN.

LeeTL1220:
Are you sure that @eitanbanks was specifying `compress` of `true`?  If it was `false`, then I believe we've misdiagnosed the issue.  Doubly odd, since `false` is the default.

scripts/mutect2_wdl/unsupported/mutect2_opt.wdl
```diff
@@ -182,14 +182,15 @@ workflow Mutect2 {
...
+                compress = compress,
                 gatk_override = gatk_override,
                 gatk_docker = gatk_docker,
                 preemptible_attempts = preemptible_attempts,
                 disk_space = m2_per_scatter_size,
                 auth = auth
         }
 
-        Float sub_vcf_size = size(M2.output_vcf, "GB")
+        Float sub_vcf_size = size(M2.vcf, "GB")
```
LeeTL1220: I prefer `output_vcf` or something more descriptive than `vcf`... `raw_vcf`?

davidbenjamin:
> Are you sure that @eitanbanks was specifying compress of true? If it was false, then I believe we've misdiagnosed the issue. Doubly odd, since false is the default.

@LeeTL1220 False is and was the default for all final outputs, but the intermediate scatter vcfs were hard-coded to be compressed.

PR #4384, opened 2018-02-09
Removed BetaFeature tag from Mutect2 and FilterMutectCalls

davidbenjamin:
@chandrans Can I give this one to you?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/FilterMutectCalls.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/FilterMutectCalls.java
@@ -68,7 +68,6 @@ import java.util.stream.Collectors;
         programGroup = VariantFilteringProgramGroup.class
 )
 @DocumentedFeature
-@BetaFeature
 public final class FilterMutectCalls extends VariantWalker {
 
     @Argument(fullName= StandardArgumentDefinitions.OUTPUT_LONG_NAME,
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
@@ -121,7 +121,6 @@ import java.util.List;
          programGroup = ShortVariantDiscoveryProgramGroup.class
  )
 @DocumentedFeature
-@BetaFeature
 public final class Mutect2 extends AssemblyRegionWalker {
 
     @ArgumentCollection


droazen:
@LeeTL1220 please also review

LeeTL1220:
@davidbenjamin What is the criteria met here?

chandrans:
@LeeTL1220 Hey Lee. There was some confusion, and my team thought Mutect2 is still in beta. Removing the beta tag will allow our docs to show the tool as not in beta. Simple fix :)

droazen:
@chandrans Untagging a tool as beta is a major step that should not be taken lightly, since it can't be reversed (we can't mark tools as beta again once they've been unmarked, as @vdauwera often reminds us). I think @LeeTL1220 is asking whether the latest master version of `Mutect2` has been run through the necessary evaluations that would justify removing the beta label.

droazen:
I'll also add here: if existing well-known arguments are going to be removed from `Mutect2` in the near future, as `-dbsnp` apparently was just recently (see https://github.com/broadinstitute/gatk/issues/4390), then it might be a good idea to keep it tagged as `@BetaFeature` for now. Once you remove the beta label, you can't really remove widely-used tool arguments without deprecating them for a period of time first (as I think @vdauwera would agree).

chandrans:
Ah, I see. My major push for this to be out of beta is that users are confused and ask about the timeline for Mutect2 to be out of beta status. My team was also confused thinking Mutect2 is still in beta because of the tag. From my position, the tool should be "not in beta" if the users can use it in production. I think the beta tag makes people hesitant to use it for proper analysis even if it is production-ready. However, I leave it to David B and team to determine whether there will be major changes in the code that would make this difficult.

davidbenjamin:
I'm for taking M2 out of beta because being in beta leaves users to the much worse alternatives of using a different Mutect.  @droazen has a good point about changing arguments, but in the case of dbsnp it was an argument that we haven't used in over a year.  I should have given it a deprecated tag long ago but forgot to do so.  I don't foresee taking anything out of the command line until we get tagged arguments, at which point we will remove the `-tumor` and `-normal` arguments.

droazen:
@chandrans @davidbenjamin I'm not actually objecting to the idea of taking `Mutect2` out of beta, for the record. I'm just trying to get people used to the idea that removing the `@BetaFeature` tag is actually a "big step" that requires careful evaluation, since it can't be undone. It's not just something that affects tool documentation -- it affects our entire release and development process now that we're out of beta. If a tool is marked stable, it needs to be kept in a constantly releasable state in master, and major changes to stable tools need to be done in such a way that existing functionality is not compromised.

If the latest master version of `Mutect2` has been run through and passed whatever evaluation criteria/scripts your team relies on @davidbenjamin, and you are comfortable at this point with the additional restrictions that come with doing development on a stable tool, then by all means take it out of beta!

davidbenjamin:
@LeeTL1220 The criteria in my opinion are being the best Mutect and being stable.  Were you suggesting waiting for some validation like MC3?

LeeTL1220:
@davidbenjamin This decision is yours.  I will support it either way.

davidbenjamin:
Okay, I'm going to do it!

Issue #4332, opened 2018-02-02
Optionally copy FASTQ's in `scripts/sv/copy_sv_results.sh`


SHuang-Broad:
And in `scripts/sv/manage_sv_pipeline.sh`.

As suggested by @cwhelan when reviewing PR #4328 

> One suggested feature enhancement if you feel like it while you're working on these scripts: most of the time I don't want to write out and copy all of the fastq files for the assemblies (mostly because it takes a long time to copy them out of the cluster), but occasionally I do. It would be nice if the scripts by default didn't move the fastqs (this could be accomplished either by not writing them or by excluding them from the copy -- @TedBrookings didn't you have code that did the latter before?), but had an optional flag to enable them.

@TedBrookings assigning to you for now, but it may not be a priority.

TedBrookings:
I used to have it in there, but once I got copy performance up I assumed it was just clutter. I need to make a minor update to the scripts anyway (option to run the debug program instead of the full pipeline) so I can put back the option to not copy fastq files as well.

SHuang-Broad:
closed by #4385

PR #4385, opened 2018-02-09
Tb sv help dump evidence

TedBrookings:
I've tried to organize these changes into three separate commits with one issue per commit.
1) Running the new ExtractSVEvidenceSpark tool via manage_sv_pipeline.sh
2) Dumping more useful info for machine learning
3) Reinstating option to not copy fastq files.

Hopefully this is sensible, it's my first time trying to organize a PR this way.

This should resolve issue #4332

### DIFF
--- a/scripts/sv/manage_sv_pipeline.sh
+++ b/scripts/sv/manage_sv_pipeline.sh
@@ -18,6 +18,8 @@ Syntax
       display help and exit
     -q or --quiet:
       run without prompting for user input
+    -n or --no-fastq
+      don't copy fastq files from cluster to cloud storage
     -u [GCS_USER] or --user [GCS_USER]:
       set GCS username (defaults to ${USER})
     -i [INIT_SCRIPT] or --init [INIT_SCRIPT]
@@ -26,12 +28,16 @@ Syntax
     -s [GCS_SAVE_PATH] or --save [GCS_SAVE_PATH]
       save results in specified bucket/folder
       (defaults to \$PROJECT_NAME/\$GCS_USER)
-    -t [MAX_LIFE] or --life:
-      life time of cluster
-      (defaults to \$CLUSTER_MAX_LIFE_HOURS)
-    -d [MAX_IDLE] or --idle:
-      maximum idling time for cluster
-      (defaults to \$CLUSTER_MAX_IDLE_MINUTES)
+    -t [MAX_LIFE] or --life [MAX_LIFE]:
+      life time of cluster (e.g. "2h" for two hours or "1d" for 1 day)
+      (defaults to \$CLUSTER_MAX_LIFE_HOURS if defined, otherwise 4h)
+    -d [MAX_IDLE] or --idle [MAX_IDLE]:
+      maximum idling time for cluster (e.g. "2h" for two hours or "10m"
+      for 10 minutes).
+      (defaults to \$CLUSTER_MAX_IDLE_MINUTES if defined, otherwise 60m)
+    -L [TOOL_NAME] or --launch-tool [TOOL_NAME]
+      specify GATK_SV_TOOL to launch
+      (defaults to "StructuralVariationDiscoveryPipelineSpark")
 
   Mandatory Positional Arguments:
     GATK_Folder: path to local copy of GATK
@@ -69,7 +75,10 @@ throw_error() {
 # parses arguments provided to this script
 QUIET=${SV_QUIET:-"N"}
 GCS_USER=${GCS_USER:-${USER}}
-
+CLUSTER_MAX_LIFE_HOURS=${CLUSTER_MAX_LIFE_HOURS:-4h}
+CLUSTER_MAX_IDLE_MINUTES=${CLUSTER_MAX_IDLE_MINUTES:-60m}
+GATK_SV_TOOL=${GATK_SV_TOOL:-"StructuralVariationDiscoveryPipelineSpark"}
+COPY_FASTQ=${COPY_FASTQ:-"Y"}
 # update gcloud
 if [[ "${QUIET}" == "Y" ]]; then
     gcloud components update --quiet
@@ -85,6 +94,10 @@ while [ $# -ge 1 ]; do
             ;;
         -q|--quiet)
             QUIET="Y"
+            shift
+            ;;
+        -n|--no-fastq)
+            COPY_FASTQ="N"
             shift
             ;;
         -u|--user)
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
```diff
@@ -244,18 +245,21 @@ protected ReadEvidence( final GATKRead read, final ReadMetadata metadata, final
             this.forwardStrand = forwardStrand;
             this.cigarString = read.getCigar().toString();
             this.mappingQuality = read.getMappingQuality();
+            this.templateSize = read.getFragmentLength();
```
cwhelan: My only thought about promoting template size to `ReadEvidence` is that for evidence types other than `WeirdTemplateSize`, the template length will probably not really follow a predictable distribution, and therefore might not be too useful to a machine learning module (or might cause it to overfit noise). Although, on the other hand, maybe excessively long template sizes could be downweighted or something... Also, what does the template size look like for inter-contig pairs? I guess it's worth trying out as a feature, but adding it here will create a bit more overhead for the system, ie. more information will have to be tracked in the evidence gathering stage. One other thing to note -- you should be able to get close to the template size by checking the distance between the evidence interval and the distal target interval.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
```diff
@@ -877,41 +890,32 @@ public SameStrandPair read( final Kryo kryo, final Input input, final Class<Same
...
 
         @Override
         public String toString() {
-            return super.toString() + "\tTemplateSize\t" + target + "\t" + templateSize;
-        }
-
-        @Override
-        public String stringRep(final ReadMetadata readMetadata, final int minEvidenceMapq) {
-            return super.stringRep(readMetadata, minEvidenceMapq) + "\t" + templateSize;
+            return super.toString() + "\tTemplateSize\t" + getTemplateSize();
```
cwhelan: Why not use `super.stringRep` which already includes the template size?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
```diff
@@ -361,6 +361,15 @@ public static void writeMetadata(final ReadMetadata readMetadata,
                                 "+" + stats.getPositiveMAD() + "\t" + stats.getCoverage() + "\t" +
                                 stats.getMeanBaseQuality() + "\t" +
                                 stats.getNReads() + "\t" + stats.getReadStartFrequency() + "\n");
+                final IntHistogram.CDF templateSizeCdf = stats.getCDF();
+                final int cdfSize = templateSizeCdf.size();
+                final long totalObservations = templateSizeCdf.getTotalObservations();
+                writer.write("template size cumulative counts:");
+                for(int idx = 0; idx < cdfSize; ++idx) {
+                    final long cumulativeCounts = Math.round(templateSizeCdf.getFraction(idx) * totalObservations);
```
cwhelan: Why not just go by the CDF -- why multiply by the number of observations?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
```diff
@@ -361,6 +361,15 @@ public static void writeMetadata(final ReadMetadata readMetadata,
                                 "+" + stats.getPositiveMAD() + "\t" + stats.getCoverage() + "\t" +
                                 stats.getMeanBaseQuality() + "\t" +
                                 stats.getNReads() + "\t" + stats.getReadStartFrequency() + "\n");
+                final IntHistogram.CDF templateSizeCdf = stats.getCDF();
+                final int cdfSize = templateSizeCdf.size();
+                final long totalObservations = templateSizeCdf.getTotalObservations();
+                writer.write("template size cumulative counts:");
```
cwhelan: Was going to suggest capitalizing to be consistent with "Library Statistics" but then realized that the headings are all over the place in this file in terms of formatting. Don't feel like you need to do it for this pull request but it would be nice to eventually convert the metadata into a more easily machine-parseable format eventually (JSON?).

scripts/sv/manage_sv_pipeline.sh
```diff
@@ -26,12 +26,15 @@ Syntax
     -s [GCS_SAVE_PATH] or --save [GCS_SAVE_PATH]
       save results in specified bucket/folder
       (defaults to \$PROJECT_NAME/\$GCS_USER)
-    -t [MAX_LIFE] or --life:
+    -t [MAX_LIFE] or --life [MAX_LIFE]:
```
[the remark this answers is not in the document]
TedBrookings: Agreed: I added documentation explaining the time format and defaults.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
TedBrookings: xgboost should be smart enough to ignore uninformative (e.g. constant or random) columns. Of course we don't want to throw stuff at it that we know is terrible, but so far, templateSize looks like it helps most of the time. In the more complex classifier type I presented, I can drop it or tell the classifier to ignore it on a per-evidenceType basis if cross-validation shows it's not helping.

The templateSize for interContigPairs is always 0.

Regarding performance: would it be smarter to just store the read object and add property getters? Then quantities that are kept principally for evidence-dumping like templateSize and cigarString would not be evaluated unless needed. I don't have a good sense of whether this would be an improvement, neutral, or make things worse.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
TedBrookings: The stringRep() functions were added so that I could dump additional info without messing with the toString() functions which you guys have been using for debug purposes. I've been trying to ensure that the two functions are largely independent and thought that it was better to not have them call each other.

Also: I just noticed that I accidentally deleted the print-out of "target", which I've reverted.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/ReadMetadata.java
TedBrookings: Since these floats all originate as integers divided by totalObservations, doing it this way allows me to output the full precision result to text with fewer characters and no special print format overrides.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/BreakpointEvidence.java
cwhelan: I think the read object would be quite a bit larger than the fields we're storing, as it contains the read sequence and base qualities as well as other tags. We've been trying to keep the evidence objects small, which is the only reason I asked if it was necessary to add more data to it.

Regarding learning from the data, one more thing to keep in mind: If the ML module decides, for example, to ignore evidence with really large template sizes because our benchmark data doesn't contain any true events of that size, what happens when we run on a sample that has an event (probably a very rare one) that is really huge? We don't want to miss that. Obviously this is a balancing act, and we could think of our evidence weighting as having a low prior on really huge events,  but if there are ways to engineer our features such that we can still overcome our priors in the face of good evidence that would be good. In this case I wonder if capping the template size feature might help? Just something to think about.

PR #4395, opened 2018-02-12
remove PhysicalLocationForMateCigar

lbergelson:
this class is unused and untested


src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/PhysicalLocationForMateCigar.java
```diff
@@ -1,78 +0,0 @@
-package org.broadinstitute.hellbender.utils.read.markduplicates;
-
-/**
- * @author nhomer
- */
-
-/** Stores the minimal information needed for optical duplicate detection. */
-public final class PhysicalLocationForMateCigar implements OpticalDuplicateFinder.PhysicalLocation {
```
jamesemery: It looks like the functionality that this class is intended to be used for has been subsumed by PairedEnds and the ReadsEnds object for the spark and non-spark versions of the tools respectively. This looks like a good removal to me.

PR #4401, opened 2018-02-13
updating to htsjdk 2.14.3

lbergelson:
* updating tests because a toString changed in htsjdk and it's value is written directly into the outputs of some tools
* fix two trivial mistakes in ExampleVariantWalkerSpark that I noticed


lbergelson:
woops, this should fix the test.

PR #4411, opened 2018-02-14
Allowed use of zero eigensamples in DenoiseReadCounts.

### LINKED ISSUE #4399: Allow zero eigensamples to be used in DenoiseReadCounts.
In case one just wants to normalize against the interval medians.

samuelklee:
Closes #4399.


samuelklee:
@mbabadi I need this to run something for @mwalker174.  OK to merge?

PR #4414, opened 2018-02-14
CNV Tumor only WDL

### LINKED ISSUE #3983: Somatic CNV WDL needs to support tumor-only mode where no matched normal is available.

LeeTL1220:
Fixes #3983 

- Added two tumor-only tests to the CNV WDL test.  I did not add all four possible combinations, since I did not see the value.  I just wanted to make sure that the WGS and WES cases were covered.

### DIFF
--- a/scripts/cnv_cromwell_tests/somatic/run_cnv_somatic_workflows.sh
+++ b/scripts/cnv_cromwell_tests/somatic/run_cnv_somatic_workflows.sh
@@ -34,7 +34,7 @@ sed -r "s/__GATK_DOCKER__/broadinstitute\/gatk\:$HASH_TO_USE/g" ${CNV_CROMWELL_T
 sed -r "s/__GATK_DOCKER__/broadinstitute\/gatk\:$HASH_TO_USE/g" ${CNV_CROMWELL_TEST_DIR}/cnv_somatic_pair_wgs_no-gc_workflow.json >cnv_somatic_pair_wgs_no-gc_workflow_mod.json
 sed -r "s/__GATK_DOCKER__/broadinstitute\/gatk\:$HASH_TO_USE/g" ${CNV_CROMWELL_TEST_DIR}/cnv_somatic_pair_wes_do-gc_workflow.json >cnv_somatic_pair_wes_do-gc_workflow_mod.json
 sed -r "s/__GATK_DOCKER__/broadinstitute\/gatk\:$HASH_TO_USE/g" ${CNV_CROMWELL_TEST_DIR}/cnv_somatic_pair_wgs_do-gc_workflow.json >cnv_somatic_pair_wgs_do-gc_workflow_mod.json
-
+sed -r "s/__GATK_DOCKER__/broadinstitute\/gatk\:$HASH_TO_USE/g" ${CNV_CROMWELL_TEST_DIR}/cnv_somatic_pair_wgs_do-gc_tumor_only_workflow.json > cnv_somatic_pair_wgs_do-gc_tumor_only_workflow_mod.json
 echo "Running ========"
 CROMWELL_JAR="cromwell-0.29.jar"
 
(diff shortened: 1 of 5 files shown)


scripts/cnv_cromwell_tests/somatic/run_cnv_somatic_workflows.sh
```diff
@@ -54,4 +56,10 @@ java -jar ~/${CROMWELL_JAR} run /home/travis/build/broadinstitute/gatk/scripts/c
 # Pair WES w/ explicit GC correction
 java -jar ~/${CROMWELL_JAR} run /home/travis/build/broadinstitute/gatk/scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl -i cnv_somatic_pair_wes_do-gc_workflow_mod.json
 # Pair WGS w/ explicit GC correction
-java -jar ~/${CROMWELL_JAR} run /home/travis/build/broadinstitute/gatk/scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl -i cnv_somatic_pair_wgs_do-gc_workflow_mod.json
\ No newline at end of file
+java -jar ~/${CROMWELL_JAR} run /home/travis/build/broadinstitute/gatk/scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl -i cnv_somatic_pair_wgs_do-gc_workflow_mod.json
+
+# Tumor only WGS w/ explicit GC correction
+java -jar ~/${CROMWELL_JAR} run /home/travis/build/broadinstitute/gatk/scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl -i cnv_somatic_pair_wgs_do-gc_tumor_only_workflow_mod.json
+
```
samuelklee: Remove white space to be consistent.

scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl
```diff
@@ -162,20 +165,22 @@ workflow CNVSomaticPairWorkflow {
...
-            ref_fasta = ref_fasta,
-            ref_fasta_fai = ref_fasta_fai,
-            ref_fasta_dict = ref_fasta_dict,
-            format = format,
-            gatk4_jar_override = gatk4_jar_override,
-            gatk_docker = gatk_docker,
-            mem_gb = mem_gb_for_collect_counts,
-            disk_space_gb = collect_counts_normal_disk,
-            preemptible_attempts = preemptible_attempts
+    if (defined(normal_bam)) {
```
samuelklee: Anything stopping you from putting all of the tasks for the normal in a single `if` block?  (If so, no biggie.)

scripts/cnv_cromwell_tests/somatic/run_cnv_somatic_workflows.sh
LeeTL1220: Done

scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl
LeeTL1220: @samuelklee I just wanted to preserve the order.  My logic was simply so that we keep the like tasks together.  If you would like me to change it, no problem.  Just tell me.

samuelklee: I think it's fine to put them all in the same block.  At some point, we can extract a single-sample subworkflow that takes allelic counts from a matched normal as an optional input to the ModelSegments task (since that is the only difference between running the normal and the tumor).  This pair workflow will just be two calls to that single-sample workflow.  However, this can definitely wait until Cromwell handles subworkflows more elegantly.

LeeTL1220:
@samuelklee Put the normal sample processing into one `if` block.

PR #4415, opened 2018-02-14
make FASTQ reader remove phred bias from quals

tedsharpe:


tedsharpe:
OK.  That would be the right thing to do.

tedsharpe:
added round-trip test that fails without the fix.
responded to rename request with explicit documentation instead.

Issue #958, opened 2015-10-07
Create URI class for tool inputs/outputs (or use java.nio.file.Path)


droazen:
We desperately need a URI abstraction for tool inputs/outputs to enforce that all tools can read/write to/from all supported sources (files, hdfs, and GCS).

droazen:
This should be used in both spark tools AND walkers, ideally.

droazen:
@tomwhite suggested looking into the possibility of using java.nio.file.Path + plugins for Hadoop and Google buckets to satisfy this ticket.

droazen:
Closing -- this was done.

PR #4416, opened 2018-02-15
Replace literal arguments with variables in several integration tests

davidbenjamin:
@takutosato Here's a bit of mainly-M2-related tech debt, nothing deep.

@lbergelson would you mind skimming the last two commits and telling me if absorbing the literals into `toolsTestDir` is objectionable?

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/GATKBaseTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/GATKBaseTest.java
@@ -30,6 +30,7 @@ public abstract class GATKBaseTest extends BaseTest {
 
     public static final String packageRootTestDir = publicTestDir + "org/broadinstitute/hellbender/";
     public static final String toolsTestDir = packageRootTestDir + "tools/";
+    public static final String exampleTestDir = toolsTestDir + "examples/";
 
     public static final String GCS_GATK_TEST_RESOURCES = "gs://hellbender/test/resources/";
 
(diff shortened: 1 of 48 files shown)


src/test/java/org/broadinstitute/hellbender/GATKBaseTest.java
```diff
@@ -30,6 +30,7 @@
 
     public static final String packageRootTestDir = publicTestDir + "org/broadinstitute/hellbender/";
     public static final String toolsTestDir = packageRootTestDir + "tools/";
+    public static final String exampleTestDir = toolsTestDir + "examples/";
```
pshapiro4broad: Someday it would be nice to modernize this code by using `File` (or, even better, `Path`) for all filesystem references. Then we can replace string concatenation in code like this `Path.resolve()`.

Also, this should use all uppercase for static finals like the rest of our code does.

droazen: Not all of the paths in this class are local files, though (some are GCS URIs), so we don't want to switch to `File` here. Ultimately we plan to unify around a "GATK URI" class that can handle all supported inputs/outputs to the GATK (including NIO `Path`-supported filesystems such GCS, as well as types not supported via NIO such as `gendb://`): https://github.com/broadinstitute/gatk/issues/958. Until this materializes continued use of `String` is the recommended practice, for now.

PR #4423, opened 2018-02-16
Adding oncotator filtering enabled in M2 WDL.

### LINKED ISSUE #4422: M2 Oncotator WDL needs to be able to filter variants in the VCF before rendering the MAF
Many users do not want mutations filtered in the VCF to show up in the MAF file.

Two things:

- Update the default M2 WDL to use `broadinstitute/oncotator:1.9.7.0` docker image.

- Add an optional flag to prune filtered mutations to the Oncotator task.  If this is set to True, add the following to the oncotator command: `--collapse-filter-cols  --prune-filter-cols`

This is unnecessary for the CNV WDL.

LeeTL1220:
Fixes #4422


LeeTL1220:
@davidbenjamin I also ran an additional test outside of our repo to make sure that this was working properly.

LeeTL1220:
@brigranger @jnkts Note when this gets into master and gets released.  To keep the previous behavior, you will want to set `filter_oncotator_maf` to false.

Issue #4367, opened 2018-02-07
Tools marked as @BetaFeature/@ExperimentalFeature should output a warning when run


droazen:
The warning should state that the tool is in a beta/experimental state, and should not be used for production work.

droazen:
@cmnbroad Could you get a PR in for this by Friday, if possible? I'd like to include it in the next point release.

cmnbroad:
I'll have something shortly.

cmnbroad:
Reopening as a placeholder since we really need t do this in Picard as well.

droazen:
Closing -- this was done.

PR #4429, opened 2018-02-20
Display a warning message for Beta and Exerimental tools.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/4367. Includes a sleep for 5 seconds; not sure thats a good idea, although the warning message scrolls off the screen pretty quickly, especially with Spark apps. This will need to be replicated in Picard as well.

Experimental:

screen shot 2018-02-20 at 3 37 06 pm

Beta:

screen shot 2018-02-20 at 3 37 22 pm

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -174,7 +174,7 @@ public abstract class CommandLineProgram implements CommandLinePluginProvider {
             printStartupMessage(startDateTime);
         }
 
-        printProductionWarning();
+        warnOnToolStatus();
 
         try {
             return runTool();
@@ -285,7 +285,7 @@ public abstract class CommandLineProgram implements CommandLinePluginProvider {
      * @param useTerminalColor true if the message should include highlighting terminal colorization
      * @return a warning message if the tool is Beta or Experimental, otherwise null
      */
-    protected String getProductionWarning(final boolean useTerminalColor) {
+    protected String getToolStatusWarning(final boolean useTerminalColor) {
         final String KNRM = "\u001B[0m"; // reset
         final String BOLDRED = "\u001B[1m\u001B[31m";
         final int BORDER_LENGTH = 60;
@@ -317,8 +317,8 @@ public abstract class CommandLineProgram implements CommandLinePluginProvider {
     /**
      * If a tool is either Experimental or Beta, log a warning against use in production a environment.
      */
-    protected void printProductionWarning() {
-        final String warningMessage = getProductionWarning(!(System.console() == null));
+    protected void warnOnToolStatus() {
+        final String warningMessage = getToolStatusWarning(true);
         if (warningMessage != null) {
             logger.warn(warningMessage);
         }
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -14,6 +14,7 @@ import htsjdk.samtools.util.Log;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.barclay.argparser.*;
+import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.utils.LoggingUtils;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.config.ConfigFactory;
@@ -24,10 +25,9 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetAddress;
-import java.net.MalformedURLException;
 import java.net.URL;
-import java.net.URLClassLoader;
 import java.text.DecimalFormat;
+import java.lang.Thread;
 import java.time.Duration;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
@@ -175,6 +175,8 @@ public abstract class CommandLineProgram implements CommandLinePluginProvider {
             printStartupMessage(startDateTime);
         }
 
+        printProductionWarning();
+
         try {
             return runTool();
         } finally {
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -278,6 +280,63 @@ protected void printStartupMessage(final ZonedDateTime startDateTime) {
...
+
+    /**
+     * If a tool is either Experimental or Beta, log a warning against use in production environment.
+     */
+    protected void printProductionWarning() {
+        final String warningMessage = getProductionWarning();
+        if (warningMessage != null) {
+            logger.warn(warningMessage);
+            try {
+                Thread.sleep(5000);
```
lbergelson: 5 seconds of sleep is going to drive me insane.  .5 seems much more tolerable but still noticeable.  We might want to add a command line option to disable the sleep and apply that to all the tests

droazen:
Let's remove the sleep delay @cmnbroad. Red text + a border made out of exclamation points should be sufficient.

lbergelson:
The one problem with using color in the output is that it makes it gross to read logs that aren't on an interactive terminal that supports color.  We might want to make an attempt to tell if we're connected to a terminal or an output file and remove the colors if so.

lbergelson:
I have this class from an ancient branch that makes dealing with colors nicer in some ways.  It tries to avoid printing colors to non-interactive things and it makes it harder to forget a reset:

```
/**
 * Provides ANSI colors for the terminal output *
 */
public final class TerminalColors {

    private TerminalColors(){};

    private enum TerminalColor{
        CYAN("\u001B[36m"),
        RED("\u001B[31m"),
        GREEN("\u001B[32m"),
        WHITE("\u001B[37m"),
        BOLD("\u001B[1m"),
        RESET("\u001B[0m"); // reset the colors

        private final String color;

        TerminalColor(String color){
            this.color = color;
        }

        public String getColorString(){
            return color;
        }

    }

    public static boolean isInteractive(){
        return !(System.console() == null);
    }

    public static String cyan(String toColor){
        return colorString(toColor, TerminalColor.CYAN);
    }

    public static String red(String toColor){
        return colorString(toColor, TerminalColor.RED);
    }

    public static String green(String toColor){
        return colorString(toColor, TerminalColor.GREEN);
    }

    public static String white(String toColor){
        return colorString(toColor, TerminalColor.WHITE);
    }

    public static String bold(String toBold){
        return colorString(toBold, TerminalColor.BOLD);
    }

    public static String colorString(String toColor, TerminalColor color) {
        if(isInteractive()) {
            return color.getColorString() + toColor + TerminalColor.RESET.getColorString();
        } else {
            return toColor;
        }
    }

    public static String stripColorsFromString(String colorString){
        String stripped = colorString;
        for(TerminalColor color : TerminalColor.values()) {
            stripped = stripped.replace(color.getColorString(),"");
        }
        return stripped;
    }
}
```

cmnbroad:
@lbergelson What do you think about putting the TerminalColors class in Barclay ? The same changes I'm making here need to be made in Picard. We could put a lot of this behavior into a shared CommandLineProgram base class that uses TerminalColors.

lbergelson:
@cmnbroad Barclay sounds like the best place for things that have to deal with the command line.  Feel free to use that class or not.  I'm not sure it's the greatest way to deal with colors, but it seems better than manually adding the escape codes everywhere.

magicDGS:
Maybe https://github.com/fusesource/jansi to handle colors?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -278,6 +279,61 @@ protected void printStartupMessage(final ZonedDateTime startDateTime) {
...
+                    useTerminalColor ? KNRM : ""
+            );
+        }
+        return warningMessage;
+    }
+
+    /**
+     * If a tool is either Experimental or Beta, log a warning against use in production a environment.
+     */
+    protected void printProductionWarning() {
```
droazen: Maybe `warnAboutToolStatus()` would be clearer?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -278,6 +279,61 @@ protected void printStartupMessage(final ZonedDateTime startDateTime) {
...
+        final String warningMessage = getProductionWarning(!(System.console() == null));
```
droazen: If this check isn't 100% reliable, I would take it out. Printing the literal codes in non-interactive use is not that big of a deal.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -278,6 +279,61 @@ protected void printStartupMessage(final ZonedDateTime startDateTime) {
         catch (final Exception e) { /* Unpossible! */ }
     }
 
+    /**
+     * If this tool is either Experimental or Beta, return a warning message advising against use in production
+     * environment.
+     * @param useTerminalColor true if the message should include highlighting terminal colorization
+     * @return a warning message if the tool is Beta or Experimental, otherwise null
+     */
+    protected String getProductionWarning(final boolean useTerminalColor) {
```
droazen: `getToolStatusWarning()` might be clearer?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
cmnbroad: Done.

PR #4430, opened 2018-02-20
VariantContextTestUtils: add a utility method to slurp an entire VCF into a collection

droazen:
This method is for unit/integration testing purposes only, and should not be called from tools.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
@@ -137,6 +137,17 @@ public final class FeatureDataSource<T extends Feature> implements GATKDataSourc
     }
 
     /**
+     * Creates a FeatureDataSource backed by the provided path. The data source will have an automatically
+     * generated name, and will look ahead the default number of bases ({@link #DEFAULT_QUERY_LOOKAHEAD_BASES})
+     * during queries that produce cache misses.
+     *
+     * @param featurePath path or URI to source of Features
+     */
+    public FeatureDataSource(final String featurePath) {
+        this(featurePath, null, DEFAULT_QUERY_LOOKAHEAD_BASES, null);
+    }
+
+    /**
      * Creates a FeatureDataSource backed by the provided File and assigns this data source the specified logical
      * name. We will look ahead the default number of bases ({@link #DEFAULT_QUERY_LOOKAHEAD_BASES}) during queries
      * that produce cache misses.
--- a/src/main/java/org/broadinstitute/hellbender/utils/test/VariantContextTestUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/test/VariantContextTestUtils.java
@@ -10,6 +10,8 @@ import com.google.common.collect.HashBiMap;
 import htsjdk.variant.variantcontext.*;
 import htsjdk.variant.vcf.*;
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.tuple.Pair;
+import org.broadinstitute.hellbender.engine.FeatureDataSource;
 import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.tools.walkers.genotyper.AlleleSubsettingUtils;
 import org.broadinstitute.hellbender.tools.walkers.genotyper.GenotypeAssignmentMethod;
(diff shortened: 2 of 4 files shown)


droazen:
@jamesemery please review

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
```diff
@@ -136,6 +136,17 @@ public FeatureDataSource(final File featureFile) {
...
     }
 
+    /**
+     * Creates a FeatureDataSource backed by the provided path. The data source will have an automatically
+     * generated name, and will look ahead the default number of bases ({@link #DEFAULT_QUERY_LOOKAHEAD_BASES})
+     * during queries that produce cache misses.
+     *
+     * @param featurePath path or URI to source of Features
+     */
+    public FeatureDataSource(final String featurePath) {
```
jamesemery: I see the purpose of this constructor but it seems a little overzealous to add this easy constructor to only use it for one test utility method. you may disagree though so I will leave it up to you.

src/main/java/org/broadinstitute/hellbender/utils/test/VariantContextTestUtils.java
```diff
@@ -22,14 +24,40 @@
...
+     * For unit/integration testing purposes only! Do not call this method from actual tools!
+     *
+     * @param vcfPath path or URI to a VCF, as a String
+     * @return A Pair with the VCFHeader as the first element, and a List of all VariantContexts from the VCF
+     *         as the second element
+     */
+    public static Pair<VCFHeader, List<VariantContext>> readEntireVCFIntoMemory(final String vcfPath) {
+        Utils.nonNull(vcfPath);
+
+        try ( final FeatureDataSource<VariantContext> vcfReader = new FeatureDataSource<>(vcfPath) ) {
```
jamesemery: What happens if it fails during the try instantiation of this reader when it points to a non-existant file? Does it throw an IOException if it can't read the file in the first place because of a bad path?

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
droazen: The new constructor seems generally useful, as currently if you want to support non-local files you have to call the 4-arg constructor, which is a bit cumbersome.

src/main/java/org/broadinstitute/hellbender/utils/test/VariantContextTestUtils.java
droazen: That case is handled by `FeatureDataSource` itself (it throws a `UserException.CouldNotReadInputFile`). I'll add a quick test to cover that case.

jamesemery:
@droazen looks good. Merge when the tests pass

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
lbergelson: should we add a Path constructor to while we're here?

src/main/java/org/broadinstitute/hellbender/utils/test/VariantContextTestUtils.java
```diff
@@ -22,14 +24,40 @@
...
+     * Reads an entire VCF into memory, returning both its VCFHeader and all VariantContext records in
+     * the vcf. Supports both local files and NIO-supported remote filesystems such as GCS.
+     *
+     * For unit/integration testing purposes only! Do not call this method from actual tools!
+     *
+     * @param vcfPath path or URI to a VCF, as a String
+     * @return A Pair with the VCFHeader as the first element, and a List of all VariantContexts from the VCF
+     *         as the second element
+     */
+    public static Pair<VCFHeader, List<VariantContext>> readEntireVCFIntoMemory(final String vcfPath) {
```
lbergelson: Returning this pair seems ugly to me.  Maybe better to return a little struct that has a `getVariants` and `getHeader`?  I hate seeing Pairs all over the place because the syntax is so nasty in java.

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
droazen: Nah, let's just add the one we need for this PR. We'll likely refactor all these constructors anyway when we get around to https://github.com/broadinstitute/gatk/issues/958

src/main/java/org/broadinstitute/hellbender/utils/test/VariantContextTestUtils.java
droazen: I think the ugliness is bearable, and outweighed by the convenience of getting the header as well. Introducing a new class used only by this method seems like a worse option to me.

PR #4431, opened 2018-02-20
refactoring in CalculateGenotypePosteriors to remove confusion

lbergelson:
refactoring code to remove confusion and unnecessary call to calculateChromosomeCounts


ldgauthier:
Nice cleanup, thanks!

Also, yesterday I verified that passing in -V <hg38VCF> -supporting <b37VCF> does throw a user error:
`A USER ERROR has occurred: Badly formed genome unclippedLoc: Contig '1' does not match any contig in the GATK sequence dictionary derived from the reference; are you sure you are using the correct reference fasta file?`
I don't feel like the error text is entirely accurate, but at least no one is able to anything too crazy.

lbergelson:
Hmn.  That's not the greatest error.  It should be failing something more along the lines of "Your sequence dictionaries don't match, please make sure your files are all aligned using the same reference"

PR #4432, opened 2018-02-21
In StrandBiasTest, use instanceof and support int[].

fnothaft:


ldgauthier:
@fnothaft Thanks for the cleanup.  I've seen that annotations take on different types depending on whether they were added during the current traversal or read in from a file, but I haven't run into any issues with the strand bias annotations so far.  Do you have a specific use case for the int[] version?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandBiasTest.java
```diff
@@ -73,14 +73,17 @@
...
-            } else if (g.getAnyAttribute(GATKVCFConstants.STRAND_BIAS_BY_SAMPLE_KEY).getClass().equals(ArrayList.class)) {
+            } else if (sbbsObj instanceof List) {
                 @SuppressWarnings("unchecked")
-                final List<Integer> sbbsList = (ArrayList<Integer>) g.getAnyAttribute(GATKVCFConstants.STRAND_BIAS_BY_SAMPLE_KEY);
+                final List<Integer> sbbsList = (ArrayList<Integer>) sbbsObj;
                 data = encodeSBBS(sbbsList);
-            } else {
+            } else if (sbbsObj instanceof int[]) {
+		data = (int[])sbbsObj;
+	    } else {
```
lbergelson: extremely nitpicky, but the indentation here is off a tiny bit

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandBiasTest.java
```diff
@@ -73,14 +73,17 @@
...
             foundData = true;
             int[] data;
-            if ( g.getAnyAttribute(GATKVCFConstants.STRAND_BIAS_BY_SAMPLE_KEY).getClass().equals(String.class)) {
-                final String sbbsString = (String)g.getAnyAttribute(GATKVCFConstants.STRAND_BIAS_BY_SAMPLE_KEY);
+	    Object sbbsObj = g.getAnyAttribute(GATKVCFConstants.STRAND_BIAS_BY_SAMPLE_KEY);
+            if (sbbsObj instanceof String) {
+                final String sbbsString = (String)sbbsObj;
                 data = encodeSBBS(sbbsString);
-            } else if (g.getAnyAttribute(GATKVCFConstants.STRAND_BIAS_BY_SAMPLE_KEY).getClass().equals(ArrayList.class)) {
+            } else if (sbbsObj instanceof List) {
```
droazen: `instanceof` can be much more expensive than `getClass()` -- we should check if there's any noticeable effect on performance here before merging.

droazen:
@fnothaft When you submit a PR, you should include a brief description of the motivation behind the  change, as well as a good unit test covering it.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/StrandBiasTest.java
lbergelson: I would argue that instanceof is actually the correct check here and comparing class is a subtle bug that will cause confusing issues if we ever encounter a different list type...

droazen: I disagree @lbergelson -- notice that there is an unconditional cast to `ArrayList` just below anyway. And `String` is a `final` class, so the `instanceof` check there has no advantage over `getClass()`.

droazen:
Closing this one, as it's been open for a year and there still isn't a test + description of the motivation behind the change.

Issue #4329, opened 2018-02-01
GATK should preserve Picard tool exit codes 


jonn-smith:
When running a Picard tool within GATK, the exit code of that Picard tool is not preserved when GATK exits.

For example, running `ExtractSequences` with no arguments produces an error and the Picard return code of `1`, but GATK returns a 0.

This seems to be around `Main.java:196` (`mainEntry`).

magicDGS:
Maybe the way is to check the returned value inside `PicardCommandLineProgramExecutor` and throw a `PicardCommandNonZeroExit` exception containing a field with the exit code, that can be handled in `Main` with a special function (as user exception, etc.).

Tests for checking the exit code returned by `mainEntry` can be found in https://github.com/broadinstitute/gatk/pull/4283.

droazen:
That sounds like a good way to do it @magicDGS

Issue #4433, opened 2018-02-21
Picard tools report 0 for exit code even if they fail


chandrans:
I think this user report sums it up nicely.

----
User Report
----

In my BASH scripts I often use "$?" to monitor the exit status of a process and normally stop if there is an error. However, I am using the latest version of GATK (4.0.0.0) and some tools return 0 exit status even if they fail. Instead, they display the following message to STDOUT:
    
    Tool returned:
    1

Though inconvenient for error handling in BASH scripts, this might be an intended behaviour, but not all tools exhibit it. To mention a few, MarkDuplicates, CollectMultipleMetrics, CollectGcBiasMetrics always have a 0 exit status, whereas VariantsToTable or CountVariants do return 1 if they encounter an error. 

A similar issue had been reported in the past for previous versions of GATK (https://gatkforums.broadinstitute.org/gatk/discussion/8618/error-handling-end-exit-codes-in-gatk).

Best regards,

Roger

This Issue was generated from your [forums] 
[forums]: https://gatkforums.broadinstitute.org/gatk/discussion/11414/exit-codes-in-gatk-4-0/p1

cmnbroad:
Yes, the return value for Picard tools are not propagated properly when run through GATK. We'll try to get the fix in for the next release. Closing as a dup of https://github.com/broadinstitute/gatk/issues/4329.

chandrans:
Thanks Chris. Did not realize this was a duplicate.

PR #4437, opened 2018-02-22
Propagate non-zero Picard tool exit codes.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/4329 and https://github.com/broadinstitute/gatk/issues/4433.


src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -201,6 +202,10 @@ protected final void mainEntry(final String[] args) {
             }
             handleUserException(e);
             System.exit(COMMANDLINE_EXCEPTION_EXIT_VALUE);
+        } catch (PicardNonZeroExitException e) {
```
magicDGS: final

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -201,6 +202,10 @@ protected final void mainEntry(final String[] args) {
...
+            // a Picard tool returned a non-zero exit code
+            handleResult(e.getToolReturnCode());
```
magicDGS: Why should the result be printed on the command line?

cmnbroad: To retain the Picard return code, which may be different than the one returned by GATK. We could make it return the actual Picard value directly, but that might be misleading since those values are not drawn from the same space as the current GATK values.

src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -201,6 +202,10 @@ protected final void mainEntry(final String[] args) {
             }
             handleUserException(e);
             System.exit(COMMANDLINE_EXCEPTION_EXIT_VALUE);
+        } catch (final PicardNonZeroExitException e) {
+            // a Picard tool returned a non-zero exit code
+            handleResult(e.getToolReturnCode());
+            System.exit(ANY_OTHER_EXCEPTION_EXIT_VALUE);
```
droazen: I think it would be better to have a separate exit status code for the case where a Picard tool exits with non-0, so that clients will know to check the log to get the underlying Picard exit code.

src/main/java/org/broadinstitute/hellbender/exceptions/PicardNonZeroExitException.java
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     *
+     * @param picardToolReturnCode the non-zero return code returned from the Picard tool
+     */
+    public PicardNonZeroExitException(final int picardToolReturnCode) {
+        this.picardToolReturnCode = picardToolReturnCode;
+        Utils.validateArg(picardToolReturnCode != 0, "Return code must be non-zero");
+    }
+
+    public int getToolReturnCode() { return picardToolReturnCode; }
```
droazen: javadoc for this method

src/test/java/org/broadinstitute/hellbender/MainTest.java
```diff
@@ -94,4 +94,17 @@ public void testMainErrorWithoutStackTrace() {
             System.setSecurityManager(backup);
         }
     }
+    @Test(singleThreaded = true)
+    public void testNonZeroPicardReturnValue() {
+        final SecurityManager backup = System.getSecurityManager();
+        try {
+            System.setSecurityManager(new ThrowOnExitSecurityManager());
+            new Main().mainEntry(new String[]{"ExtractSequences"});
```
droazen: Check the return value of `mainEntry()` to be sure that it's equal to the expected value

src/main/java/org/broadinstitute/hellbender/Main.java

src/main/java/org/broadinstitute/hellbender/exceptions/PicardNonZeroExitException.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/MainTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/Main.java
magicDGS: I though that the idea was to **exit** with the same return code as Picard, not to print out the Picard return code and then exit with a different one. For example, for `ValidateSamFile` the actual exit value has a meaning (see https://github.com/broadinstitute/picard/blob/b9b87773e9339ef6be7d61096adfc7f074b5aa74/src/main/java/picard/sam/ValidateSamFile.java#L174).

That's why I put this comment, but if that was not the idea it's fine too.

cmnbroad: The first order concern was to make sure that failure resulted in a non-zero return code. Otherwise, as mentioned above, the values themselves are hard not that reliable - I don't think Picard is even self-consistent, since ValidateSamFile returns 1 if there are validation warnings or if there is a command line parsing error.

PR #4451, opened 2018-02-23
adding missing registrations in kryo

lbergelson:
* some classes were missing registration in kryo which causes less efficient serialization
* adding registrations for a number of classes that MarkDuplicatesSpark needs that weren't registered yet

* notably, BAMRecord wasn't registered to use the correct serializer which could cause major inefficiencies
* it's not clear what circumstances we're serializing BAMRecord instead of SAMRecordToGATKReadAdapter so how much this will help is not obvious

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/spark/GATKRegistrator.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/spark/GATKRegistrator.java
@@ -21,7 +21,6 @@ public class GATKRegistrator implements KryoRegistrator {
 
     public GATKRegistrator() {}
 
-    @SuppressWarnings({"unchecked", "rawtypes"})
     @Override
     public void registerClasses(Kryo kryo) {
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/engine/spark/GATKRegistrator.java
```diff
@@ -43,6 +43,13 @@ public void registerClasses(Kryo kryo) {
         kryo.register(SAMRecordToGATKReadAdapter.class, new SAMRecordToGATKReadAdapterSerializer());
 
         kryo.register(SAMRecord.class, new SAMRecordSerializer());
+        kryo.register(BAMRecord.class, new SAMRecordSerializer());
+
+        kryo.register(SAMFileHeader.class);
```
droazen: This collides (admittedly harmlessly, in this case) with a registration for the same class in the ADAM serializer. I think we have to deal with the issue of registration precedence in this branch -- can you find a way to ensure that the GATK registrations always take precedence over any registrations for the same classes in our dependencies?

PR #4455, opened 2018-02-23
HaplotypeCaller: fix -contamination argument, and add tests

### LINKED ISSUE #4312: HaplotypeCaller failing when -contamination argument is supplied
I believe that the problem is that HaplotypeCallerGenotypingEngine doesn't use the version of StandardCallerArgumentCollection.getSampleContamination() with the sampleID, so the sampleContamination variable is never initialized and we get a NPE.  The -contamination argument is standard in our production GATK pipeline... and we're about to start telling everyone to use it!  This is an important one to fix.

Here's a stacktrace:

	at java.util.Collections$UnmodifiableMap.<init>(Collections.java:1446)
	at java.util.Collections.unmodifiableMap(Collections.java:1433)
	at org.broadinstitute.hellbender.tools.walkers.genotyper.StandardCallerArgumentCollection.getSampleContamination(StandardCallerArgumentCollection.java:89)
	at org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCallerGenotypingEngine.assignGenotypeLikelihoods(HaplotypeCallerGenotypingEngine.java:141)
	at org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCallerEngine.callRegion(HaplotypeCallerEngine.java:566)
	at org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCaller.apply(HaplotypeCaller.java:218)
	at org.broadinstitute.hellbender.engine.AssemblyRegionWalker.processReadShard(AssemblyRegionWalker.java:295)

droazen:
The -contamination argument was not hooked up properly in the HaplotypeCaller.
This patch fixes the tool argument, and adds tests on artificially contaminated
data to demonstrate that the feature works as intended.

Resolves #4312


src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
```diff
@@ -571,10 +577,127 @@ public void testMinBaseQualityScore() throws Exception {
...
+                "-L", interval.toString(),
+                "-O", output.getAbsolutePath(),
+                "-contamination", Double.toString(contaminationFraction),
+                "-ERC", (gvcfMode ? "GVCF" : "NONE")
+        };
+
+        runCommandLine(contaminationArgs);
+
+        // With contamination downsampling turned on, we should get at least expectedMinimumConcordance
+        // concordance with the calls the uncontaminated bam (ignoring reference blocks in GVCF mode)
```
jonn-smith: with the calls [to the] ....

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
```diff
@@ -571,10 +577,127 @@ public void testMinBaseQualityScore() throws Exception {
...
+        // calls on the uncontaminated bam
+        // expected concordance vs. the calls on the uncontaminated bam
+        return new Object[][] {
+                { contaminatedBam15Percent,
+                  0.15,
+                  traversalInterval,
+                  b37_reference_20_21,
+                  false, // VCF mode
+                  TEST_FILES_DIR + "expected.CEUTrio.HiSeq.WGS.b37.NA12878.calls.20.10000000-10250000.vcf",
+                  // Expect > 90% concordance vs. uncontaminated calls
```
jonn-smith: I'm curious - why would it not be ~85% without the contamination argument given that the test data are 15% contaminated?

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
```diff
@@ -571,10 +577,127 @@ public void testMinBaseQualityScore() throws Exception {
...
+                  // (NOTE: This drops to 74.49% concordance WITHOUT the -contamination argument, when last tested)
+                  0.90
+                },
+                { contaminatedBam15Percent,
+                  0.15,
+                  traversalInterval,
+                  b37_reference_20_21,
+                  true, // GVCF mode
+                  TEST_FILES_DIR + "expected.CEUTrio.HiSeq.WGS.b37.NA12878.calls.20.10000000-10250000.g.vcf",
+                  // Expect > 80% concordance vs. uncontaminated calls
```
jonn-smith: Same question for this case.

jonn-smith:
@droazen Back at you with a couple of questions about the test cases / explanations and one minor grammatical edit.

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
droazen: Well, 15% of the reads come from another sample. Depending on how they are distributed, they could easily end up affecting more than (or less than) 15% of the variant calls.

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
droazen: See explanation above

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
droazen: Fixed

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
jonn-smith: Ah, OK.  I understand now.  Thanks.

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
jonn-smith: Got it.

droazen:
I've added a bunch of additional unit tests for the `StandardCallerArgumentCollection`, and beefed up the `HaplotypeCaller` integration tests as well (we now check for concordance against the GATK3 contamination-corrected calls, in addition to just asserting that the calls in GATK4 improve with contamination correction on).

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/StandardCallerArgumentCollection.java
```diff
@@ -72,37 +72,75 @@ public void copyStandardCallerArgsFrom( final StandardCallerArgumentCollection o
...
-    public Double getSampleContamination(final String sampleId){
-        Utils.nonNull(sampleId);
-        if (sampleContamination == null){
-            setSampleContamination(new DefaultedMap<>(CONTAMINATION_FRACTION));//default to empty map
-        }
-        return sampleContamination.get(sampleId);
-    }
-
-    public void setSampleContamination(final DefaultedMap<String, Double> sampleContamination) {
+    public void setSampleContamination(final Map<String, Double> sampleContamination) {
```
jamesemery: Why does this code belong to the argument collection? I understand it is needs access to the argument collection values in order to fill defaults, but this code is not a user argument, indeed it needs to be called into by the tool itself in order to poke contamination back into its own argument collection? Could this logic live elsewhere?

src/test/java/org/broadinstitute/hellbender/utils/downsampling/AlleleBiasedDownsamplingUtilsUnitTest.java
```diff
@@ -179,8 +179,8 @@ private static void testLoadFile(final File file, final Set<String> Samples, fin
     }
 
     @Test(dataProvider = "goodContaminationFiles")
-    public void testLoadContaminationFile(final Integer ArtificalBAMnumber, final Integer numberOfSamples) throws IOException {
-        final String ArtificialBAM = String.format("contamination.case.%d.txt", ArtificalBAMnumber);
+    public void testLoadContaminationFile(final int artificalBAMnumber, final int numberOfSamples) throws IOException {
```
jamesemery: How did these changes make it into this branch?

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
```diff
@@ -571,10 +577,214 @@ public void testMinBaseQualityScore() throws Exception {
...
+    public Object[][] getContaminationCorrectionTestData() {
+        // This bam is a snippet of GATKBaseTest.NA12878_20_21_WGS_bam artificially contaminated
+        // at 15% with reads from another sample. We use such a high contamination fraction to ensure
+        // that calls will be noticeably different vs. running without -contamination at all.
+        final String contaminatedBam15Percent = largeFileTestDir + "contaminated_bams/CEUTrio.HiSeq.WGS.b37.NA12878.CONTAMINATED.WITH.HCC1143.NORMALS.AT.15PERCENT.bam";
+
+        final SimpleInterval traversalInterval = new SimpleInterval("20", 10100000, 10150000);
+
+        // File specifying the per-sample contamination for contaminatedBam15Percent
+        final String contaminationFile = TEST_FILES_DIR + "contamination_file_for_CEUTrio.HiSeq.WGS.b37.NA12878.CONTAMINATED.WITH.HCC1143.NORMALS.AT.15PERCENT";
```
jamesemery: This file only lists the contamination for one bam. What happens if there are many fields in the contamination file? or if they don't match with the samples the tool sees?  It seems the there should be tests for these cases, and since the contamination logic seems to be in obfuscated within the haplotype caller and its not tested anywhere.

jamesemery: There do seem to be tests that amount to showing the file doesn't throw, I think there should be some tests somewhere for the edge cases in these files.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/StandardCallerArgumentCollection.java
droazen: Well, there is precedent in the GATK for arg collections containing code to parse/load their arguments (see `IntervalArgumentCollection` for an example). The design of this particular class, however, in which the contamination file must be loaded externally and passed in via a setter, is not great. I'll create a ticket to refactor this in a future PR, but I'd like to get the actual fix out first.

src/test/java/org/broadinstitute/hellbender/utils/downsampling/AlleleBiasedDownsamplingUtilsUnitTest.java
droazen: I was looking at the existing tests covering the contamination downsampling code, and noticed these minor issues (which resulting in errors/warnings in my IDE), so I fixed them.

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
droazen: There are direct unit tests in `AlleleBiasedDownsamplingUtilsUnitTest` covering loading contamination files with multiple samples, as well as tests in `StandardCallerArgumentCollectionUnitTest` for the multi-sample case.

PR #4469, opened 2018-02-28
NIO support for TMP_DIR

magicDGS:
Extracted from #3998


src/main/java/org/broadinstitute/hellbender/tools/walkers/markduplicates/MarkDuplicatesGATK.java
```diff
@@ -358,7 +360,7 @@ private void generateDuplicateIndexes() {
         final int maxInMemory = (int) Math.min((Runtime.getRuntime().maxMemory() * 0.25) / SortingLongCollection.SIZEOF,
                 (double) (Integer.MAX_VALUE - 5));
         logger.info("Will retain up to " + maxInMemory + " duplicate indices before spilling to disk.");
-        this.duplicateIndexes = new SortingLongCollection(maxInMemory, TMP_DIR.toArray(new File[TMP_DIR.size()]));
+        this.duplicateIndexes = new SortingLongCollection(maxInMemory, (Path[]) TMP_DIR.stream().map(IOUtils::getPath).toArray());
```
pshapiro4broad: Instead of a cast, you can provide an argument to `toArray()`, e.g. `TMP_DIR.stream().map(IOUtils::getPath).toArray(Path[]::new)`

magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -60,8 +61,8 @@
     // abstract, this is fine (as long as no logging has to happen statically in this class).
     protected final Logger logger = LogManager.getLogger(this.getClass());
 
-    @Argument(common=true, optional=true)
-    public List<File> TMP_DIR = new ArrayList<>();
+    @Argument(common=true, optional=true, doc = "List of temp directories to use.")
+    public List<String> TMP_DIR = new ArrayList<>();
```
droazen: Can you kebabify this argument as part of this PR (`--temp-dir`)? Looks like we missed this one.

droazen: Actually, we just did a usage check on this argument, and it's ONLY used in `MarkDuplicates`. Given this, @lbergelson and I think that we should just remove this argument completely. Would you be ok with removing it @magicDGS, or are you currently depending on it?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -151,13 +156,33 @@ public Object instanceMainPostParseArgs() {
...
+                        PosixFilePermission.OWNER_WRITE, PosixFilePermission.GROUP_WRITE, PosixFilePermission.OTHERS_WRITE))) {
+                    Files.setPosixFilePermissions(p, permissions);
+                }
+            } catch (final UnsupportedOperationException | IOException e) {
+                // intentionally ignoring
+                // TODO - logging can be removed, but I think that it might be informative (maybe it should be debug)
+                logger.warn("Temp directory {}: unable to set permissions due to {}", p, e.getMessage());
+            }
+            // TODO - this should be p.toAbsolutePath().toUri().toString() to allow other FileSystems to be used (see above)
+            System.setProperty("java.io.tmpdir", p.toAbsolutePath().toString()); // in loop so that last one takes effect
```
droazen: @lbergelson suggests a special case here for file URIs to strip the `file://` prefix.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: I depend on it, but I can move it to the tools that require it.

lbergelson: @droazen Do we think a lot of users use this instead `-Djava.io.tmpdir`?  Maybe it's worth keeping a shortcut too?  Sorry to reconsider.

magicDGS: I was reconsidering it too (that's why I didn't add a commit), because maybe there are other codepaths using teh java property, which is set here, instead of directly using the parameter. Next commit will not have this fix.

magicDGS: But the name was changed on `StandardArgumentDefinitions.TMP_DIR_NAME` after rebase.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: Done in the next commit.

magicDGS:
Rebased to resolve conflicts and addressing comments.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -151,13 +153,32 @@ public Object instanceMainPostParseArgs() {
...
+                    // intentionally ignoring
+                }
+            }
+            try {
+                // add the READ/WRITE permissions to the actual permissions of the file
+                final Set<PosixFilePermission> permissions = Files.getPosixFilePermissions(p);
+                // only set the permissions if they change
+                if (permissions.addAll(Arrays.asList(
+                        PosixFilePermission.OWNER_READ, PosixFilePermission.GROUP_READ, PosixFilePermission.OTHERS_READ,
+                        PosixFilePermission.OWNER_WRITE, PosixFilePermission.GROUP_WRITE, PosixFilePermission.OTHERS_WRITE))) {
```
magicDGS: By the way, this is related with https://github.com/broadinstitute/gatk/issues/4513 - should probably resolve here.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -60,8 +61,8 @@
     // abstract, this is fine (as long as no logging has to happen statically in this class).
     protected final Logger logger = LogManager.getLogger(this.getClass());
 
-    @Argument(fullName = StandardArgumentDefinitions.TMP_DIR_NAME, common=true, optional=true)
-    public List<File> TMP_DIR = new ArrayList<>();
+    @Argument(fullName = StandardArgumentDefinitions.TMP_DIR_NAME, common=true, optional=true, doc = "List of temp directories to use.")
+    public List<String> TMP_DIR = new ArrayList<>();
```
droazen: If we're going to keep this argument in GATK, can you make the type into a `String` instead of `List<String>`, since it's not doing anything intelligent right now with the List?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -151,13 +153,32 @@ public Object instanceMainPostParseArgs() {
...
+                }
+            }
+            try {
+                // add the READ/WRITE permissions to the actual permissions of the file
+                final Set<PosixFilePermission> permissions = Files.getPosixFilePermissions(p);
+                // only set the permissions if they change
+                if (permissions.addAll(Arrays.asList(
+                        PosixFilePermission.OWNER_READ, PosixFilePermission.GROUP_READ, PosixFilePermission.OTHERS_READ,
+                        PosixFilePermission.OWNER_WRITE, PosixFilePermission.GROUP_WRITE, PosixFilePermission.OTHERS_WRITE))) {
+                    Files.setPosixFilePermissions(p, permissions);
```
droazen: Can you get rid of this code to change the permissions on the temp dir, and throw a `UserException` instead if the temp dir is not writable/readable?

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -557,6 +557,17 @@ public static Path getPath(String uriString) {
...
     }
 
+    /**
+     * Gets the absolute Path name with the URI marker, handling the special case of the default file system by removing
+     * the file:// prefix.
+     *
+     * @param path path to get the absolute name.
+     * @return a String with the absolute name.
+     */
+    public static String getAbsolutePathName(final Path path) {
```
droazen: The name of this method doesn't really tell you what it's doing -- can you rename to `getAbsolutePathWithoutFileProtocol()`?

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -557,6 +557,17 @@ public static Path getPath(String uriString) {
         }
     }
 
+    /**
+     * Gets the absolute Path name with the URI marker, handling the special case of the default file system by removing
+     * the file:// prefix.
+     *
+     * @param path path to get the absolute name.
+     * @return a String with the absolute name.
```
droazen: "a String with the absolute name, and the file:// protocol removed, if it was present"

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -151,13 +153,32 @@ public Object instanceMainPostParseArgs() {
...
             // need a tmp_dir. If this fails, the problem will be discovered downstream.
-            if (!f.exists()) f.mkdirs();
-            f.setReadable(true, false);
-            f.setWritable(true, false);
-            System.setProperty("java.io.tmpdir", f.getAbsolutePath()); // in loop so that last one takes effect
+            if (!Files.exists(p)) {
+                try {
+                    Files.createDirectories(p);
+                } catch (IOException e) {
+                    // intentionally ignoring
```
droazen: Don't create the temp dir if it doesn't exist -- just throw a `UserException` in this case

src/test/java/org/broadinstitute/hellbender/utils/io/IOUtilsUnitTest.java
```diff
@@ -196,6 +196,20 @@ private void innerTestGetPath(String s) throws IOException {
...
+                {"file:///local/example.txt", "/local/example.txt"}
+        };
+    }
+
+    @Test(dataProvider = "absoluteNames")
+    public void testGetAbsolutePathName(final String uriString, final String expected) {
+        Assert.assertEquals(IOUtils.getAbsolutePathName(IOUtils.getPath(uriString)), expected);
+    }
+
     @Test
```
droazen: Can you add a unit test to prove that the temp dir works over NIO?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: If I modify that, it changes the behavior of `MarkDuplicatesGATK` and `EstimateLibraryComplexityGATK` (I guess that they are setting the temporary directory properly)

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: I personally like specifying a non-existent directory for temporary files that is created by the program instead of explicitly using `mkdirs` - anyway, I can change in the next commit.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
droazen: We can add a separate argument to those tools if they really need to take a List of temporary directories. But having the GATK-wide argument take a List and then ignore all but the last element is confusing.

magicDGS: True. Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: Done, although in that case is a change of behavior - if the temp dir is not used at all, this can fail for every tool...

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
magicDGS: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: Removed in the next commit.

src/test/java/org/broadinstitute/hellbender/utils/io/IOUtilsUnitTest.java
magicDGS: Which kind of test do you need there? Retrieving the property and writing to it? Or running tools using it with NIO (e.g., `MarkDuplicatesGATK`)? Let me know for next round of comments...

magicDGS:
Back to you, still without tests for tmp dir working with NIO due to lack of information: should it be at the tool level (e.g., running `MarkDuplicatesGATK` with NIO tmp dir - my main use case) or just checking the property value that is correctly set?

magicDGS:
@droazen - please, clarify how the tests should be done. I'm going to be on vacation next week, so feel free to add them to this branch if you would like to speed up the process (taking into account that #3998 depends on this now). Thanks!

magicDGS:
Friendly ping here @droazen

src/test/java/org/broadinstitute/hellbender/utils/io/IOUtilsUnitTest.java
```diff
@@ -196,6 +196,20 @@ private void innerTestGetPath(String s) throws IOException {
         Assert.assertTrue(size>0);
     }
 
+    @DataProvider
+    public Object[][] absoluteNames() {
+        return new Object[][] {
+                {"/local/example.txt", "/local/example.txt"},
+                {"/local/file://example.txt", "/local/file:/example.txt"},
+                {"file:///local/example.txt", "/local/example.txt"}
```
droazen: Add a test case with a non-file URI (such as `gcs://`)

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
droazen: I think we always use the temp dir in GATK -- if for nothing else, then to extract and load the GKL library for compression/decompression.

src/test/java/org/broadinstitute/hellbender/utils/io/IOUtilsUnitTest.java
droazen: How about a test that:

1. Declares a dummy `CommandLineProgram` that executes `createTempFile()` and then asserts that it exists in its `doWork()` method

2. Runs the above command line program with `--tmp-dir` set to a jimfs URI. You can create a jimfs like this:

```
        try (FileSystem jimfs = Jimfs.newFileSystem(Configuration.unix())) {
            final Path tmpDirPath = jimfs.getPath("tmp");
```

3. You should also include a test case that runs the dummy `CommandLineProgram` with `--tmp-dir` set to a `file://` URI, since that is a different code path.

You could put the test in a new `CommandLineProgramIntegrationTest` class

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -141,8 +138,8 @@ public final Object runTool(){
 
     public Object instanceMainPostParseArgs() {
         // Provide one temp directory if the caller didn't
-        if (this.TMP_DIR == null) this.TMP_DIR = new ArrayList<>();
-        if (this.TMP_DIR.isEmpty()) TMP_DIR.add(IOUtil.getDefaultTmpDir());
+        // TODO - this should use the HTSJDK IOUtil.getDefaultTmpDirPath, which is somehow broken in the current HTSJDK version
+        if (TMP_DIR == null || TMP_DIR.isEmpty()) TMP_DIR = IOUtils.getAbsolutePathWithoutFileProtocol(IOUtils.getPath(System.getProperty("java.io.tmpdir")));
```
droazen: Curly braces around if statement body

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -60,8 +57,8 @@
     // abstract, this is fine (as long as no logging has to happen statically in this class).
     protected final Logger logger = LogManager.getLogger(this.getClass());
 
-    @Argument(fullName = StandardArgumentDefinitions.TMP_DIR_NAME, common=true, optional=true)
-    public List<File> TMP_DIR = new ArrayList<>();
+    @Argument(fullName = StandardArgumentDefinitions.TMP_DIR_NAME, common=true, optional=true, doc = "Temp directory to use.")
+    public String TMP_DIR;
```
droazen: Can you rename this variable to `tmpDir` (but keep the argument name itself as-is)?

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
```diff
@@ -151,13 +148,20 @@ public Object instanceMainPostParseArgs() {
...
+        // set the temp directory as a java property, checking for existence and read/write access
+        final Path p = IOUtils.getPath(TMP_DIR);
+        try {
+            p.getFileSystem().provider().checkAccess(p, AccessMode.READ, AccessMode.WRITE);
+            System.setProperty("java.io.tmpdir", IOUtils.getAbsolutePathWithoutFileProtocol(p));
+        } catch (final AccessDeniedException | NoSuchFileException e) {
+            // TODO: it may be that the program does not need a tmp dir
+            // TODO: if it fails, the problem can be discovered downstream
+            // TODO: should log a warning instead?
+            throw new UserException.BadInput(String.format("--%s (%s) should exists and have read/write access",
```
droazen: Can you use `UserException.BadTmpDir` instead (here and below)?

droazen: Also, "should exists" should be "should exist"

lbergelson: @magicDGS You probably want to rebase before making this change since I just made some changes to BadTmpDir

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: Is the GKL library _**always**_  extracted, even if not used at all? I don't think that it is (or should be) like that...

src/test/java/org/broadinstitute/hellbender/utils/io/IOUtilsUnitTest.java
magicDGS: I guess that you mean adding the `gs://` example, that is the prefix in GCS. Otherwise it fails because `IOUtils.getPath` does not recognize it as a non-file URI.

Done and test passing.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: Leftover from previous implementation. Changed as it is clearer.

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: Done

src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
magicDGS: Using `BadTmpDir` as currently implemented is not really pointing out where the error is: if the property was not set, and fails on checking the access it will point to a different tmp directory. I added a new constructor for the exception with a `Path` to provide the one in use.

Done.

src/test/java/org/broadinstitute/hellbender/utils/io/IOUtilsUnitTest.java
magicDGS: Which `createTempFile` should be use in the test? There are several implementations (htsjdk, gatk, `Files`, `File`), and all of them have their own peculiarities. The ones that work with a `File` object won't work with this change. Thus, I am not sure how to proceed here - for the tools using the tmpDir this should work because the codebase uses the htsjdk `Path` implementation, but otherwise I am not sure how to get this done well...

On the other hand,  there is a test class already for CLP (`CommandLineProgramUnitTest`) - should I still create the `*IntegrationTest` class or just use that one?

magicDGS:
@droazen - Still some questions about integration tests (on the comment with your suggestion, but here too):

* I am not sure if the test that you are proposing will work with all the implementations of `createTempFile`: depends on how it is handle, as a `File` or as a `Path`
* I think that this depends a lot on the parts of the codebase that we are looking at, so maybe before accepting this a pass should be done for the usages and how the `java.io.tmpdir` is used.

Waiting for your feedback before doing something that does not make sense...

droazen:
@magicDGS Could you add overloads of `IOUtils.createTempFile()` and `BaseTest.createTempFile()` that take a Path, and use them just in the tests you write for this PR? Once tests are in place, we can gradually migrate usages to the Path overloads in future PRs.

magicDGS:
Sorry for the delay here, @droazen - I had some technical problems with my hardware and did not have time to work on this. Back to you with implemented tests!

droazen:
@magicDGS Could you rebase this branch one last time? Once it's been rebased, I'll give it a final look and (hopefully) hit merge. Thanks!

magicDGS:
Rebased @droazen - back to you!

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
+/*
+ * The MIT License (MIT)
```
droazen: Remove the license from this file. GATK4 has a repository-wide BSD 3-clause license.

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+import java.io.IOException;
+import java.nio.file.FileSystem;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.UUID;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class CommandLineProgramIntegrationTest extends GATKBaseTest {
```
droazen: Integration test classes should extend `CommandLineProgramTest` rather than `GATKBaseTest` directly.

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private static FileSystem jimfs;
```
droazen: Create the `Jimfs` filesystem within your test case rather than at the class level. Static variables in test code tend to cause problems down the line.

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+        public Path doWork() {
+            return createTempPath(UUID.randomUUID().toString(), ".txt");
+        }
+    }
+
+    @DataProvider
+    public Object[][] tmpDirs() throws Exception {
+        return new Object[][]{
+                {createTempDir("local").toPath()},
+                {Files.createDirectory(jimfs.getPath("tmp"))}
```
droazen: There should be 3 test cases here rather than 2 in order to cover all the branches in your code: a local directory without a `file://` scheme specified, a local directory with `file://` specified, and a Jimfs URI.

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+        };
+    }
+
+    @Test(dataProvider = "tmpDirs", singleThreaded = true)
+    public void testCreateTempPath(final Path tmpDir) {
```
droazen: This test should be named something like `testTmpDirArgument`

src/test/java/org/broadinstitute/hellbender/utils/io/IOUtilsUnitTest.java
```diff
@@ -166,6 +166,21 @@ private void innerTestGetPath(String s) throws IOException {
         Assert.assertTrue(size>0);
     }
 
+    @DataProvider
+    public Object[][] absoluteNames() {
+        return new Object[][] {
+                {"/local/example.txt", "/local/example.txt"},
+                {"/local/file://example.txt", "/local/file:/example.txt"},
```
droazen: This is a strange one... can you explain why `"/local/file://example.txt"` becomes `"/local/file:/example.txt"`?

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+        // store the previous tmp.dir to check that it is working
+        final String previousTmpDir = System.getProperty("java.io.tmpdir");
+
+        final Path p = (Path) new TestCreateTempPathClp().instanceMain(new String[]{
+                "--" + StandardArgumentDefinitions.TMP_DIR_NAME, tmpDir.toUri().toString()});
+        Assert.assertTrue(Files.exists(p));
```
droazen: Also assert that the `java.io.tmpdir` property has been changed.

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -437,6 +439,40 @@ public static File createTempFileInDirectory(final String name, String extension
...
+     * Creates a temp path that will be deleted on exit.
+     *
+     * This will also mark the corresponding Tribble/Tabix/BAM indices matching the temp file for deletion.
+     *
+     * @param name Prefix of the file.
+     * @param extension Extension to concat to the end of the file.
+     *
+     * @return A file in the temporary directory starting with name, ending with extension, which will be deleted after the program exits.
+     */
+    public static Path createTempPath(String name, String extension) {
```
droazen: Would be good to have a standalone unit test for this method in `IOUtilsUnitTest` (since the test in `CommandLineProgramIntegrationTest` is testing the `--tmp-dir` argument rather than this method directly.

src/test/java/org/broadinstitute/hellbender/utils/io/IOUtilsUnitTest.java
```diff
@@ -166,6 +166,21 @@ private void innerTestGetPath(String s) throws IOException {
         Assert.assertTrue(size>0);
     }
 
+    @DataProvider
+    public Object[][] absoluteNames() {
+        return new Object[][] {
```
droazen: Can you also include a test case with a non-absolute path?

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
magicDGS: Removed (sorry, my IDE attach sometimes old LICENSE config...)

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
magicDGS: Done

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
magicDGS: Done

src/test/java/org/broadinstitute/hellbender/utils/io/IOUtilsUnitTest.java
magicDGS: Due to normalization of the path (removal of redundant `/`) - I added a comment here to clarify for the future.

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
magicDGS: Done.

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+        // get back to the previous tmp dir
+        System.setProperty("java.io.tmpdir", previousTmpDir);
```
magicDGS: I also put this in a `finally` clause to be sure that after the test it always resets the previous tmp directory

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
magicDGS: Actually, the one with the `file://` prefix is already tested - in the test, the argument is converted with `Path.toUri().toString()`, which adds the `file://`. I refactored the method to has the `String` argument directly, and added a test case for a local file without prefix.

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
magicDGS: Done.

src/test/java/org/broadinstitute/hellbender/utils/io/IOUtilsUnitTest.java
magicDGS: Done

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
magicDGS: I found a solution by creating a new method in `GATKBaseTest` to close `Jimfs` for a `Path`. I refactored to remove this static field.

magicDGS: For whatever reason, that method fails - it looks like GATK's `IOUtil.getPath` does not recognize `Jimfs` if the instance is created in the test method. Again, no action here.

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Test(dataProvider = "tmpDirs", singleThreaded = true)
+    public void testTmpDirArgument(final String tmpDirArg) throws Exception {
+        // store the previous tmp.dir to check that it is working
+        final String previousTmpDir = System.getProperty("java.io.tmpdir");
+
+        try {
+            final Path p = (Path) new TestCreateTempPathClp().instanceMain(new String[] {
+                    "--" + StandardArgumentDefinitions.TMP_DIR_NAME, tmpDirArg});
+            Assert.assertTrue(Files.exists(p));
```
droazen: Should also assert here that the parent directory of p is equal to the `tmpDirArg`

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+            Assert.assertNotEquals(System.getProperty("java.io.tmpdir"), previousTmpDir);
```
droazen: Ideally you would check here that `java.io.tmpdir` has been set to `tmpDirArg`, not just that it's different from `previousTmpDir`

src/test/java/org/broadinstitute/hellbender/GATKBaseTest.java
```diff
@@ -119,6 +123,16 @@ public void initGenomeLocParser() throws FileNotFoundException {
             locs.add(hg19GenomeLocParser.parseGenomeLoc(interval));
         return Collections.unmodifiableList(locs);
     }
-    
+
+    /**
+     * Closes the {@link java.nio.file.FileSystem} for a path if it is an instance from {@link Jimfs}
+     */
+    public static void closeFileSystemIfJimnfs(final Path path) throws IOException {
```
droazen: I'm not sure this method should be up in `GATKBaseTest`, since you reported that you weren't able to use it in `CommandLineProgramIntegrationTest`. You should use the same method to create the Jimfs instances across your different test classes. At this point I'd be satisfied if you just deleted this method, and used the `@BeforeClass`/`@AfterClass` approach to set up the Jimfs in `IOUtilsUnitTest`.

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.UUID;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class CommandLineProgramIntegrationTest extends CommandLineProgramTest {
+
+    private static FileSystem jimfs;
```
droazen: Can you make this non-static?

src/test/java/org/broadinstitute/hellbender/GATKBaseTest.java
magicDGS: Deleted method.

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
magicDGS: Done

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
magicDGS: Done

src/test/java/org/broadinstitute/hellbender/cmdline/CommandLineProgramIntegrationTest.java
magicDGS: Done

magicDGS:
@droazen - I also rebased locally to check the conflicting files: the differences are a new constant (`DEFAULT_TOOLKIT_SHORT_NAME`) and documentation for `TMP_DIR` (which is also added in this PR, but as it changed the `List` to `String`, it is a bit different). Do you want that I push the rebased branch, or do you prefer to wait until the end of the review?

magicDGS:
@droazen - friendly ping here!

droazen:
@magicDGS It looks like all of my comments here have been addressed, but the branch is in conflict. If you do a last rebase I can (finally) hit merge!

magicDGS:
Rebased - waiting for test passing and hopefully you can hit merge @droazen!

magicDGS:
Thanks for the review @droazen!

PR #4474, opened 2018-03-01
Add DocOnlyArgumentCollection for better downstream toolkit extensibility

magicDGS:
Can you review this one @droazen - this is the last part of the temporary solution for #3998 (mentioned on ##4468). I think that in this case, because the interface is just a marker, it is easier to just modify `GATKDocOnlyArgumentCollection` for doc-only top level arguments and does not modify the API. Thanks!

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -14,6 +14,8 @@ import htsjdk.samtools.util.Log;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.barclay.argparser.*;
+import org.broadinstitute.hellbender.cmdline.argumentcollections.DocOnlyArgumentCollection;
+import org.broadinstitute.hellbender.cmdline.argumentcollections.GATKDocOnlyArgumentCollection;
 import org.broadinstitute.hellbender.utils.LoggingUtils;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.config.ConfigFactory;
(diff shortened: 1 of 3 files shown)


magicDGS:
@droazen - can you have a look to thisone to try to make it to the next release? Thank you!

cmnbroad:
This seems like a lot of machinery (introducing two new types and a new method) just to hide the config file argument. Can't we just mark it `@Hidden` ? The only reason it even exists is because we want it to appear in the command lines we display and embed in outputs. If its `@Hidden` it will still be reflected there when it's used, but it wouldn't be displayed in tool help/usage. Its already always displayed as an arg for the gatk wrapper.

magicDGS:
@cmnbroad - it is not enough for me to make it `@Hidden`.  In my downstream toolkits , I don't want it to appear wherever command line is printed (headers, logs, etc), because I am not supporting custom configuration files. In the case that `Main` is not accounting for the configuration file and an user provide their own, then the command line might indicate wrongly that the configuration was set, but it wasn't.

I really need a way to remove completely the argument. If someone print the help with hidden arguments, then they will misunderstand that the configuration can be overwrite, even more if they know the configuration system of GATK. The only solution for my use case is to being able to remove completely the argument, not just reducing visibility.

cmnbroad:
Args like the config file that are truly optional (have no default value at all) do not show up in the command line or headers unless they're populated with some value. It should be pretty easy for ReadTools (which I think already has a common base class for its tools), to ensure a config file is never accepted by just precluding it via custom command line validation, or arg preprocessing.

BTW, all tools built with GATK already have numerous common args that may or may not apply in a given tool context. For example, all of the ReadWalkers have a `--lenientVCFProcessing` arg. So I'm not even sure we need to make this hidden. My 2 cents. Others may feel differently.

magicDGS:
That's why I am not using in ReadTools and other developmental toolkit the base class from GATK, due to the polluted command line with unused arguments. I think that for give flexibility, some of that arguments should be configurable by extending classes. For example, some tools that does not require reads at all should be able to turn off the read arguments. That will be very useful, although I am not sure how to do it in a proper way without adding more and more interfaces for argument collections.

In context case of this PR, I think that adding it does not have any real effect on the GATK codebase, and a lot is gained by downstream projects. For example, if the wrapper script adds another argument that should be parsed in `Main` and documented, the GATK team just add it to its class. If a toolkit has a similar wrapper script, it can also add its own only-doc argument by simply overriding the method...

magicDGS:
Any progress on this, @cmnbroad?

magicDGS:
Perfect @cmnbroad - I'll have a look!

PR #4493, opened 2018-03-05
Make the SV cluster creation scripts work with preemptible workers

cwhelan:
The current initialization action for dataproc workers puts the reference image in different places depending on whether or not an SSD is mounted. Preemptible dataproc workers don't have SSDs, so a mixed cluster will have references mounted on different paths depending on the worker. This change symlinks the SSD mount point onto the HDD so that paths can be consistent. 

Also increases several cluster configuration parameters relating to retries, which I saw recommended if using preemptible workers.


SHuang-Broad:
👍 Looks good to me.

PR #4494, opened 2018-03-05
exposing pair-hmm-implementation

lbergelson:
previously it was a hidden argument, now it's a visible advanced argument


droazen:
:+1: seems reasonable to me -- merge when tests pass @lbergelson

PR #4495, opened 2018-03-05
VariantEval Core Changes

bbimber:
This is part of the effort to port GATK3 VariantEval to GATK4.  Because VariantEval is a pretty big tool, I thought it would help to separate changes in the VariantEval package from core GATK changes.  If you would prefer I could split this PR into several, based on what they target.  However, here is a summary:

1) I added a method to FeatureInput so downstream code and determine if the name is actually user-supplied or whether this is the default one assigned by GATK

2) I split MultiVariantWalker into a base class that does not have arguments specified.  This is comparable to what already exists for VariantWalkerBase - this is a pretty minor change.

3) I added a method to VariantWalkerBase to return the set of intervals potentially being iterated.  The intent is to return either the full genome, or the user-supplied intervals.  

4) I restored some code to MendelianViolation that is used by components of VariantEval.  This is very close to identical from GATK3

5) Likewise to SampleDBBuilder (code had to be tweaked for GATK4)

6) IntervalUtils/MathUtils/Utils: I restored from GATK3 methods not ported to GATK4.  These should be exact copies of GATK3 unless a change was needed.

Once these are closed out, I will make a separate PR with VariantEval itself.


droazen:
Thanks for splitting this off into a separate PR @bbimber -- we'll try to have a look soon!

bbimber:
@droazen and @cmnbroad: i completely understand that this is outside the main GATK dev cycle and priorities; however, do you have any guess as to when you might be able to review?  I dont know how active development is, but I'd especially like to get that change in VariantWalker and MultiVariantWalker (which are currently really simple refactors) in before other development on them.

cmnbroad:
@bbimber I'll try to get to it this week.

cmnbroad:
@bbimber Can you point me to the branch that has the actual VariantEval changes so I can understand the motivation for MultiVarianWalkerBase and how it will be used ? Thx.

src/main/java/org/broadinstitute/hellbender/utils/text/ListFileUtils.java
```diff
@@ -0,0 +1,169 @@
...
+     * @param exactMatch If true match filters exactly, otherwise use as both exact and regular expressions
+     * @return entries from values including only values matching filters
+     */
+    public static <T> Set<T> includeMatching(Collection<T> values, StringConverter<T> converter, Collection<String> filters, boolean exactMatch) {
+        if (values == null)
+            throw new NullPointerException("values cannot be null");
+        if (converter == null)
+            throw new NullPointerException("converter cannot be null");
+        if (filters == null)
+            throw new NullPointerException("filters cannot be null");
```
cmnbroad: A couple of general style comments that apply to all of this code:
- You can use Utils.nonNull to validate argument values.
- We generally use braces for all blocks, even one liners.
- We mark all variables and arguments final unless the can't be.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerBase.java
```diff
@@ -0,0 +1,99 @@
...
+import htsjdk.samtools.SAMSequenceDictionary;
+import htsjdk.variant.variantcontext.VariantContext;
+import htsjdk.variant.vcf.VCFHeader;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.SortedSet;
+import java.util.Spliterator;
+
+public abstract class MultiVariantWalkerBase extends VariantWalkerBase {
```
cmnbroad: If the only reason for factoring this out as a base class is to allow subclasses to specify a particular command line argument name for inputs, I think it would make more sense to use an ArgumentCollection and overridable method. Take a look at how GATKTool [uses that pattern](https://github.com/broadinstitute/gatk/blob/9be98b976815e6ca68d2e786139a48f0f21d8f77/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java#L51-L58) to allow tool subclasses to control the optional/required attribute on an argument. Just add an overridable method with a default implementation that returns an instance of a class that contains the `@Argument` definition, and has a method to return the driving variants. It will require addition of a new interface or base class for the arg collection, but I think thats a cleaner solution than introducing a new level into the class hierarchy. Of course that won't work if there are other reasons for factoring out the base class.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -129,6 +130,13 @@ protected VariantFilter makeVariantFilter() {
         return VariantFilterLibrary.ALLOW_ALL_VARIANTS;
     }
 
+    /**
+     * Returns the list of intervals to iterate, either limited to the user-supplied intervals or the entire reference genome if none were specified
+     */
+    public List<SimpleInterval> getIntervals() {
+        return hasIntervals() ? intervalArgumentCollection.getIntervals(getReferenceDictionary()) : IntervalUtils.getAllIntervalsForReference(getReferenceDictionary());
```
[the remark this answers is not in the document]
bbimber: I think you're probably right.  I will simplify that to use intervalsForTraversal and add testing

src/main/java/org/broadinstitute/hellbender/utils/MathUtils.java
```diff
@@ -776,6 +776,13 @@ public static int fastRound(final double d) {
         return (d > 0.0) ? (int) (d + 0.5d) : (int) (d - 0.5d);
     }
 
+    public static double round(final double num, final int digits) {
```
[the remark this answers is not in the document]
bbimber: yes, i just carried that over from GATK3.  removed.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1233,4 +1233,37 @@ private static void removeTrailingEmptyStringsFromEnd(final List<String> result)
...
     }
+
+    /**
+     * Convenience function that formats the novelty rate as a %.2f string
+     *
+     * @param known number of variants from all that are known
+     * @param all number of all variants
+     * @return a String novelty rate, or NA if all == 0
+     */
+    public static String formattedNoveltyRate(final int known, final int all) {
```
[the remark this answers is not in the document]
bbimber: I will put this into VariantEvaluator, since it is reasonably specific to those classes.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1233,4 +1233,37 @@ private static void removeTrailingEmptyStringsFromEnd(final List<String> result)
...
+    /**
+     * Convenience function that formats a ratio as a %.2f string
+     *
+     * @param num  number of observations in the numerator
+     * @param denom number of observations in the denumerator
+     * @return a String formatted ratio, or NA if all == 0
+     */
+    public static String formattedRatio(final long num, final long denom) {
+        return denom == 0 ? "NA" : String.format("%.2f", num / (1.0 * denom));
+    }
```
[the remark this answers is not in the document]
bbimber: yeah, they're used by VariantEval code.  i'll add unit testing

src/main/java/org/broadinstitute/hellbender/utils/text/ListFileUtils.java
```diff
@@ -0,0 +1,169 @@
...
+        for (String value : values) {
+            File file = new File(value);
+            if (value.toLowerCase().endsWith(".list") && file.exists()) {
+                unpackedValues.addAll(new XReadLines(IOUtil.openFileForBufferedReading(file), true, LIST_FILE_COMMENT_START).readLines());
+            } else {
+                unpackedValues.add(value);
+            }
+        }
+        return unpackedValues;
+    }
```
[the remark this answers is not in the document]
bbimber: that's probably correct; however, do you have a GATK4 example of code using this barclay file expansion feature?

src/main/java/org/broadinstitute/hellbender/utils/text/ListFileUtils.java
bbimber: These were verbatim from GATK3, but will try to look through this more closely.  i think when we get into VariantEval's code that's going to be a lot of this kind of stuff.

src/main/java/org/broadinstitute/hellbender/utils/text/ListFileUtils.java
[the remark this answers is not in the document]
cmnbroad: Barclay does this automatically for any `@Argument` that is a collection (and that doesn't have suppressFileExpansion property set), so there isn't any code required if thats what you mean. SelectVariants, for example, relies on it - you can see an example of a test that uses it [here](https://github.com/broadinstitute/gatk/blob/f26cb5dbd7f8bb4ee549c7c1c1044e8ec7a28645/src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/SelectVariantsIntegrationTest.java#L72) if that helps.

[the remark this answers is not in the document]
cmnbroad: BTW, I noticed that SelectVariants has the filtering code from below embedded in it, since it does the same kind of thing. So that code should be updated to use whatever filtering methods result from these changes.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerBase.java
cmnbroad: I would say for now to confine it to the MultiVariantWalker hierarchy unless that can't be done cleanly.

src/main/java/org/broadinstitute/hellbender/utils/text/ListFileUtils.java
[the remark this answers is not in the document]
bbimber: That last post is exactly what I was going to write.  VariantEval is doing exactly what SelectVariants is doing, so I think we can entirely omit ListFileUtils.  Therefore we just want to pull that code related to SelectVariants.matchSamplesExpressions() into some kind of utility class.  Do you have an opinion on where in GATK4 this belongs?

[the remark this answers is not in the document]
cmnbroad: Probably Utils, assuming the code isn't sample-specific and is just filtering a list based on a regex.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalkerBase.java
bbimber: ok, done

src/main/java/org/broadinstitute/hellbender/utils/samples/SampleDBBuilder.java
```diff
@@ -53,6 +53,17 @@ public SampleDBBuilder addSamplesFromPedigreeStrings(final List<String> pedigree
...
+        for (final String sampleName : sampleNames) {
+            if (sampleDB.getSample(sampleName) == null) {
+                final Sample newSample = new Sample(sampleName, null, null, null, Sex.UNKNOWN);
+                sampleDB.addSample(newSample);
+                samplesFromDataSources.add(newSample); // keep track of data source samples
+            }
+        }
+        return this;
+    }
+
```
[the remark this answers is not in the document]
bbimber: done

src/main/java/org/broadinstitute/hellbender/utils/samples/MendelianViolation.java
```diff
@@ -167,4 +213,141 @@ private void clearInheritanceMap(){
...
+
+    //Count of violations of the type HOM_VAR/? -> HOM_REF
+    public int getParentVarChildRef(){
+        return getParentsRefVarChildRef() + getParentsVarHetChildRef() +getParentsVarVarChildRef();
+    }
+
+    //Count of violations of the type HOM_REF/? -> HOM_VAR
+    public int getParentRefChildVar(){
+        return getParentsRefVarChildVar() + getParentsRefHetChildVar() +getParentsRefRefChildVar();
+    }
```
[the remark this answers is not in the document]
bbimber: Question on this: I generally tried to carry forward GATK3 VariantEval functionality.  The only user of all these new methods is MendelianViolationEvaluator.  Most of this tracking arguably doesnt have anything to do w/ MVs either (it's just tracking of inheritance by variant type).  What do you think about simply not carrying forward the extended inheritance/MV tracking, which would mean a small functionality loss to VariantEval?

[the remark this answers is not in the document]
cmnbroad: @droazen Do you have an opinion on leaving this aspect of MV eval out of the VariantEval port ?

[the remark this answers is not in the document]
bbimber: @cmnbroad - what do you think about the MendelianViolation testing I just added?  I removed final from that class with the plan of subclassing in MendelianViolationEvaluator, where I can add the highly specialized additional code.

I think that was the only remaining piece on this PR

bbimber:
ok, thanks.  also - the travis build seems to have timed out.  from the log i think that's just a travis CI issue, not due to my commit.  is this travis timeout problem something you see frequently w/ GATK4?

cmnbroad:
@bbimber Yes, the timeouts are not unusual. I restarted the (1 of the 8) jobs in the build matrix that failed on your PR. It will probably be fine, but I'll keep an eye on it.

src/test/java/org/broadinstitute/hellbender/utils/samples/MendelianViolationUnitTest.java
```diff
@@ -146,15 +148,25 @@ public void testIsViolationFromSamples(final Genotype gMom, final Genotype gDad,
...
+        final boolean expectedWithQual = !isLowQual && expected;
+
+        Assert.assertEquals(expectedWithQual, actual);
+        Assert.assertEquals(!isLowQual && (gMom.isHomRef() && gDad.isHomRef() && gChild.isHet()), mv.getParentsRefRefChildHet() > 0);
+        Assert.assertEquals(!isLowQual && !isNoCall, mv.getFamilyCalledCount() > 0);
+
+        Assert.assertEquals(isNoCall, mv.getFamilyNoCallCount() > 0);
+        Assert.assertEquals(!isNoCall && isLowQual, mv.getFamilyLowQualsCount() > 0);
+        Assert.assertEquals(expectedWithQual, mv.getViolationsCount() > 0);
+        Assert.assertEquals((isLowQual || isNoCall || (gMom.isHomRef() && gDad.isHomRef() && gChild.isHomRef())), mv.getVarFamilyCalledCount() == 0);
```
cmnbroad: We generally try not to gang up multiple test/asserts like this in the same method if we can avoid it. Can you separate the tests for getParentsRefRefChildHet , getFamilyNoCallCount , getFamilyLowQualsCount, getVarFamilyCalledCount into separate test methods, and then they can share a minimal utility method to create the vc. Also, the assert methods generally take arguments in the order (actual, expected).

src/main/java/org/broadinstitute/hellbender/engine/FeatureInput.java
```diff
@@ -267,6 +271,14 @@ public String getName() {
         return name;
     }
 
+    /**
+     *
+     * @return true if the value for name matches the default
+     */
+    public boolean isUserSuppliedName() {
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/engine/VariantArgumentCollection.java
```diff
@@ -0,0 +1,29 @@
+package org.broadinstitute.hellbender.engine;
+
+import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class VariantArgumentCollection implements Serializable {
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/engine/VariantArgumentCollection.java
```diff
@@ -0,0 +1,29 @@
...
+    private static final long serialVersionUID = 1L;
+
+    abstract protected List<String> getDrivingVariantFiles();
```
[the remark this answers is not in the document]
bbimber: sure

src/main/java/org/broadinstitute/hellbender/engine/VariantArgumentCollection.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public static class DefaultArgumentCollection extends VariantArgumentCollection {
+        private static final long serialVersionUID = 1L;
+
+        // NOTE: using File rather than FeatureInput<VariantContext> here so that we can keep this driving source
+        //       of variants separate from any other potential sources of Features
```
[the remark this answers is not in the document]
bbimber: OK.  Aside from that, and honestly I dont really want to take this on here, but is it really actually necessary to separate the driving variants anymore?

src/main/java/org/broadinstitute/hellbender/utils/samples/MendelianViolation.java
```diff
@@ -14,12 +14,20 @@
  * - Or using the static methods to directly get information about mendelian violation in a family at a given locus
  *
  */
-public final class MendelianViolation {
+public class MendelianViolation {
+    //List of families with violations
+    private List<String> violationFamilies = new ArrayList<>();
+
+    //Call information
+    private int nocall = 0;
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/utils/samples/MendelianViolation.java
```diff
@@ -167,4 +194,32 @@ private void clearInheritanceMap(){
             }
         }
     }
+
+    //Number of families with genotype information for all members
+    public int getFamilyCalledCount(){
+        return familyCalled;
+    }
+
+    //Number of families with genotype information for all members
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/utils/samples/MendelianViolation.java
```diff
@@ -49,14 +57,24 @@ public int getParentsRefRefChildHet() {
...
+
+    public boolean isViolation(final Sample mother, final Sample father, final Sample child, final VariantContext vc){
+        resetCounts();
         updateViolations(mother.getFamilyID(), mother.getID(), father.getID(), child.getID(),vc);
-        return violations_total>0;
+        return violations_total > 0;
     }
 
-    private void updateViolations(final String familyId, final String motherId, final String fatherId, final String childId, final VariantContext vc){
+    protected void updateViolations(final String familyId, final String motherId, final String fatherId, final String childId, final VariantContext vc){
```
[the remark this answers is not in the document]
bbimber: As I said in the earlier comment, rather than re-implement all of the VariantEval-specific code in MendelianViolation (it tracked various inheritance-related values which as not really MV-related), I am planning to subclass MendelianViolation in MendelianViolationEvaluator (a VariantEval class).  Once VariantEval is merged this will be important.

src/main/java/org/broadinstitute/hellbender/utils/samples/MendelianViolation.java
```diff
@@ -49,14 +57,24 @@ public int getParentsRefRefChildHet() {
         return inheritance.get(GenotypeType.HOM_REF).get(GenotypeType.HOM_REF).get(GenotypeType.HET);
     }
 
-    public boolean isViolation(final Sample mother, final Sample father, final Sample child, final VariantContext vc){
-        violations_total=0;
+    private void resetCounts() {
```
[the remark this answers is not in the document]
bbimber: ok

src/test/java/org/broadinstitute/hellbender/utils/samples/MendelianViolationUnitTest.java
bbimber: done

bbimber:
@cmnbroad - I think that covers everything.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
```diff
@@ -22,11 +20,12 @@
...
 
-    // NOTE: using File rather than FeatureInput<VariantContext> here so that we can keep this driving source
-    //       of variants separate from any other potential sources of Features
-    @Argument(fullName = StandardArgumentDefinitions.VARIANT_LONG_NAME, shortName = StandardArgumentDefinitions.VARIANT_SHORT_NAME,
-                doc = "One or more VCF files containing variants", common = false, optional = false)
-    public List<String> drivingVariantFiles = new ArrayList<>();
+    @ArgumentCollection
+    protected MultiVariantInputArgumentCollection multiVariantInputArgumentCollection = getMultiVariantInputArgumentCollection();
+
+    protected MultiVariantInputArgumentCollection getMultiVariantInputArgumentCollection() {
```
cmnbroad: Minor nit request to move this down in the file past the variable declarations and into the section with the other methods, with a javadoc comment.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1233,4 +1234,70 @@ private static void removeTrailingEmptyStringsFromEnd(final List<String> result)
...
+
+    /**
+     * Given a collection of strings and a collection of regular expressions, generates the set of strings that match
+     * each expression
```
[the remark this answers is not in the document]
cmnbroad: Correct me if you think I'm wrong, but I still think this should say "...match any expression".

src/main/java/org/broadinstitute/hellbender/engine/VariantArgumentCollection.java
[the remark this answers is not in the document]
cmnbroad: Yes, I think we still use different caching/lookahead values for these.

cmnbroad:
@bbimber These changes look pretty good now except for a couple of things from the last round that didn't get done. I added new comments for those. Please make those changes, and also squash this PR down to single a commit, and rebase on current master. Thx.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
bbimber: ok

bbimber:
Sorry, missed those.  I fixed two comments, which I think covers it.

bbimber:
@cmnbroad There's your rebase and squash as well

src/main/java/org/broadinstitute/hellbender/engine/FeatureInput.java
```diff
@@ -267,6 +271,14 @@ public String getName() {
         return name;
     }
 
+    /**
+     *
+     * @return true if the value for name matches the default
```
cmnbroad: The javadoc for this method describes the inverse of what it does.

bbimber: done.

bbimber:
@cmnbroad do you mind just squashing via github when you do the merge?

bbimber:
@droazen and @cmnbroad: just wanted to check back in on this.  thanks.

bbimber:
@cmnbroad and @droazen : I wanted to follow up once more.  I know this is a lower priority for GATK but would love to close this out.  There's been a pretty extensive review here.  I believe all the spots still indicating change requested are false positives for situations like requesting javadoc added to some other position in the code (which was done).

It looks like we now have conflicts due to other changes that happened while this was sitting here.  I just dealt with them easily enough, but this kind of thing will probably keep happening.

Thanks

droazen:
@bbimber I have a review in progress that contains a few comments on some remaining issues in this branch -- I'll post it soon.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1274,4 +1275,70 @@ private static void removeTrailingEmptyStringsFromEnd(final List<String> result)
...
+
+        Collection<Pattern> patterns = null;
+        if (!exactMatch) {
+            patterns = compilePatterns(filterExpressions);
+        }
+        for (final String value : sourceValues) {
+            if (filterExpressions.contains(value)) {
+                filteredValues.add(value);
+            } else if (!exactMatch) {
+                for (final Pattern pattern : patterns) {
```
droazen: As @cmnbroad mentioned, this loop should end after the first match.

src/test/java/org/broadinstitute/hellbender/engine/FeatureInputUnitTest.java
```diff
@@ -257,4 +257,24 @@ public void testToString() {
...
+                {"myname,key1=value1:hdfs://localhost/user/my.vcf", true},
+                {"myname//:hdfs://localhost/user/my.vcf", true}
+        };
+    }
+
+    @Test(dataProvider = "IsUserSuppliedNameData")
+    public void testIsUserSuppliedName(final String inputString, final boolean isUserSupplied) {
+        final FeatureInput<VariantContext> input = new FeatureInput<>(inputString);
+
+        //If the name assigned to the
```
droazen: Complete this sentence

src/test/java/org/broadinstitute/hellbender/engine/FeatureInputUnitTest.java
```diff
@@ -257,4 +257,24 @@ public void testToString() {
         Assert.assertEquals(namelessGenomicsDB.toString(), "gendb://" + new File("file1").getAbsolutePath(), "String representation of nameless FeatureInput with genomicsDB path incorrect");
         Assert.assertEquals(namedGenomicsDB.toString(), "name:gendb://" + new File("file1").getAbsolutePath(), "String representation of named FeatureInput with genomicsDB path incorrect");
     }
+
+    @DataProvider(name = "IsUserSuppliedNameData")
+    public Object[][] isUserSuppliedNameData() {
+        return new Object[][] {
+                // input String, expected Feature path, expected logical name
```
droazen: Comment here is not accurate -- there are only two values per test case, not three.

src/test/java/org/broadinstitute/hellbender/utils/samples/SampleDBUnitTest.java
```diff
@@ -111,6 +111,15 @@ public void loadPEDString() {
         Assert.assertEquals(testPEDSamplesAsSet, db.getSamples());
     }
 
+    @Test()
+    public void loadPEDFromNames() {
+        List<String> names = new ArrayList<>();
+        testSAMSamples.forEach(s -> names.add(s.getID()));
+        builder.addSamplesFromSampleNames(names);
+        SampleDB db = builder.getFinalSampleDB();
+        Assert.assertEquals(testSAMSamples, db.getSamples());
```
droazen: The actual value (`db.getSamples()`) should be the first argument to `assertEquals()`, and the expected value (`testSAMSamples`) should be the second argument.

src/test/java/org/broadinstitute/hellbender/utils/samples/SampleDBUnitTest.java
```diff
@@ -111,6 +111,15 @@ public void loadPEDString() {
         Assert.assertEquals(testPEDSamplesAsSet, db.getSamples());
     }
 
+    @Test()
+    public void loadPEDFromNames() {
```
droazen: Test method should be named after the method it's testing -- eg., `testAddSamplesFromSampleNames()`

src/test/java/org/broadinstitute/hellbender/engine/FeatureInputUnitTest.java
```diff
@@ -257,4 +257,24 @@ public void testToString() {
...
+                {"hdfs://localhost/user/my.vcf", false},
+                {"myname:hdfs://localhost/user/my.vcf", true},
+                {"myname,key1=value1:hdfs://localhost/user/my.vcf", true},
+                {"myname//:hdfs://localhost/user/my.vcf", true}
```
droazen: Include at least one test case with a relative path on the right side (rather than a URI)

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerBaseUnitTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.broadinstitute.hellbender.CommandLineProgramTest;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.io.File;
+
+public class VariantWalkerBaseUnitTest extends CommandLineProgramTest {
+    final String baseVariants = packageRootTestDir + "engine/feature_data_source_test.vcf";
+
+    private static class TestVariantWalker extends VariantWalker {
```
droazen: This needs a `@CommandLineProgramProperties` annotation. Use `TestProgramGroup.class` for the `programGroup` attribute.

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerBaseUnitTest.java
```diff
@@ -0,0 +1,44 @@
+package org.broadinstitute.hellbender.engine;
+
+import htsjdk.variant.variantcontext.VariantContext;
+import org.broadinstitute.hellbender.CommandLineProgramTest;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.io.File;
+
+public class VariantWalkerBaseUnitTest extends CommandLineProgramTest {
```
droazen: When you move the `getIntervals()` method up into `GATKTool` as suggested above, this test case can move into `GATKToolIntegrationTest`

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerBaseUnitTest.java
```diff
@@ -0,0 +1,44 @@
...
+        final TestVariantWalker tool = new TestVariantWalker(baseVariants);
+        tool.instanceMain(new String[]{
+                "-V", baseVariants,
+                "-R", hg19MiniReference,
+                "-L", "1:21-21"
+        });
+
+        Assert.assertEquals(tool.getIntervals().size(), 1);
+
+        final TestVariantWalker tool2 = new TestVariantWalker(baseVariants);
```
droazen: Specify the two test cases using a `DataProvider`, as @cmnbroad suggested.

src/main/java/org/broadinstitute/hellbender/utils/samples/MendelianViolation.java
```diff
@@ -49,14 +57,32 @@ public int getParentsRefRefChildHet() {
...
+
+    /**
+     * Tests whether there is a mendelian violation between the supplied samples.  Note: this will reset any accumulated stats.
+     * @param mother
+     * @param father
+     * @param child
+     * @param vc
+     * @return Whether a violation is present
+     */
+    public boolean isViolation(final Sample mother, final Sample father, final Sample child, final VariantContext vc){
```
droazen: Having a predicate method like `isViolation()` alter object state like this is a pretty dangerous design pattern. Is there a simple refactoring we could do in this `MendelianViolation` class to fix this? What if we introduced an inner `MendelianViolationCounts` class, stored an instance of it for use by methods that need to accumulate counts across calls, and create a fresh, isolated instance of `MendelianViolationCounts` for methods that currently reset the counts on every call. What do you think @cmnbroad ?

src/main/java/org/broadinstitute/hellbender/engine/FeatureInput.java
```diff
@@ -267,6 +271,14 @@ public String getName() {
         return name;
     }
 
+    /**
+     *
+     * @return true if the value for name does not match the default, indicating it was a user supplied name (i.e. foo:file.vcf)
+     */
+    public boolean isUserSuppliedName() {
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantInputArgumentCollection.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Class that defines the variant arguments used for a MultiVariantWalker.  DefaultArgumentCollection uses the standard --variant argument; however,
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantInputArgumentCollection.java
```diff
@@ -0,0 +1,33 @@
...
+ * subclasses of MultiVariantWalker can override MultiVariantWalker.getMultiVariantInputArgumentCollection() and provide their own argument pattern.
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantInputArgumentCollection.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public abstract class MultiVariantInputArgumentCollection implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    abstract protected List<String> getDrivingVariantPaths();
```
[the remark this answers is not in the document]
bbimber: ok

src/test/java/org/broadinstitute/hellbender/engine/FeatureInputUnitTest.java
bbimber: ok

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerBaseUnitTest.java
bbimber: ok

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerBaseUnitTest.java
bbimber: I dont follow here.  @cmnbroad suggested that I use the previous computed intervalsForTraversal (see earlier version); however, who is suggesting moving to GATKTool?

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerBaseUnitTest.java
bbimber: I dont see where @cmnbroad suggested this, but OK

src/main/java/org/broadinstitute/hellbender/utils/samples/MendelianViolation.java
bbimber: I dont disagree at all here, but as @cmnbroad noted, this entire pattern existing in GATK4 before this PR.  Something like a completely separate MendelianViolationCounter class and MendelianViolationUtils (the static methods) isnt a huge refactor, but I'd really prefer to not creep the scope of this PR here.

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerBaseUnitTest.java
droazen: I suggested it above in https://github.com/broadinstitute/gatk/pull/4495#discussion_r189694685

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -129,6 +130,13 @@ protected VariantFilter makeVariantFilter() {
         return VariantFilterLibrary.ALLOW_ALL_VARIANTS;
     }
 
+    /**
+     * Returns the list of intervals to iterate, either limited to the user-supplied intervals or the entire reference genome if none were specified
+     */
+    public List<SimpleInterval> getIntervals() {
```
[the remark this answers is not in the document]
bbimber: sorry, somehow missed this.  I moved to GATKTool; however, I'm not sure if you're asking for more.  It already returns intervalsForTraversal directly if non-null; however, as the javadoc says the whole point of the method is to default back to the reference dictionary intervals if no user-supplied intervals are present.

because we're moving this to GATKTool and there is now the possibility of a walker where getReference() is false, I added a check for that situation.

droazen:
@bbimber There are still a lot of unaddressed comments from my review: https://github.com/broadinstitute/gatk/pull/4495#discussion_r189690850. Eg.,

https://github.com/broadinstitute/gatk/pull/4495#discussion_r189690850
https://github.com/broadinstitute/gatk/pull/4495#discussion_r189693827
https://github.com/broadinstitute/gatk/pull/4495#discussion_r189694685
https://github.com/broadinstitute/gatk/pull/4495#discussion_r189697161
https://github.com/broadinstitute/gatk/pull/4495#discussion_r193207051
https://github.com/broadinstitute/gatk/pull/4495#discussion_r195236794
https://github.com/broadinstitute/gatk/pull/4495#discussion_r195242052
https://github.com/broadinstitute/gatk/pull/4495#discussion_r195242436
https://github.com/broadinstitute/gatk/pull/4495#discussion_r195243668
https://github.com/broadinstitute/gatk/pull/4495#discussion_r195243837

You also missed a bunch of @cmnbroad 's comments from his earlier review. I'm guessing what's going on here is that you might not be expanding the `N Hidden Conversations...Load More` box that github inserts into the middle of the reviews to hide comments when there are more than a certain number of them?

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerBaseUnitTest.java
bbimber: got it.  I moved the test and extended it to cover some situations that are not possible for VariantWalker, but could occur for GATKTool

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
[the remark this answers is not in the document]
bbimber: Also, I renamed the method to getEffectiveIntervals() to more clearly differentiate its purpose and differentiate from other basic getters.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
```diff
@@ -46,14 +41,23 @@
...
         super.onStartup();
+
         if ( hasIntervals() ) {
             drivingVariants.setIntervalsForTraversal(intervalsForTraversal);
         }
     }
 
+    /**
+     * Return an argument collection that provides the driving variants.  This allows subclasses to override and use a different
+     * argument pattern besides the default -V
```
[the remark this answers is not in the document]
bbimber: I dont have strong feelings either; however, adding the capability for future walkers to override and provide their own names seems like a reasonable capability.

We can address in more detail when we actually get to VariantEval itself; however, as a general rule I did try to stick to what was established under GATK3, for better or worse.

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1274,4 +1275,70 @@ private static void removeTrailingEmptyStringsFromEnd(final List<String> result)
             result.remove(result.size() - 1);
         }
     }
+
+    /**
+     * Convenience function that formats the novelty rate as a %.2f string
```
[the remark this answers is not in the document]
bbimber: ok

src/test/java/org/broadinstitute/hellbender/engine/FeatureInputUnitTest.java
bbimber: ok

src/test/java/org/broadinstitute/hellbender/engine/FeatureInputUnitTest.java
bbimber: removed

src/test/java/org/broadinstitute/hellbender/utils/samples/SampleDBUnitTest.java
bbimber: right

bbimber: also corrected same error in the pre-existing loadNonExistentFile()

src/test/java/org/broadinstitute/hellbender/utils/samples/SampleDBUnitTest.java
bbimber: changed; however, that same comment could apply to virtually every other test in SampleDBUnitTest (all preexisting)

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantWalker.java
```diff
@@ -22,11 +20,12 @@
...
 
-    // NOTE: using File rather than FeatureInput<VariantContext> here so that we can keep this driving source
-    //       of variants separate from any other potential sources of Features
-    @Argument(fullName = StandardArgumentDefinitions.VARIANT_LONG_NAME, shortName = StandardArgumentDefinitions.VARIANT_SHORT_NAME,
-                doc = "One or more VCF files containing variants", common = false, optional = false)
-    public List<String> drivingVariantFiles = new ArrayList<>();
+    @ArgumentCollection
+    protected VariantArgumentCollection variantArgumentCollection = getVariantArgumentCollection();
+
+    protected VariantArgumentCollection getVariantArgumentCollection() {
```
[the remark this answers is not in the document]
bbimber: this was done already

src/main/java/org/broadinstitute/hellbender/engine/VariantArgumentCollection.java
```diff
@@ -0,0 +1,29 @@
...
+        @Argument(fullName = StandardArgumentDefinitions.VARIANT_LONG_NAME, shortName = StandardArgumentDefinitions.VARIANT_SHORT_NAME,
+                doc = "One or more VCF files containing variants", common = false, optional = false)
+        public List<String> drivingVariantFiles = new ArrayList<>();
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1233,4 +1234,70 @@ private static void removeTrailingEmptyStringsFromEnd(final List<String> result)
...
+     * Given a collection of strings and a collection of regular expressions, generates the set of strings that match
+     * each expression
+     * @param sourceValues collection of strings from which to to select
+     * @param filterExpressions list of expressions to use for matching
+     * @param exactMatch If true match filters exactly, otherwise use as both exact and regular expressions
+     * @return A new set strings from sourceValues that satisfy at least one of the expressions in sampleExpressions
+     */
+    public static Set<String> filterCollectionByExpressions(final Collection<String> sourceValues, final Collection<String> filterExpressions, final boolean exactMatch) {
+        if (filterExpressions == null) {
+            return Collections.emptySet();
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
```diff
@@ -1233,4 +1234,70 @@ private static void removeTrailingEmptyStringsFromEnd(final List<String> result)
...
+        if (!exactMatch) {
+            patterns = compilePatterns(filterExpressions);
+        }
+        for (final String value : sourceValues) {
+            if (filterExpressions.contains(value)) {
+                filteredValues.add(value);
+            } else if (!exactMatch) {
+                for (final Pattern pattern : patterns) {
+                    if (pattern.matcher(value).find()) {
+                        filteredValues.add(value);
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/utils/samples/SampleDBBuilder.java
```diff
@@ -53,6 +53,17 @@ public SampleDBBuilder addSamplesFromPedigreeStrings(final List<String> pedigree
         }
     }
 
+    public SampleDBBuilder addSamplesFromSampleNames(final Collection<String> sampleNames) {
```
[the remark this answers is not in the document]
bbimber: ok

src/test/java/org/broadinstitute/hellbender/engine/VariantWalkerBaseUnitTest.java
```diff
@@ -0,0 +1,44 @@
...
+        tool2.instanceMain(new String[]{
+                "-V", baseVariants,
+                "-R", hg19MiniReference
+        });
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/utils/samples/MendelianViolation.java
```diff
@@ -167,4 +194,32 @@ private void clearInheritanceMap(){
             }
         }
     }
+
+    //Number of families with genotype information for all members
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/utils/Utils.java
bbimber: covered above

bbimber:
Yes, I think you're right about hidden conversations.  I'm fairly sure I just covered the missing ones.

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantInputArgumentCollection.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public static class DefaultArgumentCollection extends MultiVariantInputArgumentCollection {
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/org/broadinstitute/hellbender/engine/MultiVariantInputArgumentCollection.java
```diff
@@ -0,0 +1,33 @@
...
+        private static final long serialVersionUID = 1L;
+
+        // NOTE: using List<String> rather than List<FeatureInput> here so that we can initialize the driving source of variants separately
+        // from any other potential sources of Features.
+        @Argument(fullName = StandardArgumentDefinitions.VARIANT_LONG_NAME, shortName = StandardArgumentDefinitions.VARIANT_SHORT_NAME,
+                doc = "One or more VCF files containing variants", common = false, optional = false)
+        public List<String> drivingVariantFiles = new ArrayList<>();
```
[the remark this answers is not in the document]
bbimber: covered above

bbimber:
@droazen : i just resolved the merge conflicts.  sorry for missing the hidden comments earlier, but i think this now covers everything

bbimber:
Hello - i wanted to check back on this PR.

bbimber:
@droazen @cmnbroad - just wanted to check in once more.  i see each has a few items listed changes requested; however, i am fairly sure that is just b/c either a requested change was optional (i.e. --eval arg) or i replied with the reasoning on why it wasnt done.

droazen:
@bbimber I am out of the country for a few weeks. I'm turning the last phase of the review back over to @cmnbroad. @cmnbroad, can you check whether the last round of comments have been fully addressed? If they have, and you are satisfied with the state of the branch, go ahead and merge.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -926,6 +927,14 @@ public String getToolName() {
         return String.format("%s %s", getToolkitShortName(), getClass().getSimpleName());
     }
 
+    /**
+     * Returns the list of intervals to iterate, either limited to the user-supplied intervals or the entire reference genome if none were specified.
+     * If no reference was supplied, null is returned
+     */
+    public List<SimpleInterval> getEffectiveIntervals() {
```
droazen: The method name you chose here is a bit confusing -- @cmnbroad can you suggest a better name in your final review? Perhaps this should be `getTraversalIntervals()`, and we could rename the field `intervalsForTraversal` to the somewhat more accurate `userIntervals`?

droazen:
(As part of your final review @cmnbroad, you should open a ticket to eventually refactor the `MendelianViolation` class)

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
bbimber: That would seem reasonable.  hasIntervals() should probably also get renamed hasUserSuppliedIntervals() as well?

cmnbroad:
@bbimber I went to start looking at this again and realized that you've merged master into your branch, so your commits are now distributed throughout the commit history (which makes it super hard to review, especially when reviewing in IntelliJ). Before I re-review, can you *rebase* (but don't squash) this branch onto recent master, resolving any conflicts. For the future, we generally try not to do any of merge, rebase, or squash mid-review, but at this point its the easiest way to get this branch back on track. Thx.

bbimber:
@cmnbroad : point taken on the merge; however, it was you who asked me to squash changes around May 11, when you first signed off on this.  There's been only a handful of additional changes following that merge: a few really minor things related to @droazen and two instances when I needed to make minor changes to resolve conflicts that appeared w/ master.

bbimber:
@cmnbroad - I rebased.  as you'll see there are 2 spurious commits due to my earlier merge dealing w/ trunk conflicts; however, hopefully this is good enough.  it will get squashed on the final merge.  There's only 16 files touched and it seemed clean when i was trying to review it.

Also: I have a local change staged that renamed hasIntervals() to hasUserSuppliedIntervals() and intervalsForTraversal to userIntervals (see @droazen comment above).  I'm happy to add this now, though would be equally happy to close this and immediately open a separate PR to keep it cleaner and more efficient.

bbimber:
interesting - ok, that's done.  and yes, the spurious commits are basically errors learning git rebase.  hopefully this covers everything.  the change in hasUserSuppliedIntervals() touched a lot of files, but it's a pretty trivial change in them.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -157,7 +158,7 @@
      * Walker base classes (ReadWalker, etc.) are responsible for hooking these intervals up to
      * their particular driving data source.
      */
-    List<SimpleInterval> intervalsForTraversal;
+    List<SimpleInterval> userIntervals;
```
cmnbroad: To maintain as much symmetry as possible between GATKTool and GATKSparkTool, please rename the corresponding variable in GATKSparkTool to match this one. Its private (!) so the fallout should be small.

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
```diff
@@ -945,4 +939,59 @@ private void writeHeaderAndBadVariant(final VariantContextWriter writer) {
...
+        return new Object[][]{
+                {"1:21-21", 1, hg19MiniReference, false},
+                {null, 4, hg19MiniReference, false},
+                {"1:21-21", 1, null, true},  //this will fail with an error
+                {null, null, null, false}
+        };
+    }
+
+    @Test(dataProvider = "TestGetIntervalsProvider")
+    public void testGetIntervals(@Nullable String intervals, Integer expected, String ref, boolean expectUserException) {
```
cmnbroad: Rename to testGetTraversalIntervals

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
```diff
@@ -1,12 +1,8 @@
 package org.broadinstitute.hellbender.engine;
 
+import com.sun.istack.Nullable;
```
cmnbroad: For consistency with the rest of the source, please use javax.annotation.Nullable

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
```diff
@@ -945,4 +939,59 @@ private void writeHeaderAndBadVariant(final VariantContextWriter writer) {
...
+
+        }
+    }
+
+    @DataProvider(name = "TestGetIntervalsProvider")
+    public Object[][] getTestGetIntervalsProvider() {
+        return new Object[][]{
+                {"1:21-21", 1, hg19MiniReference, false},
+                {null, 4, hg19MiniReference, false},
+                {"1:21-21", 1, null, true},  //this will fail with an error
```
cmnbroad: This is an excellent test case, but its out of place in this provider since the test fails well before it gets to `getTraversalIntervals`. It should come out of this provider but it would be worth keeping as a separate `testSequenceDictionaryRequiredForIntervalQuery` test method.

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
```diff
@@ -945,4 +939,59 @@ private void writeHeaderAndBadVariant(final VariantContextWriter writer) {
...
+            args.add(intervals);
+        }
+
+        try {
+            final TestIntervalWalker tool = new TestIntervalWalker();
+            tool.instanceMain(args.toArray(new String[args.size()]));
+
+            Assert.assertEquals(tool.getTraversalIntervals() == null ? null : tool.getTraversalIntervals().size(), expected);
+        }
+        catch (UserException e) {
```
cmnbroad: We generally try to not intermix cases that throw with cases that don't in the same method (and we generally use either the `expectedExceptions` property of the `@Test` annotation or occasionally`Assert.assertThrows`, to declare methods that throw). But in this case if you remove the test case I mentioned above, the whole try/catch block and expectUserException arg can be removed.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -8,8 +8,10 @@
 import org.broadinstitute.hellbender.engine.filters.VariantFilterLibrary;
 import org.broadinstitute.hellbender.transformers.VariantTransformer;
 import org.broadinstitute.hellbender.utils.IndexUtils;
+import org.broadinstitute.hellbender.utils.IntervalUtils;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 
+import java.util.List;
```
cmnbroad: `List` and `IntervalUtils` imports above are dead code now and can be removed.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
bbimber: ok

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
bbimber: ok.  i also updated the provider and java doc above

src/test/java/org/broadinstitute/hellbender/engine/GATKToolUnitTest.java
bbimber: ok

cmnbroad:
@bbimber Thank for working through all these issues!

PR #4497, opened 2018-03-05
HaplotypeCaller: add a good integration test for the --max-alternate-alleles argument

droazen:


droazen:
@ldgauthier Could you please review? It's a quick one

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
```diff
@@ -773,6 +773,88 @@ public void test100PercentContaminationNoCallsInGVCFMode() throws Exception {
...
+        // First, assert that there is at least one call in the VCF produced WITHOUT --max-alternate-alleles
+        // that has more than maxAlternateAlleles alt alleles, excluding NON_REF (otherwise the test below
+        // won't be meaningful).
+        boolean sawTooManyAllelesWithoutMaxAlternateArg = false;
+        for ( final VariantContext vc : callsNoMaxAlternateAlleles ) {
+            if ( getNumAltAllelesExcludingNonRef(vc) > maxAlternateAlleles ) {
+                sawTooManyAllelesWithoutMaxAlternateArg = true;
+            }
+        }
+        Assert.assertTrue(sawTooManyAllelesWithoutMaxAlternateArg,
```
ldgauthier: Ooh we should really do this more often.  I bet there are old integration tests where the data was changed and we lost coverage.

droazen:
@ldgauthier Thanks for the feedback -- I'll see if I can add some additional assertions about the actual alleles retained at each site that had more than the maximum.

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
droazen: It's a tradeoff -- it provides continuous proof that the test is meaningful, but at the expense of doubling the runtime of the test.

ldgauthier: Actually, if you're really concerned about doubling the runtime, you can merge those two loops since there should be the same number of VCs in the same order between the two files, right?  And if they don't line up then we're in big big trouble anyway.

droazen: You have to actually run the `HaplotypeCaller` twice to generate the two VCFs in the first place (with and without the argument being tested). Notice the two calls to `runCommandLine()` above -- that's what doubles the runtime.

droazen:
@ldgauthier Ok, I've addressed your comments by adding some extra assertions about the alleles and genotypes post-allele-subsetting. Ready for another look!

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
```diff
@@ -815,24 +818,58 @@ public void testMaxAlternateAlleles(final String bam, final String reference, fi
...
+
+                // CollectionUtils.isEqualCollection() will compare the lists of Alleles without
+                // regard to ordering
+                Assert.assertTrue(CollectionUtils.isEqualCollection(vc.getAlleles(), alleleSubsettingExpectedResult),
+                        "For call " + vc + " expected alleles after subsetting were: " + alleleSubsettingExpectedResult +
+                                 " but instead found alleles: " + vc.getAlleles());
+            }
+
+            // For completeness sake, also check the genotypes to ensure that no genotypes reference
+            // an allele not present in the VC:
```
ldgauthier: I'm pretty sure that htsjdk will throw an exception if there is an allele in a genotype that's not in the VC.  Don't ask me know I know.  (ahem)  So this is probably redundant.

droazen: Well, that's most likely true in practice, since `VariantContextBuilder` always requests the `VariantContext.Validation.GENOTYPES` validation when you set the `genotypes()`. But there is a way to bypass that check by calling `genotypesNoValidation()` on the builder instead of `genotypes()`, so I figure we might as well keep this sanity check around in case someone decides to do something crazy with a `VariantContextBuilder` one day...

PR #4501, opened 2018-03-06
added logging output to the bam writing step

jamesemery:
There is currently an issue with spark stderr output if running through the wrapper script, this should make it a little clearer what spark is doing after it finishes with the tools work.


jamesemery:
added logging output to the bam writing step for spark tools (#4501)

PR #4502, opened 2018-03-06
Enabled zero eigensamples to be specified for CreateReadCountPanelOfNormals.

### LINKED ISSUE #4499: Allow zero eigensamples in CreateReadCountPanelOfNormals.
See #4174, we can probably just put in a single zero eigenvalue/vector as we do there.

samuelklee:
Closes #4499.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/HDF5SVDReadCountPanelOfNormals.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/HDF5SVDReadCountPanelOfNormals.java
@@ -14,6 +14,7 @@ import org.apache.spark.mllib.linalg.Matrix;
 import org.apache.spark.mllib.linalg.SingularValueDecomposition;
 import org.apache.spark.mllib.linalg.distributed.RowMatrix;
 import org.broadinstitute.hdf5.HDF5File;
+import org.broadinstitute.hdf5.HDF5LibException;
 import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.tools.copynumber.CreateReadCountPanelOfNormals;
@@ -155,7 +156,15 @@ public final class HDF5SVDReadCountPanelOfNormals implements SVDReadCountPanelOf
 
     @Override
     public int getNumEigensamples() {
-        return (int) file.readDouble(PANEL_NUM_EIGENSAMPLES_PATH);
+        try {
+            return (int) file.readDouble(PANEL_NUM_EIGENSAMPLES_PATH);
+        } catch (final HDF5LibException e) {
+            //if the panel only contains a single sample or zero eigensamples were requested,
+            //we do not store singular values or eigenvectors in the panel;
+            //in this case, no value is written to PANEL_NUM_EIGENSAMPLES_PATH and an exception
+            //is thrown when we attempt to read it
+            return 0;
+        }
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


samuelklee:
An easy one for you, @mbabadi---thanks!

src/test/java/org/broadinstitute/hellbender/tools/copynumber/CreateReadCountPanelOfNormalsIntegrationTest.java
```diff
@@ -31,6 +31,7 @@
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.Callable;
```
samuelklee: Oops, I need to clean this up.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/SVDDenoisingUtils.java
```diff
@@ -123,13 +123,12 @@ static SVDDenoisedCopyRatioResult denoise(final SVDReadCountPanelOfNormals panel
         logger.info("Preprocessing and standardizing sample read counts...");
         final RealMatrix standardizedCopyRatioValues = preprocessAndStandardizeSample(panelOfNormals, readCounts.getCounts());
 
-        logger.info(String.format("Using %d out of %d eigensamples to denoise...", numEigensamples, panelOfNormals.getNumEigensamples()));
-
         final RealMatrix denoisedCopyRatioValues;
-        if (panelOfNormals.getOriginalReadCounts().length == 1) {
-            logger.warn("Only a single sample was used to build the panel of normals, not cannot perform denoising...");
+        if (panelOfNormals.getNumEigensamples() == 0) {
+            logger.warn("No eigensamples were available to perform denoising; denoised copy ratios will be identical to the standardized copy ratios...");
```
samuelklee: Reminder to change condition and message to include ‘|| numEigensamples == 0’.

samuelklee:
@mbabadi Can you take a quick look?

samuelklee:
@mbabadi Can we get this in the next release?  Should be quick.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/HDF5SVDReadCountPanelOfNormals.java
```diff
@@ -154,7 +155,15 @@ public double getVersion() {
 
     @Override
     public int getNumEigensamples() {
-        return (int) file.readDouble(PANEL_NUM_EIGENSAMPLES_PATH);
+        try {
+            return (int) file.readDouble(PANEL_NUM_EIGENSAMPLES_PATH);
+        } catch (final HDF5LibException e) {
```
mbabadi: Could you use a less general exception to catch this?

src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/HDF5SVDReadCountPanelOfNormals.java
```diff
@@ -192,11 +201,19 @@ public SAMSequenceDictionary getSequenceDictionary() {
 
     @Override
     public double[] getSingularValues() {
+        if (getNumEigensamples() == 0) {
+            throw new UnsupportedOperationException("No singular values were available.  " +
```
mwalker174: This is opinion, but I would avoid this kind of exception if at all possible (get these sometimes when using `Collections` and it can be a real bummer). You could have this return an empty array `new double[0]` instead and move this logic to the client code.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/HDF5SVDReadCountPanelOfNormals.java
```diff
@@ -192,11 +201,19 @@ public SAMSequenceDictionary getSequenceDictionary() {
...
+                    "This is because the panel only contains a single sample or no eigensamples were requested upon panel creation.");
+        }
         return file.readDoubleArray(PANEL_SINGULAR_VALUES_PATH);
     }
 
     @Override
     public double[][] getEigensampleVectors() {
+        if (getNumEigensamples() == 0) {
+            throw new UnsupportedOperationException("No eigensample vectors were available.  " +
```
mwalker174: Same here.

src/main/java/org/broadinstitute/hellbender/tools/copynumber/denoising/HDF5SVDReadCountPanelOfNormals.java
samuelklee: This is the exception thrown by both our wrapper and the Java HDF5 library, unfortunately---there's no more specific exception.

sooheelee:
Is this to allow for no coverage regions?

samuelklee:
No, this is to allow for denoising by the interval median only.

PR #4503, opened 2018-03-06
Changed CNV PoN to filter on equality to interval median percentile.

samuelklee:
This will require less tuning of `minimum-interval-median-percentile` to filter out completely uncovered intervals, which will avoid e.g. https://gatkforums.broadinstitute.org/gatk/discussion/11461/gatk-4-0-1-2-no-non-zero-singular-values-were-found-in-creating-a-panel-of-normals-for-somatic-cnv

For example, currently if 10% of my intervals are completely uncovered and I set the relevant parameter to 5%, the 5th percentile is then equal to zero.  However, because no intervals are strictly less than zero, none are filtered.  Changing this to filter on equality then gets rid of all 10% of the intervals as one would want to do in practice.


samuelklee:
An even easier one, @mbabadi!

samuelklee:
@mbabadi Can you take a quick look?

samuelklee:
@mbabadi Can we get this in the next release?  Should be quick.

Issue #4351, opened 2018-02-05
Add registry of unported GATK3 tools, output nice error message when user tries to run one


droazen:

cmnbroad:
Closed via #4347 .

Issue #4347, opened 2018-02-05
Implement registry of deprecated tools


droazen:
Like in GATK3.

cmnbroad:
Fixed via #4505.

PR #4505, opened 2018-03-07
Add registry for deprecated tools and annotations.

cmnbroad:
We should be able to use this single implementation to satisfy both https://github.com/broadinstitute/gatk/issues/4347 and https://github.com/broadinstitute/gatk/issues/4351.

Questions: @vdauwera What other tools that were dropped from GATK3 should be added to the list now? GATK3 also has a deprecated annotations list, which is included here, but is empty. Are there any annotations that should be listed ? I can't really implement/test that part unless I populate it with something.

Also, trying to run a missing tool is currently handled as an error, and surfaces in the context of a usage message. Perhaps that hides it too much:

screen shot 2018-03-07 at 11 25 40 am

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/Main.java
+++ b/src/main/java/org/broadinstitute/hellbender/Main.java
@@ -4,6 +4,7 @@ import com.google.cloud.storage.StorageException;
 import htsjdk.samtools.util.StringUtil;
 import org.broadinstitute.barclay.argparser.*;
 import org.broadinstitute.hellbender.cmdline.CommandLineProgram;
+import org.broadinstitute.hellbender.cmdline.DeprecatedToolsRegistry;
 import org.broadinstitute.hellbender.cmdline.PicardCommandLineProgramExecutor;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 import org.broadinstitute.hellbender.exceptions.PicardNonZeroExitException;
@@ -465,6 +466,28 @@ public class Main {
     }
 
     /**
+     * Get deprecation message for a tool
+     * @param toolName command specified by the user
+     * @return deprecation message string, or null if none
+     */
+    public String getToolDeprecationMessage(final String toolName) {
+        return DeprecatedToolsRegistry.getToolDeprecationInfo(toolName);
+    }
+
+    /**
+     * When a command does not match any known command, searches for a deprecation message, if any, or for similar
+     * commands.
+     * @return returns an error message including the closes match if relevant.
+     */
+    public String getUnknownCommandMessage(final Set<Class<?>> classes, final String command) {
+        final String deprecationMessage = getToolDeprecationMessage(command);
+        if (deprecationMessage != null) {
+            return deprecationMessage;
+        }
+        return getSuggestedAlternateCommand(classes, command);
+    }
+
+    /**
      * similarity floor for matching in getUnknownCommandMessage *
      */
     private static final int HELP_SIMILARITY_FLOOR = 7;
@@ -474,7 +497,7 @@ public class Main {
      * When a command does not match any known command, searches for similar commands, using the same method as GIT *
      * @return returns an error message including the closes match if relevant.
      */
-    public static String getUnknownCommandMessage(final Set<Class<?>> classes, final String command) {
+    public String getSuggestedAlternateCommand(final Set<Class<?>> classes, final String command) {
         final Map<Class<?>, Integer> distances = new LinkedHashMap<>();
 
         int bestDistance = Integer.MAX_VALUE;
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedToolsRegistry.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedToolsRegistry.java
@@ -1,0 +1,42 @@
+package org.broadinstitute.hellbender.cmdline;
+
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * When a tool is removed from GATK (after having been @Deprecated for a suitable period), an entry should
+ * be added to this list to issue a message when the user tries to run that tool.
+ *
+ * NOTE: Picard tools should be listed here as well, since by definition such tools will not be found in
+ * the Picard jar.
+ */
+public class DeprecatedToolsRegistry {
+
+    // Mapping from tool name to string describing the major version number where the tool first disappeared and
+    // optional recommended alternatives
+    private static Map<String, Pair<String, String>> deprecatedTools = new HashMap<>();
+
+    static {
+        // Indicate version in which the tool disappeared, and recommended replacement in parentheses if applicable
+        deprecatedTools.put("IndelRealigner", Pair.of("4.0.0.0", "Please use GATK3 to run this tool"));
+        deprecatedTools.put("RealignerTargetCreator", Pair.of("4.0.0.0", "Please use GATK3 to run this tool"));
+    }
+
+    /**
+     * Utility method to pull up the version number at which a tool was deprecated and the suggested replacement, if any
+     *
+     * @param toolName   the tool class name (not the full package) to check
+     */
+    public static String getToolDeprecationInfo(final String toolName) {
+        return deprecatedTools.containsKey(toolName) ?
+                String.format("%s is no longer included in GATK as of version %s. %s",
+                        toolName,
+                        deprecatedTools.get(toolName).getLeft(),
+                        deprecatedTools.get(toolName).getRight()
+                ) :
+                null;
+    }
+
+}
--- a/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
+++ b/src/main/java/org/broadinstitute/hellbender/cmdline/CommandLineProgram.java
@@ -26,7 +26,6 @@ import java.io.InputStream;
 import java.net.InetAddress;
 import java.net.URL;
 import java.text.DecimalFormat;
-import java.lang.Thread;
 import java.time.Duration;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
--- a/src/test/java/org/broadinstitute/hellbender/cmdline/DeprecatedToolsRegistryTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/cmdline/DeprecatedToolsRegistryTest.java
@@ -1,0 +1,37 @@
+package org.broadinstitute.hellbender.cmdline;
+
+import org.broadinstitute.hellbender.Main;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.Collections;
+
+public class DeprecatedToolsRegistryTest {
+
+    @Test
+    public void testRunMissingDeprecatedTool() {
+        final String missingTool = "IndelRealigner";
+
+        final UserException e = Assert.expectThrows(
+                UserException.class,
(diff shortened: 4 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/Main.java
```diff
@@ -480,6 +481,10 @@ public static String getUnknownCommandMessage(final Set<Class<?>> classes, final
         int bestDistance = Integer.MAX_VALUE;
         int bestN = 0;
 
+        final String deprecationMessage = DeprecatedArtifactsRegistry.getToolDeprecationInfo(command);
```
magicDGS: Please, make this configurable outside the `getUnknownCommandMessage` to override by downstream toolkits (if someone is trying to run `IndelRealignment` on my toolkit, I would like to throw an error saying that it does not exists but not that it is not included in GATK4 but I would like still to have the code for searching for similar commands...

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+public class DeprecatedArtifactsRegistry {
+
+    // Mapping from tool name to string describing the major version number where the tool first disappeared and
+    // optional recommended alternatives
+    private static Map<String, Tuple2<String, String>> deprecatedTools = new HashMap<>();
+
+    static {
+        // Indicate version in which the tool disappeared, and recommended replacement in parentheses if applicable
+        deprecatedTools.put("IndelRealigner", new Tuple2<>("4.0.0.0", "Please use GATK3 to run this tool"));
+        deprecatedTools.put("RealignerTargetCreator", new Tuple2<>("4.0.0.0", "Please use GATK3 to run this tool"));
```
magicDGS: If these tools are going to be included after I implement them, saying that they are deprecated might be a strong statement: maybe a distinction between "not supported yet (but probably future support)" and "deprecated" is a better idea. Otherwise, if they are included again after version x.x.x.x, this message in previous version will be weird.

An idea might be to include a placeholder command line programs that just throw an "unsupported in the current version" for that kind of tools and annotated with `@ExperimentalFeature`, for example. Anyway, I hope to have `IndelRealignment` sooner than later...

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+    static {
+        // Indicate version in which the annotation disappeared, and recommended replacement in parentheses if applicable
+    }
+
+    /**
+     * Utility method to pull up the version number at which a tool was deprecated and the suggested replacement, if any
+     *
+     * @param toolName   the tool class name (not the full package) to check
+     */
+    public static String getToolDeprecationInfo(final String toolName) {
```
magicDGS: Maybe instead of handle this here, a method to throw a new `UserException` in main if it is in the deprecation registry may be more informative and will handle common message....

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * When a tool is removed from GATK (after having been @Deprecated for a suitable period), an entry should
+ * be added to this list to issue a message when the user tries to run that tool.
+ *
+ * NOTE: Picard tools should be listed here as well, since by definition such tools will not be found in
+ * the Picard jar.
```
magicDGS: Maybe a registry in Picard that can add on runtime values will be helpful. The same for dowsntream tools...

src/main/java/org/broadinstitute/hellbender/Main.java

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
cmnbroad: If and when picard has such a thing we can consider integrating it here. for downstream toolkits, getUnknownCommand is overridable now, which should allow the flexibility to do whatever integration is necessary.

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
cmnbroad: I think these messages should reflect the current state, rather than speculate about the future. Hopefully we can get these tools ported; at that point we can remove them from this list.

src/main/java/org/broadinstitute/hellbender/Main.java
magicDGS: Thanks!

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
magicDGS: Fair enough.

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class DeprecatedArtifactsRegistry {
+
+    // Mapping from tool name to string describing the major version number where the tool first disappeared and
+    // optional recommended alternatives
+    private static Map<String, Tuple2<String, String>> deprecatedTools = new HashMap<>();
```
droazen: Use `org.apache.commons.lang3.tuple.Pair` instead of Scala `Tuple2`

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+
+    static {
+        // Indicate version in which the tool disappeared, and recommended replacement in parentheses if applicable
+        deprecatedTools.put("IndelRealigner", new Tuple2<>("4.0.0.0", "Please use GATK3 to run this tool"));
+        deprecatedTools.put("RealignerTargetCreator", new Tuple2<>("4.0.0.0", "Please use GATK3 to run this tool"));
+    }
+
+    // Mapping from walker name to major version number where the walker first disappeared and optional replacement options
```
droazen: `walker` -> `annotation`

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+    private static Map<String, String> deprecatedAnnotations = new HashMap<>();
```
droazen: Remove `deprecatedAnnotations` for now until there's a formal request for this feature.

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Utility method to pull up the version number at which a tool was deprecated and the suggested replacement, if any
+     *
+     * @param toolName   the tool class name (not the full package) to check
+     */
+    public static String getToolDeprecationInfo(final String toolName) {
+        return deprecatedTools.containsKey(toolName) ?
+                String.format("%s is no longer included in GATK as of version %s. (%s)",
```
droazen: `(%s)` -> `%s`

src/test/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistryTest.java
```diff
@@ -0,0 +1,21 @@
...
+    @Test
+    public void testRunMissingDeprecatedTool() {
+        final String missingTool = "IndelRealigner";
+
+        final UserException e = Assert.expectThrows(
+                UserException.class,
+                () -> new Main().instanceMain( new String[] {missingTool} )
+        );
+        Assert.assertTrue(e.getMessage().contains(DeprecatedArtifactsRegistry.getToolDeprecationInfo(missingTool)));
+    }
```
droazen: Add a second test for a tool that doesn't exist and is not in the registry.

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Map;
+
+/**
+ * When a tool is removed from GATK (after having been @Deprecated for a suitable period), an entry should
+ * be added to this list to issue a message when the user tries to run that tool.
+ *
+ * NOTE: Picard tools should be listed here as well, since by definition such tools will not be found in
+ * the Picard jar.
+ */
+public class DeprecatedArtifactsRegistry {
```
droazen: `MissingToolsRegistry` ?

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
cmnbroad: Removed completely.

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
cmnbroad: Done.

src/test/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistryTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedArtifactsRegistry.java
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/cmdline/DeprecatedToolsRegistry.java
```diff
@@ -0,0 +1,42 @@
...
+public class DeprecatedToolsRegistry {
+
+    // Mapping from tool name to string describing the major version number where the tool first disappeared and
+    // optional recommended alternatives
+    private static Map<String, Pair<String, String>> deprecatedTools = new HashMap<>();
+
+    static {
+        // Indicate version in which the tool disappeared, and recommended replacement in parentheses if applicable
+        deprecatedTools.put("IndelRealigner", Pair.of("4.0.0.0", "Please use GATK3 to run this tool"));
+        deprecatedTools.put("RealignerTargetCreator", Pair.of("4.0.0.0", "Please use GATK3 to run this tool"));
```
cmnbroad: @vdauwera Anything else that you want on this list before I merge this PR ?

Issue #4007, opened 2017-12-21
Generate new resources for CNV WDL tests.


samuelklee:
It's a little confusing that we use downsampled tumor BAMs for the gCNV WDL tests, for example.  Can wait until after release.

samuelklee:
@LeeTL1220 We'll keep you updated here.  Should be quick, but I want @asmirnov239 to finish up #4113 first.

I think if we can point @asmirnov239 to a few public germline BAMs to use for building both gCNV and ModelSegments PoNs, and we can continue using HCC1143 for the tumor.

LeeTL1220:
FYI.... @ruchim is going to use full size bams for the examples in FC.  If she changes her mind and wants to go back to the downsampled chr20 versions, she can comment here.

samuelklee:
OK.  I think that probably makes more sense.  I don't think we should use Travis test data as FC example data in general, as I'd want the latter to be able to showcase the workflows more realistically.

asmirnov239:
To generate PostprocessGermlineCNVCallsIntegrationTest test data - run gCNV on coverage files located in `gatk/src/test/resources/org/broadinstitute/hellbender/tools/copynumber/gcnv-sim-data/`
and ploidy calls in `
gatk/src/test/resources/org/broadinstitute/hellbender/tools/copynumber/gcnv-sim-data/contig-ploidy-calls/`

samuelklee:
Just noting here that I saw a lot of intermittent 60-minute Travis timeouts for the gCNV case mode WDL tests in #7411. I'm pretty sure that at some point we were running cohort + case together, so not sure why case alone is now hitting the limit. So might be worth investigating and tightening up the tests/data. Also note @ldgauthier's concerns about using a fake dictionary in the simulated data in #6957.

PR #4510, opened 2018-03-07
Changed filtering of normal hets on overlap with copy-ratio intervals…

samuelklee:
… in ModelSegments to be consistent with filtering of case hets.

Also some miscellaneous code cleanup.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/copynumber/ModelSegments.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/copynumber/ModelSegments.java
@@ -69,9 +69,10 @@ import java.util.stream.Stream;
  * <p>
  *     If both copy ratios and allele fractions are available, we perform segmentation using a combined kernel
  *     that is sensitive to changes that occur not only in either of the two but also in both.  However, in this case,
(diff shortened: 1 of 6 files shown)


samuelklee:
@sooheelee Are you comfortable reviewing this?  The code changes are minimal.

Note that there is no "regression test" since the test files do not give rise to the scenario you encountered in https://github.com/broadinstitute/dsde-docs/issues/2891.  Perhaps we can make sure that we include such a test case when we address https://github.com/broadinstitute/gatk/issues/4007.

sooheelee:
I'm only in an introductory Java class, geared towards Android app development. So I cannot comment on the code. What I have done is take the branch and test it against the data that I have and I can say the counts now match up to the lower expected value. Furthermore, the four questionable sites are now absent.

Before:
screenshot 2018-03-08 11 05 38

With changes:
screenshot 2018-03-08 11 06 06

Let me know if and how I can help with your efforts in creating a test if you decide to test for such a scenario.

samuelklee:
OK, thanks!  @MartonKN, can you take a quick look at the code?

PR #4517, opened 2018-03-09
minor doc change for -tranche argument

chandrans:
This is from https://github.com/broadinstitute/dsde-docs/issues/2967#event-1504632768. Users find the -tranche argument difficult to use. I hope this extra sentence with example will help.


ldgauthier:
Good clarification, but can you clean up the format a little? I'd like to see a colon instead of a comma after note and a hard return two wrap the long line (so that it lines up with the lines above).

droazen:
ping @chandrans

chandrans:
Done. Hope this is better.

ldgauthier:
:+1:

PR #4520, opened 2018-03-11
Option to analyze contributions of different filters in Concordance

davidbenjamin:
@takutosato This adds an option to report counts of false negatives, true negatives, and false/true negatives not flagged by any other filter for each filter.  Gives a rough picture of how valuable (or harmful) each filter is.


takutosato:
@skwalker @madduran you might find this new feature of Concordance useful for your TCGA analysis

PR #4522, opened 2018-03-12
Option to emit (but still filter) all germline sites in Mutect2

davidbenjamin:
@takutosato I made sure that when the new option is turned off, as it is by default, nothing changes.  Also, the new option surprisingly only added 30% or so to runtime because M2 already calls so many 
 false active regions.

@chandrans @sooheelee This is the feature we have been discussing.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
@@ -53,8 +53,6 @@ public class SomaticGenotypingEngine extends AssemblyBasedCallerGenotypingEngine
         public AFCalculator getInstance(final int ploidy, final int maximumAltAlleles) { return null; }
     };
 
-    private static final Logger logger = Logger.getLogger(SomaticGenotypingEngine.class);
-
     @Override
     protected String callSourceString() {
         return "M2_call";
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
```diff
@@ -222,6 +220,14 @@ private void addGenotypes(final LikelihoodMatrix<Allele> tumorLog10Matrix,
         callVcb.genotypes(genotypes);
     }
 
+    public static double[] getEffectiveCounts(final LikelihoodMatrix<Allele> log10LikelihoodMatrix) {
+        if (log10LikelihoodMatrix.numberOfReads() == 0) {
+            return new double[log10LikelihoodMatrix.numberOfAlleles()]; // zero couns for each allele
```
takutosato: couns -> counts

sooheelee:
Thanks for including us in the discussion @davidbenjamin.

> the new option surprisingly only added 30% or so to runtime because M2 already calls so many
false active regions.

This is a great value-add.

sooheelee:
Just to clarify @davidbenjamin @takutosato, the normal sample calling uses:

1. HaplotypeCaller logic for _germline_ calling OR
2. Mutect2 sensitive calling?

davidbenjamin:
@sooheelee This new option isn't doing germline calling per se.  Rather it calls all variants, somatic and germline, in the tumor sample.  To directly call germline events in the normal it wold be faster and better to use HaplotypeCaller.

sooheelee:
Thanks for the clarification. To paraphrase the following:

> This new option isn't doing germline calling per se. Rather it calls all variants, somatic and germline, in the tumor sample. 

We now have a parameter that allows inclusion of sites that are variant in the tumor that are also variant in the matched-normal and/or present in the germline resource. With the option enabled, these sites are emitted for the tumor and if the matched-normal is part of the analysis, then its variants for the sites are also noted in the callset. Finally, this parameter is set to true with `--genotype-germline-sites`.

davidbenjamin:
That's right, and just because more explicitness never hurts, here's another paraphrase.

M2 default: call somatic variation in the tumor sample.  Some possible germline variants will be included, but not intentionally.

New option: call *all* variation in the tumor sample.

In both case FilterMutectCalls is the same.  It uses the INFO fields from Mutect2 to put the `germline_risk` in the filter field.  If a user really wants to use Mutect2 as a slower HaplotypeCaller she could always set the germline posterior probability threshold to 1, effectively turning off this filter.

PR #4527, opened 2018-03-12
Add a note to the GATK README that a git clone is required to build GATK

droazen:
There have been several users confused that the sources zip file is not
buildable. Clarify that a full git clone is required for building GATK.


droazen:
@lbergelson please review

PR #4528, opened 2018-03-13
Exposed maximum chunk size in CNV panel of normals.

### LINKED ISSUE #4365: Expose or decrease CHUNK_DIVISOR in HDF5SVDReadCountPanelOfNormals.
The current hardcoded limit restricts us to 16777215 bins, which we can hit as we go below 200bp bins.  For a given matrix that needs to be broken into chunks, increasing the chunk size by decreasing CHUNK_DIVISOR will require more heap space.

In practice, it's not too hard to build a PoN per contig to get around this limit at the moment.  We might start having issues with doing SVD on the entire matrix anyway when we have this many bins.  But it might be good to have the option exposed.

samuelklee:
This enables PoNs to be built with a number of intervals greater than 16777216, which lets us push down to bin sizes <175bp if really necessary.  At some point, it is better to scatter and build a separate PoN for each contig so we don't run into memory issues.  However, the official WDL is not set up to do the scatter, so one would have to write a custom WDL.

@mwalker174 Do you mind taking a look?

Closes #4365.

### DIFF
--- a/scripts/cnv_wdl/somatic/cnv_somatic_panel_workflow.wdl
+++ b/scripts/cnv_wdl/somatic/cnv_somatic_panel_workflow.wdl
@@ -68,6 +68,7 @@ workflow CNVSomaticPanelWorkflow {
     Boolean? do_impute_zeros
     Float? extreme_outlier_truncation_percentile
     Int? number_of_eigensamples
+    Int? maximum_chunk_size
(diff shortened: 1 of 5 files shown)


samuelklee:
@mwalker174 Can you take a quick look?

samuelklee:
@mwalker174 Let me know if I should assign this to someone else.  Thanks!

PR #4542, opened 2018-03-20
removing unimplemented interface FileExtension

lbergelson:
removing an unused interface that was carried over from GATK3
removing check for file subtypes that is also no longer relevant


src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -166,7 +165,7 @@ else if (file.exists())
      * @return the absolute path to the file.
      */
     public static File absolute(File file) {
-        return replacePath(file, absolutePath(file));
+        return new File(absolutePath(file));
```
jonn-smith: It may be worth checking if we have to do this at all.

Should open a new issue for this.

lbergelson: I opened #4543 to track this issue.

PR #4547, opened 2018-03-21
setting default compression level to 2

lbergelson:
I'm honestly not sure which of the various settings take precedence, so I set them all...  We should remove the irrelevant ones.  I assume that that includes the ones in the `gatk` launch script as well as the defaults in build.gradle?


lbergelson:
@jonn-smith I forgot to change the default in the GATKConfig.java, and your test caught it.  That's awesome!

PR #4548, opened 2018-03-21
add chr20 and chr21 alts to GRCh38 reference

sooheelee:
For @droazen.

MD5 of new ref is a5e22d1d7ac48d1723df7abe4be48566


sooheelee:
Thanks for the shell script to fix unexpected behavior in samtools faidx!

Issue #4556, opened 2018-03-22
Blacklist the Picard NIO dependency in build.gradle


droazen:
The latest Picard release introduces a dependency on the Google Cloud NIO library that conflicts with GATK's dependency. We are going to have to blacklist the Picard NIO dependency for now. 

Longer term, we might want to consider having both projects depend upon a build of htsjdk that comes with the NIO plugin.

lbergelson:
The problem seems to be fixed in picard with @cmnbroad's change to the cloud configuration.  Thew pom for 2.18.1+ looks like it won't include nio.

lbergelson:
It's a good idea to force our version anyway though.

PR #4560, opened 2018-03-22
Add a force resolution strategy for google-cloud-nio.

cmnbroad:
So we don't accidentally pick up a different version via Picard. Fixes https://github.com/broadinstitute/gatk/issues/4556.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,6 +63,10 @@ final sparkVersion = System.getProperty('spark.version', '2.0.2')
 final hadoopBamVersion = System.getProperty('hadoopBam.version','7.9.1')
 final genomicsdbVersion = System.getProperty('genomicsdb.version','0.9.2-proto-3.0.0-beta-1+uuid-static')
 final testNGVersion = '6.11'
+// Using the shaded version to avoid conflicts between its protobuf dependency
+// and that of Hadoop/Spark (either the one we reference explicitly, or the one
+// provided by dataproc).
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -80,6 +84,8 @@ configurations.all {
         // force testng dependency so we don't pick up a different version via GenomicsDB
         force 'org.testng:testng:' + testNGVersion
         force 'org.broadinstitute:barclay:' + barclayVersion
+        // make sure we don't pick up a different version via Picard
+        force googleCloudNioDependency
```
cmnbroad: I just realized this isn't sufficient. We actually need to forcibly reject the google component because it has a different name.

build.gradle
```diff
@@ -80,6 +84,17 @@ configurations.all {
...
+        // make sure we don't pick up an incorrect version of the GATK variant of the google-nio library
+        // via Picard, etc., or pick up the google variant of it
+        force googleCloudNioDependency
+        componentSelection {
+            // This feature is incubating, and although it works, when its triggered it doesnt propagate
+            // the rejection message so the build fails saying it can't find google-cloud-nio.
+            withModule('com.google.cloud:google-cloud-nio') {
+                ComponentSelection selection -> selection.reject("The GATK version of this component is required")
+            }
+        }
```
cmnbroad: I tested this by including an explicit dependency on google-nio. The component-rejection statement (which is incubating) seems to work and causes the build to fail, but the error message is misleading. Instead of seeing the argument to the reject method, Gradle just says it can't find google-nio.

Issue #4111, opened 2018-01-09
Switch to the updated type & location inference tool in SV pipeline


SHuang-Broad:
As we have finished implementing the updated logic for how variants are interpreted and location inferred by studying local assembly contig alignment signatures, it is time to clean up the corresponding package in the pipeline and make the switch to the updated implementation, which now outputs not only insertion, deletion, small tandem duplication, and inversions, but also novel adjacencies (BND records whose meanings cannot be fully resolved solely from assembly alignment signatures) as well as complex variants that theoretically could be arbitrarily complex (`<CPX>`, as long as we have assembled across the full event). 

## Planed organization

the `discovery` package could be divided roughly now into

### interface

`SvDiscoveryDataBundle`, `SvDiscoverFromLocalAssemblyContigAlignmentsSpark`, `SvType`, `AnnotatedVariantProducer`

### alignment prep (sub package)

`AlignmentInterval`, `AlignedContig` (refactor `AssemblyContigWithFineTunedAlignments` into `AlignedContig`), `AlignedContigGenerator`, `AlignedAssembly`, `ContigAlignmentsModifier` (refactor `AlnModType` into it), `GappedAlignmentSplitter`, `StrandSwitch`, `FilterLongReadAlignmentsSAMSpark` (factor out the major methods in the new alignment filter by score into a 1st level class)

### type & location inference (sub package)

* imprecise: refactor out methods from to-be-deprecated `DiscoverVariantsFromContigAlignmentsSAMSpark`

* alignment classification:  `ChimericAlignment` and `NovelAdjacencyReferenceLocations` (very tricky to decouple the functionalities because both have over 50 uses), `AssemblyContigAlignmentSignatureClassifier`, `VariantDetectorFromLocalAssemblyContigAlignments`

* simple: `SimpleSVType`, `SvTypeInference`, `InsDelVariantDetector`, `BreakpointComplications` (rename to `BreakpointComplicationsForSimpleTypes`)

* complex: `BreakEndVariantType`, `SuspectedTransLocDetector`, `SimpleStrandSwitchVariantDetector`

### deprecated

`DiscoverVariantsFromContigAlignmentsSAMSpark` 

It currently provides 3 groups of functionalities:

[...]

PR #4562, opened 2018-03-22
(SV) Misc. improvements to the new location inference and type interpretation tool

SHuang-Broad:
- [x] Output bam instead of sam for assembly alignments
- [x] Instead of creating directory, new interpretation tool writes files (behavior consistent with current interpretation tool)
- [x] Prefix with sample name for output files' names
- [x] Add `INSLEN` annotation when there's `INSSEQ`
- [x] Clarify the boundary between `AlignedContig` and `AssemblyContigWithFineTunedAlignments`
- [x] Increase test coverage for `AssemblyContigAlignmentsConfigPicker`

Up to date plans for more cleanups and improvements posted in #4111


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -110,14 +111,14 @@
             = new DiscoverVariantsFromContigsAlignmentsSparkArgumentCollection();
     @Argument(doc = "sam file for aligned contigs", fullName = "contig-sam-file")
     private String outputAssemblyAlignments;
-    @Argument(doc = "filename for output vcf", shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+    @Argument(doc = "prefix for output vcf; sample name will be appended after the provided argument",
```
cwhelan: Is this assuming that if you want files to appear in a directory then the argument should end with a trailing slash? I'd like to avoid that if possible. Can we handle this with eg the Path API? 

I.e. Let's make this actually call for a directory. Then we should validate that the directory actually exists and create it if it's not there. Finally we can create the files in that directory using an API call, like `Paths.resolve(outputDirectory, filename)` for each filename we want.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -110,14 +111,14 @@
...
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
             fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME)
-    private String vcfOutputFileName;
+    private String outputPrefix;
     @Advanced
-    @Argument(doc = "directory to output results of our prototyping breakpoint and type inference tool in addition to the master VCF;" +
-            " the directory contains multiple VCF's for different types and record-generating SAM files of assembly contigs,",
-            fullName = "exp-variants-out-dir", optional = true)
-    private String expVariantsOutDir;
+    @Argument(doc = "prefix to output files of our prototyping breakpoint and type inference tool in addition to the master VCF;",
```
cwhelan: remove "in addition to the master VCF" from this doc line

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -110,14 +111,14 @@
...
+            fullName = "exp-variants-out-prefix", optional = true)
```
cwhelan: let's keep this called `directory` according to my comment above

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -163,8 +164,9 @@ protected void runTool( final JavaSparkContext ctx ) {
         if(parsedAlignments.isEmpty()) return;
 
         final Broadcast<SVIntervalTree<VariantContext>> cnvCallsBroadcast = broadcastCNVCalls(ctx, headerForReads, discoverStageArgs.cnvCallsFile);
+        final String outputPrefixWithSampleName = outputPrefix + (outputPrefix.endsWith("/") ? "" : "_") + SVUtils.getSampleId(headerForReads);
```
cwhelan: if the user accidentally leaves off the trailing slash the files will not go into the directory requested but will be in its parent with a confusing name, i think.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/SvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
```diff
@@ -306,21 +299,26 @@ private static void forNonComplexVariants(final EnumMap<RawTypes, JavaRDD<Assemb
...
 
         final Set<String> filteredReadNames = new HashSet<>( filteredContigs.map(AssemblyContigWithFineTunedAlignments::getContigName).distinct().collect() );
         toolLogger.info(filteredReadNames.size() + " contigs indicating " + rawTypeString);
+        final SAMFileHeader header = headerBroadcast.getValue();
         final JavaRDD<SAMRecord> splitLongReads = originalAlignments.filter(read -> filteredReadNames.contains(read.getName()))
-                .map(read -> read.convertToSAMRecord(headerBroadcast.getValue()));
-        SVFileUtils.writeSAMFile(outputDir+"/"+rawTypeString+".sam", splitLongReads.collect().iterator(),
-                headerBroadcast.getValue(), false);
+                .map(read -> read.convertToSAMRecord(header));
+        header.setSortOrder(SAMFileHeader.SortOrder.queryname);
```
cwhelan: Why sort these by queryname and not by position? I could be convinced either way but this requires an extra step to load the file into IGV.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -391,7 +391,7 @@ private void experimentalInterpretation(final JavaSparkContext ctx,
                         final String contigName = AlignedAssemblyOrExcuse.formatContigName(alignedAssembly.getAssemblyId(), contigIdx);
                         final List<AlignmentInterval> arOfAContig
                                 = getAlignmentsForOneContig(contigName, contigSequence, allAlignments.get(contigIdx), refNames, header);
-                        return new AlignedContig(contigName, contigSequence, arOfAContig, false);
+                        return new AlignedContig(contigName, contigSequence, arOfAContig);
```
cwhelan: Can we change this variable name? `arOfAContig` is not obvious any more. Maybe `alignmentsForContig`?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
```diff
@@ -434,11 +439,12 @@ private static AssemblyContigWithFineTunedAlignments updateContigMappingsWithGap
 
     /**
      * when two configurations are the same,
-     * put the one with less alignments,
-     * or less summed mismatches if still tie
-     * first
+     * implement ordering that
+     *      prefer the configuration with less alignments, then
```
cwhelan: "prefers"

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
```diff
@@ -434,11 +439,12 @@ private static AssemblyContigWithFineTunedAlignments updateContigMappingsWithGap
...
+     *      prefer the configuration less summed mismatches if still tie
```
cwhelan: "and prefers the configuration with a lower number of summed mismatches in case of a tie"

src/test/java/org/broadinstitute/hellbender/utils/StringSplitSpeedUnitTest.java
```diff
@@ -19,7 +19,7 @@
             "Italiam, fato profugus, Laviniaque venit " +
             "litora, multum ille et terris iactatus et alto " +
             "vi superum saevae memorem Iunonis ob iram; " +
-            "multa quoque et bello passus, dum conderet urbem, " +
+            "multa quoque et bello passus, testConfigurationSorting conderet urbem, " +
```
cwhelan: was this maybe a global search/replace change? My high school latin is pretty rusty but I don't think `testConfigurationSorting` is a word that can replace a conjunction like 'dum'. :)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
SHuang-Broad: right. thanks for catching that. corrected

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/SvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
SHuang-Broad: that was because I basically used Ribbon for reviewing those alignments, 
which is definitely easier with SAM compared to with BAM.
Now changed to adding an parameter to the utility 
* if BAM, sort by coordinate order.
* if SAM, sort by query name.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AssemblyContigAlignmentsConfigPicker.java
SHuang-Broad: done

src/test/java/org/broadinstitute/hellbender/utils/StringSplitSpeedUnitTest.java
SHuang-Broad: definitely better than mine, who initially named a function "dum" :cold_sweat:

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
SHuang-Broad: is the commit 2fd080dcd17b4d19f9efbed8028bb44f631db793 similar to what you have in mind?

droazen:
@SHuang-Broad The `testSAMWriter_chr20.bam` bam that you added here should probably have gone into git lfs (`src/test/resources/large`), rather than directly into the repo. It's too late to fix (once it's in master, it's in the git history forever), but in the future please try to store test data bigger than ~1-2 MB or so in lfs.

PR #4563, opened 2018-03-22
Fixes for HG38 exception and better logging.

### LINKED ISSUE #4521: Funcotator hg38 data source not working; configuration files contains reference to files not present
From ftp://gsapubftp-anonymous@ftp.broadinstitute.org/bundle/funcotator/ the file `gencode/hg38/gencode.config` makes reference to `gencode.v27.pc_transcripts.fasta` 

```
# Required field for GENCODE files.
# Path to the FASTA file from which to load the sequences for GENCODE transcripts:
gencode_fasta_path = gencode.v27.pc_transcripts.fasta
```

which is not present in the directory. 

```
$ ls
gencode.config
gencode.v27.chr_patch_hapl_scaff.annotation.REORDERED.gtf
gencode.v27.chr_patch_hapl_scaff.annotation.REORDERED.gtf.idx
gencode.v27.transcripts.dict
gencode.v27.transcripts.fasta
gencode.v27.transcripts.fasta.fai
```

Downloading the pc_transcript.fasta from gencode does not fix the errors thrown when trying to use the reference.  The error message is a fairly uninformative null pointer exception.  

```
00:27:35.745 INFO  Funcotator - Done initializing engine
00:27:35.758 INFO  Funcotator - Shutting down engine
[March 12, 2018 12:27:35 AM UTC] org.broadinstitute.hellbender.tools.funcotator.Funcotator done. Elapsed time: 0.01 minutes.
Runtime.totalMemory()=2258108416
	at org.broadinstitute.hellbender.tools.funcotator.Funcotator.closeTool(Funcotator.java:330)
	at org.broadinstitute.hellbender.engine.GATKTool.doWork(GATKTool.java:897)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:135)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:180)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:199)
	at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:159)
	at org.broadinstitute.hellbender.Main.mainEntry(Main.java:202)
	at org.broadinstitute.hellbender.Main.main(Main.java:288)
```

jonn-smith:
Added more logging functionality to better inform users.

Fixed NullPointerException when onTraversalStart throws.
Now logs version number information for data sources.
Now checks minimum data sources version and requires that the user
upgrade if the version is too old.

Fixes #4521 in a roundabout way - data sources need to be re-released for this version to run at all, but these new data sources contain a fix for the HG38 issue.


src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -397,10 +397,13 @@ public Object onTraversalSuccess() {
     public void closeTool() {
 
         for(final DataSourceFuncotationFactory factory : dataSourceFactories) {
-            factory.close();
+            if ( factory != null ) {
```
LeeTL1220: Would this be better as `Closeable`?

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
```diff
@@ -371,6 +410,147 @@ private static Properties readConfigFileProperties(final Path configFilePath) {
...
+            logger.error("ERROR: Given data source path is too old!  Minimum required version is: " + CURRENT_MINIMUM_DATA_SOURCE_VERSION + " (yours: " + version + ")");
+            logger.error("       You must download a newer version of the data sources from the Broad Institute FTP site: " + DATA_SOURCES_FTP_PATH);
+            logger.error("       or the Broad Institute Google Bucket: " + DATA_SOURCES_BUCKET_PATH);
+        }
+
+        return dataSourcesPathIsAcceptable;
+    }
+
+    @VisibleForTesting
+    static boolean validateVersionInformation(final int major, final int minor, final int year, final int month, final int day) {
```
LeeTL1220: No need for `final` on primitives for method parameters.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
```diff
@@ -371,6 +410,147 @@ private static Properties readConfigFileProperties(final Path configFilePath) {
...
+                }
+
+                if ( alternateSource == null ) {
+                    logger.warn("Unable to read alternate source information from data sources info/readme file: " + readmePath.toUri().toString());
+                }
+                else {
+                    logger.info("Data sources alternate source: " + alternateSource);
+                }
+            }
+            catch (final Exception ex) {
```
LeeTL1220: Please log information about the exception.  Generally, you should not be catching `Exception`, but rather a subclass of `Exception`.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
```diff
@@ -371,6 +410,147 @@ private static Properties readConfigFileProperties(final Path configFilePath) {
...
+     * We assume the data sources path is OK in the case that the version information cannot be read because the
+     * user can create their own data sources directory, which may not contain the metadata we seek.
+     * @param dataSourcesPath {@link Path} to a Data Sources directory to check.
+     * @return {@code True} if the given {@code dataSourcesPath} is equal to or newer than the minumum version, or if the version is unreadable.  {@code False} otherwise.
+     */
+    private static boolean logDataSourcesInfo(final Path dataSourcesPath) {
+
+        boolean dataSourcesPathIsAcceptable = true;
+
+        final Path readmePath = dataSourcesPath.resolve(IOUtils.getPath(README_FILE_NAME));
```
LeeTL1220: Where is documentation about the format of the README file?

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
```diff
@@ -275,10 +277,12 @@ public void spotCheck2() throws IOException {
 
         final ArgumentsBuilder arguments = new ArgumentsBuilder();
 
+        arguments.addArgument("verbosity", "INFO");
+
         arguments.addVCF(new File(inputFile));
         arguments.addReference(new File("/Users/jonn/Development/references/Homo_sapiens_assembly19.fasta"));
 
-        arguments.addArgument(FuncotatorArgumentDefinitions.DATA_SOURCES_PATH_LONG_NAME, "/Users/jonn/Development/funcotator_dataSources.v1.1.20180208");
+        arguments.addArgument(FuncotatorArgumentDefinitions.DATA_SOURCES_PATH_LONG_NAME, "/Users/jonn/Development/funcotator_dataSources_latestjnkjjk");
```
LeeTL1220: This pathname can't possibly be correct.  Must be something that would be on anyone's dev machine.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
```diff
@@ -275,10 +277,12 @@ public void spotCheck2() throws IOException {
 
         final ArgumentsBuilder arguments = new ArgumentsBuilder();
 
+        arguments.addArgument("verbosity", "INFO");
+
         arguments.addVCF(new File(inputFile));
         arguments.addReference(new File("/Users/jonn/Development/references/Homo_sapiens_assembly19.fasta"));
```
LeeTL1220: See next comment.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
```diff
@@ -235,6 +235,8 @@ public void spotCheck() throws IOException {
 
             final ArgumentsBuilder arguments = new ArgumentsBuilder();
 
+            arguments.addArgument("verbosity", "INFO");
+
             arguments.addVCF(new File("/Users/jonn/Development/M2_01115161-TA1-filtered.vcf"));
```
LeeTL1220: Please remove all references to `/Users/jonn/Development/` etc

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
```diff
@@ -78,9 +107,19 @@ private DataSourceUtils() {}
         // Go through our directories:
         final Set<String> names = new LinkedHashSet<>();
         for ( final String pathString : dataSourceDirectories ) {
+
+            logger.info("Initializing data sources from directory: " + pathString);
+
             final Path p = IOUtils.getPath(pathString);
             if ( !isValidDirectory(p) ) {
-                logger.warn("WARNING: Given path is not a valid directory: " + p.toUri().toString());
+                logger.error("ERROR: Given data source path is not a valid directory: " + p.toUri().toString());
```
LeeTL1220: I notice that you have a lot of error messages that do not throw exceptions.  Is this correct behavior?

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
```diff
@@ -371,6 +410,147 @@ private static Properties readConfigFileProperties(final Path configFilePath) {
...
+                Integer versionMonth   = null;
+                Integer versionDay     = null;
+                String source          = null;
+                String alternateSource = null;
+
+                // Get the info from our README file:
+                String line = reader.readLine();
+                while ((line != null) && ((version == null) || (source == null) || (alternateSource == null))) {
+
+                    if (version == null && line.startsWith(README_VERSION_LINE_START)) {
```
LeeTL1220: Why have a README and a config file?  Can't we merge those?

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
```diff
@@ -371,6 +410,147 @@ private static Properties readConfigFileProperties(final Path configFilePath) {
...
+                            versionMajor  = Integer.valueOf(m.group(1));
+                            versionMinor  = Integer.valueOf(m.group(2));
+                            versionYear   = Integer.valueOf(m.group(3));
+                            versionMonth  = Integer.valueOf(m.group(4));
+                            versionDay    = Integer.valueOf(m.group(5));
+
+                            version = versionMajor + "." + versionMinor + "." + versionYear + "" + versionMonth + "" + versionDay;
+                        }
+                        else {
+                            logger.warn("README file has improperly formatted version string: " + line);
```
LeeTL1220: README files typically have minimal to zero formatting.  This would indicate that we have a stricter format.  Is there a good reason not to merge with the config file?

LeeTL1220:
@jonn-smith Back to you.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
jonn-smith: Sure.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
jonn-smith: Sounds good.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
jonn-smith: Sounds good.

I want to catch any exceptions that happen there and continue so that it doesn't actually fail.  This is an optional check for logging purposes anyway.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
jonn-smith: I'll add in an explanation of what is required in the README for these purposes.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
jonn-smith: This test is intermediate and planned to be phased out.

It is checked in so that I can run my local test data and not have to worry about rewriting tests every time I need to debug something.

I added the flag `doDebugTests` to control for this and a unit test to ensure that `doDebugTests` is always set to false for the tests to pass.

I want to leave these in for now so I can continue to test locally with different settings easily.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
jonn-smith: This test is intermediate and planned to be phased out.

It is checked in so that I can run my local test data and not have to worry about rewriting tests every time I need to debug something.

I added the flag doDebugTests to control for this and a unit test to ensure that doDebugTests is always set to false for the tests to pass.

I want to leave these in for now so I can continue to test locally with different settings easily.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
jonn-smith: It's an error to have an invalid data source path.

You're right - I'll throw an exception here.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
jonn-smith: I don't understand... there is no config file for the data sources directory - only a README file.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
jonn-smith: @LeeTL1220 There is no config file for a data sources directory - only a README file.

I guess we can call it a config file, but then no one is likely to read it and it provides information to anyone who would download / use the data sources.

jonn-smith:
OK, @LeeTL1220 - First round comments addressed.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
LeeTL1220: Except that the codebase does not do this elsewhere.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
LeeTL1220: Ok... I thought this code was being run for each datasource.  Not the datasources as a whole...

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/DataSourceUtils.java
LeeTL1220: Leave w/ README.  Again, I thought this was being run for each datasource.  Not once for all datasources.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
LeeTL1220: But why can't you put these tests into the source tree and then remove if/when no longer needed?
You are asking to put these into master by issuing this PR and we cannot cut a release with these tests -- and master needs to always be release-ready.  

There should be an alternative way of doing this.

jonn-smith:
@LeeTL1220 that's totally fair.  There are a few problems with putting in the tests, though.

I don't know that I can publicly release the data that are being tested because of their provenance (pretty sure I can't).  

The second reason is that the whole reference sequences are not checked into git lfs.  Without it I can't annotate on a set of variants from around the whole genome, and the whole reference is pretty big.  

The last thing I can think of is a data source problem - I don't have all the data sources checked into git lfs (for similar reasons, especially now with dbSNP in there).

Also, these have already made their way into master.  So if this is really a problem, we should get them out of there ASAP.

LeeTL1220:
@jonn-smith Let's discuss offline.  Yes, let's get these out of master.

PR #4564, opened 2018-03-23
Replaced CollectFragmentCounts with CollectReadCounts.

### LINKED ISSUE #4519: Consider adding CollectReadCounts and further evaluate count-collection strategies.
In the process of unifying CalculateTargetCoverage / SparkGenomeReadCounts for the rewrite of the CNV pipeline, we decided to experiment with switching over to fragment-based counts due to a request from CGA.  For each fragment, CollectFragmentCounts adds a count to *the bin that overlaps with the fragment center*.  We filter to properly-paired, first-of-pair reads in order to have well formed fragments and avoid double counting.  We also filter out duplicates.

In contrast, CalculateTargetCoverage added a count to *all bins that overlapped with a read* and SparkGenomeReadCounts added a count to *the bin that contained the read start*.  These tools kept duplicates.

However, none of these collection strategies have been rigorously evaluated.

Using a small set of WGS SV tandem-duplication calls from @mwalker174 as a truth set, I did some experimenting with changing the count-collection strategy.  (We initially thought we were missing some of these simply due to over-denoising/filtering by the PoN, but as we'll see below, the count-collection strategy plays a non-trivial role.)

Subsetting to chr3, I built a small PoN of 12 normals (including the case normal) at 100bp and denoised using bin medians only (i.e., `--number-of-eigensamples 0`) to avoid denoising away common events.  In chr3, the case sample had three events:

````
chr3	8559423		8560126
chr3	64547471	64549936
chr3	90414457	90415989
````

I tried the following, running `ModelSegments` using fairly sensitive parameters (`--number-of-changepoints-penalty-factor 0.1 --maximum-number-of-segments-per-chromosome 10000 --window-size 16 --window-size 32 --maximum-number-of-smoothing-iterations 0` in copy-ratio-only mode:

1) CollectFragmentCounts.  This only recovered event 2.
2) CollectReadCounts - same as CollectFragmentCounts, but removing the properly-paired and first-of-pair filters and adding a count for each read to the bin containing its start.  This recovered all 3 events.
3) CollectFragmentOverlaps - same filters as CollectFragmentCounts, but adding counts to all bins overlapping each fragment.  Note that we need to implement a filter on maximum fragment length, otherwise we get some strange artifacts from (incorrectly mapped?) extremely long fragments; I arbitrarily chose a cutoff of 10000bp.  This recovered events 1 and 2.

Event 3 seemed to be the most difficult to recover.  Plotting the copy ratios surrounding this event (which spans ~15 100bp bins) yields some insights:

CollectFragmentCounts:
image

CollectReadCounts:
image

CollectFragmentOverlaps:
image

The increased statistical noise in the CollectFragmentCounts result (due to the lower overall count because of the pairing of reads) probably causes us to miss this event.  Also, although CollectFragmentOverlaps initially looks pretty good, I think the bin-to-bin correlations that are evident here negatively affect segmentation.

This is not an extremely rigorous evaluation, but it suggests that we should consider switching over to a CollectReadCounts-like strategy.   To appease CGA (in case they still feel strongly, which they might not), we can make this a separate tool, or perhaps just have a single tool called CollectCounts that can toggle between the two (we just have to be careful about filters in the latter case).  We should evaluate further once more rigorous automatic evaluations are in place.

@mbabadi @LeeTL1220 @asmirnov239 @sooheelee might find this of interest.  Also, I have a question for engine team, @droazen.  Is there a read filter I should be using for fragment length, and if not, can we add one (or is there already a preferred way to do this type of filtering)?

samuelklee:
Closes #4519.

Note that we should still explore and evaluate other collection strategies in the future.  However, I think we can go ahead and use this strategy as the default for now.

@asmirnov239 @mbabadi Let's get this merged quick and run the evaluations for the AACR poster using this collection strategy.  We should also run some subset of the evaluations with CollectFragmentCounts and check that results improve.

@LeeTL1220 @MartonKN We should rerun the somatic evaluations as necessary to make sure they still look good.  We may also want to show any interested parties in CGA the relevant results.

@sooheelee Note that some tutorial material, slides, etc. should be updated at the appropriate time (and may need to be updated again as we continue to tweak the strategy).


mbabadi:
PR looks good, thanks!

PR #4569, opened 2018-03-23
prevent uploadArchives from including distZip and distTar

### LINKED ISSUE #4567: Reduce size of artifactory snapshots by eliminating waste
Currently when we publish an artifactory snapshot, we upload both the maven jars (~50 MB) plus the fully-packaged jars (~500 MB). We only need the maven jars! By eliminating the packaged jars, we can reduce our artifact size by an order of magnitude.

Once this is done, we should talk to @davidbernick about increasing the artifact expiration time from the current 60 days to something longer.

lbergelson:
fixes #4567

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -610,7 +610,11 @@ artifacts {
     archives jar
     archives javadocJar
     archives sourcesJar
+
 }
+//remove zip and tar added by the application plugin
+configurations.archives.artifacts.removeAll {it.file =~ '.zip$'}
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -610,7 +610,11 @@ artifacts {
     archives jar
     archives javadocJar
     archives sourcesJar
+
 }
+//remove zip and tar added by the application plugin
```
droazen: Will our maven upload still have all the artifacts required by maven central? Isn't there a javadoc zip?

Issue #4376, opened 2018-02-08
Spark performance regression in BQSR and HC


tomwhite:
Running the reads pipeline on latest master on GCP on an exome shows a major performance regression. At the end of last year this took 40 minutes for the whole job, now using master BQSR is taking 1.4 hours, and HC is taking well over an hour (it hasn't finished yet).

#4278 is the likely culprit for the HC slowdown. Not sure about BQSR.

cc @lbergelson @droazen

frank-y-liu:
@tomwhite How many physical cores are you using? Is it a whole genome task? I was able  to finish BQSR in about an hour. From monitored activities, the CPU utilization rates are very high across all physical cores (so it's not bad). There is the long runtime of creating KnownSitesCache, as in #4264

droazen:
@tomwhite That's interesting -- @lbergelson 's profiling on `HaplotypeCallerSpark` in isolation did not show any noticeable slowdown as a result of the switch to the new code path (in addition to the output being much more correct than before), and the walker version which uses the same codepath is as fast or faster than GATK3 at this point in all modes. We should try to see whether we can replicate your results on our end. Can you provide more details on how you ran the tools?

The new `HaplotypeCaller` code path does result in the tool doing more work in some cases as compared to the early betas, particularly for genomic locations not covered by any reads -- but this work is necessary in order to produce correct output without boundary effects, and the fact that the early betas didn't do it was an oversight. I wonder if that could explain what you're seeing -- see the full discussion in https://github.com/broadinstitute/gatk/issues/4169.

There are also several ways in which `HaplotypeCallerSpark` currently makes suboptimal use of the new code path introduced in https://github.com/broadinstitute/gatk/pull/4278:

1. It uses only a single interval per shard, instead of many intervals as the walker version does (https://github.com/broadinstitute/gatk/issues/4299).

2. It materializes all of the assembly regions in a shard at once (https://github.com/broadinstitute/gatk/issues/4301), as opposed to the walker version which materializes the regions in a shard as lazily as possible.

3. The default read shard size may need adjusting (https://github.com/broadinstitute/gatk/issues/4298). The walker version creates one shard per contig, and lazily creates the assembly regions within that shard. That's obviously not possible for the Spark version, but I don't think that the effect of the shard size on performance has been measured yet for the Spark version.

lbergelson:
@tomwhite My tests were running HaplotypeCallerSpark on an exome on gcs.  It took ~40 minutes with and without the changes.

[...]

PR #4584, opened 2018-03-26
Bump to ADAM 0.24.0.

fnothaft:
This reverts commit 9fe150c801addf77bd2e58dc97c6f8bf7b999837, and bumps to ADAM 0.24.0 which reverts the code that caused #4376.


droazen:
@tomwhite When you get a chance, could you please test this branch to check whether the performance regression you reported earlier is resolved?

PR #4608, opened 2018-03-28
Fixing calculation of GC content to be correct.

### LINKED ISSUE #4600: Funcotator - GC Content is not always correct
The GC Content annotation is not always correct.

This needs to be investigated, specifically for indels, but in general it should be reviewed.

jonn-smith:
Fixes #4600


LeeTL1220:
@jonn-smith feel free to merge...

Issue #2155, opened 2016-09-08
Request: Allow custom packages for Annotations in VariantAnnotatorEngine


magicDGS:

droazen:
Adding to annotation epic #3274

magicDGS:
It will be useful if this is added to the configuration system (#2368 and #3081).

PR #4611, opened 2018-03-29
Add new configuration entry for plugins

magicDGS:
Includes:

* Configuration for packages to search read filters
* Configuration for packages to search annotations

In addition, it changes the behavior of `VariantAnnotatorEngine` to use the annotation packages from the configuration, and mimic what the plugin is doing. This closes https://github.com/broadinstitute/gatk/issues/2155

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
@@ -5,6 +5,7 @@ import htsjdk.variant.variantcontext.*;
 import htsjdk.variant.vcf.*;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.broadinstitute.barclay.argparser.ClassFinder;
 import org.broadinstitute.barclay.argparser.CommandLineException;
 import org.broadinstitute.hellbender.cmdline.GATKPlugin.DefaultGATKVariantAnnotationArgumentCollection;
 import org.broadinstitute.hellbender.cmdline.GATKPlugin.GATKAnnotationArgumentCollection;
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
```diff
@@ -499,7 +510,15 @@ static AnnotationManager ofAllMinusExcluded(final List<String> annotationsToExcl
...
         private static List<GenotypeAnnotation> makeAllGenotypeAnnotations() {
-            return ClassUtils.makeInstancesOfSubclasses(GenotypeAnnotation.class, Annotation.class.getPackage());
+            // Get our configuration:
+            final GATKConfig config = ConfigFactory.getInstance().getGATKConfig();
+
+            final ClassFinder finder = new ClassFinder();
+            for ( final String codecPackage : config.annotation_packages() ) {
+                finder.find(codecPackage, GenotypeAnnotation.class);
+            }
+            return finder.getConcreteClasses().stream().map(c -> (GenotypeAnnotation) ClassUtils.makeInstanceOf(c))
```
lbergelson: Same question as above.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/VariantAnnotatorEngine.java
```diff
@@ -490,7 +493,15 @@ static AnnotationManager ofAllMinusExcluded(final List<String> annotationsToExcl
         }
 
         private static List<InfoFieldAnnotation> makeAllInfoFieldAnnotations() {
-            return ClassUtils.makeInstancesOfSubclasses(InfoFieldAnnotation.class, Annotation.class.getPackage());
+            // Get our configuration:
+            final GATKConfig config = ConfigFactory.getInstance().getGATKConfig();
+
+            final ClassFinder finder = new ClassFinder();
+            for ( final String codecPackage : config.annotation_packages() ) {
```
lbergelson: Why do this instead of calling `ClassUtils.makeInstanceOfSubclasses` on each package and merging the results?

cmnbroad: This code is all going to be removed soon by https://github.com/broadinstitute/gatk/pull/4674. It might be best to wait for that, since it will reduce what this PR needs to do.

magicDGS: How close is that PR to be accepted? I can divide this one into the `ReadFilter` config entry and wait for the annotator to be accepted to include the entry (or include it directly in #4674). What do you prefer?

cmnbroad: I'd prefer not to add anything else to #4674 (and also to not split this one up). I would just wait until its merged (@droazen still needs to review it) and then rebase this on top of that.

lbergelson:
Closing because I've taken this over as https://github.com/broadinstitute/gatk/pull/5573

Issue #4572, opened 2018-03-23
CombineGVCFs incorrectly parses last GVCF block consisting of HLA contig


sooheelee:
Data is sensitive and bug is recapitulated in https://github.com/broadinstitute/dsde-docs/issues/3026.

CombineGVCFs gives the following error message:
```
java.lang.IllegalArgumentException: Invalid interval. Contig:HLA-DRB1*15:03:01:02 start:11569 end:11005
	at org.broadinstitute.hellbender.utils.Utils.validateArg(Utils.java:687)
	at org.broadinstitute.hellbender.utils.SimpleInterval.validatePositions(SimpleInterval.java:61)
	at org.broadinstitute.hellbender.utils.SimpleInterval.<init>(SimpleInterval.java:37)
	at org.broadinstitute.hellbender.tools.walkers.CombineGVCFs.onTraversalSuccess(CombineGVCFs.java:415)
	at org.broadinstitute.hellbender.engine.GATKTool.doWork(GATKTool.java:895)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:135)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:180)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:199)
	at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:159)
	at org.broadinstitute.hellbender.Main.mainEntry(Main.java:202)
	at org.broadinstitute.hellbender.Main.main(Main.java:288)
```
Here are the dictionary lines for two consecutive HLA-DRB1 contigs:
```
@SQ     SN:HLA-DRB1*15:03:01:02 LN:11569        M5:4e0d459b9bd15bff8645de84334e3d25     AS:38   UR:/seq/references/Homo_sapiens_assembly38/v0/Homo_sapiens_assembly38.fasta     SP:Homo sapiens
@SQ     SN:HLA-DRB1*16:02:01    LN:11005        M5:4a972df76bd3ee2857b87bd5be5ea00a     AS:38   UR:/seq/references/Homo_sapiens_assembly38/v0/Homo_sapiens_assembly38.fasta     SP:Homo sapiens
```
Notice the `LN` lengths match up. It appears that our tool is mistaking contig information.

sooheelee:
I will assign to @droazen to delegate.

[...]

PR #4615, opened 2018-03-30
Handle trailing reference blocks in CombineGVCFs.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/4572.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFs.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFs.java
@@ -412,14 +412,17 @@ public final class CombineGVCFs extends MultiVariantWalkerGroupedOnStart {
             return null;
         }
 
-        SimpleInterval interval = prevPos != null ? new SimpleInterval(prevPos.getContig(), prevPos.getStart(), variantContextsOverlappingCurrentMerge.stream().map(VariantContext::getEnd).max(Comparator.naturalOrder()).get()) :
-                storedReferenceContext.getInterval();
-
-        createIntermediateVariants(interval);
-
-        // there shouldn't be any state left unless the user cut in the middle of a gVCF block
         if ( !variantContextsOverlappingCurrentMerge.isEmpty() ) {
-            logger.warn("You have asked for an interval that cuts in the middle of one or more gVCF blocks. Please note that this will cause you to lose records that don't end within your interval.");
+            // finish off the last blocks
+            final SimpleInterval lastInterval = new SimpleInterval(
+                    variantContextsOverlappingCurrentMerge.get(0).getContig(),
+                    variantContextsOverlappingCurrentMerge.get(0).getStart(),
+                    variantContextsOverlappingCurrentMerge.stream().map(VariantContext::getEnd).max(Comparator.naturalOrder()).get());
+                createIntermediateVariants(lastInterval);
+            // there shouldn't be any state left unless the user cut in the middle of a gVCF block
+            if ( !variantContextsOverlappingCurrentMerge.isEmpty() ) {
+                logger.warn("You have asked for an interval that cuts in the middle of one or more gVCF blocks. Please note that this will cause you to lose records that don't end within your interval.");
+            }
         }
 
         return null;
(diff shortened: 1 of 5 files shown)


droazen:
@jamesemery please review

src/main/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFs.java
```diff
@@ -412,14 +412,17 @@ public Object onTraversalSuccess() {
...
-        SimpleInterval interval = prevPos != null ? new SimpleInterval(prevPos.getContig(), prevPos.getStart(), variantContextsOverlappingCurrentMerge.stream().map(VariantContext::getEnd).max(Comparator.naturalOrder()).get()) :
-                storedReferenceContext.getInterval();
-
-        createIntermediateVariants(interval);
-
-        // there shouldn't be any state left unless the user cut in the middle of a gVCF block
         if ( !variantContextsOverlappingCurrentMerge.isEmpty() ) {
-            logger.warn("You have asked for an interval that cuts in the middle of one or more gVCF blocks. Please note that this will cause you to lose records that don't end within your interval.");
+            // finish off the last blocks
+            final SimpleInterval lastInterval = new SimpleInterval(
```
jamesemery: Hmmm... So this change is mostly agnostic and is definately important for the case where the assumption that the prevPos must be on the right contig (which wasn't checked before). I'm not convinced this works universally in the case where there is a gap before the last contig. So imagine there are still reads queued to be closed at prevPos, a gap in coverage, then a stack of variant contexts at the last position. It doesn't strike me that this should handle closing that last interval if it needs to generate its own stops (for example if BREAK_BANDS_LONG_NAME) is set then it may well generate breaks in the gap between prevpos and the last stack of VCs. I think we should prefer to use prevPos in these cases and only switch to variantContextsOverlappingCurrentMerge.get(0) in the case where prevpos is on the wrong contig/starts later than it should.

src/test/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFsIntegrationTest.java
```diff
@@ -74,7 +74,10 @@
                 //Testing allele-specific annotations
                 {new File[]{getTestFile("NA12878.AS.chr20snippet.g.vcf"), getTestFile("NA12892.AS.chr20snippet.g.vcf")}, getTestFile("testAlleleSpecificAnnotations.vcf"), Arrays.asList("-G", "Standard", "-G", "AS_Standard"), b37_reference_20_21},
                 //Testing allele-specific annotations missing AS_Standard Group
-                {new File[]{getTestFile("NA12878.AS.chr20snippet.g.vcf"), getTestFile("NA12892.AS.chr20snippet.g.vcf")}, getTestFile("testAlleleSpecificAnnotationsNoGroup.vcf"), Arrays.asList("-G", "Standard", "-G", "AS_Standard"), b37_reference_20_21},};
+                {new File[]{getTestFile("NA12878.AS.chr20snippet.g.vcf"), getTestFile("NA12892.AS.chr20snippet.g.vcf")}, getTestFile("testAlleleSpecificAnnotationsNoGroup.vcf"), Arrays.asList("-G", "Standard", "-G", "AS_Standard"), b37_reference_20_21},
+                //Test that trailing reference blocks are emitted with correct intervals
```
jamesemery: You could demonstrate the issue i'm talking about if you combined two files that are already banded in step with eachother and you set BREAK_BANDS_LONG_NAME to some short value. you would expect a gap at the end of the contig given your change.

src/main/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFs.java
cmnbroad: Well, having thought about this a bit more, I'm not sure that problem is new to this change. As far as I can tell, this implementation of the tool only emits coverage for territory that is covered by the inputs, and doesn't pad out territory beyond that. In onTraversalSuccess, if prevPos is on a different contig than the remaining variants to be merged, we won't close out (pad out) the previous contig, but I don't think the tool ever does that. We might want to fix that, but I think thats a separate/bigger issue ? Otherwise, the variants that are remaining are all guaranteed to be on the same contig (right ?)  Sorry if I'm missing something, but please push back if you think I am.

jamesemery: You know, you are right. I was thinking of the code being like a normal `apply()` call where you are holding onto something "next." Your fix looks good.

jamesemery:
Feel free to merge it then :+1:

Issue #4577, opened 2018-03-26
IOUtils writeTempResource/writeTempResourceFromPath should mark files for deletion


cmnbroad:
The methods currently leave this to the caller, but this can and should be done automatically.

PR #4616, opened 2018-03-30
Automatically schedule temporary resource files for delete on exit.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/4577.


src/main/java/org/broadinstitute/hellbender/utils/R/RScriptExecutor.java
```diff
@@ -81,20 +81,19 @@ public RScriptExecutorException getScriptException(final String message) {
     }
 
     public boolean exec() {
-        List<File> tempFiles = new ArrayList<>();
+        final List<File> tempDirs = new ArrayList<>();
         try {
-            File tempLibSourceDir  = IOUtils.tempDir("RlibSources.", "");
-            File tempLibInstallationDir = IOUtils.tempDir("Rlib.", "");
```
droazen: For consistency with `BaseTest.createTempDir()`, I think that `IOUtils.tempDir()` should register the temp dir for recursive deletion on exit via `IOUtils.deleteRecursivelyOnExit()`

src/main/java/org/broadinstitute/hellbender/utils/R/RScriptLibrary.java
```diff
@@ -22,19 +24,27 @@ public String getLibraryName() {
...
 
-    public File writeLibrary(File tempDir) {
-        File libraryFile = new File(tempDir, getLibraryName());
+    /**
+     * Retrieve the resource for this library and write it to a File in {@code targetDir}. The File is not
+     * automatically scheduled for deletion and must be cleaned up by the caller.
+     * @param targetDir target directory where the File should be written
+     * @return the newly created File containing the library
+     */
+    public File writeLibrary(final File targetDir) {
```
droazen: Can you have this method register its libraryFile for deletion on exit using createTempFile(), rename the method and adjust the javadoc appropriately? You'll need to add an overload of createTempFile() that takes a directory argument.

src/main/java/org/broadinstitute/hellbender/utils/python/PythonScriptExecutor.java
```diff
@@ -115,13 +115,9 @@ public boolean executeModule(final String moduleName, final List<String> pythonP
      */
     public boolean executeScript(final Resource scriptResource, final List<String> pythonProcessArgs, final List<String> scriptArgs) {
         Utils.nonNull(scriptResource, "script resource cannot be null");
+        // this File is automatically scheduled for deletion on exit
         final File tempResourceFile = IOUtils.writeTempResource(scriptResource);
-
-        try {
-            return executeScript(tempResourceFile.getAbsolutePath(), pythonProcessArgs, scriptArgs);
-        } finally {
-            FileUtils.deleteQuietly(tempResourceFile);
```
droazen: Might want to consider keeping this -- it doesn't hurt to delete early when possible, despite the delete on exits.

src/main/java/org/broadinstitute/hellbender/utils/R/RScriptExecutor.java
cmnbroad: Done.

PR #4622, opened 2018-04-02
use latest version of fermilite jni

tedsharpe:


mwalker174:
It looks like there's some minor refactoring in your new graph handler. I'm not a real stickler about sneaking that in but just want to check it was intentional.

tedsharpe:
Thanks, Mark.  Yes, intentional.  We moved a couple of methods that we more naturally methods on the assembled objects rather than statics in client code at Chris' suggestion.  Part of the API upgrade, from my point of view.

PR #4623, opened 2018-04-03
More fix enum hashCode 

SHuang-Broad:
in `BreakpointComplications`

After this fix, the duplicated records disappear.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/BreakpointComplications.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/BreakpointComplications.java
@@ -377,8 +377,12 @@ public abstract class BreakpointComplications {
             result = 31 * result + dupSeqRepeatUnitRefSpan.hashCode();
             result = 31 * result + dupSeqRepeatNumOnRef;
             result = 31 * result + dupSeqRepeatNumOnCtg;
-            result = 31 * result + dupSeqStrandOnRef.hashCode();
-            result = 31 * result + dupSeqStrandOnCtg.hashCode();
+            for (final Strand strand : dupSeqStrandOnRef) {
+                result = 31 * result + strand.ordinal();
+            }
+            for (final Strand strand : dupSeqStrandOnCtg) {
+                result = 31 * result + strand.ordinal();
+            }
             return result;
         }
     }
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AlignedContigGenerator.java
```diff
@@ -10,4 +34,213 @@
 public abstract class AlignedContigGenerator {
```
cwhelan: It seems like this class should really just be an interface? The two subclasses could be top-level classes that just implement the interface. It's a bit weird to have this be an abstract class when it really only defines the signature of one method.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
```diff
@@ -207,11 +204,15 @@ private SvDiscoveryInputData getSvDiscoveryInputData(final JavaSparkContext ctx,
         final String outputPrefixWithSampleName = variantsOutDir + (variantsOutDir.endsWith("/") ? "" : "/")
                                                     + SVUtils.getSampleId(headerForReads) + "_";
 
+        final JavaRDD<GATKRead> reloadedContigAlignments = new ReadsSparkSource(ctx, readArguments.getReadValidationStringency())
```
cwhelan: I don't see how moving the two classes to live in the abstract class would change anything, so maybe this is what actually resolved the behavior for you? Before, how was `getReads` not returning the raw reads in the original bam file?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/BreakpointComplications.java
```diff
@@ -1072,8 +1076,16 @@ public int hashCode() {
...
+            if (dupSeqStrandOnRef != null) {
+                for (final Strand strand : dupSeqStrandOnRef) {
+                    result = 31 * result + Objects.hashCode( strand.ordinal() );
+                }
+            }
+            if (dupSeqStrandOnCtg != null) {
+                for (final Strand strand : dupSeqStrandOnCtg) {
+                    result = 31 * result + Objects.hashCode( strand.ordinal() );
+                }
+            }
```
tedsharpe: This is the actual fix, I believe.  Chris and I missed the list of enums.  (A List's hashCode combines the hashCode's of its elements, and that's no good on Spark.)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/alignment/AlignedContigGenerator.java
SHuang-Broad: Agree. But in 2nd thought, this is not related to this fix, so I'll do the refactor and the move in another PR.
I'm trying to put the implementations in one place and see if I can have one concrete class, just to avoid the two code path that exist mostly for historical reasons.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/BreakpointComplications.java
```diff
@@ -377,8 +377,12 @@ public int hashCode() {
             result = 31 * result + dupSeqRepeatUnitRefSpan.hashCode();
             result = 31 * result + dupSeqRepeatNumOnRef;
             result = 31 * result + dupSeqRepeatNumOnCtg;
-            result = 31 * result + dupSeqStrandOnRef.hashCode();
-            result = 31 * result + dupSeqStrandOnCtg.hashCode();
+            for (final Strand strand : dupSeqStrandOnRef) {
+                result = 31 * result + Objects.hashCode( strand.ordinal() );
+            }
+            for (final Strand strand : dupSeqStrandOnCtg) {
+                result = 31 * result + Objects.hashCode( strand.ordinal() );
```
tedsharpe: Objects.hashCode(strand.ordinal()) is a very expensive way of calculating strand.ordinal().  Maybe just use strand.ordinal() here, and on lines 1081 and 1086 below.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/StructuralVariationDiscoveryPipelineSpark.java
SHuang-Broad: That `getReads()` call was not right, and luckily it was actually not used so I'm going to fix it in a followup cleanup and test coverage bump PR.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/BreakpointComplications.java
SHuang-Broad: Thank you both! It's been done now.

SHuang-Broad:
Verified to be running correctly and not generating duplicates anymore.

PR #4627, opened 2018-04-05
Use GATK Travis Service Account with new key.

cmnbroad:
If this passes, we can probably remove encrypted_29f3b7c4d8c3_key from the repo variables.


cmnbroad:
Closing. The issue was resolved by one of:

- adding all DataProc permissions to the GATK Travis Service Account
- adding the DataProc "worker" role to the default service account

Issue #4541, opened 2018-03-19
Implement a way to specify combine operations per-field in GenomicsDB


droazen:

droazen:
@kgururaj Is this one still on your radar? It's mentioned in https://github.com/broadinstitute/gatk/pull/4645 as being targeted for the "next version" of the protobuf API, and @ldgauthier mentioned it to me as a feature she'd really love to have.

kgururaj:
Yep, next version

droazen:
This is done -- closing.

Issue #3689, opened 2017-10-11
Expose vid_mapping_file JSON in GenomicsDB


ldgauthier:
In order to produce the 68K WGS callset, we need to add new INFO annotations (relevant combine operations for these already exist.)  However, there's no easy way to specify how to combine them without manually editing that file. It would be great if there was a way to do that in the GATK tool args for GenomicsDBImport.

cmnbroad:
Per discussion with @kgururaj, the proposal is to have GenomicsDB expose/document the existing protocol buffers already used internally, along with protobuf-based constructors for both GenomicsDBImporter and GenomicsDBFeatureReader.

droazen:
This is a dupe of https://github.com/broadinstitute/gatk/issues/4541, I believe -- closing.

ldgauthier:
Agreed.  #4541 is a better description of the desired behavior.

Issue #3688, opened 2017-10-11
"sites-only" query for GenomicsDB


ldgauthier:
In order for the variant calling pipeline to be able to scale to process the 68K genomes in the next version of gnomAD and beyond, we need to limit the amount of genotype data we localize.  For the filtering portion of the pipeline we could greatly improve performance using a "sites-only" query from GenomicsDB.  The result could be in BCF format (as I believe it is now) and should include the first 8 fields of the VCF line (CHROM, POS, ID, REF, ALT, QUAL, FILTER, INFO) but should not return format-level/genotype data.

ldgauthier:
@kgururaj Hopefully this will be relatively easy.  We're going to need this for 68K callset we promised in the next quarter.

cmnbroad:
Per discussion with @kgururaj, this will probably take the form of a flag that suppresses materializing the genotype data. Also see the protobuf-based enhancements described [here](https://github.com/broadinstitute/gatk/issues/3689).

ldgauthier:
I did a really rough benchmarking of the new mode on my local machine.  There were ~80,000 samples and ~40,000 variants.  Master finished in 139.6 minutes and the new branch finished in 44.74.  Significant speedup!  At a glance the variants and INFO field attributes look like they match, but I'm running a more thorough comparison now.

There's something weird about the logging though.  Master output a progress update every 1000 variants, but the new branch only had an update at 1000 and then the total stats when it finished:
```
 16:19:16.342 INFO  ProgressMeter -        Current Locus  Elapsed Minutes    Variants Processed  Variants/Minute
 Chromosome chr20 position 1279618 (TileDB column 2714308521) has too many alleles in the combined VCF record : 61 : current limit : 50. Fields, such as  PL, with length equal to the number of genotypes will NOT be added for this location.
 17:03:54.604 INFO  ProgressMeter -        chr20:1269692             44.6                  1000             22.4
 GENOMICSDB_TIMER,GenomicsDB iterator next() timer,Wall-clock time(s),0.36128615400000264,Cpu time(s),0.3515559999999785
 17:03:56.445 INFO  ProgressMeter -        chr20:1308850             44.7                 40156            899.0
 17:03:56.445 INFO  ProgressMeter - Traversal complete. Processed 40156 total variants in 44.7 minutes.
```

ldgauthier:
[...]

Issue #3269, opened 2017-07-13
Add multi-interval support to GenomicsDBImport


droazen:
`GenomicsDBImport` is currently designed for scattering on a cluster, and so accepts only a single interval per invocation. To make it easier to run the tool locally as a replacement for `CombineGVCFs`, we should add the ability to pass in multiple intervals at once.

droazen:
For @kgururaj and @kdatta

ptranvan:
Multi intervals working now ?

droazen:
Not yet @ptranvan. @kgururaj and his team have a patch coming soon to add multi-interval support to this tool, so there's a good chance it will make it into a GATK release this month.

ajshultz:
Hello,
Any update to the timeline here? It looks like there is only some lasts tests to be run? I am eagerly awaiting multi-interval support for a project I am working on. Thank you for all of your hard work!

droazen:
Should not be much longer @ajshultz -- https://github.com/broadinstitute/gatk/pull/4645 just needs a final review pass, so it's likely to make the next GATK release.

droazen:
Implemented in https://github.com/broadinstitute/gatk/pull/4645

PR #4645, opened 2018-04-10
GenomicsDBImport: Modifications in order to address GATK issue #3269

francares:
This PR addresses required changes in order to use latest version of GenomicsDB which exposes new functionality such as:
- [Multi interval import and query support](https://github.com/broadinstitute/gatk/issues/3269):
  - We create multiple arrays (directories) in a single workspace - one per interval. So, if you wish to import intervals ("chr1", [ 1, 100M ]) and ("chr2", [ 1, 100M ]), you end up with 2 directories/arrays in the workspace with names chr1$1$100M and chr2$1$100M. The array names depend on the     partition bounds.
  - During the read phase, the user only supplies the workspace. The array names are obtained by scanning the entries in the workspace and reading the right arrays. For example, if you wish to read ("chr2", [ 50, 50M] ), then only the second array is queried.
  - In the previous version of the tool, the array name was a constant - _genomicsdb_array_. The new version will be backward compatible with respect to reads. Hence, if a directory named _genomicsdb_array_ is found in the workspace directory, it's passed as the array for the _GenomicsDBFeatureReader_ otherwise the array names are generated from the directory entry names.
  - Parallel import based on chromosome intervals. The number of threads to use can be specified as an integer argument to the [executeImport call](https://github.com/francares/gatk/blob/fmc_GenomicsDB_parallel_import/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java#L535). If no argument is specified, the number of threads is determined by Java's ForkJoinPool (typically equal to the \#cores in the system). 
    - The max number of intervals to import in parallel can be controlled by the command line argument --max-num-intervals-to-import-in-parallel (default 1)
    - Note that increasing parallelism increases the number of FeatureReaders opened to feed data to the importer. So, if you are using _N_ threads and your batch size is _B_, you will have _N*B_ feature readers open.
- Protobuf based API for import and read
  - [Import](https://github.com/francares/gatk/blob/fmc_GenomicsDB_parallel_import/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java#L522)
  - [Read](https://github.com/francares/gatk/blob/fmc_GenomicsDB_parallel_import/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java#L405)
    - #3688 
    - Option to produce GT field
    - [Option to produce GT for spanning deletion based on min PL value](https://github.com/Intel-HLS/GenomicsDB/issues/161)
    - https://github.com/broadinstitute/gatk/issues/2687
    - Doesn't support #4541 or #3689 yet - next version
- Bug fixes
  - #4716 
  - More error messages

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.hellbender.engine;
 
-import com.intel.genomicsdb.GenomicsDBFeatureReader;
+import com.intel.genomicsdb.model.GenomicsDBExportConfiguration;
+import com.intel.genomicsdb.reader.GenomicsDBFeatureReader;
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.tribble.*;
 import htsjdk.variant.bcf2.BCF2Codec;
@@ -23,8 +24,10 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.channels.SeekableByteChannel;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Function;
 
 /**
@@ -378,21 +381,30 @@ public final class FeatureDataSource<T extends Feature> implements GATKDataSourc
             IOUtils.canReadFile(callsetJson);
             IOUtils.canReadFile(vidmapJson);
             IOUtils.canReadFile(vcfHeader);
-        }
-        catch ( UserException.CouldNotReadInputFile e ) {
+        } catch ( UserException.CouldNotReadInputFile e ) {
             throw new UserException("Couldn't connect to GenomicsDB because the vidmap, callset JSON files, or gVCF Header (" +
                     GenomicsDBConstants.DEFAULT_VIDMAP_FILE_NAME + "," + GenomicsDBConstants.DEFAULT_CALLSETMAP_FILE_NAME + "," +
                     GenomicsDBConstants.DEFAULT_VCFHEADER_FILE_NAME + ") could not be read from GenomicsDB workspace " + workspace.getAbsolutePath(), e);
         }
 
+        GenomicsDBExportConfiguration.ExportConfiguration.Builder exportConfigurationBuilder =
+                GenomicsDBExportConfiguration.ExportConfiguration.newBuilder()
+                        .setWorkspace(workspace.getAbsolutePath())
+                        .setReferenceGenome(reference.getAbsolutePath())
+                        .setVidMappingFile(vidmapJson.getAbsolutePath())
+                        .setCallsetMappingFile(callsetJson.getAbsolutePath())
+                        .setVcfHeaderFilename(vcfHeader.getAbsolutePath());
+        File arrayFolder = new File(Paths.get(workspace.getAbsolutePath(), GenomicsDBConstants.DEFAULT_ARRAY_NAME)
+                .toAbsolutePath().toString());
+
+        if (arrayFolder.exists()) {
+            exportConfigurationBuilder.setArrayName(GenomicsDBConstants.DEFAULT_ARRAY_NAME);
+        } else {
+            exportConfigurationBuilder.setGenerateArrayNameFromPartitionBounds(true);
+        }
+
         try {
-            return new GenomicsDBFeatureReader<>(vidmapJson.getAbsolutePath(),
-                                                 callsetJson.getAbsolutePath(),
-                                                 workspace.getAbsolutePath(),
-                                                 GenomicsDBConstants.DEFAULT_ARRAY_NAME,
-                                                 reference.getAbsolutePath(),
-                                                 vcfHeader.getAbsolutePath(),
-                                                 new BCF2Codec());
+            return new GenomicsDBFeatureReader<>(exportConfigurationBuilder.build(), new BCF2Codec(), Optional.empty());
         } catch (final IOException e) {
             throw new UserException("Couldn't create GenomicsDBFeatureReader", e);
         }
--- a/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
@@ -24,6 +24,7 @@ import org.broadinstitute.hellbender.utils.nio.SeekableByteChannelPrefetcher;
 import java.io.File;
 import java.io.IOException;
 import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Iterator;
@@ -412,23 +413,22 @@ public final class FeatureDataSource<T extends Feature> implements GATKDataSourc
                         .setProduceGTWithMinPLValueForSpanningDeletions(false)
                         .setSitesOnlyQuery(false)
                         .setMaxDiploidAltAllelesThatCanBeGenotyped(GenotypeLikelihoods.MAX_DIPLOID_ALT_ALLELES_THAT_CAN_BE_GENOTYPED);
-        File arrayFolder = new File(Paths.get(workspace.getAbsolutePath(), GenomicsDBConstants.DEFAULT_ARRAY_NAME)
-                .toAbsolutePath().toString());
+        Path arrayFolder = Paths.get(workspace.getAbsolutePath(), GenomicsDBConstants.DEFAULT_ARRAY_NAME).toAbsolutePath();
 
-        //For the multi-interval support, we create multiple arrays (directories) in a single workspace -
-        //one per interval. So, if you wish to import intervals ("chr1", [ 1, 100M ]) and ("chr2", [ 1, 100M ]),
-        //you end up with 2 directories named chr1$1$100M and chr2$1$100M. So, the array names depend on the
-        //partition bounds.
+        // For the multi-interval support, we create multiple arrays (directories) in a single workspace -
+        // one per interval. So, if you wish to import intervals ("chr1", [ 1, 100M ]) and ("chr2", [ 1, 100M ]),
+        // you end up with 2 directories named chr1$1$100M and chr2$1$100M. So, the array names depend on the
+        // partition bounds.
 
-        //During the read phase, the user only supplies the workspace. The array names are obtained by scanning
-        //the entries in the workspace and reading the right arrays. For example, if you wish to read ("chr2",
-        //50, 50M), then only the second array is queried.
+        // During the read phase, the user only supplies the workspace. The array names are obtained by scanning
+        // the entries in the workspace and reading the right arrays. For example, if you wish to read ("chr2",
+        // 50, 50M), then only the second array is queried.
 
-        //In the previous version of the tool, the array name was a constant - genomicsdb_array. The new version
-        //will be backward compatible with respect to reads. Hence, if a directory named genomicsdb_array is found,
-        //the array name is passed to the GenomicsDBFeatureReader otherwise the array names are generated from the
-        //directory entries.
-        if (arrayFolder.exists()) {
+        // In the previous version of the tool, the array name was a constant - genomicsdb_array. The new version
+        // will be backward compatible with respect to reads. Hence, if a directory named genomicsdb_array is found,
+        // the array name is passed to the GenomicsDBFeatureReader otherwise the array names are generated from the
+        // directory entries.
+        if (Files.exists(arrayFolder)) {
             exportConfigurationBuilder.setArrayName(GenomicsDBConstants.DEFAULT_ARRAY_NAME);
         } else {
             exportConfigurationBuilder.setGenerateArrayNameFromPartitionBounds(true);
--- a/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
@@ -406,10 +406,26 @@ public final class FeatureDataSource<T extends Feature> implements GATKDataSourc
                         .setReferenceGenome(reference.getAbsolutePath())
                         .setVidMappingFile(vidmapJson.getAbsolutePath())
                         .setCallsetMappingFile(callsetJson.getAbsolutePath())
-                        .setVcfHeaderFilename(vcfHeader.getAbsolutePath());
+                        .setVcfHeaderFilename(vcfHeader.getAbsolutePath())
+                        .setProduceGTField(false)
+                        .setProduceGTWithMinPLValueForSpanningDeletions(false)
+                        .setSitesOnlyQuery(false);
         File arrayFolder = new File(Paths.get(workspace.getAbsolutePath(), GenomicsDBConstants.DEFAULT_ARRAY_NAME)
                 .toAbsolutePath().toString());
 
+        //For the multi-interval support, we create multiple arrays (directories) in a single workspace -
+        //one per interval. So, if you wish to import intervals ("chr1", [ 1, 100M ]) and ("chr2", [ 1, 100M ]),
+        //you end up with 2 directories named chr1$1$100M and chr2$1$100M. So, the array names depend on the
+        //partition bounds.
+
+        //During the read phase, the user only supplies the workspace. The array names are obtained by scanning
+        //the entries in the workspace and reading the right arrays. For example, if you wish to read ("chr2",
+        //50, 50M), then only the second array is queried.
+
+        //In the previous version of the tool, the array name was a constant - genomicsdb_array. The new version
+        //will be backward compatible with respect to reads. Hence, if a directory named genomicsdb_array is found,
+        //the array name is passed to the GenomicsDBFeatureReader otherwise the array names are generated from the
+        //directory entries.
         if (arrayFolder.exists()) {
             exportConfigurationBuilder.setArrayName(GenomicsDBConstants.DEFAULT_ARRAY_NAME);
         } else {
--- a/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
@@ -6,6 +6,7 @@ import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.tribble.*;
 import htsjdk.variant.bcf2.BCF2Codec;
 import htsjdk.variant.variantcontext.VariantContext;
+import htsjdk.variant.variantcontext.GenotypeLikelihoods;
 import htsjdk.variant.vcf.VCFHeader;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -409,7 +410,8 @@ public final class FeatureDataSource<T extends Feature> implements GATKDataSourc
                         .setVcfHeaderFilename(vcfHeader.getAbsolutePath())
                         .setProduceGTField(false)
                         .setProduceGTWithMinPLValueForSpanningDeletions(false)
-                        .setSitesOnlyQuery(false);
+                        .setSitesOnlyQuery(false)
+                        .setMaxDiploidAltAllelesThatCanBeGenotyped(GenotypeLikelihoods.MAX_DIPLOID_ALT_ALLELES_THAT_CAN_BE_GENOTYPED);
         File arrayFolder = new File(Paths.get(workspace.getAbsolutePath(), GenomicsDBConstants.DEFAULT_ARRAY_NAME)
                 .toAbsolutePath().toString());
 
--- a/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
@@ -387,6 +387,19 @@ public final class FeatureDataSource<T extends Feature> implements GATKDataSourc
                     GenomicsDBConstants.DEFAULT_VCFHEADER_FILE_NAME + ") could not be read from GenomicsDB workspace " + workspace.getAbsolutePath(), e);
         }
 
+        final GenomicsDBExportConfiguration.ExportConfiguration exportConfigurationBuilder =
+                createExportConfiguration(reference, workspace, callsetJson, vidmapJson, vcfHeader);
+
+        try {
+            return new GenomicsDBFeatureReader<>(exportConfigurationBuilder, new BCF2Codec(), Optional.empty());
+        } catch (final IOException e) {
+            throw new UserException("Couldn't create GenomicsDBFeatureReader", e);
+        }
+    }
+
+    private static GenomicsDBExportConfiguration.ExportConfiguration createExportConfiguration(final File reference, final File workspace,
+                                                                                               final File callsetJson, final File vidmapJson,
+                                                                                               final File vcfHeader) {
         GenomicsDBExportConfiguration.ExportConfiguration.Builder exportConfigurationBuilder =
                 GenomicsDBExportConfiguration.ExportConfiguration.newBuilder()
                         .setWorkspace(workspace.getAbsolutePath())
@@ -403,11 +416,7 @@ public final class FeatureDataSource<T extends Feature> implements GATKDataSourc
             exportConfigurationBuilder.setGenerateArrayNameFromPartitionBounds(true);
         }
 
-        try {
-            return new GenomicsDBFeatureReader<>(exportConfigurationBuilder.build(), new BCF2Codec(), Optional.empty());
-        } catch (final IOException e) {
-            throw new UserException("Couldn't create GenomicsDBFeatureReader", e);
-        }
+        return exportConfigurationBuilder.build();
     }
 
     /**
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -486,7 +486,7 @@ public final class GenomicsDBImport extends GATKTool {
 
     private Map<String, FeatureReader<VariantContext>> createSampleToReaderMap(
             final Map<String, Path> sampleNameToVcfPath, final int batchSize, final int index) {
-        //TODO: fix casting since it's really ugly
+        // TODO: fix casting since it's really ugly
         return inputPreloadExecutorService != null ?
                 getFeatureReadersInParallel((SortedMap<String, Path>) sampleNameToVcfPath, batchSize, index)
                 : getFeatureReadersSerially(sampleNameToVcfPath, batchSize, index);
@@ -548,7 +548,6 @@ public final class GenomicsDBImport extends GATKTool {
 
         final int sampleCount = sampleNameToVcfPath.size();
         final int updatedBatchSize = (batchSize == DEFAULT_ZERO_BATCH_SIZE) ? sampleCount : batchSize;
-//        final int totalBatchCount = (sampleCount/updatedBatchSize) + (sampleCount%updatedBatchSize==0 ? 0 : 1);
         final ImportConfig importConfig = createImportConfig(updatedBatchSize);
 
         GenomicsDBImporter importer;
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -113,8 +113,7 @@ import java.util.stream.Collectors;
  * <h3>Caveats</h3>
  * <ul>
  *     <li>IMPORTANT: The -Xmx value the tool is run with should be less than the total amount of physical memory available by at least a few GB, as the native TileDB library requires additional memory on top of the Java memory. Failure to leave enough memory for the native code can result in confusing error messages!</li>
- *     <li>A single interval must be provided. This means each import is limited to a maximum of one contig</li>
- *     <li>Currently, only supports diploid data</li>
+ *     <li>At least one interval must be provided</li>
  *     <li>Input GVCFs cannot contain multiple entries for a single genomic position</li>
  *     <li>The --genomicsdb-workspace-path must point to a non-existent or empty directory.</li>
  * </ul>
@@ -479,8 +478,8 @@ public final class GenomicsDBImport extends GATKTool {
                 : getFeatureReadersSerially(sampleNameToVcfPath, batchSize, index);
     }
 
-    private List<GenomicsDBImportConfiguration.Partition> generatePartitionListFromIntervals(List<ChromosomeInterval> chromosomeIntevals) {
-        return chromosomeIntevals.stream().map(interval -> {
+    private List<GenomicsDBImportConfiguration.Partition> generatePartitionListFromIntervals(List<ChromosomeInterval> chromosomeIntervals) {
+        return chromosomeIntervals.stream().map(interval -> {
             GenomicsDBImportConfiguration.Partition.Builder partitionBuilder = GenomicsDBImportConfiguration.Partition.newBuilder();
             Coordinates.ContigPosition.Builder contigPositionBuilder = Coordinates.ContigPosition.newBuilder();
             Coordinates.GenomicsDBColumn.Builder columnBuilder = Coordinates.GenomicsDBColumn.newBuilder();
@@ -529,11 +528,7 @@ public final class GenomicsDBImport extends GATKTool {
         final int updatedBatchSize = (batchSize == DEFAULT_ZERO_BATCH_SIZE) ? sampleCount : batchSize;
 //        final int totalBatchCount = (sampleCount/updatedBatchSize) + (sampleCount%updatedBatchSize==0 ? 0 : 1);
         final ImportConfig importConfig = createImportConfig(updatedBatchSize);
-        //TODO: Fix this.
-//        BiConsumer<Map<String, FeatureReader<VariantContext>>, Integer> closeReaders = (readers, batchCount) -> {
-//            progressMeter.update(intervals.get(0));
-//            logger.info("Done importing batch " + batchCount + "/" + totalBatchCount);
-//        };
+
         GenomicsDBImporter importer;
         try {
             importer = new GenomicsDBImporter(importConfig);
--- a/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
@@ -1,7 +1,12 @@
 package org.broadinstitute.hellbender.tools.genomicsdb;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
-import com.intel.genomicsdb.*;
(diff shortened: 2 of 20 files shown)


kgururaj:
Supersedes https://github.com/broadinstitute/gatk/pull/4496
* https://github.com/broadinstitute/gatk/issues/3688
* https://github.com/Intel-HLS/GenomicsDB/issues/161

kgururaj:
Fix for https://github.com/broadinstitute/gatk/issues/4716 - @cwhelan please use this branch for testing. You should be able to re-use the pre-created array - the bug was in the query part only.

cwhelan:
@kgururaj Thanks, this branch passes the test case that was failing for me before.

ldgauthier:
Thanks for the quick turnaround @kgururaj !

droazen:
There is currently one failing integration test:

```
completedicsdb.GenomicsDBImportIntegrationTest.testGenomicsDBImportFileInputsAgainstCombineGVCFWithMultipleIntervals FAILED
    java.lang.AssertionError: actual is longer than expected with at least one additional element: [VC null @ chr20:17970000 Q. of type=SYMBOLIC alleles=[G*, <NON_REF>] attr={} GT=htsjdk.variant.bcf2.BCF2Codec$LazyData@134b79ab filters=
        at org.testng.Assert.fail(Assert.java:93)
        at org.broadinstitute.hellbender.utils.test.BaseTest.assertCondition(BaseTest.java:395)
        at org.broadinstitute.hellbender.tools.genomicsdb.GenomicsDBImportIntegrationTest.lambda$checkGenomicsDBAgainstExpected$8(GenomicsDBImportIntegrationTest.java:326)
        at java.util.ArrayList.forEach(ArrayList.java:1249)
        at org.broadinstitute.hellbender.tools.genomicsdb.GenomicsDBImportIntegrationTest.checkGenomicsDBAgainstExpected(GenomicsDBImportIntegrationTest.java:319)
        at org.broadinstitute.hellbender.tools.genomicsdb.GenomicsDBImportIntegrationTest.testGenomicsDBAgainstCombineGVCFs(GenomicsDBImportIntegrationTest.java:166)
        at org.broadinstitute.hellbender.tools.genomicsdb.GenomicsDBImportIntegrationTest.testGenomicsDBImportFileInputsAgainstCombineGVCFWithMultipleIntervals(GenomicsDBImportIntegrationTest.java:107)
```

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
```diff
@@ -378,26 +381,44 @@ public static boolean isGenomicsDBPath(final String path) {
...
+                        .setVidMappingFile(vidmapJson.getAbsolutePath())
+                        .setCallsetMappingFile(callsetJson.getAbsolutePath())
+                        .setVcfHeaderFilename(vcfHeader.getAbsolutePath());
+        File arrayFolder = new File(Paths.get(workspace.getAbsolutePath(), GenomicsDBConstants.DEFAULT_ARRAY_NAME)
+                .toAbsolutePath().toString());
+
+        if (arrayFolder.exists()) {
+            exportConfigurationBuilder.setArrayName(GenomicsDBConstants.DEFAULT_ARRAY_NAME);
+        } else {
+            exportConfigurationBuilder.setGenerateArrayNameFromPartitionBounds(true);
```
droazen: Can you add a comment explaining what this `setGenerateArrayNameFromPartitionBounds(true)` call does?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -468,6 +471,51 @@ private void initializeInputPreloadExecutorService() {
...
 
+    private Map<String, FeatureReader<VariantContext>> createSampleToReaderMap(
+            final Map<String, Path> sampleNameToVcfPath, final int batchSize, final int index) {
+        //TODO: fix casting since it's really ugly
+        return inputPreloadExecutorService != null ?
+                getFeatureReadersInParallel((SortedMap<String, Path>) sampleNameToVcfPath, batchSize, index)
+                : getFeatureReadersSerially(sampleNameToVcfPath, batchSize, index);
+    }
+
+    private List<GenomicsDBImportConfiguration.Partition> generatePartitionListFromIntervals(List<ChromosomeInterval> chromosomeIntevals) {
```
lbergelson: typo: intevals

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -479,75 +527,33 @@ public void traverse() {
...
-                importer.importBatch();
-            } catch (final IOException e) {
-                throw new UserException("GenomicsDB import failed in batch " + batchCount, e);
-            }
-            closeReaders(sampleToReaderMap);
-            progressMeter.update(intervals.get(0));
-            logger.info("Done importing batch " + batchCount + "/" + totalBatchCount);
+        try {
+            importer = new GenomicsDBImporter(importConfig);
+            importer.executeImport();
```
droazen: Is this batching done internally in `executeImport()` now? If so, are there regular status messages emitted to the logger before/after each batch?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -479,75 +527,33 @@ public void traverse() {
...
         final int sampleCount = sampleNameToVcfPath.size();
         final int updatedBatchSize = (batchSize == DEFAULT_ZERO_BATCH_SIZE) ? sampleCount : batchSize;
-        final int totalBatchCount = (sampleCount/updatedBatchSize) + (sampleCount%updatedBatchSize==0 ? 0 : 1);
-
+//        final int totalBatchCount = (sampleCount/updatedBatchSize) + (sampleCount%updatedBatchSize==0 ? 0 : 1);
+        final ImportConfig importConfig = createImportConfig(updatedBatchSize);
+        //TODO: Fix this.
+//        BiConsumer<Map<String, FeatureReader<VariantContext>>, Integer> closeReaders = (readers, batchCount) -> {
+//            progressMeter.update(intervals.get(0));
+//            logger.info("Done importing batch " + batchCount + "/" + totalBatchCount);
```
droazen: Can you explain what this TODO is? Are the readers getting closed properly? (I notice that you deleted the `closeReaders()` method below...)

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -479,75 +527,33 @@ public void traverse() {
...
-        } catch (final FileNotFoundException fe) {
-            throw new UserException("Unable to write VCF Header file " + vcfHeaderFile.getAbsolutePath(), fe);
-        }
-
-
-        if (doConsolidation) {
-            logger.info("GenomicsDB consolidation started");
-            GenomicsDBImporter.consolidateTileDBArray(workspace, GenomicsDBConstants.DEFAULT_ARRAY_NAME);
-            logger.info("GenomicsDB consolidation completed");
-        }
```
droazen: I'm assuming that consolidation and writing the JSONs are handled internally now?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -115,7 +120,7 @@
  * </ul>
```
droazen: You need to update the tool documentation to reflect the fact that multiple intervals are now supported. Also included any extra information the user might need to know (eg., are the intervals required to be contiguous?)

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -52,16 +51,23 @@
     private static final List<String> LOCAL_GVCFS = Arrays.asList(HG_00096, HG_00268, NA_19625);
     private static final String GENOMICSDB_TEST_DIR = toolsTestDir + "GenomicsDBImport/";
     private static final String COMBINEGVCFS_TEST_DIR = toolsTestDir + "walkers/CombineGVCFs/";
-
     private static final String COMBINED = largeFileTestDir + "gvcfs/combined.gatk3.7.g.vcf.gz";
+    private static final String COMBINED_MULTI_INTERVAL = largeFileTestDir + "gvcfs/combined_multi_interval.gatk3.7.g.vcf.gz";
```
droazen: Include a comment describing exactly how this file was created (including the command line used).

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -52,16 +51,23 @@
...
-    private static final SimpleInterval INTERVAL = new SimpleInterval("chr20", 17960187, 17981445);
-    private static final SimpleInterval INTERVAL_3736 = new SimpleInterval("chr6",130365070,146544250);
-    private static final SimpleInterval INTERVAL_NONDIPLOID = new SimpleInterval("20", 10000000, 10100000);
-    private static final SimpleInterval SMALLER_INTERVAL = new SimpleInterval("chr20", 17960187, 17961973);
+    private static final ArrayList<SimpleInterval> INTERVAL =
+            new ArrayList<SimpleInterval>(Arrays.asList(new SimpleInterval("chr20", 17960187, 17981445)));
+    private static final ArrayList<SimpleInterval> MULTIPLE_INTERVALS = new ArrayList<SimpleInterval>(Arrays.asList(
+        new SimpleInterval("chr20", 17960187, 17970000),
+        new SimpleInterval("chr20", 17970001, 17980000),
+        new SimpleInterval("chr20", 17980001, 17981445)
```
droazen: When specifying multiple intervals, are they required to be contiguous? If not, could you please add a separate test case involving non-contiguous intervals as well?

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -123,21 +146,25 @@ private File runCombineGVCFs(final List<String> inputs, final SimpleInterval int
...
         final String workspace = createTempDir("genomicsdb-tests-").getAbsolutePath() + "/workspace";
 
-        writeToGenomicsDB(vcfInputs, interval, workspace, 0, false, 0, 1);
+        writeToGenomicsDB(vcfInputs, intervals, workspace, 0, false, 0, 1);
         checkJSONFilesAreWritten(workspace);
-        File expectedCombinedVCF = runCombineGVCFs(vcfInputs, interval, referenceFile, CombineGVCFArgs);
-        checkGenomicsDBAgainstExpected(workspace, interval, expectedCombinedVCF.getAbsolutePath(), referenceFile, true);
+        for(SimpleInterval currInterval : intervals) {
+            List<SimpleInterval> tmpList = new ArrayList<SimpleInterval>(Arrays.asList(currInterval));
+            File expectedCombinedVCF = runCombineGVCFs(vcfInputs, tmpList, referenceFile, CombineGVCFArgs);
```
droazen: Shouldn't you be running `CombineGVCFs` with multiple intervals as well, instead of 1 interval at a time?

kgururaj:
@droazen  - that failure is related to the question I sent Louis and James - I'll forward you the same email

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
```diff
@@ -590,52 +634,23 @@ public void testPreserveContigOrderingInHeader() throws IOException {
...
+                .build();
+
+       return new GenomicsDBFeatureReader<>(exportConfiguration, new BCF2Codec(), Optional.empty());
     }
 
-    private static GenomicsDBFeatureReader<VariantContext, PositionalBufferedStream> getGenomicsDBFeatureReader(final String workspace, final String reference) throws IOException {
+    private static GenomicsDBFeatureReader<VariantContext, PositionalBufferedStream> getGenomicsDBFeatureReader(
+            final String workspace, final String reference) throws IOException {
         return getGenomicsDBFeatureReader(workspace, reference, false);
     }
```
droazen: I don't see a test case covering the issue reported in https://github.com/broadinstitute/gatk/issues/4716 -- can you add one?

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
droazen: Also add a third test case involving multiple intervals from different contigs.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: Yes

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
kgururaj: For the multi-interval support, we create multiple arrays (directories) in a single workspace - one per interval. So, if you wish to import intervals ("chr1", [ 1, 100M ])  and ("chr2", [ 1, 100M ]), you end up with 2 directories named chr1$1$100M and chr2$1$100M. So, the array names depend on the partition bounds.

During the read phase, the user only supplies the workspace. The array names are obtained by scanning the entries in the workspace and reading the right arrays. For example, if you wish to read ("chr2", 50, 50M), then only the second array is queried. 

In the previous version of the tool, the array name was a constant- genomicsdb_array. The new version will be backward compatible with respect to read. Hence, if a directory named genomicsdb_array is found the array name is passed to the GenomicsDBFeatureReader otherwise the array names are generated from the directory entries.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: Yes

lbergelson:
@ldgauthier Would it be possible to run this against 1 shard of the 20k and compare to the previous output to make sure nothing has gone weird?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -468,6 +471,51 @@ private void initializeInputPreloadExecutorService() {
         }
     }
 
+    private Map<String, FeatureReader<VariantContext>> createSampleToReaderMap(
+            final Map<String, Path> sampleNameToVcfPath, final int batchSize, final int index) {
+        //TODO: fix casting since it's really ugly
+        return inputPreloadExecutorService != null ?
```
lbergelson: In the past we found that the sorting here was really important or it changed the results.  If it's still sensitive to the ordering of the map maybe the signature of `ImportConfig` constructor that this is fed to should be modified to use `SortedMaps`

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: Spurious comment - yes, the readers are being closed inside GenomicsDBImporter

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: Fixed doc - intervals need not be contiguous

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: Yes. I think we have kept the same sort order based on the following:
* [The tool sorts by sample name first](https://github.com/francares/gatk/blob/cb46c8acba2a374a2217f27056de5292c4047e5b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java#L346)
* [Using the samples in the same order as provided in the sampleToVcfName TreeMap](https://github.com/francares/gatk/blob/cb46c8acba2a374a2217f27056de5292c4047e5b/src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java#L515)
* Passing the same TreeMap as argument to getFeatureReaders()

I did a couple of tests with arbitrary ordering of samples

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
kgururaj: Intervals need not be contiguous

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
kgururaj: Related to the earlier question about CombineGVCF

src/test/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImportIntegrationTest.java
kgururaj: Done

lbergelson:
@francares We encountered a problem when @ldgauthier tried running this branch on one of the shards of our previous 20k calling computation.  The current version of gatk is able to complete the shard in a given amount of memory (7.5G I think, but @ldgauthier would know for sure), but this version ran out of memory.  Has something changed that would make the tool require more memory at runtime now?  Have you seen increased memory usage after your changes?

I think I may know what's causing the issue.  It seems like you're probably starting multiple simultaneous batch imports which is causing us to go over our memory limits. 

See `GenomicsDBImporter` line 554:
```
        final ExecutorService executor = numThreads == 0 ? ForkJoinPool.commonPool() : Executors.newFixedThreadPool(numThreads);
```

In the case of 0 threads you use `ForkJoinPool.commonPool()`, which is the case GATK is using, since it calls the no args `executeImport`  As far as I can see, this uses some number of threads automatically derived from the system.  Each additional import batch consumes a ton of memory in this case because each VCFReader uses a lot of memory.  (That's a different problem, but it's our problem).  

@francares Does that seem plausible to you?  We should try rerunning our job with threads set to 1 which should restrict it to a single batch at a time.

kgururaj:
Are you testing with a single chromosome interval or multiple intervals? If a single interval, there should be no difference since a single thread will be used.

lbergelson:
It looks like it will run with some number of threads up to `numPartitions`.  Is num partitions always going to be 1 when there's only a single interval?

kgururaj:
Yep, each chromosome interval is handled by a single thread

francares:
@lbergelson sorry for my late response. I'm currently on vacations but I will try to respond (with some delay) any question.
So, what I'm seeing here (using Github web without having a proper dev env) is that for each interval, it's going to call (in parallel) sample reader function from the configuration =>
```
final Map<String, FeatureReader<VariantContext>> sampleToReaderMap =
                                this.config.sampleToReaderMapCreator().apply(
                                        this.config.getSampleNameToVcfPath(), updatedBatchSize, index); 
```
That's is the first difference from previous implementation. If whatever you have in that function consume lots of memory, that's an issue.
Regarding the thread pool, I'm not seeing it's being starved by chromosome parallel import but it might use extra memory to execute since there is a high load of threads use due to the number of parallel imports.
Worker threads can execute only one task at the time, but the ForkJoinPool doesn’t create a separate thread for every single subtask. Instead, each thread in the pool has its own double-ended queue (or deque, pronounced deck) **which stores tasks**.

Those are the two things I'm seeing right now without having the chance to debug :(.

lbergelson:
@francares Ah, sorry to bother you on your vacation.   We can address this when you get back.  The creation of the vcf readers is incredibly memory intensive because each one needs a large index.  The number of open readers is the limiting factor that determines how large a batch size we can use. 

It doesn't make sense in this case to use multiple importers in parallel if we're also restricting batch size.  The only reason to run with a batch size smaller than the number of samples is to get around memory restrictions, so running two simultaneous imports will just force a smaller batch size which is obviously sub-optimal.

droazen:
@francares The "batch size" argument should represent an absolute upper limit on the number of simultaneous readers open at once, as its only purpose it to control memory usage, as @lbergelson noted above.

kgururaj:
* Added a command line parameter to control the number of intervals to import in parallel - default 1
* This doesn't affect Chris' case since he was importing a single interval.
* I ran the old and new versions of GATK on 1000 samples with a batch size of 20. Verified that at most 20 readers are open at a given time (GNU/Linux open file descriptors). Memory consumption difference was less than 10%.
* @cwhelan was the import command run with a list of VCF files in the command line or with the --sample-name-map argument?

ldgauthier:
@kgururaj I ran with batch size 50, using a sample map, 5 reader threads:

`java -jar GATK_GDBfork.jar GenomicsDBImport --genomicsdb-workspace-path forkTest --batch-size 50 -L chr20:45840744-45870555 --sample-name-map gnarly_reblocked_all.sample_map --reader-threads 5`

That sample map has 80K genomes because that's the project I'm working on now.  Log was as follows:
```
16:27:48.831 INFO  NativeLibraryLoader - Loading libgkl_compression.so from jar:file:/humgen/gsa-hpprojects/dev/gauthier/reblockGVCF/GATK_GDBfork.jar!/com/intel/gkl/nati
ve/libgkl_compression.so
16:27:48.947 INFO  GenomicsDBImport - ------------------------------------------------------------
16:27:48.948 INFO  GenomicsDBImport - The Genome Analysis Toolkit (GATK) v4.0.3.0-24-g8804e16-SNAPSHOT
16:27:48.948 INFO  GenomicsDBImport - For support and documentation go to https://software.broadinstitute.org/gatk/
16:27:48.949 INFO  GenomicsDBImport - Executing as gauthier@gsa5.broadinstitute.org on Linux v2.6.32-642.15.1.el6.x86_64 amd64
16:27:48.949 INFO  GenomicsDBImport - Java runtime: Java HotSpot(TM) 64-Bit Server VM v1.8.0_121-b13
16:27:48.950 INFO  GenomicsDBImport - Start Date/Time: May 4, 2018 4:27:48 PM EDT
16:27:48.950 INFO  GenomicsDBImport - ------------------------------------------------------------
16:27:48.950 INFO  GenomicsDBImport - ------------------------------------------------------------
16:27:48.950 INFO  GenomicsDBImport - HTSJDK Version: 2.14.3
16:27:48.951 INFO  GenomicsDBImport - Picard Version: 2.18.1
16:27:48.951 INFO  GenomicsDBImport - HTSJDK Defaults.COMPRESSION_LEVEL : 2
16:27:48.951 INFO  GenomicsDBImport - HTSJDK Defaults.USE_ASYNC_IO_READ_FOR_SAMTOOLS : false
16:27:48.951 INFO  GenomicsDBImport - HTSJDK Defaults.USE_ASYNC_IO_WRITE_FOR_SAMTOOLS : true
16:27:48.951 INFO  GenomicsDBImport - HTSJDK Defaults.USE_ASYNC_IO_WRITE_FOR_TRIBBLE : false
16:27:48.951 INFO  GenomicsDBImport - Deflater: IntelDeflater
16:27:48.951 INFO  GenomicsDBImport - Inflater: IntelInflater
16:27:48.951 INFO  GenomicsDBImport - GCS max retries/reopens: 20
16:27:48.951 INFO  GenomicsDBImport - Using google-cloud-java patch 6d11bef1c81f885c26b2b56c8616b7a705171e4f from https://github.com/droazen/google-cloud-java/tree/dr_all_nio_fixes
16:27:48.951 INFO  GenomicsDBImport - Initializing engine
16:28:03.989 INFO  IntervalArgumentCollection - Processing 29812 bp from intervals
16:28:03.994 INFO  GenomicsDBImport - Done initializing engine
Created workspace /humgen/gsa-hpprojects/dev/gauthier/reblockGVCF/forkTest
16:28:04.155 INFO  GenomicsDBImport - Vid Map JSON file will be written to forkTest/vidmap.json
16:28:04.155 INFO  GenomicsDBImport - Callset Map JSON file will be written to forkTest/callset.json
16:28:04.156 INFO  GenomicsDBImport - Complete VCF Header will be written to forkTest/vcfheader.vcf
16:28:04.156 INFO  GenomicsDBImport - Importing to array - forkTest/genomicsdb_array
16:28:04.158 INFO  ProgressMeter - Starting traversal
16:28:04.158 INFO  ProgressMeter -        Current Locus  Elapsed Minutes     Batches Processed   Batches/Minute
16:28:05.198 INFO  GenomicsDBImport - Starting batch input file preload
16:29:23.571 INFO  GenomicsDBImport - Finished batch preload
16:48:46.140 INFO  GenomicsDBImport - Shutting down engine
[May 4, 2018 4:48:46 PM EDT] org.broadinstitute.hellbender.tools.genomicsdb.GenomicsDBImport done. Elapsed time: 20.96 minutes.
Runtime.totalMemory()=22281715712
java.util.concurrent.CompletionException: java.lang.OutOfMemoryError: Java heap space
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
        at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
        at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1592)
        at java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1582)
        at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
        at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
        at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
        at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
Caused by: java.lang.OutOfMemoryError: Java heap space
        at com.intel.genomicsdb.importer.SilentByteBufferStream.<init>(SilentByteBufferStream.java:55)
        at com.intel.genomicsdb.importer.GenomicsDBImporterStreamWrapper.<init>(GenomicsDBImporterStreamWrapper.java:70)
        at com.intel.genomicsdb.importer.GenomicsDBImporter.addBufferStream(GenomicsDBImporter.java:397)
        at com.intel.genomicsdb.importer.GenomicsDBImporter.addSortedVariantContextIterator(GenomicsDBImporter.java:358)
        at com.intel.genomicsdb.importer.GenomicsDBImporter.<init>(GenomicsDBImporter.java:167)
        at com.intel.genomicsdb.importer.GenomicsDBImporter.lambda$null$2(GenomicsDBImporter.java:598)
        at com.intel.genomicsdb.importer.GenomicsDBImporter$$Lambda$58/15335646.get(Unknown Source)
        at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1590)
        ... 5 more
```

lbergelson:
@ldgauthier You're running 80k?  Does that run using the current master version of GATK?  I assumed you were rerunning a 20k shard with the same settings we had used for the 20k.

ldgauthier:
80k is what I had easy access to and what I'm the most invested in
benchmarking right now. Master does fine with the same params. It's slow,
but no failures. We decided to split into 1000 shards (Eric is convinced
that there's a substantial startup cost per shard so we do better in total
cpu-hours on fewer shards) and each of those takes about 24 hours.

kgururaj:
\* Hangs head in shame *
I made a mistake in the buffer size computation in the Java side - over allocated .
Fixed now - consumes approximately the same amount of memory now

droazen:
Thanks @kgururaj !  @ldgauthier let us know whether the new version runs successfully, and we'll proceed to the next round of review if it does.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: Correction - progress isn't printed now. Will try to figure this out

droazen: Ok, thanks -- regular logger updates before/after each batch are crucial for us, as our `GenomicsDBImport` jobs tend to be very long-running.

ldgauthier: In the latest version I'm seeing a lot of 
```
16:26:48.128 INFO  GenomicsDBImport - Starting batch input file preload
16:27:11.629 INFO  GenomicsDBImport - Finished batch preload
```
but I prefer the old "batch X out of Y" format.  We have import jobs that run for 24 hours and it's nice to know how much progress has been made.

ldgauthier:
My 80K sample run has been going for 15 minutes with no problems.  (The old version failed very quickly.)  It will probably take about 3 hours to finish, but I think it's worth moving on with the review.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: Done

ldgauthier:
The 80K shard did finish successfully in a rather reasonable amount of time.  I consider the memory issue fixed.

lbergelson:
@ldgauthier Is going to do a check that we get equivalent output for the 80k shard that we did from the previous version since there are pretty substantial changes going on here.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
```diff
@@ -233,6 +235,14 @@
             minValue = 1)
     private int vcfInitializerThreads = 1;
 
+    @Advanced
+    @Argument(fullName = MAX_NUM_INTERVALS_TO_IMPORT_IN_PARALLEL,
+            shortName = MAX_NUM_INTERVALS_TO_IMPORT_IN_PARALLEL,
+            doc = "Max number of intervals to import in parallel; higher values may improve performance",
```
lbergelson: maybe edit to -> "higher values may be faster but require more memory"?

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
ldgauthier: I'm not getting any logging from querying the GDB either, e.g. running GenotypeGVCFs.

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: Done

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: Fixed

kgururaj:
Modified the CI test that compares with CombineGVCFs so that the interval(s) queried does/do not hit the issue discussed on Monday

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: I see lines:
```
14:46:01.216 INFO  ProgressMeter -              1:69417              0.3                  1000           3505.9
14:46:05.059 WARN  InbreedingCoeff - Annotation will not be calculated, must provide at least 10 samples
14:50:01.511 INFO  ProgressMeter -             1:876592              4.3                  7000           1631.6
14:50:13.315 INFO  ProgressMeter -             1:892516              4.5                 12000           2674.5
```
Is that adequate?

ldgauthier: How many samples was that?  My problem is that I seems to be getting updates every X variants, but for 80,000 samples that takes a very long time.

src/main/java/org/broadinstitute/hellbender/engine/FeatureDataSource.java
```diff
@@ -378,26 +382,61 @@ public static boolean isGenomicsDBPath(final String path) {
...
+                        .setWorkspace(workspace.getAbsolutePath())
+                        .setReferenceGenome(reference.getAbsolutePath())
+                        .setVidMappingFile(vidmapJson.getAbsolutePath())
+                        .setCallsetMappingFile(callsetJson.getAbsolutePath())
+                        .setVcfHeaderFilename(vcfHeader.getAbsolutePath())
+                        .setProduceGTField(false)
+                        .setProduceGTWithMinPLValueForSpanningDeletions(false)
+                        .setSitesOnlyQuery(false)
+                        .setMaxDiploidAltAllelesThatCanBeGenotyped(GenotypeLikelihoods.MAX_DIPLOID_ALT_ALLELES_THAT_CAN_BE_GENOTYPED);
+        File arrayFolder = new File(Paths.get(workspace.getAbsolutePath(), GenomicsDBConstants.DEFAULT_ARRAY_NAME)
```
pshapiro4broad: It's unnecessary to convert the `Path` to a `File` here. It would be simpler to write
```
Path arrayFolder = Paths.get(...);
if (Files.exists(arrayFolder)) {
   ...
```

lbergelson:
@ldgauthier Did you ever get a chance to compare the new 80k shard to the old one for any weirdnesses?  I'd really like to make sure we have comparable results and no weird bugs that didn't show up in small numbers of samples before we merge this.  We've had to many weird bug

ldgauthier:
Not yet.  I had one attempt hang (no progress all weekend) and I'm trying to see if I can find a tiny interval that works.

ldgauthier:
@lbergelson I just realized I didn't specify `-new-qual` so the hang is my fault.  Trying again.  (This is a process because it takes 15 hours to import 80K samples, even if it's just over 4Kb.)

ldgauthier:
Good news!  With the exception of the commandline in the header (also my fault) my two outputs from 4.0.3.0-27-g173ed25 vs. 4.0.4.0-30-gec36b13 for chr20:1265769-1269692 on ~80K samples are BITWISE EQUIVALENT!  Runtime was also pretty similar.  Admittedly I didn't compare memory usage, but I trust Karthik's assessment and I feel good giving this my :+1:

src/main/java/org/broadinstitute/hellbender/tools/genomicsdb/GenomicsDBImport.java
kgururaj: Added test for non adjancent intervals

droazen:
Looks like the only remaining TODOs here are test-related. Before we can merge, we need:

1. Better tests for the case of multiple, non-adjacent intervals (in particular, a test case involving multiple intervals from different contigs, if that is supported, or a test showing that a sensible exception is thrown if it's not supported).
2. A test covering the issue reported in #4716
3. Test cases covering the new sites-only query support (#3688) and support for retrieving the GT field

lbergelson:
@droazen Yay!

PR #4646, opened 2018-04-11
update sv scripts to only copy a single bam file and index, and respect project parameter

cwhelan:
This PR makes four changes to the SV pipeline management scripts:  First it uses the new glob argument for copying data into HDFS to only copy the requested BAM file and its index, rather than everything in the GCS bucket directory (this is useful when the bucket directory contains multiple samples, as is the case with our HGSV trios). Second, the scripts now respect the project argument at each phase of the pipeline. Third, the output directory will include the name of the cluster so that if multiple samples are being processed by the same branch in parallel it's easier to figure out which results directory contains which sample and the results won't collide if two runs happen to start at the same timestamp. Finally, if the user requests not to copy the FASTQ files from the main management script, the script will not direct the SV discovery tool to write them to disk.

### DIFF
--- a/scripts/sv/manage_sv_pipeline.sh
+++ b/scripts/sv/manage_sv_pipeline.sh
@@ -269,8 +269,8 @@ while true; do
                     INIT_ARGS="${INIT_SCRIPT} gs://${GCS_SAVE_PATH}/init/"
                 fi
 
-                echo "create_cluster.sh ${GATK_DIR} ${PROJECT_NAME} ${CLUSTER_NAME} ${CLUSTER_MAX_LIFE_HOURS} ${CLUSTER_MAX_IDLE_MINUTES} ${GCS_REFERENCE_DIR} ${GCS_BAM_DIR} ${INIT_ARGS} 2>&1 | tee -a ${LOCAL_LOG_FILE}" | tee -a ${LOCAL_LOG_FILE}
-                create_cluster.sh ${GATK_DIR} ${PROJECT_NAME} ${CLUSTER_NAME} ${CLUSTER_MAX_LIFE_HOURS} ${CLUSTER_MAX_IDLE_MINUTES} ${GCS_REFERENCE_DIR} ${GCS_BAM_DIR} ${INIT_ARGS} 2>&1 | tee -a ${LOCAL_LOG_FILE}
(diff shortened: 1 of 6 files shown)


scripts/sv/manage_sv_pipeline.sh
```diff
@@ -285,7 +285,7 @@ done
 ########################################################################################################################
 # call runWholePipeline
 # set output directory to datetime-git branch-git hash stamped folder
-OUTPUT_DIR="/results/$(date "+%Y-%m-%d_%H.%M.%S")-${GIT_BRANCH}-${GATK_JAR_HASH}${UNTRACKED_COMMIT}"
+OUTPUT_DIR="/results/$(date "+%Y-%m-%d_%H.%M.%S")-${CLUSTER_NAME}-${GIT_BRANCH}-${GATK_JAR_HASH}${UNTRACKED_COMMIT}"
```
TedBrookings: Since the default CLUSTER_NAME duplicates some of the information already in OUTPUT_DIR, would it meet the stated goals of the PR (while keeping OUTPUT_DIR more concise) to replace it with SANITIZED_BAM?
Independently of this I'd suggest giving OUTPUT_DIR an optional override, which yields the easiest scripting of processing the results. e.g. my suggestion for this line would be something like:
OUTPUT_DIR="/results/${SV_OUTPUT_DIR:-"$(date "+%Y-%m-%d_%H.%M.%S")-${SANITIZED_BAM}-${GIT_BRANCH}-${GATK_JAR_HASH}${UNTRACKED_COMMIT}"}"

cwhelan:
@TedBrookings I implemented these suggestions of yours:

- replace CLUSTER_NAME with SANITIZED_BAM in the results directory
- allow output directory to be overridden by setting SV_OUTPUT_DIR
- rename `runWholePipeline` to `run-whole-pipeline`

I also found and fixed two other bugs:

- added a `set -f` to the create cluster script to avoid having bash expand the wildcard glob for that step (this caused a problem if a file matching the pattern was present locally)
- changed how the copy results script got cluster info so that it parses the result header (this fixes a problem when using preemptible workers because the number of columns in the results of the `gcloud dataproc clusters list` command had a different number of columns

Can you give these changes a quick re-review?

PR #4652, opened 2018-04-12
Numerical consistency between python and java, transpose fix

lucidtronix:
This PR brings pure python inference and java-python streaming inference into much better agreement. also a transpose bugfix.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/filters/ReadGroupBlackListReadFilter.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/filters/ReadGroupBlackListReadFilter.java
@@ -39,14 +39,15 @@ public final class ReadGroupBlackListReadFilter extends ReadFilter implements Se
     // Command line parser requires a no-arg constructor
     public ReadGroupBlackListReadFilter() {};
 
-        /**
-         * Creates a filter using the lists of files with blacklisted read groups.
-         * Any entry can be a path to a file (ending with "list" or "txt" which
-         * will load blacklist from that file. This scheme works recursively
-         * (ie the file may contain names of further files etc).
-         */
+    /**
+     * Creates a filter using the lists of files with blacklisted read groups.
+     * Any entry can be a path to a file (ending with "list" or "txt" which
+     * will load blacklist from that file. This scheme works recursively
+     * (ie the file may contain names of further files etc).
+     */
     public ReadGroupBlackListReadFilter(final List<String> blackLists, final SAMFileHeader header) {
         super.setHeader(header);
+        this.blackList.addAll(blackLists);
         final Map<String, Collection<String>> filters = new TreeMap<>();
         for (String blackList : blackLists) {
             try {
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariants.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariants.java
@@ -10,9 +10,9 @@ import org.broadinstitute.barclay.argparser.*;
 import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 import org.broadinstitute.hellbender.engine.*;
-import org.broadinstitute.hellbender.engine.filters.VariantFilter;
-import org.broadinstitute.hellbender.engine.filters.VariantFilterLibrary;
+import org.broadinstitute.hellbender.engine.filters.*;
 import org.broadinstitute.hellbender.exceptions.GATKException;
+import org.broadinstitute.hellbender.utils.haplotype.HaplotypeBAMWriter;
 import org.broadinstitute.hellbender.utils.io.Resource;
 import org.broadinstitute.hellbender.utils.io.IOUtils;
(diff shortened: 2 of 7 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariants.java
```diff
@@ -319,10 +319,14 @@ private void transferReadsToPythonViaFifo(final VariantContext variant, final Re
         }
         Iterator<GATKRead> readIt = readsContext.iterator();
         if (!readIt.hasNext()) {
-            logger.warn("No reads at contig:" + variant.getContig() + "site:" + String.valueOf(variant.getStart()));
+            logger.warn("No reads at contig:" + variant.getContig() + " site:" + String.valueOf(variant.getStart()));
         }
         while (readIt.hasNext()) {
-            sb.append(GATKReadToString(readIt.next()));
+            GATKRead r  = readIt.next();
+            if (r.getReadGroup().toLowerCase().contains("artificial")){
```
cmnbroad: @lucidtronix Whats the reason for this change ? It seems to be casting a pretty wide net. HC-generated artificial read groups ?

lucidtronix: Yeah exactly, we've always excluded them from the read tensor.  But I agree, I will add an argument to toggle this.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariants.java
```diff
@@ -319,10 +322,14 @@ private void transferReadsToPythonViaFifo(final VariantContext variant, final Re
         }
         Iterator<GATKRead> readIt = readsContext.iterator();
         if (!readIt.hasNext()) {
-            logger.warn("No reads at contig:" + variant.getContig() + "site:" + String.valueOf(variant.getStart()));
+            logger.warn("No reads at contig:" + variant.getContig() + " site:" + String.valueOf(variant.getStart()));
         }
         while (readIt.hasNext()) {
-            sb.append(GATKReadToString(readIt.next()));
+            GATKRead r  = readIt.next();
+            if (!keepArtificialReadGroup && r.getReadGroup().toLowerCase().contains("artificial")){
```
cmnbroad: I see. I was thinking it was overly broad because it matches any group name containing the string "artificial", rather than because of lack of command line control. Instead of doing this manually, I would add an override of getDefaultReadFilters that adds a ReadGroupBlackListReadFilter initialized with HaplotypeBamWriter.DEFAULT_HAPLOTYPE_READ_GROUP_ID (which is the actual string used by HC - will have to be made public), prefixed with the string "ID:". That will provide an exact match to the synthetic HC read group, the engine will filter the reads out automatically so the tool never sees them, and it can be disabled using the existing builtin argument --disable-read-filter.

lucidtronix: Got it. I also had to make the read group black list argument optional and for gatk3 compatibility also filter "ArtificialHaplotype" read groups.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariants.java
```diff
@@ -208,6 +207,16 @@ protected VariantFilter makeVariantFilter(){
         }
     }
 
+    @Override
+    public List<ReadFilter> getDefaultReadFilters() {
+        List<ReadFilter> readFilters = new ArrayList<>();
```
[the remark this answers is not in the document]
lucidtronix: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariants.java
```diff
@@ -208,6 +207,16 @@ protected VariantFilter makeVariantFilter(){
...
+        List<String> filterList = new ArrayList<>();
+        filterList.add("ID:" + HaplotypeBAMWriter.DEFAULT_HAPLOTYPE_READ_GROUP_ID);
+        filterList.add("ID:ArtificialHaplotype"); // GATK3 Compatibility
```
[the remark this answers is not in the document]
lucidtronix: done

src/main/java/org/broadinstitute/hellbender/engine/filters/ReadGroupBlackListReadFilter.java
```diff
@@ -39,14 +39,15 @@
...
-         */
+    /**
+     * Creates a filter using the lists of files with blacklisted read groups.
+     * Any entry can be a path to a file (ending with "list" or "txt" which
+     * will load blacklist from that file. This scheme works recursively
+     * (ie the file may contain names of further files etc).
+     */
     public ReadGroupBlackListReadFilter(final List<String> blackLists, final SAMFileHeader header) {
         super.setHeader(header);
+        this.blackList = blackLists;
```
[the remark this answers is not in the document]
lucidtronix: fixed

cmnbroad:
@lucidtronix This looks good now. Can you squash it down to one commit and rebase  and then we can merge once the tests pass.

PR #4653, opened 2018-04-12
Updates to `gatk` launch script to fix properties and config file definitions.

### LINKED ISSUE #4484: Remove hardcoded system properties from gatk frontend script that collide with those in the GATK config file
We should rely on the GATK config file for the default values for these system properties, rather than hardcoding them into the `gatk` frontend script.

jonn-smith:
Fixes #4484.


droazen:
@jamesemery Can you check whether the system properties like `samjdk.compression_level` are currently being set on Spark workers. Eg., when running on a Spark cluster, within something like a `mapPartitions()` call, check `System.getProperty("samjdk.compression_level")` to see whether it's 2 (the GATK default) or 5 (the htsjdk default).

We need to know the answer to this before we can make a decision on the approach taken in this PR to propagating system properties via owner.

jamesemery:
@droazen @jonn-smith I just checked with a cluster quickly, within each runner both the system property and the htsjdk property were correctly set to 2 or whatever the user requests on the command line on each node.

droazen:
@jamesemery Great, thanks for checking. Could you do a review pass on this when you get a chance? It's not clear that the approach taken here of sending the owner config file around is what we want....it seems like instead we need a way to load the owner config from the launcher script itself.

PR #4655, opened 2018-04-12
Updated top-level class documentation for Funcotator.

### LINKED ISSUE #4654: Funcotator - Update documentation
Documentation for Funcotator needs to be updated to be consistent with current functionality.

jonn-smith:
Fixes #4654

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
@@ -177,7 +177,7 @@ import java.util.stream.Stream;
  *   ./gatk Funcotator \
  *   -R reference.fasta \
  *   -V input.vcf \
- *   -O output.vcf \
+ *   -O outputFile \
  *   --output-file-format MAF \
  *   --data-sources-path dataSourcesFolder/ \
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -172,6 +174,7 @@
  *   -R reference.fasta \
  *   -V input.vcf \
  *   -O output.vcf \
```
LeeTL1220: You sure the output file should still have a name of ".vcf" if the output format is MAF?

jonn-smith:
It's supposed to ignore transcript version numbers, but it looks like right now it doesn't.

I'll add in an issue for it.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
jonn-smith: Good point.  Fixed!

jonn-smith:
It's still accurate, but this info needs to be added to.  I'll merge it, but we'll need to do another one of these in relatively short order.

PR #4671, opened 2018-04-18
Mark --disable-tool-default-read-filters as advanced, and add a warning to its documentation string

droazen:


droazen:
@cmnbroad please review

PR #4676, opened 2018-04-18
updating picard to 2.18.2

lbergelson:
upgrading picard dependency from 2.18.1 -> 2.18.2

this way we'll be on the latest release when we start doing MarkDuplicates tieout


jamesemery:
👍 if the tests pass merge

PR #4677, opened 2018-04-19
fix 4649

### LINKED ISSUE #4649: Error in StructuralVariationDiscoveryPipelineSpark: UnhandledCaseSeen in CpxVariantInterpreter
When run with the current master build against our snapshot HG00514 sample, the experimental variant interpretation pipeline fails with the following exception:

```
18/04/11 20:27:50 WARN org.apache.spark.scheduler.TaskSetManager: Lost task 32.0 in stage 42.0 (TID 56552, cwhelan-hg00514-1-cram-samtools-bam-feature-w-4.c.broad-dsde-methods.internal, executor 28): org.broadinstitute.hellbender.tools.spark.sv.discovery.inference.CpxVariantInterpreter$UnhandledCaseSeen: 1st segment is not overlapping with head alignment but it is not immediately before/after the head alignment either
AssemblyContigWithFineTunedAlignments{sourceTig=(asm022672:tig00004, [1_1430_chr9:130955309-130956738_-_1430M2216S_60_-1_-1_S, 1587_1763_chr9:130955156-130955308_-_1586S54M24I99M1883S_60_-1_-1_S, 1824_2015_chr9:130954964-130955155_-_1823S192M1631S_60_-1_-1_S, 2164_3646_chr9:130953867-130955307_-_2163H167M42I1274M_60_55_1318_O]), insertionMappings=[1963_2177_chr9:130955093-130955304_-_1962H179M3I33M1469H_19_14_138_O], hasEquallyGoodAlnConfigurations=false, saTAGForGoodMappingToNonCanonicalChromosome='NONE'}
        at org.broadinstitute.hellbender.tools.spark.sv.discovery.inference.CpxVariantCanonicalRepresentation.extractAltHaplotypeSeq(CpxVariantCanonicalRepresentation.java:338)
        at org.broadinstitute.hellbender.tools.spark.sv.discovery.inference.CpxVariantCanonicalRepresentation.<init>(CpxVariantCanonicalRepresentation.java:143)
        at org.broadinstitute.hellbender.tools.spark.sv.discovery.inference.CpxVariantInterpreter.lambda$inferCpxVariant$b3be3b47$1(CpxVariantInterpreter.java:53)
        at org.apache.spark.api.java.JavaPairRDD$$anonfun$pairFunToScalaFun$1.apply(JavaPairRDD.scala:1043)
        at org.apache.spark.api.java.JavaPairRDD$$anonfun$pairFunToScalaFun$1.apply(JavaPairRDD.scala:1043)
        at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
        at org.apache.spark.shuffle.sort.BypassMergeSortShuffleWriter.write(BypassMergeSortShuffleWriter.java:149)
        at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:96)
        at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:53)
        at org.apache.spark.scheduler.Task.run(Task.scala:108)
        at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:338)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

SHuang-Broad:
fix #4649 

The cause of the exception is a new edge case that was not imagined when the reference region segmenting logic was initially written.
It is now covered, with updated tests.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
@@ -13,6 +13,7 @@ import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.tools.spark.sv.discovery.AnnotatedVariantProducer;
 import org.broadinstitute.hellbender.tools.spark.sv.discovery.alignment.AlignmentInterval;
 import org.broadinstitute.hellbender.tools.spark.sv.discovery.alignment.AssemblyContigWithFineTunedAlignments;
+import org.broadinstitute.hellbender.tools.spark.sv.discovery.alignment.StrandSwitch;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import scala.Tuple2;
 
@@ -136,8 +137,9 @@ final class CpxVariantCanonicalRepresentation {
             checkBoundedBySharedSingleBase(cpxVariantInducingAssemblyContig);
         }
 
+        final Set<SimpleInterval> twoBaseBoundaries = cpxVariantInducingAssemblyContig.getTwoBaseBoundaries();
         affectedRefRegion = getAffectedReferenceRegion(segmentingLocations);
-        referenceSegments = extractRefSegments(basicInfo, segmentingLocations);
+        referenceSegments = extractRefSegments(basicInfo, segmentingLocations, twoBaseBoundaries);
         eventDescriptions = extractAltArrangements(basicInfo, contigAlignments, jumps, referenceSegments);
 
         altSeq = extractAltHaplotypeSeq(cpxVariantInducingAssemblyContig.getPreprocessedTig(), referenceSegments, basicInfo);
@@ -152,7 +154,8 @@ final class CpxVariantCanonicalRepresentation {
 
     @VisibleForTesting
     static List<SimpleInterval> extractRefSegments(final CpxVariantInducingAssemblyContig.BasicInfo basicInfo,
-                                                   final List<SimpleInterval> segmentingLocations) {
+                                                   final List<SimpleInterval> segmentingLocations,
+                                                   final Set<SimpleInterval> twoBaseBoundaries) {
 
         if (segmentingLocations.size() == 1) { // for case described in {@link checkBoundedBySharedSingleBase}
             return segmentingLocations;
@@ -168,8 +171,14 @@ final class CpxVariantCanonicalRepresentation {
             // there shouldn't be a segment constructed if two segmenting locations are adjacent to each other on the reference
             // this could happen when (in the simplest case), two alignments are separated by a mapped insertion (hence 3 total alignments),
             // and the two alignments' ref span are connected
+            // more generally: only segment when the two segmenting locations are boundaries of alignments that overlap each other (ref. span)
             if (rightBoundary.getStart() - leftBoundary.getEnd() > 1) {
                 segments.add(new SimpleInterval(eventPrimaryChromosome, leftBoundary.getStart(), rightBoundary.getStart()));
+            } else if (rightBoundary.getStart() - leftBoundary.getEnd() == 1) {
+                final SimpleInterval twoBaseSegment = new SimpleInterval(eventPrimaryChromosome, leftBoundary.getEnd(), rightBoundary.getStart());
+                if ( twoBaseBoundaries.contains(twoBaseSegment) ) {
+                    segments.add(twoBaseSegment);
+                }
             }
             leftBoundary = rightBoundary;
         }
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
@@ -335,7 +335,7 @@ final class CpxVariantCanonicalRepresentation {
                                                                                           : (head.referenceSpan.getStart() - firstSegment.getEnd() == 1);
             if ( ! firstSegmentNeighborsHeadAlignment )
                 throw new CpxVariantInterpreter.UnhandledCaseSeen("1st segment is not overlapping with head alignment but it is not immediately before/after the head alignment either\n"
-                        + tigWithInsMappings.toString());
+                        + tigWithInsMappings.toString() + "\nSegments:\t" + segments.toString());
             start = head.endInAssembledContig;
         } else {
             final SimpleInterval intersect = firstSegment.intersect(head.referenceSpan);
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
```diff
@@ -168,8 +171,14 @@ static SimpleInterval getAffectedReferenceRegion(final List<SimpleInterval> even
             // there shouldn't be a segment constructed if two segmenting locations are adjacent to each other on the reference
             // this could happen when (in the simplest case), two alignments are separated by a mapped insertion (hence 3 total alignments),
             // and the two alignments' ref span are connected
+            // more generally: only segment when the two segmenting locations are boundaries of alignments that overlap each other (ref. span)
             if (rightBoundary.getStart() - leftBoundary.getEnd() > 1) {
                 segments.add(new SimpleInterval(eventPrimaryChromosome, leftBoundary.getStart(), rightBoundary.getStart()));
+            } else if (rightBoundary.getStart() - leftBoundary.getEnd() == 1) {
+                final SimpleInterval twoBaseSegment = new SimpleInterval(eventPrimaryChromosome, leftBoundary.getEnd(), rightBoundary.getStart());
```
TedBrookings: Should this be leftBoundary.getEnd() ?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
```diff
@@ -168,8 +171,14 @@ static SimpleInterval getAffectedReferenceRegion(final List<SimpleInterval> even
...
+                if ( twoBaseBoundaries.contains(twoBaseSegment) ) {
+                    segments.add(twoBaseSegment);
+                }
```
TedBrookings: Assuming I'm correct about the above being a typo, you might decrease scope for this kind of error be refactoring thusly:
final int start = rightBoundary.getStart();
final int end = leftBoundary.getEnd();
final int segmentLength = end - start;
if( segmentLength >= 1) {
    final SimpleInterval newSegment = new SimpleInterval(eventPrimaryChromosome, start, end);
    if(segmentLength > 1 || twoBaseBoundaries.contains(newSegment)) {
        segments.add(newSegment);
    }
}

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
```diff
@@ -168,8 +171,14 @@ static SimpleInterval getAffectedReferenceRegion(final List<SimpleInterval> even
             // there shouldn't be a segment constructed if two segmenting locations are adjacent to each other on the reference
             // this could happen when (in the simplest case), two alignments are separated by a mapped insertion (hence 3 total alignments),
             // and the two alignments' ref span are connected
+            // more generally: only segment when the two segmenting locations are boundaries of alignments that overlap each other (ref. span)
             if (rightBoundary.getStart() - leftBoundary.getEnd() > 1) {
                 segments.add(new SimpleInterval(eventPrimaryChromosome, leftBoundary.getStart(), rightBoundary.getStart()));
```
TedBrookings: Should the new SimpleInterval use leftBoundary.getEnd() ?

SHuang-Broad: It actually doesn't matter, because the boundaries are 1bp "point" locations.
I'll make it clear in the docs.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantInducingAssemblyContig.java
```diff
@@ -178,6 +179,31 @@ BasicInfo getBasicInfo() {
...
+     * @return two-base boundaries of contig alignments that are in the valid region, i.e. [alpha, omega]
+     */
+    @VisibleForTesting
+    Set<SimpleInterval> getTwoBaseBoundaries() {
+        // only look at alignments that are not disjoint from [alpha, omega]
+        final SimpleInterval validRegion = new SimpleInterval(basicInfo.eventPrimaryChromosome, basicInfo.alpha.getStart(), basicInfo.omega.getEnd());
+
+        final String chr = basicInfo.eventPrimaryChromosome;
+
+        final Set<SimpleInterval> result = new HashSet<>(contigWithFineTunedAlignments.getAlignments().size());
```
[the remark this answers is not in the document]
SHuang-Broad: yep and done as suggested

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
SHuang-Broad: I see what you mean, but like I replied in the other place, it's a "point" location so won't matter in this case.
And the suggestion you made is good, except that end is always >= start because it is from the "left" (I know, so many mind-numbing details...)

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/CpxVariantCanonicalRepresentation.java
SHuang-Broad: also changed to `getEnd()`

TedBrookings:
Like I said, I don't know enough to say if this approach is feasible, but sometimes it's possible to test an algorithm with a problem-generating function. This works well when it's easy to generate self-consistent solutions and test data consistent with a given solution. I was able to do this for some of the interval overlap functions on the python side of my code. Other methods were too complicated (for me to figure out anyway).

Let's say you have an algorithm function f_algo that you want to test. You design a function test_data_factory that takes as input either a known solution, or a set of parameters from which it can easily generate a known solution (by a route that is logically distinct from the way f_algo works). Then f_tester generates and returns test_data. The flow looks like
```
// get either random test solutions or some distinct set that tests particular edge cases
test_params = get_test_params();
test_solution = easy_solution_from_params(test_params);

test_data = test_data_factory(test_params, test_solution);
assert f_algo(test_data) == test_solution;
```
In some sense you've already done that, but you've personally stepped in as test_data_factory(), and I'm wondering if it would be possible to automate that to get rid of all those numbers in the test code.

SHuang-Broad:
Hmm, I see what you mean.
It’s an interesting idea, and I think Valentin does that for some CIGAR related tests; but I think it is difficult to apply in our case without having a “simulator”. I’ll keep this in mind.

Issue #4633, opened 2018-04-08
CombineGVCFs: java.lang.ClassCastException


huangzhibo:
When the gvcf was merged by "bcftools concat", the following error will be happen.

more info: 
https://gatkforums.broadinstitute.org/gatk/discussion/10817/gatk-runtime-error-on-genotypegvcfs-java-lang-double-cannot-be-cast-to-java-lang-integer

**java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Integer**
	at java.lang.Integer.compareTo(Integer.java:52)
	at java.util.Comparators$NaturalOrderComparator.compare(Comparators.java:52)
	at java.util.Comparators$NaturalOrderComparator.compare(Comparators.java:47)
	at java.util.TimSort.countRunAndMakeAscending(TimSort.java:351)
	at java.util.TimSort.sort(TimSort.java:216)
	at java.util.Arrays.sort(Arrays.java:1507)
	at java.util.stream.SortedOps$SizedRefSortingSink.end(SortedOps.java:302)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:502)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
	at org.broadinstitute.hellbender.utils.Utils.getMedianValue(Utils.java:1137)
	at org.broadinstitute.hellbender.tools.walkers.ReferenceConfidenceVariantContextMerger.mergeAttributes(ReferenceConfidenceVariantContextMerger.java:277)
	at org.broadinstitute.hellbender.tools.walkers.ReferenceConfidenceVariantContextMerger.merge(ReferenceConfidenceVariantContextMerger.java:101)
	at org.broadinstitute.hellbender.tools.walkers.CombineGVCFs.endPreviousStates(CombineGVCFs.java:340)
	at org.broadinstitute.hellbender.tools.walkers.CombineGVCFs.createIntermediateVariants(CombineGVCFs.java:189)
	at org.broadinstitute.hellbender.tools.walkers.CombineGVCFs.apply(CombineGVCFs.java:134)
[...]

Issue #4525, opened 2018-03-12
CombineGVCFs java.lang.IllegalArgumentException: Unexpected base in allele bases '*AACC'


chandrans:
### Affected tool(s)
CombineGVCFs

### Affected version(s)
4.0.2.0

### Description 
CombineGVCFs fails with an IllegalArgumentException when trying to combine GVCFs. However, when trying to combine a different subset of the GVCFs, it fails with a ClassCastException. The GVCFs all validate with ValidateVariants and were produced with GATK4 HaplotypeCaller.

#### Steps to reproduce
Details in https://github.com/broadinstitute/dsde-docs/issues/2990

This Issue was generated from your [forums] 
[forums]: https://gatkforums.broadinstitute.org/gatk/discussion/11503/combinegvcfs-java-lang-illegalargumentexception-unexpected-base-in-allele-bases-aacc/p1

chandrans:
A second user has reported it as well. https://gatkforums.broadinstitute.org/gatk/discussion/11693/combine-gvcfs-error

droazen:
@cmnbroad Could you please have a look at this one when you get a chance, since you're already in this tool at the moment for https://github.com/broadinstitute/gatk/issues/4572? We've been asked to prioritize this ticket. Thanks!

cmnbroad:
@chandrans I don't have access to dsde-docs so I can't see the ticket/test files (we asked @vdauwera to give me access last week for a different issue, but I don't have it yet).

chandrans:
Oh. I will slack you Chris.

cmnbroad:
Fixed via https://github.com/broadinstitute/gatk/pull/4680.

PR #4261, opened 2018-01-25
GenomicsDB Allele-specific Annotation Support and Bug Fixes

psfoley:
This PR adds:
- Allele Annotation Support #3707 (w/ changes requested in [4047](https://github.com/broadinstitute/gatk/pull/4047))

and fixes:
- [4160](https://github.com/broadinstitute/gatk/issues/4160)
- [4106](https://github.com/broadinstitute/gatk/issues/4106)
- [3736](https://github.com/broadinstitute/gatk/issues/3736)
- [3745](https://github.com/broadinstitute/gatk/issues/3745)
- Bug fix - Resize LUTs when dealing with spanning deletions

PR #4680, opened 2018-04-19
ReferenceConfidenceVariantContextMerger fixes for spanning deletions, attribute merging.

cmnbroad:
Fixes for https://github.com/broadinstitute/gatk/issues/4525 and https://github.com/broadinstitute/gatk/issues/4633. Longer term we should do a more significant rewrite/refactoring.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
@@ -4,6 +4,8 @@ import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.util.Locatable;
 import htsjdk.variant.variantcontext.*;
 import htsjdk.variant.vcf.VCFConstants;
+import htsjdk.variant.vcf.VCFHeader;
+import htsjdk.variant.vcf.VCFInfoHeaderLine;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.tools.walkers.annotator.VariantAnnotatorEngine;
 import org.broadinstitute.hellbender.tools.walkers.annotator.allelespecific.AlleleSpecificAnnotationData;
@@ -28,12 +30,17 @@ import java.util.stream.Stream;
 public final class ReferenceConfidenceVariantContextMerger {
 
     private final GenotypeLikelihoodCalculators calculators;
+    private final VCFHeader vcfInputHeader;
     protected final VariantAnnotatorEngine annotatorEngine;
-    protected final OneShotLogger warning = new OneShotLogger(this.getClass());
+    protected final OneShotLogger oneShotAnnotationLogger = new OneShotLogger(this.getClass());
+    protected final OneShotLogger oneShotHeaderLineLogger = new OneShotLogger(this.getClass());
 
-    public ReferenceConfidenceVariantContextMerger(VariantAnnotatorEngine engine){
+    public ReferenceConfidenceVariantContextMerger(VariantAnnotatorEngine engine, final VCFHeader inputHeader) {
+        Utils.nonNull(inputHeader, "A VCF header must be provided");
+
         calculators = new GenotypeLikelihoodCalculators();
         annotatorEngine = engine;
+        vcfInputHeader = inputHeader;
     }
 
     /**
@@ -173,6 +180,9 @@ public final class ReferenceConfidenceVariantContextMerger {
 
         for (final Allele a : vc.getAlternateAlleles()) {
             if (a.isSymbolic()) {
+                result.add(a);
+            } else if ( a == Allele.SPAN_DEL ) {
+                // add SPAN_DEL directly so we don't try to extend the bases
                 result.add(a);
             } else if (a.isCalled()) {
                 result.add(extendAllele(a, extraBaseCount, refBases));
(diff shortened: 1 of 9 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
```diff
@@ -379,19 +388,42 @@ private void addReferenceConfidenceAttributes(final VCWithNewAlleles vcPair, fin
...
     }
 
-    private Comparable<?> parseNumber(String stringValue) {
-        if (stringValue.contains(".")) {
-            return Double.parseDouble(stringValue);
-        } else {
-            return Integer.parseInt(stringValue);
+    // Use the VCF header's declared type for the given attribute to ensure that all the values for that attribute
+    // across all the VCs being merged have the same boxed representation. Some VCs have a serialized value of "0"
+    // for FLOAT attributes, with no embedded decimal point, but we still need to box those into Doubles, or the
```
ldgauthier: Of all the terrible things that happen to annotations in htsjdk, at least it guarantees some decimal precision (VCFEncoder.formatVCFDouble()). I'm guessing this was an issue related to GenomicsDB output (which uses htslib)?  I mentioned this to Karthik a while ago (https://github.com/broadinstitute/gatk/pull/4047#issuecomment-356063579)  It should have been addressed by #4261, but I haven't tested that explicitly.  Not that I object to the belt and suspenders approach.

cmnbroad: @ldgauthier Good to know - thanks. At least two users have reported this, one of them seemed to [implicate bcftools](https://github.com/broadinstitute/gatk/issues/4633), but that may be a red herring. I do think htsjdk can produce such output though, since it appears to rely on the in-memory representation of the value provided by the caller, rather than the header, to determine the format. I’d love to tighten that up in the next round of htsjdk updates.

src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
```diff
@@ -379,19 +388,42 @@ private void addReferenceConfidenceAttributes(final VCWithNewAlleles vcPair, fin
...
+    // subsequent sorting required to obtain the median will fail due to the list having a mix of Comparable<Integer>
+    // and Comparable<Double>.
+    private Comparable<?> parseNumericInfoAttributeValue(final VCFHeader vcfHeader, final String key, final String stringValue) {
+        final VCFInfoHeaderLine infoLine = vcfHeader.getInfoHeaderLine(key);
+        if (infoLine == null) {
+            warning.warn(String.format("No header line for INFO attribute %s was found in the input header", key));
```
jamesemery: This warning seems problematic. If it gets triggered it will probably get spit out for EVERY read in a VCF which would result in a torrential span. Perhaps us a OneShotLogger here to save peoples poor poor bash histories...

jamesemery: Looking at the code leading up to this I don't expect this warning will be a problem for any non-reducible annotations except i'm not sure i completely trust our VCFStandardHeaderLines to have the correct types. We have a number of issues with the header lines either here or in HTSJDK being wrong. This is still better I would say.

cmnbroad: Excellent point - but `warning` *is* a `OneShotLogger`. Also, I did keep the fallback code for the old behavior if the header line isn't found.

jamesemery: Right, good point about `warning`. The next point is that currently its just a dumb implementation that only emits once. If there are multiple error states in a file its probably best to open multiple loggers so if both are triggered they will each be written out, warning is currently used for a different case to do with merging and numerical parsing. maybe split them into 2 loggers?

cmnbroad: Right. Split out into a second logger, with a slightly more generic message.

cmnbroad: Also, I rebased this on top of my master to pick up my last CombineGVCFs branch/fix.

src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
```diff
@@ -379,19 +389,41 @@ private void addReferenceConfidenceAttributes(final VCWithNewAlleles vcPair, fin
...
-            return Integer.parseInt(stringValue);
+    // Use the VCF header's declared type for the given attribute to ensure that all the values for that attribute
+    // across all the VCs being merged have the same boxed representation. Some VCs have a serialized value of "0"
+    // for FLOAT attributes, with no embedded decimal point, but we still need to box those into Doubles, or the
+    // subsequent sorting required to obtain the median will fail due to the list having a mix of Comparable<Integer>
+    // and Comparable<Double>.
+    private Comparable<?> parseNumericInfoAttributeValue(final VCFHeader vcfHeader, final String key, final String stringValue) {
+        final VCFInfoHeaderLine infoLine = vcfHeader.getInfoHeaderLine(key);
+        if (infoLine == null) {
+            oneShotHeaderLineLogger.warn(String.format("At least one attribute was found (%s) for which there no corresponding header line", key));
```
jamesemery: there no -> there is no

src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
```diff
@@ -28,12 +30,17 @@
 public final class ReferenceConfidenceVariantContextMerger {
 
     private final GenotypeLikelihoodCalculators calculators;
+    private final VCFHeader vcfInputHeader;
     protected final VariantAnnotatorEngine annotatorEngine;
     protected final OneShotLogger warning = new OneShotLogger(this.getClass());
+    protected final OneShotLogger oneShotHeaderLineLogger = new OneShotLogger(this.getClass());
```
jamesemery: While you're at it, could you rename warning to something like `nonNumericalAttributeWarning` or something to that end so its clearer in the future.

Issue #4672, opened 2018-04-18
CombineGVCFs --break-bands-at-multiples-of can generate invalid intermediate intervals 


cmnbroad:
This happens whenever the start position of an interval for which intermediate bands must be created is less than the value `of break-bands-at-multiples-of`. For example, an input reference block record with a `start` position (say 1) that is less than the value of `of break-bands-at-multiples-of` (say 10000) would result in the invalid intermediate band interval:
```
java.lang.IllegalArgumentException: Invalid interval. Contig:chr21 start:-1 end:-1

	at org.broadinstitute.hellbender.utils.Utils.validateArg(Utils.java:687)
	at org.broadinstitute.hellbender.utils.SimpleInterval.validatePositions(SimpleInterval.java:61)
	at org.broadinstitute.hellbender.utils.SimpleInterval.<init>(SimpleInterval.java:37)
	at org.broadinstitute.hellbender.tools.walkers.CombineGVCFs.createIntermediateVariants(CombineGVCFs.java:191)
	at org.broadinstitute.hellbender.tools.walkers.CombineGVCFs.apply(CombineGVCFs.java:134)

```
This doesn't happen in GATK3.

PR #4681, opened 2018-04-19
Fix and test for calculating CombineGVCFs intermediate band interval start locations.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/4672.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFs.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFs.java
@@ -152,15 +152,15 @@ public final class CombineGVCFs extends MultiVariantWalkerGroupedOnStart {
      */
     @VisibleForTesting
     void createIntermediateVariants(SimpleInterval intervalToClose) {
-        Set<Integer> sitesToStop = new HashSet<>();
         resizeReferenceIfNeeded(intervalToClose);
 
         // Break up the GVCF according to the provided reference blocking scheme
-        if ( multipleAtWhichToBreakBands > 0) {
-            for (int i = ((intervalToClose.getStart())/multipleAtWhichToBreakBands)*multipleAtWhichToBreakBands; i <= intervalToClose.getEnd(); i+=multipleAtWhichToBreakBands) {
-                sitesToStop.add(i-1); // Subtract 1 here because we want to split before this base
-            }
-        }
+        // The values returned from getIntermediateStopSites represent a proposed set of stop sites that may include
+        // intervals that are outside the actual interval being closed. These sites are filtered out below.
+        // Note: Precomputing these is really inefficient when large reference blocks are closed with
+        // fine band resolution because it results in very large collections of stop sites (tens or hundreds of millions)
+        // that must subsequently be sorted.
+        final Set<Integer> sitesToStop = getIntermediateStopSites(intervalToClose, multipleAtWhichToBreakBands);
 
         // If any variant contexts ended (or were spanning deletions) the last context compute where we should stop them
         for (VariantContext vc : variantContextsOverlappingCurrentMerge) {
@@ -181,7 +181,7 @@ public final class CombineGVCFs extends MultiVariantWalkerGroupedOnStart {
         List<Integer> stoppedLocs = new ArrayList<>(sitesToStop);
         stoppedLocs.sort(Comparator.naturalOrder());
 
-        // For each stopped loc, create a fake QueuedContextState and pass it to endPreviousStats
+        // For each stopped loc that is within the interval being closed, create a fake QueuedContextState and pass it to endPreviousStats
         for (int stoppedLoc : stoppedLocs) {
             SimpleInterval loc = new SimpleInterval(intervalToClose.getContig(), stoppedLoc, stoppedLoc);
             if (( stoppedLoc <= intervalToClose.getEnd() && stoppedLoc>= intervalToClose.getStart()) && isWithinInterval(loc)) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFs.java
```diff
@@ -157,8 +157,14 @@ void createIntermediateVariants(SimpleInterval intervalToClose) {
 
         // Break up the GVCF according to the provided reference blocking scheme
         if ( multipleAtWhichToBreakBands > 0) {
-            for (int i = ((intervalToClose.getStart())/multipleAtWhichToBreakBands)*multipleAtWhichToBreakBands; i <= intervalToClose.getEnd(); i+=multipleAtWhichToBreakBands) {
-                sitesToStop.add(i-1); // Subtract 1 here because we want to split before this base
+            // if the intermediate interval to close starts before the end of the first interval,
```
jamesemery: This seems like a perfectly reasonable fix but i would probably add an extra layer of protection here. Below where sitesToStop is converted into intervalLocs you should also add a check that the intervals generated by this or the other break code are not going to be invalid. Also i think this is still broken with multipleAtWhichToBreakBands = 1 because you get a start position = 0

src/test/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFsIntegrationTest.java
```diff
@@ -74,7 +74,15 @@
...
                 //Testing allele-specific annotations missing AS_Standard Group
-                {new File[]{getTestFile("NA12878.AS.chr20snippet.g.vcf"), getTestFile("NA12892.AS.chr20snippet.g.vcf")}, getTestFile("testAlleleSpecificAnnotationsNoGroup.vcf"), Arrays.asList("-G", "Standard", "-G", "AS_Standard"), b37_reference_20_21},};
+                {new File[]{getTestFile("NA12878.AS.chr20snippet.g.vcf"), getTestFile("NA12892.AS.chr20snippet.g.vcf")}, getTestFile("testAlleleSpecificAnnotationsNoGroup.vcf"), Arrays.asList("-G", "Standard", "-G", "AS_Standard"), b37_reference_20_21},
+                //Test that trailing reference blocks are emitted with correct intervals
+                {new File[]{getTestFile("gvcfExample1WithTrailingReferenceBlocks.g.vcf"), getTestFile("gvcfExample2WithTrailingReferenceBlocks.g.vcf")}, getTestFile("gvcfWithTrailingReferenceBlocksExpected.g.vcf"), NO_EXTRA_ARGS, b38_reference_20_21},
+                // same test as the previous one, except with a band multiple specified
+                {new File[]{getTestFile("gvcfExample1WithTrailingReferenceBlocks.g.vcf"), getTestFile("gvcfExample2WithTrailingReferenceBlocks.g.vcf")},
```
jamesemery: Add a test with either/both of useBPResolution or bands of size 1 to demonstrate that there isn't a band being made at 0.

src/main/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFs.java
cmnbroad: @jamesemery I'm not clear on what kind of interval validation you're suggesting - if we ever do create invalid interval I think it would fail the same way it does now. You're right about the start position 0 - I wound up factoring out the banded stop site calculation for bands into a separate method that can be unit tested, and added a bunch of test cases that verify the stop sites that are generated, and that they result in valid intervals.

src/test/java/org/broadinstitute/hellbender/tools/walkers/CombineGVCFsUnitTest.java
```diff
@@ -0,0 +1,54 @@
...
+                { new SimpleInterval("contig", 1, 2), 2, Arrays.asList(1) },
+                { new SimpleInterval("contig", 1, 10), 2, Arrays.asList(1, 3, 5, 7, 9) },
+                { new SimpleInterval("contig", 1, 10), 5, Arrays.asList(4, 9) },
+                { new SimpleInterval("contig", 1, 100), 25, Arrays.asList(24, 49, 74, 99) },
+
+                { new SimpleInterval("contig", 10, 10), 2, Arrays.asList(9) },
+                { new SimpleInterval("contig", 10, 10), 5, Arrays.asList(9) },
+
+                // TODO ?? is this the correct result ?
+                { new SimpleInterval("contig", 10, 10), 10, Arrays.asList(9) },
```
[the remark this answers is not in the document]
jamesemery: ehhh... i don't consider this critical or even important at all, but all of these tests that return Collections.EMPTY_LIST as a result are only ever failing on the first breakpoint interval, which technically goes through a new codepath. If you wanted you could add a test { new SimpleInterval("contig", 110, 120), 100, Collections.EMPTY_LIST }, to be more through.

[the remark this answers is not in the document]
cmnbroad: Agreed on the first comment. I view the purpose of this test as validating that we don't create invalid intermediate intervals at this stage of the calculation, since thats what was originally failing - I think the fact that they get filtered out later is ok. Ideally we'd have additional unit tests for the subsequent parts of the calculation, but they require much more state.

I think the additional test you mentioned should return {"99"}, not the empty list, for this method. Again, that stop site will get filtered later as its outside of the interval. Its testing the case where the band multiple is smaller than the start of the close interval, which is essentially the same as the `{ new SimpleInterval("contig", 10, 20), 5, Arrays.asList(9, 14, 19) }` test.  Doesn't hurt to add it though.

PR #4690, opened 2018-04-22
Better default AF for M2 tumor-normal mode; improved M2 STR filter

davidbenjamin:
@takutosato The first commit gives tumor-normal the old default AF.  The second commit is a more intelligent STR contraction filter that improves both sensitivity and precision.


davidbenjamin:
@takutosato The one test failure occurs because when there is low normal coverage germline risk might get triggered too often.  I might need to adjust the default value but it won't affect the code otherwise.

PR #4697, opened 2018-04-24
Replace StreamingProcessController println with a logger info message.

cmnbroad:
Simple one liner to replace an errant println.


droazen:
:+1:

Issue #4702, opened 2018-04-24
barclay fails to set collection arguments initialized with an immutable collection


droazen:
If you have a barclay `@Argument` field of type `List`, barclay will fail to set the field value properly when the argument is specified if the `List` is initialized using an immutable Collection, such as that returned by `Collections.emptyList()`. Example error:

```
	at java.util.AbstractList.add(AbstractList.java:148)
	at java.util.AbstractList.add(AbstractList.java:108)
	at org.broadinstitute.barclay.argparser.CommandLineArgumentParser.setArgument(CommandLineArgumentParser.java:706)
	at org.broadinstitute.barclay.argparser.CommandLineArgumentParser.parseArguments(CommandLineArgumentParser.java:427)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.parseArgs(CommandLineProgram.java:220)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:194)
	at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:160)
	at org.broadinstitute.hellbender.Main.mainEntry(Main.java:203)
	at org.broadinstitute.hellbender.Main.main(Main.java:289)
```

droazen:
For @cmnbroad

magicDGS:
There's a solution here: https://github.com/broadinstitute/barclay/pull/102

magicDGS:
Also, some discussion about what to do with immutable collections in barclay is here: https://github.com/broadinstitute/barclay/issues/100

PR #4703, opened 2018-04-24
Fix two instances of immutable collections being used to initialize List arguments

droazen:
Barclay can't currently handle immutable collections in `@Argument` values due to
https://github.com/broadinstitute/gatk/issues/4702

Tests for these arguments are coming in a separate, larger branch, but I
wanted to get the fixes in first since it's such a simple fix.


jonn-smith:
Looks good to me.

PR #4704, opened 2018-04-25
A bit of tidying Mutect2 code

davidbenjamin:
@takutosato


src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
```diff
@@ -187,12 +185,7 @@ public void apply(final AssemblyRegion region, final ReferenceContext referenceC
 
     @Override
     public void closeTool() {
-        if ( vcfWriter != null ) {
```
takutosato: I have no objections to this change but I think @lbergelson has once told me to check whether the writer is null in closeTool().

PR #4706, opened 2018-04-25
add method to compare two locatables assumed to have the same contig

SHuang-Broad:


SHuang-Broad:
engine team want to take a look?

src/main/java/org/broadinstitute/hellbender/utils/IntervalUtils.java
```diff
@@ -84,17 +84,36 @@ public static final int compareLocatables(Locatable first, Locatable second, SAM
             // compare the contigs
             result = compareContigs(first, second, dictionary);
             if (result == 0) {
-                // compare start position
-                result = Integer.compare(first.getStart(), second.getStart());
-                if (result == 0) {
-                    // compare end position
-                    result = Integer.compare(first.getEnd(), second.getEnd());
-                }
+                result = compareLocatablesOfSameContig(first, second);
```
magicDGS: This change checks again tha the contigs are the same, which was already done in `compareContigs`. I think that this produces a performance regression...

SHuang-Broad: Thanks, and I see your concern.
The reason I did that check in the new refactored out method, which is `public`, is to prevent callers from accidentally using it for locatables from different contigs, so a safe-guard.
Do you have better suggestions?

magicDGS: What's about a private method `comparePosition` without the check, which is called by both `compareLocatables` and `compareLocatablesOfSameContig`? Like that, there is no code repetition (less error prone) and no overhead of checking twice in the methods that does not need it.

src/main/java/org/broadinstitute/hellbender/utils/IntervalUtils.java
```diff
@@ -84,17 +84,36 @@ public static final int compareLocatables(Locatable first, Locatable second, SAM
...
         return result;
     }
 
+    /**
+     * Compare two locations assuming they are of the same contig.
+     *
+     * @return 0 if first and second are equal, a negative value if first < second or a positive value if first > second,
+     * @throws IllegalArgumentException if two locatables are not of the same contig
+     */
+    public static final int compareLocatablesOfSameContig(final Locatable first, final Locatable second) {
```
lbergelson: @SHuang-Broad What's the need for this method?  Are you seeing that checking the contigs for equality is expensive somewhere in your code?

SHuang-Broad: I was trying to avoid hauling around sequence dictionary around, which the general method requires.

SHuang-Broad:
closing after discussing with the engine team that it's simpler to roll up one's own comparator
Thanks @magicDGS for looking!

PR #4710, opened 2018-04-26
M2 NIO wdl -- updated for Firecloud pet service accounts

davidbenjamin:
@LeeTL1220 Ready for review.  I will concurrently test in Firecloud.


scripts/mutect2_wdl/mutect2_nio.wdl
```diff
@@ -0,0 +1,972 @@
...
+    File ref_fai
+    File ref_dict
+    File tumor_bam
+    File tumor_bai
+    File? normal_bam
+    File? normal_bai
+    File? pon
+    Int scatter_count
+    File? gnomad
+    File? variants_for_contamination
```
LeeTL1220: Why eliminate all the index files?  Seems like that will make it harder to keep the files in sync.  Your call.

LeeTL1220:
If it works in FC, then feel free to merge.  Addressing my comment is your decision.

davidbenjamin:
Working on Firecloud after one tweak: `if [ -f ${normal_bam} ]` doesn't work when `normal_bam` is a `String` because nothing is localized, so I replaced it with `if [ ! -z ${normal_bam} ]`.  I will merge around 6 am after Travis tests hopefully pass.

PR #4711, opened 2018-04-26
minor improvements to code in IOUtils and UserException.BadTmpDir

lbergelson:
Two commits here:

- The first is to fix a no longer accurate message in `UserException.BadTmpDir`
- The second is a few improvements to IOUtils.  
  1.  Rename and simplify `tmpDir` -> `createTempDir` and make it automatically scheduled for deletion
  2. Add documentation to the confusing `absolute` method so that I stop wondering what it's for

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/copynumber/PostprocessGermlineCNVCalls.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/copynumber/PostprocessGermlineCNVCalls.java
@@ -432,7 +432,7 @@ public final class PostprocessGermlineCNVCalls extends GATKTool {
             logger.info("Generating segments VCF file...");
 
             /* perform segmentation */
-            final File pythonScriptOutputPath = IOUtils.createTmpDir("gcnv-segmented-calls");
+            final File pythonScriptOutputPath = IOUtils.createTempDir("gcnv-segmented-calls");
             final boolean pythonScriptSucceeded = executeSegmentGermlineCNVCallsPythonScript(
                     sampleIndex, contigPloidyCallsPath, sortedCallsShardPaths, sortedModelShardPaths,
                     pythonScriptOutputPath);
--- a/src/main/java/org/broadinstitute/hellbender/tools/copynumber/PostprocessGermlineCNVCalls.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/copynumber/PostprocessGermlineCNVCalls.java
@@ -432,7 +432,7 @@ public final class PostprocessGermlineCNVCalls extends GATKTool {
             logger.info("Generating segments VCF file...");
 
             /* perform segmentation */
-            final File pythonScriptOutputPath = IOUtils.tempDir("gcnv-segmented-calls", "");
+            final File pythonScriptOutputPath = IOUtils.createTmpDir("gcnv-segmented-calls");
             final boolean pythonScriptSucceeded = executeSegmentGermlineCNVCallsPythonScript(
                     sampleIndex, contigPloidyCallsPath, sortedCallsShardPaths, sortedModelShardPaths,
                     pythonScriptOutputPath);
--- a/src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
@@ -62,20 +62,25 @@ public final class IOUtils {
     }
 
     /**
-     * Creates a temp directory with the prefix and optional suffix.
+     * Creates a temp directory with the given prefix.
      *
      * The directory and any contents will be automatically deleted at shutdown.
+     *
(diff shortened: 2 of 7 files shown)


src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -64,36 +64,18 @@ public static boolean isBamFileName(final String inputFileName) {
     /**
      * Creates a temp directory with the prefix and optional suffix.
```
droazen: You don't accept a suffix any more...

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -64,36 +64,18 @@ public static boolean isBamFileName(final String inputFileName) {
...
-     * @param suffix       Optional suffix for the directory name.
-     * @return The created temporary directory.
-     */
-    public static File tempDir(String prefix, String suffix) {
-        return tempDir(prefix, suffix, null);
-    }
-
-    /**
-     * Creates a temp directory with the prefix and optional suffix.
+     * The directory and any contents will be automatically deleted at shutdown.
```
droazen: Add a note that the method only works if the temp directory is a File

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -158,17 +140,18 @@ else if (file.exists())
...
+     *   ex:    /path/./to/./directory/. -> /path/to/directory/
+     *
+     * It differs from {@link File#getCanonicalFile()} in that it doesn't dereference symbolic links or completely
+     * simplify the file path.  For instance ".." is not simplified by this method.
      *
      * @param file the file.
-     * @return the absolute path to the file.
+     * @return the absolute path to the file with redundant "./" removed.
      */
     public static File absolute(File file) {
```
droazen: Can you either: find a better name for this method, or just use `getCanonicalFile()` directly (if you're uncomfortable with the latter option, try asking @kshakir about the genesis of this method in GATK3)

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
```diff
@@ -214,7 +197,7 @@ public static File writeTempResource(Resource resource) {
         try {
             temp = File.createTempFile(FilenameUtils.getBaseName(resource.getPath()) + ".", "." + FilenameUtils.getExtension(resource.getPath()));
         } catch (IOException e) {
-            throw new UserException.BadTmpDir(e.getMessage());
+            throw new UserException.BadTmpDir(e.getMessage(), e);
```
droazen: `BadTmpDir` -> `BadTempDir`

src/main/java/org/broadinstitute/hellbender/tools/copynumber/PostprocessGermlineCNVCalls.java
```diff
@@ -432,7 +432,7 @@ private void generateSegmentsVCFFileFromAllShards() {
             logger.info("Generating segments VCF file...");
 
             /* perform segmentation */
-            final File pythonScriptOutputPath = IOUtils.tempDir("gcnv-segmented-calls", "");
+            final File pythonScriptOutputPath = IOUtils.createTmpDir("gcnv-segmented-calls");
```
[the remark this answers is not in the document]
lbergelson: done

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
lbergelson: getAbsoluteAndPartiallyCanonicalizedFile?

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java
lbergelson: Ah.  That makes sense.  paths.normalize seems like a better plan.

lbergelson: absolute -> Paths.normalize()

PR #4720, opened 2018-04-30
gcnvkernel tunings

mbabadi:
Notes:

- We still need to decide about `GermlineCNVCaller` default arguments. See issue #4719.
- The case denoising and calling is unlikely to benefit from the marginalized warmup strategy. Therefore, it is not included in this PR.


mwalker174:
I tested this on a collection of 32 tandem duplications (~10kbp-50kbp events, 5kbp padding, 100bp bins) from an HGSV proband case. Using the latest release docker it nailed almost all of them (hooray!) but it hardly called any dups when I tried to do the calling/postprocessing locally using this branch (log reports successful convergence). This could just be an error on my end but we should check before merging.

mwalker174:
@mbabadi @samuelklee After our discussion, I suspect my issue was related to sharding and is probably not related to any changes in this branch.

samuelklee:
@mbabadi can you make any changes and merge this when you get a chance?

mbabadi:
@samuelklee done!

Issue #4340, opened 2018-02-05
Create a roadmap document for HTSJDK 3.0


droazen:
Work with project maintainers and stakeholders to create a roadmap document for the next generation of the HTSJDK file-parsing library, and obtain buy-in from all stakeholders.

From our point of view, should address major pain points for GATK:

* NIO support everywhere
* Interfaces for all data types
* Easily extensible/pluggable, easy to add parsers for new versions of our formats (such as the still-unsupported VCF 4.3 and BCF 2.2)
* Fix dumb design decisions that are killing GATK performance (such as always slurping tabix indices into memory)

magicDGS:
Nice that HTSJDK is moving forward to version 3. The points that I would like to address are similar to yours, with some inclussions

* Regarding NIO support, I would go to remove completely `File` support. If API users need to use the `File` abstraction, they should convert to a `java.nio.Path` using the `toPath` method.
* In addition, I would like that HTTP/S and FTP is handled also with NIO. For HTTP/S, I am working in a simple `FileSystemProvider` that should be good enough for using in combination with HTSJDK ([jsr203-http](https://github.com/magicDGS/jsr203-http)), and I can speed up the development there for needs in HTSJDK; for FTP, maybe [ftp-fs](https://github.com/robtimus/ftp-fs) can be used or a simple implementation can be derived from the HTTP/S implementation (without credentials). This will remove the special handling of HTTP/S and FTP paths in HTSJDK in favor of a consistent and pluggable manner.
* Interfaces for the data types are great, and maybe it will be good to have codec interfaces for both encoding and decoding. For example, I am missing encoders in tribble (an attempt in https://github.com/samtools/htsjdk/pull/822 for writing support).
* For VCF, I would like to have a less diploid-centric interface and design, or at least a way of configure the catching of genotype-related attributes. Currently there are methods for homozygotes/heterozygotes that aren't really useful for triploids or even VCFs without variation (for example, in Pool-Seq data).
* Modular design for artifacts: thus, a project with only SAM/BAM requirements will require only `htsjdk-sam`, and if they also want CRAM support, `htsjdk-cram`. See https://github.com/samtools/htsjdk/issues/896 for more info about it.
* Common license for all HTSJDK, or at least for each module. This will be good for taking into account legal concerns when including the library, because now there is a mixture depending on the files that are used.

This is what is coming to my mind now. Maybe I added something else in https://github.com/samtools/htsjdk/issues/520

[...]

PR #4731, opened 2018-05-02
adding new ReadQueryNameComparator

lbergelson:
* this should match the results of SAMRecordQueryNameComparator exactly
* it operates on GATKRead instead of SAMRecord

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/read/ReadQueryNameComparator.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/read/ReadQueryNameComparator.java
@@ -1,0 +1,61 @@
+package org.broadinstitute.hellbender.utils.read;
+
+import htsjdk.samtools.SAMRecordQueryNameComparator;
+import htsjdk.samtools.SAMTag;
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+/**
+ * compare {@link GATKRead} by queryname
+ * duplicates the exact ordering of {@link SAMRecordQueryNameComparator}
+ */
+public class ReadQueryNameComparator implements Comparator<GATKRead>, Serializable {
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public int compare(final GATKRead read1, final GATKRead read2) {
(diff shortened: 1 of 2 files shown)


src/test/java/org/broadinstitute/hellbender/utils/read/ReadQueryNameComparatorUnitTest.java
```diff
@@ -0,0 +1,134 @@
...
+        for(GATKRead left: reads){
+            for(GATKRead right: reads){
+                tests.add(new Object[]{left, right});
+            }
+        };
+        return tests.iterator();
+    }
+
+    @Test(dataProvider = "getReads")
+    public void testTieBreakers(GATKRead left, GATKRead right){
```
jamesemery: this looks like a much cleaner approach than the equivalent approach in ReadsCoordinateComparator

src/main/java/org/broadinstitute/hellbender/utils/read/ReadQueryNameComparator.java
```diff
@@ -0,0 +1,61 @@
...
+
+import htsjdk.samtools.SAMRecordQueryNameComparator;
+import htsjdk.samtools.SAMTag;
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+/**
+ * compare {@link GATKRead} by queryname
+ * duplicates the exact ordering of {@link SAMRecordQueryNameComparator}
```
magicDGS: Duplication of code from HTSJDK to GATK (and other downstream projects) is a sign that htsjdk needs to evolve to an interface-based - as @lbergelson is a maintainer, I point out again that #4340 and htsjdk-related issues should be a high priority!

lbergelson: @magicDGS You're very right.  We had hoped to begin work on a major htsjdk revamp last quarter, but we didn't have enough bandwidth to take it on along with other requirements we had.  We're hoping to start very soon.

PR #4732, opened 2018-05-02
Fixed MarkDuplicatesSpark handling of unsorted bams

### LINKED ISSUE #4701: Fix inconsistent marking with coordinate sorted bams in ReadsSparkSource
Currently there is a bug in MarkDuplicatesSpark where for non queryname sorted bams the index is incorrectly assumed to be sufficient for reads groups, which wis guaranteed to be adequate for queryname sorted bams. This means that under some circumstances reads can be inconsistently marked as duplicates when its mate is a non-duplicate if they were spread across multiple partitions. 

There is a disabled `MarkDuplicatesSparkUtilsUnitTests` test in #4656 that should hopefully demonstrate this issue.

jamesemery:
Added a global sort to the beginning of the tool to ensure we are always working with name grouped bams. In the future we should evaluate if alternatives that avoid sorting are necessary.  

Fixes  #4701

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSpark.java
@@ -89,12 +89,18 @@ public final class MarkDuplicatesSpark extends GATKSparkTool {
                                          final MarkDuplicatesScoringStrategy scoringStrategy,
                                          final OpticalDuplicateFinder opticalDuplicateFinder,
                                          final int numReducers, final boolean dontMarkUnmappedMates) {
-        SAMFileHeader headerForSort = header.clone();
-        headerForSort.setSortOrder(SAMFileHeader.SortOrder.coordinate);
-        JavaRDD<GATKRead> sortedReadsForMarking = ReadUtils.isReadNameGroupedBam(header)? reads :
-                ReadsSparkSource.putPairsInSamePartition(SparkUtils.sortReads(reads, headerForSort, numReducers),headerForSort,reads.context());//TODO this can and should be switched to a better comparator
+        JavaRDD<GATKRead> sortedReadsForMarking;
+        SAMFileHeader headerForTool = header.clone();
 
-        JavaPairRDD<MarkDuplicatesSparkUtils.IndexPair<String>, Integer> namesOfNonDuplicates = MarkDuplicatesSparkUtils.transformToDuplicateNames(header, scoringStrategy, opticalDuplicateFinder, sortedReadsForMarking, numReducers);
+        // If the input isn't queryname sorted, sort it before duplicate marking
+        if (ReadUtils.isReadNameGroupedBam(header)) {
+            sortedReadsForMarking = reads;
+        } else {
+            headerForTool.setSortOrder(SAMFileHeader.SortOrder.queryname);
+            sortedReadsForMarking = ReadsSparkSource.putPairsInSamePartition(headerForTool, SparkUtils.querynameSortReads(reads, numReducers), new JavaSparkContext(reads.context()));
+        }
+
+        JavaPairRDD<MarkDuplicatesSparkUtils.IndexPair<String>, Integer> namesOfNonDuplicates = MarkDuplicatesSparkUtils.transformToDuplicateNames(headerForTool, scoringStrategy, opticalDuplicateFinder, sortedReadsForMarking, numReducers);
 
         // Here we explicitly repartition the read names of the unmarked reads to match the partitioning of the original bam
         final JavaRDD<Tuple2<String,Integer>> repartitionedReadNames = namesOfNonDuplicates
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSpark.java
@@ -17,6 +17,8 @@ import org.broadinstitute.hellbender.cmdline.argumentcollections.OpticalDuplicat
 import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.spark.GATKSparkTool;
+import org.broadinstitute.hellbender.engine.spark.datasources.ReadsSparkSource;
+import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.utils.Utils;
 import org.broadinstitute.hellbender.utils.read.GATKRead;
 import org.broadinstitute.hellbender.utils.read.ReadUtils;
@@ -24,6 +26,7 @@ import org.broadinstitute.hellbender.utils.read.SAMRecordToGATKReadAdapter;
 import org.broadinstitute.hellbender.utils.read.markduplicates.DuplicationMetrics;
 import org.broadinstitute.hellbender.utils.read.markduplicates.MarkDuplicatesScoringStrategy;
 import org.broadinstitute.hellbender.utils.read.markduplicates.OpticalDuplicateFinder;
+import org.broadinstitute.hellbender.utils.spark.SparkUtils;
 import picard.cmdline.programgroups.ReadDataManipulationProgramGroup;
 import scala.Tuple2;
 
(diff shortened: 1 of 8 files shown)


lbergelson:
Compile warning:

```
:compileTestJava/home/travis/build/broadinstitute/gatk/src/test/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSourceUnitTest.java:339: warning: [static] static method should be qualified by type name, ReadsSparkSource, instead of by an expression
        readsSparkSource.putPairsInSamePartition(header, problemReads, ctx);
```

src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtils.java
```diff
@@ -167,9 +172,7 @@ public int getIndex() {
             keyedReads = spanReadsByKey(indexedReads);
         } else {
             // sort by group and name (incurs a shuffle)
-            JavaPairRDD<String, IndexPair<GATKRead>> keyReadPairs = indexedReads.mapToPair(read -> new Tuple2<>(ReadsKey.keyForRead(
-                    read.getValue()), read));
-            keyedReads = keyReadPairs.groupByKey(numReducers);
+            throw new GATKException("MarkDuplicatesSparkUtils.mark() requires input reads to be queryname sorted, yet the header indicated otherwise");
```
lbergelson: Could you have it print the sort order it thinks its in?

src/test/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtilsUnitTest.java
```diff
@@ -66,4 +67,58 @@ private String getReadGroupId(final SAMFileHeader header, final int index) {
         return new Tuple2<>(key, ImmutableList.copyOf(reads));
     }
 
+    @Test
+    // Test that asserts the duplicate marking is sorting agnostic, specifically this is testing that when reads are scrambled across
+    // partitions in the input that all reads in a group are getting properly duplicate marked together as they are for queryname sorted bams
+    public void testSortOrderParitioningCorrectness() throws IOException {
+
+        JavaSparkContext ctx = SparkContextFactory.getTestSparkContext();
+        JavaRDD<GATKRead> unsortedReads = generateUnsortedReads(10000,3, ctx, 100, true);
```
lbergelson: stupid nitpick: spaces here are wonky, and on the next line

src/test/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtilsUnitTest.java
```diff
@@ -66,4 +67,58 @@ private String getReadGroupId(final SAMFileHeader header, final int index) {
...
+
+        Random rand = new Random(10);
+        for (int i = 0; i < numReadGroups; i++ ) {
+            int start1 = rand.nextInt(SAMRecordSetBuilder.DEFAULT_CHROMOSOME_LENGTH);
+            int start2 = rand.nextInt(SAMRecordSetBuilder.DEFAULT_CHROMOSOME_LENGTH);
+            for (int j = 0; j < numDuplicatesPerGroup; j++) {
+                samRecordSetBuilder.addPair("READ" + readNameCounter++, 0, start1, start2);
+            }
+        }
+        final ReadCoordinateComparator coordinateComparitor = new ReadCoordinateComparator(hg19Header);
```
lbergelson: coordinateComparitor is unused, and misspelled.

src/test/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtilsUnitTest.java
```diff
@@ -66,4 +67,58 @@ private String getReadGroupId(final SAMFileHeader header, final int index) {
...
+        // Comparing the output reads to ensure they are all duplicate marked correctly
+        while (sortedReadsFinal.hasNext()) {
+            GATKRead read1 = sortedReadsFinal.next();
+            GATKRead read2 = unsortedReadsFinal.next();
+            Assert.assertEquals(read1.getName(), read2.getName());
+            Assert.assertEquals(read1.isDuplicate(), read2.isDuplicate());
+        }
+    }
+
+    private JavaRDD<GATKRead> generateUnsortedReads(int numReadGroups, int numDuplicatesPerGroup, JavaSparkContext ctx, int numPartitions, boolean coordinate) {
```
lbergelson: these are sorted...  rename to `generateReadsWithDuplicates` or something like that

lbergelson: could you add a bit of javadoc to this method explaining what it does

src/test/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtilsUnitTest.java
```diff
@@ -66,4 +67,58 @@ private String getReadGroupId(final SAMFileHeader header, final int index) {
...
+        JavaRDD<GATKRead> unsortedReads = generateUnsortedReads(10000,3, ctx, 100, true);
+        JavaRDD<GATKRead> pariedEndsQueryGrouped = generateUnsortedReads(10000,3, ctx,1, false);
+
+        SAMFileHeader unsortedHeader = hg19Header.clone();
+        unsortedHeader.setSortOrder(SAMFileHeader.SortOrder.unsorted);
+        SAMFileHeader sortedHeader = hg19Header.clone();
+        sortedHeader.setSortOrder(SAMFileHeader.SortOrder.queryname);
+
+        // Using the header flagged as unsorted will result in the reads being sorted again
+        JavaRDD<GATKRead> unsortedReadsMarked = MarkDuplicatesSpark.mark(unsortedReads,unsortedHeader, MarkDuplicatesScoringStrategy.SUM_OF_BASE_QUALITIES,new OpticalDuplicateFinder(),100,true);
```
lbergelson: this is called unsorted, but isn't it actually coordinate sorted?

lbergelson: Why the different num reducers?  Is  that to find issues with edge fixing?  If it is, I think you'd be better off with a specific (and possibly similar) test for that.  Since we're always generating pairs, it seems to me that they might never get split across partitions if we're creating an even number of partitions.

src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSpark.java
```diff
@@ -64,22 +67,50 @@
...
                                          final int numReducers, final boolean dontMarkUnmappedMates) {
+        JavaRDD<GATKRead> sortedReadsForMarking;
+        SAMFileHeader headerForTool = header.clone();
+
+        // If the input isn't queryname sorted, sort it before duplicate marking
+        if (ReadUtils.isReadNameGroupedBam(header)) {
+            sortedReadsForMarking = reads;
+        } else {
+            headerForTool.setSortOrder(SAMFileHeader.SortOrder.queryname);
+            sortedReadsForMarking = ReadsSparkSource.putPairsInSamePartition(headerForTool, SparkUtils.querynameSortReads(reads, numReducers), new JavaSparkContext(reads.context()));
```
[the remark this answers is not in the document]
lbergelson: I might extract this whole sorting operation into a function, "queryNameSortReadsIfNecessary"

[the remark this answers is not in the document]
jamesemery: done

src/main/java/org/broadinstitute/hellbender/utils/spark/SparkUtils.java
```diff
@@ -144,6 +144,28 @@ public static boolean pathExists(final JavaSparkContext ctx, final Path targetPa
...
+
+        // do a total sort so that all the reads in partition i are less than those in partition i+1
+        final Comparator<GATKRead> comparator = new ReadQueryNameComparator();
+        final JavaPairRDD<GATKRead, Void> readVoidPairs;
+        if (numReducers > 0) {
+            readVoidPairs = rddReadPairs.sortByKey(comparator, true, numReducers);
+        } else {
+            readVoidPairs = rddReadPairs.sortByKey(comparator);
+        }
+        return readVoidPairs.keys();
```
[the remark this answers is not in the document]
jamesemery: done

src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtils.java
jamesemery: done

src/test/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtilsUnitTest.java
jamesemery: Yeah, it really does have a poor lot in life doesn't it

src/test/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtilsUnitTest.java
jamesemery: done

jamesemery: done

src/test/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtilsUnitTest.java
jamesemery: So the reason for different numbers of partitions was that when I first wrote this test this test there was no exposed way to do the edge fixing for a queryname sorted bam. I didn't want to deal with the  problems of having a mispartitioned bam so I let the queryname sorted reads reside on one partition so the spanning couldn't be wrong. Since this is a test of the coordinate sorted bam marking across partitions and not the edge fixing i'm not worried.

src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSpark.java
```diff
@@ -103,6 +129,21 @@
...
+     * Sort reads into queryname order if they are not already sorted
+     */
+    protected static JavaRDD<GATKRead> querynameSortReadsIfNecessary(JavaRDD<GATKRead> reads, int numReducers, SAMFileHeader headerForTool) {
+        JavaRDD<GATKRead> sortedReadsForMarking;
+        if (ReadUtils.isReadNameGroupedBam(headerForTool)) {
+            sortedReadsForMarking = reads;
+        } else {
+            headerForTool.setSortOrder(SAMFileHeader.SortOrder.queryname);
+            JavaRDD<GATKRead> sortedReads = SparkUtils.querynameSortReads(reads, numReducers);
+            sortedReadsForMarking = ReadsSparkSource.putPairsInSamePartition(headerForTool, sortedReads, new JavaSparkContext(reads.context()));
```
lbergelson: @jamesemery I would use JavaSparkContext.fromSparkContext instead of new JavaSparkContext.  I think it's the same, but it might change in future sparks.

Issue #4586, opened 2018-03-26
Funcotator - Need Performance Upgrades


jonn-smith:
Funcotator can be slow.

Need to profile it and make it fast all the time.

droazen:
A large speedup has already been achieved in https://github.com/broadinstitute/gatk/pull/4740. According to @lbergelson:

```
Original (46a8661): 31.1 minutes, 1658 variants / minute
Changes (#4740): 1.26 minutes, 40933 variants / minute

~24.6x speedup
```

jonn-smith:
We should investigate tuning the caching parameter to increase speed further.

We still need to profile, increase speed further, and compare to Oncotator.

jonn-smith:
We also need to profile on multiple machines and verify that `Funcotator` will run at the same speed on all of them.

jonn-smith:
Turns out the caching fix is only a partial fix because of the discrepancies between b37 contig names, hg19 contig names, and the fact that the datasources are inconsistent between them.

PR #4740, opened 2018-05-07
making optimizations to funcotator

lbergelson:
* improving slow String handling
* enabling feature caching

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
+++ b/src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
@@ -805,18 +805,24 @@ public abstract class GATKTool extends CommandLineProgram {
     /**
      * A method to allow a user to inject data sources after initialization that were not specified as command-line
      * arguments.
+     *
+     * @param filePath path to the Feature file to register
+     * @param name what to call the Feature input
+     * @param featureType class of features
+     * @param featureQueryLookahead look ahead this many bases during queries that produce cache misses
      * @return The {@link FeatureInput} used as the key for this data source.
      */
     protected FeatureInput<? extends Feature> addFeatureInputsAfterInitialization(final String filePath,
                                                                                   final String name,
-                                                                                  final Class<? extends Feature> featureType) {
+                                                                                  final Class<? extends Feature> featureType,
+                                                                                  final int featureQueryLookahead) {
 
-        final FeatureInput<? extends Feature> featureInput = new FeatureInput<>(name + FeatureInput.FEATURE_ARGUMENT_TAG_DELIMITER + filePath);
+        final FeatureInput<? extends Feature> featureInput = new FeatureInput<>(filePath, name);
 
         //Add datasource to the feature manager too so that it can be queried. Setting lookahead to 0 to avoid caching.
         //Note: we are disabling lookahead here because of windowed queries that need to "look behind" as well.
         features.addToFeatureSources(
-                0,
+                featureQueryLookahead,
                 featureInput,
                 featureType,
                 cloudPrefetchBuffer,
--- a/src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
@@ -211,7 +211,7 @@ public class Funcotator extends VariantWalker {
     /**
      * The current version of {@link Funcotator}.
      */
-    public static final String VERSION = "0.0.2";
+    public static final String VERSION = "0.0.3";
 
     //==================================================================================================================
     // Arguments:
@@ -568,9 +568,9 @@ public class Funcotator extends VariantWalker {
         }
     }
 
-    private void addFeaturesForLocatableDataSource( final Path dataSourceFile,
-                                                    final Properties dataSourceProperties,
-                                                    final Class<? extends Feature> featureClazz ) {
+    private void addFeaturesForLocatableDataSource(final Path dataSourceFile,
+                                                   final Properties dataSourceProperties,
+                                                   final Class<? extends Feature> featureClazz) {
 
         final String name = dataSourceProperties.getProperty(DataSourceUtils.CONFIG_FILE_FIELD_NAME_NAME);
 
@@ -580,7 +580,8 @@ public class Funcotator extends VariantWalker {
                         IOUtils.getPath( dataSourceProperties.getProperty(DataSourceUtils.CONFIG_FILE_FIELD_NAME_SRC_FILE) )
                 ).toUri().toString(),
                 name,
-                featureClazz
+                featureClazz,
+                VariantWalkerBase.FEATURE_CACHE_LOOKAHEAD
         );
 
         // Add our feature input to our list of manual inputs:
--- a/src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfCodec.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfCodec.java
@@ -139,16 +139,9 @@ final public class GencodeGtfCodec extends AbstractFeatureCodec<GencodeGtfFeatur
         return new FeatureCodecHeader(readActualHeader(lineIterator), FeatureCodecHeader.NO_HEADER_END);
     }
 
-    @SuppressWarnings( "deprecation" )
     @Override
     public LocationAware makeIndexableSourceFromStream(final InputStream bufferedInputStream) {
-        final PositionalBufferedStream pbs;
-        if (bufferedInputStream instanceof PositionalBufferedStream) {
-            pbs = (PositionalBufferedStream) bufferedInputStream;
-        } else {
-            pbs = new PositionalBufferedStream(bufferedInputStream);
-        }
-        return new AsciiLineReaderIterator(new AsciiLineReader(pbs));
+        return new AsciiLineReaderIterator(AsciiLineReader.from(bufferedInputStream));
     }
 
     // ============================================================================================================
--- a/src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
@@ -4,6 +4,7 @@ import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.util.Locatable;
 import htsjdk.tribble.Feature;
 import htsjdk.tribble.annotation.Strand;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.hellbender.exceptions.UserException;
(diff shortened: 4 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -809,14 +809,15 @@ private String createProgramGroupID(final SAMFileHeader header) {
      */
     protected FeatureInput<? extends Feature> addFeatureInputsAfterInitialization(final String filePath,
                                                                                   final String name,
-                                                                                  final Class<? extends Feature> featureType) {
+                                                                                  final Class<? extends Feature> featureType,
+                                                                                  final int featureQueryLookahead) {
```
droazen: Add javadoc for method args

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
```diff
@@ -108,23 +109,24 @@ protected GencodeGtfFeature(final String[] gtfFields) {
...
-            // Each optional field is in a key/value pair:
-            final String fieldName = fieldParts[EXTRA_FIELD_KEY_INDEX].trim();
-
-            // The value of the field may be between two quotes.
-            // We remove them here.
-            final String fieldValue = fieldParts[EXTRA_FIELD_VALUE_INDEX].trim().replaceAll("\"", "");
+            final int splitPoint = trimmedExtraField.indexOf(EXTRA_FIELD_KEY_VALUE_SPLITTER);
+            final String fieldName;
+            final String fieldValue;
+            if(splitPoint == -1 ) {
```
pshapiro4broad: need a space after `if`

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -212,6 +212,7 @@
      * The current version of {@link Funcotator}.
      */
     public static final String VERSION = "0.0.2";
+    public static final int FEATURE_QUERY_LOOKAHEAD = 100000;
```
droazen: Do you want to reference one of the other predefined constants here?

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
```diff
@@ -108,23 +109,24 @@ protected GencodeGtfFeature(final String[] gtfFields) {
...
+                throw new UserException.MalformedFile("Extraneous optional field data - not in a key/value pair: " + extraField);
+            } else {
```
pshapiro4broad: `else` is unnecessary here since the `if` body throws. Once you get rid of `else` you can move the declarations for `fieldName` and `fieldValue` down here where they're assigned

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
```diff
@@ -108,23 +109,24 @@ protected GencodeGtfFeature(final String[] gtfFields) {
...
+                fieldName = trimmedExtraField.substring(0,splitPoint).trim();
```
pshapiro4broad: need space after `,`

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -580,7 +581,8 @@ private void addFeaturesForLocatableDataSource( final Path dataSourceFile,
                         IOUtils.getPath( dataSourceProperties.getProperty(DataSourceUtils.CONFIG_FILE_FIELD_NAME_SRC_FILE) )
                 ).toUri().toString(),
                 name,
-                featureClazz
+                featureClazz,
+                FEATURE_QUERY_LOOKAHEAD
```
droazen: Can you confirm that Funcotator never "looks back" when querying its `FeatureContext`? If it does, the existing caching scheme might be non-ideal.

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfCodec.java
```diff
@@ -139,16 +139,9 @@ public FeatureCodecHeader readHeader(final LineIterator lineIterator) throws IOE
...
     @Override
     public LocationAware makeIndexableSourceFromStream(final InputStream bufferedInputStream) {
-        final PositionalBufferedStream pbs;
-        if (bufferedInputStream instanceof PositionalBufferedStream) {
-            pbs = (PositionalBufferedStream) bufferedInputStream;
-        } else {
-            pbs = new PositionalBufferedStream(bufferedInputStream);
-        }
-        return new AsciiLineReaderIterator(new AsciiLineReader(pbs));
+        return new AsciiLineReaderIterator(AsciiLineReader.from(bufferedInputStream));
```
droazen: Is this method guaranteed to wrap the stream in a `PositionalBufferedStream`?

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
```diff
@@ -108,23 +109,24 @@ protected GencodeGtfFeature(final String[] gtfFields) {
...
+
+                // The value of the field may be between two quotes.
+                // We remove them here.
+                final String rawFieldValue = trimmedExtraField.substring(splitPoint + 1, trimmedExtraField.length());
```
droazen: If there's more than 1 occurrence of the delimiter, the extra tokens will now all be concatenated together in `rawFieldValue`, whereas before they would be ignored. What's the desired behavior here? Should multiple delimiters be an error condition?

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
```diff
@@ -108,23 +109,24 @@ protected GencodeGtfFeature(final String[] gtfFields) {
...
+                fieldValue = StringUtils.remove(rawFieldValue.trim(), '"');
```
droazen: Can we assume that the quotes are always at the beginning/end? If yes, we could make this even faster using a `substring()` call.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
```diff
@@ -211,6 +211,20 @@ public void metaTestEnsureTempDirs() {
         Assert.assertEquals(doDebugTests, false);
     }
 
+    @Test
+    public void testLocalThing(){
```
droazen: Need a better name...

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
droazen: Also, do you want a different value for Gencode vs. other feature inputs? Should we expose a command-line arg at the `GATKTool` and/or `VariantWalker` level?

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
lbergelson: I think we might want different values for gencode... I'm not sure though.  I think maybe we can leave that decision to the future?

lbergelson: I think a good option would be to put the look ahead value as a datasource property, that way each one could customize it.

lbergelson: @jonn-smith Any thoughts on that?

droazen:
@lbergelson Can you add a summary of the performance improvements introduced in this branch as a comment to the master Funcotator performance ticket (https://github.com/broadinstitute/gatk/issues/4586)

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
lbergelson: @jonn-smith Can you answer the case of what should happen if there are multiple delimiters here? Previous behavior was that everything after the second element in the split array would be ignored.  It seems like maybe it should be an error condition to have >2 elements though?

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
lbergelson: It looks like it only ever calls it on the same location as the driving variant.  It does some contig translation, but it should always translate the same way so I don't think that's a problem.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
lbergelson: For now I've set it to use the same value of 100_000, but I'm referring to the value in `VariantWalkerBase`.  It looks like all variant walker features, both driving and side inputs are set to 100k.

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfCodec.java
lbergelson: Yes

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
lbergelson: good point. done

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
lbergelson: added

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
lbergelson: I'm assuming that's an error case and added a check for now

lbergelson:
@droazen Running a larger sample to have a better number.

lbergelson:
Original (46a86618d9dd29eb05a8aa222fab7a4d3673a547): 31.1 minutes, 1658 variants / minute
Changes  (https://github.com/broadinstitute/gatk/pull/4740): 1.26 minutes,  40933 variants / minute

~24.6x speedup

lbergelson:
@droazen Any further comments?

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
jonn-smith: I think that's fine for now.  I don't remember if I ran into an issue that required that I ignore the second delimiter on.

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
jonn-smith: @lbergelson - I think we can assume that quotes would wrap the strings.

That said, we will need to check to see if quotes are in the string before blindly substringing them.

jonn-smith:
@lbergelson - might make sense to update the Funcotator version from `0.0.2` to `0.0.3` since this is a big improvement.

src/main/java/org/broadinstitute/hellbender/utils/codecs/gencode/GencodeGtfFeature.java
```diff
@@ -108,23 +109,26 @@ protected GencodeGtfFeature(final String[] gtfFields) {
...
-                    throw new UserException.MalformedFile("Extraneous optional field data - not in a key/value pair: " + extraField);
-                }
+            final int splitPoint = trimmedExtraField.indexOf(EXTRA_FIELD_KEY_VALUE_SPLITTER);
+            if( splitPoint == -1 ) {
+                throw new UserException.MalformedFile("Extraneous optional field data - not in a key/value pair: " + extraField);
             }
 
-            // Each optional field is in a key/value pair:
-            final String fieldName = fieldParts[EXTRA_FIELD_KEY_INDEX].trim();
+            final String fieldName = trimmedExtraField.substring(0, splitPoint).trim();
```
droazen: Couldn't you end up with an empty `fieldName`/`fieldValue` after parsing? Should that be an error condition?

droazen: After discussing in person with @lbergelson, we decided that we don't know enough about the data to add extra safety checks here. If leading whitespace is typical/common rather than an error condition, we don't want to overly-strict extra checks for empty keys.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
lbergelson: @jonn-smith I'm not sure I understand what you mean by making it a parameter on the datasource type.  Would we have a different parameter for each datasource type?  I'm going to merge this now and we can revisit if we need more granular control I think.

PR #4751, opened 2018-05-09
Sf mito fix

lucidtronix:
Small bugfix to prevent crashes on the ends of the mitochondrial contig.

### DIFF
--- a/src/main/python/org/broadinstitute/hellbender/vqsr_cnn/vqsr_cnn/inference.py
+++ b/src/main/python/org/broadinstitute/hellbender/vqsr_cnn/vqsr_cnn/inference.py
@@ -103,6 +103,8 @@ def reference_string_to_tensor(reference):
             dna_data[i, defines.DNA_SYMBOLS[b]] = 1.0
         elif b in defines.AMBIGUITY_CODES:
             dna_data[i] = defines.AMBIGUITY_CODES[b]
+        elif b == '\x00':
+            break
         else:
             raise ValueError('Error! Unknown code:', b)
 
--- a/src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariantsIntegrationTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariantsIntegrationTest.java
@@ -88,15 +88,25 @@ public class CNNScoreVariantsIntegrationTest extends CommandLineProgramTest {
         final ArgumentsBuilder argsBuilder = new ArgumentsBuilder();
         argsBuilder.addArgument(StandardArgumentDefinitions.VARIANT_LONG_NAME, inputVCF)
                 .addArgument(StandardArgumentDefinitions.OUTPUT_LONG_NAME, "%s")
-                .addArgument(StandardArgumentDefinitions.REFERENCE_LONG_NAME, b37_reference_20_21)
-                .addArgument("architecture", architecture1D)
-                .addArgument("inference-batch-size", "8")
-                .addArgument("transfer-batch-size", "16")
+                .addArgument(StandardArgumentDefinitions.REFERENCE_LONG_NAME, hg19MiniReference)
                 .addArgument(StandardArgumentDefinitions.ADD_OUTPUT_VCF_COMMANDLINE, "false");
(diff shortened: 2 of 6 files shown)


src/main/python/org/broadinstitute/hellbender/vqsr_cnn/vqsr_cnn/inference.py
```diff
@@ -103,6 +103,8 @@ def reference_string_to_tensor(reference):
             dna_data[i, defines.DNA_SYMBOLS[b]] = 1.0
         elif b in defines.AMBIGUITY_CODES:
             dna_data[i] = defines.AMBIGUITY_CODES[b]
+        elif b == '\x00':
```
cmnbroad: @lucidtronix - question on this - it seems like this issue (zeros in the ref bases) can happen whenever the position of a variant is within windowsize/2 of the end of the reference contig, since thats what causes the tool to pad out the ref base string. This makes me wonder if the reference tensor is being constructed correctly when a SNP occurs near the edges of a contig - normally the reference base that corresponds to the variant base in question is in the middle of the read tensor, and padded out on each side. But not so at the edges, where it will be offset. Is that expected, or should we be padding out from the middle on both sides of the base ? I think the answer to that will determine whether its ok to just break out at the first 0.

Also, if this really isn't mito-specific, I don't think we need to add a mito reference for the test case - we might be able to get away with creating a test vcf that uses an existing reference, as long as it has a variant near the edge of a contig.

lucidtronix: Yes, you're right variants at the very beginning (i.e. within 64bp of the start) of contigs are not correctly constructed.  Since these variants are probably only going to show up in humans in the mitochondria, their CNN scores are pretty meaningless anyway.  After chatting with @ldgauthier we think it's best to get this fix in now to prevent crashes, and then revisit if necessary when mitochondrial best practices are ready.  

I removed the mitochondrial reference files and updated the test.  Back to you @cmnbroad

src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariantsIntegrationTest.java
```diff
@@ -100,11 +100,11 @@ public void testSmallBatchInference()throws IOException {
     }
 
     @Test(groups = {"python"})
-    public void testOnMitochondria() throws IOException{
-        final String mitoVcf = largeFileTestDir + "VQSR/errorM.vcf";
+    public void testOnContigEdge() throws IOException{
+        final String edgeVcf = largeFileTestDir + "VQSR/errorM.vcf";
```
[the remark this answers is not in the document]
lucidtronix: Moved and renamed test file, rebased and squashed if tests pass good to merge?

src/test/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariantsIntegrationTest.java
```diff
@@ -99,6 +99,19 @@ public void testSmallBatchInference()throws IOException {
         spec.executeTest("testInference", this);
     }
 
+    @Test(groups = {"python"})
+    public void testOnContigEdge() throws IOException{
+        final String edgeVcf = toolsTestDir + "walkers/VQSR/variantNearContigEdge.vcf";
+        final ArgumentsBuilder argsBuilder = new ArgumentsBuilder();
+        argsBuilder.addArgument(StandardArgumentDefinitions.VARIANT_LONG_NAME, edgeVcf)
+                .addArgument(StandardArgumentDefinitions.REFERENCE_LONG_NAME, hg19MiniReference)
+                .addArgument("architecture", architecture1D)
```
cmnbroad: @lucidtronix I just noticed this test is using 1d, which doesn't exercise the changed code path.

lucidtronix: I think the users who encountered this error were running the 1D model, but either way both models will call the reference_string_to_tensor function in inference.py where the null character check was added.

PR #4756, opened 2018-05-09
Added a new WEX test bam to src/test/resources/large, with a companion target interval list

droazen:
This is a ~130 MB snippet of the NexPond-377866 NA12878 bam from
https://github.com/broadinstitute/palantir/wiki/Gold-Standard-Datasets
with:

-all reads from chromosomes 20 and 21
-50,000 unmapped reads added to the end

It can be used with the existing human_g1k_v37.20.21.fasta reference in
src/test/resources/large

Tests that use this new file will be coming in a future branch.


lbergelson:
@droazen one thing I just noticed, the header of this includes the full hg19 dict, do we want to replace it with just 20/21?

Issue #4696, opened 2018-04-24
[CNNScoreVariants] A timeout ocurred waiting for output from the remote Python command


heliac2000:
In GATK 4.0.3, add `tensorflow-gpu==1.4.0` to `gatkcondaenv.yml `and create a conda environment.

    $ conda env create -n gatk_gpu -f gatkcondaenv.yml

Activate this environment and execute GATK CNNScoreVariants(1D model).

    $ gatk-4.0.3.0/gatk --java-options '-Xmx8G -DGATK_STACKTRACE_ON_USER_EXCEPTION=true' CNNScoreVariants -V INPUT_VCF -R REF -O OUTPUT_VCF

Then, GATK outputs the following:

```
18:30:57.468 INFO  CNNScoreVariants - Initializing engine
18:30:57.985 INFO  FeatureManager - Using codec VCFCodec to read file ...
18:30:58.183 INFO  IntervalArgumentCollection - Processing 48129895 bp from intervals
18:30:58.188 INFO  CNNScoreVariants - Done initializing engine
18:31:00.188 INFO  CNNScoreVariants - Using key:CNN_1D for CNN architecture:.../1d_cnn_mix_train_full_bn.2560984151625233621.json and weights:.../1d_cnn_mix_train_full_bn.2397909300265264152.hd5
18:31:19.873 INFO  ProgressMeter - Starting traversal
18:31:19.874 INFO  ProgressMeter -        Current Locus  Elapsed Minutes    Variants Processed  Variants/Minute
18:31:50.095 INFO  CNNScoreVariants - Shutting down engine
[2018/04/24 18:31:50 JST] org.broadinstitute.hellbender.tools.walkers.vqsr.CNNScoreVariants done. Elapsed time: 0.88 minutes.
Runtime.totalMemory()=2141716480
***********************************************************************
A USER ERROR has occurred: A timeout ocurred waiting for output from the remote Python command.
***********************************************************************
org.broadinstitute.hellbender.exceptions.UserException: A timeout ocurred waiting for output from the remote Python command.
        at org.broadinstitute.hellbender.utils.python.StreamingPythonScriptExecutor.getAccumulatedOutput(StreamingPythonScriptExecutor.java:219)
        at org.broadinstitute.hellbender.utils.python.StreamingPythonScriptExecutor.sendSynchronousCommand(StreamingPythonScriptExecutor.java:153)
[...]

Issue #4444, opened 2018-02-22
StreamingPythonScriptExecutor: explore use of a 2nd FIFO for synchronization (instead of relying on the Python prompt)


droazen:
Use of the second FIFO should be encapsulated behind an easy-to-use API on the Python side, perhaps with methods such as "writeBatchOutput" that would both write a batch's worth of output and signal to Java that the batch is done using the second FIFO.

cmnbroad:
Closed via https://github.com/broadinstitute/gatk/pull/4757.

Issue #4443, opened 2018-02-22
StreamingPythonScriptExecutor: encapsulate handshake process within a single method (for both Python and Java)


droazen:

cmnbroad:
Closed via https://github.com/broadinstitute/gatk/pull/4757.

Issue #4342, opened 2018-02-05
Refine and finalize StreamingPythonScriptExecutor design


droazen:
Let's have another look at the `StreamingPythonScriptExecutor` and evaluate whether we are using the best synchronization and inter-process communication primitives, devise ways to reduce the amount of boilerplate code on the Python side, and generally ask whether any improvements could be made to the current design.

droazen:
For @cmnbroad and @jonn-smith

PR #4757, opened 2018-05-10
Replace StreamingPythonExecutor prompt synchronization with ack FIFO and remove timeouts.

cmnbroad:
- Fixes https://github.com/broadinstitute/gatk/issues/4696, https://github.com/broadinstitute/gatk/issues/4342, https://github.com/broadinstitute/gatk/issues/4443, https://github.com/broadinstitute/gatk/issues/4444.
- Use a second FIFO for command acknowledgement instead of relying on prompt synchronization.
- Add a Python module for managing the Python side of GATK/Python interaction.
- Removed all timeouts.
- Install a Python exception handler for handling uncaught Python exceptions.
- Update CNNScoreVariants to use the new protocol.


cmnbroad:
@lucidtronix Specifically it would be useful to try this with various transfer/inference batch sizes now that the timeouts are gone. I tried with the original defaults (256/128), and got no timeouts, but the larger values seemed to result in longer runtimes.

cmnbroad:
Note to self: we should get rid of the MAX_READ_BATCH check as part of this, since that was motivated by timeouts IIRC.

src/main/python/org/broadinstitute/hellbender/gatktool/tool.py
```diff
@@ -0,0 +1,175 @@
...
+Most of the functions in this module are intended to be called by GATK via
+the StreamingPythonScriptExecutor Java class, and are not called by Python
+code directly. The one exception is the readDataFIFO function, which can be
+used to read data that had been passed to Python by GATK Java code.
+"""
+
+import sys
+import os
+
+_ackFIFO = None
```
jonn-smith: If you refactored this as another Python class, you could get rid of these globals.  It might make it a little cleaner.

src/main/python/org/broadinstitute/hellbender/gatktool/tool.py
```diff
@@ -0,0 +1,175 @@
...
+    global _ackFIFO
+    _ackFIFO.writeNack()
+
+
+def terminateGATK():
+    """
+    Called by GATK when no more Python commands will be executed
+    """
+    global _ackFIFO
+    assert _ackFIFO != None
```
jonn-smith: Should this be an exception, rather than an assert?

src/main/python/org/broadinstitute/hellbender/gatktool/tool.py
```diff
@@ -0,0 +1,175 @@
...
+    _dataFIFO = DataFIFO(dataFIFOName)
+
+
+def closeDataFIFO():
+    """"
+    Close the data FIFO. Afteer this method hass been called, the
+    data FIFO can no longer be read.
+    """
+    global _dataFIFO
+    assert _dataFIFO != None
```
jonn-smith: Same assert question as above.

src/main/python/org/broadinstitute/hellbender/gatktool/tool.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class DataFIFO:
+    """
+    Manage the FIFO stream used for transferring data from the GATK tool to
+    Python code.
+
+    The FIFO is written by GATK and read by Python.
+    """
+    # TODO: Need a way to provide a per-line deserializer
```
jonn-smith: Does this mean that each line is a new command?

src/test/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessControllerUnitTest.java
```diff
@@ -23,163 +19,75 @@
...
-    public void testSerialCommands() throws TimeoutException, IOException {
-        // start an interactive Python session with unbuffered IO
-        final ProcessSettings processSettings = new ProcessSettings(new String[] {"python", "-i", "-u"});
-
-        final StreamingProcessController controller = new StreamingProcessController(processSettings, PYTHON_PROMPT, true);
-        Assert.assertTrue(controller.start());
-
-        // consume the Python startup banner debris, but don't try validate it
-        controller.getProcessOutputByPrompt();
+    public void testSerialCommands() throws IOException {
```
jonn-smith: Just for completeness, it might be good to have a few more test cases here.  You could just make a data provider with 5 or so simple chained expressions.

src/test/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessControllerUnitTest.java
```diff
@@ -190,13 +98,13 @@ public void testStderrRedirect() throws TimeoutException {
     }
 
     @Test(timeOut = 10000)
-    public void testFIFOLifetime() throws TimeoutException {
+    public void testFIFOLifetime() {
```
jonn-smith: Does this test really test the lifetime of a fifo?  It looks like it just makes sure that it's opened and closed properly.

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
```diff
@@ -107,48 +90,44 @@ public void apply(GATKRead read, ReferenceContext referenceContext, FeatureConte
...
-        // to the FIFO. Sending a *SYNCHRONOUS* command here would immediately block the background thread
-        // since this statement will be executed BEFORE any data from the batch is written to the stream.
-        pythonExecutor.sendAsynchronousCommand(String.format(
-                "for i in range(%s):\n    tempFile.write(fifoFile.readline())" + NL + NL, batchCount));
-        asyncWriter.startAsynchronousBatchWrite(batchList);
+        // to the FIFO through the stream writer. Sending a *SYNCHRONOUS* command would block indefinitely
+        // waiting for the ack to be received, but it never would be because no data has been written to the
+        // stream yet.
+        pythonExecutor.sendAsynchronousCommand(
+                String.format("for i in range(%s):\n    tempFile.write(tool.readDataFIFO())" + NL + NL, batchCount)
```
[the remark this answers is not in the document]
cmnbroad: The embedded Python code here is tool-specific, so I'm not sure how to encapsulate it more. Somewhere, a Python tool has to have Python code that reads from the FIFO (by calling tool.readDataFIFO), and does something with the result. In this example tool, I could put that call inside a tool-specific function in Python, and then embed a call to that function here (thats what the CNN tool does). But for illustration purposes, I think having the Python source embedded in the Java source allows you to see the entire lifecycle here, all in one place. Let me know if you have a different take on that.

src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/CNNScoreVariants.java
```diff
@@ -17,14 +17,13 @@
...
-import org.broadinstitute.hellbender.utils.runtime.AsynchronousStreamWriterService;
+import org.broadinstitute.hellbender.utils.runtime.AsynchronousStreamWriter;
 import org.broadinstitute.hellbender.utils.variant.GATKVCFConstants;
 import org.broadinstitute.hellbender.utils.variant.GATKVCFHeaderLines;
 import picard.cmdline.programgroups.VariantEvaluationProgramGroup;
 
 import java.io.*;
 import java.util.*;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.StreamSupport;
```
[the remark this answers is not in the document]
cmnbroad: It might if the tool class hierarchy weren't partitioned by traversal type, but given that it is, I don't see any clean way to do it in a traversal-independent way.

src/main/java/org/broadinstitute/hellbender/utils/python/StreamingPythonScriptExecutor.java
```diff
@@ -171,110 +180,134 @@ public void sendAsynchronousCommand(final String line) {
...
-     * See if any output is currently available. This is non-blocking, and can be used to determine if a blocking
-     * call can be made; it is always safe to call getAccumulatedOutput if isOutputAvailable is true.
-     * @return true if data is available from the remote process.
+     * Wait for an acknowledgement (which must have been previously requested).
+     * @return true if a positive acknowledgement (ack) is recieved, false if negative (nck)
      */
-    public boolean isOutputAvailable() {
-        return spController.isOutputAvailable();
+    public ProcessOutput waitForAck() {
+        if (!isAckOutstanding) {
```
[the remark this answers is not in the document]
cmnbroad: Good catch. Its the exception message that needs to change. A precondition for calling this method (to wait for an ack) is that an ack request must have been previously sent.

Having said that, I do think the variable name should also be changed to `isAckRequestOutstanding`.

src/main/java/org/broadinstitute/hellbender/utils/python/StreamingPythonScriptExecutor.java
```diff
@@ -171,110 +180,134 @@ public void sendAsynchronousCommand(final String line) {
...
-            throw new UserException("A timeout ocurred waiting for output from the remote Python command.", e);
-        }
+    public String getApproximateCommandLine() {
+        return curatedCommandLineArgs.stream().collect(Collectors.joining(" "));
     }
 
     /**
-     * Terminate the remote process, closing the fifo if any.
+     *
+     * @param itemSerializer
```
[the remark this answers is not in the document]
cmnbroad: Yeah, done.

src/main/java/org/broadinstitute/hellbender/utils/python/StreamingPythonScriptExecutor.java
```diff
@@ -171,110 +180,134 @@ public void sendAsynchronousCommand(final String line) {
...
 
     /**
-     * Return a (not necessarily executable) string representing the current command line for this executor
-     * for error reporting purposes.
-     * @return A string representing the command line used for this executor.
+     * Return all data accumulated since the last call to {@link #getAccumulatedOutput} (either directly, or
+     * indirectly through {@link #sendSynchronousCommand}.
+     *
+     * Note that the output returned is somewhat non-deterministic, in that there is no guaranty that all of
+     * the output from th prevous command has been flushed at the time this call is made.
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/AsynchronousStreamWriter.java
```diff
@@ -76,28 +76,24 @@ public void startAsynchronousBatchWrite(final List<T> batchList) {
...
                 } catch (ExecutionException | InterruptedException e) {
-                    throw new GATKException("Interrupted during background stream write");
-                } catch (TimeoutException e) {
-                    throw new GATKException("Timeout waiting for background stream write to complete");
+                    throw new GATKException("Interrupted during background stream write", e);
                 }
                 streamWriter.flush();
             } catch (IOException e) {
-                throw new GATKException("IOException waiting for asynchrouns batch completion", e);
+                throw new GATKException("IOException waiting for asynchrnous batch completion", e);
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/ProcessOutput.java
```diff
@@ -30,4 +30,23 @@ public StreamOutput getStderr() {
         return stderr;
     }
 
+    public String toString() {
```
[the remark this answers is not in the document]
cmnbroad: This whole method is no longer used so I'm removing it.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -8,75 +8,68 @@
...
+ * detected by {@link #waitForAck}. Creates separate threads for reading stdout and stderr and provides access
+ * to stdout and stderr as strings.
  */
 public final class StreamingProcessController extends ProcessControllerBase<CapturedStreamOutputSnapshot> {
     private static final Logger logger = LogManager.getLogger(StreamingProcessController.class);
 
     private final ProcessSettings settings;
-    private final String promptForSynchronization;
+
+    private static String ACK_FIFO_FILE_NAME = "gatkStreamingControllerAck.fifo";
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -243,92 +160,81 @@ private StreamOutput getOutputOptimistic(final Future<CapturedStreamOutputSnapsh
...
+     */
+    public boolean waitForAck() {
+        if (ackFuture != null) {
+            throw new GATKException("An ack is already outstanding");
+        }
+        ackFuture = executorService.submit(
+                () -> {
+                    try {
+                        byte[] ack = new byte[ACK_MESSAGE_SIZE];
+                        int nBytes = ackFIFOInputStream.read(ack, 0, ACK_MESSAGE_SIZE);
```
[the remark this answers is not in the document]
cmnbroad: I didn't see any reason to use a delimiter so I didn't - its just string of `ack' or `nck`.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -337,13 +243,16 @@ private ProcessOutput getProcessOutput() throws TimeoutException {
...
      * @return a FIFO File
      */
-    public File createFIFO() {
-        if (fifoTempDir != null || fifoFile != null) {
-            throw new IllegalArgumentException("Only one FIFO per controller is supported");
+    public File createDataFIFO() {
+        if (dataFIFOFile != null) {
+            throw new IllegalArgumentException("Only one data FIFO per controller is supported");
         }
+        dataFIFOFile = createFIFOFile(DATA_FIFO_FILE_NAME);;
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessController.java
```diff
@@ -526,7 +397,7 @@ public void enable(final String commandString) {
                 journalingFileWriter.write("Initial process command line: ");
                 journalingFileWriter.write(settings.getCommandString() + "\n\n");
             } catch (IOException e) {
-                throw new GATKException(String.format("Error creating streaming process journaling file %s for command \"%s\"",
+                throw new GATKException(String.format("Error creating straeming process journaling file %s for command \"%s\"",
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/python/org/broadinstitute/hellbender/gatktool/tool.py
cmnbroad: These could be removed and instead created by the having the executor create them, but at least this way they explicitly appear in the Python code. Per discussion yesterday I'll leave these as is.

src/main/python/org/broadinstitute/hellbender/gatktool/tool.py
cmnbroad: Yes, I originally did this to keep myself honest, but I think all of the asserts in this file should raise exceptions instead. Done.

src/main/python/org/broadinstitute/hellbender/gatktool/tool.py
cmnbroad: Done.

src/main/python/org/broadinstitute/hellbender/gatktool/tool.py
cmnbroad: This was just a reminder for myself that ultimately we'll probably want a way to allow heterogenous data types to be passed through the FIFO (i.e., a variant followed by one or more reads), and if we used something structured like protobufs, we'll need some way to handle them on both sides (Java and Python). But for now this comment can be removed. Done.

src/test/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessControllerUnitTest.java
cmnbroad: The FIFO's lifetime is dictated by the lifetime of the controller; this test validates that bye checking that the FIFO doesn't survive termination of the controller. Maybe there is a better name ?

src/test/java/org/broadinstitute/hellbender/utils/runtime/StreamingProcessControllerUnitTest.java
cmnbroad: Sure, done.

src/main/python/org/broadinstitute/hellbender/gatktool/tool.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    sendNack()
+    sys.__excepthook__(exceptionType, value, traceback)
+
+
+def sendAck():
+    """
+    Send a positive acknowledgment to GATK. This should generally only
+    be called by python code that is embedded in Java, since the executor
+    keeps track of whether an ack reuest is outstanding.
```
droazen: reuest -> request

src/main/java/org/broadinstitute/hellbender/tools/examples/ExampleStreamingPythonExecutor.java
```diff
@@ -57,44 +55,29 @@
...
-        pythonExecutor.sendAsynchronousCommand(String.format("fifoFile = open('%s', 'r')" + NL, fifoFile.getAbsolutePath()));
-        try {
-            fifoWriter = new FileOutputStream(fifoFile);
-            asyncWriter = pythonExecutor.getAsynchronousStreamWriterService(fifoWriter, AsynchronousStreamWriterService.stringSerializer);
-        } catch ( IOException e ) {
-            throw new GATKException("Failure opening FIFO for writing", e);
-        }
-        // synchronize on the output prompt before executing the next statement
-        pythonExecutor.getAccumulatedOutput();
+        asyncStreamWriter = pythonExecutor.getStreamWriter(AsynchronousStreamWriter.stringSerializer);
```
droazen: Encapsulate the writer within the executor, as discussed in person.

jonn-smith:
@cmnbroad back to you for conflict resolution, then good to go by me

PR #4764, opened 2018-05-11
Added --sites-only argument to GATK engine

### LINKED ISSUE #4763: Resurrect --sites-only engine arg for VCF output
We've had a request to restore the `--sites-only` engine argument that was present in GATK3. This should be trivial: add the argument to `GATKTool`, have `GATKTool.createVCFWriter()` check the argument and add `Options.DO_NOT_WRITE_GENOTYPES` if necessary.

jamesemery:
Fixes #4763


jamesemery:
@ldgauthier This is what you requested correct?

ldgauthier:
@jamesemery Those of us that work with thousands of samples thank you.
If you could change "GT" to "genotype" in the arg description then this can be good to go.  (Technically the GT field is part of the genotype field, along with AD, DP and all our other friends.  Obviously it would be silly to output those without GT, but I aim for semantic correctness.)

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -113,6 +113,10 @@
             optional = true)
     public boolean disableBamIndexCaching = false;
 
+    @Argument(fullName = StandardArgumentDefinitions.SITES_ONLY_LONG_NAME,
+            doc = "If true, don't emit GT fields when writing vcf file output.", optional = true)
+    public boolean sitesOnlyMode = false;
```
droazen: Name of field should make it clear that this applies to VCFs specifically -- eg., `outputSitesOnlyVCFs` instead of `sitesOnlyMode`

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -728,6 +732,10 @@ public VariantContextWriter createVCFWriter(final File outFile) {
             }
         }
 
+        if (sitesOnlyMode) {
+            options.add(Options.DO_NOT_WRITE_GENOTYPES);
```
droazen: Can you check to see whether there are any other `GATKTools` that bypass `GATKTool.createVCFWriter()` when writing a VCF and call `GATKVariantContextUtils.createVCFWriter()` directly instead? (besides the ones you dealt with below)

Don't worry about non-GATKTools.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -322,17 +322,20 @@ private void initializeActiveRegionEvaluationGenotyperEngine() {
      * @return a VCF or GVCF writer as appropriate, ready to use
      */
     public VariantContextWriter makeVCFWriter( final String outputVCF, final SAMSequenceDictionary readsDictionary,
-                                               final boolean createOutputVariantIndex, final boolean  createOutputVariantMD5 ) {
+                                               final boolean createOutputVariantIndex, final boolean sitesOnlyMode,
+                                               final boolean  createOutputVariantMD5 ) {
         Utils.nonNull(outputVCF);
         Utils.nonNull(readsDictionary);
 
+        List<Options> options = new ArrayList<>(2);
```
droazen: `final`

src/test/java/org/broadinstitute/hellbender/engine/FeatureSupportIntegrationTest.java
```diff
@@ -83,6 +92,25 @@ public void testFeaturesAsIntervalsUnrecognizedFormatFile() throws IOException {
         testSpec.executeTest("testFeaturesAsIntervals", this);
     }
 
+    @Test
+    public void testSitesOnlyMode() {
```
droazen: Can you move this to a new `GATKToolIntegrationTest` class instead? I regret my original suggestion of `FeatureSupportIntegrationTest`

src/test/java/org/broadinstitute/hellbender/engine/FeatureSupportIntegrationTest.java
```diff
@@ -83,6 +92,25 @@ public void testFeaturesAsIntervalsUnrecognizedFormatFile() throws IOException {
...
+        File out = createTempFile("GTStrippedOutput", "vcf");
+        String[] args = new String[] {
+                "-V",  FEATURE_INTEGRATION_TEST_DIRECTORY + "vcf_with_genotypes.vcf",
+                "--" + StandardArgumentDefinitions.SITES_ONLY_LONG_NAME,
```
droazen: Can you confirm that this test fails if you remove the `--sites-only` arg temporarily?

src/test/java/org/broadinstitute/hellbender/engine/FeatureSupportIntegrationTest.java
```diff
@@ -83,6 +92,25 @@ public void testFeaturesAsIntervalsUnrecognizedFormatFile() throws IOException {
...
+
+        // Assert that the genotype field has been stripped from the file
+        Pair<VCFHeader, List<VariantContext>> results = VariantContextTestUtils.readEntireVCFIntoMemory(out.getAbsolutePath());
+
+        Assert.assertFalse(results.getLeft().hasGenotypingData());
+        for (VariantContext v: results.getRight()) {
+            Assert.assertFalse(v.hasGenotypes());
+        }
+    }
+
```
droazen: Could you also add a similar test to `HaplotypeCallerIntegrationTest` to prove that it honors `--sites-only`, since that is a special case?

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jamesemery: I have already checked. It is bypassed in a number of tests, the HaplotypeCaller, and `CreateSomaticPanelOfNormals` which doesn't extend `GATKTool` so I elected not to include it here.

src/test/java/org/broadinstitute/hellbender/engine/FeatureSupportIntegrationTest.java
jamesemery: just checked, it fails without --sites-only

src/test/java/org/broadinstitute/hellbender/GatkToolIntegrationTest.java
```diff
@@ -0,0 +1,37 @@
+package org.broadinstitute.hellbender;
```
droazen: This class should live in the same package as `GATKTool` (`org.broadinstitute.hellbender.engine`)

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -322,17 +322,20 @@ private void initializeActiveRegionEvaluationGenotyperEngine() {
      * @return a VCF or GVCF writer as appropriate, ready to use
      */
     public VariantContextWriter makeVCFWriter( final String outputVCF, final SAMSequenceDictionary readsDictionary,
-                                               final boolean createOutputVariantIndex, final boolean  createOutputVariantMD5 ) {
+                                               final boolean createOutputVariantIndex, final boolean sitesOnlyMode,
+                                               final boolean  createOutputVariantMD5 ) {
```
droazen: It would be slightly more logical if the ordering of the boolean args were: `createOutputVariantIndex, createOutputVariantMD5, sitesOnlyMode` instead

src/test/java/org/broadinstitute/hellbender/GatkToolIntegrationTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Test
+    public void testSitesOnlyMode() {
+        File out = createTempFile("GTStrippedOutput", "vcf");
+        String[] args = new String[] {
+                "-V",  TEST_DIRECTORY + "vcf_with_genotypes.vcf",
+                "--" + StandardArgumentDefinitions.SITES_ONLY_LONG_NAME,
+                "-O",
+                out.getAbsolutePath()};
+        new Main().instanceMain(makeCommandLineArgs(Arrays.asList(args), SelectVariants.class.getSimpleName()));
```
droazen: Can you add a new overload of `runCommandLine()` to `CommandLineProgramTest` that takes 2 arguments: a List of args, and an explicit tool name, and call that new overload here? Should also add it to the `CommandLineProgramTester` interface.

src/test/java/org/broadinstitute/hellbender/engine/FeatureSupportIntegrationTest.java
```diff
@@ -1,14 +1,23 @@
 package org.broadinstitute.hellbender.engine;
 
+import htsjdk.variant.variantcontext.VariantContext;
+import htsjdk.variant.vcf.VCFHeader;
+import org.apache.commons.lang3.tuple.Pair;
 import org.broadinstitute.hellbender.CommandLineProgramTest;
+import org.broadinstitute.hellbender.Main;
```
droazen: Remove stray imports from this class

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
```diff
@@ -71,6 +73,30 @@ public void testVCFModeIsConsistentWithPastResults(final String inputFileName, f
         IntegrationTestSpec.assertEqualTextFiles(output, expected);
     }
 
+    @Test
+    public void testSitesOnlyMode() {
```
droazen: Move this test after the standard `testVCFMode*` and `testGVCFMode*` tests, rather than sticking it in the middle of them.

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
```diff
@@ -71,6 +73,30 @@ public void testVCFModeIsConsistentWithPastResults(final String inputFileName, f
...
+        File out = createTempFile("GTStrippedOutput", "vcf");
+        final String[] args = {
+                "-I", NA12878_20_21_WGS_bam,
+                "-R", b37_reference_20_21,
+                "-L", "20:10000000-10100000",
```
droazen: This seems like an unnecessarily large interval for this test. How about 10k bases instead of 100k?

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
droazen: Also call `Utils.resetRandomGenerator()` at the beginning of the test.

src/main/java/org/broadinstitute/hellbender/cmdline/StandardArgumentDefinitions.java
```diff
@@ -37,6 +37,7 @@ private StandardArgumentDefinitions(){}
     public static final String ANNOTATIONS_TO_EXCLUDE_LONG_NAME = "annotations-to-exclude";
     public static final String SAMPLE_NAME_LONG_NAME = "sample-name";
     public static final String PEDIGREE_FILE_LONG_NAME = "pedigree";
+    public static final String SITES_ONLY_LONG_NAME = "sites-only";
```
droazen: Can you rename to the more explicit `sites-only-vcf-output`?

src/test/java/org/broadinstitute/hellbender/GatkToolIntegrationTest.java
jamesemery: done

src/test/java/org/broadinstitute/hellbender/engine/FeatureSupportIntegrationTest.java
jamesemery: doen

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
jamesemery: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
jamesemery: done

jamesemery:
@droazen Responded to comments

PR #4767, opened 2018-05-14
Fix bug in sv copy script

cwhelan:
The SV copy results script has to figure out the total number of workers including standard and preemptibles. The current version of the script had an error in parsing the `gcloud compute dataproc clusters list` command, which left a blank field when no preemptibles were used that couldn't be parsed because it was displayed in fixed-width format. This switches to the CSV format option for the cluster list command, making it easier to parse.


SHuang-Broad:
tested working. merge when ready.
Thanks!

Issue #4712, opened 2018-04-27
java.lang.IllegalArgumentException: Invalid interval. 


chatchawit:
I've got this error.

java.lang.IllegalArgumentException: Invalid interval. Contig:ENST00000342066.7|ENSG00000187634.11|OTTHUMG00000040719.10|OTTHUMT00000276866.2|SAMD11-202|SAMD11|2551|protein_coding| start:0 end:0

I'm working on hg38. The VCF input file was produced from Mutect2.

First, I've the error "java.lang.NullPointerException". So I put only "gencode" folder in the data-source folder. Then, I've got the error message above.

lbergelson:
Hi @chatchawit,

It's hard to tell what's going on from the information you provided.  Could you include the exact commandline that you're trying to run.  Also, the complete stack trace from both the original error and the new error message would be very helpful.

jonn-smith:
@chatchawit Like Louis said, there have been quite a few updates.

The issue you're seeing is a data sources problem relating to HG38.  This has been fixed in the latest release of funcotator and the data sources.

If you grab the latest software and the latest data sources release this problem should go away.

Sorry for the inconvenience.

chatchawit:
I've tried the latest GATK today. The error message changed. Please help. Thanks.

Using GATK jar /omics/chatchawit/gatk2/gatk-package-4.0.4.0-local.jar
Running:
    java -Dsamjdk.use_async_io_read_samtools=false -Dsamjdk.use_async_io_write_samtools=true -Dsamjdk.use_async_io_write_tribble=false -Dsamjdk.compression_level=2 -jar /omics/chatchawit/gatk2/gatk-package-4.0.4.0-local.jar Funcotator -R /omics/chatchawit/bundle/hsa38.fasta -V /omics/chatchawit/sm/out/sample21.vcf -O /omics/chatchawit/sm/anno/sample21.vcf --output-file-format VCF --data-sources-path /omics/chatchawit/bundle/dsrc/ --ref-version hg38
10:24:13.971 INFO  NativeLibraryLoader - Loading libgkl_compression.so from jar:file:/omics/chatchawit/gatk2/gatk-package-4.0.4.0-local.jar!/com/intel/gkl/native/libgkl_compression.so
10:24:14.182 INFO  Funcotator - ------------------------------------------------------------
10:24:14.183 INFO  Funcotator - The Genome Analysis Toolkit (GATK) v4.0.4.0-0.0.2
[...]

PR #4770, opened 2018-05-15
Addressing issue #4712

jonn-smith:
Added in a script to pull down the latest Gencode data source.
Fixed an issue in 5' UTR processing that would cause variant alleles with length > 1 to throw an exception (issue 4712).
Added three test cases to prevent regression of issue 4712.
Updated Gencode codec to be compatible with latest Gencode release (v28).
Fixed a bug in the version detection for Funcotator data sources that would prevent newer data source versions from being detected as compatible (date comparison error).


src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/gencode/GencodeFuncotationFactory.java
```diff
@@ -1074,32 +1073,38 @@ private GencodeFuncotation createUtrFuncotation(final VariantContext variant,
             // Get the 5' UTR sequence here.
             // Note: We grab 3 extra bases at the end (from the coding sequence) so that we can check for denovo starts
             //       even if the variant occurs in the last base of the UTR.
+            final int numExtraLeadingBases = 2;
```
LeeTL1220: Extract constant.

PR #4776, opened 2018-05-16
Fixes PairedStrandedIntervalTree

cwhelan:
The current implementation of `PairedStrandedIntervalTree` contains a few bugs and a few inefficiencies. The most important bug is the fact that it could not store two different entries with the same left interval but different left strands. This PR attempts to fix that. Unfortunately fixing that issue required a near-total re-write of the class. The rewritten version also introduces several efficiency improvements such as not calling `contains()` whenever `put()` is called (which essentially did two lookups) and tracking the size of the data structure instead of re-computing it each time.


droazen:
@cwhelan @tedsharpe Can you guys either review this ancient PR or close it if it's no longer needed? Thanks!

cwhelan:
Closing this for now until its ready to be revived.

PR #4780, opened 2018-05-16
Vrr single contig reference aligner

vruano:


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SingleReferenceSequenceAligner.java
```diff
@@ -0,0 +1,136 @@
...
+ * Encompasses an aligner to a single-contig reference.
+ * <p>
+ *     This is an {@link AutoCloseable} mean to be use in try-with-resource constructs.
+ * </p>
+ * <p>
+ *     If you don't do so, please remember to call {@link #close} at the end to free
+ *     resources.
+ * </p>
+ */
+public class SingleReferenceSequenceAligner implements AutoCloseable {
```
vruano: FYI I will rename The class to SingleContigReferenceAligner and make it final.

vruano:
It can easely be extended to multiple sequences... it just happened that I didn't need it personally. 

I took a quick pick to your branch... do you really care about the contig descriptions?.... I would say that this "aligner" class should not be responsible of compose the multi sequence fasta file but rather accept one as a constructor argument and the construction of the fasta is delegated back to the invoker code; in this new more general aligner the current single contig could be implemented as a public static method call.

vruano:
In addition just to clarify that the "contig" word reference to the sequences in the given reference not the number of contigs that you might align to it, so single-contig-reference-aligner means an aligner that will align several sequences (read, contigs, random-stuff) vs a single sequence/contig reference. 

I guess I shall change contig for chromosome or sequence to make it more clear.

src/test/java/org/broadinstitute/hellbender/utils/RandomDNA.java
```diff
@@ -20,11 +18,44 @@
  *     Returned bases are always in upper case and one of the valid four nucleotides 'A', 'C', 'G' and 'T'.
  * </p>
  *
+ * <p>
+ *     This class is written in a way that ensure that the sequence of bases returned would only depend on the seed and
+ *     not the sequence of calls to different base generating methods.
+ *
+ *     For example a single call asking for 132 bases using {@link #nextBases(int)} would generate the same sequence
+ *     as calling 132 times {@link #nextBase()}.
```
SHuang-Broad: if initialized with the same seed

vruano:
@SHuang-Broad 
I have addressed comments and changed the SSRAligner interface to make it more general in terms of the type of inputs (base sequence providers) and the type of outputs that it will generate. 

Also I added and example as to how to address your SVFastqUtil.FastqRead to SAMRecord realigner may look like but is untested. perhaps you can take on that commit and add the test, 

does not need to be part of this PR push though.

PR #4789, opened 2018-05-19
Inversion breakpoint linking (no merge yet)

SHuang-Broad:
This is for demo purpose only, so the code is not ready yet to be merged:

### background & goal
Currently, there are two parallel code path for structural variation breakpoint location and type inference using local assembly contig alignment signatures in the pipeline `StructuralVariationDiscoveryPipelineSpark`

* the stable code path: scanning neighbor chimeric alignment pairs of a contig iteratively and outputs inversion breakpoints as symbolic variant `<INV>`, annotated with `INV55` and `INV33` for signaling if it is the left or right breakpoint of the assumed inversion.
* the experimental code path that separates the alignment pre-processing step from the inference step, and studying the alignments in whole; this code path, in addition to outputting insertion, deletion and small duplication calls as does the stable path, outputs 
    * BND records representing assembled breakpoints for which type could not be completely determined using only the contig alignments; this includes supposedly inversion breakpoints
    * complex (`<CPX>`) variants from assembly contigs with more than 2 alignments
    
The tool proposed in this PR is based on [manual review](https://github.com/broadinstitute/dsde-methods-sv/tree/sh_inv_filter_init/docs/knowledgeBase/variantReview/inversion/chm) of a callset generated a long time ago (but still useful for studying filtering inversion breakpoints), and is designed to be integrated with the experimental code path.

#### input:
* the "INV55/INV33"-annotated `BND` records output by the upstream experimental code path
  * BND's have related concepts of `MATE` and `PARTNER` (see figure below, left)
  * `MATE`: novel adjacency, i.e. contiguity on sample that is absent on reference (e.g. mobile element insertions, deletions)
  * `PARTNER`: novel disruption, i.e. contiguity on reference disrupted on sample (e.g. insertions, deletions)

inversion_demo

* complex variants detected by the upstream experimental code path; the reason is that sometimes inversion calls are incorporated as part of a larger, more complex event and the logic implemented in the upstream code, theoretically, allows for arbitrarily complex rearrangement; shown above on the right is a table of inversion calls (TP/FP 12/7 using PacBio calls on CHM-1 & 13 cell lines as truth) extracted from the `<CPX>` calls, they were extracted by the tool proposed in PR #4602.

#### stages:

* Primitive filter on breakpoints 
    * low MQ of assembly contigs' mappings that evidenced the BND records, 
    * suspiciously large distance between mates (mate pairs whose distance are over $10^5$bp (~1/3 of input, see blelow) are more likely to be artifact or dispersed/segmental duplications)

  <p align="center"></p>
    * if overlaps with CPX (supposedly they should be captured already, or is more complex than what can be comprehended by the logic proposed here)

The mates are then converted to intervals bounded by the mates' locations. These "normal sized" variants are sent down for further analysis.

* Filtering based on overlap signatures. Here we have several possible scenarios (total ~130 pairs of mates):
  * no overlappers (~ 50 mate pairs, balanced between ++/--)
  * multiple overlappers (~ 10 mate pairs, balanced between ++/--)
  * unique overlapping pairs of mate pairs (~ 60 mate pairs)
     * which overlaps with same type (++/++ or --/--, ~ 10 mate pairs)
     * which overlaps with opposite type (++/-- overlap, ~ 50 mate pairs). These are the overlapping pairs sent down for breakpoint linking, expecting a maximum of 20~30 inversion calls.

* Type inference. Here we have four possible cases, each signaling what could be involved (primed block is inverted):
  * INV55 interval left/right boundary upstream of INV33 interval's left/right boundary: `ABC ->    B'`
  * INV33 interval left/right boundary upstream of INV55 interval's left/right boundary: `ABC -> AC'B'A'C`
  * INV33 interval contains INV55 interval: `ABC -> ABA' or AB'A'`
  * INV55 interval contains INV33 interval: `ABC -> C'BC or C'B'C`

Note that for the last two cases, where the inverted dispersed duplication is guaranteed, the two possible alternate alleles are reverse complement&mdash;inversion&mdash;of each other, hence signatures of contig alignments along is not enough, and alignments of short reads within the affected region cannot break the degeneracy either.
So we need to attach left and right flanking regions to the affected region, and align short reads back to these two haplotypes and study the pair orientations of the alignments to break the degeneracy.

#### output:

* VCF containing the inversion and flanking deletion and dispersed duplication calls (together with CPX-derived inversion calls, total number of `INV` calls are ~ 30)

* BED file on filtered BND's citing reason for filtering

Relevant files, including an IGV session can be found in this [bucket](https://console.cloud.google.com/storage/browser/broad-dsde-methods/shuang/archive/inversion-algo-demo/?project=broad-dsde-methods&organizationId=548622027621)

## Todo:

* Large inversions. The primitive size-based filtering step and the filtering step requiring matched mate pairs undoubtedly will cause us false negatives, as sometimes we don't expect assembly of all breakpoints for inversions complicated by copy number events. The inversions currently captured tend to be small inversions
  
```
Min.     1st Qu.    Median      Mean     3rd Qu.       Max.
77.0       188.0     359.0    1144.5       823.2    12697.0
```
    
* run check against reference annotation: known Seg. Dup., RC-STR, centromere, as well as consistent pair support from short reads
    
  * __Question: is this pre-filtering a bad idea; if not, how to report?__

  *  __Question: is the number $10^5$ reasonable?__
  
  * __Question: any other suggestion on filtering criteria?__
 
* Corner cases here & there in the proposed tool

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/InversionBreakendPreFilter.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/InversionBreakendPreFilter.java
@@ -29,21 +29,20 @@ import static org.broadinstitute.hellbender.tools.spark.sv.utils.SVLocalContext.
  * Strategy is to perform pre-filtering upfront because they are likely to be artifact or different types:
  *
  * 1) first filter out BND records that are (see {@link PrimitiveFilteringResult}):
- *      a) inter-chromosome, no SS, 2nd in mate (redundant),
- *      b) badly supported by MQ,
- *      c) mates too far away,
- *      d) covered by CPX variants
+ *      a) badly supported by MQ,
+ *      b) mates too far away,
+ *      c) covered by CPX variants
  *
  * then convert the paired mate locations into boundaries of an interval, then
  *
- * 2) filter out intervals that are (see {@link OverlapBasedFilteringResult}:
+ * 2) filter out intervals which (see {@link OverlapBasedFilteringResult}:
  *      a) overlaps with multiple other intervals
  *      b) overlaps with no other intervals (including those that only overlaps with the "popular" ones mentioned above)
  *      c) overlaps with one other interval of the same type (INV55/55 or INV33/33 pairs)
  *
- * The resulting INV33/55 pairs are then returned.
+ * The resulting INV33/55 pairs are then returned via {@link OverlappingPair}.
  */
-final class InversionBreakendPreFilter {
+public final class InversionBreakendPreFilter {
 
     private enum ReasonForFilter {
         LOW_MQ, MATE_DISTANCE, COVERED_BY_CPX,
@@ -51,17 +50,22 @@ final class InversionBreakendPreFilter {
         SAME_DIRECTION_OVERLAPPER // INV55/33 overlapping with INV55/33
     }
 
+    /**
+     * Holding information on two BND records
+     * (INV55 and INV33, can NOT be INV33/INV33 or INV55/INV55)
+     * whose mates' spanning intervals overlap.
+     */
     public static final class OverlappingPair implements Serializable {
         private static final long serialVersionUID = 1L;
 
-        protected final InvBreakEndContext fivePrimeBreakEnd;
-        protected final InvBreakEndContext threePrimeBreakEnd;
+        final InvBreakEndContext fivePrimeBreakEnd;
+        final InvBreakEndContext threePrimeBreakEnd;
 
-        protected final String chr;
-        protected final int fiveIntervalLeftBoundary;
-        protected final int fiveIntervalRightBoundary;
-        protected final int threeIntervalLeftBoundary;
-        protected final int threeIntervalRightBoundary;
+        final String chr;
+        final int fiveIntervalLeftBoundary;
+        final int fiveIntervalRightBoundary;
+        final int threeIntervalLeftBoundary;
+        final int threeIntervalRightBoundary;
 
         OverlappingPair(final SVLocalContext.InvBreakEndContext first, final SVLocalContext.InvBreakEndContext second) {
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/InversionBreakendPreFilter.java
```diff
@@ -0,0 +1,440 @@
...
+     *   3) is paired with one intra-chromosome SS BND, but a INV55/INV55 or INV33/INV33 pair
+     */
+    private static final class OverlapBasedFilteringResult {
+        private final Set<InvBreakEndContext> noOverlappers;                                        // variants whose has no overlapping INV BND suspects, or whose unique overlapper is one of {@code multipleOverlappers}
+        private final Map<InvBreakEndContext, List<String>> multipleOverlappers;                    // variants who overlaps multiple other records (note: not using map because VC.equals() is hard)
+        private final Set<Tuple2<InvBreakEndContext, InvBreakEndContext>> uniqueStrangeOverlappers; // pairs of variants (INV55/55, INV33/33) that overlap each other uniquely
+
+        private final Set<OverlappingPair> uniqueHetOverlappers;                                    // pairs of variants (INV55/33, INV33/55) that overlap each other uniquely, pairs that will be analyzed
+
+        OverlapBasedFilteringResult(final Set<InvBreakEndContext> noOverlappers,
```
cwhelan: Have you thought about storing these breakpoints in a graph structure?

Then you could traverse nodes in the graph and get their category on the fly instead of splitting them into multiple collections.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/LinkedInversionBreakpointsInference.java
```diff
@@ -0,0 +1,935 @@
...
+    }
+
+    //==================================================================================================================
+
+    /**
+     * Scenario where re-alignment of short reads back to artificial reference haplotypes
+     * is NOT necessary
+     * to resolve which variants are here.
+     */
+    abstract static class NonDegenerateScenario extends OverlappingPairHandler {
```
cwhelan: In general I think naming abstract subclasses after their superclass is a good idea for readability, ie call this `NonDegenerateOverlappingPairHandler`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/LinkedInversionBreakpointsInference.java
```diff
@@ -0,0 +1,935 @@
...
+            final List<VariantContextBuilder> result = new ArrayList<>(3);
+
+
+            byte[] refBase = getRefBase(reference, new SimpleInterval(overlappingPair.chr,
+                    overlappingPair.threeIntervalLeftBoundary, overlappingPair.threeIntervalLeftBoundary));
+            final SimpleInterval invertedInterval = new SimpleInterval(overlappingPair.chr,
+                    overlappingPair.threeIntervalLeftBoundary, overlappingPair.fiveIntervalRightBoundary);
+            final VariantContextBuilder inversionBuilder = makeInversion(invertedInterval, Allele.create(refBase, true));
+
+            if ( overlappingPair.threeIntervalLeftBoundary - overlappingPair.fiveIntervalLeftBoundary - 1 > 49 ) {
```
cwhelan: Don't we have the 50 value stored as a constant somewhere? Also, what's the rationale for using 50 bases here?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/LinkedInversionBreakpointsInference.java
```diff
@@ -0,0 +1,935 @@
...
+     */
+    abstract static class NonDegenerateScenario extends OverlappingPairHandler {
+        private static final long serialVersionUID = 1L;
+
+        NonDegenerateScenario(final OverlappingPair overlappingPair) {
+            super(overlappingPair);
+        }
+    }
+    
+    private static final class FiveInterleavesThree extends NonDegenerateScenario {
```
cwhelan: A lot of the code between this class and the three -> five class below looks pretty similar. It'd be good to find opportunities for reuse to remove code duplication.

SHuang-Broad:
> I've looked through at a high level although I'd prefer not to do a line-by-line code review until you think this is ready to be merged). Here are some comments on your proposed algorithm description:

Answer: thanks for looking! That's what I intended as I'm not sure if the algorithm itself would face strong critics. If that's the case, time spent on coding is not worth it IMO. In fact ideally I'd like to write a design doc or something similar, and only code after the design is agreed on.

--------

> I would of course prefer not to have to have a hard filter on length. This would mean we would never call a large inversion even if it exists. 

Answer: Totally agree. Now looking back, it get clearer to me that this proposal contains two parts: the filtering part, and the breakpoint linking part, separated into two major classes `InversionBreakendPreFilter` and `LinkedInversionBreakpointsInference`. That being said, it doesn't make much sense to separate them into two PRs because _currently_ the filtering part is designed around the linking part, i.e. it is trying to check which BND's are suitable to the logic implemented in the linking part, and if the logic isn't applicable to an BND, the BND simply slips through without generating any new interpretations. So `InversionBreakendPreFilter` is a filter and a classifier at the same time, it function is really diverting different BND's to be handled by different logics, and it definitely should be improved.
If you buy this argument, I am also fully aware of the code design issue that it is preferable to NOT divert&mdash;gather&mdash;send through different handlers like it currently is for calling variants from the assembly contigs, instead it should be a single stream pass through all the BND's. I'll try to follow the preferred design.

> What about some other filters more specifically aimed at the artifacts that cause these false large calls? I think it's a good idea to check annotations -- ie. do the mates lie at two regions that are segmental duplications of each other, or one side of the mate looks like a transposable element insertion? I guess it's ok to put in a tool with this limit temporarily, though.

Answer: Agree. And I think these kind of checking are not only good, but a mandate for a good caller, i.e. to take advantage of prior knowledge. I am also thinking about improving it by checking if there are short read evidence supporting the BND's (sometimes there aren't any, which is strange for the alignments of the assembly to point to such novel adjacencies in the first place).
However, I might have to push back a little on this particular request at this point because such features can be later added on like you said, and probably now is not the best time to do it, because the linking logic is unlikely to be affected by presence of these reference annotations.
One note though, and I think you don't mean it literally either, the large ones are not necessarily artifacts. Like what we discussed during the group meetings, they are "filtered" partly because they are more likely to overlap with multiple other BND mates, and are more likely to be artifacts or mobile element insertions. In other words, they are not sent for linking because they are less likely to be suited for the linking logic in `LinkedInversionBreakpointsInference`. But if they do, the linking is totally agnostic about size.

> Building in an ability to check the copy number of the region in which an inversion breakpoint lies (by checking against a CNV call for example) would probably be really helpful.

Answer: Agree. And it shouldn't be too hard to do that. But again, may be in a different PR?

> When you say 'overlaps with CPX' it might be helpful have a more particular set of criteria.. A larger inversion event might span over a smaller complex event.

Answer: Note taken. I'll improve on this in the next polished implementation that is ready for line-by-line review.

> I'd check to see if there are additional filters implemented in SV-Pipe that you could apply here.

Answer: Yes. But I'm not sure if by SV-Pipe you mean RDTest repo? Sorry I wasn't around when I was added to that repo so I don't know the context. In addition, I believe this could be another future improvements?

> I don't think it's that necessary to provide reports on things that didn't become breakpoints. If they still exist as BNDs in the original perhaps there are some codes we could add as INFO field annotations to indicate that they were considered for inversion calling but filtered, and the reason for doing so.

Answer: Agree. And I think the final format can definitely change, as long as the resolved variants, be it deletions, inverted dispersed duplications, and/or real inversions are registered in a VCF. The BED file that is currently produced is for development use, i.e. for me to check why a certain BND is not sent for the analysis so that I can develop ideas on how to catch them in the future (you know, BED are IGV-friendly), and it can certainly be dropped in the final output. 
Regarding adding another INFO field saying they are not used for _THIS PARTICULAR_ logic, I prefer not to do it this way, because we can add more logics in the future, and capturing/resolving these BND's that are not suitable for _THIS PARTICULAR_ logic. I guess in general my personal preference is to put less algorithm-related information in VCF for analysts (less reading for them), and produce add on files for tool developers. What's your thoughts?

> Does this even have to be a spark tool? It looks like you are just reading the variants into a parallel spark context, filtering, and then collecting them to actually process them. Why not just make this a non-spark tool and process it all in memory on one node?

Answer: Agree. It doesn't have to be, at least in theory, and it probably is going to be faster as we don't need to incur the Spark overhead for such a typically small job. But (I'm saying too many buts....) up to this point all SV tools are under the package `hellbender.tools.spark.sv`, so I'm following suit here. Note the two classes's main interface methods mentions nothing about RDDs (that's on purpose). 
On the other hand, this is an engineering question I believe, and it depends on whether we want to put as much of discovery code as possible into `StructuralVariationDiscoveryPipelineSpark` (the last commit actually hooks the two classes into it, so a single invocation of the tool produces more variants), or we go wdl in pipelining the whole process.


-------

All in all, I think the comments and critics are generally about the "filtering"/"classifying" part, and the most serious concern about it is false negatives. Am I understanding correctly? If so, given that the filtering step is only picking the BND's that are suitable to the linking logic, I can imagine the false negative problem be solved in the future by other logics (e.g. more relaxed requirement on pair matching, or not even requiring matching INV55/INV33 pairs, etc.) In fact, that's what I'm planning on.
Another part of the problem is how much I can accomplish in this single PR, and how large it should be&mdash;the forever existing problem for new tools.

SHuang-Broad:
closing as it won't be merged anytime soon (only I care about it now).

PR #4801, opened 2018-05-23
New qual doesn't count spanning deletions as support for variant qual

davidbenjamin:
@ldgauthier The new qual was not doing the right thing for spanning deletions.  This fixes it.  Would you mind reviewing?  And would you like me to re-run those GVCFs just in case before going ahead with #4614?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/AlleleFrequencyCalculator.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/AlleleFrequencyCalculator.java
@@ -3,6 +3,7 @@ package org.broadinstitute.hellbender.tools.walkers.genotyper.afcalc;
 import htsjdk.variant.variantcontext.Allele;
 import htsjdk.variant.variantcontext.Genotype;
 import htsjdk.variant.variantcontext.VariantContext;
+import it.unimi.dsi.fastutil.ints.Int2ObjectArrayMap;
 import org.apache.commons.math3.util.MathArrays;
 import org.broadinstitute.hellbender.tools.walkers.genotyper.GenotypeAlleleCounts;
 import org.broadinstitute.hellbender.tools.walkers.genotyper.GenotypeLikelihoodCalculator;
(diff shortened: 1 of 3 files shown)


davidbenjamin:
`testSpanningDeletionIsNotConsideredVariant` fails in master.  `testPresenceOfUnlikelySpanningDeletionDoesntAffectResults` passes here and in master, as it should.  Its purpose is to check that the spanning deletion logic doesn't break anything.

ldgauthier:
If the second test passed in master then I don't consider it to be a good test of your fix. Can you make a test where the PLs with and without span del would lead to different QUALs in the old version? Maybe a case with a confident deletion in sample1 and a low quality SNP in sample2 that causes a spanning deletion in sample1.  You could probably do this by modifying your first test by adding a sample that has A or B called with low GQ with respect to the reference.

davidbenjamin:
The point of the second test is to make sure that the new code path involving the spanning deletion has no effect if the spanning deletion has low likelihood.  You could imagine that I introduced a bug where you lose sensitivity whenever a spanning deletion appears in the alleles list even if there's very little evidence for it, just because its potential existence sends the code down some new, incorrect, logic.

I think `testSpanningDeletionIsNotConsideredVariant` is already doing what you want because before this fix the example VC is definitely considered to be variant even though the only non-ref evidence is for the symbolic `*` allele.  The PLs in this example are 50, 100, 100, 0, 100, 100, which means that it is overwhelmingly likely that this sample's genotype is REF/SPAN_DEL.  In the old code you get a qual of about 50 (the hom ref PL), whereas in the new code you get that the chances of a variant are about 1 in 10^10 since the only truly variant PLs are 100.

Now, if you think a multi-sample case would be useful that sounds good to me.  However, a "confident deletion in sample1 and a low quality SNP in sample2" would not look very different since the deletion would only exist in an upstream VC, not the VC being tested, so basically we would have sample1 = REF/SPAN_DEL, sample2 = REF/SNP with low quality.  The current test is the same thing with just sample1.  Is this what you had in mind?

ldgauthier:
Let's be rigorous and use two samples so that you're testing a case that could actually occur in the wild. It should be easy enough to add sample2 with PLs like [10,0,40,100,70,300].  Then can you add a comment about where the "magic number" in the assert statement comes from?  Or maybe to be super duper rigorous you could compare the QUAL from that variant to the QUAL for sample3 and sample4 where 3 and 4 have the same PLs as 1 and 2, but without the spanning deletion.

davidbenjamin:
@ldgauthier I did all of those things.  Note that to be precise, when you remove the spanning deletion the "het" REF / SPAN_DEL must be treated as haploid, which the new tests do.

ldgauthier:
@davidbenjamin I kicked Travis and now tests are green so you can merge.

PR #4827, opened 2018-05-30
Adds a class to represent expected insert-size distribution (normal a…

vruano:
…nd log-normal distributed) parametrized by

insert size mean and stddev


vruano:
I use this class to allow the user to pass a Normal or LogNormal expected distribution for insert sizes; notice that the String parameter constructor allows for it to use as an [at]Argument type in a tool. 

@tedsharpe  please review.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/InsertSizeDistribution.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Holds the information characterizing and insert size distribution.
+ */
+public class InsertSizeDistribution implements Serializable {
```
tedsharpe: I thought we had a convention for class layout that put all fields first, then constructors, then other stuff.  I could be wrong.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/InsertSizeDistribution.java
```diff
@@ -0,0 +1,157 @@
...
+    private final String description;
+
+    private transient AbstractRealDistribution dist;
+
+    private AbstractRealDistribution dist() {
+        initializeDistribution();
+        return dist;
+    }
+
+    public double average() {
```
tedsharpe: Maybe put this down with the other methods like min, max, density?  Should it be called "mean" rather than "average"?

vruano:
Added support for the empirical distribution read-metadata txt file generated in the pipeline.
Also rebased to current master.

vruano:
@tedsharpe 
Could you review the last commit adding support for the empirical distribution?

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/InsertSizeDistribution.java
```diff
@@ -0,0 +1,205 @@
...
+
+    @VisibleForTesting
+    static final Type[] SUPPORTED_TYPES = { new NormalType(), new LogNormalType() };
+
+    public interface Type {
+        List<String> getNames();
+
+        AbstractRealDistribution fromMeanAndStdDeviation(final double mean, final double stddev);
+
+        default AbstractRealDistribution fromReadMetadataFile(final String meanString) {
```
tedsharpe: It seems rather fragile to parse a text file for this information.  We already have a means of serializing the ReadMetadata to a file.  Why don't you just read that?
You could change the ReadMetadata code to always write read-metadata.txt as well as read-metadata.bin.

vruano:
Added a pure "empirical" isize distribution where the prob of each size is determine by the fraction of cases with that insert size + some smoothing so that 0-count insert size don't have 0 probability.

Also I moved the supported isize distribution "shapes" (normal, lognormal, empirical) to a enum.

vruano:
@tedsharpe please check on the last commit.

tedsharpe:
This is difficult to review because there isn't any client code:  I don't know how this is going to be used.  I suspect I'd have a lot of "YAGNI" comments if I knew.
For example, you are basing all your implementations on Apache's AbstractIntegerDistribution.  That class, it seems to me, is really intended to allow you to do  sampling from a distribution.  But I suspect you won't be sampling, you'll only be asking questions about density.  If so, there's a lot of baggage that gets pulled into your anonymous implementations of this class:  random number generators, boundary information, etc.  Lots of extra boilerplate.

Couldn't this be clearer if reorganized as an abstract class implementing AbstractIntegerDistribution, 3 concrete classes for each case (rather than the current anonymous classes), a factory that takes a spec and returns the correct distribution, and a simple enum class?

It seems weird that the distributions you allow users to realize using a spec are both two-tailed distributions, when fragment size is a one-tailed distribution.

It seems awkward that failure to parse a distribution spec leads to a code path where you try to extract a file name and read serialized read metadata.  Wouldn't it be clearer to have two completely distinct code paths with a different program argument for the empirical case?

The read metadata gives per library distributions.  It seems suspect that you are folding them all together.  Different libraries can have rather different fragment size stats.

Still don't like that you're providing the possibility of reading the metadata text file.  Seems fragile.  Why don't you modify the ReadMetadata code to always produce just the data you need.  Then you could eliminate the text-file code.  And you could simplify the code that processes the serialized ReadMetadata which now has this awkward code path:  CDF -> density -> sum across libs -> density+CDF stored in memory.  If you have the CDF you can trivially produce density on demand.

Notwithstanding all this, if you're happy with the code as it stands, feel free to merge.
Back to you, review done.

vruano:
Noted, will mege, I would rather address those concerns in a separate pull-request(s), later on.

PR #4834, opened 2018-05-31
Sheila apply vqsr quick fix

chandrans:
User pointed out --ts_filter_level does not work. https://gatkforums.broadinstitute.org/gatk/discussion/12040/errata-documentation-and-help-text-apparently-incorrect-for-applyvqsr-ts-filter-level


src/main/java/org/broadinstitute/hellbender/tools/walkers/vqsr/ApplyVQSR.java
```diff
@@ -81,7 +81,7 @@
  *   -R Homo_sapiens_assembly38.fasta \
  *   -V input.vcf.gz \
  *   -O output.vcf.gz \
- *   --ts_filter_level 99.0 \
+ *   --truth-sensitivity-filter-level 99.0 \
```
cmnbroad: @chandrans Might as well change the other one of these below on line 97.

chandrans: Thanks Chris! Done

PR #4841, opened 2018-06-01
Deleted an unused duplicated method in GenotypeGivenAllelesUtils

### LINKED ISSUE #4182: Duplicate composeGivenHaplotypes methods
`composeGivenHaplotypes` in `GenotypingGivenAllelesUtils` is an *exact* duplicate of the same method in `ReadThreadingAssembler` and is never invoked in the GATK.  It can be deleted.

davidbenjamin:
Closes #4182.  Note the duplicate method in `ReadThreadingAssembler`. @droazen Let me know if I should bother someone outside the engine team for things like this.

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
@@ -1538,71 +1538,7 @@ public final class GATKVariantContextUtilsUnitTest extends GATKBaseTest {
             }
         };
     }
-
-    @Test(dataProvider="overlapWithData")
-    public void testOverlapsWith(final VariantContext vc, final GenomeLoc genomeLoc) {
-        final boolean expected;
-
-        if (genomeLoc.isUnmapped())
-            expected = false;
-        else if (vc.getStart() > genomeLoc.getStop())
-            expected = false;
-        else if (vc.getEnd() < genomeLoc.getStart())
-            expected = false;
-        else if (!vc.getContig().equals(genomeLoc.getContig()))
-            expected = false;
-        else
-            expected = true;
-
-        Assert.assertEquals(GATKVariantContextUtils.overlapsRegion(vc, genomeLoc), expected);
-    }
-
-
-    private final String[] OVERLAP_WITH_CHROMOSOMES =  { "1", "2" };
-    private final int[] OVERLAP_WITH_EVENT_SIZES =  { -10, -1, 0, 1, 10 }; // 0 == SNP , -X xbp deletion, +X xbp insertion.
-    private final int[] OVERLAP_WITH_EVENT_STARTS = { 1000, 1001,
-            1005, 1010,
-            1009, 1011,
-            2000 };
-
-    @DataProvider(name="overlapWithData")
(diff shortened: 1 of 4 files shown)


src/test/java/org/broadinstitute/hellbender/utils/variant/GATKVariantContextUtilsUnitTest.java
```diff
@@ -1483,71 +1483,7 @@ public void remove() {
             }
         };
     }
-
-    @Test(dataProvider="overlapWithData")
-    public void testOverlapsWith(final VariantContext vc, final GenomeLoc genomeLoc) {
```
droazen: Instead of deleting this test case, can you instead convert it into test coverage on the remaining `GATKVariantContextUtils.overlapsRegion(VariantContext, SimpleInterval)` method?

davidbenjamin: Could I instead get rid of that method and replace it with the equivalent `Locatable.overlaps(Locatable)`?  They are equivalent:

The `GATKVariantContextUtils` version is
```
if (variantContext.getEnd() < region.getStart())
    return false;
if (variantContext.getStart() > region.getEnd())
    return false;
return variantContext.getContig().equals(region.getContig());
```
i.e.
```
return sameContig && vcDoesntEndBeforeRegionStarts && vcDoesntStartAfterRegionEnds;
```
The Locatable method is (I'm inlining `withinDistanceOf`, `contigsMatch`, and `CoordMath.overlaps` and expanding `Objects.equal` to delegate to `String.equals`):
```
return vc.getContig().equals(region.getContig()) && 
(region.getStart() >= vc.getStart() && region.getStart() <= vc.getEnd()) || 
(region.getEnd() >=vc.getStart() && region.getEnd() <= vc.getEnd()) ||
vc.getStart() >= region.getStart() && vc.getEnd() <= region.getEnd());
```
i.e.
```
return sameContig && 
(regionStartsInVariant || regionEndsInVariant || variantStartsAndEndsInRegion);
```
You can satisfy yourself that these are equivalent.

Other than that, the `Locatable` version returns `false` if an argument is `null`, while the `GATKVariantContextUtils` version throws an error.  But that cant be much of an obstacle.

davidbenjamin:
Whoops, forgot to re-assign @droazen after that last question.

davidbenjamin:
@droazen I put that last change I mentioned in a commit.

davidbenjamin:
@droazen ping

davidbenjamin:
@lbergelson Do you have thoughts on that last commit since David R is away for a while?

davidbenjamin:
@cmnbroad Do you have thoughts on replacing `GATKVariantContextUtils.overlapsRegion()` with `Locatable.overlaps()` ie whether I am correct that the former is a duplicate of the latter?

cmnbroad:
@davidbenjamin :+1: They do appear to be equivalent - I don't think the differences in null handling will matter for existing code since the only call site you're changing in `ReadThreadingAssembler` is already protected by `Utils` preconditions. There are a few dead imports in the files you changed that can be removed now, though.

davidbenjamin:
Thanks @cmnbroad!  I optimized the imports, now waiting for Travis.

PR #4847, opened 2018-06-01
Fixes a ArrayIndexOutOfBound exception into a more informative error message …

vruano:
…for the user.

The problem occurs when there is no `=` sign in the property string passed by the user.


vruano:
@lbergelson or @cmnbroad please review, this one should be a short one.

vruano:
Travis building issues, rebased against master in case it makes a difference.

PR #4848, opened 2018-06-04
realignment filter in M2 wdl

davidbenjamin:
@LeeTL1220 This is worth getting in before the release.


scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -73,6 +73,8 @@ workflow Mutect2 {
     File? gnomad_index
     File? variants_for_contamination
     File? variants_for_contamination_index
+    File? realignment_index_bundle
```
LeeTL1220: Can you update the comments and any relevant READMEs?

scripts/mutect2_wdl/mutect2_nio.wdl
```diff
@@ -74,6 +74,8 @@ workflow Mutect2 {
     Int scatter_count
     File? gnomad
     File? variants_for_contamination
+    File? realignment_index_bundle
```
LeeTL1220: Could you add comments above?

PR #4851, opened 2018-06-06
tweaked Mutect read position filter to better handle non-biological insertions

davidbenjamin:
@LeeTL1220 This is for Junko's issue.


LeeTL1220:
@davidbenjamin As per offline conversation and that this code is actually simpler.... feel free to merge.

PR #4858, opened 2018-06-07
bamout no longer shows deletions in reference-supporting reads ending in STRs

### LINKED ISSUE #4829: Bamout aligns non-spanning reads to false deletions in STRs
The following screenshot of a bamout is common in HaplotypeCaller and Mutect2:

screen shot 2018-05-30 at 2 36 58 pm

Notice how all of the reads that appear to support a single-base deletion actually do not because they do not span the poly-T homopolymer.  This makes the bamout harder to interpret.  In this particular case (of a false positive M2 insertion) it is a red herring because it suggests that the multiallelicness of the site is relevant to the false positive deletion.

One could ask what in the GATK engine is responsible.

* The assembly engine, perhaps?  No, it is the assembly engine's job to propose possible haplotypes, not to call them.  In any case, there *is* one spanning read with the deletion above the reads shown, so it is a valid path in the graph.
* Pair-HMM?  This one confused me for a while, but no.  The engine is *not* saying that these reads' best alignment to the reference has a deletion, which would be false because there is a gap opening penalty.  Rather, it says that they align equally well (with no deletions) to the ref haplotype and to the deletion haplotype.  The deletion shown in IGV is the deletion of the alt haplotype relative to the reference, not of the reads relative to their best haplotype.
* The bamout writer?  Nope, that code is really straightforward and does the right thing.

So what's the issue?  Well, the bamout writer gets its read alignments from the `readLikelihoods` after the reads have been realigned to their best haplotype.  In these cases, it turns out that the alignment of the reads to their best haplotype, the deletion has a log likelihood better than the alignment to the ref haplotype by about 0.00001.  The simplest solution would be to give an extremely modest prior in favor of the reference and break these near-ties in favor of the reference.  @droazen @ldgauthier @yfarjoun if you think this is a good idea I can fix it for both HC and M2.  Otherwise I'll do an M2-only fix.

davidbenjamin:
Closes #4829.

Also cleans up many other implausible deletions in the bamout, a few tens of thousands per genome.

@ldgauthier I would especially like your thoughts on the ungainly use of `useReferenceIfUninformative`.  I wanted to be conservative about switching things like contamination downsampling to the new version, but it's inelegant.


src/main/java/org/broadinstitute/hellbender/utils/read/CigarUtils.java
```diff
@@ -18,6 +18,14 @@
     // these values were chosen via optimization against the NA12878 knowledge base
     public static final SWParameters NEW_SW_PARAMETERS = new SWParameters(200, -150, -260, -11);
 
+    // In Mutect2 and HaplotypeCaller reads are realigned to their *best* haplotypes, which is very different from a generic alignment.
+    // The {@code NEW_SW_PARAMETERS} penalize a substitution error more than an indel up to a length of 9 bases!
+    // Suppose, for example, that a read has a single substitution error, say C -> T, on its last base.  Those parameters
+    // would prefer to extend a deletion until the next T on the reference is found in order to avoid the substitution, which is absurd.
+    // Since these parameters are for aligning a read to the biological sequence we believe it comes from, the parameters
+    // we choose should correspond to sequencer error.  They *do not* have anything to do with the prevalence of true variation!
```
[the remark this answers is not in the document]
davidbenjamin: Basically, I chose values that would prefer a substitution to any indel and prefer a single-base deletion over two substitutions in a row.  You would need to tweak the parameters to get the perfect balance of "do I prefer an n-base deletion over m substitutions?" but remember, we're only using this to align to the *best* haplotype, so if there's any more discrepancy than a couple of bases the GATK probably made a mistake upstream.

Also, the fact that we're aligning to a haplotype that we believe this read actually came from means our parameters should be different from most other implementations of sequence alignment.  Here, our score function assumes errors come from sequencing, whereas in most applications errors are assumed to come from divergent evolution of species or biological mutation.  Thus, for example, these parameters should be much more reluctant to create an indel, at least for Illumina sequencing.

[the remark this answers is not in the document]
davidbenjamin: > Can we trim down this overloading proliferation?

Yes, I would be glad to.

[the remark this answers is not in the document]
davidbenjamin: Did it.

davidbenjamin:
>I'm also seeing that bestAlleles(void) gets called from the RankSumTest, but I don't think that's the right thing to do because we used to exclude uninformative reads.

`RankSumTest` still excludes uninformative reads, in `fillQualsFromLikelihood`.  And the change in this PR never changes an uninformative read into an informative one, or vice versa.

davidbenjamin:
@ldgauthier Done with the changes to overloading and commented back on your questions.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RankSumTest.java
```diff
@@ -79,7 +79,7 @@
     }
 
     protected void fillQualsFromLikelihood(VariantContext vc, ReadLikelihoods<Allele> likelihoods, List<Double> refQuals, List<Double> altQuals, int refLoc) {
-        for (final ReadLikelihoods<Allele>.BestAllele bestAllele : likelihoods.bestAlleles()) {
+        for (final ReadLikelihoods<Allele>.BestAllele bestAllele : likelihoods.bestAllelesBreakingTies()) {
```
ldgauthier: If fillQualsFromLikelihood doesn't change, then why do the rank sum values change in some of the variants in your new expected outputs?  It still excludes uninformative reads, but your new searchBestAllele has _made_ them informative.

davidbenjamin: I need to figure that out.  The new code in `searchBestAllele` is
```
if ( bestLikelihood - referenceLikelihood < BestAllele.INFORMATIVE_THRESHOLD ) {
    secondBestLikelihood = bestLikelihood;
    bestAlleleIndex = referenceAlleleIndex;
    bestLikelihood = referenceLikelihood;
}
```
and so the `BestAllele` constructor gets a best likelihood of `referenceLikelihood` and a second-best likelihood of `bestLikelihood`.  The confidence is then `referenceLikelihood - bestLikelihood`, which is negative and thus less than the informative threshold.

It's also puzzling why the other rank sums don't change.  I'll have to think some more.

davidbenjamin: Ah, I think I've got it.  The new Smith-Waterman parameters affect the pileup for some informative reads when they turn deletions into base errors, which shifts some bases into/out of some pileups.

ldgauthier: Yep, I just came to the same conclusion.  I think changing the pileups is fine.

One last thing: do the likelihoods for uninformative reads get changed more than once?  There are multiple calls to searchBestAllele on the same likelihoods -- e.g. each rank sum test.

davidbenjamin: The likelihoods themselves are never modified, just the `BestAllele`s that are determined from the likelihoods.  That is, every annotation sees the same `ReadLikelihoods` as before the change, and every annotation calculates the `BestAllele` for each read afresh.

ldgauthier: For some reason I thought bestLikelihood and the like were member variables, but clearly they are not, in which case this is fine.

PR #4862, opened 2018-06-08
forgot to delete --infer-onps in oncotator task

davidbenjamin:


davidbenjamin:
@takutosato A very quick PR before we update the Firecloud M2.

PR #4877, opened 2018-06-11
increased boot disk size on GATK tasks in M2 wdl to accommodate 4.0.5.0 docker

davidbenjamin:
@LeeTL1220 For you if tests pass


davidbenjamin:
@LeeTL1220 I copied the attribute name from the oncotator task.

PR #4878, opened 2018-06-11
Added a second layer of deconvolution for pairs that were causing problems in MarkDuplicatesSpark

jamesemery:
Here is the non-string  key solution.


src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtils.java
```diff
@@ -338,6 +339,12 @@ private static long getGroupKey(MarkDuplicatesSparkRecord record) {
         }
     }
 
+    // Note, this uses bitshift operators in order to perform only a single groupBy operation for all the merged data
+    private static long getGroupsForPairs(Pair record) {
```
fleharty: final

src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtils.java
```diff
@@ -338,6 +339,12 @@ private static long getGroupKey(MarkDuplicatesSparkRecord record) {
         }
```
droazen: Are you checking the sequence dictionary on startup to make sure that we don't have > 64k contigs?

src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtils.java
```diff
@@ -313,8 +314,8 @@ public String toString() {
                     nonDuplicates.add(bestFragment);
                 }
 
-                if (Utils.isNonEmpty(pairs)) {
-                    nonDuplicates.add(handlePairs(pairs, finder));
+                for (List<Pair> pairList : pairsStratified) {
```
fleharty: fina?

src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtils.java
cmnbroad: Can this method be made an instance method on MarkDuplicatesSparkRecord?

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/MarkDuplicatesSparkIntegrationTest.java
```diff
@@ -187,6 +187,25 @@ public void testSupplementaryReadUnmappedMate() {
...
 
+    @Test
+    public void testHashCollisionHandling() {
+        File output = createTempFile("supplementaryReadUnmappedMate", "bam");
+        final ArgumentsBuilder args = new ArgumentsBuilder();
+        args.addOutput(output);
+        args.addInput(getTestFile("hashCollisionedReads.bam"));
+        runCommandLine(args);
+
+        try ( final ReadsDataSource outputReadsSource = new ReadsDataSource(output.toPath()) ) {
```
fleharty: Add a comment about this, it's not clear that this bam has reads that have a hash collision.

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/MarkDuplicatesSparkIntegrationTest.java
```diff
@@ -187,6 +187,25 @@ public void testSupplementaryReadUnmappedMate() {
         }
     }
 
+    @Test
+    public void testHashCollisionHandling() {
+        File output = createTempFile("supplementaryReadUnmappedMate", "bam");
```
fleharty: final

jamesemery:
I'm posting an experimental version of this branch for performance purposes. This is no longer in a state where it should be merged.

src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtils.java
jamesemery: I added some checks for libraries and contigs so an exception is thrown in the case where we expect the keys to break.

jamesemery:
@lbergelson I updated this branch with the new key representation. After some performance runs it appears that these lead to a slightly faster mapping operation and approximately 15% less serialization for the step where they are used. Can you take a look?

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/MarkDuplicatesSparkIntegrationTest.java
lbergelson: I think this test is out of date now.

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/ReadsKey.java
```diff
@@ -22,33 +25,124 @@ public static int hashKeyForFragment(int strandedUnclippedStart, boolean reverse
...
+        }
+
+        @Override
+        public String toString() {
+            return fragmentValue + " " + keyValue;
+        }
+    }
+
+
+    private static long longKeyForFragment(int strandedUnclippedStart, boolean reverseStrand, int referenceIndex, byte library) {
```
lbergelson: It still seems weird to encode these as longs when we could either A) keep them as the individual values and let serialization handle it, or B) convert them to a byte array and save a few bits.  We can revisit later if we think it would help I guess.  Metrics seem higher priority than shaving a few more bits here.

src/test/java/org/broadinstitute/hellbender/tools/spark/pipelines/MarkDuplicatesSparkIntegrationTest.java
lbergelson: Could you add tests that generate some keys from the different reads and show that they are not equal for non-equal reads?  Nothing too fancy.

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/ReadsKey.java
```diff
@@ -4,13 +4,16 @@
...
  * Encodes a unique key for read, read pairs and fragments. Used to identify duplicates for MarkDuplicatesGATK.
  *
  * This class was changed to primarily operate on key hashing instead of generating long string keys as it was discovered
  * that it had performance implications for serialization in MarkDuplicatesSpark. 
  */
-public final class ReadsKey {
+public class ReadsKey {
 
     /**
      * Makes a hash key for the fragment.
```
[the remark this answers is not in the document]
jamesemery: huh... I forgot to push a branch that had this change already in it.

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/ReadsKey.java
lbergelson: Do we have a list written somewhere?

src/test/java/org/broadinstitute/hellbender/utils/read/markduplicates/ReadsKeyUnitTest.java
```diff
@@ -75,4 +95,70 @@ public void testKeyForFragmentPerformanceEvidence() throws Exception {
...
+                        createTestRead("name1", "1", 1200, "100M", library1.getReadGroupId(), true),
+                        false,
+                        createTestRead("name2", "1", 1000, "100M", library2.getReadGroupId(), false),
+                        createTestRead("name2", "1", 1200, "100M", library2.getReadGroupId(), true),},
+
+                // Test of two equivalent groups, but one was softclipped to a different start
+                {header, createTestRead("name1", "1", 1010, "10S90M", library1.getReadGroupId(), false),
+                        createTestRead("name1", "1", 1200, "100M", library1.getReadGroupId(), true),
+                        true,
+                        createTestRead("name2", "1", 1000, "100M", library1.getReadGroupId(), false),
```
lbergelson: add a case for read 2 on a different contig, otherwise the same

src/main/java/org/broadinstitute/hellbender/tools/spark/transforms/markduplicates/MarkDuplicatesSparkUtils.java
```diff
@@ -178,11 +178,12 @@ public String toString() {
...
     public static Map<String, Byte> constructLibraryIndex(final SAMFileHeader header) {
         final List<String> discoveredLibraries = header.getReadGroups().stream()
-                .map(r -> {String library = r.getLibrary(); return library==null? LibraryIdGenerator.UNKNOWN_LIBRARY : library;} )
+                .map(r -> { String library = r.getLibrary();
+                            return library==null? LibraryIdGenerator.UNKNOWN_LIBRARY : library;} )
                 .distinct()
                 .collect(Collectors.toList());
         if (discoveredLibraries.size() > 255) {
-            throw new GATKException("Detected too many read libraries among read groups header, currently MarkDuplciatesSpark only supports up to 256 unique readgroup libraries");
+            throw new GATKException("Detected too many read libraries among read groups header, currently MarkDuplciatesSpark only supports up to 256 unique readgroup libraries but " + discoveredLibraries.size() + " were found");
```
lbergelson: typo MarkDuplciates

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/ReadsKey.java
```diff
@@ -101,48 +76,51 @@ public String toString() {
...
         public String toString() {
-            return fragmentValue + " " + keyValue;
+            return firstReadKeyValue + " " + secondReadKeyValue;
         }
     }
 
-
+    // Helper methods for generating summary longs
     private static long longKeyForFragment(int strandedUnclippedStart, boolean reverseStrand, int referenceIndex, byte library) {
         return (((long)strandedUnclippedStart) << 32) |
```
lbergelson: could you add a comment explaining the reason you need to mask these, it wasn't obvious to us at first, so it probably won't be when we've come back to look at it later.

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/sparkrecords/Passthrough.java
```diff
@@ -14,8 +13,8 @@
     Passthrough(GATKRead read, int partitionIndex) {
         super(partitionIndex, read.getName());
 
-        // use a
-        this.key = ReadsKey.hashKeyForRead(read);
+        // use a hash key here instead of a normal key because collisions don't matter here
```
lbergelson: 👍

PR #4889, opened 2018-06-12
Modify the behavior of getNextReferenceVertex for non-ref paths

DonFreed:
There was a subtle change in the behavior of getNextReferenceVertex in the migration from GATK 3.x to the GATK 4.x that may be worth another look.

The original 3.x behavior is described in a comment of the function, if `@param allowNonRefPaths` is true, allow sub-paths that are non-reference _if there is only a single outgoing edge_. The 4.x migration copied the comment, but changed the behavior to if `@param allowNonRefPaths` is true, allow sub-paths that are non-reference _and pick a random outgoing edge_. 

This pull request restores the 3.x behavior, which we believe is the design intent.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/BaseGraph.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/BaseGraph.java
@@ -225,8 +225,8 @@ public abstract class BaseGraph<V extends BaseVertex, E extends BaseEdge> extend
         final Set<E> blacklistedEdgeSet = blacklistedEdge.isPresent() ? Collections.singleton(blacklistedEdge.get()) : Collections.emptySet();
 
         // if we got here, then we aren't on a reference path
-        final Optional<E> edge = outgoingEdges.stream().filter(e -> !blacklistedEdgeSet.contains(e)).findAny();
-        return edge.isPresent() ? getEdgeTarget(edge.get()) : null;
+        final List<E> edges = outgoingEdges.stream().filter(e -> !blacklistedEdgeSet.contains(e)).limit(2).collect(Collectors.toList());
+        return edges.size() == 1 ? getEdgeTarget(edges.get(0)) : null;
     }
 
     /**


lbergelson:
@davidbenjamin Are you able to review this?

davidbenjamin:
@lbergelson Sure!

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/graphs/BaseGraph.java
```diff
@@ -225,8 +225,8 @@ public final V getNextReferenceVertex( final V v, final boolean allowNonRefPaths
         final Set<E> blacklistedEdgeSet = blacklistedEdge.isPresent() ? Collections.singleton(blacklistedEdge.get()) : Collections.emptySet();
 
         // if we got here, then we aren't on a reference path
-        final Optional<E> edge = outgoingEdges.stream().filter(e -> !blacklistedEdgeSet.contains(e)).findAny();
-        return edge.isPresent() ? getEdgeTarget(edge.get()) : null;
+        final List<E> edges = outgoingEdges.stream().filter(e -> !blacklistedEdgeSet.contains(e)).limit(2).collect(Collectors.toList());
+        return edges.size() == 1 ? getEdgeTarget(edges.get(0)) : null;
```
davidbenjamin: Getting the first edge is still essentially random in that it depends on the order in which reads arrived.  Why not be principled and traverse the edge with the greatest weight, which seems like a good heuristic if you're trying to arrive back at the reference?
```
final Optional<E> edge = outgoingEdges.stream().max(Comparator.comparingInt(BaseEdge::getMultiplicity).reversed())
```

davidbenjamin: Sorry, I misread the code.  But I still think my suggestion to do a greedy search before giving up on finding the reference is reasonable and more robust to things like high depth where you might have lots of low-weight non-reference edges.

DonFreed: Right, the current behavior is to pick a random edge. The PR reverts to the 3.x behavior, which is to return a non-reference edge iff there 1 non-reference edge.

It would be great to try a more exhaustive search. However, my intuition is that the we are unlikely to make it back to the reference when the paths start forking and the computational cost can be substantial. We can use this PR to revert back to the deterministic logic and save implementation (and evaluation) of a greedy search for a future issue.

DonFreed: @davidbenjamin thoughts?

davidbenjamin: @DonFreed I'm fine with the 3.x behavior.  Anything fancier can wait for a later thorough revision of the dangling end merging code.

lbergelson:
Thank you @DonFreed!

 @davidbenjamin for reviewing, if you feel we should make the changes you've suggested could you open an issue so we don't forget?

PR #4899, opened 2018-06-15
Fixed a missing special case in MarkDuplicates ReadsKey code

jamesemery:


src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/sparkrecords/Pair.java
```diff
@@ -55,21 +55,36 @@ public Pair(final GATKRead read1, final GATKRead read2, final SAMFileHeader head
         final int read1UnclippedStart = ReadUtils.getStrandedUnclippedStart(read1);
         final int read2UnclippedStart = ReadUtils.getStrandedUnclippedStart(read2);
 
-        if( read1UnclippedStart < read2UnclippedStart ){
+        int read1ReferenceIndex = ReadUtils.getReferenceIndex(read1,header);
+        int read2ReferenceIndex = ReadUtils.getReferenceIndex(read2,header);
+
+        if( read1ReferenceIndex != read2ReferenceIndex ? read1ReferenceIndex < read2ReferenceIndex : read1UnclippedStart <= read2UnclippedStart ){
```
lbergelson: Wouldn't this would be easier to read as 
```
if( read1ReferenceIndex <= read2ReferenceIndex && read1UnclippedStart <= read2UnclippedStart) {
```

unless I missed something there that makes them not equivalent

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/sparkrecords/Pair.java
```diff
@@ -55,21 +55,36 @@ public Pair(final GATKRead read1, final GATKRead read2, final SAMFileHeader head
...
+            second = tmp;
+        }
+
+        this.key = ReadsKey.getKeyForPair(header, first, second, headerLibraryMap);
+
         isRead1ReverseStrand = first.isReverseStrand();
         isRead2ReverseStrand = second.isReverseStrand();
 
+        // Keep track of the orientation of read1 and read2 as it is important for optical duplicate marking
+        wasFlipped = second.isFirstOfPair(); //TODO be sure this is still correct...
```
lbergelson: todo worries me

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/sparkrecords/Pair.java
```diff
@@ -55,21 +55,36 @@ public Pair(final GATKRead read1, final GATKRead read2, final SAMFileHeader head
...
         } else {
             first = read2;
             second = read1;
         }
-        // Keep track of the orientation of read1 and read2 as it is important for optical duplicate marking
-        wasFlipped = second.isFirstOfPair();
-
         firstStartPosition = first.getAssignedStart();
 
+        // Ensuring that the orientation is consistent for reads starting at the same place there is no propper order
```
lbergelson: I think the comment in picard is clearer

lbergelson: // if the two read ends are in the same position, pointing in opposite directions,
                            // the orientation is undefined and the procedure above
                            // will depend on the order of the reads in the file.
                            // To avoid this, we set it explicitly (to FR):

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/sparkrecords/Pair.java
lbergelson: There should be a test for this edge case that we were missing.

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/sparkrecords/Pair.java
lbergelson: also, propper -> proper

src/main/java/org/broadinstitute/hellbender/utils/read/markduplicates/sparkrecords/Pair.java
lbergelson: Oh, heh, you're write.  I didn't think that through.

jamesemery:
@lbergelson back to you

Issue #4901, opened 2018-06-15
Inconsistent unmapped but placed query results


cmnbroad:
When querying bams by interval, htsjdk only returns unmapped, placed reads if the read's start position overlaps the query interval, whereas the read's mapped mate is returned if any part of it overlaps the interval. Unclear if this has an affect on things like finding breakpoint evidence, but it can result in unintentionally splitting up pairs, and can give subtly inconsistent results such as when querying the reads that back a variant, since sometimes a given query can include unmapped reads, sometimes not, and sometimes have a mix.

lbergelson:
This seems really bad.

cmnbroad:
@lbergelson Yeah, it would seem more rational to always keep them together. It would require some lookahead/state to use the same interval for both, since empirically there seems to be no ordering maintained between the reads in a pair.

cmnbroad:
Just noticed that there is an old open ticket for this in htsjdk https://github.com/samtools/htsjdk/issues/278.

PR #4902, opened 2018-06-15
Engine read caching.

cmnbroad:
Allow tools to opt-in for read caching (enabled in this PR for CNNScoreVariants and FilterAlignmentArtifacts) when issuing forward-only queries such as those used to query a ReadsContext for all reads backing each variant.

The cache matches the (sometimes surprising - see https://github.com/broadinstitute/gatk/issues/4901) results that are returned when caching is not used, specifically that an unmapped but placed read that is mated with a mapped read is only returned if the *start position* overlaps the interval query interval, whereas the mapped read is returned if any part of the read overlaps the query interval.


src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections/ReadInputArgumentCollection.java
```diff
@@ -37,6 +38,11 @@
               optional = true)
     protected List<String> readIndices;
 
+    @Advanced
+    @Argument(fullName = "reads-look-ahead-window-size", shortName="reads-look-ahead-window-size",
+                doc = "Window size for reads look-ahead query caching")
+    protected int readsLookaheadWindowSize = 100 * 1024;
```
magicDGS: I don't think that this should be exposed to the user, but being similar to the feature-cache: it should be choosen by the walker-class what is the best value.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -508,7 +511,16 @@ public boolean requiresReads() {
     }
 
     /**
-     * Does this tool require intervals? Traversals types and/or tools that do should override to return true.
+     * Should this tool use read caching to optimize forward-only queries on coordinate sorted read inputs ?
+     *
+     * @return true if read caching should be used, otherwise false
+     */
+    public boolean useReadCaching() {
```
magicDGS: This could be a `getReadsLookaheadWindowSize` instead, returning a value with the meaning of disable (0? -1?) and let each walker-class to choose how the cache is handled. A tool can always implement an argument to modify this value.

cmnbroad:
Rebased this to do some testing with current master.

src/test/java/org/broadinstitute/hellbender/engine/cache/LocatableCacheUnitTest.java
```diff
@@ -0,0 +1,431 @@
...
+                emptyRegion,
+                "Should get back empty List for empty region");
+    }
+
+    /***************************************************************************************************
+     * End of LocatableCache tests using DrivingFeatureInputCacheStrategy to test basic cache operations
+     ***************************************************************************************************/
+
+    /********************************************************
+     * LocatableCache tests using SideReadInputCacheStrategy
```
cmnbroad, line 311: This test is really testing `SideReadInputCacheStrategy`,so maybe it should be moved to a SideReadInputCacheStrategyUnitTest class.

PR #4915, opened 2018-06-19
XsvTableFeatures now always put out the right number of columns.

jonn-smith:
Fixes #4193

`XsvTableFeature` no longer removes an extra column if start and end in
the config file for a `LocatableXsv` data source are the same.


jonn-smith:
Yeah - the files I needed are all in there and the tests that rely on them pass.

Merging...

PR #4917, opened 2018-06-20
Added custom MAF fields to MafOutputRenderer

### LINKED ISSUE #4634: Funcotator - Must implement t_alt_count / t_ref_count, alt_count/ref_count
Somehow the annotations for  t_alt_count / t_ref_count and alt_count/ref_count are missing.

They need to be implemented.

The way to properly do this is to pull them out of the genotyped samples, but you must know which is ref and which is alt.  This information is primarily in the VCF header, then by sample name (variations on "matched reference" / "alternate").  This is the priority in which they should be determined.

From there you can check the field for the counts.

There is also a field in the INFO field that needs to be used for this information.

LeeTL1220:
- Implements t_alt_count, t_ref_count, n_alt_count, n_ref_count, and tumor_f fields for MAF only. Rules for automatic determination of pairs is implemented to determine which is the tumor and which is the normal.
- Closes #4634


LeeTL1220:
@jonn-smith I am going to need to discuss with you how to fix the remaining broken test.  I believe that the the issue is with the design of the test itself, not the code in this PR.  There is an implicit assumption that the MafOutputRenderer will never create a Funcotation, which this PR violates.  I believe that we should eliminate this assumption.

LeeTL1220:
@jonn-smith I put in a fix.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/metadata/SamplePairExtractor.java
```diff
@@ -0,0 +1,67 @@
...
+            final VCFHeaderLine normalMetaDataLine = vcfHeader.getMetaDataLine(Mutect2Engine.NORMAL_SAMPLE_KEY_IN_VCF_HEADER);
+            return Collections.singletonList(new TumorNormalPair(vcfHeader.getMetaDataLine(Mutect2Engine.TUMOR_SAMPLE_KEY_IN_VCF_HEADER).getValue(),
+                    (normalMetaDataLine == null ?
+                            NO_NORMAL : normalMetaDataLine.getValue())));
+        }
+
+        // Then try sample names (e.g. "TUMOR", "NORMAL")
+        final List<String> sampleNames = vcfHeader.getSampleNamesInOrder();
+        if (sampleNames.size() == 1) {
+            return Collections.singletonList(new TumorNormalPair(sampleNames.get(0), NO_NORMAL));
```
jonn-smith: As long as this doesn't happen for the germline case, I'm fine with assuming the first sample is the tumor sample.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/mafOutput/MafOutputRenderer.java
```diff
@@ -116,6 +121,18 @@ public MafOutputRenderer(final Path outputFilePath,
         this.inputFileHeader = inputFileHeader;
         dataSourceFactories = dataSources;
 
+        this.tnPairs = SamplePairExtractor.extractPossibleTumorNormalPairs(this.inputFileHeader);
```
jonn-smith: Can you add in a new comment in the constructor and the output format argument for MAF saying that this should not be used for Germline data?

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
```diff
@@ -399,7 +407,7 @@ public void exhaustiveArgumentTest(final String dataSourcesPath,
     }
 
     /**
-     * Test that we can annotate a hg19 datasource when GENCODE is using "chr*" and the datasource is not.
+     * Test that we can annotate a b37 (here it is clinvar) datasource when GENCODE is using "chr*" and the datasource is not.
```
jonn-smith: Good catch.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/metadata/SamplePairExtractor.java
LeeTL1220: This does happen in the case of single-sample germline VCF.  As per the rules I laid out, if it sees one sample, it will assume that is the tumor sample.  I do not really see this as an issue, though.  Anyone generating a MAF from a germline single sample is going to ignore the t_alt_count, etc.  And in the case of multi-sample, those fields would be blank anyway.

No action.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/mafOutput/MafOutputRenderer.java
LeeTL1220: Can I just say it is unsupported?  People are welcome to use it -- might be useful with some big caveats....

Done.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
LeeTL1220: Thanks... No action.

PR #4926, opened 2018-06-20
Clone read base qualities rather than reference them

DonFreed:
When adjusting the base quality of overlapping read pairs, the modifications are made in place. If the modified reads are later used in another active region, the results from the later active region will be changed by the earlier modification.

This pull request clones the read's base quality, so that in place modifications do not affect later active regions.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
@@ -86,7 +86,7 @@ public final class AssemblyBasedCallerUtils {
             if ( ! clippedRead.isEmpty() && clippedRead.getCigar().getReadLength() > 0 ) {
                 clippedRead = ReadClipper.hardClipToRegion( clippedRead, region.getExtendedSpan().getStart(), region.getExtendedSpan().getEnd() );
                 if ( region.readOverlapsRegion(clippedRead) && clippedRead.getLength() > 0 ) {
-                    readsToUse.add(clippedRead);
+                    readsToUse.add((clippedRead == myRead) ? clippedRead.deepCopy() : clippedRead);
                 }
             }
         }
--- a/src/main/java/org/broadinstitute/hellbender/utils/fragments/FragmentUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/fragments/FragmentUtils.java
@@ -45,9 +45,9 @@ public final class FragmentUtils {
         final int numOverlappingBases = Math.min(clippedFirstRead.getLength() - firstReadStop, clippedSecondRead.getLength());
 
         final byte[] firstReadBases = clippedFirstRead.getBases();
-        final byte[] firstReadQuals = clippedFirstRead.getBaseQualities();
+        final byte[] firstReadQuals = clippedFirstRead.getBaseQualities().clone();
         final byte[] secondReadBases = clippedSecondRead.getBases();
-        final byte[] secondReadQuals = clippedSecondRead.getBaseQualities();
+        final byte[] secondReadQuals = clippedSecondRead.getBaseQualities().clone();
 
         for ( int i = 0; i < numOverlappingBases; i++ ) {
             final int firstReadIndex = firstReadStop + i;
--- a/src/main/java/org/broadinstitute/hellbender/utils/fragments/FragmentUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/fragments/FragmentUtils.java
@@ -45,9 +45,9 @@ public final class FragmentUtils {
         final int numOverlappingBases = Math.min(clippedFirstRead.getLength() - firstReadStop, clippedSecondRead.getLength());
 
         final byte[] firstReadBases = clippedFirstRead.getBases();
-        final byte[] firstReadQuals = clippedFirstRead.getBaseQualities().clone();
+        final byte[] firstReadQuals = clippedFirstRead.getBaseQualities();
         final byte[] secondReadBases = clippedSecondRead.getBases();
-        final byte[] secondReadQuals = clippedSecondRead.getBaseQualities().clone();
+        final byte[] secondReadQuals = clippedSecondRead.getBaseQualities();
 
         for ( int i = 0; i < numOverlappingBases; i++ ) {
             final int firstReadIndex = firstReadStop + i;


src/main/java/org/broadinstitute/hellbender/utils/fragments/FragmentUtils.java
```diff
@@ -45,9 +45,9 @@ public static void adjustQualsOfOverlappingPairedFragments(final GATKRead clippe
         final int numOverlappingBases = Math.min(clippedFirstRead.getLength() - firstReadStop, clippedSecondRead.getLength());
 
         final byte[] firstReadBases = clippedFirstRead.getBases();
-        final byte[] firstReadQuals = clippedFirstRead.getBaseQualities();
+        final byte[] firstReadQuals = clippedFirstRead.getBaseQualities().clone();
```
magicDGS: `GATKRead.getBases()` and `GATKRead.getBaseQualities()` are safe methods and return a copy of the array. Thus, it is not really modified in place. I think that this patch is not needed for that reason.

lbergelson:
@DonFreed, I agree with @magicDGS's assessment about it.  This feels like a fix that was applied to Gatk3 but doesn't translate to 4?  Of course, there could be implementations of GATKRead that don't obey the given contract about copying, but it's worth fixing those since we were more careful to think about copy/no copy when we wrote the new interface.  

Of note: if you haven't seen it, `GATKRead` provides a set of unsafe `getBaseQualitiesNoCopy()` methods for times when the copy is a performance bottleneck and you can guarantee safe use of the underlying array.  

I'm going to close this.  Feel free to reopen if you disagree / can provide a unit test that demonstrates the issue still exists.

DonFreed:
@lbergelson and @magicDGS, thank you for correcting me here. You are both right, this PR is based on an outdated understanding of the behavior of GATKRead.getBaseQualities(). Thanks for closing it!

lbergelson:
@donfreed Thank you for your continued interest and additions to GATK!

DonFreed:
@lbergelson Could we reopen this issue? We were able to track down the source of the modified BQs across active regions.

At the end of the function `adjustQualsOfOverlappingPairedFragments(final GATKRead clippedFirstRead, final GATKRead clippedSecondRead)`, the base qualities are set in the clipped reads:
```
clippedFirstRead.setBaseQualities(firstReadQuals);
clippedSecondRead.setBaseQualities(secondReadQuals);
```

In some cases, the clipped read is actually the original read, modifying the original read's BQ. I've committed a simple fix to my branch that ensures that the clipped read is always a copy of the original read.

droazen:
@lbergelson Can you re-review?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
```diff
@@ -86,7 +86,7 @@ public static void finalizeRegion(final AssemblyRegion region,
             if ( ! clippedRead.isEmpty() && clippedRead.getCigar().getReadLength() > 0 ) {
                 clippedRead = ReadClipper.hardClipToRegion( clippedRead, region.getExtendedSpan().getStart(), region.getExtendedSpan().getEnd() );
                 if ( region.readOverlapsRegion(clippedRead) && clippedRead.getLength() > 0 ) {
-                    readsToUse.add(clippedRead);
+                    readsToUse.add((clippedRead == myRead) ? clippedRead.deepCopy() : clippedRead);
```
droazen: I wonder if we can get away with a less-expensive shallow copy (ie., `copy()`) rather than a costlier `deepCopy()` here?

Also, can you add a unit test that fails before the fix and passes after it?

droazen:
Re-assigning to @jamesemery

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtilsUnitTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @BeforeClass
+    public void init() {
+        header = ArtificialReadUtils.createArtificialSamHeader(1, 1, 100000000);
+        SAMReadGroupRecord rg = new SAMReadGroupRecord("tumor");
+        rg.setSample("tumor");
+        header.addReadGroup(rg);
+    }
+
+    @Test
+    public void testfinalizeRegion() {
```
jamesemery: This test looks good, though in this case since there is only one test in the AssemblyBasedCallerUtilsUnitTests I wouldn't use the `@BeforeClass` to initialize files. Instead just put them into static fields or build the header etc... into this test.

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtilsUnitTest.java
```diff
@@ -0,0 +1,56 @@
...
+        SAMRecord orgRead0 = parser.parseLine("HWI-ST807:461:C2P0JACXX:4:2204:18080:5857\t83\t1\t42596803\t39\t1S95M5S\t=\t42596891\t-7\tGAATCATCATCAAATGGAATCTAATGGAATCATTGAACAGAATTGAATGGAATCGTCATCGAATGAATTGAATGCAATCATCGAATGGTCTCGAATAGAAT\tDAAAEDCFCCGEEDDBEDDDGCCDEDECDDFDCEECCFEECDCEDBCDBDBCC>DCECC>DBCDDBCBDDBCDDEBCCECC>DBCDBDBGC?FCCBDB>>?\tRG:Z:tumor");
+        SAMRecord orgRead1 = parser.parseLine("HWI-ST807:461:C2P0JACXX:4:2204:18080:5857\t163\t1\t42596891\t39\t101M\t=\t42596803\t7\tCTCGAATGGAATCATTTTCTACTGGAAAGGAATGGAATCATCGCATAGAATCGAATGGAATTAACATGGAATGGAATCGAATGTAATCATCATCAAATGGA\t>@>:ABCDECCCEDCBBBDDBDDEBCCBEBBCBEBCBCDDCD>DECBGCDCF>CCCFCDDCBABDEDFCDCDFFDDDG?DDEGDDFDHFEGDDGECB@BAA\tRG:Z:tumor");
+
+        reads.add(new SAMRecordToGATKReadAdapter(orgRead0.deepCopy()));
+        reads.add(new SAMRecordToGATKReadAdapter(orgRead1.deepCopy()));
+        activeRegion.addAll(reads);
+
+        SampleList sampleList = SampleList.singletonSampleList("tumor");
+        Byte minbq = 9;
+        AssemblyBasedCallerUtils.finalizeRegion(activeRegion, false, false, minbq, header, sampleList);
```
jamesemery: You should add a comment explaining what this test is doing here, (like explaining why it was failing due to unsafe operations on original reads and read qualities in the engine) so people know what is going on down the line.

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtilsUnitTest.java
jamesemery: Also this is a good test

DonFreed:
Thanks @jamesemery! I've updated the PR to address your comments.

PR #4927, opened 2018-06-21
Additional caching speed fixes for Funcotator.

### LINKED ISSUE #4586: Funcotator - Need Performance Upgrades

jonn-smith:
Fixes #4586

Released new version of datasources to go with this release (1.4.20180615).
This was necessary because the data sources needed to be made
consistent with hg19 (before they were a mix of hg19 and b37
contig names).

Now Funcotator assumes all data sources for the hg19 reference are
compliant with hg19 contig names.

Updated the minimum data source version to the new release (1.4.20180615).

Simplified `Funcotator::enqueueAndHandleVariant`.

Not clear that the `--allow-hg19-gencode-b37-contig-matching-override`
flag does anything anymore.

Updated the `getDbSNP.sh` and `createSqliteCosmicDb.sh` data source
scripts to preprocess those data sources to be have hg19-compliant
contigs names.

New speeds are ~20k variants/minute for hg19 and ~200k variants/minute for hg38


src/main/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorUtils.java
```diff
@@ -1482,6 +1486,8 @@ public static String getBasesInWindowAroundReferenceAllele( final Allele refAlle
                                                                 final ReferenceContext referenceContext) {
         // TODO: This is generating too long a string for INDELs because of VCF format alleles - see Issue #4407
 
+        // TODO: this seems to be the same as GencodeFunotationFactory::getReferenceBases - should that method call into this?
```
LeeTL1220: Can we determine an answer and apply a fix?  Otherwise, file an issue and reference it here.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/gencode/GencodeFuncotationFactory.java
```diff
@@ -1292,6 +1293,8 @@ private GencodeFuncotation createIntronFuncotation(final VariantContext variant,
     @VisibleForTesting
     static String getReferenceBases(final Allele refAllele, final Allele altAllele, final ReferenceContext reference, final Strand strand ) {
 
+        // TODO: this seems to be the same as FuncotatorUtils::getBasesInWindowAroundReferenceAllele - should this method call into that?
```
LeeTL1220: See my previous comment.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/gencode/GencodeFuncotationFactory.java
```diff
@@ -1306,12 +1309,29 @@ static String getReferenceBases(final Allele refAllele, final Allele altAllele,
                     reference.getWindow().getStart() - referenceWindow,
                     reference.getWindow().getEnd() + referenceWindow + indelAdjustment);
 
+        // Get the reference bases for this interval.
+        // This grossness is required for now as a fix for the hg19 vs b37 reference contig name issue:
+        byte[] referenceBases;
+        try {
+            referenceBases = reference.getBases(refBasesInterval);
+        }
+        catch (final UserException.MissingContigInSequenceDictionary | UserException.NoDataAtRequestedContig ex) {
```
LeeTL1220: We should never include throwing exceptions, except in erroneous code paths.  This goes double for code that will be run per variant.
Detect whether there will be a missing contig and then use an `if` statement, please.  Or some similar approach.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/gencode/GencodeFuncotationFactory.java
```diff
@@ -1584,10 +1604,11 @@ public static double calculateGcContent( final Allele refAllele,
                                              final ReferenceContext referenceContext,
                                              final int windowSize ) {
 
+        // TODO: this seems to do something similar to FuncotatorUtils::getBasesInWindowAroundReferenceAllele - should this method call into that?
```
LeeTL1220: See previous comment.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/gencode/GencodeFuncotationFactory.java
```diff
@@ -1610,10 +1631,23 @@ public static double calculateGcContent( final Allele refAllele,
...
         // Get the bases:
-        bases = referenceContext.getBases(leadingWindowSize, trailingWindowSize);
+        // This grossness is required for now as a fix for the hg19 vs b37 reference contig name issue:
+        try {
+            bases = referenceContext.getBases(leadingWindowSize, trailingWindowSize);
+        }
+        catch (final UserException.MissingContigInSequenceDictionary | UserException.NoDataAtRequestedContig ex) {
+
+            final SimpleInterval otherPossibleRefInterval = new SimpleInterval(
+                    FuncotatorUtils.convertHG19ContigToB37Contig(referenceContext.getWindow().getContig()),
```
LeeTL1220: Can we use an `if` statement here?  (see previous comment)

src/main/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorUtils.java
jonn-smith: This is the "ugliness" I mentioned to you in person.  My plan is to file an issue and keep this in for now so the fix can go in sooner.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/gencode/GencodeFuncotationFactory.java
jonn-smith: Will commit and PR after.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/gencode/GencodeFuncotationFactory.java
jonn-smith: Will commit and PR after.

src/main/java/org/broadinstitute/hellbender/engine/FeatureCache.java
```diff
@@ -223,11 +223,23 @@ public void trimToNewStartPosition( final int newStart ) {
...
      */
     public void printCacheStatistics() {
+        printCacheStatistics("");
+    }
+
+    /**
+     * Print statistics about the cache hit rate for debugging.
+     * @param sourceName The source for the features in this cache.
+     */
+    public void printCacheStatistics(final String sourceName) {
```
cmnbroad: I like the sourceName idea - would support making this a constructor arg since this could for error messages, etc. Alternatively its something I can do in my caching PR it not here.

jonn-smith: @cmnbroad Sounds good.  I'd like to get this one in sooner rather than later - would you mind putting it in your branch?

src/main/java/org/broadinstitute/hellbender/tools/funcotator/dataSources/gencode/GencodeFuncotationFactory.java
jonn-smith: Totally obviated via changes discussed in person.

src/main/java/org/broadinstitute/hellbender/engine/FeatureCache.java
cmnbroad: @jonn-smith Sure, I can do it in my branch.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -137,17 +138,17 @@
     /**
      * Our source of reference data (null if no reference was provided)
      */
-    ReferenceDataSource reference;
+    protected ReferenceDataSource reference;
```
magicDGS: Finally these fields get protected. I was looking forward to access the data sources and I was planning to do a PR for this. But it won't be better to have a getter that throws if `null`? This will prevent users of `GATKTool` without `require*` methods returning `true` to use a field that isn't set (and get rid of implementation discrepancies...)

src/main/java/org/broadinstitute/hellbender/engine/ReferenceContext.java
```diff
@@ -352,7 +352,7 @@ private int trimToContigLength(final String contig, final int end){
 
         final SAMSequenceRecord sequence = dataSource.getSequenceDictionary().getSequence(contig);
         if ( sequence == null ) {
-            throw new UserException("Given reference file does not have data at the requested contig(" + contig + ")!");
+            throw new UserException.NoDataAtRequestedContig("Given reference file does not have data at the requested contig(" + contig + ")!");
```
magicDGS: With the current implementation of `NoDataAtRequestedContig` you should only pass `contig` to the constructor. Otherwise the message will be weird.

jonn-smith: Yup.  Fixed!

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -300,12 +292,14 @@
     )
     private int lookaheadFeatureCachingInBp = FuncotatorArgumentDefinitions.LOOKAHEAD_CACHE_IN_BP_DEFAULT_VALUE;
 
+    @Advanced
+    @Hidden
     @Argument(
-            fullName = FuncotatorArgumentDefinitions.ALLOW_HG19_GENCODE_B37_CONTIG_MATCHING_OVERRIDE_LONG_NAME,
+            fullName = FuncotatorArgumentDefinitions.FORCE_B37_TO_HG19_REFERENCE_CONTIG_CONVERSION,
```
LeeTL1220: Thanks for catching this.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -383,38 +386,46 @@ public void onTraversalStart() {
...
-            inputReferenceIsB37 = FuncotatorUtils.isSequenceDictionaryUsingB37Reference(getSequenceDictionaryForDrivingVariants());
+            // so the lengths of the contigs are the same, the bases are just _slightly_ different.
+            // ALSO, the contig naming convention is different between hg19 and hg38:
+            //      hg19 uses contigs of the form "chr1"
+            //      b37 uses contigs of the form  "1"
+            // This naming convention difference causes a LOT of issues and was a bad idea.
+            inputReferenceIsB37 = forceB37ToHg19ContigNameConversion || FuncotatorUtils.isSequenceDictionaryUsingB37Reference(getSequenceDictionaryForDrivingVariants());
             if ( inputReferenceIsB37 ) {
                 logger.warn("WARNING: You are using B37 as a reference.  " +
                         "Funcotator will convert your variants to GRCh37, and this will be fine in 99.9% of cases.  " +
```
LeeTL1220: Can you say "vast majority"?  "99.9%" makes it seem like you really studied this and derived an actual number....

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -532,8 +549,9 @@ private void enqueueAndHandleVariant(final VariantContext variant, final Referen
                 .flatMap(List::stream)
                 .map(gf -> (GencodeFuncotation) gf).collect(Collectors.toList());
 
-
         //==============================================================================================================
+        // Create the rest of the funcotations:
+
```
LeeTL1220: Extra blank line here.

LeeTL1220: Actually, comments seem incorrect.  Please correct.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
jonn-smith: Sure.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jonn-smith: I like the idea, but I think those methods would have to query the `require*` methods to see if they should throw anyway.  I'd prefer to get this in and then fix the interfaces at a later. 

@magicDGS - can you add fixing the interfaces to these objects as an issue?

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
jonn-smith: no prob.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
jonn-smith: I went over the comments and made them a little more descriptive.

I prefer to leave in the extra newline so that from a visual perspective the sections are a little more clear.  I made the dividers formatted the same way.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -378,32 +386,89 @@ public void onTraversalStart() {
         }
 
         // Check for reference version (in)compatibility:
-        if ( allowHg19ContigNamesWithB37 &&
-             referenceVersion.equals(FuncotatorArgumentDefinitions.HG19_REFERENCE_VERSION_STRING) ) {
+        if ( forceB37ToHg19ContigNameConversion || referenceVersion.equals(FuncotatorArgumentDefinitions.HG19_REFERENCE_VERSION_STRING) ) {
```
droazen: I think that you need to check in `onTraversalStart()` that the reference passed in via `-R` and the sequence dictionary for the set of variants you're annotating (`-V`) are both B37 or both HG19, if either of them is B37/HG19. There is a sequence dictionary compatibility check in the engine, but it's not necessarily sufficient to detect the case of an HG19 fasta + a B37 VCF (for example), since it only checks that there is at least one shared contig between dictionaries.

It would also be good (while you're on this topic) to explore whether the current check in `GATKTool.validateSequenceDictionaries()` would sometimes conclude that a B37 and HG19 dictionary were compatible. If it would, you should file a ticket against it.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorArgumentDefinitions.java
```diff
@@ -48,16 +48,13 @@
...
-    public static final String ALLOW_HG19_GENCODE_B37_CONTIG_MATCHING_LONG_NAME = "allow-hg19-gencode-b37-contig-matching";
-
     public static final String HG19_REFERENCE_VERSION_STRING = "hg19";
     public static final String HG38_REFERENCE_VERSION_STRING = "hg38";
 
     public static final String LOOKAHEAD_CACHE_IN_BP_NAME = "lookahead-cache-bp";
     public static final int LOOKAHEAD_CACHE_IN_BP_DEFAULT_VALUE = VariantWalkerBase.FEATURE_CACHE_LOOKAHEAD;
 
-    public static final String ALLOW_HG19_GENCODE_B37_CONTIG_MATCHING_OVERRIDE_LONG_NAME = "allow-hg19-gencode-b37-contig-matching-override";
+    public static final String FORCE_B37_TO_HG19_REFERENCE_CONTIG_CONVERSION = "force-b37-to-hg19-reference-contig-lookup";
```
droazen: `lookup` -> `conversion`?

src/main/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorUtils.java
```diff
@@ -1492,9 +1498,26 @@ public static String getBasesInWindowAroundReferenceAllele( final Allele refAlle
 
         final String referenceBases;
 
+        // Get the reference bases for this interval.
+        // This grossness is required for now as a fix for the hg19 vs b37 reference contig name issue:
```
droazen: This hack can now be removed, as discussed in person.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
```diff
@@ -422,7 +451,9 @@ public void testCanAnnotateMixedContigHg19Clinvar() {
         arguments.addArgument(FuncotatorArgumentDefinitions.DATA_SOURCES_PATH_LONG_NAME, DS_PIK3CA_DIR);
         arguments.addArgument(FuncotatorArgumentDefinitions.REFERENCE_VERSION_LONG_NAME, FuncotatorTestConstants.REFERENCE_VERSION_HG19);
         arguments.addArgument(FuncotatorArgumentDefinitions.OUTPUT_FORMAT_LONG_NAME, outputFormatType.toString());
-        arguments.addBooleanArgument(FuncotatorArgumentDefinitions.ALLOW_HG19_GENCODE_B37_CONTIG_MATCHING_LONG_NAME, true);
+
+        // We need this argument since we are testing on a subset of b37
+        arguments.addBooleanArgument(FuncotatorArgumentDefinitions.FORCE_B37_TO_HG19_REFERENCE_CONTIG_CONVERSION, true);
```
droazen: If your B37/HG19 detection mechanism were a bit more permissive, you might not need these shenanigans. In particular, if you require only 1 autosome to match a known contig+length from B37 or HG19, instead of requiring all contigs to match, then I think that you should be able to work with these subsetted test files without hacks like this.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
```diff
@@ -297,12 +297,28 @@ public void testFuncotatorWithoutValidatingResults(final String dataSourcesPath,
         runCommandLine(arguments);
     }
 
+    private void validateFuncotationsOnVcf(final Iterable<VariantContext> vcfIterable, final String[] funcotationFieldNames) {
```
droazen: Do you have at least one test case that fails without this patch, and passes with it?

src/main/java/org/broadinstitute/hellbender/engine/FeatureCache.java
```diff
@@ -223,11 +223,23 @@ public void trimToNewStartPosition( final int newStart ) {
...
+
+        final String sourceNameString = sourceName.isEmpty() ? "" : "for data source " + sourceName + " ";
+
         final int totalQueries = getNumCacheHits() + getNumCacheMisses();
-        logger.debug(String.format("Cache hit rate was %.2f%% (%d out of %d total queries)",
+        logger.info(String.format("Cache hit rate %swas %.2f%% (%d out of %d total queries)",
```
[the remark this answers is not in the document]
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/engine/FeatureContext.java
```diff
@@ -39,7 +39,7 @@
      * FeatureManager containing backing data sources for all discovered Feature arguments.
      * Null if there are no sources of Features.
      */
-    private final FeatureManager featureManager;
+    public final FeatureManager featureManager;
```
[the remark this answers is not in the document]
jonn-smith: Fixed

src/main/java/org/broadinstitute/hellbender/engine/FeatureManager.java
```diff
@@ -86,7 +86,7 @@
     /**
      * Mapping from FeatureInput argument to query-able FeatureDataSource for that source of Features
      */
-    private final Map<FeatureInput<? extends Feature>, FeatureDataSource<? extends Feature>> featureSources;
+    public final Map<FeatureInput<? extends Feature>, FeatureDataSource<? extends Feature>> featureSources;
```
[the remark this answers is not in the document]
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -137,17 +138,17 @@
...
 
     /**
      * Our source of reads data (null if no source of reads was provided)
      */
-    ReadsDataSource reads;
+    protected ReadsDataSource reads;
```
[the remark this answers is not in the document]
jonn-smith: Sure.

Fixed!

src/main/java/org/broadinstitute/hellbender/engine/ReferenceContext.java
```diff
@@ -352,7 +352,7 @@ private int trimToContigLength(final String contig, final int end){
 
         final SAMSequenceRecord sequence = dataSource.getSequenceDictionary().getSequence(contig);
         if ( sequence == null ) {
-            throw new UserException("Given reference file does not have data at the requested contig(" + contig + ")!");
+            throw new UserException.NoDataAtRequestedContig(contig);
```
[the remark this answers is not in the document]
jonn-smith: I needed this to handle one of the error cases when I was using error handling as signal passing.  I will remove the class.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -378,32 +386,89 @@ public void onTraversalStart() {
...
+            // This naming convention difference causes a LOT of issues and was a bad idea.
+            inputReferenceIsB37 = forceB37ToHg19ContigNameConversion || FuncotatorUtils.isSequenceDictionaryUsingB37Reference(getSequenceDictionaryForDrivingVariants());
             if ( inputReferenceIsB37 ) {
                 logger.warn("WARNING: You are using B37 as a reference.  " +
-                        "Funcotator will convert your variants to GRCh37, and this will be fine in 99.9% of cases.  " +
-                        "There MAY be some errors in the Y chromosome due to changes between the two references.");
+                        "Funcotator will convert your variants to GRCh37, and this will be fine in the vast majority of cases.  " +
+                        "There MAY be some errors (e.g. in the Y chromosome, but possibly in other places as well) due to changes between the two references.");
+
+                mustConvertInputContigsToHg19 = forceB37ToHg19ContigNameConversion || referenceVersion.equals(FuncotatorArgumentDefinitions.HG19_REFERENCE_VERSION_STRING);
```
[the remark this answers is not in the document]
jonn-smith: Sure.  Fixed.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -378,32 +386,89 @@ public void onTraversalStart() {
...
+                });
+    }
+
     @Override
     public void apply(final VariantContext variant, final ReadsContext readsContext, final ReferenceContext referenceContext, final FeatureContext featureContext) {
 
         // Place the variant on our queue to be funcotated:
         enqueueAndHandleVariant(variant, referenceContext, featureContext);
+
+        ++numVariantsAnnotated;
```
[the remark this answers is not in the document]
jonn-smith: I had this in as some debugging code.  I'll remove it.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorArgumentDefinitions.java
jonn-smith: Sure.  Fixed.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorUtils.java
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -378,32 +386,89 @@ public void onTraversalStart() {
...
+            // Ignore variants that have been filtered if the user requests it:
+            if ( removeFilteredVariants && variant.isFiltered() ) {
+                return false;
+            }
+            return true;
+        };
+    }
+
+    @Override
+    public void traverse() {
```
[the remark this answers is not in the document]
jonn-smith: Sounds good.

In order to adjust the interval I have had to create a `setInterval` method in `ReferenceContext`.

Fixed!

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
jonn-smith: A test case for what specifically?  This PR was created to increase the speed of annotation - I don't have a test that goes slowly without and faster with this patch.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
jonn-smith: Being permissive with references already makes me uncomfortable.  I'm of the opinion that files should match in their entirety and that subsets aren't sufficient.  I'd prefer not to put in any more fuzzy matching with references.

As an aside, the reference issue in Funcotator should be solved by adding in another reference set for `b37` so that every reference has its own backing data sources.  This patch to me is something of a stop-gap to make Funcotator fast in the mean time.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
jonn-smith: I'll add this check in the form of a call to `SequenceDictionaryUtils.validateDictionaries` that will require the reference sequence dictionary to be a superset of the variant sequence dictionary.

I will need to fix many tests that have broken because of this (and change some file names to be more descriptive).


Also:
I've verified that giving `b37` variants with the UCSC hg19 reference causes `validateSequenceDictionaries` to get upset (this is correct).

It does not get upset when I give `b37` variants and the GRCh37 reference (it should).

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -149,6 +150,42 @@
      */
     FeatureManager features;
 
+    /**
+     * Get the {@link ReferenceDataSource} for this {@link GATKTool}.
+     * Will throw a {@link GATKException} if the reference is null.
+     * @return the {@link ReferenceDataSource} for this {@link GATKTool}.  Never {@code null}.
+     */
+    protected ReferenceDataSource getReferenceDataSource() {
```
droazen: After discussing with @lbergelson, it turns out that this change of introducing these datasource getters and then overriding them in the walker base classes to try to preserve encapsulation is slightly controversial and deserves separate discussion. Since this change is no longer actually needed for this PR (as far as I can determine), can I suggest that you:

1. In this branch, remove the new datasource getters from `GATKTool` and the overrides of the getters from the walker base classes.
2. Create a separate standalone PR that introduces the new getters, so that they can be reviewed/discussed independently of this `Funcotator` PR.
3. In the separate PR, instead of repeating the overrides of the datasource getters in every walker base class, could you introduce a new abstract `Walker` class that extends `GATKTool` and contains the overrides that throw an exception? Then you can just make the walker base classes like `ReadWalker` extend `Walker` rather than `GATKTool` directly, and you won't have to repeat the overrides all over the place. This change might improve the chances of the PR being accepted.

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -330,7 +342,12 @@ public boolean requiresReference() {
     @Override
     public void onTraversalStart() {
 
-        // First set up our transcript list:
+        if (!disableSequenceDictionaryComparison) {
+            // Check that the VCF and the Reference file have compatible sequence dictionaries:
```
droazen: "Ensure that the reference dictionary is a superset of the variant dictionary"

src/main/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorArgumentDefinitions.java
```diff
@@ -48,16 +48,14 @@
...
-
     public static final String HG19_REFERENCE_VERSION_STRING = "hg19";
     public static final String HG38_REFERENCE_VERSION_STRING = "hg38";
 
     public static final String LOOKAHEAD_CACHE_IN_BP_NAME = "lookahead-cache-bp";
     public static final int LOOKAHEAD_CACHE_IN_BP_DEFAULT_VALUE = VariantWalkerBase.FEATURE_CACHE_LOOKAHEAD;
 
-    public static final String ALLOW_HG19_GENCODE_B37_CONTIG_MATCHING_OVERRIDE_LONG_NAME = "allow-hg19-gencode-b37-contig-matching-override";
+    public static final String FORCE_B37_TO_HG19_REFERENCE_CONTIG_CONVERSION = "force-b37-to-hg19-reference-contig-conversion";
+    public static final String DISABLE_SEQUENCE_DICTIONARY_COMPARISON = "disable-sequence-dictionary-comparison";
```
droazen: Remove `DISABLE_SEQUENCE_DICTIONARY_COMPARISON` if possible, as discussed above.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
droazen: Good to know this workaround will be short-lived.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
```diff
@@ -975,10 +1050,42 @@ private void runSomaticVcf(final String vcfFile) {
         arguments.addArgument(FuncotatorArgumentDefinitions.DATA_SOURCES_PATH_LONG_NAME, DS_PIK3CA_DIR);
         arguments.addArgument(FuncotatorArgumentDefinitions.REFERENCE_VERSION_LONG_NAME, FuncotatorTestConstants.REFERENCE_VERSION_HG19);
         arguments.addArgument(FuncotatorArgumentDefinitions.OUTPUT_FORMAT_LONG_NAME, outputFormatType.toString());
-        arguments.addBooleanArgument(FuncotatorArgumentDefinitions.ALLOW_HG19_GENCODE_B37_CONTIG_MATCHING_LONG_NAME, true);
+        arguments.addBooleanArgument(FuncotatorArgumentDefinitions.FORCE_B37_TO_HG19_REFERENCE_CONTIG_CONVERSION, true);
         arguments.addArgument(FuncotatorArgumentDefinitions.TRANSCRIPT_SELECTION_MODE_LONG_NAME, TranscriptSelectionMode.CANONICAL.toString());
 
+        // Disable the sequence dictionary check for the tests:
+        arguments.addBooleanArgument(FuncotatorArgumentDefinitions.DISABLE_SEQUENCE_DICTIONARY_COMPARISON, true);
```
droazen: Perhaps you should extract a helper method that returns an `ArgumentsBuilder` initialized with the arguments you need in every test case, and add arguments like this one there.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
```diff
@@ -975,10 +1050,42 @@ private void runSomaticVcf(final String vcfFile) {
...
+    @Test(expectedExceptions = UserException.IncompatibleSequenceDictionaries.class)
+    public void testSequenceDictionaryCheck() {
+        final FuncotatorArgumentDefinitions.OutputFormatType outputFormatType = FuncotatorArgumentDefinitions.OutputFormatType.VCF;
+        final File outputFile = getOutputFile(outputFormatType);
+
+        final ArgumentsBuilder arguments = new ArgumentsBuilder();
+
+        arguments.addVCF(new File(XSV_CLINVAR_COL_TEST_VCF));
+        arguments.addOutput(outputFile);
+        arguments.addReference(new File(b37Chr2Ref));
```
droazen: Document in a comment why the dictionaries of these inputs are incompatible. Also, are they failing the check in `GATKTool`, or just the stricter check in `Funcotator`? (the latter case would make for a better test)

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
```diff
@@ -975,10 +1050,42 @@ private void runSomaticVcf(final String vcfFile) {
...
+        arguments.addArgument(FuncotatorArgumentDefinitions.DATA_SOURCES_PATH_LONG_NAME, DS_XSV_CLINVAR_TESTS);
+        arguments.addArgument(FuncotatorArgumentDefinitions.REFERENCE_VERSION_LONG_NAME, FuncotatorTestConstants.REFERENCE_VERSION_HG19);
+        arguments.addArgument(FuncotatorArgumentDefinitions.OUTPUT_FORMAT_LONG_NAME, outputFormatType.toString());
+
+        arguments.addBooleanArgument(FuncotatorArgumentDefinitions.FORCE_B37_TO_HG19_REFERENCE_CONTIG_CONVERSION, true);
+
         runCommandLine(arguments);
```
droazen: Isn't everything after `runCommandLine()` in this test case unnecessary, since you expect to get an exception at runtime?

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -149,6 +150,42 @@
...
+    }
+
+    /**
+     * Get the {@link FeatureManager} for this {@link GATKTool}.
+     * Will throw a {@link GATKException} if the features are null.
+     * @return the {@link FeatureManager} for this {@link GATKTool}.  Never {@code null}.
+     */
+    protected FeatureManager getFeatureManager() {
+        if ( features == null ) {
+            throw new GATKException("Attempted to retrieve null reference!");
```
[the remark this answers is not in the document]
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -149,6 +150,42 @@
      */
     FeatureManager features;
 
+    /**
+     * Get the {@link ReferenceDataSource} for this {@link GATKTool}.
+     * Will throw a {@link GATKException} if the reference is null.
```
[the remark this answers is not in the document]
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/engine/ReferenceContext.java
```diff
@@ -248,6 +248,32 @@ public SimpleInterval getWindow() {
...
+    /**
+     * Set the location on the reference represented by this context, without including
+     * any extra bases of requested context around this interval.
+     *
+     * Will set the window around the given interval to be the same as the previous window.
+     * Will invalidate the cache.
+     *
+     * @param newInterval The new interval on which this reference will be based.
+     */
+    public void setInterval(final SimpleInterval newInterval) {
```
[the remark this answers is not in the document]
jonn-smith: Sounds good.  Fixed!

src/main/java/org/broadinstitute/hellbender/engine/ReferenceContext.java
```diff
@@ -248,6 +248,32 @@ public SimpleInterval getWindow() {
         return window;
     }
 
+    /**
+     * Set the location on the reference represented by this context, without including
+     * any extra bases of requested context around this interval.
```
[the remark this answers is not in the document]
jonn-smith: OBE - I'm removing the `setInterval` method in favor of the copy constructor.

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -93,17 +111,57 @@ public final SAMSequenceDictionary getBestAvailableSequenceDictionary() {
      */
     protected abstract Spliterator<VariantContext> getSpliteratorForDrivingVariants();
 
+    /**
+     * Returns the pre-filter variant transformer (simple or composite) that will be applied to the variants before filtering.
+     * The default implementation uses the {@link VariantTransformer#identity()}.
+     * Default implementation of {@link #traverse()} calls this method once before iterating over the reads and reuses
```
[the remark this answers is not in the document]
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -93,17 +111,57 @@ public final SAMSequenceDictionary getBestAvailableSequenceDictionary() {
...
+     *
+     * Subclasses can extend to provide own transformers (i.e. override and call super).
+     * Multiple transformers can be composed by using {@link VariantTransformer} composition methods.
+     */
+    public VariantTransformer makePreVariantFilterTransformer() {
+        return VariantTransformer.identity();
+    }
+
+    /**
+     * Returns the post-filter variant transformer (simple or composite) that will be applied to the reads after filtering.
```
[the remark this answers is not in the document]
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/engine/VariantWalkerBase.java
```diff
@@ -93,17 +111,57 @@ public final SAMSequenceDictionary getBestAvailableSequenceDictionary() {
...
+     * The default implementation uses the {@link VariantTransformer#identity()}.
+     * Default implementation of {@link #traverse()} calls this method once before iterating over the reads and reuses
```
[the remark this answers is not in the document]
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -297,12 +293,23 @@
...
+    private boolean forceB37ToHg19ContigNameConversion = false;
+
+    @Advanced
+    @Hidden
+    @Argument(
+            fullName = FuncotatorArgumentDefinitions.DISABLE_SEQUENCE_DICTIONARY_COMPARISON,
+            optional = true,
+            doc = "(Advanced / DO NOT USE*) If you select this flag, Funcotator will not perform a comparison between the sequence dictionary of the input VCF and the sequence dictionary of the given Reference FASTA.  *This option is useful in integration tests (written by devs) only."
+    )
+    private boolean disableSequenceDictionaryComparison = false;
```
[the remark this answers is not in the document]
jonn-smith: Sure.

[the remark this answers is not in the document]
jonn-smith: Fixed!

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -378,30 +395,93 @@ public void onTraversalStart() {
         }
 
         // Check for reference version (in)compatibility:
-        if ( allowHg19ContigNamesWithB37 &&
-             referenceVersion.equals(FuncotatorArgumentDefinitions.HG19_REFERENCE_VERSION_STRING) ) {
+        determineReferenceAndDatasourceCompatibility();
+    }
+
+    private void checkVcfAndReferenceHaveSameSequenceDictionaries() {
```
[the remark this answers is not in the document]
jonn-smith: Sure.  Fixed!

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -378,30 +395,93 @@ public void onTraversalStart() {
...
+
+        final SAMSequenceDictionary referenceDictionary = getReferenceDictionary();
+        final SAMSequenceDictionary variantDictionary = getSequenceDictionaryForDrivingVariants();
+
+        if ( referenceDictionary == null ) {
+            throw new UserException.BadInput("Reference fasta sequence dictionary is null!");
+        }
+
+        if ( variantDictionary == null ) {
+            throw new UserException.BadInput("Variant input file sequence dictionary is null!");
```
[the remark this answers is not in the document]
jonn-smith: Sounds good.

MAF supports comments so you can put the sequence dictionary in the commented metadata.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
jonn-smith: Sounds good.  Fixed!

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
jonn-smith: Sure.

It's failing in the new check I added in Funcotator.

src/test/java/org/broadinstitute/hellbender/tools/funcotator/FuncotatorIntegrationTest.java
jonn-smith: Yup.  Fixed!

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jonn-smith: Sounds good.  

Submitted as PR #4964

src/main/java/org/broadinstitute/hellbender/engine/ReferenceContext.java
```diff
@@ -40,7 +40,7 @@
     /**
      * Interval representing our location on the reference. May be null if, eg., we're dealing with unmapped data.
      */
-    private final SimpleInterval interval;
+    private SimpleInterval interval;
```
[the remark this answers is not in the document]
jonn-smith: Fixed!

PR #4944, opened 2018-06-25
add option to invert logic in ExtractOriginalAlignmentRecordsByNameSpark

SHuang-Broad:
by turning on flag `-v`

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/ExtractOriginalAlignmentRecordsByNameSpark.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/ExtractOriginalAlignmentRecordsByNameSpark.java
@@ -2,6 +2,7 @@ package org.broadinstitute.hellbender.tools.spark.sv.utils;
 
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.JavaSparkContext;
+import org.apache.spark.api.java.function.Function;
 import org.apache.spark.broadcast.Broadcast;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.BetaFeature;
@@ -22,7 +23,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 /**
- * Subsets reads by name
+ * Subsets reads by name (basically a parallel version of "grep -f", or "grep -vf")
  *
  * <p>Reads a file of read (i.e., template) names, and searches a SAM/BAM/CRAM to find names that match.
  * The matching reads are copied to an output file.</p>
@@ -65,8 +66,12 @@ import java.util.stream.Collectors;
 public final class ExtractOriginalAlignmentRecordsByNameSpark extends GATKSparkTool {
     private static final long serialVersionUID = 1L;
 
-    @Argument(doc = "file containing list of read names", fullName = "read-name-file")
+    @Argument(doc = "file containing list of read names", shortName = "f", fullName = "read-name-file")
     private String readNameFile;
+
+    @Argument(doc = "invert the list, i.e. filter out reads whose name appear in the given file",
+            shortName = "v", fullName = "invert-match", optional = true)
+    private Boolean invertFilter = false;
 
     @Argument(doc = "file to write reads to", shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
             fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME)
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/ExtractOriginalAlignmentRecordsByNameSpark.java
```diff
@@ -82,14 +87,21 @@ protected void runTool( final JavaSparkContext ctx ) {
 
         final Broadcast<Set<String>> namesToLookForBroadcast = ctx.broadcast(parseReadNames());
 
-        final JavaRDD<GATKRead> reads =
-                getUnfilteredReads().filter(read -> namesToLookForBroadcast.getValue().contains(read.getName())).cache();
+        final Function<GATKRead, Boolean> predicate = getGatkReadBooleanFunction(namesToLookForBroadcast);
```
TedBrookings: Super-picky, ignore if you like: "predicate" is a confusing name to me in this context, presumably meaning "to assert" in this context. Since it's an object, I might suggest "nameFilter" which indicates what it is. Again, this is just a suggestion, feel free to ignore.

cwhelan: This is a predicate (noun form) meaning a boolean valued function of one argument. https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/integration/ExtractOriginalAlignmentRecordsByNameSparkIntegrationTest.java
```diff
@@ -10,44 +10,61 @@
...
+    private Object[][] createTestData() throws IOException {
+        final List<Object[]> data = new ArrayList<>(20);
 
         final File tempWorkingDir = BaseTest.createTempDir("extractOriginalAlignmentRecordsByNameSparkIntegrationTest");
 
         FileUtils.writeLines(new File(tempWorkingDir, "names.txt"), Collections.singleton("asm013903:tig00002"));
 
         final SAMFileHeader expectedHeader;
-        final List<SAMRecord> expectedRecords;
+        List<SAMRecord> expectedRecords;
```
TedBrookings: Another nitpick: consider renaming this to final List<SAMRecord> expectedRecordsNormalMatch  (or similar)
and later creating List<SAMRecord> expectedRecordsInvertedMatch (or similar)
The reason being it took a bit of parsing for me to figure out why there were two blocks here that looked (at first glance) identical.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/ExtractOriginalAlignmentRecordsByNameSpark.java
TedBrookings: Ah, I did some googling but missed that definition. I now see it's super-standard, so I retract this comment.

src/test/java/org/broadinstitute/hellbender/tools/spark/sv/integration/ExtractOriginalAlignmentRecordsByNameSparkIntegrationTest.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/ExtractOriginalAlignmentRecordsByNameSpark.java
SHuang-Broad: Thanks for clarifying, Chris!
One note to add: returning a `Function` makes it serializable, which we can use in Spark.
I'm being lazy :see_no_evil: by NOT implementing a class that implements both `Predicate` and `Serializable`

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/ExtractOriginalAlignmentRecordsByNameSpark.java
```diff
@@ -65,9 +66,13 @@
 public final class ExtractOriginalAlignmentRecordsByNameSpark extends GATKSparkTool {
     private static final long serialVersionUID = 1L;
 
-    @Argument(doc = "file containing list of read names", fullName = "read-name-file")
+    @Argument(doc = "file containing list of read names", shortName = "f", fullName = "read-name-file")
     private String readNameFile;
 
+    @Argument(doc = "revert the list, i.e. filter out reads whose name appear in the give file",
```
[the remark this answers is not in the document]
SHuang-Broad: Corrected. Thanks!

PR #4945, opened 2018-06-26
(SV) output "bai" for local assembly alignments BAM

SHuang-Broad:


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/SvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
```diff
@@ -208,17 +208,22 @@ private void writeSAMfilesForUnknown(final String outputPrefix, final JavaRDD<GA
 
             final List<GATKRead> contigRawAlignments = assemblyRawAlignments
                     .filter(read -> namesOfInterest.contains(read.getName())).collect();
+            contigRawAlignments.sort(Comparator.comparing(GATKRead::getName));
```
cwhelan: Do we always want to write these files sorted by name even though the other bam file's sort order is now choosable from the command line? You might want to put a note in the documentation for the new parameter that it will not affect these BAM files.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/AlignedAssemblyOrExcuse.java
```diff
@@ -326,12 +326,23 @@ static void writeAssemblySAMFile(final String outputAssembliesFile,
         final SAMFileHeader cleanHeader = new SAMFileHeader(header.getSequenceDictionary());
         cleanHeader.addReadGroup(contigAlignmentsReadGroup);
         cleanHeader.setSortOrder(assemblyAlnSortOrder);
+        final SAMRecordComparator samRecordComparator;
```
cwhelan: This is a trivial suggestion but if you extracted a method to get the comparator you could just have a `return` statement in each `case` clause and wouldn't need the `break`s.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
```diff
@@ -130,6 +130,8 @@ public boolean requiresReads()
     @Override
     protected void runTool( final JavaSparkContext ctx ) {
 
+        validateParams();
```
cwhelan: Does `StructuralVariationDiscoveryPipelineSpark` need to call this too? 

Would it make sense to move this method to the argument collection class itself? Then the tools that use it could call `validate` on it.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/utils/SVFileUtils.java
```diff
@@ -19,36 +20,16 @@ public static void writeSAMFile(final String outputName, final Iterator<SAMRecor
         Utils.nonNull(header, "provided header for outputting sam file is null");
         Utils.nonNull(outputName, "provided output name is null");
 
-        try (final SAMFileWriter writer = getSAMFileWriter(outputName, header, preOrdered)) {
+        final SAMFileWriterFactory factory = new SAMFileWriterFactory()
+                .setCreateIndex(preOrdered && outputName.endsWith(".bam")
```
cwhelan: I'd make constants for these extensions. For bam you could use htsjdk's `BamFileIoUtils.BAM_FILE_EXTENSION`.

SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/FindBreakpointEvidenceSpark.java
SHuang-Broad: Yes. That's better. Though each tool should have its own `validate` as well preferably.

Just talked with Engine team, such CLI-arg validation methods must be called explicitly right now, so automatically validating them in the constructor is not supported, yet.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/evidence/AlignedAssemblyOrExcuse.java
SHuang-Broad: done. Method implemented and tested in `SVUtils`.

src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/SvDiscoverFromLocalAssemblyContigAlignmentsSpark.java
SHuang-Broad: Yes. Will do.

Users most likely will not care about these bams, because they are more of a debugging use. 

These are local assembly contigs whose alignment signatures that we don't know how to emit un-ambiguous calls from, so their alignments are saved just incase we want to emit BND's, or which ever format we can report, and annotate accordingly.

Personally I generally look at all alignments of a contig to figure out why they are classified that way, and get a sense of if we can come up with rules to squeeze more signal from them.

I also tried with an option of NOT outputting alignments of these contigs, but just names and the reason for failing to produce a call. But in the end I find myself using `ExtractOriginalAlignmentRecordsByNameSpark` again to extract the alignments anyway, so just decided to have some data redundancy in exchange for convenience.

PR #4950, opened 2018-06-26
remove jtransforms

lbergelson:
removing the unused dependency  'com.github.wendykierp:JTransforms:3.1'


lbergelson:
@mbabadi I think you may have been the person who was using this?  Any objections to removing it?

Issue #4766, opened 2018-05-14
Add profiling option to StreamingPythonExecutor


cmnbroad:
Once https://github.com/broadinstitute/gatk/pull/4757 is in, it would be pretty easy to add an `@Advanced` or `@Hidden` tool argument that runs the Python code in (cProfile) profiling mode and writes the profile or summary stats to  a file.

PR #4953, opened 2018-06-27
Enable Python profiling for StreamingPythonScriptExecutor.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/4766. We might want to do a similar thing for the non-streaming executor, in a separate PR.


droazen:
@jonn-smith and @lucidtronix please review when you get a chance. Thanks!

cmnbroad:
@jonn-smith Yes, I think if this turns out to be useful, we may want to add more options, i.e. which stats to collect, but I think well have to use it a bit to figure out which ones.

Issue #4747, opened 2018-05-08
Migrate VariantWalkerBase (and subclasses) to CountingVariantFilter


takutosato:
Currently it uses `VariantFilterLibrary.ALLOW_ALL_VARIANTS`

droazen:
Motivation here is to allow summary statistics for variant filtration to be printed at the end of the `VariantWalker` traversals.

PR #4954, opened 2018-06-27
Integrate CountingVariantFilter.

cmnbroad:
Fixes https://github.com/broadinstitute/gatk/issues/4747.


cmnbroad:
So, the problem with this is that the counting filters use the filter class name to report summaries, and most of the variant filters that we have are implemented using lambdas, so they have no recognizable name. Before this is merged I'll want to reimplement the handful of existing filters used by CNNScoreVariants and Funcotator to use static, named classes in VariantFilterLibrary.

cmnbroad:
@lbergelson I had to add another commit to this after you reviewed it (see my note above), so back to you for another look.

lbergelson:
Looks like this needs a rebase, feel free to merge after unless there are crazy changes needed.

PR #4955, opened 2018-06-27
Refactored the docker build script to only only include the gatk bundle in order to shrink the docker image size

### LINKED ISSUE #3930: Refactor the docker image to correspond to the minimum necessary for user execution
Right now our docker image is much larger than it needs to be. This is at least in part because it contains our entire git clone as well as the packaged jars. This is not necessary and could potentially come with it shrinking our docker image substantially. 

This change would involve a major refactoring of how we execute our tests through gralde inside the docker image, as removing the test dependencies will mean we probably have to externally mount the git clone from the docker image in order to pull in the proper dependencies.

jamesemery:
fixes #3930

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -174,19 +174,6 @@ configurations.all {
     all*.exclude group: 'junit', module: 'junit'
 }
 
-
-jacocoTestReport {
-    dependsOn test
-    group = "Reporting"
-    description = "Generate Jacoco coverage reports after running tests."
-    additionalSourceDirs = files(sourceSets.main.allJava.srcDirs)
-
-    reports {
-        xml.enabled = true
-        html.enabled = true
-    }
-}
-
 //NOTE: we ignore contracts for now
 compileJava {
   options.compilerArgs = ['-proc:none', '-Xlint:all', '-Werror', '-Xdiags:verbose']
@@ -379,7 +366,6 @@ version = (isRelease ? gitVersion() : gitVersion() + "-SNAPSHOT").replaceAll(".d
 logger.info("build for version:" + version)
 group = 'org.broadinstitute'
 
-
 tasks.withType(Jar) {
     manifest {
         attributes 'Implementation-Title': 'The Genome Analysis Toolkit (GATK)',
@@ -390,7 +376,33 @@ tasks.withType(Jar) {
     }
 }
 
-test {
+task wrapper(type: Wrapper) {
+    gradleVersion = '3.1'
+}
+
+tasks.withType(ShadowJar) {
+    from(project.sourceSets.main.output)
+    baseName = project.name + '-package'
+    mergeServiceFiles()
+    relocate 'com.google.common', 'org.broadinstitute.hellbender.relocated.com.google.common'
+    zip64 true
+    exclude 'log4j.properties' // from adam jar as it clashes with hellbender's log4j2.xml
+    exclude '**/*.SF' // these are Manifest signature files and
+    exclude '**/*.RSA' // keys which may accidentally be imported from other signed projects and then fail at runtime
+
+    // Suggested by the akka devs to make sure that we do not get the spark configuration error.
+    // http://doc.akka.io/docs/akka/snapshot/general/configuration.html#When_using_JarJar__OneJar__Assembly_or_any_jar-bundler
+    transform(com.github.jengelman.gradle.plugins.shadow.transformers.AppendingTransformer) {
+        resource = 'reference.conf'
+    }
+}
+
+//============================================================================================================================
+// WARNING
+//============================================================================================================================
+// This is duplicated code with the build script in scripts/docker/dockertest.gradle. If you alter the tests in any way
+// please make sure to alter this test method there as well.
+tasks.withType(Test) {
     outputs.upToDateWhen { false }  //tests will never be "up to date" so you can always rerun them
     String TEST_VERBOSITY = "$System.env.TEST_VERBOSITY"
 
@@ -401,7 +413,11 @@ test {
      * anything else             : run the non-cloud tests
      */
     String TEST_TYPE = "$System.env.TEST_TYPE"
-
+//============================================================================================================================
+// WARNING
+//============================================================================================================================
+// This is duplicated code with the build script in scripts/docker/dockertest.gradle. If you alter the tests in any way
+// please make sure to alter this test method there as well.
     useTestNG {
         if (TEST_TYPE == "cloud") {
             // run only the cloud tests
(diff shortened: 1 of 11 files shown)


lbergelson:
@jamesemery How big was the docker and how big is it now?

jamesemery:
@lbergelson 
now: 4.1GB (without ~0.3-5 GB extra off from the base image changes currently in the branch)
then:  5.22GB
A moderate but not irrelevant reduction in size. With these changes some more drastic changes will be easier, namely trying to slice down the python packages or refactoring the build jars should be easier. I intend to open another PR to upgrade the base image shortly.

droazen:
@jamesemery So to clarify: including the size reductions to the base image, the total reduction will be around 1.6-1.8 GB, correct?

jamesemery:
@droazen correct. 
Generally, one issue is that this slows down the docker image creation in a somewhat substantial way. Around 10 minutes currently. Half of this is unzipping the bundled jar, and another piece is some redundant gradle downloading that can be alleviated with cache shenanigans.

build.gradle
```diff
@@ -524,7 +509,25 @@ task sparkJar(type: ShadowJar) {
     }
 }
 
-task bundle(type: Zip) {
+// A jar that only contains the test classes and resources (to be extracted for testing)
+task shadowTestClassJar(type: ShadowJar){
+    group = "Shadow"
```
lbergelson: Add description lines to the new tasks so that people who see them have some idea what they're for.

scripts/docker/build.gradle
```diff
@@ -0,0 +1,174 @@
+//Note: this section 'buildscript` is only for the dependencies of the buildscript itself.
+// See the second 'repositories' section below for the actual dependencies of GATK itself
+buildscript {
```
lbergelson: I would probably rename this to something other than build.gradle  Maybe dockertest.gradle?

scripts/docker/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+    mavenCentral()
+    jcenter()
+    maven {
+        url "https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot/" //for htsjdk snapshots
+    }
+    mavenLocal()
+}
+
+////Note: the test suite must use the same defaults. If you change system properties in this list you must also update the one in the test task
+//applicationDefaultJvmArgs = ["-Dsamjdk.use_async_io_read_samtools=false","-Dsamjdk.use_async_io_write_samtools=true", "-Dsamjdk.use_async_io_write_tribble=false", "-Dsamjdk.compression_level=2"]
```
lbergelson: delete these

scripts/docker/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Valid options for TEST_TYPE are:
+     * cloud, integration, unit  : run one of the three disjoint partitions of the test suite
+     * all                       : run all the tests
+     * anything else             : run the non-cloud tests
+     */
+    String TEST_TYPE = "$System.env.TEST_TYPE"
+
+    useTestNG {
```
lbergelson: We need to add a big obvious comment to this identical block in the regular build.gradle that this has to be updated when it is.  Alternatively, can we factor it out into a shared subscript?  We're going to have nasty bugs where we aren't running some tests when someone updates the regular test block but not this one.

scripts/docker/build.gradle
```diff
@@ -0,0 +1,174 @@
+//Note: this section 'buildscript` is only for the dependencies of the buildscript itself.
+// See the second 'repositories' section below for the actual dependencies of GATK itself
```
lbergelson: this second comment doesn't make sense here

scripts/docker/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+    repositories {
+        mavenCentral()
+        jcenter() // for shadow plugin
+     }
+}
+
+
+plugins {
+    id "java"           // set up default java compile and test tasks
+    id 'maven'
```
lbergelson: I doubt we need maven here

scripts/docker/build.gradle
```diff
@@ -0,0 +1,174 @@
+//Note: this section 'buildscript` is only for the dependencies of the buildscript itself.
+// See the second 'repositories' section below for the actual dependencies of GATK itself
+buildscript {
+    repositories {
+        mavenCentral()
+        jcenter() // for shadow plugin
```
lbergelson: You can drop jcenter from this one

scripts/docker/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+
+
+plugins {
+    id "java"           // set up default java compile and test tasks
+    id 'maven'
+    id "jacoco"
+}
+
+repositories {
+    mavenCentral()
```
lbergelson: You can remove this whole block I think.

scripts/docker/build.gradle
lbergelson: If we can share the code between the two scripts, can we merge this tasks.withType block with the actual test below, it's weird to split it when there's only 1.

scripts/docker/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+            if (!desc.parent) { // will match the outermost suite
+                println "Results: ${result.resultType} (${result.testCount} tests, ${result.successfulTestCount} successes, ${result.failedTestCount} failures, ${result.skippedTestCount} skipped)"
+            }
+        }
+    }
+}
+
+task unpackTestJar(type: Copy){
+    String testClassesJar = "$System.env.TEST_JAR"
+    String testClassesDir = "$System.env.CP_DIR"
```
lbergelson: Add a comment explaining why this is necessary for future generations.

lbergelson: and a description line

build.gradle
```diff
@@ -390,7 +376,28 @@ tasks.withType(Jar) {
...
+    exclude '**/*.RSA' // keys which may accidentally be imported from other signed projects and then fail at runtime
+
+    // Suggested by the akka devs to make sure that we do not get the spark configuration error.
+    // http://doc.akka.io/docs/akka/snapshot/general/configuration.html#When_using_JarJar__OneJar__Assembly_or_any_jar-bundler
+    transform(com.github.jengelman.gradle.plugins.shadow.transformers.AppendingTransformer) {
+        resource = 'reference.conf'
+    }
+}
+
+tasks.withType(Test) {
```
lbergelson: You can put this back as just a modification of the test task since there's no second one here anymore.

scripts/docker/build.gradle
jamesemery: i think we need it to download jacoco? we point to maven central still.

lbergelson: The maven plugin is for releasing to maven, not downloading from it.  That's a core gradle thing I think.

build.gradle
jamesemery: i'd actually rather leave it, because we are using warnings i would rather the two test methods that need to be updated in lock step are byte identical so its as easy as possible for someone to update.

PR #4956, opened 2018-06-27
Fix position annotations to use position in original, not clipped, read

davidbenjamin:
`ReadPosition` and `ReadPositionRankSum` had used the position of the variant within the read *after* it had been hard-clipped to the assembly region, making these annotations less informative.


davidbenjamin:
@ldgauthier Tests are still failing because I haven't updated test vcfs.  Before I do so could I get your opinion?  Here's a summary of the rationale behind each commit:

* fa34758 When we get to `realignReadToBestHaplotype` we have discarded bases outside of the assembly region but we still have hard clips in the read Cigar.  This change forces us to keep those Cigar elements after realigning to the best haplotype so that we know how far past the assembly window the read extended.
* 6af7ad4 After the above change, `BaseQualityRankSum` was liable to look for discarded bases in the hard-clipped part of the read.  This fixes that.
* 952d217 The `Clipping` annotation doesn't do anything.  It counts the number of hard clips, but pre-this PR there are no hard clips because those bases don't get realigned to the best haplotype and post-this PR the "hard clips" are just bases outside the assembly region.  As a placeholder I'm setting it to zero (note how this doesn't break any tests!) but really I think we should just get rid of it.
* This PR introduced some off-by-one errors in the depth annotation (but not the ADs).  While looking into this I found an apparent bug where some reads that don't overlap a variant get counted in the depth.  The issue was that we were counting clipped bases in the overlap.  I don't think this is correct because by this point in the code we have unclipped soft clips and gone through local reassembly.  Therefore, anything clipped here is a part of the read we truly don't believe belongs anywhere near the assembly region.  This change alone breaks the tests with a few off-by-one DP fields.
* 8c51c0a Uses hard clips in the Cigar to correct read position annotations.

davidbenjamin:
@ldgauthier Ready for final review.

src/test/resources/org/broadinstitute/hellbender/tools/haplotypecaller/expected.testVCFMode.gatk4.vcf
```diff
@@ -26,161 +25,161 @@
...
 20	10004094	.	A	C	1825.77	.	AC=2;AF=1.00;AN=2;DP=53;ExcessHet=3.0103;FS=0.000;MLEAC=2;MLEAF=1.00;MQ=52.75;QD=34.45;SOR=1.524	GT:AD:DP:GQ:PL	1/1:0,53:53:99:1854,157,0
 20	10004147	.	A	G	1837.77	.	AC=2;AF=1.00;AN=2;DP=52;ExcessHet=3.0103;FS=0.000;MLEAC=2;MLEAF=1.00;MQ=54.99;QD=30.67;SOR=1.136	GT:AD:DP:GQ:PL	1/1:0,52:52:99:1866,156,0
-20	10004193	.	G	T	916.77	.	AC=1;AF=0.500;AN=2;BaseQRankSum=-0.392;ClippingRankSum=0.000;DP=52;ExcessHet=3.0103;FS=2.369;MLEAC=1;MLEAF=0.500;MQ=59.30;MQRankSum=3.385;QD=17.98;ReadPosRankSum=0.995;SOR=0.941	GT:AD:DP:GQ:PL	0/1:23,28:51:99:945,0,776
-20	10004222	.	C	CA	12.96	.	AC=1;AF=0.500;AN=2;BaseQRankSum=0.335;ClippingRankSum=0.000;DP=55;ExcessHet=3.0103;FS=1.494;MLEAC=1;MLEAF=0.500;MQ=60.05;MQRankSum=-2.042;QD=0.33;ReadPosRankSum=-0.052;SOR=0.434	GT:AD:DP:GQ:PL	0/1:31,8:39:50:50,0,923
-20	10004223	.	A	AG	47.73	.	AC=1;AF=0.500;AN=2;BaseQRankSum=-0.480;ClippingRankSum=0.000;DP=63;ExcessHet=3.0103;FS=0.000;MLEAC=1;MLEAF=0.500;MQ=58.33;MQRankSum=1.886;QD=1.02;ReadPosRankSum=2.234;SOR=0.638	GT:AD:DP:GQ:PL	0/1:36,11:47:85:85,0,982
```
ldgauthier: I spot checked this site and it looks pretty respectable.  I think this variant benefited from your changes.

PR #4960, opened 2018-06-27
Adding in a config file for Funcotator.

### LINKED ISSUE #4581: Funcotator - Need to create funcotator config file as an alternative to CLI options
Currently `Funcotator` uses many command-line options for configuration.  This is inefficient and potentially confusing.

`Funcotator` should be modified to take in a configuration file (using `Owner`) that can configure all of the CLI options for the tool.

jonn-smith:
Fixes #4581

Had to change a configuration method to allow Funcotator tests to work.

Currently funcotator will not use the config cache - it will always
create a new configuration on startup.  This should be fine for now.


src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
```diff
@@ -427,6 +457,35 @@ public void closeTool() {
 
     //==================================================================================================================
 
+    /**
```
LeeTL1220: Shouldn't this be in the superclass?  Or the Argument sub-system?

LeeTL1220: This seems like useful functionality to automate and put in every command line....

LeeTL1220:
@jonn-smith Just a question...

src/main/java/org/broadinstitute/hellbender/tools/funcotator/Funcotator.java
jonn-smith: That's a good point.  

@LeeTL1220 to be clear, are you saying it should be generic and accept any argument for a tool as part of the config file?  That is, you can give the config file arg and the config file can contain all args for the tool?  

I think making it completely generic would be difficult because of the way `Owner` is built (but not impossible).

@droazen care to weigh in?

LeeTL1220: @jonn-smith That is what I meant.

magicDGS: In barclay, there is a special argument to load the arguments from a file (`--arguments_file`, see [barclay#SpecialArgumentsCollection](https://github.com/broadinstitute/barclay/blob/master/src/main/java/org/broadinstitute/barclay/argparser/SpecialArgumentsCollection.java)). This is included in `CommandLineProgram`, so it should be doing that job. I think that configuration files makes more sense for concrete arguments/tools that are system-wide.

In the case of Funcotator makes sense because the config file can be provided with the sources and ask the user to modify them instead of provide single arguments, but it can also be an `arguments_file` instead.

jonn-smith:
The Engine Team discussed this internally and we're going to pull out a subset of all the configuration options into the config file.  These options should be those that will change only infrequently (like the data sources directory).

PR #4962, opened 2018-06-28
Fix a new edge case in alignment deoverlapping

SHuang-Broad:
The problem is that when alignments are de-overlapped (i.e. they overlap on the read) in the CPX logic, some resulting alignment could be only 1 base long, leading to problems.

This was not caught before because in practice a heuristic alignment filtering step is in place in `AssemblyContigAlignmentsConfigPicker` to filter out, in a post-hoc way, such small alignments.
When I experimented with switching orders of the filtering steps, this behavior was observed.

The aim is to make the downstream logic agnostic to upstream filtering detail.


cwhelan:
Would it make more sense to have a much more stringent cutoff for alignment size after de-overlap? For example, I'm not sure I'd really trust a 5, 7, or even 8 base alignment. Have you experimented with higher values? What's the downstream effect of changing this cutoff that you're proposing here; and would it make sense to make it something much higher, like say 19 to match the minimum BWA-MEM seed length?

SHuang-Broad:
Thanks for the suggestion!

To answer the questions:
>Would it make more sense to have a much more stringent cutoff for alignment size after de-overlap? 

Yes! The filtering step is actually done in the class `AssemblyContigAlignmentsConfigPicker` in the `alignment` package, where the unique read span length filter is defaulted to 10 base.

I put it this way so that the contigs won't be "re-classified" in `CpxVariantInterpreter` as having a simple chimera and having to be sent back to `SimpleNovelAdjacencyInterpreter`. So, the idea was to separate the concerns of alignment picking from type inference.

> What's the downstream effect of changing this cutoff that you're proposing here; and would it make sense to make it something much higher, like say 19 to match the minimum BWA-MEM seed length?

I'll experiment with the new suggested length.
The idea behind settling on this size-10 filter was to be more permissive when it comes to alignment filtering in `AssemblyContigAlignmentsConfigPicker`, and filter variants later in VCF.

SHuang-Broad:
Results are in:

Using the branch for PR #4971 with the value `ALIGNMENT_LOW_READ_UNIQUENESS_THRESHOLD` set to 10 and 19, while keeping the gap split children together (that is, method 
`private static GoodAndBadMappings splitGaps(final GoodAndBadMappings configuration, final boolean keepSplitChildrenTogether)` is called with `false` for its second parameter).

Here are the comparisons:
```
                  simple variants          unique TP          unique FP
size-10 filter:        10756                      24                101
size-19 filter:        10755                       1                  0
```

So I think your suggestion is a better trade off!

What I'll do is make that parameter an (advanced) CLI argument in PR #4971 , and experiment more to settle on a good default value.

SHuang-Broad:
Assigning to @TedBrookings assuming @cwhelan is busy with the inversion PR #4889.
Please feel free to re-assign if I'm wrong.

cwhelan:
It's OK, I can still look at this. Sorry for the delay. I don't want @TedBrookings to be distracted from trying to wrap up his change set.

If I understand correctly, you'd like to keep the lower threshold of 2 here and then apply the more strict threshold downstream, in `AssemblyContigAlignmentsConfigPicker`? In that case, why make it 2 and not 1, ie. not have a limit at all here?

SHuang-Broad:
I see. 

Yes. That's what I'm planning on (except that `AssemblyContigAlignmentsConfigPicker` is upstream of this unit), and here's the thought for why:
* I'd try to place the alignment picking step in a single place as much as possible, this makes improvements to the alignment picking/filtering step easier
* the size-based filter can be tuned, even by an CLI argument, this would affect the number of segments in the CPX logic, and the alt_arrangment annotations, and the simple variants re-interpreted by `CpxVariantReInterpreterSpark`, but it won't affect the alt haplotype sequence, which IMO is what really is important. 
* I'm developing a downstream variant filter, which hopefully can cut down the false-positives.

And for the question of "why 2 instead of 1", I think what you are suggesting is to change
```java
public static final int MIN_READ_SPAN_AFTER_DEOVERLAP = 2;
if (one.getSizeOnRead() >= MIN_READ_SPAN_AFTER_DEOVERLAP) result.add(one);
```
to
```java
public static final int MIN_READ_SPAN_AFTER_DEOVERLAP = 1;
if (one.getSizeOnRead() > MIN_READ_SPAN_AFTER_DEOVERLAP) result.add(one);
```
Am i right?

cwhelan:
For the "why 2 instead of 1" I guess what I meant is what's the point of filtering based on length at all then? Why not allow any size alignment? Why is having an alignment with length 1 worse than an alignment with length 2? I was proposing dropping the check entirely.

SHuang-Broad:
Oh, I see.

I think I didn't explain the situation very well (my bad), the middle 1bp alignment is what I'm trying to drop in this PR (so the two segments will become one):

untitled

If I don't drop it, the resulting variant will have two segment as illustrated in the figure, with empty alt arrangement. The tool `CpxVariantReInterpreterSpark` is not expecting such cases, so will throw exceptions.
Alternatively, I can fix `CpxVariantReInterpreterSpark` and merge the two deletion calls.
Would that be better?

cwhelan:
Oh, right. Sorry, you did say that. I'm guess I'm still not sure I really understand why the 1bp alignment causes problems, though, in a way that a 2bp alignment wouldn't? Is the 1bp alignment really zero length somehow -- is that the issue?

SHuang-Broad:
> Is the 1bp alignment really zero length somehow -- is that the issue?

Yes! In a different sense.
The explanation below is a bit mind-numbing, so please bear with me:

--------

So, to illustrate further, this is the alignment that prompted this PR: [ribbon plot](http://www.genomeribbon.com/?perma=Vg2BPO7Yay)

Turning off the filter step in `AssemblyContigAlignmentsConfigPicker` by setting `ALIGNMENT_LOW_READ_UNIQUENESS_THRESHOLD = 0;` (hence the comment _"The aim is to make the downstream logic agnostic to upstream filtering detail"_), the middle alignment--actually only 2 alignment but we do gap-split--is uniquely covering only 1 base.

It would generate the following without this PR
```
chr21	42283722	CPX_chr21:42283722-42284027	C	<CPX>	.	.	ALIGN_LENGTHS=733;ALT_ARRANGEMENT;CTG_NAMES=asm029245:tig00000;END=42284027;HQ_MAPPINGS=1;MAPPING_QUALITIES=60;MAX_ALIGN_LENGTH=733;SEGMENTS=chr21:42283722-42283784,chr21:42283784-42284027;SEQ_ALT_HAPLOTYPE=CAC;SVLEN=-303;SVTYPE=CPX;TOTAL_MAPPINGS=1
```
with this PR, it would generate:
```
chr21	42283722	CPX_chr21:42283722-42284027	C	<CPX>	.	.	ALIGN_LENGTHS=733;ALT_ARRANGEMENT=UINS-1;CTG_NAMES=asm029245:tig00000;END=42284027;HQ_MAPPINGS=1;MAPPING_QUALITIES=60;MAX_ALIGN_LENGTH=733;SEGMENTS=chr21:42283722-42284027;SEQ_ALT_HAPLOTYPE=CAC;SVLEN=-303;SVTYPE=CPX;TOTAL_MAPPINGS=1
```
The differences are in annotations `ALT_ARRANGEMENT` and `SEGMENTS`:
 * without the PR, the code hyper-segments the region into 2 segments, with no `ALT_ARRANGEMENT`. This is technically wrong.
 * with the PR, the two segmentations are "merged" into one, and the `ALT_ARRANGEMENT` annotation correctly says there's one "unmapped" base (technically not unmapped, but it would bring in more noise because a single base could map to 1/4 of the genome).

And yes, the hyper-segmentation problem is still there for cases where de-overlapped alignments are 2bp long (or very short, 3bp, 5bp, etc.), but in that case the `ALT_ARRANGEMENT` won't be empty. And the re-interpretation unit (code that turns the CPX variant into INS, DEL, INV) can handle that with some rules.

------

The technical issue at the core is in `CpxVariantInducingAssemblyContig.extractJumpsOnReference()`: 
imagine a little stick man (yes, that's how I mentally picture it always....) walking along the reference holding a map (the alignments), the map tell it sometimes to walk, sometimes to jump, sometimes to walk backwards; with the 1 base long alignment, it lands at POSITION and immediately jumps away, whereas with a 2bp long alignment, it lands at POSITION, walks one base, and jumps away from POSITION+1.

Issue #4716, opened 2018-04-30
GenomicsDB Import/Select fails on record containing a spanning deletion allele


cwhelan:
Reading from GenomicsDB fails when a some records containing spanning deletion alleles are imported into a workspace. Not all records seem to cause this to fail; I haven't been able to figure out what specific properties of the records cause the error. Here's the contents (minus header) of a VCF file that causes the error:

```
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	NA12878
20	10097436	.	CTTTTCTTTCTTTCTTTCTTTCTTTCTTTCTTTCTTT	C,<NON_REF>	1054.73	.	BaseQRankSum=1.820;ClippingRankSum=0.000;DP=89;ExcessHet=3.0103;MLEAC=1,0;MLEAF=0.500,0.00;MQRankSum=-6.464;RAW_MQ=262143.00;ReadPosRankSum=-3.231	GT:AD:DP:GQ:PL:SB	0/1:57,32,0:89:99:1092,0,2241,1263,2338,3601:23,34,11,21
20	10097437	.	TTTTC	*,T,<NON_REF>	2089.73	.	DP=76;ExcessHet=3.0103;MLEAC=1,1,0;MLEAF=0.500,0.500,0.00;RAW_MQ=217330.00	GT:AD:DP:GQ:PL:SB	1/2:0,32,23,0:55:99:2127,940,1799,1195,0,1125,2201,1453,1262,2642:0,0,16,39
```

Steps to reproduce:

```
./gatk GenomicsDBImport -R src/test/resources/large/human_g1k_v37.20.21.fasta -L 20 -V test_gdb_import.vcf.gz -genomicsdb-workspace-path spanDelWorkspace
./gatk SelectVariants -V gendb://spanDelWorkspace -R src/test/resources/large/human_g1k_v37.20.21.fasta -O test.vcf -L 20
```

Error:

```
java.lang.IllegalArgumentException: Duplicate allele added to VariantContext: T
    at htsjdk.variant.variantcontext.VariantContext.makeAlleles(VariantContext.java:1490)
    at htsjdk.variant.variantcontext.VariantContext.<init>(VariantContext.java:380)
    at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:494)
    at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:488)
[...]

PR #4963, opened 2018-06-28
Make HaplotypeCaller genotype and output spanning deletions

cwhelan:
This PR modifies `HaplotypeCaller` so that it can output and genotype spanning deletion alleles represented by the `*` allele. 

Currently, the output of `HaplotypeCaller` will not include spanning deletion alleles when run in single sample VCF mode or in genotype given alleles mode, even when that genotype would be more appropriate. In the joint calling workflow `GenotypeGVCFs` adds genotypes for spanning deletions, although the input likelihoods will not be broken out to specifically account for spanning deletion alleles.

Some implementation notes:

- I also fixed some behavior specific to GGA mode that I encountered while testing this bug. In particular, when GGA mode was run with multiple variants with the same start position or with spanning events, `HaplotypeCaller` used to emit the warning `"Multiple valid VCF records detected in the alleles input file at site " + loc + ", only considering the first record"` for each such site. This was a bit of a misleading message, since the other variants were in fact taken into account UNLESS HC decided to emit an empty variant context, for example due to zero coverage.
- I rewrote the `createAlleleMapper` method in `AssemblyBasedCallerGenotypingEngine`. The old version had a very brittle mapping scheme that depended heavily on the ordering of alleles in the variant context created by `AssemblyBasedCallerUtils.makeMergedVariantContext` and `getEventsAtThisLoc`. This proved to be difficult to ensure when spanning deletions were added in, and there was an ominous TODO in the old method saying that the logic was not good enough, so I ended up re-writing it from scratch. The new version is longer but I hope it is easier to read and less fragile, but let me know if I've missed anything.

Test currently fail on this branch and therefore it should not be merged. To make them pass we need a fix to https://github.com/broadinstitute/gatk/issues/4716 which is currently being worked on in https://github.com/broadinstitute/gatk/pull/4645. However, since that PR is taking a while to make it through code review, I thought it might be good to start the review process for these changes.


cwhelan:
Not sure who the best reviewer(s) would be but it would be good to get some experienced eyes on this because it touches a few core methods of haplotype caller.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -168,24 +273,82 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
...
+                    if (spanningEvent.getReference().length() == mergedVC.getReference().length()) {
+                        // reference allele lengths are equal; we can just use the spanning event's alt allele
+                        // in the case of GGA mode the spanning event might not match an allele in the mergedVC
+                        if (result.containsKey(spanningEvent.getAlternateAllele(0))) {
+                            result.get(spanningEvent.getAlternateAllele(0)).add(h);
+                        }
+                    } else if (spanningEvent.getReference().length() < mergedVC.getReference().length()) {
+                        // spanning event has shorter ref allele than merged VC; we need to pad out its alt allele
+                        final Map<Allele, Allele> spanningEventAlleleMappingToMergedVc
+                                = GATKVariantContextUtils.createAlleleMapping(mergedVC.getReference(), spanningEvent, new ArrayList<>());
```
davidbenjamin: If you feel like going on a tangent, its' weird that `createAlleleMapping` returns a map even though you're only applying it to a single alt allele.  There should be a method to pad a single alt allele to a longer reference.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -168,24 +273,82 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
...
+                        final Allele remappedSpanningEventAltAllele = spanningEventAlleleMappingToMergedVc.get(spanningEvent.getAlternateAllele(0));
+                        // in the case of GGA mode the spanning event might not match an allele in the mergedVC
```
davidbenjamin: Why do we do nothing with the spanning event in such a case?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -168,24 +273,82 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
...
             }
+
+            if (! hasSpanningEvents) {    //no events --> this haplotype supports the reference at this locus
+                result.get(ref).add(h);
+            }
         }
         return result;
     }
 
+    private static VariantContext findMatchingGivenAllele(final List<VariantContext> activeAllelesToGenotype, final VariantContext spanningEvent) {
```
davidbenjamin: Couldn't this conceivably return `List<VariantContext>` in the case of deletions spanning multiallelic GGA vcs?  (Or maybe, multiple GGA vcfs with different alt alleles at the same site).

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -442,7 +442,7 @@ public void writeHeader( final VariantContextWriter vcfWriter, final SAMSequence
     public ActivityProfileState isActive( final AlignmentContext context, final ReferenceContext ref, final FeatureContext features ) {
 
         if ( hcArgs.genotypingOutputMode == GenotypingOutputMode.GENOTYPE_GIVEN_ALLELES ) {
-            final VariantContext vcFromAllelesRod = GenotypingGivenAllelesUtils.composeGivenAllelesVariantContextFromRod(features, ref.getInterval(), false, hcArgs.genotypeFilteredAlleles, logger, hcArgs.alleles);
+            final VariantContext vcFromAllelesRod = GenotypingGivenAllelesUtils.composeGivenAllelesVariantContextFromRod(features, ref.getInterval(), false, hcArgs.genotypeFilteredAlleles, hcArgs.alleles);
```
davidbenjamin: While you're at it you can rename variables with "rod" in their name.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -230,7 +227,7 @@ public void shutdown() {
     @Override
     public ActivityProfileState isActive(final AlignmentContext context, final ReferenceContext ref, final FeatureContext featureContext) {
         if ( MTAC.genotypingOutputMode == GenotypingOutputMode.GENOTYPE_GIVEN_ALLELES ) {
-            final VariantContext vcFromAllelesRod = GenotypingGivenAllelesUtils.composeGivenAllelesVariantContextFromRod(featureContext, ref.getInterval(), false, MTAC.genotypeFilteredAlleles, logger, MTAC.alleles);
+            final VariantContext vcFromAllelesRod = GenotypingGivenAllelesUtils.composeGivenAllelesVariantContextFromRod(featureContext, ref.getInterval(), false, MTAC.genotypeFilteredAlleles, MTAC.alleles);
```
davidbenjamin: Feel free to change "rod" names in Mutect code, too.

src/main/java/org/broadinstitute/hellbender/utils/haplotype/EventMap.java
```diff
@@ -395,6 +396,10 @@ public String toString() {
         return startPosKeySet;
     }
 
+    public Iterator<VariantContext> getSpanningEvents(final int loc) {
```
davidbenjamin: I would call this `getOverlappingEvents` because otherwise it suggests that you're *only* getting spanning deletions.

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtilsUnitTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.testng.annotations.Test;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public class GenotypingGivenAllelesUtilsUnitTest {
+
+    @Test
+    public void testComposeGivenAllelesVariantContextFromRod() {
```
davidbenjamin: More griping about "ROD" names.

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtilsUnitTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+        final SimpleInterval loc = new SimpleInterval("20", 10093568, 10093568);
```
davidbenjamin: These loci don't need to be 8-digit numbers.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtils.java
```diff
@@ -28,36 +29,41 @@
...
-                    }
-                }
-            }
+        final List<VariantContext> variantContextsInFeatureContext = tracker.getValues(allelesBinding, new SimpleInterval(loc));
+        return composeGivenAllelesVariantContextFromVariantList(variantContextsInFeatureContext, loc, snpsOnly, keepFiltered);
+    }
+
+    @VisibleForTesting
+    protected static VariantContext composeGivenAllelesVariantContextFromVariantList(final List<VariantContext> variantContextsInFeatureContext, final Locatable loc, final boolean snpsOnly, final boolean keepFiltered) {
+        final List<VariantContext> rodVcsAtLoc = variantContextsInFeatureContext
```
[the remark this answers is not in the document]
cwhelan: Yes, I was trying to step lightly on renaming variables in this PR but since you want them to go I'm more than happy to.. renamed.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtils.java
```diff
@@ -28,36 +29,41 @@
...
+                .stream()
+                .filter(vc -> vc.getStart() == loc.getStart() &&
+                        (keepFiltered || vc.isNotFiltered()) &&
+                        (!snpsOnly || vc.isSNP()))
```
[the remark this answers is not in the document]
cwhelan: Must have been a legacy thing. Removed.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtils.java
```diff
@@ -28,36 +29,41 @@
...
+                .collect(Collectors.toList());
+
+
+        if (rodVcsAtLoc.isEmpty()) {
+            return null;
         }
+        final List<String> haplotypeSources = rodVcsAtLoc.stream().map(VariantContext::getSource).collect(Collectors.toList());
+        final VariantContext mergedVc = GATKVariantContextUtils.simpleMerge(rodVcsAtLoc, haplotypeSources,
+                GATKVariantContextUtils.FilteredRecordMergeType.KEEP_IF_ANY_UNFILTERED,
```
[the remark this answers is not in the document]
cwhelan: Thanks for catching that. Changed to your suggestion.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -111,24 +112,98 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
         return startPosKeySet;
     }
 
+    /**
+     * Utility class to represent a location and set of Alleles, used to track redundant variant contexts by adding
```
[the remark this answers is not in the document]
cwhelan: Re-wrote the comment, look better to you now?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -111,24 +112,98 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
...
+            final LocationAndAlleles that = (LocationAndAlleles) o;
+
+            if (loc != that.loc) return false;
+            return alleles != null ? alleles.equals(that.alleles) : that.alleles == null;
+        }
+
+        @Override
+        public int hashCode() {
+            int result = loc;
+            result = 31 * result + (alleles != null ? alleles.hashCode() : 0);
```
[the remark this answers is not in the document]
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -111,24 +112,98 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
...
+     * Returns the list of variant contexts active at this location, using either the list of active haplotypes or
+     * the list of active alleles to genotypes if we are in GGA mode as candidates. This method will include events
+     * that span the current location if includeSpanningEvents is set to true; otherwise it will only include events that
+     * have loc as their start position.
+     * @param haplotypes Haplotypes for the current active region.
+     * @param loc The start position we are genotyping
+     * @param activeAllelesToGenotype The list of given alleles for the current active region, empty unless we are in GGA mode
+     * @param includeSpanningEvents If true, will also return events that span loc
+     */
+    protected static List<VariantContext> getVCsAtThisLocation(final List<Haplotype> haplotypes,
```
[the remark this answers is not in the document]
cwhelan: I've done this refactoring and it does look a lot cleaner I think; thanks for the suggestion. Let me know how it looks now.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -111,24 +112,98 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
...
+                        if (! uniqueLocationsAndAlleles.contains(locationAndAlleles)) {
+                            uniqueLocationsAndAlleles.add(locationAndAlleles);
+                            results.add(v);
+                        }
+                    });
         } else { // we are in GGA mode!
             int compCount = 0;
             for( final VariantContext compVC : activeAllelesToGenotype ) {
-                if( compVC.getStart() == loc ) {
+                if( compVC.getStart() <= loc && compVC.getEnd() >= loc) {
```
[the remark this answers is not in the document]
cwhelan: yeah, more weird old legacy names.. done, although I did keep the "Comp" in the `vcSourceName` in case anyone depended on that downstream.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -111,24 +112,98 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
...
+                    if (! (includeSpanningEvents || compVC.getStart() == loc)) {
```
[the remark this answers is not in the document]
cwhelan: Passing on this suggestion for now since I feel like it wouldn't be much clearer to my eye. Let me know if you really want me to re-write it that way.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -168,24 +273,82 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
...
-            } else {
-                //TODO: this only works if eventsAtThisLoc's and mergedVC's alleles are ordered identically.
-                //TODO: this is in fact the case (see AssemblyBasedCallerUtils::makeMergedVariantContext), but
-                //TODO: that's not good enough
-                for (int n = 0; n < eventsAtThisLoc.size(); n++) {
-                    if (haplotypeVC.hasSameAllelesAs(eventsAtThisLoc.get(n))) {
-                        result.get(mergedVC.getAlternateAllele(n)).add(h);
-                        break;
+
+            final Iterator<VariantContext> spanningEvents = h.getEventMap().getSpanningEvents(loc);
```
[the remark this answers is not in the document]
cwhelan: I was avoiding creating a new collection but ended up changing this at your suggestion.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -168,24 +273,82 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
...
+
+            final boolean hasSpanningEvents = spanningEvents.hasNext();
+
```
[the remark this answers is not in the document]
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
```diff
@@ -168,24 +273,82 @@ public CalledHaplotypes(final List<VariantContext> calls, final Set<Haplotype> c
...
+            while (spanningEvents.hasNext()) {
+                final VariantContext spanningEvent = spanningEvents.next();
+                if (spanningEvent.getStart() == loc) {
+                    // the event starts at the current location
+
+                    if (spanningEvent.getReference().length() == mergedVC.getReference().length()) {
+                        // reference allele lengths are equal; we can just use the spanning event's alt allele
+                        // in the case of GGA mode the spanning event might not match an allele in the mergedVC
```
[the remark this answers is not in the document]
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
cwhelan: I think you're right but I'm going to pass on this refactor for now.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
cwhelan: After looking at this again, changed it to a boolean method that returns true if it finds a matching given allele, since that's all it was being used for anyway.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/utils/haplotype/EventMap.java
cwhelan: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtilsUnitTest.java
cwhelan: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/genotyper/GenotypingGivenAllelesUtilsUnitTest.java
cwhelan: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
cwhelan: What's in the merged VC is either the union of the alt alleles from the active haplotypes or the GGA alleles. The comment refers to the case when an assembled haplotype supports an event that isn't present in the merged VC. HaplotypeCaller's GGA mode behavior is to ignore active events that aren't in the given alleles list so my assumption was that those haplotypes should be dropped from the event map. Does that seem correct to you?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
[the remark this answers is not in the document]
davidbenjamin: Fine with me.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
davidbenjamin: No problem.

cwhelan:
@davidbenjamin Thanks for your review; I've tried to address your comments but let me know if anything still looks like it could use some work or if you spot anything new. Still waiting for tests to pass on this update (bar the GenomicsDB tests that I expect to fail).

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerGenotypingEngine.java
davidbenjamin: That does sound correct, and I see that you hooked up `SomaticGenotypingEngine` to the version of this method that doesn't get the GGA alleles, which is correct because they have already been injected into the haplotypes.  So, yes, I think everything is okay.

ldgauthier:
I agree that AD should be 6 for the * allele.  It doesn't look like it's entirely trivial to implement in the GATK code since when we do the allele index remapping we expect dropped alleles to be low quality and don't aggregate their ADs.  But that issue aside, I'm concerned that the CombineGVCF and GenomicsDB results don't match.  @kgururaj is it possible #4645 changed something relevant to this bug?

kgururaj:
I'll try to explain my understanding of how spanning deletions and the associated attributes are computed in GenomicsDB - you can let me know what part of the logic I should fix.

1. The spanning deletion allele at 20 : 10068160 corresponds to the deletion at 20 : 10068158
1. The spanning deletion alleles at positions 20 : 10068160-10068175 correspond to the deletion at 20 : 10068160.
1. In GenomicsDB:
    1. ALT allele GTATATATATATGTA -> G at 20 : 10068160 is considered a deletion.
    1. GTATATATATATGTA -> * at 20 : 10068160 is NOT considered a deletion. Should it be considered a deletion ALT allele and be part of the min PL computation?
    1. Hence, GTATATATATATGTA -> G is the deletion ALT allele with min PL value. All the AD and PL values in the subsequent positions correspond to this ALT allele.

Let me know if I should fix 3.ii

ldgauthier:
Yes, the * at 10068160 should be considered a deletion.  It looks like that must be necessary to match the CombineGVCFs results.

src/test/resources/org/broadinstitute/hellbender/tools/haplotypecaller/expected.testGVCFMode.gatk4.alleleSpecific.g.vcf
```diff
@@ -1124,8 +1124,8 @@
 20	10068149	.	T	<NON_REF>	.	.	END=10068151	GT:DP:GQ:MIN_DP:PL	0/0:31:63:31:0,63,945
 20	10068152	.	A	<NON_REF>	.	.	END=10068155	GT:DP:GQ:MIN_DP:PL	0/0:29:57:28:0,57,855
 20	10068156	.	A	<NON_REF>	.	.	END=10068157	GT:DP:GQ:MIN_DP:PL	0/0:27:51:27:0,51,765
-20	10068158	.	GTGTATATATATA	G,<NON_REF>	66.73	.	AS_RAW_BaseQRankSum=|-0.7,1|NaN;AS_RAW_MQ=5282.00|8882.00|0.00;AS_RAW_MQRankSum=|0.3,1|NaN;AS_RAW_ReadPosRankSum=|0.5,1|NaN;AS_SB_TABLE=0,3|2,2|0,0;BaseQRankSum=-0.652;DP=29;ExcessHet=3.0103;MLEAC=1,0;MLEAF=0.500,0.00;MQRankSum=0.328;RAW_MQ=93364.00;ReadPosRankSum=0.524	GT:AD:DP:GQ:PGT:PID:PL:SB	0/1:3,4,0:7:57:0|1:10068158_GTGTATATATATA_G:104,0,57,114,69,183:0,3,2,2
-20	10068160	.	GTATATATATATGTA	G,<NON_REF>	7.57	.	AS_RAW_BaseQRankSum=|-0.6,1|NaN;AS_RAW_MQ=9723.00|1682.00|0.00;AS_RAW_MQRankSum=|-0.8,1|NaN;AS_RAW_ReadPosRankSum=|0.0,1|NaN;AS_SB_TABLE=2,3|0,2|0,0;BaseQRankSum=-0.566;DP=28;ExcessHet=3.0103;MLEAC=1,0;MLEAF=0.500,0.00;MQRankSum=-0.712;RAW_MQ=87005.00;ReadPosRankSum=0.000	GT:AD:DP:GQ:PGT:PID:PL:SB	0/1:5,2,0:7:44:1|0:10068158_GTGTATATATATA_G:44,0,134,60,141,201:2,3,0,2
```
ldgauthier: These changes are all great.

ldgauthier:
@cwhelan my understanding is that this is blocked because tests won't pass until there's a GDB update -- yes?

cwhelan:
@ldgauthier Yes, this was just waiting on some resolution to the double-spanning deletion issue reported above. If a new GDB update makes the output of `GenomicsDBImport` concordant with that of `CombineGVCFs` in terms of the DP and AD's listed above, I think I can update the test files to reflect the new behavior. I still think that the behavior of Combine and GDBI is not exactly right as I mentioned in the last line of my long comment above, but maybe that's something that could be spun out into its own issue to fix independently, as long as we're OK with running with the current CombineGVCFs behavior in the meantime.

ldgauthier:
I know you don't need this for your project anymore, but I'd rather get this in and leave the AD correction for overlapping deletions for another issue.

cwhelan:
@ldgauthier I've updated `GenomicsDBImportIntegrationTest.testGenomicsDBImportFileInputsAgainstCombineGVCFWithNonDiploidData` to use its own expected output VCF rather than running `CombineGVCFS` and comparing to the output of that. Let me know if you want to take one more look, or have someone else do so, before giving this a final thumb.

ldgauthier:
If you add an issue for the combining of AD for multiple spanning deletions (which will ultimately improve your new expected results here) then I'll consider this :+1:

PR #4964, opened 2018-06-28
Added a new `Walker` abstract class to be used for all built-in walkers.

jonn-smith:
Added the following methods to `GATKTool`:

- `getReferenceDataSource()`
- `getReadsDataSource()`
- `getFeatureManager()`

`Walker` inherits directly from `GATKTool` and overrides these methods to throw an exception if they are called.

No walker should need to directly access the data.


src/main/java/org/broadinstitute/hellbender/engine/Walker.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * An abstract class to represent built-in walkers that inherit directly from {@link GATKTool}.
+ * Created by jonn on 6/28/18.
+ */
+public abstract class Walker extends GATKTool {
+
+    @Override
+    final protected ReferenceDataSource getReferenceDataSource() {
+        throw new GATKException("Should never access ReferenceDataSource in child classes of AssemblyRegionWalker.");
```
magicDGS: `AssemblyRegionWalker` should be `Walker` - here and below!

magicDGS:
Sorry for the multiple edits, something was wrong with my keyboard and it was submitting without my action.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
jonn-smith: Good catch!

Fixed!

droazen:
@magicDGS The problem with exposing the datasources to walkers is that they would be able to invalidate the entire traversal. For example, a `ReadWalker` could alter the traversal intervals on the reads datasource mid-way through traversal from within `apply()`, or it could cause the reads iterator used by the engine to get closed by issuing a separate `iterator()` call on the datasource, which would cause the rest of the traversal to fail.

This is why I feel strongly that the datasource objects should not be directly accessible to walker-based tools. Note that it's still possible for walkers to create their own, separate datasources without reaching into the ones used by the engine, or a tool author can extend `GATKTool` directly rather than one of the walker base classes and have the freedom to access everything (which was not possible before this PR).

magicDGS:
I am aware that those methods should be definetively implemented in the abstract class - but it could be recommended *NOT TO OVERRIDE* unless you know what you are doing. I know the problems of having them exposed, but also I can see some potential to be accessible from an implemented walker when the user knows what is doing...

src/main/java/org/broadinstitute/hellbender/engine/FeatureWalker.java
```diff
@@ -79,7 +78,7 @@ private void initializeDrivingFeatures() {
      * Subclasses can override to provide their own behavior but default implementation should be suitable for most uses.
      */
     @Override
-    public void traverse() {
+    public final void traverse() {
```
droazen: This `final` conflicts with the "Subclasses can override to provide their own behavior" comment above. We typically don't make `traverse()` final to allow for specializations of walker traversals (like `TwoPassVariantWalker` overriding `traverse()` from `VariantWalker`). I think the comment should be modified to read:

"You should only override traverse() if you are writing a new walker base class in the engine package that extends this class. It is not meant to be overridden by tools outside of the engine package."

While you're at it, can you make the same change to the comments for `traverse()` in the other walker base classes (and remove the finals).

src/main/java/org/broadinstitute/hellbender/engine/IntervalWalker.java
```diff
@@ -56,7 +54,7 @@ protected final void onStartup() {
     }
 
     @Override
-    public void traverse() {
+    public final void traverse() {
```
droazen: As discussed above, don't make this final, but do add the comment "You should only override traverse() if you are writing a new walker base class in the engine package that extends this class. It is not meant to be overridden by tools outside of the engine package."

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
```diff
@@ -146,7 +146,7 @@ protected final void onStartup() {
      * and including deletions only if {@link #includeDeletions()} returns {@code true}.
      */
     @Override
-    public void traverse() {
+    public final void traverse() {
```
droazen: As discussed above, don't make this final, but do modify the comment to read: "You should only override traverse() if you are writing a new walker base class in the engine package that extends this class. It is not meant to be overridden by tools outside of the engine package."

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
```diff
@@ -29,7 +24,7 @@
  * ReadWalker authors must implement the apply() method to process each read, and may optionally implement
  * onTraversalStart() and/or onTraversalSuccess(). See the PrintReadsWithReference walker for an example.
  */
-public abstract class ReadWalker extends GATKTool {
+public abstract class ReadWalker extends Walker {
```
droazen: Update docs for `ReadWalker.traverse()` to read "You should only override traverse() if you are writing a new walker base class in the engine package that extends this class. It is not meant to be overridden by tools outside of the engine package." instead of the current "Subclasses can override to provide own behavior"

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
```diff
@@ -43,7 +39,7 @@
  * Internally, the reads are loaded in chunks called read shards, which are then subdivided into active/inactive regions
  * for processing by the tool implementation. One read shard is created per contig.
  */
-public abstract class AssemblyRegionWalker extends GATKTool {
+public abstract class AssemblyRegionWalker extends Walker {
```
droazen: Update docs for `AssemblyRegionWalker.traverse()` to read "You should only override traverse() if you are writing a new walker base class in the engine package that extends this class. It is not meant to be overridden by tools outside of the engine package." instead of the current "Subclasses can override to provide own behavior"

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
```diff
@@ -0,0 +1,25 @@
+package org.broadinstitute.hellbender.engine;
+
+import org.broadinstitute.hellbender.exceptions.GATKException;
+
+/**
+ * An abstract class to represent built-in walkers that inherit directly from {@link GATKTool}.
```
droazen: Change to: 

"Base class for pre-packaged walker traversals in the GATK engine. Classes such as ReadWalker and VariantWalker in the engine package that implement a standardized traversal should extend this class rather than GATKTool. Actual concrete tool classes should extend one of the specific walker base types (such as ReadWalker) or GATKTool directly rather than this class.

Classes that extend Walker will be unable to directly access engine data sources unless they are in the engine package. This is to allow walker base classes such as ReadWalker direct datasource access while disallowing it for concrete walker tool implementations, which should get their data via apply()."

src/main/java/org/broadinstitute/hellbender/engine/AbstractConcordanceWalker.java
```diff
@@ -34,7 +33,7 @@
  * Created by Takuto Sato 1/30/17, abstractified by David Benjamin on 2/22/17.
  * {@link #onTraversalStart}, {@link #onTraversalSuccess} and/or {@link #closeTool}.
  */
-public abstract class AbstractConcordanceWalker extends GATKTool {
+public abstract class AbstractConcordanceWalker extends Walker {
```
droazen: In the time since you opened this PR, more walker base classes have been added to the engine. Can you modify them to extend `Walker` as well?

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -157,6 +158,45 @@
      */
     List<SimpleInterval> intervalsForTraversal;
 
+    /**
+     * Get the {@link ReferenceDataSource} for this {@link GATKTool}.
+     * Will throw a {@link GATKException} if the reference is null.
+     * Clients are expected to call the {@link #hasReference()} method prior to calling this.
+     * @return the {@link ReferenceDataSource} for this {@link GATKTool}.  Never {@code null}.
+     */
+    protected ReferenceDataSource getReferenceDataSource() {
```
droazen: Rename to `directlyAccessEngineReferenceDataSource()` to deter casual callers and improve the odds that they'd check the javadoc.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -157,6 +158,45 @@
...
+        return reference;
+    }
+
+    /**
+     * Get the {@link ReadsDataSource} for this {@link GATKTool}.
+     * Will throw a {@link GATKException} if the reads are null.
+     * Clients are expected to call the {@link #hasReads()} method prior to calling this.
+     * @return the {@link ReadsDataSource} for this {@link GATKTool}.  Never {@code null}.
+     */
+    protected ReadsDataSource getReadsDataSource() {
```
droazen: Rename to `directlyAccessEngineReadsDataSource()` to deter casual callers and improve the odds that they'd check the javadoc.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -157,6 +158,45 @@
...
+        return reads;
+    }
+
+    /**
+     * Get the {@link FeatureManager} for this {@link GATKTool}.
+     * Will throw a {@link GATKException} if the features are null.
+     * Clients are expected to call the {@link #hasFeatures()} method prior to calling this.
+     * @return the {@link FeatureManager} for this {@link GATKTool}.  Never {@code null}.
+     */
+    protected FeatureManager getFeatureManager() {
```
droazen: Rename to `directlyAccessEngineFeatureManager()` to deter casual callers and improve the odds that they'd check the javadoc.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -157,6 +158,45 @@
      */
     List<SimpleInterval> intervalsForTraversal;
 
+    /**
+     * Get the {@link ReferenceDataSource} for this {@link GATKTool}.
```
droazen: Add disclaimer: "Should only be called by walker base classes in the engine (such as ReadWalker), or by "free-form" tools that extend the GATKTool class directly rather than one of the built-in walker types. Tools that extend a walker type should get their data via apply() rather than directly accessing the engine datasources."

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -157,6 +158,45 @@
...
+     */
+    protected ReferenceDataSource getReferenceDataSource() {
+        if ( reference == null ) {
+            throw new GATKException("Attempted to retrieve null reference!");
+        }
+        return reference;
+    }
+
+    /**
+     * Get the {@link ReadsDataSource} for this {@link GATKTool}.
```
droazen: Add disclaimer: "Should only be called by walker base classes in the engine (such as ReadWalker), or by "free-form" tools that extend the GATKTool class directly rather than one of the built-in walker types. Tools that extend a walker type should get their data via apply() rather than directly accessing the engine datasources."

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
```diff
@@ -157,6 +158,45 @@
...
+     */
+    protected ReadsDataSource getReadsDataSource() {
+        if ( reads == null ) {
+            throw new GATKException("Attempted to retrieve null reads!");
+        }
+        return reads;
+    }
+
+    /**
+     * Get the {@link FeatureManager} for this {@link GATKTool}.
```
droazen: Add disclaimer: "Should only be called by walker base classes in the engine (such as ReadWalker), or by "free-form" tools that extend the GATKTool class directly rather than one of the built-in walker types. Tools that extend a walker type should get their data via apply() rather than directly accessing the engine datasources."

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
```diff
@@ -0,0 +1,25 @@
...
+ * Created by jonn on 6/28/18.
+ */
+public abstract class Walker extends GATKTool {
+
+    @Override
+    final protected ReferenceDataSource getReferenceDataSource() {
+        throw new GATKException("Should never access ReferenceDataSource in child classes of Walker.");
```
droazen: Change to: "Walker tools outside of the engine package should not directly access the engine reference datasource. They should get their reference data via apply() instead. Tools that need direct datasource access (eg., to implement custom traversal patterns) should extend GATKTool directly rather than a walker class, or introduce a new walker base class for the new traversal."

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    final protected ReadsDataSource getReadsDataSource() {
+        throw new GATKException("Should never access ReadsDataSource in child classes of Walker.");
```
droazen: Change to: "Walker tools outside of the engine package should not directly access the engine reads datasource. They should get their reads data via apply() instead. Tools that need direct datasource access (eg., to implement custom traversal patterns) should extend GATKTool directly rather than a walker class, or introduce a new walker base class for the new traversal."

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    final protected FeatureManager getFeatureManager() {
+        throw new GATKException("Should never access FeatureManager in child classes of Walker.");
```
droazen: Change to: "Walker tools outside of the engine package should not directly access the engine feature manager. They should get their feature data via apply() instead. Tools that need direct datasource access (eg., to implement custom traversal patterns) should extend GATKTool directly rather than a walker class, or introduce a new walker base class for the new traversal."

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.broadinstitute.hellbender.exceptions.GATKException;
+
+/**
+ * An abstract class to represent built-in walkers that inherit directly from {@link GATKTool}.
+ * Created by jonn on 6/28/18.
+ */
+public abstract class Walker extends GATKTool {
+
+    @Override
```
droazen: Add method-level javadoc comment stating that we are overriding this method to prevent walker tool implementations outside of the engine package from directly accessing the engine datasources, since walker tools should get their data via apply() instead.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
```diff
@@ -0,0 +1,25 @@
...
+    final protected ReferenceDataSource getReferenceDataSource() {
+        throw new GATKException("Should never access ReferenceDataSource in child classes of Walker.");
+    }
+
+    @Override
```
droazen: Add method-level javadoc comment stating that we are overriding this method to prevent walker tool implementations outside of the engine package from directly accessing the engine datasources, since walker tools should get their data via apply() instead.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
```diff
@@ -0,0 +1,25 @@
...
+    final protected ReadsDataSource getReadsDataSource() {
+        throw new GATKException("Should never access ReadsDataSource in child classes of Walker.");
+    }
+
+    @Override
```
droazen: Add method-level javadoc comment stating that we are overriding this method to prevent walker tool implementations outside of the engine package from directly accessing the engine datasources, since walker tools should get their data via apply() instead.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
```diff
@@ -0,0 +1,25 @@
+package org.broadinstitute.hellbender.engine;
+
+import org.broadinstitute.hellbender.exceptions.GATKException;
+
+/**
+ * An abstract class to represent built-in walkers that inherit directly from {@link GATKTool}.
+ * Created by jonn on 6/28/18.
+ */
+public abstract class Walker extends GATKTool {
```
lbergelson: It might be good to rename this to `WalkerBase` or `AbstractWalker`.  Walker sounds like a concrete class.

src/main/java/org/broadinstitute/hellbender/engine/FeatureWalker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/IntervalWalker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/LocusWalker.java
jonn-smith: Fixed

src/main/java/org/broadinstitute/hellbender/engine/ReadWalker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/AssemblyRegionWalker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/AbstractConcordanceWalker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/GATKTool.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
jonn-smith: Fixed.

src/main/java/org/broadinstitute/hellbender/engine/Walker.java
jonn-smith: `WalkerBase` it is!

PR #4967, opened 2018-06-29
Qual = 0 sites don't count as polymorphic for GVCF mode

davidbenjamin:
@ldgauthier This is for #4614.


davidbenjamin:
I was assuming I would change the default in the same PR that makes new qual default.

ldgauthier:
Fair point.  I'll try to help remember.

davidbenjamin:
Is this PR good to merge then?

ldgauthier:
:+1:

PR #4969, opened 2018-06-29
Improve MQ calculation accuracy

ldgauthier:
After the switch to rawMQ for the reducible implementation we no longer take the median of MQ values across all samples (yay!) but there were some accuracy problems at sites with a lot of uninformative reads (boo!)

I tried to calculate the depth over variant samples previously, but the approximations weren't perfect.  Now we keep track of the depth used to find the RMS of MQ at the end of joint calling.


ldgauthier:
@jamesemery Can you take a look?  (This doesn't have to go into the next release.)

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
```diff
@@ -181,6 +186,24 @@ public void assertMatchingGenotypesFromTileDB(File input, File expected, Locatab
...
 
+    @DataProvider(name="VCFdata")
+    public Object[][] getVCFdata() {
+        return new Object[][]{
+                {new File[]{getTestFile("../CombineGVCFs/YRIoffspring.chr20snippet.g.vcf"), NA12878_HG37}, getTestFile("newMQcalc.genotyped.vcf"), "20:10000000-10100000", NO_EXTRA_ARGS, b37_reference_20_21}
+        };
+    }
+
+    @Test (dataProvider = "VCFdata")
+    public void assertMatchingGenotypesFromGenomicsDB_vidmapHack(File[] inputs, File expected, String interval, List<String> additionalArguments, String reference) throws IOException {
```
jamesemery: Could you add a comment explaining this test? It doesn't seem related to the rest of the branch.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
```diff
@@ -42,9 +42,12 @@
 @DocumentedFeature(groupName=HelpConstants.DOC_CAT_ANNOTATORS, groupSummary=HelpConstants.DOC_CAT_ANNOTATORS_SUMMARY, summary="Root mean square of the mapping quality of reads across all samples (MQ)")
 public final class RMSMappingQuality extends InfoFieldAnnotation implements StandardAnnotation, ReducibleAnnotation {
     private static final RMSMappingQuality instance = new RMSMappingQuality();
+    public static final int NUM_LIST_ENTRIES = 2;
+    public static final int SUM_OF_SQUARES_INDEX = 0;
+    public static final int TOTAL_DEPTH_INDEX = 1;
 
     @Override
-    public String getRawKeyName() { return GATKVCFConstants.RAW_RMS_MAPPING_QUALITY_KEY;}
+    public String getRawKeyName() { return GATKVCFConstants.RAW_MAPPING_QUALITY_WITH_DEPTH_KEY;}   //new key for the two-value MQ data to prevent version mismatch catastrophes
```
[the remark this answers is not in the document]
jamesemery: Perhaps we could add a check and emit a warning to the user if there is a mismatch between the versions of the rmsMappingQualityKey in their files?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
```diff
@@ -42,9 +42,12 @@
 @DocumentedFeature(groupName=HelpConstants.DOC_CAT_ANNOTATORS, groupSummary=HelpConstants.DOC_CAT_ANNOTATORS_SUMMARY, summary="Root mean square of the mapping quality of reads across all samples (MQ)")
 public final class RMSMappingQuality extends InfoFieldAnnotation implements StandardAnnotation, ReducibleAnnotation {
     private static final RMSMappingQuality instance = new RMSMappingQuality();
+    public static final int NUM_LIST_ENTRIES = 2;
```
jamesemery: Could you update the header comments to reflect the new annotation? Just add a line specifying what the count means

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
```diff
@@ -97,6 +100,8 @@
                 {getTestFile( "testAlleleSpecificAnnotations.CombineGVCF.output.g.vcf"), getTestFile( "testAlleleSpecificAnnotations.CombineGVCF.expected.g.vcf"), Arrays.asList( "-A", "ClippingRankSumTest", "-G", "AS_StandardAnnotation", "-G", "StandardAnnotation"), b37_reference_20_21},
                 //all sites not supported yet see https://github.com/broadinstitute/gatk-protected/issues/580 and  https://github.com/broadinstitute/gatk/issues/2429
                 //{getTestFile(basePairGVCF), getTestFile( "gvcf.basepairResolution.includeNonVariantSites.gatk3.7_30_ga4f720357.expected.vcf"), Collections.singletonList("--"+GenotypeGVCFs.ALL_SITES_LONG_NAME) //allsites not supported yet
+                //Test for new RAW_MQandDP annotation format
```
jamesemery: This test doesn't actually test  RAW_MWandDP output as they are added to attributes to ignore. Maybe spin this off into its own test to make sure that the assertions made on the annotations are run?

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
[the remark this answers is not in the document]
ldgauthier: I added a check in the finalize method in case GenotypeGVCFs gets run on old data.  GenomicsDB behavior is largely outside of our control here.  I don't have a great idea for CombineGVCFs though.  I thought about adding a check against ReducibleAnnotation.getDeprecatedKeyName(), but I feel like having getDeprecatedKeyName() return null for the other annotations and querying the annotationMap for null is an invitation for trouble.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java
```diff
@@ -223,6 +226,13 @@ private static double parseRawDataString(String rawDataString) {
     @VisibleForTesting
     static int getNumOfReads(final VariantContext vc,
                              final ReadLikelihoods<Allele> likelihoods) {
+        if(vc.hasAttribute(GATKVCFConstants.RAW_MAPPING_QUALITY_WITH_DEPTH_KEY)) {
```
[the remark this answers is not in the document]
ldgauthier: I updated the GATK Javadoc for the annotation -- is that what you meant?

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
ldgauthier: I goofed and copied the genotypes comparison when I meant do compare annotations.  I fixed the name and made sure MQ is actually being compared.  (FYI the vidmap hack is because I need to change the GDB json since I changed the MQ annotation name.  I'll update this once the protobuf update goes in.)

ldgauthier:
The AS_MQ never suffered from this issue because it uses AD for (allele-specific) depth instead of the INFO DP.  The sum of the squared MQs there was allocated based on informative reads and the AD represents informative reads, so the data there was always in lock-step.

ldgauthier:
I put this into a different branch because I upgraded GDB to fix the weird error.  I don't want this feature to go into the 4.0.9.0 release so I'll do a PR of the new branch after.

ldgauthier:
Provided Travis tests pass, does this still have your :+1: @jamesemery ?  The ReblockGVCF tool had a less elegant MQ solution so now I'm having it output both versions so that we don't need to re-reprocess the gnomAD v3 GVCFs.

indraniel:
Just as a note.  This change in line 251 of [`gatk/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java`](https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/RMSMappingQuality.java) of function `parseRawDataString`:

```
final int squareSum = Integer.parseInt(parsed[SUM_OF_SQUARES_INDEX]);
```

results in the following type of GVCF/VCF parsing error:

```
A USER ERROR has occurred: Bad input: malformed RAW_MQ annotation: 3415207168,1749038
```

 when the `RAW_MQ` first index is greater than `Integer.MAX_VALUE`.

PR #4970, opened 2018-06-29
(SV) slim down REF column for CPX variants

SHuang-Broad:
Made a stupid decision to output all bases of the affected reference region, which significantly increased the VCF file size.

Now we follow the example for `DEL` variants, i.e. we only put the reference base in POS in the REF column.


src/main/java/org/broadinstitute/hellbender/tools/spark/sv/discovery/inference/SegmentedCpxVariantSimpleVariantExtractor.java
```diff
@@ -835,9 +835,10 @@ private void extractFrontAndRearInsertions(final VariantContext complexVC, final
             }
 
             if (firstRefSegmentIdx != altArrangement.size() - 1) {
-                final Allele anchorBaseRefAlleleRear = Allele.create(refBases[refBases.length - 2], true);
-                final SimpleInterval startAndStop = makeOneBpInterval(complexVC.getContig(), complexVC.getEnd());
-                final VariantContextBuilder rearIns = getInsFromOneEnd(false, firstRefSegmentIdx, startAndStop, anchorBaseRefAlleleRear, refSegmentLengths, altArrangement, true);
+                final int pos = complexVC.getEnd();
```
cwhelan: Why is it the end of the VC here and not the start?

SHuang-Broad: The reason was because this is for extracting rear insertion.
Example:
```
SEGMENTS=chr1:100-200,chr1:200-300,chr1:300-400,chr1:400-500
ALT_ARRANGEMENT=chrOther:100-158,1,-3,4,UINS-433
```
Here we would emit four records:
a front insertion of 59 bases at position chr1:100, a deletion of chr1:200-300, an inversion of chr1:300-400, and a rear insertion of 433 bases at position of chr1:500.

PR #4980, opened 2018-07-03
fixes new qual regression in 4.0.5.0

davidbenjamin:
@droazen Can you review this once tests pass?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/AlleleFrequencyCalculator.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/AlleleFrequencyCalculator.java
@@ -102,7 +102,7 @@ public final class AlleleFrequencyCalculator extends AFCalculator {
                 final int[] nonVariantIndices = nonVariantIndicesByPloidy.get(ploidy);
                 final double[] nonVariantLog10Posteriors = MathUtils.applyToArray(nonVariantIndices, n -> log10GenotypePosteriors[n]);
                 // when the only alt allele is the spanning deletion the probability that the site is non-variant
-                // may be so close to 1 thatfinite precision error in log10SumLog10 yields a positive value,
+                // may be so close to 1 that finite precision error in log10SumLog10 yields a positive value,
                 // which is bogus.  Thus we cap it at 0.
                 log10PNoVariant += Math.min(0,MathUtils.log10SumLog10(nonVariantLog10Posteriors));
             }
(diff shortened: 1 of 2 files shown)


ldgauthier:
This is the second precision bug we've had related to the AlleleFrequencyCalculator recently.  Adding epsilons and min(zero, littleValue) seems like a code smell, but no suggestions immediately jump to mind.

EvanTheB:
https://github.com/broadinstitute/gatk/commit/9a4fb6d4e5fd2a226db820a8b4062c139b66e2ef
There is a standard math variant of log(x) for calculating the common log(1+x). It avoids numerical error for small x.

davidbenjamin:
@EvanTheB You have a point, because the implementation of `logSumLog` (okay it's log10 but that's ugly so let's pretend we live in a parallel universe where the GATK uses only natural log) is (assume that `B < A` below)
```
log (10^A + 10^B . . .) = A + log(1 + 10^(B - A) . . .)
```
the accuracy of which could certainly be improved via the dedicated `log1p` method, especially when `B << A`.  But accuracy alone doesn't necessarily address the issue.  I mean, the results we're getting are like 10^-21 instead of 0.  If the numerical answer is epsilon, the crucial thing is that epsilon must be negative, not that abs(epsilon) be minimized.  Now, one could always truncated the (alternating) Taylor series for `log(1 + 10^(B - A) . . .)` at some fixed order that guarantees it's an underestimate, but first, even if that's the implementation of `log1p` it's brittle to rely on that; and second, it's not guaranteed that the leading term in dominant.

Another alternative is instead of calculating the probability of all the non-variant (that is, hom ref but also het ref / span del and hom span del) genotypes add up just the variant genotypes and subtract those probabilities from 1.  I hesitate to do this because it's less efficient when there are many alt alleles (there are many more variant genotypes than non-variant).  Granted, it's only when span dels are present, but I still don't like it.

src/main/java/org/broadinstitute/hellbender/tools/walkers/genotyper/afcalc/AlleleFrequencyCalculator.java
```diff
@@ -101,7 +101,10 @@ public AFCalculationResult getLog10PNonRef(final VariantContext vc, final int de
                 nonVariantIndicesByPloidy.computeIfAbsent(ploidy, p -> genotypeIndicesWithOnlyRefAndSpanDel(p, alleles));
                 final int[] nonVariantIndices = nonVariantIndicesByPloidy.get(ploidy);
                 final double[] nonVariantLog10Posteriors = MathUtils.applyToArray(nonVariantIndices, n -> log10GenotypePosteriors[n]);
-                log10PNoVariant += MathUtils.log10SumLog10(nonVariantLog10Posteriors);
+                // when the only alt allele is the spanning deletion the probability that the site is non-variant
+                // may be so close to 1 thatfinite precision error in log10SumLog10 yields a positive value,
+                // which is bogus.  Thus we cap it at 0.
+                log10PNoVariant += Math.min(0,MathUtils.log10SumLog10(nonVariantLog10Posteriors));
```
droazen: Is it possible to add a unit test that fails before this fix, and passes after it?

PR #4982, opened 2018-07-04
Add optional summary table output to ValidateBasicSomaticShortMutations

davidbenjamin:
With this addition, summarizing the results of the MC3 validation will use more GATK, less scripting.


davidbenjamin:
@LeeTL1220 do you mind reviewing?

PR #4984, opened 2018-07-05
Created "variantcalling" test group and split them off from the rest of the integration tests for runtime purposes

### LINKED ISSUE #4983: Split travis integration tests into two jobs to reduce test runtime
Currently the travis CI integration test task is the longest running task, which currently clocks in uncomfortably close to 70 minute hard limit on task runtime. One solution to this would be to split off a second job encompassing some of the longest running integration tests, namely a `VariantCallingIntegrationTests` tag. If we wanted to split the integration tests further then we could also consider splitting off the spark integration tests, which also take some time.

jamesemery:
Fixes #4983

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -423,7 +423,7 @@ test {
         } else if (TEST_TYPE == "all"){
             //include everything
         } else {
-            excludeGroups "cloud", "bucket", "python", "funcotatorValidation", "variantcalling"
+            excludeGroups "cloud", "bucket", "python", "funcotatorValidation"
         }
     }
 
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -408,19 +408,22 @@ test {
...
             includeGroups "spark"
-            excludeGroups "cloud", "bucket", "python", "funcotatorValidation"
+            excludeGroups "cloud", "bucket", "python", "funcotatorValidation", "variantcalling"
         } else if (TEST_TYPE == "python") {
             includeGroups "python"
         } else if (TEST_TYPE == "all"){
             //include everything
         } else {
-            excludeGroups "cloud", "bucket", "python", "funcotatorValidation"
+            excludeGroups "cloud", "bucket", "python", "funcotatorValidation", "variantcalling"
```
droazen: I think that the `variantcalling` tests should not be excluded in the `else` clause here, since the intention of the else clause is to include all the "normal" (unit/integration) tests, I believe.

PR #4988, opened 2018-07-06
Memory-less HMM for baseline state and silent CNV class

mbabadi:
This slight change in the copy-number calling model is expected to (1) bring the interpretation of the model hyperparameters closer to what is implied, and (2) perhaps represent the reality of CNV events better.


mbabadi:
@samuelklee this is a short PR, though, please review my math carefully.

samuelklee:
Thanks for putting this together!  Any chance we can run some subset of the evaluations while I review?  @asmirnov239 this might be a good time to set up some of the evaluations in FC.

samuelklee:
This looks OK to me.  If I understand correctly, the decaying probability to stay in a class/state now only applies to the silent class and the baseline copy-number state?

@asmirnov239 Let's try to do a quick run on GPC2 to see how performance changes.  You might want to sweep `p-alt`, `p-active`, and `cnv-coherence-length` again, since I don't think we should expect the optimal parameters to be the same across both HMMs; perhaps just sample the corners of the hypercube.  Then let's rerun a SFARI cohort with 1) the filtered target list, 2) this HMM change, and 3) both, using optimal parameters.

samuelklee:
@asmirnov239 Evaluations can wait until they've been automated, but don't forget to test this change at some point.

ldgauthier:
@asmirnov239 is this worth going back to now that we have the SFARI and 1000G evaluations set up?

asmirnov239:
@ldgauthier Yes, definitely. I believe Jack has done some evaluations in the past that were inconclusive, but I can benchmark this branch now and see if we get improvements with certain types of CNV categories.

samuelklee:
@asmirnov239 did you ever run this with the latest version of your evaluation? If so, it might be good to record the conclusions here, and then merge/close as appropriate.

